PR #634, opened 2020-07-05
Server spans

iNikem:
With this PR all instrumentations that create SERVER spans use `HttpServerTracer` and its subclasses.

Some subclasses of `HttpServerDecorator` still exist, but they are used in instrumentation which don't create SERVER spans at all. I have left them as they are, but propose to replace them with `BaseDecorator`.


agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -56,25 +59,31 @@ public HttpServerTracer() {
...
 
-  public Span startSpan(REQUEST request, Method origin, String originType) {
+  public Span startSpan(REQUEST request, CONNECTION connection, Method origin, String originType) {
+    String spanName = spanNameForMethod(origin);
+    return startSpan(request, connection, spanName, originType);
+  }
+
+  public Span startSpan(
+      REQUEST request, CONNECTION connection, String spanName, String originType) {
     final Span.Builder builder =
```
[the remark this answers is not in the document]
iNikem: It was @trask wish to never return `null` from `startSpan` methods. And to let every instrumentation to check the existence of the server span themselves. Unfortunately :) I saw his point, so couldn't object too much.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -218,23 +234,24 @@ private void setStatus(Span span, int status) {
...
 
-  /** Stores given context in the given request in implementation specific way. */
-  protected abstract void attachContextToRequest(Context context, REQUEST request);
+  /** Stores given context in the given storage in implementation specific way. */
+  protected abstract void attachServerSpanContext(Context context, STORAGE storage);
 
   /**
-   * Returns context stored to given request by {@link #attachContextToRequest(Context, REQUEST)}.
+   * Returns context stored to the given storage by {@link #attachServerSpanContext(Context,
+   * STORAGE)}.
```
trask, line 252:
```suggestion
   * Returns context stored to the given request-response-loop storage by {@link #attachServerSpanContext(Context,
   * STORAGE)}.
```

instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/auto/instrumentation/akkahttp/AkkaHttpServerTracer.java
```diff
@@ -17,25 +17,28 @@
...
 import java.net.URI;
 import java.net.URISyntaxException;
 
-public class AkkaHttpServerDecorator
-    extends HttpServerDecorator<HttpRequest, HttpRequest, HttpResponse> {
-  public static final AkkaHttpServerDecorator DECORATE = new AkkaHttpServerDecorator();
-
-  public static final Tracer TRACER =
-      OpenTelemetry.getTracerProvider().get("io.opentelemetry.auto.akka-http-10.0");
+public class AkkaHttpServerTracer extends HttpServerTracer<HttpRequest, HttpRequest, HttpRequest> {
```
trask, line 26: to self doc that it doesn't use storage
```suggestion
public class AkkaHttpServerTracer extends HttpServerTracer<HttpRequest, HttpRequest, Void> {
```

instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/auto/instrumentation/akkahttp/AkkaHttpServerTracer.java
```diff
@@ -17,25 +17,28 @@
...
+  public static final AkkaHttpServerTracer TRACER = new AkkaHttpServerTracer();
 
   @Override
   protected String method(final HttpRequest httpRequest) {
     return httpRequest.method().value();
   }
 
+  @Override
+  protected void attachServerSpanContext(Context context, HttpRequest httpRequest) {}
```
trask, line 35:
```suggestion
  protected void attachServerSpanContext(Context context, Void storage) {}
```

instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/auto/instrumentation/akkahttp/AkkaHttpServerTracer.java
```diff
@@ -17,25 +17,28 @@
...
+
+  @Override
+  public Context getServerSpanContext(HttpRequest httpRequest) {
```
trask, line 38:
```suggestion
  public Context getServerSpanContext(Void storage) {
```

instrumentation/netty/netty-3.8/src/main/java/io/opentelemetry/auto/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
```diff
@@ -48,43 +45,25 @@ public void messageReceived(final ChannelHandlerContext ctx, final MessageEvent
...
-    final Span span = spanBuilder.startSpan();
-    try (final Scope scope = TRACER.withSpan(span)) {
-      DECORATE.afterStart(span);
-      DECORATE.onConnection(span, ctx.getChannel());
-      DECORATE.onRequest(span, request);
-
-      channelTraceContext.setServerSpan(span);
-
+    Span span = TRACER.startSpan(request, ctx.getChannel(), "netty.request", null);
+    try (final Scope ignored = TRACER.startScope(span, channelTraceContext)) {
```
trask, line 62: can merge inner try/catch into the outer try-with-resources (and i think same in v4_0 and v4_1 copies)

instrumentation/netty/netty-3.8/src/main/java/io/opentelemetry/auto/instrumentation/netty/v3_8/server/HttpServerResponseTracingHandler.java
```diff
@@ -44,26 +45,21 @@ public void writeRequested(final ChannelHandlerContext ctx, final MessageEvent m
...
+    final Context context = TRACER.getServerSpanContext(channelTraceContext);
+    if (context == null || !(msg.getMessage() instanceof HttpResponse)) {
       ctx.sendDownstream(msg);
       return;
     }
 
-    try (final Scope scope = TRACER.withSpan(span)) {
-      final HttpResponse response = (HttpResponse) msg.getMessage();
-
+    try (final Scope ignored = ContextUtils.withScopedContext(context)) {
```
trask, line 54: can be single-level try, e.g. (and same in v4_0 and v4_1 copies)
```
    Span span = TracingContextUtils.getSpan(context);
    try (final Scope ignored = ContextUtils.withScopedContext(context)) {
      ctx.write(msg, prm);
    } catch (final Throwable throwable) {
      TRACER.endExceptionally(span, throwable, 500);
      throw throwable;
    }
    TRACER.end(span, ((HttpResponse) msg).status().code());
```

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/auto/instrumentation/netty/v4_1/AttributeKeys.java
```diff
@@ -44,10 +45,10 @@
    * io.opentelemetry.auto.instrumentation.ratpack.server.TracingHandler, so if this changes, that
    * must also change.
    */
-  // TODO understand and change to context
-  public static final AttributeKey<Span> SERVER_ATTRIBUTE_KEY =
+  public static final AttributeKey<Context> SERVER_ATTRIBUTE_KEY =
       attributeKey(HttpServerTracingHandler.class.getName() + ".span");
```
trask, line 49:
```suggestion
      attributeKey(HttpServerTracingHandler.class.getName() + ".context");
```

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/auto/instrumentation/netty/v4_1/ChannelFutureListenerInstrumentation.java
```diff
@@ -69,7 +69,7 @@ public ChannelFutureListenerInstrumentation() {
       packageName + ".client.HttpClientResponseTracingHandler",
       packageName + ".client.HttpClientTracingHandler",
       // server helpers
-      packageName + ".server.NettyHttpServerDecorator",
+      packageName + ".server.NettyHttpServerTracer",
```
trask, line 72: i think all these server helpers can go away

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/auto/instrumentation/netty/v4_1/ChannelFutureListenerInstrumentation.java
```diff
@@ -27,7 +27,7 @@
 
 import com.google.auto.service.AutoService;
 import io.netty.channel.ChannelFuture;
-import io.opentelemetry.auto.instrumentation.netty.v4_1.server.NettyHttpServerDecorator;
+import io.opentelemetry.auto.instrumentation.netty.v4_1.client.NettyHttpClientDecorator;
```
trask, line 30: i think this is a good catch, it seems to me this instrumentation only applies to netty client, should we move it to the client subpackage?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -218,23 +234,24 @@ private void setStatus(Span span, int status) {
...
 
   protected abstract HttpTextFormat.Getter<REQUEST> getGetter();
 
   protected abstract URI url(REQUEST request) throws URISyntaxException;
 
   protected abstract String method(REQUEST request);
 
-  /** Stores given context in the given request in implementation specific way. */
-  protected abstract void attachContextToRequest(Context context, REQUEST request);
+  /** Stores given context in the given storage in implementation specific way. */
```
trask, line 247: trying to differentiate more between the different storages
```suggestion
  /** Stores given context in the request-response-loop storage in implementation specific way. */```

File: agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java [hunk 12]
```diff
@@ -218,23 +234,24 @@ private void setStatus(Span span, int status) {
...
+  protected abstract void attachServerSpanContext(Context context, STORAGE storage);
```
trask, line 248: i found "server span context" a bit confusing. i got used to it by the end of reviewing the PR, but maybe just "server context"?

and this method, being defined in `HttpServerTracer` could be just

```suggestion
  protected abstract void attachContext(Context context, STORAGE storage);
```

instrumentation/akka-http-10.0/src/lagomTest/groovy/LagomTest.groovy
```diff
@@ -124,7 +123,7 @@ class LagomTest extends AgentTestRunner {
     }
   }
 
-  String expectedOperationName(String method) {
-    return method != null ? "HTTP $method" : HttpServerDecorator.DEFAULT_SPAN_NAME
+  String expectedOperationName() {
+    return "akka.request"
```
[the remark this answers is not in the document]
iNikem: Main reason is that I didn't want to migrate `HttpServerDecorator#spanNameForRequest` to `HttpServerTracer`

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/auto/instrumentation/netty/v4_1/ChannelFutureListenerInstrumentation.java
iNikem: Why? From class loading perspective they don't differ from old decorator, do they?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
iNikem: I can rename `attach` method because it is protected, but `getServerSpanContext` is public, meant to be used by instrumentations. And I think there the longer, more explicit name, is better.

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/auto/instrumentation/netty/v4_1/ChannelFutureListenerInstrumentation.java
iNikem: A, I see what you mean... Let me try

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
trask: what do u think of  `attachServerContext()` and `getServerContext()`?

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/auto/instrumentation/netty/v4_1/ChannelFutureListenerInstrumentation.java
iNikem: Nope, they cannot. `ChannelFutureListenerInstrumentation` uses `AttributeKeys`, which depends on `HttpServerTracingHandler`

instrumentation/akka-http-10.0/src/lagomTest/groovy/LagomTest.groovy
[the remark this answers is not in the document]
trask: In cases where we don't have a nice route-based span name, I think we most commonly fall back to `Class(simpleName).method`

I guess in this case that would be something like `HttpExt.bindAndHandleSync` or `HttpExt.bindAndHandleAsync`, which isn't super awesome either.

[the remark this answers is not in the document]
iNikem: Yeah, that fallback works if we have some sensible class.method at hand. I don't have good ideas here

Issue #593, opened 2020-06-28
Remove HttpClientTest.READ_TIMEOUT_MS


trask:
Having a small-ish value for `HttpClientTest.CONNECT_TIMEOUT_MS` is helpful to keep the connection error tests from taking a long time (the tests that attempt to connect to `http://www.google.com:81`) .

But I'm not seeing why we need to override the default readTimeout everywhere with `HttpClientTest.READ_TIMEOUT_MS`.

malafeev:
based on submitted PR looks like removing makes tests more unstable.

malafeev:
can be closed.

iNikem:
@malafeev just in case you don't know it yet: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #635, opened 2020-07-06
Remove HttpClientTest.READ_TIMEOUT_MS

malafeev:
fix for #593

### DIFF
--- a/instrumentation/khttp-0.1/src/test/groovy/KHttpClientTest.groovy
+++ b/instrumentation/khttp-0.1/src/test/groovy/KHttpClientTest.groovy
@@ -22,7 +22,7 @@ class KHttpClientTest extends HttpClientTest {
   @Override
   int doRequest(String method, URI uri, Map<String, String> headers, Closure callback) {
     // khttp applies the same timeout for both connect and read
-    def timeoutSeconds = Math.max(CONNECT_TIMEOUT_MS, READ_TIMEOUT_MS) / 1000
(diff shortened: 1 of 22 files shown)


malafeev:
looks like tests more unstable with this PR

instrumentation/khttp-0.1/src/test/groovy/KHttpClientTest.groovy
```diff
@@ -22,7 +22,7 @@ class KHttpClientTest extends HttpClientTest {
   @Override
   int doRequest(String method, URI uri, Map<String, String> headers, Closure callback) {
     // khttp applies the same timeout for both connect and read
-    def timeoutSeconds = Math.max(CONNECT_TIMEOUT_MS, READ_TIMEOUT_MS) / 1000
+    def timeoutSeconds = CONNECT_TIMEOUT_MS / 1000
```
trask, line 25: i think this probably fine, even though `khttp` uses this for both connect and read timeouts

let's keep and if it causes issues we can bump

malafeev: @trask should I set  `READ_TIMEOUT_MS` only for `KHttpClientTest`? I mean define `READ_TIMEOUT_MS` only in `KHttpClientTest` ?

trask: yeah, that's probably wise, and self-documenting, since khttp is a bit unusual

PR #639, opened 2020-07-06
[2] Create instrumentation-core:spring:spring-webflux-5.0-core 

mabdinur:
Move core instrumentation from instrumentation:spring-webflux-5.0 to instrumentation-core.


instrumentation/spring-webflux-5.0/spring-webflux-5.0.gradle
```diff
@@ -30,8 +30,7 @@ testSets {
 }
 
 dependencies {
-  compileOnly group: 'org.springframework', name: 'spring-webflux', version: '5.0.0.RELEASE'
-  compileOnly group: 'io.projectreactor.ipc', name: 'reactor-netty', version: '0.7.0.RELEASE'
```
trask, line 34: i think it makes sense to keep these, since these classes are used directly in this module (even though they also come in as transitive dependencies now)

instrumentation/spring-webflux-5.0/spring-webflux-5.0.gradle
```diff
@@ -30,8 +30,7 @@ testSets {
...
+  implementation project(':instrumentation-core:spring:spring-webflux-5.0-core')
```
trask, line 33: check out #644 for how to shade this, so the agent and manual instrumentation don't conflict

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/WebClientTracingFilter.java
```diff
@@ -34,14 +35,24 @@
 
 public class WebClientTracingFilter implements ExchangeFilterFunction {
 
+  private final Tracer tracer;
+
+  public WebClientTracingFilter() {
+    this.tracer = TRACER;
+  }
+
```
trask, line 43: I suggest removing the no-arg constructor and passing `TRACER` in to the constructor below

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/WebClientTracingFilter.java
```diff
@@ -34,14 +35,24 @@
...
+  public WebClientTracingFilter(Tracer tracer) {
```
trask, line 44: can be `private` or at least package-private

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/WebClientTracingFilter.java
```diff
@@ -34,14 +35,24 @@
...
+    this.tracer = tracer;
+  }
+
   public static void addFilter(final List<ExchangeFilterFunction> exchangeFilterFunctions) {
```
trask, line 48: I think add an overload that takes a `Tracer` arg for users who want to provide their own `Tracer`

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/SpringWebfluxHttpClientDecorator.java
```diff
@@ -14,21 +14,18 @@
...
-import lombok.extern.slf4j.Slf4j;
 import org.springframework.web.reactive.function.client.ClientRequest;
 import org.springframework.web.reactive.function.client.ClientResponse;
 
-@Slf4j
-public class SpringWebfluxHttpClientDecorator
-    extends HttpClientDecorator<ClientRequest, ClientResponse> {
+class SpringWebfluxHttpClientDecorator extends HttpClientDecorator<ClientRequest, ClientResponse> {
 
   public static final Tracer TRACER =
```
saturnism, line 30: not familiar w/ the tracer provider and the value passed in `get`. Is it something we can inject instead? i.e., `Tracer` should prob be produced as a bean.

iNikem: Nice to meet you here, Ray! :) 

@mabdinur you have done this injection for `HttpServerTracer`, remember? Please do the same for this decorator too. Just add one more constructor here which takes an instance of `Tracer`.

mabdinur: I was tried to minimize the changes to the instrumentation code because  I wanted to push out small pull requests and to avoid introducing regressions. @Nikita I will open another pull request to clean up the injection

PR #642, opened 2020-07-07
Add jvmargs to gradle.properties to provide more RAM when building

anuraaga:
I got an out-of-heap space when working in the project in IntelliJ and it suggests raising the RAM like this. I have a feeling previously, the metaspace flag of 1G was implicitly causing Gradle to use more RAM while now it's the default, I think is just hundreds of MB.

I went with 2G arbitrarily since in other large projects I've had better experience with it than 1G and haven't needed more.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 
-org.gradle.jvmargs=-Xmx2048m
+org.gradle.jvmargs=-Xmx1024m
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,4 @@
 org.gradle.parallel=true
 org.gradle.caching=true
+
+org.gradle.jvmargs=-Xmx2048m
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,5 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 
+# Gradle default is 256m which causes issues with our build - https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 
-org.gradle.jvmargs=-Xmx1024m
+org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m


iNikem:
I think this is better put into local `~/.gradle/gradle.properties` file. It may differ from env to env, how much memory one can give to gradle.

anuraaga:
How about setting to 1g then so at least it matches the previous behavior? It's probably enough to not crash which is a good contributor experience (never had a crash before), and user can still tune it up with their home directory if they want.

iNikem:
How does it “match”? :) It was metaspace option, not it is heap :)

anuraaga:
It's just a guess, but because metaspace is part of heap I'm assuming Java defaults the heap flag to be at least as high as the metaspace option, which would make sense. It never crashed before but did right away today :)

Default is an awesome 64MB
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/gradlew#L47

iNikem:
```
 java -XX:+PrintFlagsFinal -version | grep MaxHeapSize          
   size_t MaxHeapSize                              = 4294967296                                {product} {ergonomic}
openjdk version "11.0.7" 2020-04-14 LTS
OpenJDK Runtime Environment (build 11.0.7+10-LTS)
OpenJDK 64-Bit Server VM (build 11.0.7+10-LTS, mixed mode)
```

```
 java -XX:+PrintFlagsFinal -XX:MaxMetaspaceSize=1g -version | grep MaxHeapSize
   size_t MaxHeapSize                              = 4294967296                                {product} {ergonomic}
openjdk version "11.0.7" 2020-04-14 LTS
OpenJDK Runtime Environment (build 11.0.7+10-LTS)
OpenJDK 64-Bit Server VM (build 11.0.7+10-LTS, mixed mode)
```

Metaspace does not affect max heap size.

iNikem:
> Default is an awesome 64MB
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/gradlew#L47

I expect all actual work to happen in Gradle daemons, not in the main process. Look here: https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory

iNikem:
But I have also had OOM in my Idea recently. That's interesting...

anuraaga:
Thanks for the link, I didn't realize the daemon had different defaults. It sounds like our previous flag was setting metaspace limit to effectively 512MB (heap size) instead of Gradle default of 256MB. And because the build started crashing today after that flag change, I think it's a good guess that this is a threshold for our build. How about if we just restore the metaspace flag?

trask:
Oh yes this makes sense now. Thanks @anuraaga @iNikem 👍

PR #643, opened 2020-07-07
More fixes for Gradle warnings

iNikem:
This PR renames configuration used for dependencies declarations.

PR #633, opened 2020-07-05
[WIP] Add a (trait | rule) to encapsulate test agent bootstrap logic …

anuraaga:
…to allow sharing test hierarchy between auto and non-auto instrumentation

I was examining sharing test code for auto and manual instrumentation (my first candidate is a WIP extraction of the lettuce-51 instrumentation where the test cases would be identical) and found the blocker is that spock will only find test methods (feature methods I guess is what spock calls them) in the class hierarchy. This makes some sense and reminds me of junit4 where it was strongly recommended to migrate from using base classes for behavior to rules to avoid constricting the test hierarchy.

So I tried adding a junit rule to use alternatively to `AgentTestRunner`. It's a big unwieldy though because we need both a classrule and a rule, and I don't think there's a way to do that in junit4 with a single class (jupiter has no problem but spock / sputnik doesn't work on jupiter.

So I tried a different pattern of using a Groovy trait, which spock does support for reading behavior (fixture methods). Unfortunately it has its own problem, where child classes can't define their own fixture methods using the normal names due to this.

https://stackoverflow.com/questions/56464191/public-groovy-method-must-be-public-says-the-compiler

The workaround of using a different method like `childSetup` works ok though.

The other change is with these approaches, `@RunWith` needs to be added to the test class. I'm ok with this since it's a very small amount of boilerplate and I've seen many codebases where they used a custom runner and had this on all tests, it's fairly idiomatic I guess.

Any thoughts? I lean towards the trait since codifying the groovy restrictions isn't that hard and it is a bit easier to use in the tests themselves. Not doing anything is ok too, but I don't know if we'd ever be able to share tests between auto / manual in that case.

https://github.com/spockframework/spock/issues/106

As an aside, I found several long-standing open issues in spock during my investigation of this, and including issues like that one we had with the weird closure / parent class mock, have a mostly negative impression. If using junit5 + assertj + mockito, since we're a Java project, is on the table I'm happy to explore it :) I'm also happy with not though, projects should use the framework they prefer.

Issue #597, opened 2020-06-28
Better structure for the repo to accommodate different clients' needs


trask:
Rename `agent-bootstrap` directory to `auto-bootstrap`, and rename `agent-tooling` directory to `auto-tooling`.

iNikem:
Or rename module names? :) Why auto is better than agent?

anuraaga:
I'm happy with any that make the folder / gradle task match up, including just `bootstrap` and `tooling`. The actual archive names generally aren't important during normal development so I've always found it annoying when the tasks don't match up with the folder names and I have to look up what the actual name is supposed to be in a file. Instead, using `archivesBaseName` to tweak the archives to be different from folder names has always been easier for me to use.

anuraaga:
Let me add here instead of filing a new issue for now, it might be alleviated by naming or similar.

I currently find the distinction between bootstrap and tooling to be confusing. I'd expect classes like `TracerInstaller` to be in bootstrap, and others like `ClientDecorator` to be in tooling. I feel like every time I'm looking for a class I look in the wrong one first :)

Should more of tooling be moved into bootstrap and instrumentation stuff like `ClientDecorator` moved closer to `typed`, and we rename that from `tooling` to `instrumentation-common`?

trask:
oh ya, `bootstrap` is stuff that goes in the bootstrap class loader, not stuff that is used for bootstrapping the agent

it would be nice for that to be more self-explanatory

some initial thoughts for a more descriptive name: `bootstrap-class-loader`, `bootstrap-cl`, `global`, `global-namespace`, ...

trask:
trying to think about `auto-bootstrap` vs `agent-bootstrap`...

(setting aside temporarily whether/how to rename `bootstrap` and `tooling`...)

i think `auto` goes better with the current `groupId`, e.g.

* `io.opentelemetry.instrumentation.auto:opentelemetry-javaagent`
* `io.opentelemetry.instrumentation.auto:opentelemetry-auto-bootstrap`
* `io.opentelemetry.instrumentation.auto:opentelemetry-auto-tooling`

but that matches the pattern for instrumentation:

* `io.opentelemetry.instrumentation.auto:opentelemetry-auto-X`

which doesn't seem ideal.

what about having a third `groupId` for javaagent infrastructure `io.opentelemetry.javaagent`, e.g.

* `io.opentelemetry.javaagent:opentelemetry-javaagent`
[...]

PR #644, opened 2020-07-07
Shade AWS instrumentation library when using in agent

### LINKED ISSUE #488: Shade aws-sdk from instrumentation-core when used in auto instrumentation
Otherwise those classes could conflict if manual instrumentation also on the application's class path.

Suggestion for shaded package name:

```
io.opentelemetry.auto.instrumentation.awssdk.v2_2.shaded.AwsSdk
```

anuraaga:
Fixes #488

I considered shading in the aws-sdk auto module instead but don't think it's simple to get tests to use the shaded classes in that case. Using an unpublished shaded jar generated by the manual module instead seems to work quite well.


instrumentation-core/aws-sdk/aws-sdk-2.2/aws-sdk-2.2.gradle
```diff
@@ -48,3 +53,11 @@ dependencies {
   latestDepTestCompile group: 'software.amazon.awssdk', name: 'dynamodb', version: '+'
   latestDepTestCompile group: 'software.amazon.awssdk', name: 'kinesis', version: '+'
 }
+
+shadowJar {
+  archiveClassifier.set('agent')
```
trask, line 58: is this the same?
```suggestion
  archiveClassifier = 'agent'
```

gradle/publish.gradle
```diff
@@ -10,7 +10,7 @@ publishing {
   }
   publications {
     maven(MavenPublication) {
-      if (project.tasks.findByName("shadowJar") != null) {
+      if (project.tasks.findByName("shadowJar") != null && !findProperty('noShadowPublish')) {
```
trask, line 13: 👍

iNikem:
I think some documentation is absent somewhere. About this code reuse and why this shady business is needed :)

iNikem:
Doc looks good for now, thank you.

Re "reuse": I think it is time to discuss and decide on how auto and manual/assisted instrumentations will look like. Their final package structure, how they communicate with each other etc. And then we can document all that properly in `CONTRIBUTING` doc.

anuraaga:
Agree it's a good time to decide on that. I think the conversation has gotten split over many issues and PRs like #597 #601 #633 But hopefully there's enough information now with these examples

iNikem:
@anuraaga as you have dealt with this the most, can you write a small document, a proposal of how code should be structured, what should use what etc etc? Together with motivation and any other info that could help future developers as well? I think both google doc and a new section in contribution doc are both good options.

anuraaga:
Sure will do!

anuraaga:
Will merge this after #643

PR #646, opened 2020-07-07
Merge changes from dd-trace-java 0.57.0

trask:
https://github.com/DataDog/dd-trace-java/releases/tag/v0.57.0

### DIFF
--- a/agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
+++ b/agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
@@ -89,6 +89,10 @@ public class GlobalIgnoresMatcher<T extends TypeDescription>
       }
       return true;
     }
+    // clojure
+    if (name.startsWith("clojure.") || name.contains("$fn__")) {
+      return true;
+    }
 
     if (name.startsWith("io.opentelemetry.auto.")) {
       // FIXME: We should remove this once
--- a/agent-tooling/src/test/groovy/io/opentelemetry/auto/test/ClassLoaderMatcherTest.groovy
+++ b/agent-tooling/src/test/groovy/io/opentelemetry/auto/test/ClassLoaderMatcherTest.groovy
@@ -19,6 +19,7 @@ package io.opentelemetry.auto.test
 import io.opentelemetry.auto.bootstrap.AgentClassLoader
 import io.opentelemetry.auto.tooling.ClassLoaderMatcher
 import io.opentelemetry.auto.tooling.ExporterClassLoader
+import io.opentelemetry.auto.tooling.log.LogContextScopeListener
 import io.opentelemetry.auto.util.test.AgentSpecification
 
 class ClassLoaderMatcherTest extends AgentSpecification {
@@ -59,5 +60,10 @@ class ClassLoaderMatcherTest extends AgentSpecification {
   def "ExporterClassLoader class name is hardcoded in ClassLoaderMatcher"() {
     expect:
     ExporterClassLoader.name == "io.opentelemetry.auto.tooling.ExporterClassLoader"
+  }
+
+  def "helper class names are hardcoded in Log Instrumentations"() {
+    expect:
+    LogContextScopeListener.name == "io.opentelemetry.auto.tooling.log.LogContextScopeListener"
   }
 }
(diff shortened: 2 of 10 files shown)


agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
```diff
@@ -63,110 +63,154 @@ private GlobalIgnoresMatcher(final boolean skipAdditionalLibraryMatcher) {
...
-        || name.startsWith("org.springframework.core.$Proxy")) {
-      return true;
-    }
-
-    if (COM_MCHANGE_PROXY.matcher(name).matches()) {
-      return true;
-    }
-
-    if (!skipAdditionalLibraryMatcher && additionalLibraryIgnoreMatcher.matches(target)) {
+    if (name.contains("javassist") || name.contains(".asm.")) {
```
anuraaga, line 209: Aren't these so broad they'll result in false positives? I notice it's not a new change but noticed it now, especially since I had the same question about the clojure patterns, which also seem like they have a higher chance of colliding with a user class than seems safe.

agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
```diff
@@ -63,110 +63,154 @@ private GlobalIgnoresMatcher(final boolean skipAdditionalLibraryMatcher) {
...
-      // FIXME: We should remove this once
-      // https://github.com/raphw/byte-buddy/issues/558 is fixed
-      if (name.equals(
-              "io.opentelemetry.auto.bootstrap.instrumentation.java.concurrent.RunnableWrapper")
-          || name.equals(
-              "io.opentelemetry.auto.bootstrap.instrumentation.java.concurrent.CallableWrapper")) {
-        return false;
-      }
-      return true;
+    switch (name.charAt(0) - 'a') {
```
iNikem, line 66: Can we have at least documentation for this code? _Every_ engineer that will see it will ask "WTH is going on here?!". Can some benchmark results be added as well to demonstrate that this is indeed required? 

The comment on the corresponding [upstream PR](https://github.com/DataDog/dd-trace-java/pull/1643) is confusing "This currently actually increases the startup time".

anuraaga: Heh yeah was confused by that too but assumed it's a typo or something.

 I would suggest a bar where for these sort of unreadable changes for performance we add in the comment the actual performance gain found in JMH.

`// Speeds up startup on clojure apps 5x without affecting others`.

FWIW if the comment said 10%, we probably shouldn't accept it.

richardstartin: Agreed, this code looks _horrendous_, hence the comment below (possibly understatement) but we it gets called for every loaded class and we have real reports that our instrumentation stresses clojure start up quite a lot because clojure apps load so many classes. This translated to over 25% improvement in a clojure application, which led to a meaningful outcome. I would be tempted not to take the change unless you have a clojure userbase. 

The top comment on the PR related to a matcher which could identify and exclude any clojure class, which actually led to regressions in startup and was removed in favour of this - I'll change it now. There's another comment on the PR about how it would be nice to implement this matcher with longest common prefix matches against a trie, rather than a hack like this, but there wasn't really time to do it. That's definitely we'd merge back.

iNikem: @trask what do you think about excluding this change? IMO it is a big trade-off of complexity vs performance and ideally we should do such trade-offs only with some evidence attached to these changes. So that future developers would be able to reproduce results and think of better solutions.

trask: startup overhead is a big pain point, so I'm in favor of anything that helps, even if it's only 10%. i agree we need some benchmarks, and we can definitely add more of this explanation directly in the code. let's pause on this PR and i'll get back to you

agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
richardstartin: Possibly, but they aren't new, they just moved.

agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
trask: I wrote a benchmark #768, but it wasn't clojure specific, so didn't show any improvement with this change. In any case, I removed this change, but will keep it in mind as we look at startup performance more in the future.

agent-tooling/src/main/java/io/opentelemetry/auto/tooling/matcher/GlobalIgnoresMatcher.java
trask: Opened #769 to review all global ignores

PR #647, opened 2020-07-08
[3.1] Create HttpServerTracerBase decorator 

mabdinur:
This dectorator will be used to instrument spring-webmvc-core. The existing decorator `HttpServerTracer` creates a default tracer field. I need to be able to inject my own tracer. I moved most of the logic from `HttpServerTracer` to `HttpServerTracerBase` and I created wrapper methods which use the default tracer. This way I limit the code changes required.


iNikem:
If everything from here is in #648 then please close this PR

PR #632, opened 2020-07-03
[3.3] Refactor instrumentation:spring:spring-webmvc-3.1

mabdinur:
I broke up this pull requests in to smaller changes that build on each other:
  1) [3.1] Create decorator https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/647
  2) [3.2] Create instrumentation-core:spring:spring-webmvc-3.1-core https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/648

This changes extracts code which can be used for manual instrumentation from instrumentation:spring-webmvc-3.1 to instrumentation-core:spring:spring-webmvc-3.1-core. This change includes:
- Moved HandlerMappingResourceNameFilter.java  to spring-webmvc-3.1-core
- Moved SpringWebMvcDecorator.java to spring-webmvc-3.1-core (this class will now extend the new decorator HttpServerTracingBase)
     - Override abstract methods in HttpServerTracingBase
- Update references to HandlerMappingResourceNameFilter and SpringWebMvcDecorator in instrumentation:spring-webmvc-3.1
   - Update package names in DispatcherServletInstrumentation, HandlerAdaptorInstrumentation, WebApplicationContextInstrumentation

**I am not blocked by this pull request. All this change does is move things around. I'm blocked by https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/640, https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/639, https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/647, and https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/648**

PR #648, opened 2020-07-08
[3.2] Create spring-webmvc-3.1-core project

mabdinur:
This pull requests builds on this change https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/647.

This pull request introduces a new spring-webmvc-3.1-core project 
- Move spring-webmvc:3.1 and javax dependencies from `spring-webmvc-3.1.gradle` to `spring-webmvc-3.1-core.gradle`
- Add `compileOnly project(`instrumentation-core:spring:spring-webmvc-3.1-core')` to `spring-webmvc-3.1.gradle
- Implement the HttpServerTracingBase decorator 
- Use the new decorator to instrument spring-webmvc using a spring-web filter (This filter will be used to add tracing to spring-web controllers)


instrumentation-core/spring/spring-webmvc-3.1-core/src/main/java/io/opentelemetry/instrumentation/springwebmvc/SpringWebMvcDecorator.java
```diff
@@ -0,0 +1,88 @@
...
+    return span;
+  }
+
+  public Span beforeFinish(final Span span) {
+    assert span != null;
+    return span;
+  }
+
+  @Override
+  public void onError(final Span span, final Throwable throwable) {
```
iNikem, line 45: There is no need to override and expose this method. Clients should use `end` or `endExceptionally` from the base class instead.

instrumentation-core/spring/spring-webmvc-3.1-core/src/main/java/io/opentelemetry/instrumentation/springwebmvc/WebMVCTracingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+      HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+      throws ServletException, IOException {
+    HttpServletRequest req = (HttpServletRequest) request;
+
+    Span serverSpan = DECORATE.startSpan(tracer, req, req, FILTER_METHOD, FILTER_CLASS);
+
+    try (Scope scope = tracer.withSpan(serverSpan)) {
+      filterChain.doFilter(req, response);
+    } catch (Throwable t) {
+      DECORATE.onError(serverSpan, t);
```
iNikem, line 53: Use `endExceptionally` here

instrumentation-core/spring/spring-webmvc-3.1-core/src/main/java/io/opentelemetry/instrumentation/springwebmvc/WebMVCTracingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+    this.tracer = tracer;
+  }
+
+  @Override
+  public void doFilterInternal(
+      HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+      throws ServletException, IOException {
+    HttpServletRequest req = (HttpServletRequest) request;
+
+    Span serverSpan = DECORATE.startSpan(tracer, req, req, FILTER_METHOD, FILTER_CLASS);
```
iNikem, line 48: I am not sure if my thoughts below are correct... @trask @anuraaga what do you think about it?

Among other things there are two objectives in introducing manual instrumentations in this repo.

- To produce roughly the same result as auto-instrumentation does
- To allow auto-instrumentation to reuse code from manual one, if possible.

This filter violates at least the first one. Auto-instrumentation for Spring MVC creates spans with `Span.Kind=INTERNAL`, this filter produces `Span.King=SERVER`.

instrumentation/spring-webmvc-3.1/spring-webmvc-3.1.gradle
```diff
@@ -1,31 +1,7 @@
...
-    extraDependency "javax.servlet:javax.servlet-api:3.0.1"
-  }
-}
-
 dependencies {
-  compileOnly group: 'org.springframework', name: 'spring-webmvc', version: '3.1.0.RELEASE'
-  compileOnly group: 'javax.servlet', name: 'javax.servlet-api', version: '3.1.0'
-//  compileOnly group: 'org.springframework', name: 'spring-webmvc', version: '2.5.6'
-//  compileOnly group: 'javax.servlet', name: 'servlet-api', version: '2.4'
+  compileOnly project(':instrumentation-core:spring:spring-webmvc-3.1-core')
```
iNikem, line 4: What exactly does this auto-instrumentation modules (re)-uses from that "core" module?

instrumentation-core/spring/spring-webmvc-3.1-core/spring-webmvc-3.1-core.gradle
```diff
@@ -0,0 +1,52 @@
...
+dependencies {
+  // copy-paste from instrumentation-core-aws-sdk:aws-sdk-2.2-core:
+  // TODO(anuraaga): We currently include common instrumentation logic like decorators in the
+  // bootstrap, but we need to move it out so manual instrumentation does not depend on code from
+  // the agent, like Agent.
+  api project(':auto-bootstrap')
+
+  // TODO(mabdinur):
+  // Move servlet-common to instrumentation-core
+  compileOnly project(':instrumentation:servlet:servlet-common')
```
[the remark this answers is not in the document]
mabdinur: I need this getter: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/auto/instrumentation/servlet/HttpServletRequestGetter.java

instrumentation/spring-webmvc-3.1/spring-webmvc-3.1.gradle
mabdinur: I packaged the javax.servlet-api and spring-webmvc dependencies in the core module. I thought it would be best to keep the versions in sync. I also have plans to refactor [HandlerMappingResourceNameFilter.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/spring-webmvc-3.1/src/main/java/io/opentelemetry/auto/instrumentation/springwebmvc/HandlerMappingResourceNameFilter.java) and [SpringWebMvcDecorator.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/spring-webmvc-3.1/src/main/java/io/opentelemetry/auto/instrumentation/springwebmvc/SpringWebMvcDecorator.java) and move it from the auto instrumentation module to the core module. This pull request https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/632 has these changes.

instrumentation-core/spring/spring-webmvc-3.1-core/src/main/java/io/opentelemetry/instrumentation/springwebmvc/WebMVCTracingFilter.java
mabdinur: The auto-instrumentation uses bytebuddy advice to add instrumentation to WebApplicationContexts, javax DispatchServerlets, and spring-web HandlerAdapters. I could mirror this approach by using spring aspects to keep things consistent. I thought a cleaner approach would be to use [a spring-web filter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/filter/OncePerRequestFilter.html) to auto-configure traces. I'm still figuring out the pros and cons for each approach but using filters seems to be the most direct.

instrumentation-core/spring/spring-webmvc-3.1-core/spring-webmvc-3.1-core.gradle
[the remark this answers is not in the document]
mabdinur: I wanted to keep this pull request small. I plan to extract the servlet-common project to instrumentation-core in my next pull request.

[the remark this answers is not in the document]
iNikem: Take a look at #601. Probably I have done there exactly this.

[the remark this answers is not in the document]
mabdinur: Awesome. I noticed that you moved (HttpServletRequestGetter)[https://github.com/iNikem/opentelemetry-java-instrumentation/blob/servlet-common/instrumentation-core/servlet/src/main/java/io/opentelemetry/instrumentation/servlet/HttpServletRequestGetter.java] to the core module. I'll update this dependency to map to :instrumentation-core:servlet:servlet-common when your change gets in. Keep an eye on this in case my change gets in first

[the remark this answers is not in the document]
trask: hey @mabdinur, I just merged #601 👍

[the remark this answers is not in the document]
mabdinur: @trask Thanks. I just rebased from master and updated this dependency. This should be good to go

instrumentation/spring-webmvc-3.1/spring-webmvc-3.1.gradle
mabdinur: :instrumentation-core:spring:spring-webmvc-3.1-core dependency for now

mabdinur:
> I am not sure if my thoughts below are correct... @trask @anuraaga what do you think about it?
> 
> Among other things there are two objectives in introducing manual instrumentations in this repo.
> 
> * To produce roughly the same result as auto-instrumentation does
> * To allow auto-instrumentation to reuse code from manual one, if possible.
> 
> This filter violates at least the first one. Auto-instrumentation for Spring MVC creates spans with `Span.Kind=INTERNAL`, this filter produces `Span.King=SERVER`.

@trask @iNikem Since this manual instrumentation is out of sync with the auto-instrumentation in instrumentation:spring-webmvc. Does it make more sense to move this change to https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/master/instrumentation-core/spring/spring-web-3.1 under a new package (something like io.opentelemetry.instrumentation.spring.web)?

instrumentation-core/spring/spring-webmvc-3.1/src/main/java/io/opentelemetry/instrumentation/springwebmvc/SpringWebMvcServerTracer.java
```diff
@@ -0,0 +1,80 @@
...
+import io.grpc.Context;
+import io.opentelemetry.auto.bootstrap.instrumentation.decorator.HttpServerTracer;
+import io.opentelemetry.context.propagation.HttpTextFormat.Getter;
+import io.opentelemetry.instrumentation.servlet.HttpServletRequestGetter;
+import io.opentelemetry.trace.Tracer;
+import java.net.URI;
+import java.net.URISyntaxException;
+import javax.servlet.http.HttpServletRequest;
+
+public class SpringWebMvcServerTracer
```
trask, line 28: i like the idea from an earlier review of keeping public api surface of manual instrumentation minimal 👍
```suggestion
class SpringWebMvcServerTracer
```

PR #650, opened 2020-07-08
 Separate out core reactor instrumentation 

anuraaga:
I know I owe a doc! But I mostly finished this before committing to that and was waiting on #643 before sending out.

This just copies the unit test as-is for now until we find a nicer approach, hopefully in #633 (main reason I started working on this was to give an example for it that's smaller than lettuce)


instrumentation/reactor-3.1/reactor-3.1.gradle
```diff
@@ -1,7 +1,6 @@
 // Set properties before any plugins get loaded
 ext {
   minJavaVersionForTests = JavaVersion.VERSION_1_8
-  maxJavaVersionForTests = JavaVersion.VERSION_1_8
```
trask, line 4: 👍

instrumentation-core/reactor-3.1/reactor-3.1.gradle
```diff
@@ -0,0 +1,56 @@
...
+  implementation deps.slf4j
+
+  testImplementation project(':testing')
+
+  latestDepTestImplementation group: 'io.projectreactor', name: 'reactor-core', version: '3.+'
+  // Looks like later versions on reactor need this dependency for some reason even though it is marked as optional.
+  latestDepTestImplementation group: 'io.micrometer', name: 'micrometer-core', version: '1.+'
+}
+
+shadowJar {
```
iNikem, line 50: Will we need this shadow in every manual instrumentation module? Should we create an issue to extract this to reusable function?

iNikem:
@anuraaga I think you have genuine build failure here

PR #652, opened 2020-07-09
Docs for writing new instrumentation...

anuraaga:
... and migrates AWS SDK 2.2 instrumentation to the pattern proposed in these docs.

Hope this is what you had in mind @iNikem. I tried to keep things mostly conceptual and process, but not code since it has a higher chance of drifting vs the real code. But can add more or less detail as anyone thinks makes sense!


iNikem:
Thanks! This is a good starting place to iterate :) Do you want to migrate your AWS instrumentation to the proposed structure?

anuraaga:
Sure as long as the direction seems sound - it's going to be a lot of grungy Gradle work, ideally don't have to revert it :)

iNikem:
I like the general direction. Some details need ironing our. Like the name `core`. I don’t like the current pair `core` and `auto`

anuraaga:
Cool let me play with it, just renaming later isn't a problem at all.

trask:
Yeah, this is a great doc!

And I really like putting `auto`, `core`, and `testing` together (especially the `testing` being close to both of them seems important).

Agree with @iNikem, I'm not sold on `core` either, but could get used to it if we don't find a better option.

anuraaga:
I have applied the pattern to AWS SDK

- Playing with names, I tried `library` (the library we publish for people to use for instrumentation) and `agent` (the module that loads the library for use by our agent). No strong feelings

- Temporarily named `testing` as `testing_`, there's a long-standing Gradle bug that a module importing a root project with the same name gets treated as a circular dependency and fails. I'll rename the top-level to `testing-common` or something if the approach is sane

Let me know if you have any thoughts

iNikem:
In general, I start liking it. A log of nitpicking will follow later :), but I think this is sound approach and we should follow it to the end.

Some questions:

1. What will be artefact names?
2. How can we share code between different versions of the same library? E.g. in `servlet` there are both shared tracer/extractor and "shared" instrumentations, that apply to both versions 2.2 and 3.0

anuraaga:
Thanks, think this is mostly ready for the nitpicking :)

> What will be artefact names?

I like the current names,

`yarpc-1.0/library` -> `io.opentelemetry.instrumentation:opentelemetry-yarpc-1.0`
`yarpc-1.0/agent` -> `io.opentelemetry.instrumentation.auto:opentelemetry-auto-yarpc-1.0`

> How can we share code between different versions of the same library?

Yeah I didn't really address the multi-version case yet, I'd figure

`yarpc/yarpc-common/library` -> `io.opentelemetry.instrumentation:opentelemetry-yarpc-common`
`yarpc/yarpc-1.0/library` - depends on `:yarpc:yarpc-common:library`

I can try adding to the README if this sounds ok

iNikem:
Sorry @anuraaga flooded with meetings :( Will certainly review and comment tomorrow

CONTRIBUTING.md
```diff
@@ -78,6 +78,123 @@ available in bootstrap classloader.
...
+
+In addition to automatically initializing manual instrumentation, auto instrumentation can be used
+for libraries where manual instrumentation is not possible, such as `URLConnection`, because it can
+intercept even the JDK's classes. Such libraries will not have manual instrumentation but will have
+auto instrumentation.
+
+#### Folder Structure
+
+When writing new instrumentation, create a new subfolder of `instrumentation` to correspond to the
+instrumented library and the version being targeted. Ideally an old version of the library is
```
iNikem, line 118: It is worth mentioning here "and the oldest version being target"

CONTRIBUTING.md
```diff
@@ -78,6 +78,123 @@ available in bootstrap classloader.
...
+`agent`, and `testing`.
+
+For example, if we are targeting an RPC framework `yarpc` at version `1.0` we would have a tree like
+
+```
+instrumentation ->
+    ...
+    yarpc-1.0 ->
+        agent
+            agent.gradle
```
iNikem, line 132: Gradle files names should be unique, so likely `yarpc-1.0-agent.gradle`

CONTRIBUTING.md
```diff
@@ -78,6 +78,123 @@ available in bootstrap classloader.
...
+and in the top level `settings.gradle`
+
+```groovy
+
+include 'instrumentation:yarpc-1.0:agent'
+include 'instrumentation:yarpc-1.0:library'
+include 'instrumentation:yarpc-1.0:testing'
+```
+
+#### Writing manual instrumentation
```
iNikem, line 148: We should point to an example instrumentation library right away, so that reader could see actual code behind following explanations.

CONTRIBUTING.md
```diff
@@ -78,6 +78,123 @@ available in bootstrap classloader.
...
+
+Begin by writing the instrumentation for the library in `library`. This generally involves defining a
```
iNikem, line 150: This doc is a very good starting point. But we have to write more extensive guide with code examples and, probably, method-by-method explanations. Or may be just a series of blog posts ;)

gradle/instrumentation-library.gradle
```diff
@@ -0,0 +1,29 @@
+ext {
+  noShadowPublish = true
```
iNikem, line 2: If we have naming convention for modules, as suggested above, we may replace this configuration with convention. `publish.gradle` should figure out itself that instrumentations should publish regular jars and only exporters and agent itself should publish fat jar.

gradle/instrumentation.gradle
```diff
@@ -3,6 +3,10 @@
 apply plugin: 'net.bytebuddy.byte-buddy'
 apply plugin: 'muzzle'
 
+ext {
+  packageInAgentBundle = true
```
iNikem, line 7: Same remark about module naming and convention over configuration

instrumentation/aws-sdk/aws-sdk-2.2/agent/agent.gradle
```diff
@@ -0,0 +1,36 @@
+ext {
+  minJavaVersionForTests = JavaVersion.VERSION_1_8
+}
+
```
iNikem, line 4: Do we need extra folder layer? `instrumentation/aws-sdk/aws-sdk-2.2` vs `instrumentation/aws-sdk-2.2`

instrumentation/aws-sdk/aws-sdk-2.2/library/library.gradle
```diff
@@ -0,0 +1,27 @@
...
+apply plugin: 'org.unbroken-dome.test-sets'
+
+testSets {
+  latestDepTest {
+    dirName = 'test'
+  }
+}
+
+dependencies {
+  api group: 'software.amazon.awssdk', name: 'aws-core', version: '2.2.0'
```
iNikem, line 16: should this be `compileOnly`?

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
```diff
@@ -72,15 +72,15 @@ class Aws2ClientCoreTest extends InstrumentationTestRunner {
     }
   }
 
-  def "send #operation request with builder {#builder.class.getName()} mocked response"() {
+  abstract void configureSdkClient(SdkClientBuilder builder)
+
+  def "send #operation request with builder #builder.class.getName() mocked response"() {
     setup:
+    configureSdkClient(builder)
```
iNikem, line 79: Are you sure you want to call this on every method? Not in `setupSpec`?

testing-common/src/main/groovy/io/opentelemetry/auto/test/AgentTestTrait.groovy
```diff
@@ -0,0 +1,83 @@
...
+import groovy.transform.stc.ClosureParams
+import groovy.transform.stc.SimpleType
+import io.opentelemetry.auto.test.asserts.InMemoryExporterAssert
+import io.opentelemetry.sdk.trace.data.SpanData
+
+/**
+ * A trait which initializes agent tests, including bytecode manipulation and a test span exporter.
+ * All agent tests should implement this trait.
+ */
+trait AgentTestTrait {
```
iNikem, line 29: Somewhere should be documentation about `childSetupSpec` and `childSetup` methods

testing-common/src/main/groovy/io/opentelemetry/auto/test/InstrumentationTestTrait.groovy
```diff
@@ -0,0 +1,69 @@
...
+import groovy.transform.stc.ClosureParams
+import groovy.transform.stc.SimpleType
+import io.opentelemetry.auto.test.asserts.InMemoryExporterAssert
+import io.opentelemetry.sdk.trace.data.SpanData
+
+/**
+ * A trait which initializes instrumentation library tests, including a test span exporter. All
+ * library tests should implement this trait.
+ */
+trait InstrumentationTestTrait {
```
iNikem, line 29: Why do we need 2 traits?

instrumentation/aws-sdk/aws-sdk-2.2/agent/agent.gradle
trask: The extra folder layer helps keep the `instrumentation` directory size down (e.g. in Intellij) which I think makes navigation easier

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
anuraaga: I might be misreading, but I think it's not possible here since the builder is actually created in the where clause of the spec, it's one AWS SDK but there are actually many different types of builders

testing-common/src/main/groovy/io/opentelemetry/auto/test/InstrumentationTestTrait.groovy
anuraaga: Only only initializes the test span reporter `TEST_WRITER`, the other also initializes bytecode manipulation

gradle/instrumentation-library.gradle
anuraaga: I don't know if it's worth adding that which is sort of magic, we've already abstracted it away to `instrumentation-library.gradle`. Or are you suggesting projects shouldn't have to do `apply from: "instrumentation-library.gradle"`?

In my experience with some large Gradle projects, I've had better luck having top level logic respond to small signals from the subprojects, for example this `apply from:`, or for publish I'd usually just apply `maven-publish` in the subprojects and `plugins.withId('maven-publish')` in the top-level. Seems to reduce top-level spaghetti from what I've seen. But let me know if the removal of the `apply from:` is actually your idea here.

CONTRIBUTING.md
trask: Maybe @iNikem meant this?

```
instrumentation ->
  yarpc ->
    yarpc-1.0 ->
      agent ->
        yarpc-1.0-agent.gradle
      library ->
        yarpc-1.0-library.gradle
```

and module names
* `:yarpc:yarpc-1.0:agent`
* `:yarpc:yarpc-1.0:library`

anuraaga: I think with that, we would need something like this in settings.gradle

```
project(':instrumentation:yarpc:yarpc-1.0:agent').name = 'yarpc-1.0-agent'
```

which is pretty tedious. Don't believe there is a workaround for that since it's at the initial loading of the projects.

trask: not saying it's a good idea, but we could probably add that logic to this existing code in `settings.gradle`:

```
def setBuildFile(project) {
  project.buildFileName = "${project.name}.gradle"
  project.children.each {
    setBuildFile(it)
  }
}

rootProject.children.each {
  setBuildFile(it)
}
```

gradle/instrumentation-library.gradle
iNikem: No, `apply from:` is fine, I don't object it at all.

CONTRIBUTING.md
anuraaga: I never noticed this code I could indeed tweak it. @inikem is @trask's suggested structure what you were thinking?

iNikem: Yes

gradle/instrumentation-library.gradle
anuraaga: So hopefully this is OK, as long as this detail isn't leaked into the project gradle files themselves, it seems a little nicer to have less magic here.

CONTRIBUTING.md
trask: It doesn't bother me, but I'm not well-versed in gradle convention, so nothing stands out to me as particularly off.

Just throwing out another option if you want to keep gradle file in sync with module name:

```
instrumentation ->
  yarpc ->
    v1.0 ->    (i'm just guessing using only 1.0 without the "v" prefix might be an issue?)
      yarpc-1.0-agent ->
        yarpc-1.0-agent.gradle
      yarpc-1.0-library ->
        yarpc-1.0-library.gradle
    v2.0 ->
      yarpc-2.0-agent ->
        yarpc-2.0-agent.gradle
      yarpc-2.0-library ->
        yarpc-2.0-library.gradle
```

anuraaga: I don't feel too strongly either, if both of you are ok with it than I'm ok with the current form too :)

trask:
@mabdinur check this out 🎉

Issue #536, opened 2020-06-16
Document muzzle


iNikem:
The whole story around `muzzle` is extremely confusing, but one encounters it almost everywhere in this repo. It has 0 documentation and huge number of moving parts:

- Muzzle plugin in `buildSrc` folder
- Two muzzle plugins in `auto-tooling` module
- Muzzle task and its configuration
- Muzzle task's logs that one has to read and understand when muzzle task fails
- What does it mean "instrumentation was muzzled"??
- What is muzzle's relationship with different class loaders? I sense there is one, but I am not sure.
- What is `MuzzleMatcher` and how is it connected to muzzle task and muzzle plugins?
- What does the javadoc on `io.opentelemetry.auto.tooling.Instrumenter.Default#getInstrumentationMuzzle` actually mean? And why do we generate our own code with "compile-time bytecode transformations"?
- Muzzle plugin does some extremely convoluted things with Gradle. Such as arbitrarily reaching to other projects. This makes it impossible to use Gradle's configuration on demand. That could significantly increase build time in some cases.
- Why `MuzzleVisitor` uses ASM and not ByteBuddy?

anuraaga:
+100 - getting a presentation from @trask helped me a lot but indeed there's so much muzzle around I do get re-lost.

Want to give credit for the help in working around the issues from reviewers but it's naturally not as scalable as us having a proper guide to muzzle. Based on the name, I assume it's originally datadog and I hope they can help us with documenting this to improve the contributor experience.

trask:
I will take a first stab at this, and call in @tylerbenson for backup as needed

trask:
See #654 for another muzzle thing to investigate and doc

trask:
Another interesting thing to doc: #696

trask:
Another good thing to doc: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/701#issuecomment-659814186

trask:
And another: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/694#issuecomment-659000584

trask:
Another quirk: updates to muzzle java sources in `auto-tooling` don't seem to take effect when running

```
./gradlew :instrumentation:<module>:muzzle
```

without adding `--rerun-tasks`

anuraaga:
@trask https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1121#discussion_r482686479 :)

[...]

PR #654, opened 2020-07-09
Remove a redundant configuration for Muzzle

anuraaga:
I don't see any reason for this setting to have an effect given it's overwritten in `afterEvaluate`. Unfortunately it's hard to confirm this doesn't actually slow down the build.

### DIFF
--- a/gradle/instrumentation.gradle
+++ b/gradle/instrumentation.gradle
@@ -3,14 +3,6 @@
 apply plugin: 'net.bytebuddy.byte-buddy'
 apply plugin: 'muzzle'
 
-byteBuddy {
-  transformation {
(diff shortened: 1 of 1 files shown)


anuraaga:
Apparently it does do something :-) Trying to figure out how to rephrase the comment to make more sense though since I still don't really understand what. Is it `// Make sure byte buddy transformation is not applied to test sources.`?

iNikem:
Maybe we should wait for #536 ? :)

iNikem:
@anuraaga Do we want to keep this PR open? My OCD triggers on 11 PRs in this repo :)

anuraaga:
Let's close it for now :)

PR #656, opened 2020-07-09
Kafka context propagation is configurable for consumer as well.

iNikem:
`ota.kafka.client.propagation.enabled` configuration property now governs both injecting context on the producer side and reading context on the consumer side.

### DIFF
--- a/agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java
+++ b/agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java
@@ -75,7 +75,7 @@ public class Config {
   public static final String RUNTIME_CONTEXT_FIELD_INJECTION =
       "trace.runtime.context.field.injection";
 
-  public static final String KAFKA_INJECT_HEADERS = "kafka.inject.headers";
+  public static final String KAFKA_CLIENT_PROPAGATION_ENABLED = "kafka.client.propagation.enabled";
 
   public static final String LOG_INJECTION_ENABLED = "log.injection.enabled";
   public static final String EXPERIMENTAL_LOG_CAPTURE_THRESHOLD =
@@ -95,7 +95,7 @@ public class Config {
   public static final boolean DEFAULT_LOG_INJECTION_ENABLED = false;
   public static final String DEFAULT_EXPERIMENTAL_LOG_CAPTURE_THRESHOLD = null;
 
-  public static final boolean DEFAULT_KAFKA_INJECT_HEADERS = true;
+  public static final boolean DEFAULT_KAFKA_CLIENT_PROPAGATION_ENABLED = true;
 
   private static final String DEFAULT_TRACE_ANNOTATIONS = null;
   private static final boolean DEFAULT_TRACE_EXECUTORS_ALL = false;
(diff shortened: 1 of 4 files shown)


trask:
Do you think we need this change?

Check out the setting's origin story: https://github.com/DataDog/dd-trace-java/pull/1448

I don't think there's any harm attempting to extract the header and using it if it's available.

iNikem:
I got this request internally from Splunk as part of upstreaming our codebase to Otel.

From one POV, there is probably no harm indeed. From another, if I as an user disable Kafka header propagation, I _may_ reasonable expect that my spans will NOT join remote spans. E.g. if I receive them from another system and I want my traces separate from those traces.

trask:
We don't support disabling propagation for other instrumentation.

This setting seems needed just for working around a specific issue.

Maybe if we rename the setting to something that doesn't give the wrong expectation?

iNikem:
But we already disable propagation via Kafka for _outgoing_ requests. This PR just extends this to _incoming_ requests, which makes sense to me.

Kafka header propagator is not even actual propagator in the `io.opentelemetry.context.propagation.ContextPropagators` sense. So probably renaming the configuration property is indeed sensible.

trask:
> Kafka header propagator is not even actual propagator in the `io.opentelemetry.context.propagation.ContextPropagators` sense.

I didn't follow what you mean?

iNikem:
Technically, Otel spec now defines only Http propagators: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md . Thus passing span context via Kafka headers is not specified. Thus I agree with you on renaming configuration option.

trask:
Because I ❤️ naming 😂, some ideas:

* `ota.kafka.headers.enabled`
* `ota.kafka.add.headers.enabled`
* `ota.kafka.set.headers.enabled`
* `ota.kafka.inject.headers.enabled`

I think I like the last one most(?)

anuraaga:
Getting ready for a bike ride across the park? :) Just kidding, but one idea I had from reading these is `ota.kafka.injection.enabled`, or otherwise just `ota.kafka.inject.headers` also seems nice. `inject.enabled` is verb followed by adjective so reads a bit weird.

iNikem:
My vote goes to `ota.kafka.inject.headers`. Unless somebody objects, I will go with it.

trask:
With the new name `ota.kafka.inject.headers`, it seems more clear that it only applies to injection and not extraction, do you still think we should apply it to extraction also?

iNikem:
:( As this was my original intention, yes, I still want this. I can rename config again, to something more obviously double-sided

trask:
Ok, it makes sense to me for it to be double-sided, to help prevent confusion for people who need to use the setting.

E.g. (for myself when I come back and wonder why we did this):

If you have a multiple kafka services, some running <0.11 and some running >0.11, the easiest option is to apply this setting across all nodes. But then if it's one-sided, you will see propagation sometimes (between >0.11 nodes) and not other times. And you could argue that propagation when possible is better than no propagation, but it can also lead to more questions and confusion about "missing data" and inconsistent experience.

iNikem:
So, are we back to naming game? :) We cannot use any name with `inject` in it then. And you still object to `ota.kafka.client.propagation.enabled`?

trask:
No, I don't object to `ota.kafka.client.propagation.enabled` for double-sided setting.

I was objecting previously to it being double-sided, and thought we were going down the road of renaming it to reflect it being single-sided.

But I have since changed my view and agree with your original proposal.

Thanks for bearing with me 😂.

PR #659, opened 2020-07-09
Gradle warnings

iNikem:
The majority of remaining Gradle warnings is fixed.

Remaing warnings are mostly due to Playframework gradle plugin. See https://github.com/gradle/playframework/issues/114


gradle/spotless.gradle
```diff
@@ -12,6 +12,7 @@ spotless {
   scala {
     scalafmt()
     licenseHeaderFile rootProject.file('gradle/enforcement/spotless.license.java'), '(package|import|public)'
+    target 'src/**/*.scala'
```
trask, line 15: nice catch 👍

smoke-tests/play/play.gradle
```diff
@@ -1,30 +1,20 @@
 plugins {
-  id 'play'
+  id "org.gradle.playframework" version "0.9"
```
trask, line 2: 👍

PR #662, opened 2020-07-10
Update to otel-java 0.6.0

iNikem:


agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -226,8 +226,11 @@ protected Throwable unwrapThrowable(Throwable throwable) {
   }
 
   private <C> SpanContext extract(final C carrier, final HttpTextFormat.Getter<C> getter) {
+    // Using Context.ROOT here may be quite unexpected, but the reason is simple.
+    // We want either span context extracted from the carrier or invalid one.
+    // We DO NOT want any span context potentially lingering in the current context.
```
trask, line 231: 👍

auto-exporters/logging/src/main/java/io/opentelemetry/auto/exporters/logging/LoggingExporter.java
```diff
@@ -17,10 +17,10 @@
 package io.opentelemetry.auto.exporters.logging;
 
 import io.opentelemetry.common.AttributeValue;
+import io.opentelemetry.common.ReadableKeyValuePairs.KeyValueConsumer;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.util.Collection;
-import java.util.Map;
 
 public class LoggingExporter implements SpanExporter {
```
jkwatson, line 25: Not directly relevant to this PR, but why do we have one of these here, when we have one in the main repo?

instrumentation/opentelemetry-api-beta/src/test/groovy/MeterTest.groovy
```diff
@@ -96,34 +92,34 @@ class MeterTest extends AgentTestRunner {
...
     metricData.descriptor.type == SUMMARY
-    metricData.descriptor.constantLabels == ["m": "n", "o": "p"]
+    metricData.descriptor.constantLabels == io.opentelemetry.common.Labels.of("m", "n", "o", "p")
     metricData.instrumentationLibraryInfo.name == instrumentationName
     metricData.instrumentationLibraryInfo.version == "1.2.3"
     metricData.points.size() == 1
     def point = metricData.points.iterator().next()
     if (bind) {
-      point.labels == ["w": "x", "y": "z"]
+      point.labels == io.opentelemetry.common.Labels.of("w", "x", "y", "z")
```
jkwatson, line 120: import?

auto-exporters/logging/src/main/java/io/opentelemetry/auto/exporters/logging/LoggingExporter.java
iNikem: That is a great question. Please, file a bug to us :)

instrumentation/opentelemetry-api-beta/src/test/groovy/MeterTest.groovy
iNikem: We already have an import for `unshaded.io.opentelemetry.common.Labels`, which is used for creating instruments.

auto-exporters/logging/src/main/java/io/opentelemetry/auto/exporters/logging/LoggingExporter.java
jkwatson: will do. I think yours has much cleaner output, but it's also not using a logging framework, but just using `System.out` so maybe they're not compatible?

instrumentation/opentelemetry-api-beta/src/test/groovy/MeterTest.groovy
jkwatson: oh I see. carry on, then.

Issue #660, opened 2020-07-10
Stop using Lombok


iNikem:
Just an idea. Do we really want to use Lombok? Its presence requires a support from the IDE and sometimes it breaks. As for me now in latest IntelliJ IDEA EAP.

trask:
👍 and tagged with a new tag "new contributor experience"

PR #672, opened 2020-07-12
Remove Lombok

donbeave:
Closes: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/660.

In addition to this PR also have some minor changes:

1) Upgraded Gradle to 6.5.1 (before was 6.5).
2) Added `mavenLocal` to some repositories config, otherwise, it always tries to use the remote repository.
3) By default disabled downloading javadoc/sources for IntelliJ IDEA, otherwise it really takes a lot of time. But it's not completely switched off such downloading, from IDEA still able to download javadoc/sources manually by one click.


build.gradle
```diff
@@ -16,6 +16,13 @@ plugins {
   id "com.github.spotbugs" version "4.0.1"
 }
 
+idea {
+    module {
+        downloadJavadoc = false
+        downloadSources = false
+    }
+}
+
```
trask, line 25: is this needed since also in `allprojects` block below?

agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java
```diff
@@ -329,15 +329,13 @@ public static String getSettingFromEnvironment(final String name, final String d
    * Calls {@link #getSettingFromEnvironment(String, String)} and converts the result to a list by
    * splitting on `,`.
    */
-  @NonNull
   private static List<String> getListSettingFromEnvironment(
       final String name, final String defaultValue) {
-    return parseList(getSettingFromEnvironment(name, defaultValue));
+    return Objects.requireNonNull(parseList(getSettingFromEnvironment(name, defaultValue)));
```
anuraaga, line 334: `requireNonNull` shouldn't be used with return statements, it's really for validating parameters. I understand it's to replace the usage of `@lombok.NonNull` but I think we can just delete it here, especially since they're all private methods. Best would be for us to consistently add `@Nullable` and enable `NullAway` (not for this PR).

agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java
```diff
@@ -401,7 +397,8 @@ private static String propertyNameToSystemPropertyName(final String setting) {
    * @throws NumberFormatException
    */
   private static <T> T valueOf(
-      final String value, @NonNull final Class<T> tClass, final T defaultValue) {
+      final String value, final Class<T> tClass, final T defaultValue) {
+    Objects.requireNonNull(tClass);
```
anuraaga, line 401: This is ok, though also rare to validate parameters for a private method.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/jdbc/DBInfo.java
```diff
@@ -32,5 +31,139 @@
...
+            .host(host)
+            .port(port);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
```
anuraaga, line 95: Use `!(o instanceof DbInfo)` pattern throughout

iNikem:
IDEA change - Ok.

Let us move AutoValue discussion out of this PR :)

trask:
Great. I created #676 to move the AutoValue discussion out of this PR.

build.gradle
donbeave: Actually not need, I'll remove it.

agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java
donbeave: Agree, I'll remove them.

agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java
donbeave: I'll remove it, if dig into public methods which use this variable, they already have `Objects.requireNonNull` validation.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/jdbc/DBInfo.java
donbeave: Fixed.

donbeave:
> Does this improve build performance, or does it help with something else?

@trask It improves speed downloading dependencies. If mix contributing to Maven and Gradle projects together, then this will help to reuse Maven local dependencies repo and during importing new project it will copy dependencies from local maven repo directly to Gradle cache.

> What do you think about using AutoValue for the data objects?

Agree AutoValue or Immutables will be better for such cases. I'll add my opinion in #676 thread.

iNikem:
@donbeave you have quite stable test failures in this PR. Can you confirm that those are fine on your machine?

donbeave:
@iNikem pls don't merge first, I'll check tests locally first.

iNikem:
@donbeave any luck with test verification?

donbeave:
Hi, @iNikem!

Sorry, I didn't check them yet, a little bit busy with my work now. I'll do it tomorrow or the day after tomorrow.

iNikem:
No worries :) Just checking if this PR is still active.

donbeave:
@iNikem I apologize for waiting a long time, now is ok to merge this branch.

iNikem:
@donbeave Thanks for contribution! :)

Issue #602, opened 2020-06-28
What to do about instrumentation-specific span attributes?


trask:
We currently capture lots of instrumentation-specific span attributes that are not defined by semantic conventions.

Questions:
* Should we only capture span attributes that are defined by semantic conventions?
* Should we propose instrumentation-specific span attributes (e.g. for `aws-sdk`, `elasticsearch`, etc) in an attempt to have consistency for these across languages?
* Is there naming convention we should follow for instrumentation-specific span attributes (e.g. standard prefixing like `elasticsearch.*`)?

anuraaga:
I think the reality is we'll need instrumentation-specific attributes to capture instrumentation domain-specific issues that are important for observability. Prefixing seems like a good idea (e.g., agree that the kafka instrumentation should do `kafka.tombstone` instead of just `tombstone`. Having these spec'd would be even nicer to allow them to be consistent across languages, I've already seen the x-ray instrumentation for AWS SDK have different naming strategies which results in pain like this

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/master/exporter/awsxrayexporter/translator/aws.go#L30

iNikem:
Below is my understanding of the ideal world:

* All attributes have a clear use-case explaining their existence
* All attributes are defined in semantic convention
* All attributes are documented

While point 2 from above is harder to achieve, point 1 and 3 are more under our control. So in order to bring new attribute in we have to understand the use-case and then document it.

anuraaga:
Definitely +1 to all attributes are documented, it's a clear bar to set and will naturally lend to point 1 and maybe point 2 depending on the attribute.

anuraaga:
Filed spec issue https://github.com/open-telemetry/opentelemetry-specification/issues/968

iNikem:
@anuraaga I think we can close this issue, right? This discussion is concluded and we seem to have an answer from Spec.

trask:
I think worth adding something to `docs/` since this will probably be a common question we get about adding new span attributes (assigning to myself)

iNikem:
#1874 and open-telemetry/opentelemetry-specification#968 concludes this discussion.

PR #681, opened 2020-07-13
Lettuce and Cassandra migrated from deprecated Decarator to new Tracer

iNikem:


instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
```diff
@@ -38,11 +36,9 @@
   private static final Set<CommandType> NON_INSTRUMENTING_COMMANDS = EnumSet.of(SHUTDOWN, DEBUG);
 
   public static SpanWithScope beforeCommand(final RedisCommand<?, ?, ?> command) {
-    final String spanName = command == null ? "Redis Command" : command.getType().name();
-    final Span span =
-        LettuceClientDecorator.TRACER.spanBuilder(spanName).setSpanKind(CLIENT).startSpan();
-    LettuceClientDecorator.DECORATE.afterStart(span);
-    return new SpanWithScope(span, currentContextWith(span));
+    //    final String spanName = command == null ? "Redis Command" : command.getType().name();
```
trask, line 39: remove?
```suggestion
```

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
```diff
@@ -52,45 +48,41 @@ public static void afterCommand(
...
+            if (ex == null) {
+              LettuceDatabaseClientTracer.TRACER.end(span);
             } else {
-              LettuceClientDecorator.DECORATE.onError(span, ex);
+              if (ex instanceof CancellationException) {
+                span.setAttribute("db.command.cancelled", true);
+                LettuceDatabaseClientTracer.TRACER.end(span);
+              } else {
+                LettuceDatabaseClientTracer.TRACER.endExceptionally(span, ex);
+              }
```
trask, line 63: can be pulled up into outer, e.g. `if / else if / else`

instrumentation/lettuce/lettuce-5.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v5_0/ConnectionFutureAdvice.java
```diff
@@ -16,41 +16,39 @@
...
-      @Advice.Enter final SpanWithScope spanWithScope,
       @Advice.Thrown final Throwable throwable,
-      @Advice.Return final ConnectionFuture<?> connectionFuture) {
-    final Span span = spanWithScope.getSpan();
+      @Advice.Return final ConnectionFuture<?> connectionFuture,
+      @Advice.Local("otelSpan") Span span,
+      @Advice.Local("otelScope") Scope scope) {
+    if (scope != null) {
+      scope.close();
+    }
```
trask, line 46: conditional not needed(?)
```suggestion
    scope.close();
```

instrumentation/lettuce/lettuce-5.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v5_0/LettuceAsyncCommandsAdvice.java
```diff
@@ -16,55 +16,48 @@
...
-      @Advice.Enter final SpanWithScope spanWithScope,
+      @Advice.Argument(0) final RedisCommand<?, ?, ?> command,
       @Advice.Thrown final Throwable throwable,
-      @Advice.Return final AsyncCommand<?, ?, ?> asyncCommand) {
+      @Advice.Return final AsyncCommand<?, ?, ?> asyncCommand,
+      @Advice.Local("otelSpan") Span span,
+      @Advice.Local("otelScope") Scope scope) {
+    if (scope != null) {
+      scope.close();
+    }
```
trask, line 49: conditional not needed(?)
```suggestion
    scope.close();
```

instrumentation/lettuce/lettuce-5.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v5_0/rx/LettuceFluxTerminationRunnable.java
```diff
@@ -74,25 +73,18 @@ public void accept(final Signal signal) {
 
   @Override
   public void run() {
-    if (span != null) {
-      finishSpan(true, null);
-    } else {
-      LoggerFactory.getLogger(Flux.class)
-          .error(
-              "Failed to finish this.span to indicate cancellation, LettuceFluxTerminationRunnable"
-                  + " cannot find this.span because it probably wasn't started.");
```
trask, line 83: nice catch on this duplicatation

instrumentation/lettuce/lettuce-5.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v5_0/ConnectionFutureAdvice.java
iNikem: I used the same pattern as we have in `HttpServerTracer`. And the same pattern will be used later in `HttpClientTracer`. But I can remove it here.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
```diff
@@ -40,37 +47,60 @@ public DatabaseClientTracer() {
...
-
-  private String getVersion() {
-    return null;
-  }
-
-  public Scope withSpan(Span span) {
-    return tracer.withSpan(span);
-  }
-
   public Span startSpan(CONNECTION connection, QUERY query, String originType) {
```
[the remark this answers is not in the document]
iNikem: alas, no :( Some subclasses use `String` as `QUERY` and they will have 2 methods with same signature.

instrumentation/lettuce/lettuce-5.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v5_0/ConnectionFutureAdvice.java
trask: i think remove, otherwise the reader wonders the question when can scope be `null`

trask:
Btw, this is very nice cleanup / structure coming out!

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
```diff
@@ -40,37 +47,60 @@ public DatabaseClientTracer() {
...
     String normalizedQuery = normalizeQuery(query);
 
+    return startSpan(normalizedQuery, connection, originType);
+  }
+
+  public Span startSpan(String spanName, CONNECTION connection, String originType) {
     final Span span =
         tracer
-            .spanBuilder(spanName(normalizedQuery))
+            .spanBuilder(spanName(spanName))
```
anuraaga, line 59: Shouldn't `spanName(query)` be called in the overload that accepts a query, not here?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
```diff
@@ -40,37 +47,60 @@ public DatabaseClientTracer() {
...
             .startSpan();
 
-    onConnection(span, connection);
-    onPeerConnection(span, connection);
-    onStatement(span, normalizedQuery);
+    if (connection != null) {
+      onConnection(span, connection);
+      onPeerConnection(span, connection);
+    }
+    onStatement(span, spanName);
```
anuraaga, line 69: Shouldn't `onStatement` only be called in the overload that accepts a query, `onStatement(span, normalizedQuery)`?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
```diff
@@ -87,17 +117,14 @@ protected Span afterStart(final Span span) {
...
 
   /** This should be called when the connection is being used, not when it's created. */
-  public Span onConnection(final Span span, final CONNECTION connection) {
-    assert span != null;
-    if (connection != null) {
-      span.setAttribute(SemanticAttributes.DB_USER.key(), dbUser(connection));
-      span.setAttribute(SemanticAttributes.DB_INSTANCE.key(), dbInstance(connection));
-      span.setAttribute(SemanticAttributes.DB_URL.key(), dbUrl(connection));
-    }
+  protected Span onConnection(final Span span, final CONNECTION connection) {
```
anuraaga, line 120: Just a note not necessarily this PR but I was confused by having both `onConnection` and `onPeerConnection`, I think a connection always has a peer. Based on the javadoc, maybe it's `onConnectionAttempt` and `onConnectionComplete` but not quite sure

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -84,6 +80,57 @@ public Span startSpan(
...
+  }
+
+  /** Ends given span exceptionally with default response status code 500. */
+  public void endExceptionally(Span span, Throwable throwable) {
+    endExceptionally(span, throwable, 500);
+  }
+
+  public void endExceptionally(Span span, Throwable throwable, int responseStatus) {
+    if (responseStatus == 200) {
+      // TODO I think this is wrong.
```
anuraaga, line 109: Filed #688 since seems important

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
```diff
@@ -40,37 +47,60 @@ public DatabaseClientTracer() {
...
+  }
+
+  public Span startSpan(String spanName, CONNECTION connection, String originType) {
     final Span span =
         tracer
-            .spanBuilder(spanName(normalizedQuery))
+            .spanBuilder(spanName(spanName))
             .setSpanKind(CLIENT)
             .setAttribute(SemanticAttributes.DB_TYPE.key(), dbType())
             .setAttribute("span.origin.type", originType)
```
anuraaga, line 62: I think we can be setting null here which is surprising, but same as previous code

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
```diff
@@ -38,11 +36,8 @@
   private static final Set<CommandType> NON_INSTRUMENTING_COMMANDS = EnumSet.of(SHUTDOWN, DEBUG);
 
   public static SpanWithScope beforeCommand(final RedisCommand<?, ?, ?> command) {
-    final String spanName = command == null ? "Redis Command" : command.getType().name();
-    final Span span =
-        LettuceClientDecorator.TRACER.spanBuilder(spanName).setSpanKind(CLIENT).startSpan();
-    LettuceClientDecorator.DECORATE.afterStart(span);
-    return new SpanWithScope(span, currentContextWith(span));
+    final Span span = LettuceDatabaseClientTracer.TRACER.startSpan(null, command, null);
```
anuraaga, line 39:
```suggestion
    final Span span = LettuceDatabaseClientTracer.TRACER.startSpan(/* connection= */ null, command, /* originType= */ null);
```

instrumentation/cassandra/cassandra-4.0/src/main/java/io/opentelemetry/auto/instrumentation/cassandra/v4_0/TracingCqlSession.java
```diff
@@ -232,36 +224,36 @@ public ResultSet execute(@NonNull String query) {
   @NonNull
   public CompletionStage<AsyncResultSet> executeAsync(@NonNull Statement<?> statement) {
     final String query = getQuery(statement);
-    final Span span = startSpan(query);
 
-    try (final Scope scope = currentContextWith(span)) {
+    final Span span = TRACER.startSpan(session, query, null);
```
anuraaga, line 228:
```suggestion
    final Span span = TRACER.startSpan(session, query, /* originType= */ null);
```

Though using null origin seems to be a pattern, maybe can have an overload

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
anuraaga: So many optional parameters does make this API a bit suspicious - would be good to confirm we got the right abstraction here. I sort of wonder if it's possible for the instrumentation tracer to define the API expected for the instrumentation rather than leaking the base class API over here.

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
```diff
@@ -52,45 +47,39 @@ public static void afterCommand(
...
-    // span may be finished by handleAsync call above.
   }
 
   public static SpanWithScope beforeConnect(final RedisURI redisURI) {
-    final Span span =
-        LettuceClientDecorator.TRACER.spanBuilder("CONNECT").setSpanKind(CLIENT).startSpan();
-    LettuceClientDecorator.DECORATE.afterStart(span);
-    LettuceClientDecorator.DECORATE.onConnection(span, redisURI);
-    return new SpanWithScope(span, currentContextWith(span));
+    final Span span = LettuceDatabaseClientTracer.TRACER.startSpan("CONNECT", redisURI, null);
```
anuraaga, line 72: I think we do need to solve @trask's suggestion, probably by having methods with different names. It's super-confusing to see methods used in the same file with different ordering, looked like a bug at first.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
iNikem: Hm, I will take a look...

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
iNikem: I am always confused by that as well. There will be a round of cleanup of all new tracers, when all/majority of instrumentations are migrated.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
iNikem: Setting null will remove attribute, so no harm done. But I really dislike this attribute :)

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
iNikem: Partly you are correct and I want to remove that `originType` as soon as I understand WTH is this.

Partly Lettuce instrumentations are somewhat special in that they have connection info only in some places and not another. Otherwise I still think that `startSpan(Connection, Query)` is a good API.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
trask: I will open an issue to discuss/remove `span.origin.type` 👍

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
iNikem: Agreed this is bad. But this IS `startSpan` method... I will take another look if I can improve it somehow...

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
iNikem: For me this is part of #602

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
anuraaga: This is Java, so I think it's pretty idiomatic when running into issues with overloads, instead of reordering arguments, to use `with`, e.g., `startSpanWithQuery` / `startSpanWithName`. If we expect one of these to be the vast majority, than that one can be shortened to `startSpan`. What do you think?

iNikem: I have reworked this a little, PTAL.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
iNikem: Fixed

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/DatabaseClientTracer.java
iNikem: Thanks, @anuraaga , for spotting this bug.

instrumentation/cassandra/cassandra-4.0/src/main/java/io/opentelemetry/auto/instrumentation/cassandra/v4_0/TracingCqlSession.java
iNikem: Let it leave be for a moment and we will address it in #689

iNikem:
> Just curious any reason not to knock out lettuce-5.1 here too?

5.1 does not use `DatabaseClientDecorator`

instrumentation/lettuce/lettuce-4.0/src/main/java/io/opentelemetry/auto/instrumentation/lettuce/v4_0/InstrumentationPoints.java
anuraaga: Makes sense. Still recommend the "param docs" to make this a but clearer.

instrumentation/cassandra/cassandra-3.0/src/main/java/io/opentelemetry/auto/instrumentation/cassandra/v3_0/TracingSession.java
```diff
@@ -81,78 +76,79 @@ public Session apply(final Session session) {
   }
 
   @Override
-  public ResultSet execute(final String query) {
-    final Span span = startSpan(query);
-    try (final Scope scope = currentContextWith(span)) {
+  @NonNull
```
anuraaga, line 79: Believe we don't want to add this to prevent conflict with other PR

PR #686, opened 2020-07-13
Move webflux context injection to the decorator

mabdinur:
Clean up for changes introduced in this pull request: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/639

### DIFF
--- a/instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/WebClientTracingFilter.java
+++ b/instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/WebClientTracingFilter.java
@@ -16,13 +16,10 @@
 
 package io.opentelemetry.instrumentation.springwebflux.client;
 
-import static io.opentelemetry.instrumentation.springwebflux.client.HttpHeadersInjectAdapter.SETTER;
 import static io.opentelemetry.instrumentation.springwebflux.client.SpringWebfluxHttpClientDecorator.DECORATE;
 import static io.opentelemetry.instrumentation.springwebflux.client.SpringWebfluxHttpClientDecorator.TRACER;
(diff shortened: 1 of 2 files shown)


instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/springwebflux/client/WebClientTracingFilter.java
```diff
@@ -52,29 +49,21 @@ public static void addFilter(
...
 
       return next.exchange(mutatedRequest)
           .doOnSuccessOrError(
               (clientResponse, throwable) -> {
-                if (throwable != null) {
-                  DECORATE.onError(span, throwable);
-                } else {
-                  DECORATE.onResponse(span, clientResponse);
-                }
+                DECORATE.onError(span, throwable);
```
mabdinur, line 65: There's a `throwable != null` check in DECORATE.onError(). This if statement seems redundant.

anuraaga: I think we have some deep checks like that but would prefer code to read better, rather than be minimum LoC - I'm thinking some day such a deep null check will be gone.

Here, it doesn't really sound right to call `onError` when it's not an error, so having the if check makes this code more readable than without it IMO (and if I understood him correctly, @iNikem too :) )

iNikem: Decorators usually have null checks inside methods. New Tracers usually has them outside.

But in this particular case, you have to decide either call this method or that method. Calling `onError` always assuming/hoping that it is a no-op if `throwable` is null does not seem write to me.

iNikem:
@mabdinur Can you add a motivation for this change into PR's description? As there is no associated issue it will be hard to tell later, why this change was made.

PR #565, opened 2020-06-23
Now publish all instrumentations as well.

iNikem:
Closes #135

Changed groupId to `io.opentelemetry.instrumentation`.
Removed `opentelemetry` from auto-exporters' artifactIds.
Some instrumentations will need moving around to have better names, e.g. servlets. Should I do it right away?

PR #692, opened 2020-07-14
Allows actually consume published instrumentation artifacts

iNikem:
Address concern raised in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/565#issuecomment-658094194

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -57,11 +57,14 @@ publishing {
   }
 }
 
-private String artifactPrefix(Project p){
-  if(p.name.startsWith("opentelemetry")){
+private String artifactPrefix(Project p) {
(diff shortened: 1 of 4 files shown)


gradle/publish.gradle
```diff
@@ -57,11 +57,14 @@ publishing {
...
 }
 
-private String artifactPrefix(Project p){
-  if(p.name.startsWith("opentelemetry")){
+private String artifactPrefix(Project p) {
+  if (p.name.startsWith("opentelemetry")) {
     return ''
   }
-  if(p.group == 'io.opentelemetry.instrumentation.auto'){
+  if (p.name.startsWith("auto")) {
```
anuraaga, line 64: A bit funny but makes sense

PR #693, opened 2020-07-14
Bigger timeout for release job

iNikem:


trask:
@iNikem no idea why muzzle failure only started happening, or why the failure is not reproducible locally (I couldn't repro it locally either), but I think this will fix the issue: #696

PR #694, opened 2020-07-14
Create instrumentation-core apache-httpclient-4.0

mabdinur:
Extract code that can be used in manual instrumentation from instrumentation:apache-httpclient:apache-httpclient-4.0 to instrumentation-core.

### DIFF
--- a/instrumentation-core/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_0/ApacheHttpClientDecorator.java
+++ b/instrumentation-core/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_0/ApacheHttpClientDecorator.java
@@ -19,18 +19,14 @@ package io.opentelemetry.instrumentation.apachehttpclient.v4_0;
 import io.grpc.Context;
 import io.opentelemetry.OpenTelemetry;
 import io.opentelemetry.auto.bootstrap.instrumentation.decorator.HttpClientDecorator;
-import io.opentelemetry.trace.Tracer;
 import java.net.URI;
 import org.apache.http.Header;
 import org.apache.http.HttpMessage;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpUriRequest;
 
-class ApacheHttpClientDecorator extends HttpClientDecorator<HttpUriRequest, HttpResponse> {
+public class ApacheHttpClientDecorator extends HttpClientDecorator<HttpUriRequest, HttpResponse> {
   public static final ApacheHttpClientDecorator DECORATE = new ApacheHttpClientDecorator();
-
-  public static final Tracer TRACER =
-      OpenTelemetry.getTracerProvider().get("io.opentelemetry.auto.apache-httpclient-4.0");
 
   public void inject(Context context, HttpUriRequest request) {
     OpenTelemetry.getPropagators()
--- a/instrumentation-core/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_0/ApacheHttpClientDecorator.java
+++ b/instrumentation-core/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_0/ApacheHttpClientDecorator.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.auto.instrumentation.apachehttpclient.v4_0;
+package io.opentelemetry.instrumentation.apachehttpclient.v4_0;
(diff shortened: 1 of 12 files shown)


instrumentation-core/spring/spring-webflux-5.0/spring-webflux-5.0.gradle
```diff
@@ -9,7 +9,7 @@ ext {
 
 apply from: "$rootDir/gradle/java.gradle"
 
-group = 'io.opentelemetry.instrumentation.spring.webflux'
+group = 'io.opentelemetry.instrumentation.spring'
```
trask, line 12: I think let's not create maven subgroups, and put manual instrumentation directly in `io.opentelemetry.instrumentation` (which matches what we are doing in `io.opentelemetry.instrumentation.auto`)
```suggestion
group = 'io.opentelemetry.instrumentation'
```

mabdinur: I'll make this change in the spring-web and spring-webmvc projects as well

mabdinur:
Muzzle is failing on last commit.

Muzzle passed for all commits except the last one (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/694/commits/7d02d9dbff4d28a34dcb8910fbb6b54dee5100e5). I did not change any gradle files in this commit so I'm not sure why `:instrumentation:apache-httpclient:apache-httpclient-4.0:muzzle-AssertPass` tests are failing. Has anyone seen this issue before?

iNikem:
Make your `ApacheHttpClientDecorator` public. It is called directly from instrumented class, so should be reachable from library code.

trask:
Hey @mabdinur! The `muzzle` issue is because of the static `TRACER` field in `ApacheHttpClientInstrumentation`.

The references to that `TRACER` field then pull in the whole of `ApacheHttpClientInstrumentation`, which then pulls in those bytebuddy classes. This is a common issue I need to doc in #536.

I suggest having overloaded methods in `ApacheHttpClientHelper` that either take a `Tracer` or use the default one, e.g.

```
public static SpanWithScope doMethodEnter(HttpUriRequest request, Tracer tracer)
```

and

```
public static SpanWithScope doMethodEnter(HttpUriRequest request)
```

This will be useful for people using manual instrumentation, and also can be used then from auto instrumentation and you can remove the `TRACER` field from the auto instrumentation.

instrumentation-core/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_0/ApacheHttpClientDecorator.java
```diff
@@ -19,19 +19,15 @@
...
 import io.opentelemetry.auto.bootstrap.instrumentation.decorator.HttpClientDecorator;
-import io.opentelemetry.trace.Tracer;
 import java.net.URI;
 import org.apache.http.Header;
 import org.apache.http.HttpMessage;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpUriRequest;
 
-class ApacheHttpClientDecorator extends HttpClientDecorator<HttpUriRequest, HttpResponse> {
+public class ApacheHttpClientDecorator extends HttpClientDecorator<HttpUriRequest, HttpResponse> {
```
trask, line 28: Can you try again with this package-private, now that `muzzle` is fixed? Better not to expose this to manual instrumenters if possible.
```suggestion
class ApacheHttpClientDecorator extends HttpClientDecorator<HttpUriRequest, HttpResponse> {
```

instrumentation-core/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_0/HttpHeadersInjectAdapter.java
```diff
@@ -19,7 +19,7 @@
 import io.opentelemetry.context.propagation.HttpTextFormat;
 import org.apache.http.client.methods.HttpUriRequest;
 
-class HttpHeadersInjectAdapter implements HttpTextFormat.Setter<HttpUriRequest> {
+public class HttpHeadersInjectAdapter implements HttpTextFormat.Setter<HttpUriRequest> {
```
trask: same here
```suggestion
class HttpHeadersInjectAdapter implements HttpTextFormat.Setter<HttpUriRequest> {
```

mabdinur: Done

PR #695, opened 2020-07-14
Bigger timeout for release job

trask:
Same as #693, troubleshooting muzzle failure.


trask:
Done troubleshooting. Opened #696 with the fix.

PR #696, opened 2020-07-14
Fix muzzle failure

trask:
No idea why `muzzle` only started failing, but it looks like the failure is because `JettyHttpServerTracer` extends `Servlet3HttpServerTracer`, but due to the ordering in `helperClassNames()`, `JettyHttpServerTracer` is added to the class loader first, before `Servlet3HttpServerTracer` (it's super class) has been added.

### DIFF
--- a/instrumentation/jetty-8.0/src/main/java/io/opentelemetry/auto/instrumentation/jetty/JettyHandlerInstrumentation.java
+++ b/instrumentation/jetty-8.0/src/main/java/io/opentelemetry/auto/instrumentation/jetty/JettyHandlerInstrumentation.java
@@ -52,6 +52,8 @@ public final class JettyHandlerInstrumentation extends Instrumenter.Default {
 
   @Override
   public String[] helperClassNames() {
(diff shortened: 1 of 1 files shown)


instrumentation/jetty-8.0/src/main/java/io/opentelemetry/auto/instrumentation/jetty/JettyHandlerInstrumentation.java
```diff
@@ -53,11 +53,11 @@ public JettyHandlerInstrumentation() {
   @Override
   public String[] helperClassNames() {
     return new String[] {
-      packageName + ".JettyHttpServerTracer",
+      "io.opentelemetry.instrumentation.servlet.HttpServletRequestGetter",
```
anuraaga, line 56: Can you add a comment, at least "Order matters here"?

trask: yes, excellent point, done

PR #701, opened 2020-07-15
Remove http server decorator

iNikem:
Last subclasses of `HttpServerDecorator` are gone. These instrumentation used it, but in fact produced `INTERNAL` spans. This PR removes from them those semantic attributes which belongs to `SERVER` spans as was briefly discussed on Gitter.


iNikem:
@trask I would like your opinion here. Ratpack instrumentation muzzle config is this:
```
    versions = "[1.5.0,)"
    skipVersions += ["0.9.12", "0.9.13", "0.9.14",]
    assertInverse = true
```
Thus it expect muzzle to fail on version prior to 1.5. This PR removed a lot of code from `RatpackServerDecorator`, including 
```
  protected String peerHostIP(final Request request) {
    return request.getRemoteAddress().getHost();
  }
```
which, apparently, allows muzzle to pass on versions 1.4.x. Thus muzzle is now failing with `MUZZLE PASSED ServerErrorHandlerInstrumentation BUT FAILURE WAS EXPECTED`.

Why do we insist that muzzle should fail on versions prior to 1.5.0? Can I remove `assertInverse = true` from muzzle configuration for Ratpack?

trask:
> Why do we insist that muzzle should fail on versions prior to 1.5.0?

I think of `assertInverse = true` as just a nice extra check that we have the right baseline version, and that it won't match something earlier that we weren't expecting. But it's not required (and not used everywhere). cc: @tylerbenson in case he has different thoughts

> Can I remove `assertInverse = true` from muzzle configuration for Ratpack?

Ideally, we would lower the baseline (e.g. to 1.4.0) and lower the tested version (assuming it passes the lower version) if that works out. And then drop the module name to `ratpack-1.4`.

If that doesn't work out, or is a pain, then yeah, that's fine.

iNikem:
Interesting... Muzzle configuration `versions = "[1.5.0,)"` fails with
```
> Task :instrumentation:ratpack-1.5:muzzle-AssertFail-io.ratpack-ratpack-core-1.4.0 FAILED
MUZZLE PASSED ServerErrorHandlerInstrumentation BUT FAILURE WAS EXPECTED
```
Muzzle configuration `versions = "[1.4.0,)"` fails with
```
> Task :instrumentation:ratpack-1.5:muzzle-AssertPass-io.ratpack-ratpack-core-1.4.0 FAILED
FAILED MUZZLE VALIDATION: io.opentelemetry.auto.instrumentation.ratpack.ContinuationInstrumentation mismatches:
-- io.opentelemetry.auto.instrumentation.ratpack.ContinuationInstrumentation$ResumeAdvice:80 Missing method getHost#()Ljava/lang/String;
```
Is there a bug in muzzle or?

iNikem:
I have decided to remove `assertInverse = true` for the time being.

instrumentation/play/play-2.3/src/main/java/io/opentelemetry/auto/instrumentation/play/v2_3/PlayAdvice.java
```diff
@@ -37,17 +36,17 @@
   public static SpanWithScope onEnter(@Advice.Argument(0) final Request req) {
     final Span.Builder spanBuilder = TRACER.spanBuilder("play.request");
     if (!TRACER.getCurrentSpan().getContext().isValid()) {
+      // TODO should we do that for INTERNAL span?
       final SpanContext extractedContext = extract(req.headers(), GETTER);
       if (extractedContext.isValid()) {
         spanBuilder.setParent(extractedContext);
       }
```
trask, line 43: I agree with your TODO, delete 😄
```suggestion
```

instrumentation/play/play-2.3/src/main/java/io/opentelemetry/auto/instrumentation/play/v2_3/PlayAdvice.java
```diff
@@ -37,17 +36,17 @@
...
-    } else {
-      // An upstream framework (e.g. akka-http, netty) has already started the span.
-      // Do not extract the context.
     }
+    // An upstream framework (e.g. akka-http, netty) has already started the span.
+    // Do not extract the context.
+
```
trask, line 47: and delete
```suggestion
```

instrumentation/play/play-2.4/src/main/java/io/opentelemetry/auto/instrumentation/play/v2_4/PlayAdvice.java
```diff
@@ -40,13 +40,11 @@ public static SpanWithScope onEnter(@Advice.Argument(0) final Request req) {
       if (extractedContext.isValid()) {
         spanBuilder.setParent(extractedContext);
       }
-    } else {
-      // An upstream framework (e.g. akka-http, netty) has already started the span.
-      // Do not extract the context.
     }
+    // An upstream framework (e.g. akka-http, netty) has already started the span.
+    // Do not extract the context.
```
trask, line 45: delete `!isValid` condition above, and delete this comment ⚡
```suggestion
```

instrumentation/play/play-2.6/src/main/java/io/opentelemetry/auto/instrumentation/play/v2_6/PlayAdvice.java
```diff
@@ -40,13 +40,12 @@ public static SpanWithScope onEnter(@Advice.Argument(0) final Request req) {
       if (extractedContext.isValid()) {
         spanBuilder.setParent(extractedContext);
       }
-    } else {
-      // An upstream framework (e.g. akka-http, netty) has already started the span.
-      // Do not extract the context.
     }
+    // An upstream framework (e.g. akka-http, netty) has already started the span.
+    // Do not extract the context.
+
```
trask, line 46: and again, let's delete condition above, and this comment 👍
```suggestion
```

testing/src/main/groovy/io/opentelemetry/auto/test/utils/ConfigUtils.groovy
```diff
@@ -17,10 +17,9 @@
 package io.opentelemetry.auto.test.utils
 
 import io.opentelemetry.auto.config.Config
-import lombok.SneakyThrows
-
 import java.lang.reflect.Modifier
 import java.util.concurrent.Callable
+import lombok.SneakyThrows
```
trask, line 22: Can you check your Intellij import settings for groovy?

> With java use the following import layout (groovy should still use the default)

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/CONTRIBUTING.md#intellij-idea

trask:
> Is there a bug in muzzle or?

That message is saying:

`ServerErrorHandlerInstrumentation` is compatible with, and will be applied when running version `1.4.0`

while `ContinuationInstrumentation` is not compatible with, and will not be applied when running version `1.4.0`

This is not ideal, because that means if you are using ratpack `1.4.0`, we will apply some instrumentation but not others, which could lead to odd behavior (and possibly app failure if we are careless in how those two instrumentations happen to depend on each other).

Btw, I just noticed this in `ContinuationInstrumentation`:

```
    public void muzzleCheck(final PathBinding binding, final HostAndPort host) {
      // This was added in 1.4.  Added here to ensure consistency with other instrumentation.
      binding.getDescription();

      // This is available in Guava 20 which was required starting in 1.5
      host.getHost();
    }
```

This is one approach to keeping all the instrumentation within a given module in sync with each other version-wise. Search the codebase for other methods called `muzzleCheck`. These methods are never called (and don't need to be named that, it's just convention), but their presence in their respective Advice adds the methods inside of them to the muzzle verification process and manipulates the version of the instrumented library that will be matched.

In this case, it looks like removing the call to `host.getHost()` may drop the version down to `1.4` consistently across all instrumentation.

testing/src/main/groovy/io/opentelemetry/auto/test/utils/ConfigUtils.groovy
iNikem: What do you mean? Please don't tell me you want to have `java.*` import separately for groovy

trask: I'm just following the instructions from `CONTRIBUTING.md`:

image

trask: I don't think anyone would objection if you want to change the groovy import ordering to match the java import ordering

iNikem:
Great, now, after removing SpanContext extraction from wire, play instrumentations for 2.3 and 2.4 work perfectly on all later versions as well :)

```
> Task :instrumentation:play:play-2.3:muzzle-AssertFail-com.typesafe.play-play_2.11-2.6.17 FAILED
MUZZLE PASSED PlayInstrumentation BUT FAILURE WAS EXPECTED
```

It seems that 2.3 and 2.4 can be merged together, but 2.6 instrumentation is somewhat different (see `PlayDecorator`). Will see what can I do...

iNikem:
Configured `muzzleCheck` to preserver current play 2.3, 2.4 and 2.6 separation. Will file a separate ticket to merge some of them.

Ratpack version downgraded to 1.4

PR #705, opened 2020-07-15
Make sure that filter chain always runs

trask:
`filterChain.doFilter(request, response)` was accidentally nested inside of a conditional.

cc: @tylerbenson, you may want this in dd-trace-java


iNikem:
No test failures nor changes?

trask:
Generally that condition is always true, unless something bad happens, which I managed to make happen in my fork. Not sure how to repro the issue in this repository, but I think the code should be fixed anyways.

tylerbenson:
@trask thanks for the heads up!

PR #707, opened 2020-07-15
Minor update to the contributing doc

heyams:


CONTRIBUTING.md
```diff
@@ -86,10 +86,23 @@ snapshot builds of the `master` branch. They are available from
...
+
+```bash
+./gradlew assemble
+```
+
+and then generate the -all artifact
+
+```bash
+./gradlew :opentelemetry-javaagent:shadowJar
+```
```
trask, line 101: Good catch! I opened #708 to see about generating the `-all` artifact as part of `assemble` in the future, but good to update docs now to reflect the current state.

CONTRIBUTING.md
```diff
@@ -143,6 +156,12 @@ Our build will fail if source code is not formatted according to that style.
 The main goal is to avoid extensive reformatting caused by different IDEs having different opinion
 about how things should be formatted by establishing.
 
+For Windows users, please set autocrlf to true and line endings will be converted to Unix style line endings when adding files to the repo:
+
+```bash
+git config --global core.autocrlf true
```
trask, line 162: I like to set it globally, but maybe we should just recommend setting it locally
```suggestion
git config core.autocrlf true
```

CONTRIBUTING.md
```diff
@@ -143,6 +156,12 @@ Our build will fail if source code is not formatted according to that style.
 The main goal is to avoid extensive reformatting caused by different IDEs having different opinion
 about how things should be formatted by establishing.
 
+For Windows users, please set autocrlf to true and line endings will be converted to Unix style line endings when adding files to the repo:
```
trask, line 159:
```suggestion
Note for Windows users: if you are seeing all of your files being changed by `./gradlew spotlessApply` (below), it's probably the line endings, and make sure you are using a `core.autocrlf` setting of either `true` or `input`.
```

CONTRIBUTING.md
```diff
@@ -143,6 +156,12 @@ Our build will fail if source code is not formatted according to that style.
 The main goal is to avoid extensive reformatting caused by different IDEs having different opinion
 about how things should be formatted by establishing.
 
+Note for Windows users: if you are seeing all of your files being changed by `./gradlew spotlessApply` (below), it's probably the line endings, and make sure you are using a `core.autocrlf` setting of either `true` or `input`.
+
+```bash
+git config core.autocrlf true
```
anuraaga, line 162: @trask What do you think about adding `.gitattributes` with something like this?

```
* text=auto eol=lf

*.bat text eol=crlf
*.cmd text eol=crlf
```

I often develop on Windows, and I have only had pain with repositories without gitattributes and have never seen problems after adding it (modern OS / tools handle LF just fine on Windows too, even Notepad!).

trask: oh, yes, that does work, as spotless respects the line endings in `.gitattributes`! (https://github.com/diffplug/spotless/issues/23)

trask: @heyams let's remove this section and add the `.gitattributes` file suggested by @anuraaga above 👍

PR #712, opened 2020-07-16
Remove attribute `span.origin.type`

### LINKED ISSUE #689: Remove span.origin.type attribute
I'm not seeing that it provides much value over other attributes which are already captured:

* `org.h2.jdbc.JdbcPreparedStatement` (`db.url` already tells us we are connecting to `h2`, though we don't capture whether it's a prepared statement)
* `forwardToCompileError_jsp` (`spanName` already has this info: `Render /forwards/forwardToCompileError.jsp`)
* `org.apache.catalina.core.ApplicationFilterChain` (not too interesting)
* `org.apache.activemq.ActiveMQMessageConsumer` (this is sort of interesting, we don't capture currently that we are connecting to `activemq`)

Also, I think all general-purpose attributes (like this) should be defined by the semantic conventions.

If someone wants to capture one or more of these attribute values in some instrumentation-specific attribute(s) that would seem ok to me (see ongoing discussing of instrumentation-specific attributes in #602).

But I would suggest we remove this general-purpose attribute.

iNikem:
Closes #689


iNikem:
> for deleting code!

That is literally my favourite part :)

Issue #529, opened 2020-06-15
Sporadic test failure (3x): context.FieldBackedProviderFieldInjectionDisabledTest


jkwatson:
```
java.lang.AssertionError: 3 Instrumentation errors during test
	at io.opentelemetry.auto.test.AgentTestRunner.agentCleanup(AgentTestRunner.java:214)
	at org.spockframework.util.ReflectionUtil.invokeMethod(ReflectionUtil.java:200)
	at org.spockframework.runtime.model.MethodInfo.invoke(MethodInfo.java:113)
	at io.opentelemetry.auto.test.SpockRunner.run(SpockRunner.java:158)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:110)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:38)
	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.processTestClass(TestWorker.java:118)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedObjectConnection$DispatchWrapper.dispatch(MessageHubBackedObjectConnection.java:182)
[...]

Issue #619, opened 2020-07-01
Sporadic test failure (4x) testDisabledFieldInjection


anuraaga:
```
FAILURE: Build failed with an exception.



* What went wrong:

Execution failed for task ':testing:testDisabledFieldInjection'.
```

```
    classMethod - context.FieldBackedProviderFieldInjectionDisabledTest

    java.lang.AssertionError: 3 Instrumentation errors during test
    	at io.opentelemetry.auto.test.AgentTestRunner.agentCleanup(AgentTestRunner.java:215)
    	at org.spockframework.util.ReflectionUtil.invokeMethod(ReflectionUtil.java:200)
    	at org.spockframework.runtime.model.MethodInfo.invoke(MethodInfo.java:113)
    	at io.opentelemetry.auto.test.SpockRunner.run(SpockRunner.java:159)
    	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.runTestClass(JUnitTestClassExecutor.java:110)
    	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:58)
    	at org.gradle.api.internal.tasks.testing.junit.JUnitTestClassExecutor.execute(JUnitTestClassExecutor.java:38)
    	at org.gradle.api.internal.tasks.testing.junit.AbstractJUnitTestClassProcessor.processTestClass(AbstractJUnitTestClassProcessor.java:62)
    	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.processTestClass(SuiteTestClassProcessor.java:51)
    	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
    	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
    	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
    	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
    	at org.gradle.api.internal.tasks.testing.worker.TestWorker.processTestClass(TestWorker.java:119)
    	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
[...]

PR #725, opened 2020-07-17
Faster tests

iNikem:
This PR tries to speed up test runs. I have compared it to master branch while running 3 gradle workers and 1. I used `./gradlew clean test --no-build-cache`, running Liberica jdk 11.0.7, on quad-core i7 iMac, macOs Catalina.

1 worker
master branch: 27m 44s
this PR: 8m 12s

3 workers:
master branch: 16m 56s
this PR: 5m 1s

Three main things were done here.

First, `test.forkEvery=1` was removed. Before we spawned a new JVM for every test. Now Gradle spawns new jvm for every  instrumentation, because they all have separate classpath. All tests for a given module thus run in one JVM.

Second, `AgentTestRunner` was changed so that it installs bytebuddy agent and all our instrumentations only once. If test changes configuration, that resets BB agent and forces it to reinstall itself.

Third, tests requiring bytecode modifications got separated from all others into junit category of its own, because they don't mix well in the same JVM.


gradle/instrumentation.gradle
```diff
@@ -54,3 +51,21 @@ afterEvaluate {
...
   }
 }
+
+tasks.withType(Test) {
+  useJUnit {
+    excludeCategories 'io.opentelemetry.auto.test.BytecodeTests'
+  }
+}
+
+task integrationTest(type: Test) {
```
anuraaga, line 61: Looks like we need to add `-x integrationTest` as well to circleci. Though probably better to define a property to handle that with `withType(Test)` globally, maybe `-PskipTests`.

iNikem: Yeah, as usually I forget about CI. Have to review all jobs there

trask:
Wow 🔥

gradle/instrumentation.gradle
anuraaga: Found that `-PskipTests` is already defined :-)
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/gradle/java.gradle#L279

iNikem:
Please wait till I solve all CI problems :( Probably will need another round of reviews then...

iNikem:
I have all tests succeed locally, but failing in CI. Still don't understand why. And tomorrow I will have to deal with some employer work :(

.circleci/config.yml
```diff
@@ -201,11 +201,8 @@ workflows:
       - default_test_job:
           requires:
             - build
-          prefixTestTask: true
-          name: test_<< matrix.testTask >>
-          matrix:
-            parameters:
-              testTask: ["8"]
```
trask, line 208: Do you mind keeping the parameterization here? Really hoping to bring back testing on Java 11 and Java 14 soon

gradlew.bat
```diff
@@ -1,104 +1,104 @@
...
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@rem
```
trask, line 1: thanks 😄, i noticed git keeps wanting to change line endings on this file now that we added the `.gitattributes`

instrumentation/jms-1.1/src/latestDepTest/groovy/JMS2Test.groovy
```diff
@@ -45,6 +46,13 @@ import static io.opentelemetry.trace.Span.Kind.CONSUMER
 import static io.opentelemetry.trace.Span.Kind.PRODUCER
 
 class JMS2Test extends AgentTestRunner {
+  static {
+    ConfigUtils.updateConfig {
+      System.setProperty("ota.trace.classes.exclude", "org.springframework.jms.config.JmsListenerEndpointRegistry\$AggregatingCallback,org.springframework.context.support.DefaultLifecycleProcessor\$1")
```
trask, line 51: This seems surprising. I thought `ota.trace.classes.exclude` only applied to annotation instrumentation?

testing-common/src/main/groovy/io/opentelemetry/auto/test/AgentTestRunner.java
```diff
@@ -68,17 +68,17 @@
...
  * </ul>
  */
+@Category(BytecodeTests.class)
 @RunWith(SpockRunner.class)
 @SpecMetadata(filename = "AgentTestRunner.java", line = 0)
 public abstract class AgentTestRunner extends AgentSpecification {
 
   private static final org.slf4j.Logger log = LoggerFactory.getLogger(AgentTestRunner.class);
 
-  private static final long TIMEOUT_MILLIS = 10 * 1000;
```
trask, line 81: Yes! I tried to remove this before but ended up with totally inexplicable test failures even though it's not used anywhere

anuraaga:
Can you explain https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/725/commits/41a30d9a991b8a210cdcdb21baa3db011f25ff83 ? I'm worried we're flakier than before and just were able to get through CI by reducing the probability with this flag. But not sure, probably missing something :)

gradlew.bat
anuraaga: Can you confirm the new version is using CRLF?

.circleci/config.yml
iNikem: I hope to switch to GitHub Actions sooner :)

iNikem:
> Can you explain [41a30d9](https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/41a30d9a991b8a210cdcdb21baa3db011f25ff83) ? I'm worried we're flakier than before and just were able to get through CI by reducing the probability with this flag. But not sure, probably missing something :)

My local machine has 4 cores (8 with HT) and 3 Gradle workers comfortably saturate it while running tests without excessive contention. 4 workers did not give any benefits.
CI machine has 8 cores, probably 16 with HT. And we run 3 jobs in parallel. So before this change we had 18 workers. I suspect it was too much and it overwhelmed build server.

instrumentation/jms-1.1/src/latestDepTest/groovy/JMS2Test.groovy
iNikem: It is used in `io.opentelemetry.auto.tooling.AgentInstaller#matchesConfiguredExcludes` so applies to everything.

gradlew.bat
iNikem: Nope, it uses LF now. Will convert to CRLF.

iNikem: Hm... converted to CRLF, ran `git add` and now I see `nothing to commit, working tree clean`

anuraaga: Yeah it means that the checked in version is already CRLF which is correct, so that seems OK. One of the problems with changing gitattributes is it works great for new clones, but not so much for existing ones, so I often just reclone after a gitattributes change (there is other magic that might work that is much more complex :) ).

trask:
> we run 3 jobs in parallel

Do you mean `test8`, `muzzle`, and `test_latest`? I think each of those get their own hardware (we used to run ~7 parallel jobs).

instrumentation/jms-1.1/src/latestDepTest/groovy/JMS2Test.groovy
trask: Oh, I didn't realize that. Still curious why this is needed? If related to your [gitter question](https://gitter.im/open-telemetry/opentelemetry-java-instrumentation?at=5f155d76a874ff08d288cdb5), maybe these classes need to be added to `AdditionalLibraryIgnoresMatcher`?

iNikem:
Hm... then I don't know what happened with this PR and why reducing workers count let it pass immediately... What do you think, should I investigate tests behaviour under load in this PR or to do it separately? After this is merged.

trask:
> What do you think, should I investigate tests behaviour under load in this PR or to do it separately? After this is merged.

I'm going to hit re-run on CI a few times for this PR.

I looked at one of the prior failures, and will investigate a bit.

https://13910-210933087-gh.circle-artifacts.com/0/reports/instrumentation/hibernate/hibernate-4.3/tests/integrationTest/classes/SpringJpaTest.html

```
01:55:42.224 [Test worker] DEBUG io.opentelemetry.auto.tooling.bytebuddy.matcher.LoggingFailSafeMatcher - Instrumentation type matcher unexpected exception: io.opentelemetry.auto.instrumentation.hibernate.v4_3.ProcedureCallInstrumentation
java.lang.IllegalStateException: Cannot resolve type description for spring.jpa.Customer_$$_jvst56_0
	at net.bytebuddy.pool.TypePool$Resolution$Illegal.resolve(TypePool.java:157)
	at io.opentelemetry.auto.tooling.bytebuddy.AgentCachingPoolStrategy$CachingResolution.resolve(AgentCachingPoolStrategy.java:285)
	at net.bytebuddy.pool.TypePool$Default$WithLazyResolution$LazyTypeDescription.delegate(TypePool.java:912)
	at net.bytebuddy.description.type.TypeDescription$AbstractBase$OfSimpleType$WithDelegation.getModifiers(TypeDescription.java:8331)
	at net.bytebuddy.matcher.ModifierMatcher.matches(ModifierMatcher.java:48)
	at net.bytebuddy.matcher.ModifierMatcher.matches(ModifierMatcher.java:27)
	at net.bytebuddy.matcher.NegatingMatcher.matches(NegatingMatcher.java:46)
	at net.bytebuddy.matcher.ElementMatcher$Junction$Conjunction.matches(ElementMatcher.java:122)
	at io.opentelemetry.auto.tooling.bytebuddy.matcher.LoggingFailSafeMatcher.matches(LoggingFailSafeMatcher.java:62)
	at net.bytebuddy.agent.builder.AgentBuilder$RawMatcher$ForElementMatchers.matches(AgentBuilder.java:1299)
	at net.bytebuddy.agent.builder.AgentBuilder$RawMatcher$Conjunction.matches(AgentBuilder.java:1150)
	at net.bytebuddy.agent.builder.AgentBuilder$RawMatcher$Conjunction.matches(AgentBuilder.java:1150)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$Transformation$SimpleMatcher.matches(AgentBuilder.java:9853)
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Collector.consider(AgentBuilder.java:6941)
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy.apply(AgentBuilder.java:4821)
	at net.bytebuddy.agent.builder.AgentBuilder$Default.doInstall(AgentBuilder.java:9463)
	at net.bytebuddy.agent.builder.AgentBuilder$Default.installOn(AgentBuilder.java:9384)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$Delegator.installOn(AgentBuilder.java:10986)
	at io.opentelemetry.auto.tooling.AgentInstaller.installBytebuddyAgent(AgentInstaller.java:147)
	at io.opentelemetry.auto.test.AgentTestRunner.agentSetup(AgentTestRunner.java:152)
```

iNikem:
> I looked at one of the prior failures, and will investigate a bit.

I will be very grateful if you can later explain, how one investigate such things... I suspect some class loading race, but without consistent reproduction, don't knot where to start

instrumentation/jms-1.1/src/latestDepTest/groovy/JMS2Test.groovy
trask: Btw, I also opened #737 to consider applying `trace.classes.exclude` only to annotation instrumentation

iNikem: I will try to formulate my problem and will open an issue regarding `AdditionalLibraryIgnoresMatcher`

trask:
I have a theory!

I think the order of the two `@BeforeClass` methods is not deterministic.

Most of the time they run in file order.

But the exception above is under the first `@BeforeClass` method:

```
io.opentelemetry.auto.test.AgentTestRunner.agentSetup(AgentTestRunner.java:152)
```

and the exception will only increment `AgentTestRunner.INSTRUMENTATION_ERROR_COUNT` if an active test has been registered with `AgentTestRunner.TEST_LISTENER`, which happens in the second `@BeforeClass` method.

It looks like they are split out into two method like this because the second one is not static (which is weird for `@BeforeClass`) so that we can override `shouldTransformClass()` in one of the tests (`FieldBackedProviderTest`).

Another odd thing about `TestRunnerListener` is that it is modeled to have multiple `activeTests`, while we never run tests in parallel (in the same JVM).

Maybe we can do a better cleanup here, but quick fix to test theory is to combine both `@BeforeClass` methods into a single non-static `@BeforeClass`.

🤞

trask:
If we are lucky, this may fix sporadic failures tracked in #529 and #619

iNikem:
Sorry, @trask, I don't follow how the wrong order of those 2 methods may lead to `java.lang.IllegalStateException: Cannot resolve type description for spring.jpa.Customer_$$_jvst56_0`

iNikem:
And I have found at least one reason why we have test failures:
https://scans.gradle.com/s/ss76zhrd32fa6/tests/:instrumentation:servlet:servlet-3.0:latestDepTest/TomcatServlet3TestDispatchAsync/test%20success%20with%204%20requests#1

There is this line: `SPAN START: TEST_SPAN id=4dcdbbcf5e59b875 traceid=af7df4ba6f7ecb5234c60b1cd5ff04e1 parent=03bbac4d9cfc4529`. As you, paren is not null, but it must be for this span created by `HttpServerTestAdvice`. Because of this this trace is not counted and test fails. Looks like we have span context leak somewhere...

trask:
> Sorry, @trask, I don't follow how the wrong order of those 2 methods may lead to `java.lang.IllegalStateException: Cannot resolve type description for spring.jpa.Customer_$$_jvst56_0`

Run `SpringJpaTest` (hiberate-4.3) locally.

It always logs:

```
23:50:18.672 [Test worker] DEBUG io.opentelemetry.auto.tooling.bytebuddy.matcher.LoggingFailSafeMatcher - Instrumentation type matcher unexpected exception: io.opentelemetry.auto.instrumentation.hibernate.v4_3.ProcedureCallInstrumentation
java.lang.IllegalStateException: Cannot resolve type description for spring.jpa.Customer_$$_jvstbdc_0
```

But it only fails the test if there's an active test in `TEST_LISTENER` at the time that message is logged.

You can repro locally by combining the two `@BeforeClass` methods into a single (non-static) method in the wrong order:

```
  @BeforeClass
  public void agentSetup() {
    TEST_LISTENER.activateTest(this);
    if (activeTransformer == null) {
      activeTransformer =
          AgentInstaller.installBytebuddyAgent(INSTRUMENTATION, true, TEST_LISTENER);
    }
  }
```

iNikem:
I moved my investigation of Servlet tests failure to https://github.com/iNikem/opentelemetry-java-instrumentation/runs/893986276?check_suite_focus=true to not spend money on CircleCI

trask:
Oh no! It looks like the `@BeforeClass` change didn't fix the problem? 😢

trask:
@iNikem it looks like github is a bit confused, I'm not sure why, it seems to be showing CI failure and conflicting files from an earlier commit(?)

iNikem:
@trask disregard this PR for the time being. I am rerunning it in GH actions to reduce cost and trying to do something with flaky tests there

iNikem:
Potentially closes #529 and #619

iNikem:
Will I ever learn to use git? :(

iNikem:
@trask @anuraaga Now it is ready for merge 🎉 

I will take a look at flaky test separately from this. Soon™️

PR #726, opened 2020-07-18
feat(servlet): content length

FrankSpitulski:
add content length tag for v3 servlets as per https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#common-attributes


FrankSpitulski:
I signed the EasyCLA doc. Not sure what else I have to do to make the bot happy.

agent-bootstrap/src/main/java/io/opentelemetry/auto/instrumentation/api/MoreAttributes.java
```diff
@@ -19,6 +19,7 @@
 public class MoreAttributes {
   public static final String HTTP_QUERY = "http.query.string";
   public static final String HTTP_FRAGMENT = "http.fragment.string";
+  public static final String HTTP_RESPONSE_CONTENT_LENGTH = "http.response_content_length";
```
anuraaga, line 22: If I recall correctly this was already released in otel 0.6.0, is it not part of `SemanticAttributes`?

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
```diff
@@ -0,0 +1,86 @@
...
+
+package io.opentelemetry.auto.instrumentation.servlet.v3_0;
+
+import java.io.IOException;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
+
+/** HttpServletResponseWrapper since servlet 2.3, not applicable to 2.2 */
+public class CountingHttpServletResponse extends HttpServletResponseWrapper {
```
anuraaga, line 25: I think we may need counting for cases like chunked responses but a huge majority will have a content length header. What do you think about starting with that to get most bang for buck? We could do it in `HttpServerTracer` I think.

Either way we probably want to avoid adding this wrapper when the response has a content length header. We could use the content length in a future PR instead.

FrankSpitulski: From my testing the set Content Length method is not invoked by springboot's rest controller. Counting the bytes was the only reliable way to get the size. I also tried to stick it in the http server tracer but the wrapper was introduced in servlet 2.3 and causes issues in the servlet 2.2 instrumentation.

agent-bootstrap/src/main/java/io/opentelemetry/auto/instrumentation/api/MoreAttributes.java
FrankSpitulski: I did a project wide search for the string before adding it. Is that defined elsewhere?

anuraaga: Yeah it's in the SDK repo

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/attributes/SemanticAttributes.java

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
anuraaga: I see that sounds like for servlet we'll need to ignore the content length header always, thanks for the background

FrankSpitulski:
For completeness we should wrap getWriter, but in practice I don't think it is used on the httpservletresponse. The httpservletresponse overrides the outputstream and the use of the writer vs the output stream is mutually exclusive according to the docs. I can add a wrapper for it but I don't have a real world test case that uses it to verify.

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public void write(byte[] b) throws IOException {
+      delegate.write(b);
+      counter += b.length;
+    }
+
+    @Override
+    public void write(byte[] b, int off, int len) throws IOException {
+      delegate.write(b, off, len);
+      counter += len - off;
```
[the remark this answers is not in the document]
FrankSpitulski: Good catch

trask:
> I don't have a real world test case that uses it to verify.

I think the `jsp-2.3` module tests (which also test the `servlet-3.0` instrumentation) should use `getWriter()`, which would give us test coverage

instrumentation-core/servlet/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -109,4 +109,8 @@ public void setPrincipal(Span span, HttpServletRequest request) {
       span.setAttribute(MoreAttributes.USER_NAME, principal.getName());
     }
   }
+
+  public void setContentLength(Span span, int length) {
```
iNikem, line 113: The design principal of Tracers (as opposed to old Decorators) is to expose as small API as possible. For this reason I don't really like this new public method here. I don't have better idea though right now.

FrankSpitulski: It could go in the servlet 3 advice class. I originally put it here because that seemed to be the pattern.

trask: I opened #733 so we can leave this as-is in this PR

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/AsyncContextInstrumentation.java
```diff
@@ -50,6 +50,9 @@ public AsyncContextInstrumentation() {
     return new String[] {
       "io.opentelemetry.instrumentation.servlet.HttpServletRequestGetter",
       "io.opentelemetry.instrumentation.servlet.ServletHttpServerTracer",
+      "io.opentelemetry.auto.instrumentation.servlet.v3_0.CountingHttpServletResponse",
```
iNikem, line 53: Use `packageName` as below

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/Servlet3Instrumentation.java
```diff
@@ -54,6 +54,9 @@ public Servlet3Instrumentation() {
     return new String[] {
       "io.opentelemetry.instrumentation.servlet.HttpServletRequestGetter",
       "io.opentelemetry.instrumentation.servlet.ServletHttpServerTracer",
+      "io.opentelemetry.auto.instrumentation.servlet.v3_0.CountingHttpServletResponse",
```
iNikem, line 57: Use `packageName` as below

instrumentation/servlet/servlet-3.0/src/test/groovy/AbstractServlet3Test.groovy
```diff
@@ -88,6 +88,8 @@ abstract class AbstractServlet3Test<SERVER, CONTEXT> extends HttpServerTest<SERV
         "${SemanticAttributes.HTTP_URL.key()}" { it == "${endpoint.resolve(address)}" || it == "${endpoint.resolveWithoutFragment(address)}" }
         "${SemanticAttributes.HTTP_METHOD.key()}" method
         "${SemanticAttributes.HTTP_STATUS_CODE.key()}" endpoint.status
+        "${SemanticAttributes.HTTP_RESPONSE_CONTENT_LENGTH.key()}" { it == null || it == getContentLength(endpoint) }
```
trask, line 91: curious under what conditions is response content length not captured? (why we need `it == null`)

FrankSpitulski: @trask There are quite a few where response length is not captured in the tests. 
image

I haven't figured out how to track down why that is happening yet. The groovy tests are a lot less straightforward than junit.

FrankSpitulski: All of the sync tests are happy, maybe my async implementation is bad.

FrankSpitulski: checking the actual okhttp response body length in the test fixed the async tests

gradlew.bat
```diff
@@ -1,104 +1,104 @@
-@rem
```
FrankSpitulski, line 1: I cannot get git to leave this file alone. It changes it to have CRLF but according to the .gitattributes that should already be the case.

trask: yeah, we just changed the `.gitattributes` file.

can you try:

```
rm .gitattributes
git checkout -- .gitattributes
```

trask: oh i just got stuck in this again now too, will debug, really not understanding this either

trask: this seems to make it go away at least temporarily:

```
sed -i '/.bat/d' .gitattributes && git status > /dev/null && git checkout -- .gitattributes
```

trask: i think i found permanent fix #743, sorry for the confusion 😅

FrankSpitulski: 👍

FrankSpitulski:
all tests good now

FrankSpitulski:
I think the ci is having some interment failures. I'll squash a commit to trigger it again.

trask:
> I think the ci is having some interment failures. I'll squash a commit to trigger it again.

Yeah, we have several sporadic failures currently: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues?q=is%3Aissue+is%3Aopen+label%3A%22sporadic+test+failure%22

We'll watch the build and hit retry if it fails again.

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public void close() throws IOException {
+      delegate.close();
+    }
+  }
+
+  static class CountingPrintWriter extends PrintWriter {
+    // PrintWriter is synchronised, so the counter has to be atomic
+    private AtomicInteger counter = new AtomicInteger(0);
```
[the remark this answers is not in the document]
FrankSpitulski: The jdk implementation uses an internal lock to synchronise pretty much every method. Since I can't grab that lock I can imagine a situation under high load where two of those methods are called, pulled off of the cpu, then scheduled on two cpus at the same time.

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
```diff
@@ -0,0 +1,146 @@
...
+  public void sendError(int sc, String msg) throws IOException {
+    super.sendError(sc, msg);
+    if (msg != null) {
+      errorLength += msg.length();
+    }
+  }
+
+  static class CountingServletOutputStream extends ServletOutputStream {
+
+    private ServletOutputStream delegate;
```
trask, line 80:
```suggestion
    private final ServletOutputStream delegate;
```

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
trask, line 118:
```suggestion
    private final AtomicInteger counter = new AtomicInteger(0);
```

testing-common/src/main/groovy/io/opentelemetry/auto/test/base/HttpServerTest.groovy
```diff
@@ -453,7 +453,7 @@ abstract class HttpServerTest<SERVER> extends AgentTestRunner {
   }
 
   // parent span must be cast otherwise it breaks debugging classloading (junit loads it early)
-  void serverSpan(TraceAssert trace, int index, String traceID = null, String parentID = null, String method = "GET", ServerEndpoint endpoint = SUCCESS) {
+  void serverSpan(TraceAssert trace, int index, String traceID = null, String parentID = null, String method = "GET", ServerEndpoint endpoint = SUCCESS, Response response = null) {
```
trask, line 456: can you change this to pass in `int responseContentLength` instead of the okhttp `Response`, and also swap the last two params, so then can leave off `endpoint` from the callers since most don't use it?

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/CountingHttpServletResponse.java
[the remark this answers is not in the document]
trask: I think the Servlet Spec comes to our rescue here:

> Other than the startAsync and complete methods, implementations of the request
and response objects are not guaranteed to be thread safe. This means that they
should either only be used within the scope of the request handling thread or the
application must ensure that access to the request and response objects are thread
safe.

My reading of this is that we don't even need volatile here (similar to implementation above in `CountingServletOutputStream`).

testing-common/src/main/groovy/io/opentelemetry/auto/test/base/HttpServerTest.groovy
FrankSpitulski: sure

FrankSpitulski: I'll need to make it a nullable object for the optional parameters to work, and also a Long to remove unnecessary type casting. Perhaps the SemanticAttributes.HTTP_RESPONSE_CONTENT_LENGTH should have a long type instead of string as well.

trask: 👍 I opened https://github.com/open-telemetry/opentelemetry-java/issues/1442 this morning while reviewing your PR

FrankSpitulski:
@trask not sure what's going on with that failed test

```groovy.lang.MissingMethodException: No signature of method: FinatraServer.started() is applicable for argument types: () values: []
Possible solutions: start(), stage(), trace(scala.Function0), trace(org.slf4j.Marker, scala.Function0), trace(scala.Function0, java.lang.Throwable), trace(org.slf4j.Marker, scala.Function0, java.lang.Throwable)
	at FinatraServerTest.startServer(FinatraServerTest.groovy:57)```

startServer is an implicit function from the twitter scala library. The IDE seems to find it fine and I don't think I changed anything that would cause it to fail.

trask:
yeah, the `test_latest` build runs the tests against latest dependencies (in this case the latest finatra dependency) available in maven central, so sometimes it fails when there's a new version released, and i can see there was a new finatra version released just today

https://search.maven.org/search?q=g:com.twitter%20a:finatra-http_2.11

so that maybe the issue, i'll check it out and get back to you

trask:
Yeah, that's the problem. I submitted #747 to fix. And then I think I can merge master into your PR to get clean build, or if not I will ping you to merge master into your branch. thx for your patience as we go through growing pains 😅

trask:
Thanks again @FrankSpitulski 🎉

PR #734, opened 2020-07-20
Add GA requirements doc

trask:
Trying to draft GA requirements and prioritization. This is just a starting place. Looking for feedback here. Will also discuss in this Thu meeting. Thanks!


docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
+P0 (e.g. cannot GA without these):
+* End-to-end tests
+  * OTLP, Jaeger and Zipkin
+  * Tomcat, Wildfly, Spring Boot
```
anuraaga, line 4: This list seems sort of arbitrary. Is it P0 to test 3 servlet implementations? My suggestion based on back-of-hand estimation of popularity nowadays is one servlet, spring-boot, and grpc, while others could be P1 or P2

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
...
+  * All Java LTS versions + latest Java version if non-LTS
+* Benchmarking
+  * Startup overhead benchmark
```
anuraaga, line 7: This feels lower than P0, maybe P2.

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
...
+  * Non-required semantic attributes are nice to have, but not required for GA
+* Basic configuration points
+  * Add custom auto-instrumentation
+  * Ability to build "custom distro"
+* Documentation (tag "documentation" minus tag "contributor experience")
+  * All configuration options
+    * Standard OpenTelemetry SDK + Exporter configuration options
+    * Auto-instrumentation configuration options (e.g. disable/enable, peer.service mapping)
+  * For each instrumentation
+    * Document all spans that it captures
```
anuraaga, line 19: I don't think I've ever seen such detailed documentation per-instrumentation in other tracing libraries. More docs is always better than less, but I don't think it's P0, maybe P2, and it needs to come with the commitment that we will keep the docs up to date.

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
...
+      * Span attributes (including explanation of any non-semantic attributes)
+      * Events
+    * Document any instrumentation-specific configuration
+    * Document any other effects (e.g. updating SERVER span name with route)
+  * Document the basic configuration points (above)
+    * How to write your own auto-instrumentation
+      * (much of this can be shared with contributor documentation below)
+    * How to build your own "custom distro"
+  * How to troubleshoot (start documenting common issues somewhere)
+* Library (manual) instrumentations corresponding to at least 20 auto-instrumentations, with shared code and tests
```
anuraaga, line 30: I consider Brave to have very high instrumentation coverage, and it's a total of 25 (including some that are multi-version)
https://github.com/openzipkin/brave/tree/master/instrumentation

20 feels very high for a P0, my suggestion would be making this more use-case oriented.

    Libraries commonly used on Android
    Libraries commonly used with spring-boot

for example.

docs/ga-requirements.md
iNikem: Spring Boot certainly covers Tomcat or Jetty, which is all the same. Wildfly is useful, because JBoss is tricky classloader-wise. It uses its own module system and is very sensitive to logging infra problems.

I think current smoke tests selection, Boot, Wildfly and Play, is quite good and propose to continue with that

trask: Wildfly is problematic for us, because of both jboss modules (class loading) and jboss logging (we need avoid touching java.util.logging before it loads).

I think our normal instrumentation tests cover Tomcat, Spring Boot and gRPC decently, so I would be good with pushing them to lower priority.

```suggestion
  * Wildfly, due to common regressions around jboss modules (class loading) and jboss logging (we need avoid touching java.util.logging before it loads)
```

docs/ga-requirements.md
iNikem: Not sure. P1 - can agree, P2 - probably not.

docs/ga-requirements.md
iNikem: Unfortunately must agree with Anuraag.

docs/ga-requirements.md
trask: kept this at P0 for now (well, now P0 is P1) since the next priority level is all about "contributor experience", but agree this is flexible

trask:
Addressed initial feedback, please re-review 👍

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
+P1 (e.g. cannot GA without these):
+* End-to-end tests
+  * OTLP, Jaeger and Zipkin
+  * Spring Boot and Wildfly
+    * (Wildfly chosen due to common javaagent issues around jboss modules and jboss logging)
+  * All Java LTS versions + latest Java version if non-LTS
```
anuraaga, line 6:
```suggestion
  * Java 8, 11, and the latest Java version
```

I want to be optimistic we GA before Java 17 :)

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
+P1 (e.g. cannot GA without these):
+* End-to-end tests
+  * OTLP, Jaeger and Zipkin
+  * Spring Boot and Wildfly
+    * (Wildfly chosen due to common javaagent issues around jboss modules and jboss logging)
+  * Java 8, 11, and the latest Java version
+* Benchmarking
+  * Runtime overhead benchmark
+  * Startup overhead benchmark
+* Implement all required semantic attributes
```
iNikem, line 10: There is this idea: https://github.com/open-telemetry/opentelemetry-specification/issues/653#issuecomment-662156100

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
...
+* Library (manual) instrumentations for libraries commonly used with Spring
+  * Spring WebMVC, Spring WebFlux, Spring RestTemplate, Spring Kafka, Spring AMQP,
+    Reactor, JDBC, java.util.concurrent
+* Library (manual) instrumentations for libraries commonly used with Android
+  * OkHttp, gRPC
+
+P2:
+* Contributor experience (tag "contributor experience" plus tag "cleanup" plus tag "sporadic test failure")
+  * New contributor documentation
+    * How to write new instrumentation (auto, library, tests)
```
iNikem, line 35: But P1 already has "How to write your own auto-instrumentation". And I think that without these documentation here nobody will be able to "Add custom auto-instrumentation". But putting all "New contributor documentation" in P1 seems like too aggressive to me :(

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
...
+    * Standard OpenTelemetry SDK + Exporter configuration options
+    * Auto-instrumentation configuration options (e.g. disable/enable, peer.service mapping)
+  * For each instrumentation
+    * Document any instrumentation-specific configuration
+  * Document the basic configuration points (above)
+    * How to write your own auto-instrumentation
+      * (much of this can be shared with contributor documentation below)
+    * How to build your own "custom distro"
+  * How to troubleshoot (start documenting common issues somewhere)
+* Library (manual) instrumentations for libraries commonly used with Spring
```
iNikem, line 26: I am not sure this is hard requirement for GA

docs/ga-requirements.md
```diff
@@ -0,0 +1,51 @@
...
+  * Spring WebMVC, Spring WebFlux, Spring RestTemplate, Spring Kafka, Spring AMQP,
+    Reactor, JDBC, java.util.concurrent
+* Library (manual) instrumentations for libraries commonly used with Android
```
iNikem, line 29: I am not sure this is hard requirement for GA

docs/ga-requirements.md
trask: I'd like to have some kind of hard requirement for manual instrumentation, in order to show our commitment to supporting both auto and manual instrumentation.

anuraaga: Not sure if it matters much, but FWIW, the work for extracting manual instrumentation for okhttp and grpc seems very small compared to Spring which is a lot of new instrumentation, though @mabdinur is doing a great job tackling it :)

PR #741, opened 2020-07-20
Add new triager @malafeev

trask:
Malafeev is a regular contributor to this repo.

This is a new role that gives regular contributors the github Triage permission so that they can (among other things) apply labels and assign issues.


trask:
🥳

PR #743, opened 2020-07-21
Turn gradlew.bat into a text file

trask:
I think this fixes the issue we have been seeing after adding the `.gitattributes` file, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/725#discussion_r457093672 and https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/726#discussion_r457815576.

My understanding (now) is that previously the `gradlew.bat` file was not stored in git as a "text file". Instead it was stored in git with CRLFs. Now we want it to be a "text file" (stored in git without CRLFs). And our new `.gitattributes` file tells git to use CRLFs in working directory on checkout.


trask:
🤞

Issue #588, opened 2020-06-27
Document InstrumentationContext in CONTRIBUTING


trask:
This is an important part of the codebase that is super non-obvious.

Compare/contrast with Context (and some instrumentation keep some state in, e.g. HttpRequest attributes as well. So we have at least 3 options. How a developer should choose which one of them to use?)

ref to gitter discussion: https://gitter.im/open-telemetry/opentelemetry-java-instrumentation?at=5ef7a1ceb8152d3484697a86

trask:
Include that `InstrumentationContext` can only be used inside of advice (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/746#discussion_r458243293)

trask:
> Include that `InstrumentationContext` can only be used inside of advice ([#746 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/746#discussion_r458243293))

This is documented in the `InstrumentationContext` javadoc now.

The rest is resolved by #2793.

PR #746, opened 2020-07-21
Target OTel SDK snapshot

### LINKED ISSUE #728: Have master branch depend on OTel time-stamped snapshots?
Given there's a lot of work being put into the spec, e.g., #723, I wonder whether instead of having the otel-snapshot branch just make master depend on the snapshots until GA? I think it's fairly common to do SNAPSHOTs for dev, non for release, and is codified in maven-release-plugin and gradle-release-plugin. This would allow us to keep up faster which changes in the spec and probably speed up towards GA. The downside is from what I understand it can be a lot of work to maintain our API shim, so targeting snapshots might increase that work so much that it's not worth it. Wonder what people's thoughts are.

anuraaga:
While `recordException` will be the snapshot I really want figured it's good to get practice.

Fixes #728


anuraaga:
As an aside, I think it would be good to update this page, claims about API stability feel a bit silly...

https://opentelemetry.io/project-status/#04-additional-beta-release

iNikem:
`depends-on-otel-java-snapshot` branch can be deleted then?

anuraaga:
Yeah think so.

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/trace/Bridging.java
```diff
@@ -169,6 +172,12 @@ public void consume(String key, AttributeValue attributeValue) {
         .build();
   }
 
+  public static io.grpc.Context toShaded(Context context) {
+    ContextStore<Context, io.grpc.Context> contextStore =
+        InstrumentationContext.get(Context.class, io.grpc.Context.class);
```
trask, line 177: unfortunately InstrumentationContext is not documented yet (#588), but this won't work here. InstrumentationContext can only be used inside of advice classes because we change those call sites while injecting the advice, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/agent-tooling/src/main/java/io/opentelemetry/auto/tooling/context/FieldBackedProvider.java#L82-L86

trask:
> As an aside, I think it would be good to update this page, claims about API stability feel a bit silly...
> 
> https://opentelemetry.io/project-status/#04-additional-beta-release

👍 I opened a PR to update this page: https://github.com/open-telemetry/opentelemetry.io/pull/223

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/trace/Bridging.java
anuraaga: Oh bummer. Any pointer on how to convert this `Context`? :) If it's a hard problem, I can ask about reverting the API change, it seems like a very minor convenience.

trask: for this PR i suggest

passing `ContextStore` all the way down to the `Span.Builder` starting here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/OpenTelemetryApiInstrumentation.java#L68

similar to https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/OpenTelemetryApiInstrumentation.java#L90

and let's discuss what to do in the future in #752 👍

anuraaga: Thanks, plumbed

PR #756, opened 2020-07-22
Add instrumentation for Armeria (server)

anuraaga:
Why?

- Armeria is my go-to framework so I want it instrumented!
- I wanted an example of a instrumented a framework that provides detailed logging, including timings, in the framework itself (so use its timestamps). Realized I forgot to add to end, will update it
- I hadn't written an instrumentation from scratch. It's a good experience :)


instrumentation/armeria-0.99/library/src/main/java/io/opentelemetry/instrumentation/armeria/v0_99/server/OpenTelemetryService.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private OpenTelemetryService(HttpService delegate, ArmeriaServerTracer serverTracer) {
+    super(delegate);
+    this.serverTracer = serverTracer;
+  }
+
+  @Override
+  public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) throws Exception {
+    String route = route(ctx);
+    String spanName = route != null ? route : "HTTP " + req.method().name();
```
trustin, line 57: Curious if we can utilize `ctx.log().serviceName()`

anuraaga: Yeah let me check it in a followup PR where I add tests for armeria-grpc with it, for now I want to focus on the minimum HTTP behavior. Especially with auto instrumentation we might have some issue due to applying both instrumentations.

In the meantime maybe you can help with this one :)
 https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/756/files#diff-a63d5fb32f3eb41b3cde769b7807b2e6R87

instrumentation/armeria-0.99/testing/src/main/groovy/io/opentelemetry/instrumentation/armeria/v_0_99/AbstractArmeriaServerTest.groovy
```diff
@@ -0,0 +1,126 @@
...
+          }
+        }
+      }
+    }
+
+    where:
+    path          | spanName     | method          | code
+    "/exact"      | "/exact"     | HttpMethod.GET  | 200
+    // TODO(anuraaga): Seems to be an Armeria bug not to have :objectId here
+    "/items/1234" | "/items/:"   | HttpMethod.POST | 200
```
minwoox, line 88: So you mean the `spanName` should be `/items/:itemsId`?

trustin, line 88: IIRC it's working as expected, although not very nice: https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/server/ParameterizedPathMapping.java#L64-L70

Armeria's `PathMapping` probably needs to provide a dedicated getter method for distributed tracing spans.

anuraaga: Yeah

anuraaga: Having a method like that to give a human readable representation of the route would be nice.

trustin: Created an issue for solving this better: https://github.com/line/armeria/issues/2932

instrumentation/armeria-0.99/library/src/main/java/io/opentelemetry/instrumentation/armeria/v0_99/server/ArmeriaServerTracer.java
```diff
@@ -0,0 +1,104 @@
...
+import io.opentelemetry.auto.bootstrap.instrumentation.decorator.HttpServerTracer;
+import io.opentelemetry.context.propagation.HttpTextFormat.Getter;
+import io.opentelemetry.instrumentation.armeria.v0_99.internal.ContextUtil;
+import io.opentelemetry.trace.Tracer;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+import java.net.URI;
+
+final class ArmeriaServerTracer
+    extends HttpServerTracer<HttpRequest, ServiceRequestContext, RequestContext> {
```
trask, line 33: it looks like we aren't using the `STORAGE` parameter, e.g. no calls to `getServerContext`

if that's the case, maybe good to be explicit about this (similar to `AkkaHttpServerTracer`)

```suggestion
    extends HttpServerTracer<HttpRequest, ServiceRequestContext, Void> {
```

instrumentation/armeria-0.99/library/src/main/java/io/opentelemetry/instrumentation/armeria/v0_99/server/OpenTelemetryService.java
```diff
@@ -0,0 +1,120 @@
...
+    Span span = serverTracer.startSpan(req, ctx, spanName, requestStartTimeNanos);
+
+    // For non-recording spans, nothing special to do.
+    if (!span.isRecording()) {
+      try (Scope ignored = serverTracer.startScope(span, ctx)) {
+        return unwrap().serve(ctx, req);
+      }
+    }
+
+    ctx.log()
```
trask, line 72: maybe wrap this in `span.isRecording()` and then don't need above condition/short-circuit?

instrumentation/armeria-0.99/testing/src/main/groovy/io/opentelemetry/instrumentation/armeria/v_0_99/AbstractArmeriaServerTest.groovy
```diff
@@ -0,0 +1,126 @@
...
+import com.linecorp.armeria.testing.junit4.server.ServerRule
+import io.opentelemetry.auto.test.InstrumentationSpecification
+import io.opentelemetry.auto.test.utils.TraceUtils
+import io.opentelemetry.trace.attributes.SemanticAttributes
+import spock.lang.Ignore
+import spock.lang.Shared
+
+import static io.opentelemetry.trace.Span.Kind.SERVER
+
+abstract class AbstractArmeriaServerTest extends InstrumentationSpecification {
```
trask, line 34: did you try extending `HttpServerTest`?

trask:
Oh yes, and add it to the README!

instrumentation/armeria-0.99/library/src/main/java/io/opentelemetry/instrumentation/armeria/v0_99/server/ArmeriaServerTracer.java
trask: ya, it's attached in `startScope`, but then never read after that

instrumentation/armeria-0.99/auto/armeria-0.99-auto.gradle
```diff
@@ -0,0 +1,24 @@
...
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+// TODO(anuraaga): Move into instrumentation.gradle
+archivesBaseName = projectDir.parentFile.name
+
+muzzle {
+  pass {
+    group = "com.linecorp.armeria"
+    module = "armeria"
+    versions = "[0.99.8,)"
```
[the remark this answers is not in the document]
trask: would it be weird to rename the module to armeria-1.0 now, just to avoid unnecessary maven artifact naming churn later? we can still doc the "correct" as of now version in README

README.md
```diff
@@ -143,6 +143,7 @@ provide the path to a JAR file including an SPI implementation using the system
 | [Akka HTTP](https://doc.akka.io/docs/akka-http/current/index.html)                                                                    | 10.0+                          |
 | [Apache HttpAsyncClient](https://hc.apache.org/index.html)                                                                            | 4.0+                           |
 | [Apache HttpClient](https://hc.apache.org/index.html)                                                                                 | 2.0+                           |
+| [Armeria](https://armeria.dev)                                                                                                        | 0.98+                          |
```
trask, line 146: is this a typo? `0.99.8` was min version in the muzzle range

instrumentation/armeria-0.99/auto/armeria-0.99-auto.gradle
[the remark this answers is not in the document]
anuraaga: Yeah think that makes sense

instrumentation/armeria-0.99/library/src/main/java/io/opentelemetry/instrumentation/armeria/v0_99/server/ArmeriaServerTracer.java
anuraaga: Hmm I think I want to keep it like this, Armeria does provide a storage implementation, whether that's actually used or not is more a decision of `HttpServerTracer` than the armeria instrumentation from what I can tell.

trask: > whether that's actually used or not is more a decision of `HttpServerTracer` than the armeria instrumentation from what I can tell

so, confusingly, this is not the case. it's up to the individual instrumentation to use what's in `STORAGE` if they need it for propagation

it seems likely we will need it in the armeria instrumentation in the future, e.g. for propagating context to some async handlers, so we can keep it, though maybe a comment about it not being used currently?

@iNikem please correct me if I'm missing something about the `STORAGE` param (either current or future plans).

trask: oh yes, and i opened #791 to document the `STORAGE` param, it is confusing when/if to use it

anuraaga: Ok less complexity makes sense. I'll revisit after instrumenting client and introducing some more complex async scenarios.

PR #763, opened 2020-07-23
Add tests to spring-boot-autoconfigure

mabdinur:
Add auto configuration tests
Add bean post processor tests

### DIFF
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/httpclients/webclient/WebClientBeanPostProcessorTest.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/httpclients/webclient/WebClientBeanPostProcessorTest.java
@@ -65,7 +65,7 @@ public class WebClientBeanPostProcessorTest {
   }
 
   @Test
-  public void should_add_ONE_exchange_filter_to_WebClient() {
+  public void should_add_exchange_filter_to_WebClient() {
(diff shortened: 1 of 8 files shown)


instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/httpclients/webclient/WebClientBeanPostProcessorTest.java
```diff
@@ -0,0 +1,101 @@
...
+        webClientBeanPostProcessor.postProcessAfterInitialization(
+                webClientBuilder, "testWebClientBuilder")
+            instanceof WebClient.Builder);
+  }
+
+  @Test
+  public void should_add_ONE_exchange_filter_to_WebClient() {
+    WebClient webClient = WebClient.create();
+    Object processedWebClient =
+        webClientBeanPostProcessor.postProcessAfterInitialization(webClient, "testWebClient");
```
trask, line 71: did you mean to call this multiple times?

mabdinur: It wouldn't make sense to call webClientBeanPostProcessor.postProcessAfterInitialization multiple times since it returns a new WebClient object. I changed the name of the test from `should_add_ONE_exchange_filter_to_WebClient` to `should_add_exchange_filter_to_WebClient`

PR #768, opened 2020-07-23
Add type matching benchmark

trask:
I added this to benchmark the change discussed in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/646#discussion_r451452916, and the benchmark seems useful to keep around (and probably improve).

### DIFF
--- a/benchmark/benchmark.gradle
+++ b/benchmark/benchmark.gradle
@@ -12,6 +12,9 @@ dependencies {
   jmh 'com.google.http-client:google-http-client:1.19.0'
   jmh 'org.eclipse.jetty:jetty-server:9.4.1.v20170120'
   jmh 'org.eclipse.jetty:jetty-servlet:9.4.1.v20170120'
+
+  // used to provide lots of classes for TypeMatchingBenchmark
+  jmh 'org.springframework:spring-web:4.3.28.RELEASE'
 }
 
 jmh {
@@ -22,10 +25,16 @@ jmh {
   fork = 1 // How many times to forks a single benchmark. Use 0 to disable forking altogether
 //  jvmArgs += ["-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints", "-XX:StartFlightRecording=delay=5s,dumponexit=true,name=jmh-benchmark,filename=$rootDir/benchmark/build/reports/jmh/jmh-benchmark.jfr"]
 //  jvmArgs += ["-agentpath:$rootDir/benchmark/src/jmh/resources/libasyncProfiler.so=start,collapsed,file=$rootDir/benchmark/build/reports/jmh/profiler.txt"]
-  failOnError = true // Should JMH fail immediately if any benchmark had experienced the unrecoverable error?
+
+  // uncomment to use for TypeMatchingBenchmark
+//  jvmArgs += ["-javaagent:${project(':opentelemetry-javaagent').buildDir}/libs/opentelemetry-javaagent-${project.version}-all.jar".toString(), "-Dota.exporter=logging"]
+
+  // Should JMH fail immediately if any benchmark had experienced the unrecoverable error?
+  failOnError = true
   warmup = '5s' // Time to spend at each warmup iteration.
 //  warmupBatchSize = 10 // Warmup batch size: number of benchmark method calls per operation.
-  warmupForks = 0 // How many warmup forks to make for a single benchmark. 0 to disable warmup forks.
+  // How many warmup forks to make for a single benchmark. 0 to disable warmup forks.
+  warmupForks = 0
   warmupIterations = 1 // Number of warmup iterations to do.
 
 //  profilers = ['stack:lines=5;detailLine=true;period=5;excludePackages=true']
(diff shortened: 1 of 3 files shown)


benchmark/benchmark.gradle
```diff
@@ -22,10 +25,16 @@ jmh {
   fork = 1 // How many times to forks a single benchmark. Use 0 to disable forking altogether
 //  jvmArgs += ["-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints", "-XX:StartFlightRecording=delay=5s,dumponexit=true,name=jmh-benchmark,filename=$rootDir/benchmark/build/reports/jmh/jmh-benchmark.jfr"]
 //  jvmArgs += ["-agentpath:$rootDir/benchmark/src/jmh/resources/libasyncProfiler.so=start,collapsed,file=$rootDir/benchmark/build/reports/jmh/profiler.txt"]
-  failOnError = true // Should JMH fail immediately if any benchmark had experienced the unrecoverable error?
+
+  // uncomment to use for TypeMatchingBenchmark
```
anuraaga, line 29: Is it worth having a separate Gradle project?

iNikem: Or separate Gradle task

benchmark/src/jmh/java/io/opentelemetry/benchmark/TypeMatchingBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+@BenchmarkMode(Mode.SingleShotTime)
+@Fork(5)
+@Warmup(iterations = 0)
+@Measurement(iterations = 1)
+@OutputTimeUnit(MILLISECONDS)
+@State(Scope.Thread)
+public class TypeMatchingBenchmark {
+
+  @Benchmark
+  public void loadLotsOfClasses() throws Exception {
```
iNikem, line 44: You would create a list of classes to load in `@Setup` method and benchmark itself will just load classes. And as a baseline, add a benchmark which just iterates over list of classes

iNikem:
@trask do you want to address my comments and will merge?

trask:
Yes, I do want to address your comments, they are good, just haven't got to it yet

benchmark/benchmark.gradle
trask: I just tried this, but the jmh task seems different from normal gradle tasks(?): https://github.com/melix/jmh-gradle-plugin/blob/master/src/main/groovy/me/champeau/gradle/JMHTask.java#L32

and so I run into this problem:

https://github.com/melix/jmh-gradle-plugin/issues/161#issuecomment-662367084

trask: for now, I modeled it similar to existing benchmarks, and I added #782

benchmark/src/jmh/java/io/opentelemetry/benchmark/TypeMatchingBenchmark.java
```diff
@@ -0,0 +1,97 @@
...
+    int errorCount = 0;
+    for (String className : classNames) {
+      try {
+        Class.forName(className, false, TypeMatchingBenchmark.class.getClassLoader());
+        successCount++;
+      } catch (Throwable t) {
+        errorCount++;
+      }
+    }
+    System.out.println("[loaded: " + successCount + ", failed to load: " + errorCount + "]");
```
anuraaga, line 87: Delete? Seems like it'd affect the benchmark result / be spammy

benchmark/src/jmh/java/io/opentelemetry/benchmark/TypeMatchingBenchmark.java
```diff
@@ -0,0 +1,97 @@
...
+@Warmup(iterations = 0)
+@Measurement(iterations = 1)
+@OutputTimeUnit(MILLISECONDS)
+@State(Scope.Thread)
+public class TypeMatchingBenchmark {
+
+  private Set<String> classNames;
+
+  @Setup
+  public void setup() throws IOException {
```
anuraaga, line 50: Ah just noticed maybe can stick with a static final. Find it makes benchmarks simpler not to worry about`State` unless really having to.

benchmark/src/jmh/java/io/opentelemetry/benchmark/TypeMatchingBenchmark.java
```diff
@@ -0,0 +1,87 @@
...
+      try {
+        Class.forName(className, false, TypeMatchingBenchmark.class.getClassLoader());
+      } catch (NoClassDefFoundError e) {
+        // many classes in the jar files have optional dependencies which are not present
+      }
+    }
+  }
+
+  @Fork(jvmArgsAppend = "-javaagent:/path/to/opentelemetry-javaagent-master.jar")
+  public static class WithAgentMaster extends TypeMatchingBenchmark {}
```
iNikem, line 81: What is agent master?

benchmark/src/jmh/java/io/opentelemetry/benchmark/TypeMatchingBenchmark.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @Fork(
+      jvmArgsAppend =
+          "-javaagent:/path/to/opentelemetry-java-instrumentation/java-agent/build/libs/opentelemetry-javaagent.jar")
```
iNikem, line 85: `java-agent`->`opentelemetry-agent`

benchmark/src/jmh/java/io/opentelemetry/benchmark/TypeMatchingBenchmark.java
trask: i copied from another benchmark, i think the idea was to test against multiple versions of the agent, but i don't think needed so removed in both places

PR #770, opened 2020-07-23
Run tests using JUnit5 platform and remove SpockRunner

anuraaga:
I noticed that we're adding junit platform tests in the spring instrumentation and figured it'd be good if we go ahead and upgrade to give us the option of using it. I'm still interested in doing a side-by-side comparison of spock and junit5 just to see :)

This also tries to simplify the tests by removing `SpockRunner` - only three tests actually depended on its behavior, and it seems OK to let Gradle take care of the classloader isolation instead of using a JUnit runner which constrains options for the future.


iNikem:
@anuraaga what are your plans about this?

anuraaga:
I'm going to examine it more - didn't expect content length and exceptions to eat up my days :P

anuraaga:
I'm going to play a bit more with this, most tests work. But some still have the dreaded "SEVERE: java.lang.LinkageError: loader constraint violation: loader 'bootstrap' wants to load class io.opentelemetry.trace.Span$Kind. A different class with the same name was previously loaded by 'app'. (io.opentelemetry.trace.Span$Kind is in unnamed module of loader 'app')" type of error, so may have to give up. Seems to be a difference in behavior of class loading when using the junit5 runner, even when delegating to vintage.

testing-common/src/main/groovy/io/opentelemetry/auto/test/base/HttpClientTest.groovy
```diff
@@ -42,7 +42,7 @@ import static org.junit.Assume.assumeTrue
 abstract class HttpClientTest extends AgentTestRunner {
   protected static final BODY_METHODS = ["POST", "PUT"]
   protected static final CONNECT_TIMEOUT_MS = 1000
-  protected static final BASIC_AUTH_KEY = "custom authorization header"
+  protected static final BASIC_AUTH_KEY = "custom-authorization-header"
```
anuraaga, line 45: In this PR PlayJavaWEClientTest fails because of invalid header name characters (space) which looks correct to me... The Play tests have a couple other assertions that fail (TimeoutException vs ConnectException) that seem similarly like they should be failing, as if on current master we're not actually running these tests so I need to investigate more what's going on...

trask:
it's gone green!!! 🟢🟩🟢🟩🟢

agent-tooling/src/test/groovy/io/opentelemetry/auto/test/HelperInjectionTest.groovy
```diff
@@ -33,13 +32,15 @@ import java.util.concurrent.atomic.AtomicReference
...
 import static io.opentelemetry.auto.util.gc.GCUtils.awaitGC
+import static org.junit.Assume.assumeTrue
 
 class HelperInjectionTest extends AgentSpecification {
 
   @Timeout(10)
   def "helpers injected to non-delegating classloader"() {
     setup:
     String helperClassName = HelperInjectionTest.getPackage().getName() + '.HelperClass'
+    assumeTrue(!isClassLoaded(helperClassName, Utils.getAgentClassLoader()))
```
anuraaga, line 43: @trask Yup! My biggest concern right now is this test, not sure why it's not passing with JUnit5. The issues with assumeTrue I'll need to file an issue with JUnit team for.

instrumentation/grizzly-client-1.9/src/test/groovy/GrizzlyAsyncHttpClientTest.groovy
```diff
@@ -20,10 +20,13 @@ import com.ning.http.client.Request
 import com.ning.http.client.RequestBuilder
 import com.ning.http.client.Response
 import com.ning.http.client.uri.Uri
+import io.opentelemetry.auto.test.SpockRunner
 import io.opentelemetry.auto.test.base.HttpClientTest
+import org.junit.runner.RunWith
 import spock.lang.AutoCleanup
 import spock.lang.Shared
 
+@RunWith(SpockRunner)
```
anuraaga, line 29: When I thought I finally had a good mental model of what's going on, a grizzly shows up 😅 

I think one of the reasons a couple of test failed is because with JUnit5, there is no way to intercept class loading
https://github.com/junit-team/junit5/issues/201

This is a fundamental aspect of JUnit5 so presumably even when using the vintage engine, the limitation applies.

I noticed only two tests in the entire repo actually seem to rely on this behavior though, so I gave them special treatment. At that point, it seemed `SpockRunner` doesn't actually do anything anymore, so I tried removing it completely. Everything worked, except for this one test. 

@trask if you feel like it, maybe you could look at this test without and with SpockRunner and teach me what function it's serving and why it would only affect this test :)

trask: Yeah, that's a tricky one...

There are two things conspiring against that test:

1. the test code itself gets instrumented, not only the library under test (`GrizzlyAsyncHttpClientTest$AsyncCompletionHandlerMock`)
2. the instrumentation that gets injected into that test code uses `InstrumentationContext`, and tries to put a `null` value into it

Because `AsyncCompletionHandlerMock` is a nested class inside of the test, it appears that class is loaded before we install bytebuddy, which means we can only retransform it, which means `InstrumentationContext` cannot use the (fast) field-backed provider, and instead has to use the `WeakMap`.

This is fine.

But then comes (2) and tries to put a null value into the `WeakMap`, which is what then causes the NPE.

Moving `AsyncCompletionHandlerMock` out into a top-level class resolves the issue.

I'll also open an issue about the `WeakMap` implementation not supporting `null` values, though in general we don't use the the `WeakMap` when running as `-javaagent:...`, since in that case we are typically able to do initial transform (not structure limited re-transform) and so we can inject a field into the class and use the field-backed provider.

iNikem:
@anuraaga can you please summarise the intent of this PR? I mean, you want to run tests with junit5. Ok, but what tests? All tests? Some tests?

anuraaga:
The intent is to run all the tests with JUnit5 - I think it's more future-proof, including if we want to move to Spock 3 which is based on JUnit5 at some point.

I was hoping the vintage runner would work the exact same way but of course hopes don't always pan out :) I'm not married to this PR, it is a significant regression for `assumeTrue` to not work correctly so I'll need to follow up with JUnit team to get that fixed during the 5.7 milestones. I do lean towards doing it though to keep up with JUnit. But if we don't merge it it's ok too, during investigating this I learned a lot about the agent :D

iNikem:
I am confused :) What is a relationship between Spock and Junit? You don't propose to remove Spock, do you?

anuraaga:
Nope :)

JUnit is a test runner framework as well as a framework for writing tests
Spock is a framework for writing tests
Sputnik is a JUnit custom test runner for running Spock tests (and is the standard way of running them)

So this PR changes from

JUnit4 + Sputnik -> Spock
to
JUnit5 + Vintage Runner + Sputnik -> Spock

And from my reading of Spock 2.0, I think (not sure) sputnik is gone too and it becomes
JUnit5 -> Spock directly (JUnit4 would not be supported)

gradle/dependencies.gradle
```diff
@@ -24,6 +24,7 @@ ext {
     kotlin            : "1.3.72",
     coroutines        : "1.3.0",
     springboot        : "2.3.1.RELEASE",
+    junit5            : "5.7.0-M1",
```
iNikem, line 27: Do we need a milestone or current stable is enough?

gradle/java.gradle
```diff
@@ -99,6 +99,12 @@ repositories {
 }
 
 dependencies {
+  testImplementation enforcedPlatform(group: 'org.junit', name: 'junit-bom', version: versions.junit5)
+  testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
+  testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-params'
+  testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
+  testRuntimeOnly group: 'org.junit.vintage', name: 'junit-vintage-engine'
```
iNikem, line 106: Do we need both old and new engines?

instrumentation/annotations/annotations.gradle
```diff
@@ -28,3 +28,17 @@ dependencies {
...
   }
 }
+
+test {
+  filter {
+    excludeTestsMatching 'TraceProvidersTest'
+  }
+}
+
+// Needs a fresh classloader.
```
iNikem, line 38: Why?

gradle/dependencies.gradle
anuraaga: Yeah this milestone fixed a spock issue

gradle/java.gradle
anuraaga: We need new engine for any Java tests, we have one `PatchLoggerTest`. If we get to Spock 2.0 we'll remove vintage at that point.

instrumentation/annotations/annotations.gradle
anuraaga: No clue :P Filed an issue

testing-common/src/main/groovy/io/opentelemetry/auto/test/base/HttpClientTest.groovy
trask: > as if on current master we're not actually running these tests

oh, wow, they were not, before this PR!!!

cc: @tylerbenson

tylerbenson: @anuraaga could you elaborate... which task isn't failing that should be?  Any idea why they aren't being executed?

anuraaga: I didn't end up digging on the why, but I think it was this test that started failing with this PR

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/874b157fe5fdaad80fd82f3ece0600b6860924e7/instrumentation/play-ws/play-ws-2.1/src/test/groovy/PlayWSClientTest.groovy

Netty threw a validation error because of the spaces in this header name which seems right to me.

PR #785, opened 2020-07-24
Update database semantic conventions

heyams:
Fix [#697](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/697)

I'll have a follow-up PR to use the latest SemanticAttributes from opentelemetry-java.
db.operation hasn't been updated yet since we need to parse query string in order to determine the database operation.

PR #786, opened 2020-07-25
Update OTel snapshot to 723 for recordException

anuraaga:
This doesn't update to the latest snapshot which would have lots of conflicts with #785

### DIFF
--- a/instrumentation/jetty-8.0/src/main/java/io/opentelemetry/auto/instrumentation/jetty/JettyHandlerAdvice.java
+++ b/instrumentation/jetty-8.0/src/main/java/io/opentelemetry/auto/instrumentation/jetty/JettyHandlerAdvice.java
@@ -26,7 +26,6 @@ import io.opentelemetry.context.Scope;
 import io.opentelemetry.trace.Span;
 import java.lang.reflect.Method;
 import java.util.concurrent.atomic.AtomicBoolean;
-import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import net.bytebuddy.asm.Advice;
@@ -95,14 +94,8 @@ public class JettyHandlerAdvice {
 
     // Check again in case the request finished before adding the listener.
     if (!request.isAsyncStarted() && responseHandled.compareAndSet(false, true)) {
-      contentLengthHelper(span, response);
+      JettyHttpServerTracer.contentLengthHelper(span, response);
       TRACER.end(span, response.getStatus());
-    }
-  }
-
-  public static void contentLengthHelper(Span span, ServletResponse response) {
-    if (response instanceof CountingHttpServletResponse) {
-      TRACER.setContentLength(span, ((CountingHttpServletResponse) response).getContentLength());
     }
   }
 }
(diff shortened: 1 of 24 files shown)


instrumentation/opentelemetry-api-beta/src/test/groovy/TracerTest.groovy
```diff
@@ -287,6 +288,32 @@ class TracerTest extends AgentTestRunner {
     }
   }
 
+  def "capture exception()"() {
```
trask, line 291: 👍

iNikem:
Do you want also to fix all Tracers to use new API for exceptions?

anuraaga:
I'll do it in a separate PR I just want to get the baseline up on this one. There seems to be something wrong with async servlet which I'm digging

iNikem:
I have spent 1.5 days digging into async servlets :( Please wait for me to file a bug with my findings to not duplicate my efforts.

iNikem:
A, you have another issue. Mine is here: #787

anuraaga:
@trask Would you be able to take a look at this branch? Having trouble getting servlet async tests passing. From what I can tell, previously the assertions weren't correct since they were effectively `"0" || false` always true. After fixing the tests, I noticed that `TagSettingAsyncListener` doesn't have access to the counting response if a user just calls `startAsync` so I am wrapping that too now to make sure the overridden response is used. But the count is still always 0, and with step debugging I noticed that even before the call to `startAsync`, it seems the original `HttpServletResponse` is already filled with a successful response. So stuck on why that could be happening

trask:
oh, you're right, that test validation was not correct, good find! i sent a PR to your branch https://github.com/anuraaga/opentelemetry-auto-instr-java/pull/2, i couldn't get rid of the `it == 0` issue which seems to be needed to make the dispatch tests pass for now (i think `it == 0` was the intention of that groovy `?: 0`). maybe open an issue to fix the dispatch tests and remove the `it == 0`?

anuraaga:
Ok - yeah if nothing is obvious it's good to get this baseline update in and we'll debug more separately.

instrumentation/jetty-8.0/src/main/java/io/opentelemetry/auto/instrumentation/jetty/JettyHandlerAdvice.java
```diff
@@ -86,7 +95,14 @@ public static void stopSpan(
 
     // Check again in case the request finished before adding the listener.
     if (!request.isAsyncStarted() && responseHandled.compareAndSet(false, true)) {
+      contentLengthHelper(span, response);
       TRACER.end(span, response.getStatus());
     }
   }
+
+  public static void contentLengthHelper(Span span, ServletResponse response) {
+    if (response instanceof CountingHttpServletResponse) {
```
anuraaga, line 104: Notice JettyHandler didn't seem to have implemented counting, so thought it'd be a quick fix to add it, but some conflict between unnamed classloader and app classloader here :( Tomorrow will give another small look and then revert this and remove the assertion for now.

anuraaga: Found it was because I was missing this advice from the helper classes (was referring to servlet-3.0 while writing this). Filed #796 since this was confusing.

PR #788, opened 2020-07-25
New way to make patch releases

iNikem:
Previous patch release job assumed that we have a branch started from a release tag. Patch release then trigged on any new commit to that branch.

Now, before, GA, we actually don't use that. We just want to push a tag, e.g. `v0.6.1` to an already verified commit, a release a version with that tag. In exactly the same way as minor releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,20 +171,6 @@ jobs:
           command: ./gradlew -Prelease.useLastTag=true final
           no_output_timeout: 6h
 
-  patch_release:
-    <<: *defaults
-    steps:
-      - checkout
-
-      - attach_workspace:
-          at: .
-
-      - restore_cache:
-          <<: *cache_keys
-
-      - run: ./gradlew -Prelease.useLastTag=true -Prelease.scope=patch final
-
-
 workflows:
   build_test_deploy:
     jobs:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -259,56 +254,10 @@ workflows:
...
-            - test_14
-            - test_latest
-            - muzzle
           filters:
             branches:
-              only: /\d+\.\d+\.x$/
+              ignore: /.*/
+            tags:
+              # Only for major and minor releases
+              only: /^v.*\.[1-9]+$/
```
trask, line 263: comment update, and I really hope we never have 10 patch releases for one version, but just in case 😭
```suggestion
              # Only for patch releases
              only: /^v.*\.[1-9][0-9]*$/
```

.circleci/config.yml
```diff
@@ -249,15 +235,5 @@ workflows:
...
-    jobs:
-      - patch_release:
-          filters:
-            branches:
-              ignore: /.*/
-            tags:
-              # Only for major and minor releases
-              only: /^v.*\.[1-9]+$/
+              # As long as we don't have support branches, we release all tags in the same manner
+              only: /^v.*$/
```
trask, line 239: nice 👍

iNikem:
Realised that everything can be much simpler

Issue #787, opened 2020-07-25
Context leak in Tomcat


iNikem:
`TomcatServlet3TestDispatchAsync.test success with 4 requests` test fails constantly in Github Action https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions?query=workflow%3A%22Nightly+build%22 . The reason for a failure can be seen from this log:
```

13:55:36.123 [Test worker] DEBUG http-client - --> GET http://localhost:36897/tomcat-context/dispatch/success |  
-- | --
  | 13:55:36.124 [http-nio-36897-exec-3] DEBUG io.opentelemetry.auto.tooling.Instrumenter$Default - Applying instrumentation: java_concurrent -- io.opentelemetry.auto.instrumentation.javaconcurrent.RunnableInstrumentation on jdk.internal.loader.ClassLoaders$AppClassLoader@5c29bfd |  
  | 13:55:36.129 [http-nio-36897-exec-3] DEBUG io.opentelemetry.auto.tooling.context.FieldBackedProvider - Found context-store access in io.opentelemetry.auto.instrumentation.javaconcurrent.RunnableInstrumentation |  
  | 13:55:36.131 [http-nio-36897-exec-3] DEBUG io.opentelemetry.auto.tooling.context.FieldBackedProvider - Rewriting context-store map fetch for instrumenter io.opentelemetry.auto.instrumentation.javaconcurrent.RunnableInstrumentation: java.lang.Runnable -> io.opentelemetry.auto.bootstrap.instrumentation.java.concurrent.State |  
  | 13:55:36.132 [http-nio-36897-exec-3] DEBUG io.opentelemetry.auto.tooling.AgentInstaller$TransformLoggingListener - Transformed org.apache.tomcat.util.net.NioBlockingSelector$BlockPoller$RunnableRemove -- jdk.internal.loader.ClassLoaders$AppClassLoader@5c29bfd |  
  | 13:55:36.143 [http-nio-36897-exec-3] DEBUG io.opentelemetry.auto.test.InMemoryExporter - >>> SPAN START: TEST_SPAN id=70ec8a85d67d75a7 traceid=d559ee1f62c109196811925b12ae92c0 parent=f42a8abecfe071db, library=InstrumentationLibraryInfo{name=io.opentelemetry.auto, version=null} |  
[...]

PR #792, opened 2020-07-26
Simplify javaconcurrent instrumentation

trask:
Randomly ran across this (well, ok, not totally randomly https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/787#issuecomment-663934243)


trask:
> This seems reasonable but can you refresh me on when we need to use `SpanWithScope`?

Good question, opened #794 to answer and officially deprecate it 😄

PR #795, opened 2020-07-26
Deprecate SpanWithScope

### LINKED ISSUE #794: Mark SpanWithScope as deprecated and document new recommendation in it's javadoc
Originally, we used `SpanWithScope` to pass the span and scope between `@Advice.OnMethodEnter` and `@Advice.OnMethodExit`, e.g.

```
  @Advice.OnMethodEnter(...)
  public static SpanWithScope onEnter(...) {
    ...
    Span span = ...
    return new SpanWithScope(span, currentContextWith(span));
  }
```

```
  @Advice.OnMethodExit(...)
  public static void stopSpan(
      ...
      @Advice.Enter final SpanWithScope spanWithScope) {
    Span span = spanWithScope.getSpan();
    ...
    span.end();
    spanWithScope.closeScope();
  }
```

We are (slowly) migrating to a new pattern using `@Advice.Local`:

```
  @Advice.OnMethodEnter(...)
  public static void onEnter(
      ...
      @Advice.Local("otelSpan") Span span,
    ...
    span = ...
    scope = ...
  }
```

```
  @Advice.OnMethodExit
  public static void onExit(
      ...
      @Advice.Local("otelSpan") Span span,
      ...
    span.end();
    scope.close();
  }
```

This new pattern has the following benefits:
* The new pattern is more efficient since it doesn't require instantiating the `SpanWithScope` holder object
* The new pattern extends nicely in the common case where we also need to pass `CallDepthThreadLocalMap.Depth` between the methods

trask:
Closes #794

### DIFF
--- a/agent-bootstrap/src/main/java/io/opentelemetry/auto/instrumentation/api/SpanWithScope.java
+++ b/agent-bootstrap/src/main/java/io/opentelemetry/auto/instrumentation/api/SpanWithScope.java
@@ -44,8 +44,7 @@ import io.opentelemetry.trace.Span;
  *   }
  * </pre>
  *
- * But we are (slowly) migrating to a new style using {@code @Advice.Local} that doesn't require
- * {@code SpanWithScope} anymore, e.g.
+ * We are (slowly) migrating to a new pattern using `@Advice.Local`:
  *
  * <pre>
  *   &#64;Advice.OnMethodEnter(...)
@@ -68,6 +67,15 @@ import io.opentelemetry.trace.Span;
  *     scope.close();
  *   }
  * </pre>
+ *
+ * This new pattern has the following benefits:
+ *
+ * <ul>
+ *   <li>The new pattern is more efficient since it doesn't require instantiating the {@code
+ *       SpanWithScope} holder object
+ *   <li>The new pattern extends nicely in the common case where we also need to pass {@link
+ *       io.opentelemetry.auto.bootstrap.CallDepthThreadLocalMap.Depth} between the methods
+ * </ul>
  *
  * @deprecated
  */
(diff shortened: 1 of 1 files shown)


agent-bootstrap/src/main/java/io/opentelemetry/auto/instrumentation/api/SpanWithScope.java
```diff
@@ -19,7 +19,62 @@
...
-// intentionally (for now) not implementing Closeable since not clear what it would close
+/**
+ * This is deprecated.
+ *
+ * <p>Originally, we used {@code SpanWithScope} to pass the {@link Span} and {@link Scope} between
+ * {@code @Advice.OnMethodEnter} and {@code @Advice.OnMethodExit}, e.g.
+ *
+ * <pre>
+ *   &#64;Advice.OnMethodEnter(...)
+ *   public static CallDepth onEnter(
```
anuraaga, line 30: Don't think this is the old pattern, though it's not a big deal to document in detail either I think.

trask: thx! bad copy-paste 😂

agent-bootstrap/src/main/java/io/opentelemetry/auto/instrumentation/api/SpanWithScope.java
```diff
@@ -19,7 +19,59 @@
...
+ *       ...
+ *       &#64;Advice.Enter final SpanWithScope spanWithScope) {
+ *     Span span = spanWithScope.getSpan();
+ *     ...
+ *     span.end();
+ *     spanWithScope.closeScope();
+ *   }
+ * </pre>
+ *
+ * But we are (slowly) migrating to a new style using {@code @Advice.Local} that doesn't require
```
iNikem, line 47: I think it deserves the explanation: why are we migrating? What was bad with old approach? Document your decisions, not actions :)

PR #799, opened 2020-07-27
Only run nightly build nightly

anuraaga:
Guess we didn't mean to schedule on master push.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -1,9 +1,6 @@
 name: Nightly build
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


iNikem:
Why? I think we do to ensure that main branch still works after PR merge. Probably need to rename file though.

anuraaga:
Yeah I thought we were separating out "standard build" for master / PR and nightly, nightly would have the `latestDepTest` which isn't very useful when run multiple times a day?

iNikem:
Good point about `latestDepTest`.

PR #803, opened 2020-07-27
Remove logging instrumentation

### LINKED ISSUE #720: Remove log instrumentation
I randomly looked at the logging instrumentation and found `experimentalLogCaptureThreshold`. I'm not aware of anything close to this style of converting logs to ~0 duration spans in OpenTelemetry specs and suspect this is carried over from datadog. Does it make sense to remove this instrumentation to prevent confusion?

trask:
Closes #720


iNikem:
I am somewhat confused about logging story :) These were instrumentations to create spans from log messages, right? The story of injecting traceId/spanId into logging context is a separate story and was recently done by @anuraaga in java repo, right? Do we want to provide the same injecting functionality for earlier versions of log libraries?

anuraaga:
I think we definitely want to add injection to older libraries. Most tracers do it with scope decorators (callbacks run when creating / closing scope) but AFAIU SDK currently doesn't provide that - so much more work than the log4j2 plugin

trask:
> AFAIU SDK currently doesn't provide that

Ya, see https://github.com/open-telemetry/opentelemetry-java/issues/922#issuecomment-661969038

PR #809, opened 2020-07-27
Use archivesBaseName when publishing.

anuraaga:
By default `project.name` (gradle tasks) == `archivesBaseName` (jar built by assemble) == artifactId (jar published to maven). But for these new directories, we override `archivesBaseName` since project.name is generic. So we also need to set maven `artifactId` to use `archivesBaseName` to ensure assemble and publish have the same naming.

### DIFF
--- a/gradle/instrumentation.gradle
+++ b/gradle/instrumentation.gradle
@@ -15,6 +15,11 @@ byteBuddy {
   }
 }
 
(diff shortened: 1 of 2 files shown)


iNikem:
I fail to connect PR description with changes :)

anuraaga:
Tried rewording let me know if it's still unclear.

Issue #807, opened 2020-07-27
Certain instrumentation does not record exception in many tests


anuraaga:
Grizzly does not seem to record exception in many tests. Need to confirm this is expected, for now I have added `testException() return false` without knowing if this is precise or not.

Vertx as well.

Probably same reason as Vertx but Netty too. Exception is literally caught by the test and translated into HTTP so doesn't seem to really test anything (presumably instrumentation should be doing something like what the test is doing)

Servlet Async too, there seems to be no way of completing an async context with a throwable. Our tests just throw the exception without it being handled by the instrumentation.

anuraaga:
My guess is this and #806 will be solved at the same time, those loose assertions allowed these to pass in the first place from what I can tell.

PR #813, opened 2020-07-27
Use Span.recordException for logging throwable

### LINKED ISSUE #806: Suspicious assertion - "error.*" { it == null || foo }
I've noticed some suspicious assertions recently and will file issues when I file.

`HttpServerTest` checks error attributes but has a `it == null` fallback in the assertions. This seems imprecise and can hide real bugs.

anuraaga:
I found some quirks on the way

- Previously, our exception assertions allowed no exception. This seems like a very fragile check, and instead I have gone though and explicitly disabled the exception test itself and we can investigate whether these are limitations of instrumentation, bugs of instrumentation, or expected in #807 

- Many tests actually handle exception and convert to HTTP responses. This makes exception and error to effectively be the same test and suppressing the exception test is probably correct for them. Will confirm in #807 

Note the removal of tags from some server spans is expected - the new attributes are only for exceptions, we don't map HTTP errors onto them anymore.

I expected this PR to take a while but was even more digging than I expected :) Filed some other issues along the way.

In the process, looks like I fixed the `null` assertions everywhere, so fixes #806


instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
```diff
@@ -282,6 +282,18 @@ class JSPInstrumentationBasicTests extends AgentTestRunner {
           operationName expectedOperationName()
           spanKind SERVER
           errored true
+          event(0) {
+            eventName(SemanticAttributes.EXCEPTION_EVENT_NAME)
```
iNikem, line 286: Why not usual `errorEvent`?

instrumentation/ratpack-1.4/src/test/groovy/server/RatpackHttpServerTest.groovy
```diff
@@ -137,4 +135,11 @@ class RatpackHttpServerTest extends HttpServerTest<EmbeddedApp> {
   String expectedServerSpanName(String method, ServerEndpoint endpoint) {
     return endpoint.status == 404 ? "/" : endpoint == PATH_PARAM ? "/path/:id/param" : endpoint.path
   }
+
+  @Override
+  boolean testException() {
+    // TODO(anuraaga): We record exception both in the ratpack controller and in the error handler.
+    // Do we need the latter? https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/810
+    return false
```
iNikem, line 143: If this is `false`, how can the `if` on line 128 be ever `true`?

instrumentation/google-http-client-1.19/src/main/java/io/opentelemetry/auto/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
```diff
@@ -143,7 +142,6 @@ public static void methodExit(
           // for a failed request.  Thus, check the response code
           if (response != null && !response.isSuccessStatusCode()) {
             span.setStatus(Status.UNKNOWN);
-            span.setAttribute(MoreAttributes.ERROR_MSG, response.getStatusMessage());
```
iNikem, line 146: Why was this just removed?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/BaseDecorator.java
```diff
@@ -114,12 +111,7 @@ public static void setPeer(final Span span, String peerName, String peerIp) {
   }
 
   public static void addThrowable(final Span span, final Throwable throwable) {
-    span.setAttribute(MoreAttributes.ERROR_MSG, throwable.getMessage());
-    span.setAttribute(MoreAttributes.ERROR_TYPE, throwable.getClass().getName());
-
-    StringWriter errorString = new StringWriter();
-    throwable.printStackTrace(new PrintWriter(errorString));
-    span.setAttribute(MoreAttributes.ERROR_STACK, errorString.toString());
+    span.recordException(throwable);
```
trask, line 114: 🎉

instrumentation/grizzly-2.0/src/test/groovy/GrizzlyTest.groovy
```diff
@@ -111,4 +111,10 @@ class GrizzlyTest extends HttpServerTest<HttpServer> {
   String expectedServerSpanName(String method, ServerEndpoint endpoint) {
     return "HttpCodecFilter.handleRead"
   }
+
+  @Override
+  boolean testException() {
+    // TODO(anuraaga): https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
```
trask, line 117: 👍

instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
```diff
@@ -126,11 +124,6 @@ class DropwizardTest extends HttpServerTest<DropwizardTestSupport> {
         "${SemanticAttributes.HTTP_RESPONSE_CONTENT_LENGTH.key()}" { it == responseContentLength || /* async */ it == null }
         "servlet.context" ""
         "servlet.path" ""
-        if (endpoint.errored) {
-          "error.msg" { it == null || it == EXCEPTION.body }
-          "error.type" { it == null || it == Exception.name }
-          "error.stack" { it == null || it instanceof String }
-        }
```
trask, line 133: no replacement for this?

instrumentation/vertx-reactive-3.5/src/test/groovy/server/VertxRxHttpServerTest.groovy
```diff
@@ -68,6 +68,12 @@ class VertxRxHttpServerTest extends HttpServerTest<Vertx> {
     return false
   }
 
+  @Override
+  boolean testException() {
+    // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
```
trask, line 73: just helping with the future search 😄 
```suggestion
    // TODO(anuraaga): https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
```

instrumentation/netty/netty-3.8/src/latestDepTest/groovy/Netty38ClientTest.groovy
```diff
@@ -97,17 +98,19 @@ class Netty38ClientTest extends HttpClientTest {
...
+            errorClass = Class.forName('io.netty.channel.AbstractChannel$AnnotatedConnectException')
+          } catch (ClassNotFoundException e) {
+            // Older versions use 'java.net.ConnectException' and do not have 'io.netty.channel.AbstractChannel$AnnotatedConnectException'
+          }
+          event(0) {
+            eventName(SemanticAttributes.EXCEPTION_EVENT_NAME)
+            attributes {
+              "${SemanticAttributes.EXCEPTION_TYPE.key()}" errorClass.name
+              "${SemanticAttributes.EXCEPTION_MESSAGE.key()}" ~/Connection refused:( no further information:)? \/127.0.0.1:$UNUSABLE_PORT/
+              "${SemanticAttributes.EXCEPTION_STACKTRACE.key()}" String
```
trask, line 112: there are a few places that look like they could use `errorEvent`

instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
anuraaga: Don't think so - the exception is recorded in the handler span so isn't on this one. Not sure if this is an issue with the instrumentation or expected though.

instrumentation/google-http-client-1.19/src/main/java/io/opentelemetry/auto/instrumentation/googlehttpclient/GoogleHttpClientInstrumentation.java
anuraaga: This is setting an HTTP response as an error tag, but our new convention is only about exceptions, not HTTP status line. Keeping this would be confusing wrt to our other tags.

instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
anuraaga: The assertions are more complex not just matching

instrumentation/netty/netty-3.8/src/latestDepTest/groovy/Netty38ClientTest.groovy
anuraaga: This one's because the message is matched using regex. I could make errorevent more powerful but didn't find enough of these places to warrant it yet.

instrumentation/ratpack-1.4/src/test/groovy/server/RatpackHttpServerTest.groovy
anuraaga: It can't - but I would remove that if along with the TODO if this can't ever be fixed, similar to all the tests I've only supressed the exception cases for further debugging.

testing-common/src/main/groovy/io/opentelemetry/auto/test/base/HttpServerTest.groovy
```diff
@@ -464,6 +462,16 @@ abstract class HttpServerTest<SERVER> extends AgentTestRunner {
       } else {
         parent()
       }
+      if (endpoint == EXCEPTION && !hasHandlerSpan()) {
+        event(0) {
+          eventName(SemanticAttributes.EXCEPTION_EVENT_NAME)
+          attributes {
+            "${SemanticAttributes.EXCEPTION_TYPE.key()}" { it == null || it == Exception.name }
```
anuraaga, line 469: This one I'll update

instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
trask: ah, makes sense, i think this is expected

instrumentation/netty/netty-3.8/src/latestDepTest/groovy/Netty38ClientTest.groovy
trask: it looks like you are using `errorEvent` in other places with regex?

anuraaga: Ah didn't even notice - `message` is the only one in my function without a type parameter mostly by accident :D Yeah I'll use it then

instrumentation/jsp-2.3/src/test/groovy/JSPInstrumentationBasicTests.groovy
```diff
@@ -38,7 +38,7 @@ class JSPInstrumentationBasicTests extends AgentTestRunner {
   static {
     // skip jar scanning using environment variables:
     // http://tomcat.apache.org/tomcat-7.0-doc/config/systemprops.html#JAR_Scanning
-    // having this set allows us to test with old versions of the tomcat api since
+    // having this set allows us to test with old verseions of the tomcat api since
```
trask, line 41:
```suggestion
    // having this set allows us to test with old versions of the tomcat api since
```

PR #814, opened 2020-07-27
Temporary workaround for #787

iNikem:
This is NOT a proper fix, but it will at least unblock GitHub Actions

### DIFF
--- a/instrumentation/netty/netty-4.1/src/test/groovy/Netty41ClientTest.groovy
+++ b/instrumentation/netty/netty-4.1/src/test/groovy/Netty41ClientTest.groovy
@@ -23,6 +23,9 @@ import io.netty.channel.embedded.EmbeddedChannel
 import io.netty.handler.codec.http.HttpClientCodec
 import io.opentelemetry.auto.instrumentation.netty.v4_1.client.HttpClientTracingHandler
 import io.opentelemetry.auto.test.base.HttpClientTest
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
 import org.asynchttpclient.AsyncCompletionHandler
 import org.asynchttpclient.AsyncHttpClient
 import org.asynchttpclient.DefaultAsyncHttpClientConfig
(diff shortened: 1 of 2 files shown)


instrumentation/netty/netty-4.1/src/test/groovy/Netty41ClientTest.groovy
```diff
@@ -115,7 +114,7 @@ class Netty41ClientTest extends HttpClientTest {
               "error.type" AbstractChannel.AnnotatedConnectException.name
               "error.stack" String
               // slightly different message on windows
-              "error.msg" ~/Connection refused:( no further information:)? localhost\/[0-9.:]+:$UNUSABLE_PORT/
+              "error.msg" ~/Connection refused:( no further information:)? localhost\/\[?[0-9.:]+\]?:$UNUSABLE_PORT/
```
iNikem, line 117: This is to allow both IPv4 and IPv6 addresses. GH Action has both.

instrumentation/netty/netty-4.1/src/test/groovy/Netty41ClientTest.groovy
```diff
@@ -23,6 +23,9 @@ import io.netty.channel.embedded.EmbeddedChannel
 import io.netty.handler.codec.http.HttpClientCodec
 import io.opentelemetry.auto.instrumentation.netty.v4_1.client.HttpClientTracingHandler
 import io.opentelemetry.auto.test.base.HttpClientTest
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
```
trask, line 28: #816 👍

trask:
@iNikem, if we are in agreement that this is no longer just temporary workaround, let's rename (since there are several links to this issue and so is a bit confusing)?

iNikem:
Do you want to rename PR? Or is it sufficient to change commend in code?

trask:
I'd suggest both, but if you prefer not to rename the PR that's ok

PR #817, opened 2020-07-28
Run integrationTest as part of test

### LINKED ISSUE #790: Should we make the "test" task also run "integrationTest" task?
I started to update the contributing docs to reflect that we now have two testing tasks to run, and I ended up wondering if it makes sense to have the `test` task depend on the `integrationTest` so that everything can be run using the `test` task?

And if we want a way to run non-`integrationTest` tests separately, maybe add `unitTest` and have `test` depend on both of those?

EDIT: motivation for this is from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/785#issuecomment-663892862

trask:
Closes #790

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
           name: Build Project
           command: >-
             GRADLE_OPTS="-Dorg.gradle.jvmargs='-Xmx1G -Xms64M' -Dota.forkedMaxHeapSize=1G -Dota.forkedMinHeapSize=64M"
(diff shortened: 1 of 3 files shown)


iNikem:
AAA! @anuraaga @trask Are you sure that if I run `./gradlew testJava14`, then integration tests will also run on java 14? I am ready to bet they will not.

anuraaga:
@iNikem Seems to work ok. Tried `testJava8` and it just ends up running `test` (presumably after configuring JVM) which gets finalized by `integrationTest`.

PR #820, opened 2020-07-28
Remove ratpack duplicate error capture

### LINKED ISSUE #810: Ratpack instrumentation seems to over-record exceptions
I found that Ratpack instrumentation records exceptions twice when a user defines an error handler

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/ratpack-1.4/src/main/java/io/opentelemetry/auto/instrumentation/ratpack/TracingHandler.java#L71

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/ratpack-1.4/src/main/java/io/opentelemetry/auto/instrumentation/ratpack/ErrorHandlerAdvice.java#L32

Gut feeling is latter doesn't need to be instrumented.

trask:
Closes #810


anuraaga:
My build went green faster :D Could you remove https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/ratpack-1.4/src/test/groovy/server/RatpackHttpServerTest.groovy#L140

PR #14, opened 2019-11-15
Merge dd-master into master

trask:
Resolves #3

Issue #12, opened 2019-11-15
Change instrumentation to use OpenTelemetry API directly


trask:
The reason behind this is to be able to use the same instrumentation in the future for both auto-instrumentation and whitebox instrumentation where possible.

Since instrumentation can be injected into any class loader (including the bootstrap class loader), the instrumentation's use of the OpenTelemetry API will need to be shaded, and the agent will need to provide a shaded version of the OpenTelemetry API in the bootstrap class loader. 

This may have a downstream impact whether the OpenTelemetry SDK exporters need to be shaded (or at least partially shaded to match the API shading), but there are several options available to us here, including shading a user-supplied exporter at runtime if needed (vendors could shade their own exporter at compile time when bundling everything together).

----------

Tracking modules as they are updated to use the OpenTelemetry API directly below.

If you want to work on one that has not already been updated, open an issue (e.g. #83, #84, #85) to let everyone know that you are going to work on it (to avoid two people working on the same one). Someone will assign it to you if you don't have rights to assign it to yourself.

See PRs #55, #67, #72, #75, #81, #87 to see how several of the modules have already been updated to use the OpenTelemetry API directly.

- [x] akka-http-10.0
- [x] apache-httpasyncclient-4
- [x] apache-httpclient-4
- [x] aws-java-sdk-1.11.0
- [x] cdi-1.2
- [x] couchbase-2.0
- [x] couchbase-2.6
- [x] datastax-cassandra-3
- [x] dropwizard
- [x] elasticsearch
- [x] glassfish
- [x] google-http-client
- [x] grizzly-2
- [x] grpc-1.5
- [x] hibernate
- [x] http-url-connection
- [x] hystrix-1.4
- [x] java-concurrent
- [x] jax-rs-annotations-1
- [x] jax-rs-annotations-2
- [x] jax-rs-client-1.1
- [x] jax-rs-client-2.0
- [x] jboss-classloading
- [x] jdbc
- [x] jedis-1.4
- [x] jedis-3.0
- [x] jetty-8
- [x] jms
- [x] jsp-2.3
- [x] kafka-clients-0.11
- [x] lettuce-5
- [x] mongo
- [x] netty-4.0
- [x] netty-4.1
- [x] okhttp-3
- [x] osgi-classloading
- [x] play-2.4
- [x] play-2.6
- [x] play-ws-1
- [x] play-ws-2
- [x] play-ws-2.1
- [x] ratpack-1.4
[...]

PR #823, opened 2020-07-28
Apply new database semantic attributes

heyams:
Apply new database semantic attributes.

In addition, since updating to the latest SNAPSHOT version, needed to update JaegerExporterFactory to use the new JaegerGrpcSpanExporter builder methods.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -9,21 +9,21 @@ configurations.all {
 ext {
   versions = [
     // Check https://oss.jfrog.org/libs-snapshot/io/opentelemetry/ for latest snapshot version.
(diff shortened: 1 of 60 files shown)


gradle/dependencies.gradle
```diff
@@ -9,21 +9,21 @@ configurations.all {
 ext {
   versions = [
     // Check https://oss.jfrog.org/libs-snapshot/io/opentelemetry/ for latest snapshot version.
-    opentelemetry: '0.7.0-20200723.154929-27',
```
heyams, line 12: I reformatted this file using IntelliJ, no difference.  only line #12 and #14 were updated.

PR #815, opened 2020-07-27
Create exporter starters

mabdinur:
Create an opentelemetry-spring-starter
Add spring starters for exporters (zipkin, otlp, jaeger)

Issue #723, opened 2020-07-17
Use new Span.recordException method


iNikem:
When open-telemetry/opentelemetry-java#1430 gets implemented, update Decorators/Tracers to use it.

iNikem:
@anuraaga As you did open-telemetry/opentelemetry-java#1430, do you want/plan to do this soon as well? Or I can take it. I want to close this in the next few days if possible :)

anuraaga:
Yup I'll get it :)

anuraaga:
Fixed in #813

PR #824, opened 2020-07-28
Fix build

trask:
Fix build due to collisions between #723 and #781 and #815


instrumentation-core/spring/spring-boot-autoconfigure/spring-boot-autoconfigure.gradle
```diff
@@ -13,26 +13,28 @@ dependencies {
...
     exclude group: 'org.slf4j', module: 'slf4j-simple'
   }
-   
+
   compileOnly group: 'io.grpc', name: 'grpc-api', version: '1.24.0'
   compileOnly deps.opentelemetryLogging
   compileOnly deps.opentelemetryJaeger
   compileOnly deps.opentelemetryOtlp
   compileOnly deps.opentelemetryZipkin
+
```
mabdinur, line 36:
```suggestion

testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: versions.springboot
```

I'm not sure why I didn't catch this before. This should fix the failed test in WebMVCFilterAutoConfigurationTest

trask: thx @mabdinur!

anuraaga:
Oops accidentally double approved

Issue #776, opened 2020-07-24
HttpServlet Tag Enhancements


FrankSpitulski:
I think we should also tag headers, query params, and maybe fragments in the servlet spans. I'd be happy to implement this.

trask:
hey @FrankSpitulski! I forgot about https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v0.6.0/agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java#L133, see if this is what you are looking for around query params/fragments

trask:
for http headers, are you looking to capture standard headers like http host and user agent, or others?

FrankSpitulski:
That is exactly what I'm looking for. Standard headers are good. I'd also like to tag the rest of them if possible. Looking at the query parameter setting, maybe it can be similar such as a verbose headers flag.

trask:
i just realized that query param/fragment are already part of `http.url`. do you need those captured separately as well? i'm kind of wondering if we can drop those config settings i linked to above (and associated attributes), since they aren't defined by the [http semantic attributes spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#common-attributes)

for standard headers, there's a couple of headers (Host, User-Agent) already spec'd out that we definitely want to capture

for capturing all headers, what would you propose for the span attribute name / value format? also, ideally we'd like to go through the spec process for new span attributes like this

in the meantime, if we want to go through spec process for this, we've been discussing the ability to subclass the `HttpServerTracer` implementations in custom builds of the javaagent which could be used to implement additional behavior like this

FrankSpitulski:
It's helpful to have them as separate tags so I'd like to keep that behaviour. For the name value pairs I'd suggest http.headers.{header}.

FrankSpitulski:
What sort of plans are there for this subclassing? Will users have to fork and repackage the repo?

iNikem:
I am very reluctant of adding all headers as span attributes. They may contain sensitive data such as authorisation tokens and even password in case of Basic Auth. It will not be trivial to know which headers to skip.

iNikem:
> What sort of plans are there for this subclassing? Will users have to fork and repackage the repo?

[...]

PR #827, opened 2020-07-28
feat(tags): http

FrankSpitulski:
flavour, user agent, client ip

I had to make a new type of tracer method that takes both the request and the connection because netty has the flavour on the request while armeria has the flavour on the connection. Also, I needed a fallback for the client ip which relies on the connection.

Closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/776


FrankSpitulski:
fixing tests, but I figured I'd put it up here for people to review

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +201,44 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+  protected void onConnectionAndRequest(Span span, CONNECTION connection, REQUEST request) {
+    String flavor = flavor(connection, request);
+    if (flavor != null) {
+      SemanticAttributes.HTTP_FLAVOR.set(span, flavor);
+    }
+    SemanticAttributes.HTTP_CLIENT_IP.set(span, clientIP(connection, request));
+  }
+
+  protected String clientIP(CONNECTION connection, REQUEST request) {
+    // try Forwarded
```
trask, line 213: do you think we need to support this header?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +201,44 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+          return split[1];
+        }
+      }
+    }
+
+    // try X-Forwarded-For
+    forwarded = requestHeader(request, "X-Forwarded-For");
+    if (forwarded != null) {
+      // may be split by ,
+      String[] split = forwarded.split(", *", 2); // only need the first
```
trask, line 232: should be slightly faster to split by `,` and then trim result

agent-bootstrap/src/main/java/io/opentelemetry/auto/typedspan/DelegatingSpan.java
```diff
@@ -91,6 +91,11 @@ public void setStatus(final Status status) {
     delegate.setStatus(status);
   }
 
+  @Override
+  public void recordException(Throwable throwable) {
+    delegate.recordException(throwable);
+  }
+
```
trask, line 98: ya, fix should be merged shortly in #824 👍

instrumentation/netty/netty-3.8/src/main/java/io/opentelemetry/auto/instrumentation/netty/v3_8/server/NettyHttpServerTracer.java
```diff
@@ -67,6 +72,11 @@ protected String peerHostIP(final Channel channel) {
     return null;
   }
 
+  @Override
+  protected String flavor(Channel channel, HttpRequest request) {
+    return null; // TODO flavor is on the request, not the connection for this tracer
```
trask, line 77: does this work?
```suggestion
    return request.getProtocolVersion().toString();
```

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +201,44 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+    }
+    SemanticAttributes.HTTP_CLIENT_IP.set(span, clientIP(connection, request));
+  }
+
+  protected String clientIP(CONNECTION connection, REQUEST request) {
+    // try Forwarded
+    String forwarded = requestHeader(request, "Forwarded");
+    if (forwarded != null) {
+      // may be split by ; or ,
+      for (String forwardedValues : forwarded.split(";|(, *)")) {
```
anuraaga, line 217: `split(regex)` is really slow, if we need a regex we should use `Pattern.compile` at least. But it'd be best if we can avoid split too due to the array allocations, do you think we can loop through to find the indexes instead of splitting?

@trask By the way, is it ok to use Guava here? `CharMatcher` can help with such searches.

trask: > is it ok to use Guava here?

Currently we don't, maybe better not to add this to our instrumentation api contract. We could expose something in our instrumentation api that wraps specific guava functionality if we need. I'd probably lean towards doing the index looping here for now (assuming we need to support the "Forwarded" header).

anuraaga: Ah forgot this is the instrumentation API, also used in library instrumentation. Agree not to add dependencies.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
FrankSpitulski: yeah, mozilla says it's the real standard while x-forwarded-for is the defacto standard. best to get both imo. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

instrumentation/netty/netty-3.8/src/main/java/io/opentelemetry/auto/instrumentation/netty/v3_8/server/NettyHttpServerTracer.java
FrankSpitulski: yep, I just missed this one when I added the request param.

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
FrankSpitulski: I'll change it to pattern.compile and incorporate the trim optimisation

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
FrankSpitulski: I got rid of the regex in the x-forwarded-for case

agent-bootstrap/src/main/java/io/opentelemetry/auto/typedspan/DelegatingSpan.java
FrankSpitulski: rebased

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +205,48 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+      }
+    }
+
+    // try X-Forwarded-For
+    forwarded = requestHeader(request, "X-Forwarded-For");
+    if (forwarded != null) {
+      // may be split by ,
+      int endIndex = forwarded.indexOf(',');
+      if (endIndex > 0) {
+        forwarded = forwarded.substring(0, endIndex);
```
trask, line 239: did we need trim here?
```suggestion
        forwarded = forwarded.substring(0, endIndex).trim();
```

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +205,48 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+  }
+
+  protected String clientIP(CONNECTION connection, REQUEST request) {
+    // try Forwarded
+    String forwarded = requestHeader(request, "Forwarded");
+    if (forwarded != null) {
+      // may be split by ; or ,
+      for (String forwardedValues : FORWARDED_SPLIT_PATTERN.split(forwarded)) {
+        forwardedValues = forwardedValues.trim();
+        String[] split = EQUALS_PATTERN.split(forwardedValues, 2);
```
trask, line 223: can you convert `EQUALS_PATTERN` to indexOf/substring to avoid that regex? I don't think that would add code complexity.

replacing `FORWARDED_SPLIT_PATTERN` would definitely add code complexity, so i don't mind keeping

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
FrankSpitulski: no, there aren't going to be any spaces before the first comma and we only need the first value

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
anuraaga: Out of curiosity, what do you think about this? I noticed the lines of code are very close to what we have right now even though we define a helper method.

```
int forIndex = forwarded.indexOf("for=");
int endIndex = findForwardedEnd(forwarded, forIndex + 4);
forwarded.substring(forIndex + 4, endIndex);

int find forwardedEnd(String s, int start) {
  for (int i = start; i < s.length(); i++) {
    char c = s.charAt(i);
    if (c == ',' || c == ';') {
      return i;
    }
  }
  return s.length()
```

FrankSpitulski: @anuraaga this would work for regular cases but would not catch a malformed header such as "for=, for=1.1.1.1" That's probably okay though. I will use your implementation.

FrankSpitulski:
@trask I think this may be a sporadic test failure, works on my machine and it's a timeout error 🤷 

https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-instrumentation/2264/workflows/139f7258-b674-4f8f-abfc-87311eecdd25/jobs/14878/tests

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +201,59 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+      if (!forwarded.isEmpty()) {
+        return forwarded;
+      }
+    }
+
+    // fallback to peer IP if there are no proxy headers
+    return peerHostIP(connection);
+  }
+
+  private static String extractForwardedFor(String forwarded) {
```
trask, line 239: Can you add unit test for this method (to mitigate against the extra complexity we asked for :smile:)?

FrankSpitulski: added tests, caught an edge case

FrankSpitulski: getting another issue with latestDepTest though.

./gradlew :instrumentation:grpc-1.5:latestDepTest fails
./gradlew :instrumentation:grpc-1.5:test works

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/decorator/HttpServerTracer.java
```diff
@@ -195,6 +201,63 @@ protected void onRequest(final Span span, final REQUEST request) {
...
+      }
+    }
+
+    // fallback to peer IP if there are no proxy headers
+    return peerHostIP(connection);
+  }
+
+  // VisibleForTesting
+  static String extractForwardedFor(String forwarded) {
+    int start = forwarded.toLowerCase().indexOf("for=");
```
anuraaga, line 241: Since the spec seems to use lowercase, and it's not a commonly used header anyways, I probably wouldn't be so permissive and just not do the `toLowerCase` but it's ok too. From what I can tell, we consider it unimportant enough to not have unit tests :P

FrankSpitulski: the mozilla spec says it can be any case, so I just stuck that in there. I've never seen it used in practice but it's apparently the standard.

FrankSpitulski: there's an uppercase example in the RFC https://tools.ietf.org/html/rfc7239#section-4

anuraaga: Ah thanks I couldn't find it, in that case can we add one test case for it?

FrankSpitulski: sure, added

trask:
I forgot that I can merge master into your branch 😄

PR #829, opened 2020-07-29
Remove unnecessary 3rd party license file

### LINKED ISSUE #826: Github is not detecting license for this project
Github is not detecting/displaying the license for this project:

https://github.com/open-telemetry?q=java&type=&language=

Also you can see the difference in these two pages:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/LICENSE

https://github.com/open-telemetry/opentelemetry-java/blob/master/LICENSE

trask:
Testing on a forked repo leads me to believe this will resolve #826


trask:
> Do we need to get approval from anyone at CNCF?

This file is woefully outdated anyways, so let's go ahead and remove it, and I opened #836, and marked it `required-for-ga` `p1`, to seek guidance from community repo / CNCF since we embed our dependencies which makes the CNCF requirements a bit less clear to me.

PR #839, opened 2020-07-30
Fix tests on newly released grpc version 1.31.0

trask:


trask:
I had to split out the `forAddress` and the `build` instrumentation because the new `declaresField(named("interceptors"))` restriction prevented the `forAddress` instrumentation from being applied where it was needed.

Issue #831, opened 2020-07-29
Document test bytecode manipulation infrastructure


anuraaga:
I dove into `SpockRunner`, `AgentSpecification` yesterday, primarily with the goal of understanding why we need to have separate `test` and `integrationTest` - our agent goes to so much effort to not affect a user's app, I sort of would expect the same machinery to work for tests.

I think some documentation on what's going on in the tests can be helpful, I'll jot down my notes from digging and questions here

`SpockRunner` has several functions
- "Shadow" the test class - it causes the test class to be loaded in a custom classloader instead of junit's default which is the app classloader
- It sets this custom classloader as the context classloader, presumably for servlet engines like tomcat
- Setup the bootstrap classpath. This seems to be injecting the same classes into the bootstrap loader which our agent would. Primarily the OpenTelemetry API, logging, and servlet
- It ensures there aren't any bootstrap classes in test classes
- Initializes `ByteBuddy` by calling `ByteBuddy.install`. This creates an `Instrumentation` object which can be used by `AgentTestRunner` to install the agent and reset it

`AgentSpecification` seem to only be to override `private static final Config INSTANCE` to be public and non-final.

Some questions

- The custom classloader in `SpockRunner` still delegates to `SpockRunner.class.getClassLoader()`, the app classloader. What is the purpose of `shadowTestClass`?
- `SpockRunner` asserts there are no bootstrap classes in test classes. Why do we need this check? Why do I feel this has never restricted me when writing tests? `AgentTestRunner`, inherited by the test classes, has a field of type `Tracer` which seems to work ok

And some findings

- Trying to add `SpockRunner` to `AgentSpecification` complains with the bootstrap class check, because tests have a field of type `Tracer`. I'm still not sure why this isn't allowed but it works ok for `AgentTestRunner`
- Trying to add `SpockRunner` to `AgentSpecification` and temporarily disable the bootstrap class check results in missing classes like `org.groovy.lang.Object`. Since the custom classloader created by `SpockRunner` delegates to the app classloader, not sure why this would happen

trask:
@anuraaga ok to close this as resolved by #1643?

PR #840, opened 2020-07-30
Generate bootstrap jar with gradle instead of when running tests.

anuraaga:
Currently, we create an entire JAR file at runtime every time we run a test, and this takes many seconds. This changes it so it's Gradle that generates and sets up the bootstrap classpath, not `SpockRunner`. It's essentially the same as the previous state otherwise.

A clean build may be a bit faster than before, not sure. But more importantly, running the same test over and over goes much faster since the bootstrap jar is up-to-date by Gradle. (I have been running tests over and over a lot recently :P).

I went with a plugin in Java mainly to be able to continue to copy-paste from `Constants.java`. I find it easier to maintain personally too.

It took a while to get this to work (finding the use of `files(Callable)` was a doozy) so want to merge and follow up with even more optimizations.

- There seem to be only two types of bootstrap jars, normal and normal+servlet. I should be able to tweak it so only these two jars are ever created, which will speed up clean build substantially.

- See if we can loosen some restrictions now that the bootstrap classpath is set up at startup instead of at runtime (maybe can merge test and integrationTest back into one? not sure haven't tried yet but I think it might be possible).

- Remove the unit test bootstrap proxy from our agent code, less test hacks in business logic is good :D


iNikem:
Can you remind me, why our tests do that in the first place? Why do have some special jar in bootstrap class loader?

anuraaga:
I think, but don't know hence #831, that it's where we need classes to go that are used both by the agent and the user app since it's the only common ClassLoader parent.

buildSrc/build.gradle.kts
```diff
@@ -10,6 +10,10 @@ gradlePlugin {
       id = "muzzle"
       implementationClass = "MuzzlePlugin"
     }
+    create("auto-instrumentation-plugin") {
+      id = "io.opentelemetry.auto.auto-instrumentation"
+      implementationClass = "io.opentelemetry.auto.gradle.instrumentation.AutoInstrumentationPlugin"
```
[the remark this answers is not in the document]
anuraaga: Yeah - I did put about a minute into it but couldn't come up with a better package. Let me know if you have a suggestion so I can use it :)

buildSrc/src/main/java/io/opentelemetry/auto/gradle/instrumentation/AutoInstrumentationPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+      "org.slf4j",
+      "ch.qos.logback",
+      // Tomcat's servlet classes must be on boostrap
+      // when running tomcat test
+      "javax.servlet.ServletContainerInitializer",
+      "javax.servlet.ServletContext"
+    };
+    TEST_BOOTSTRAP_PREFIXES =
+        Arrays.copyOf(
+            BOOTSTRAP_PACKAGE_PREFIXES_COPY,
```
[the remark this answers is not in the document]
anuraaga: I think it's so one can be pure copy-paste from the business logic.

buildSrc/src/main/java/io/opentelemetry/auto/gradle/instrumentation/AutoInstrumentationPlugin.java
```diff
@@ -0,0 +1,146 @@
...
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.file.DuplicatesStrategy;
+import org.gradle.api.plugins.JavaLibraryPlugin;
+import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.tasks.TaskProvider;
+import org.gradle.api.tasks.testing.Test;
+import org.gradle.jvm.tasks.Jar;
+
+public class AutoInstrumentationPlugin implements Plugin<Project> {
```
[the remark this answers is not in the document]
anuraaga: I added a bit, I'll add some more when moving more logic into here, this and instrumentation.gradle are serving the same layer so I may as well move everything into here.

anuraaga:
Oops forgot to check `latestDepTest` locally will look at it

buildSrc/build.gradle.kts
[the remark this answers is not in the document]
anuraaga: Tried a new name let me know :)

anuraaga:
Very strange, works great on Mac and even checked docker with the datadog builder image and can't repo any failures locally. Will do some debug commits

anuraaga:
I have always blatantly ignored Gradle's advice to use `buildDir` instead of just the string `build` and it finally came to haunt me 👻

testing-common/src/test/groovy/context/FieldBackedProviderTest.groovy
```diff
@@ -177,8 +180,12 @@ class FieldBackedProviderTest extends AgentTestRunner {
 
   def "context classes are redefine safe"() {
     when:
-    ByteBuddyAgent.getInstrumentation().redefineClasses(new ClassDefinition(KeyClass, ClasspathUtils.convertToByteArray(KeyClass)))
-    ByteBuddyAgent.getInstrumentation().redefineClasses(new ClassDefinition(UntransformableKeyClass, ClasspathUtils.convertToByteArray(UntransformableKeyClass)))
+    try {
```
anuraaga, line 183: @trask This started throwing an exception even though the assertions pass, any idea if this is a valid fix?

iNikem: It seems strange, that test named "context classes are redefine safe" is expected to throw an exception.

trask: i don't think it's a valid fix 😄

i tried it out, looks like jacoco instrumentation is getting applied to `KeyClass` and `UntransformableKeyClass` in your branch, which is causing the redefinition to fail, e.g.:

```
  def "context classes are redefine safe"() {
    when:
    for (Field field : KeyClass.getDeclaredFields()) {
      println field.getName()
    }
```

shows `$jacocoData` in your branch, but no jacoco field in master.

it looks like those classes are supposed to be excluded from jacoco:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/testing-common/testing-common.gradle#L13-L14

so maybe your changes broke that exclusion somehow?

anuraaga: Thanks for the debugging! Yeah that must be it I see what needs to be fixed. For reference though, does this mean that this behavior is only guaranteed when there isn't another agent running? I wonder how many of our features have that restriction.

anuraaga: Also I don't know why `UntransformableKeyClass` is not transformable that would be helpful if you know :)

testing-common/src/test/groovy/context/FieldBackedProviderTest.groovy
```diff
@@ -175,6 +178,9 @@ class FieldBackedProviderTest extends AgentTestRunner {
     new UntransformableKeyClass().incrementContextCount() == 1
   }
 
+  // NB: This test will fail if some other agent is also running that fills the class structure
+  // before we can. A likely culprit is jacoco if you start seeing failure here due to a change
+  // make sure jacoco exclusion is working.
```
trask, line 183:
```suggestion
  // NB: This test will fail if some other agent is also running that modifies the class structure
  // in a way that is incompatible with redefining the class back to its original bytecode.
  // A likely culprit is jacoco if you start seeing failure here due to a change make sure jacoco
  // exclusion is working.
```

testing-common/src/test/java/context/ContextTestInstrumentation.java
```diff
@@ -194,7 +194,12 @@ public void putContextCount(final int value) {
     }
   }
 
-  /** A class which cannot be transformed by our instrumentation. */
+  /**
+   * A class which cannot be transformed by our instrumentation. {@link KeyClass} is an empty object
+   * meaning it has one field available for us to add context through redefinition. {@link
+   * UntransformableKeyClass} has an implicit reference to its parent class, {@link KeyClass} and
+   * thus has no available space for us to add context via redefinition.
```
trask, line 201:
```suggestion
   * A class which will not be transformed by our instrumentation due to {@link
   * FieldBackedProviderTest#shouldTransformClass(String, ClassLoader)}.
```

anuraaga:
I haven't measured timings in detail but anecdotally, when I run a test in IntelliJ every time it takes between 5 and 10 seconds between Gradle task starting and instrumentation debug spam starting, while now that is just gone.

PR #842, opened 2020-07-30
Rename trace.methods.exclude config

### LINKED ISSUE #760: Rename trace.methods.exclude to trace.annotated.methods.exclude
And to change our doc to speak about two exclude options separately

* trace.classes.excludes - applies to all instrumentation, useful for:
  * completely silence spans from a given class/package
  * quick workaround for an instrumentation bug, when byte code in one specific class is problematic.
* trace.annotated.methods.exclude - applies only to annotation instrumentation, useful for:
  * if you have over-instrumented your code with these annotations

trask:
Closes #760

I included `README.md` changes directly in this PR since this config setting is not very common, so I think ok for it's docs to be out of sync until next week, but I can separate out `README.md` change if anyone prefers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,15 +274,14 @@ public class MyClass {
 Each time the application invokes the annotated method, it creates a span
 that denote its duration and provides any thrown exceptions.
 
-#### Configuration
+#### Suppressing `@WithSpan` instrumentation
 
-The `@WithSpan` annotation requires code changes to implement. You can
-disable the annotation at runtime via the exclude configuration or
-environment variables:
+This is useful in case you have code that is over-instrumented using `@WithSpan`,
+and you want to suppress some of them without modifying the code.
 
-| System property                 | Environment variable            | Purpose                                                                                                                         |
-|---------------------------------|---------------------------------|---------------------------------------------------------------------------------------------------------------------------------|
-| trace.annotated.methods.exclude | TRACE_ANNOTATED_METHODS_EXCLUDE | Exclude methods with the `@WithSpan` annotation, the format "my.package.MyClass1[method1,method2];my.package.MyClass2[method3]" |
+| System property                 | Environment variable            | Purpose                                                                                                                                  |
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -276,10 +280,9 @@ The `@WithSpan` annotation requires code changes to implement. You can
 disable the annotation at runtime via the exclude configuration or
 environment variables:
 
-| System property                  | Environment variable             | Purpose                                                              |
-|----------------------------------|----------------------------------|----------------------------------------------------------------------|
-| trace.classes.exclude            | TRACE_CLASSES_EXCLUDE            | Exclude classes with the `@WithSpan` annotation                      |
```
anuraaga, line 281: Isn't this property still there?

trask: I moved documentation of that property to the "Suppressing specific auto-instrumentation" section above (which links to a new doc page)

anuraaga: Ah sorry missed that markdown in the middle of the java files!

docs/suppressing-instrumentation.md
```diff
@@ -0,0 +1,26 @@
...
+
+This option should not be used lightly, as it can leave some instrumentation partially applied,
+which could have unknown side-effects.
+
+If you find yourself needing to use this, it would be great if you could drop us an issue explaining why,
+so that we can try to come up with a better solution to address your need.
+
+| System property       | Environment variable  | Purpose                                                                                                                         |
+|-----------------------|-----------------------|---------------------------------------------------------------------------------------------------------------------------------|
+| trace.classes.exclude | TRACE_CLASSES_EXCLUDE | Exclude methods with the `@WithSpan` annotation, the format "my.package.MyClass1[method1,method2];my.package.MyClass2[method3]" |
```
iNikem, line 26: Documentation is wrong. We exclude classes here, not methods :)

trask: fixed!

PR #844, opened 2020-07-30
Faster spotless

### LINKED ISSUE #736: Markdown spotless should also be per-project
I always get confused by the fact that all markdown in the repo is checked by the top level spotless check, even when the file is in a subproject. Presumably this also reduces the chance for up-to-date checks to work. Is it possible to make markdown in a subproject also checked by the subproject's spotless?

trask:
Closes #736

The change in `instrumentation.gradle` was so that `spotless.gradle` wouldn't get applied at that level (which was also slow).

EDIT this ^^^ didn't work, re-working...

### DIFF
--- a/gradle/spotless.gradle
+++ b/gradle/spotless.gradle
@@ -19,6 +19,12 @@ spotless {
     ktlint().userData(['indent_size': '2', 'continuation_indent_size': '2'])
     licenseHeaderFile rootProject.file('gradle/enforcement/spotless.license.java'), '(package|import|public)'
   }
+  format 'misc', {
+    target '**/.gitignore', '**/*.md', '**/*.sh'
+    indentWithSpaces()
+    trimTrailingWhitespace()
+    endWithNewline()
+  }
 }
 
 task formatCode(dependsOn: ['spotlessApply'])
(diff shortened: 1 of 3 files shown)


gradle/spotless.gradle
```diff
@@ -19,6 +19,12 @@ spotless {
     ktlint().userData(['indent_size': '2', 'continuation_indent_size': '2'])
     licenseHeaderFile rootProject.file('gradle/enforcement/spotless.license.java'), '(package|import|public)'
   }
+  format 'misc', {
+    target '.gitignore', '*.md', 'docs/**/*.md', '*.sh'
```
iNikem, line 23: We don't have `docs` in modules, do we? And shouldn't it be `**/*.md`?

trask: > We don't have `docs` in modules, do we?

not yet (and maybe never), so removed for now

> And shouldn't it be `**/*.md`?

not using `**/` patterns to keep spotless fast, added a comment here about it

iNikem: So that we will format md files only in the root of modules? Is it ok?

trask: interesting, we do have some md files not in the root of modules:

```
./agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/instrumentation/jdbc/README.md
./agent-bootstrap/src/test/resources/classloader-test-jar/testjar-readme.md
```

how about I limit it to `*.md,src/**/*.md`?

iNikem: That's good

PR #857, opened 2020-07-31
Create issue if nightly build fails

### LINKED ISSUE #753: Create an issue if nightly build fails
We postpone several of mandatory code checks to nightly builds, e.g. #748 . But nightly build failures are very easy to ignore. Should at least create a new issue whenever build fails. Probably requires #678

iNikem:
Closes #753

### DIFF
--- a/.github/workflows/nightly-failed.md
+++ b/.github/workflows/nightly-failed.md
@@ -1,0 +1,6 @@
+---
+title: Nightly build failed
+assignees: iNikem
+labels: bug, area:build, priority:p1
+---
+Nightly job failed. Please take a look and fix it ASAP
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -97,3 +97,20 @@ jobs:
 
       - name: Test with Gradle
         run: ./gradlew latestDepTest
+
+  issue:
+    name: Open issue on failure
+    needs: [test, testLatestDep]
+    runs-on: ubuntu-latest
+    if: always()
+    steps:
+      # run this action to get workflow conclusion
+      # You can get conclusion by env (env.WORKFLOW_CONCLUSION)
+      - uses: technote-space/workflow-conclusion-action@v1
+      - uses: actions/checkout@v2
+      - uses: JasonEtco/create-an-issue@v2
+        if: env.WORKFLOW_CONCLUSION == 'failure' # notify only if failure
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          filename: .github/workflows/nightly-failed.md


.github/workflows/nightly-failed.md
```diff
@@ -0,0 +1,6 @@
+---
+title: Nightly build failed
+assignees: iNikem
```
anuraaga, line 3: Is it only possible to assign one person to assignee? If not, maybe can assign the team if it's possible, or otherwise the manually maintained list of the maintainers if that's possible. Otherwise... thanks! ;)

.github/workflows/nightly.yaml
```diff
@@ -97,3 +97,20 @@ jobs:
 
       - name: Test with Gradle
         run: ./gradlew latestDepTest
+
+  issue:
+    name: Open issue on failure
+    needs: [test, testLatestDep]
+    runs-on: ubuntu-latest
```
anuraaga, line 104: Does this require Ubuntu?

.github/workflows/nightly-failed.md
iNikem: Several assignees can be defined. Don't know about a team. I am OK for now with being single assignee or I can add you and @trask, if you want :)

.github/workflows/nightly.yaml
iNikem: Don't know actually. Just copy-paste from examples and it's the same as in all other jobs. Is it important?

.github/workflows/nightly-failed.md
anuraaga: I can't speak for @trask but go ahead and add me :)

.github/workflows/nightly.yaml
anuraaga: My impression is JavaScript will work faster than docker but that's it so it's not too important.

.github/workflows/nightly-failed.md
trask: does it need an assignee? or can we have no assignee, and then whichever one of us picks it up can assign ourselves so we don't duplicate effort fixing

iNikem: I propose to leave it as it is, assigning to me (it will help me react faster), but you are more than welcomed to steal these issue from me if you reach them earlier :)

trask: that works for me, what time does the nightly build run, and should we optimize that timing for you?

iNikem: 03:24 UTC. That is usual nightly time for me :) So both you and Anuraag can grab it immediately, or I will wake up and deal with it.

trask: > or I will wake up and deal with it

please don't 😄

and feel free to move the timing of the nightly build any time you want

iNikem: I mean "wake up in my usual time in the morning and will deal with it in a due course during my work day" :)

PR #859, opened 2020-07-31
Fix some sporadic test failures

trask:
Based on @iNikem's findings in #787.

CI results seem promising.

But going to look a bit more into preventing the context leak in the first place using #860.


iNikem:
@trask any reasons why you don’t want to merge this for now?

trask:
Closing, we decided to chase down thread pool leakage issues using new new context leak debugger added in #860

PR #867, opened 2020-08-01
Retry test steps of nightly builds

### LINKED ISSUE #862: Nightly build should rerun a failed build a couple of times
Now that we file an issue on failure we will deal with a lot of issues due to our current problem of sporadic test failures. Since it only runs once a night anyways, it's ok for the build to take some time so we should probably add some retry functionality to it to reduce flakiness.

iNikem:
Closes #862


trask:
nice 👍

PR #871, opened 2020-08-02
Merge integration tests back into test.

anuraaga:
#840 seems to have fixed the issue where unit tests and instrumentation tests had to be kept separate. This is probably because previously it was possible for a unit test to load a class into the app class loader, and an instrumentation would later try to load it into bootstrap and fail. Now, the bootstrap classpath is setup at startup time so there is no chance of an ordering conflict.

It could be argued that it's still useful to have separation between "small tests" and "large tests" but I think we'll get that naturally when moving more to the library/auto pattern, and the `Category` adds complexity that probably isn't paying itself off if we don't require the separation. In practice, our overall build should be a tiny bit faster with less tasks to initialize.


iNikem:
Thank you :)

Issue #821, opened 2020-07-28
Failed to transform class files


iNikem:
Sometimes our builds fail with the following exception:

```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':instrumentation:play:play-2.3:compileJava'.
> Failed to transform class files in /home/runner/work/opentelemetry-java-instrumentation/opentelemetry-java-instrumentation/workspace/instrumentation/play/play-2.3/build/classes/java/main
```

I suspect this to be muzzle plugin which injects some code during class compilation.

Example failures:

https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-instrumentation/2232/workflows/75c8327e-befa-45b5-8e97-6e8afde0da89/jobs/14768

trask:
Yeah, I've seen the error linked in the report above frequently also:

```
Caused by: java.util.zip.ZipException: ZipFile closed
```

I'm interested to see if this somehow resolves itself by switching over to GitHub Actions.

mateuszrzeszutek:
@trask @iNikem I think we can close this one - I don't think we had any muzzle/bytebuddy transformation issues after switching ot GHA.

PR #873, opened 2020-08-02
Add --stacktrace to nightly build commands

anuraaga:
A final stacktrace on failure is almost never particularly noisy compared to the rest of the build but can help find out about issues like #821. I often wonder why Gradle doesn't default to it...


anuraaga:
Yeah back in the day people used it on CI but now Gradle always spews a warning presumably because the cost of spawning and not reusing the daemon is so small. So I don't bother :)

PR #876, opened 2020-08-03
Use assertJ in tests

mabdinur:
Use assertJ assertions instead of junit5

### DIFF
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
@@ -59,7 +59,7 @@ class TracerAutoConfigurationTest {
         .withConfiguration(AutoConfigurations.of(TracerAutoConfiguration.class))
         .run(
             (context) -> {
-              assertThat(context.containsBean("otelTracer")).isEqualTo(true);
+              assertThat(context.containsBean("otelTracer")).isTrue();
             });
   }
 
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
@@ -16,12 +16,11 @@
 
 package io.opentelemetry.instrumentation.spring.autoconfigure;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
+import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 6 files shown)


jkwatson:
Rather than adding a few assertJ tests, I'd rather continue to use Truth until we convert everything at once. I definitely don't want to have both Truth and assertJ at the same time.

anuraaga:
This is the `opentelemetry-java-instrumentation` repo we don't have any Truth here yet. Since Spring uses assertj natively (we don't even have to import it since it's included in their test starter) I'd like to keep consistency with Spring for these modules rather than SDK project here if that sounds OK.

instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
```diff
@@ -41,45 +40,39 @@ public Tracer customTestTracer() {
...
         .run(
             (context) -> {
-              assertTrue(
-                  "Application Context contains customTestTracer bean",
-                  context.containsBean("customTestTracer"));
-
-              assertFalse(
-                  "Application Context DOES NOT contain the otelTracer bean defined in TracerAutoConfiguration",
-                  context.containsBean("otelTracer"));
+              assertThat(context.containsBean("customTestTracer")).isEqualTo(true);
```
anuraaga, line 50:
```suggestion
              assertThat(context.containsBean("customTestTracer")).isTrue();
```

instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
```diff
@@ -41,45 +40,39 @@ public Tracer customTestTracer() {
...
+              assertThat(context.containsBean("otelTracer")).isEqualTo(false);
```
anuraaga, line 51:
```suggestion
              assertThat(context.containsBean("otelTracer")).isFalse();
```

instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfigurationTest.java
```diff
@@ -41,45 +40,39 @@ public Tracer customTestTracer() {
...
+  @DisplayName("when Application Context DOES NOT contain Tracer bean should initialize otelTracer")
+  void initalizeTracer() {
     this.contextRunner
         .withConfiguration(AutoConfigurations.of(TracerAutoConfiguration.class))
         .run(
             (context) -> {
-              assertTrue(
-                  "Application Context contains otelTracer bean",
-                  context.containsBean("otelTracer"));
+              assertThat(context.containsBean("otelTracer")).isEqualTo(true);
```
anuraaga, line 62:
```suggestion
              assertThat(context.containsBean("otelTracer")).isTrue();
```

jkwatson:
oh yes...Very sorry. I got confused as to what repo I was looking at.  Never Mind Me.  ;)

PR #878, opened 2020-08-03
clean up spring dependencies

mabdinur:
1. Apply gradle/instrumentation-library.gradle to spring library instrumentation
2. Change spring api/implementation dependencies to compileOnly 
3. Exclude `org.slf4j.slf4j-simple` from agent-bootstrap (agent-bootstrap exposes sl4j-simple artifact and this can conflict with loggers in a user's application)

Next steps will include:
- following the instructions in [writing-instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/writing-instrumentation.md) to move library instrumentation from instrumentation-core to instrumentation/..../library

### DIFF
--- a/gradle/instrumentation-library.gradle
+++ b/gradle/instrumentation-library.gradle
@@ -15,7 +15,9 @@ dependencies {
   // TODO(anuraaga): We currently include common instrumentation logic like decorators in the
   // bootstrap, but we need to move it out so manual instrumentation does not depend on code from
   // the agent, like Agent.
-  api project(':auto-bootstrap')
+  api project(':auto-bootstrap'){
+    exclude group: 'org.slf4j', module: 'slf4j-simple'
+  }
 
   api deps.opentelemetryApi
 
--- a/gradle/instrumentation-library.gradle
+++ b/gradle/instrumentation-library.gradle
@@ -7,8 +7,6 @@ group = 'io.opentelemetry.instrumentation'
 apply from: "$rootDir/gradle/java.gradle"
 apply from: "$rootDir/gradle/publish.gradle"
 
-apply plugin: 'com.github.johnrengelman.shadow'
-
 archivesBaseName = projectDir.parentFile.name
 
 dependencies {
@@ -24,12 +22,17 @@ dependencies {
   testImplementation project(':testing-common')
 }
 
-shadowJar {
-  archiveClassifier = 'agent'
+if (!ext.properties.noShadow) {
 
-  configurations = []
+  apply plugin: 'com.github.johnrengelman.shadow'
 
-  relocate "io.opentelemetry.instrumentation.${javaSubPackage}", "io.opentelemetry.auto.instrumentation.${javaSubPackage}.shaded"
+  shadowJar {
+    archiveClassifier = 'agent'
+
+    configurations = []
+
+    relocate "io.opentelemetry.instrumentation.${javaSubPackage}", "io.opentelemetry.auto.instrumentation.${javaSubPackage}.shaded"
+  }
 }
 
 afterEvaluate {
--- a/gradle/instrumentation-library.gradle
+++ b/gradle/instrumentation-library.gradle
@@ -15,7 +15,7 @@ dependencies {
   // TODO(anuraaga): We currently include common instrumentation logic like decorators in the
   // bootstrap, but we need to move it out so manual instrumentation does not depend on code from
   // the agent, like Agent.
-  api project(':auto-bootstrap'){
+  api(project(':auto-bootstrap')){
     exclude group: 'org.slf4j', module: 'slf4j-simple'
   }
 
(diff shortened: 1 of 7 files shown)


gradle/instrumentation-library.gradle
```diff
@@ -15,7 +15,9 @@ dependencies {
   // TODO(anuraaga): We currently include common instrumentation logic like decorators in the
   // bootstrap, but we need to move it out so manual instrumentation does not depend on code from
   // the agent, like Agent.
-  api project(':auto-bootstrap')
+  api project(':auto-bootstrap'){
+    exclude group: 'org.slf4j', module: 'slf4j-simple'
+  }
```
mabdinur, line 21:
```suggestion
  api deps.opentelemetrySdk
```
Would it make sense to add this dependency or are we giving user's the flexibility to supply their own sdk

trask: i think leave it to users, since they have to pull in sdk to configure exporters anyways

instrumentation-core/servlet/servlet.gradle
```diff
@@ -1,14 +1,11 @@
-group = 'io.opentelemetry.instrumentation'
-
-apply from: "$rootDir/gradle/java.gradle"
-apply from: "$rootDir/gradle/publish.gradle"
+ext {
+  noShadowPublish = false
```
trask, line 2: i think the shadow jar is only for use by the corresponding auto-instrumentation modules, so i think we don't want to publish the shadow jar for anything under `:instrumentation-core`
```suggestion
```

instrumentation-core/servlet/servlet.gradle
```diff
@@ -1,14 +1,10 @@
-group = 'io.opentelemetry.instrumentation'
-
-apply from: "$rootDir/gradle/java.gradle"
-apply from: "$rootDir/gradle/publish.gradle"
+ext {
+  javaSubPackage = ''
```
trask, line 2: this is used by servlet auto-instrumentation. and i think you'll also need to update those auto-instrumentation to use `configuration: 'shadow'`, and use shaded package name
```suggestion
  javaSubPackage = 'servlet'
```

instrumentation-core/spring/spring-web-3.1/spring-web-3.1.gradle
```diff
@@ -1,25 +1,9 @@
-group = 'io.opentelemetry.instrumentation'
+ext {
+  javaSubPackage = ''
+}
```
trask, line 3: maybe something more explicit which disables the `shadowJar` task in `instrumentation-library.gradle`, e.g.

we could also have lack of `javaSubPackage` disable the `shadowJar` task, but don't want people to accidentally forget about shading these dependencies for auto-instrumentation, so think maybe this is safer

```suggestion
  noShadow = true
```

instrumentation-core/spring/spring-webmvc-3.1/spring-webmvc-3.1.gradle
```diff
@@ -1,24 +1,12 @@
-group = 'io.opentelemetry.instrumentation'
-
-apply from: "$rootDir/gradle/java.gradle"
-apply from: "$rootDir/gradle/publish.gradle"
+ext {
+  javaSubPackage = ''
+}
```
trask, line 3: same comment as above

instrumentation-core/spring/spring-web-3.1/spring-web-3.1.gradle
trask: oh sorry, i accidentally removed the surrounding `ext { ... }`

mabdinur: I run into this error when I don't define javaSubPackage:

```
A problem occurred evaluating script.
> Could not get unknown property 'javaSubPackage'
```
Is there a way to disable the shadowJar configuration in instrumentation-library.gradle so that I don't run into this compile time error? A work around would be to just set javaSubPackage to an empty string but that doesn't seem ideal.

trask: there may be a better way, but wrapping parts of `instrumentation-library.gradle` in conditionals seems to do the trick:

```
if (!ext.properties.noShadow) {
  apply plugin: 'com.github.johnrengelman.shadow'
}
```

if (!ext.properties.noShadow) {
  shadowJar {
    archiveClassifier = 'agent'

    configurations = []

    relocate "io.opentelemetry.instrumentation.${javaSubPackage}", "io.opentelemetry.auto.instrumentation.${javaSubPackage}.shaded"
  }
}

mabdinur: Awesome thanks!

PR #879, opened 2020-08-03
Remove unused code

trask:
Noticed while reviewing #877


mabdinur:
Hey, 

I think this method is used here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/spring-webflux-5.0/src/main/java/io/opentelemetry/auto/instrumentation/springwebflux/client/WebClientFilterAdvice.java

trask:
Eeek, you're right! And I forgot the idea was also for this method to be part of our public API if people didn't want to supply their own tracer 😅

PR #884, opened 2020-08-04
Extract instrumentation api module

### LINKED ISSUE #755: Adding auto instrumentation after library instrumentation requires building the shadow jar
Since auto instrumentation depends on the shaded version of the library, you need to run e.g., `:shadowJar` in the library project before IntelliJ can find the classes. We need to document this.

trask:
Part of #597

A few things:
* Introduces two new modules
  * `library-api` (`io.opentelemetry.instrumentation:opentelemetry-library-api`)
    * "library" instrumentation only depends on this (they do not depend on `auto-bootstrap` or `auto-api` below)
  * `auto-api` (`io.opentelemetry.instrumentation.auto:opentelemetry-auto-api`)
    * "auto" instrumentation no longer depends on `auto-bootstrap`. they do still depend on `auto-tooling`, we probably want to extract `auto-matcher-api` out from `auto-tooling` at some point, and remove auto instrumentation dependency on all of `auto-tooling`
* Moved `:instrumentation-core:apache-httpclient-4.0` classes back to  `:instrumentation:apache-httpclient:apache-httpclient-4.0` because they still needed `auto-api`, and weren't really standalone yet anyways

There's a couple of odd dependencies that this revealed, e.g.
* `InternalJarURLHandler` and `AgentClassLoader` shouldn't be in public API
* `Config` shouldn't be in `library-api`
* `MoreAttributes` shouldn't be in public API (we should probably just remove it and the associated attributes)

I will open issues to track these and any other odd dependencies that this extraction highlights.

Fixes #755


trask:
I still need to shade the instrumentation-api module when it is used in the agent.

We could run shadowJar in the instrumentation-api module with `archiveClassifier = 'agent'`, like we are doing in `instrumentation-library.gradle`, and consume that shadow configuration everywhere.

But then we need to publish that shadow classifier for 3rd party auto-instrumentation to use. Which is not the worst, but seems likely to cause confusion.

So thinking instead to wait and shade it in the `opentelemetry-javaagent` build (which is where we shade some other stuff anyways).

That doesn't answer how we manage 3rd party auto-instrumentation, but if we are promoting "make your own custom build" then it's not a problem. Or if we want to load dynamically, we could shade at runtime like the dynamic exporter loading.

Let me know if you have thoughts.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/SafeServiceLoader.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.auto.bootstrap.instrumentation;
+package io.opentelemetry.instrumentation.api;
```
anuraaga, line 17: Is this class needed by manual isntrumentation?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/CallDepthThreadLocalMap.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.auto.bootstrap;
+package io.opentelemetry.instrumentation.api;
```
anuraaga, line 17: Are these three classes needed by manual instrumentation?

trask:
I went with shading later in the build due to concerns I raised in #892

agent-tooling/src/main/java/io/opentelemetry/auto/tooling/DefaultExporterConfig.java
```diff
@@ -16,7 +16,7 @@
 
 package io.opentelemetry.auto.tooling;
 
-import io.opentelemetry.auto.config.Config;
+import io.opentelemetry.instrumentation.library.api.config.Config;
```
anuraaga, line 19: Since we don't have `library` in our artifact names for library, guess we wouldn't have it in this package name?

gradle/instrumentation-library.gradle
```diff
@@ -10,12 +10,7 @@ apply from: "$rootDir/gradle/publish.gradle"
 archivesBaseName = projectDir.parentFile.name
 
 dependencies {
-  // TODO(anuraaga): We currently include common instrumentation logic like decorators in the
```
anuraaga, line 13: Yay!

iNikem:
I promise I will take a look at this soon!

library-api/library-api.gradle
```diff
@@ -0,0 +1,14 @@
+group = 'io.opentelemetry.instrumentation'
+
+apply from: "$rootDir/gradle/java.gradle"
+apply from: "$rootDir/gradle/publish.gradle"
+
+dependencies {
+  api deps.opentelemetryApi
+  compileOnly deps.opentelemetrySdk
```
iNikem, line 8: Is this needed?

testing-common/src/main/groovy/io/opentelemetry/auto/test/utils/TraceUtils.groovy
```diff
@@ -18,13 +18,14 @@ package io.opentelemetry.auto.test.utils
 
 import io.grpc.Context
 import io.opentelemetry.OpenTelemetry
-import io.opentelemetry.auto.bootstrap.instrumentation.decorator.BaseDecorator
-import io.opentelemetry.auto.bootstrap.instrumentation.decorator.BaseTracer
 import io.opentelemetry.auto.test.asserts.TraceAssert
 import io.opentelemetry.context.Scope
+import io.opentelemetry.instrumentation.library.api.decorator.BaseDecorator
+import io.opentelemetry.instrumentation.library.api.decorator.BaseTracer
```
iNikem, line 24: I think `io.opentelemetry.instrumentation.api` would be better package, without `library`

iNikem:
I don't quite understand the purpose of `auto-api` module. Why is it separate from `auto-tooling`?

I think we have to update `Understanding the javaagent components` document to describe modules first.

agent-tooling/src/main/java/io/opentelemetry/auto/tooling/DefaultExporterConfig.java
trask: makes sense to me, done

library-api/library-api.gradle
trask: i hope not, removed, we'll see 😄

testing-common/src/main/groovy/io/opentelemetry/auto/test/utils/TraceUtils.groovy
trask: 👍

trask:
> I don't quite understand the purpose of `auto-api` module. Why is it separate from `auto-tooling`?

`auto-api` lives in the bootstrap class loader. `auto-tooling` lives in the `AgentClassLoader`.

> I think we have to update `Understanding the javaagent components` document to describe modules first.

👍 will do in this PR (tomorrow). I don't want to merge this PR until after #893 anyways to avoid causing merge conflicts for #893

anuraaga:
Added Fixes #755 to the PR description

iNikem:
> `auto-api` lives in the bootstrap class loader.

Then what's the difference with `agent-bootstrap`?

trask:
> Then what's the difference with `agent-bootstrap`

`auto-api` is public api surface for people writing their own auto-instrumentation

trask:
> I think we have to update `Understanding the javaagent components` document to describe modules first.

done 👍

trask:
I'm going to send merge PRs shortly to some of the existing PRs to help mitigate the conflict pain that this PR is going to cause.

Issue #846, opened 2020-07-31
Library instrumentation experience confusing because of SDK dependency requirement?


anuraaga:
Just want to throw this out there, I wonder if the requirement that instrumentation only depends on the OTel API and not SDK is too surprising to users. If someone depends on `opentelemetry-instrumentation-grpc-1.5`, it seems natural to expect there to be actual instrumentation happening but it'd be no-op without also adding `opentelemetry-sdk`. At the very least, I haven't checked yet, but we might not even have a scary log message for this case.

trask:
I think that without the user adding an exporter (which would then bring in `opentelemetry-sdk`), nothing (much) is going to happen even with `opentelemetry-sdk`

iNikem:
@anuraaga This task looks like it should have `discussion` label, shouldn't it?

trask:
hey @anuraaga do you have additional thoughts about this?

@mabdinur asked about this too: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/878#discussion_r464546751

anuraaga:
Yeah it's true that if adding the exporter, it's not a huge step to also add the SDK so I guess it's ok as-is.

For @mabdinur I think we do have to decide for the spring boot starters - spring boot starters are usually more self-contained and I think it'd be idiomatic to have both SDK and OTLP exporter included for batteries included.

trask:
Does adding an exporter pull in the sdk transitively?

iNikem:
As it was conscious decision from Otel to clearly separate API vs SDK, I think we should continue doing so. Which means that user has to opt-in into actually collecting monitoring data. Even if some of their libraries are instrumented. This opt-in happens either by adding SDK to _application_ dependencies or by installing java agent.

Thus I think our library instrumentations should not depend on SDK.

anuraaga:
Yeah I think we can close this since adding the exporter will generally be enough!

PR #902, opened 2020-08-06
Add @Withspan functionality using spring-aop

mabdinur:
Extend this functionality to spring projects: https://github.com/open-telemetry/opentelemetry-java-instrumentation#configure-a-withspan-annotation.

### DIFF
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
@@ -25,7 +25,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
-/** Configures {@link WithSpanAspect} to trace spring beans. */
+/** Configures {@link WithSpanAspect} to trace bean methods annotated with {@link WithSpan}. */
 @Configuration
 @EnableConfigurationProperties(TraceAspectProperties.class)
 @ConditionalOnProperty(
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
@@ -16,8 +16,8 @@
 
 package io.opentelemetry.instrumentation.spring.autoconfigure.aspects;
 
+import io.opentelemetry.extensions.auto.annotations.WithSpan;
 import io.opentelemetry.trace.Tracer;
-
 import org.aspectj.lang.annotation.Aspect;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -25,18 +25,18 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
-/** Configures {@link WithSpanAspect} */
+/** Configures {@link WithSpanAspect} to trace spring beans. */
 @Configuration
 @EnableConfigurationProperties(TraceAspectProperties.class)
 @ConditionalOnProperty(
     prefix = "opentelemetry.trace.aspects",
     name = "enabled",
     matchIfMissing = true)
-@ConditionalOnClass({Aspect.class})
+@ConditionalOnClass({Aspect.class, WithSpan.class})
 public class TraceAspectAutoConfiguration {
 
   @Bean
-  public WithSpanAspect traceClassAspect(Tracer tracer) {
+  public WithSpanAspect withSpanAspect(Tracer tracer) {
     return new WithSpanAspect(tracer);
   }
 }
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
@@ -17,6 +17,9 @@
 package io.opentelemetry.instrumentation.spring.autoconfigure.aspects;
 
 import io.opentelemetry.trace.Tracer;
+
+import org.aspectj.lang.annotation.Aspect;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
@@ -25,7 +28,11 @@ import org.springframework.context.annotation.Configuration;
 /** Configures {@link WithSpanAspect} */
 @Configuration
 @EnableConfigurationProperties(TraceAspectProperties.class)
-@ConditionalOnProperty(prefix = "opentelemetry.trace.aspects", name = "enabled", matchIfMissing = true)
+@ConditionalOnProperty(
+    prefix = "opentelemetry.trace.aspects",
+    name = "enabled",
+    matchIfMissing = true)
+@ConditionalOnClass({Aspect.class})
 public class TraceAspectAutoConfiguration {
 
   @Bean
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.spring.autoconfigure.aspects;
+
+import io.opentelemetry.trace.Tracer;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/** Configures {@link WithSpanAspect} */
+@Configuration
+@EnableConfigurationProperties(TraceAspectProperties.class)
+@ConditionalOnProperty(prefix = "opentelemetry.trace.aspects", name = "enabled", matchIfMissing = true)
+public class TraceAspectAutoConfiguration {
+
+  @Bean
+  public WithSpanAspect traceClassAspect(Tracer tracer) {
+    return new WithSpanAspect(tracer);
+  }
+}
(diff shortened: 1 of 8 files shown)


trask:
@mabdinur can you merge in master, it will fix the muzzle failure (#899)

mabdinur:
@trask I plan to add tests after I get this change approved. Is that okay?

anuraaga:
@mabdinur I think this is a great feature so let's go ahead and add tests now

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  private String getSpanName(final ProceedingJoinPoint pjp) {
+    MethodSignature signature = (MethodSignature) pjp.getSignature();
+    Method method = signature.getMethod();
+    WithSpan withSpan = method.getAnnotation(WithSpan.class);
+
+    String spanName = withSpan.value();
+    if (spanName.isEmpty()) {
+      spanName = method.getName();
```
iNikem, line 67: It should be `className.methodName`

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,76 @@
...
+  private final Tracer tracer;
+
+  public WithSpanAspect(Tracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Around("@annotation(io.opentelemetry.extensions.auto.annotations.WithSpan)")
+  public Object traceMethod(final ProceedingJoinPoint pjp) throws Throwable {
+
+    Span span = tracer.spanBuilder(getSpanName(pjp)).startSpan();
```
iNikem, line 49: `WithSpan` now supports `kind` attribute as well for setting span kind. Do you want to support this as well?

mabdinur:
I added tests for the aspect and the autoconfiguration

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectAutoConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+import io.opentelemetry.extensions.auto.annotations.WithSpan;
+import io.opentelemetry.trace.Tracer;
+import org.aspectj.lang.annotation.Aspect;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/** Configures {@link WithSpanAspect} to trace spring beans. */
```
anuraaga, line 28:
```suggestion
/** Configures {@link WithSpanAspect} to trace bean methods annotated with {@link WithSpan}. */
```

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/TraceAspectProperties.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.spring.autoconfigure.aspects;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/** Configuration for enabling aspects. */
```
anuraaga, line 21:
```suggestion
/** Configuration for enabling tracing aspects. */
```

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * <p>Ensure methods annotated with {@link WithSpan} are implemented on beans managed by the Spring
+ * container.
+ */
+@Aspect
+public class WithSpanAspect {
+
+  private final Tracer tracer;
+
+  private static final Logger logger = LoggerFactory.getLogger(WithSpanAspect.class);
```
anuraaga, line 45: Statics before non-statics

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public WithSpanAspect(Tracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Around("@annotation(io.opentelemetry.extensions.auto.annotations.WithSpan)")
+  public Object traceMethod(final ProceedingJoinPoint pjp) throws Throwable {
```
anuraaga, line 52: Don't think we need final in the arguments

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,81 @@
...
+    Span span = tracer.spanBuilder(getSpanName(pjp)).setSpanKind(Kind.INTERNAL).startSpan();
+    logger.info(getSpanName(pjp));
```
anuraaga, line 54:
```suggestion
```

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private static final Logger logger = LoggerFactory.getLogger(WithSpanAspect.class);
+
+  public WithSpanAspect(Tracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Around("@annotation(io.opentelemetry.extensions.auto.annotations.WithSpan)")
+  public Object traceMethod(final ProceedingJoinPoint pjp) throws Throwable {
+    Span span = tracer.spanBuilder(getSpanName(pjp)).setSpanKind(Kind.INTERNAL).startSpan();
```
iNikem, line 53: You should take `kind` from `WithSpan` annotation, if that is present.

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
mabdinur: I removed the logger

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
trask: ya our convention here is not clear, i opened #921 to clarify and sync up our codebase to match our desired convention

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,76 @@
...
+    if (spanName.isEmpty()) {
+      return method.getDeclaringClass().getSimpleName() + "." + method.getName();
+    }
+    return spanName;
+  }
+
+  private void errorHandler(Span span, Throwable t) {
+    span.recordException(t);
+    span.setStatus(Status.INTERNAL);
+  }
```
trask, line 75: i think these two lines will read better inlined into the catch block

instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
```diff
@@ -0,0 +1,150 @@
...
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.aop.aspectj.annotation.AspectJProxyFactory;
+
+/** Spring AOP Test for {@link WithSpanAspect} */
+@ExtendWith(MockitoExtension.class)
+public class WithSpanAspectTest {
```
trask, line 42: nice tests 👍

instrumentation-core/spring/starters/spring-starter/spring-starter.gradle
```diff
@@ -11,7 +11,9 @@ sourceCompatibility = '8'
 
 dependencies { 
   api group: "org.springframework.boot", name: "spring-boot-starter", version: versions.springboot
+  api group: 'org.springframework.boot', name: 'spring-boot-starter-aop', version: versions.springboot
   api project(':instrumentation-core:spring:spring-boot-autoconfigure')
+  api deps.opentelemetryApiAutoAnnotations
   api deps.opentelemetryApi
   api deps.opentelemetryLogging
   api deps.opentelemetrySdk
```
trask, line 19: based on discussion in #846, does this make sense?
```suggestion
  api deps.opentelemetryApi
  compileOnly deps.opentelemetryApiAutoAnnotations
  compileOnly deps.opentelemetrySdk
```

mabdinur: For the starter I'd like to provide all the opentelemetry dependencies required to instrument a spring application. Users can create their own custom starters and supply their own implementation of the opentelemetry-sdk and opentelemetryApiAutoAnnotations using opentelemetry-spring-boot-autoconfigure project. Here's that gradle file:
 - https://github.com/mabdinur/opentelemetry-java-instrumentation/blob/add_with_span_spring_aop/instrumentation-core/spring/spring-boot-autoconfigure/spring-boot-autoconfigure.gradle. 

Here's a pull request that documents this behavior: 
https://github.com/mabdinur/opentelemetry-java-instrumentation/pull/14/files

mabdinur: What would make sense is removing the opentelemetrySdk dependency since this is provided by the default logger (opentelemetryLogging). I'm not sure if this change would reduce readability. 

```suggestion
  api deps.opentelemetryApiAutoAnnotations
  api deps.opentelemetryApi
  api deps.opentelemetryLogging
```

anuraaga: Yeah that discussion wasn't about starter :) I think it should have out-of-the-box deps. Could even add otlp by default, reconfigure would be similar to the process for reconfiguring logging to use log4j instead of logback (exclude and add dependency)

instrumentation-core/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
```diff
@@ -0,0 +1,150 @@
...
+  static class WithSpanTester {
+    @WithSpan
```
anuraaga, line 44: Can you add a test with a constructor? We document that it works so should test it (not sure if it would, since usually constructors happen before becoming a bean). If it doesn't work let's update the javadoc

mabdinur: I will update the java docs. This spring aspect won't work on a constructor. It only proxies beans after initialization

instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspect.java
```diff
@@ -0,0 +1,72 @@
...
+import io.opentelemetry.trace.Status;
+import io.opentelemetry.trace.Tracer;
+import java.lang.reflect.Method;
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.aspectj.lang.reflect.MethodSignature;
+
+/**
+ * Uses Spring-AOP to wrap methods and constructors marked by {@link WithSpan} in a {@link
```
anuraaga, line 31: Ah sorry I meant this doc

instrumentation-core/spring/starters/spring-starter/spring-starter.gradle
trask: > Could even add otlp by default, reconfigure would be similar to the process for reconfiguring logging to use log4j instead of logback (exclude and add dependency)

I like this idea. I opened #923 to track/discuss.

PR #908, opened 2020-08-06
Replace Spock retry with Gradle retry.

anuraaga:
I randomly found Gradle has an official retry plugin

https://github.com/gradle/test-retry-gradle-plugin

It seems a little nicer than using spock's retry since the retries are reported, even in the build scan

https://github.com/gradle/test-retry-gradle-plugin#gradle
https://github.com/gradle/test-retry-gradle-plugin#gradle-enterprise

Aside: I was surprised to see `@Retry` already on the akka tests despite how much the fail :O

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -277,6 +277,9 @@ tasks.withType(Test).configureEach {
   enabled = isJavaVersionAllowed(JavaVersion.current()) && !project.rootProject.hasProperty("skipTests")
 
   retry {
+    // Retry is opt-in - we don't want our tests to be flaky but currently go ahead and retry them to allow CI
+    // to work. Please continue remove this property from as many projects as possible. You can see tests that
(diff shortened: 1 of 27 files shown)


anuraaga:
Ah found the ones that give us grief weren't actually in akka-http, added retry to them too

gradle/java.gradle
```diff
@@ -274,4 +275,8 @@ tasks.withType(Test).configureEach {
   // Disable all tests if current JVM doesn't match version requirements
   // Disable all tests if skipTests property was specified
   enabled = isJavaVersionAllowed(JavaVersion.current()) && !project.rootProject.hasProperty("skipTests")
+
+  retry {
```
iNikem, line 279: Please add a big warning "we still need to fix flaky tests not just retry them forever!"

trask:
nice 👍

PR #910, opened 2020-08-06
Add document describing current problem with context propagation

iNikem:


docs/contributing/inter-thread-context-propagation.md
```diff
@@ -0,0 +1,87 @@
...
+
+This means that mechanism described in the previous section will propagate the execution context
+of one request processing to a thread accepting some next, unrelated, request.
+This will result in spans representing the accepting and processing of the second request will join
+the same trace as those of the first span. This mistakenly correlates unrelated requests and may lead
+to huge traces being active for hours and hours.
+
+In addition this makes some of our tests extremely flaky.
+
+## The currently accepted trade-offs
```
iNikem, line 83: Add the note about cleaning context on SERVER span creation.

docs/contributing/inter-thread-context-propagation.md
```diff
@@ -0,0 +1,87 @@
...
+}
+```
+
+In both cases request processing requires some potentially long operation and application developer
+wants to do them off the main thread. In the first case this hand-off between request accepting thread
+and request processing thread happens manually, by submitting work into some thread pool.
+In the second case it is the framework that handles separate thread pool and passing work to it.
+
+In cases like this proper tracing solution should still combine into a single trace all the work
+required for request processing, irregardless in what thread that work happened. With proper
```
trask, line 40:
```suggestion
required for request processing, regardless in what thread that work happened. With proper
```

docs/contributing/inter-thread-context-propagation.md
```diff
@@ -0,0 +1,87 @@
...
+
+## The drawback
+There are some runtime environments which, simplifying, do the following:
+```
+pool.submit(new AcceptRequestRunnable() {
+    Request req = readRequest()
+    pool.submit(new ProcessRequestRunnable(req) {
+        writeResponse(process(req))
+        pool.submit(new AcceptRequestRunnable() {
+        ... repeat untill shutdown
```
trask, line 70:
```suggestion
        ... repeat until shutdown
```

docs/contributing/inter-thread-context-propagation.md
```diff
@@ -0,0 +1,87 @@
...
+        try(Scope scope = withScopedContext(this.context)) {
+            return userRepository.queryShippingAddress(requet)
+        }}
+    job.context = Context.current()
+    Future f1 = pool.submit()
+
+```
+
+## The drawback
+There are some runtime environments which, simplifying, do the following:
```
trask, line 63: the example in this section shows how if we leak to the first `pool.submit`, then it will keep leaking, but i don't think it answers the question of how we leak to the first `pool.submit`.

maybe the more concrete example you tracked down with tomcat / async dispatch would be more illustrative of the problem?

iNikem: I don't quite understand what do you mean by "leak the first pool.submit"? My idea was to show that any context, which was active during the first `pool.submit` will continue to be active for all runnables submitted via this started chain. That second `AcceptRequestRunnable` will have the same active context as the first one.

trask: I'm not clear, are you showing that the context (e.g. `spanId`) from `REQUEST#1` below will leak to `REQUEST#2`? I don't think it will, at least I don't think it will via the `java-concurrent` instrumentation alone.

This is what it looks like to me:

```
pool.submit(new AcceptRequestRunnable() {    <!--- assuming clean context to start with
    Request req = readRequest()    <--- REQUEST#1
    pool.submit(new ProcessRequestRunnable(req) {    <!--- the clean context from above will be propagated
        writeResponse(process(req))    <--- REQUEST#1
        pool.submit(new AcceptRequestRunnable() {    <!--- the clean context from above will be propagated
            Request req = readRequest()    <--- REQUEST#2
            ... repeat until shutdown
    })
})
```

iNikem: I tried to add clarifications, PTAL

docs/contributing/inter-thread-context-propagation.md
```diff
@@ -0,0 +1,94 @@
...
+```
+pool.submit(new AcceptRequestRunnable() {
+    Request req = readRequest()
+    // At this point Context.current() will have a started span
```
trask, line 67: > At this point Context.current() will have a started span

this is the part of the example that I'm not following, how does the (thread-bound) `Context` with the started span escape `readRequest()`?

iNikem: And now?

docs/contributing/inter-thread-context-propagation.md
```diff
@@ -64,15 +64,17 @@ There are some runtime environments which, simplifying, do the following:
...
-        pool.submit(new AcceptRequestRunnable() {
-        // The same context from above is propagated here as well
-        ... repeat until shutdown
-    })
+    service(req){ <- This method is instrumented and we start new scope here
+        // At this point Context.current() will have a started span
+        // recording monitoring data about `req`
+        pool.submit(new ProcessRequestRunnable(req) {
+            // The same context from above is active here
+            writeResponse(process(req))
```
trask, line 72: why would we instrument `service(req)` above, and not instrument `writeResponse(process(req))`, which would solve the propagation issue for this example? let's add this to agenda for tomorrow's special topic meeting, will be easier to discuss, i have a feeling i'm not expressing my thoughts about this (complicated topic) very well

PR #911, opened 2020-08-06
Extract Akka context propagation into separate module

iNikem:


settings.gradle
```diff
@@ -40,6 +40,8 @@ include ':smoke-tests:wildfly'
 include ':smoke-tests:springboot'
 
 // instrumentation:
+include ':instrumentation:akka-context-propagation'
+include ':instrumentation:akka-context-propagation:akka-testing'
```
anuraaga, line 44: Is there any benefit to having this separate testing module?

iNikem: This is one strange module. I think I will merge it indeed...

PR #914, opened 2020-08-06
Add spring web documentation

mabdinur:
Documentation for spring-web manual instrumentation

### DIFF
--- a/instrumentation-core/spring/spring-web-3.1/README.md
+++ b/instrumentation-core/spring/spring-web-3.1/README.md
@@ -22,18 +22,19 @@ For Maven add to your `pom.xml`:
     <version>OPENTELEMETRY_VERSION</version>
   </dependency>
 
+   <!-- provides opentelemetry-sdk -->
    <dependency>
     <groupId>io.opentelemetry</groupId>
-    <artifactId>opentelemetry-sdk</artifactId>
+    <artifactId>opentelemetry-exporters-logging</artifactId>
     <version>OPENTELEMETRY_VERSION</version>
   </dependency>
 
-  <!-- spring web -->
+  <!-- required to instrument spring-web -->
+  <!-- this artifact should already be present in your application -->
   <dependency>
     <groupId>org.springframework</groupId>
     <artifactId>spring-web</artifactId>
     <version>SPRING_VERSION</version>
-    <scope>runtime</scope>
   </dependency>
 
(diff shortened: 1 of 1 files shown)


instrumentation-core/spring/spring-web-3.1/README.md
```diff
@@ -0,0 +1,86 @@
...
+    <groupId>io.opentelemetry</groupId>
+    <artifactId>opentelemetry-sdk</artifactId>
+    <version>OPENTELEMETRY_VERSION</version>
+  </dependency>
+
+  <!-- spring web -->
+  <dependency>
+    <groupId>org.springframework</groupId>
+    <artifactId>spring-web</artifactId>
+    <version>SPRING_VERSION</version>
```
anuraaga, line 35: Since this is part of the user's app maybe we don't need to document it. But if we do, isn't it compile scope?

anuraaga:
Comment applies to the three PRs - other than that looks good though thanks for the docs!

instrumentation-core/spring/spring-web-3.1/README.md
mabdinur: Setting the spring-web dependency to runtime works since the instrument-core:spring:spring-web project already has spring compile only dependencies. But you're right this can be confusing. I'll remove the compile scope.

I think it's important to document the spring-web dependency in this readme. I don't see a case where a user would add this artifact to their project and not already have spring-web in their build/pom file. However if this dependency is missing this project will throw a ClassDefNotFoundException. I'll add a note to clear this up.

instrumentation-core/spring/spring-web-3.1/README.md
```diff
@@ -0,0 +1,89 @@
...
+</dependencies>
+```
+
+For Gradle add to your dependencies:
+```groovy
+implementation 'io.opentelemetry.instrumentation:opentelemetry-spring-web-3.1:OPENTELEMETRY_VERSION'
+implementation 'io.opentelemetry:opentelemetry-exporters-logging:OPENTELEMETRY_VERSION'
+
+//this artifact should already be present in your application
+runtime 'org.springframework:spring-web:SPRING_VERSION'
```
trask, line 49: i think to match the change you made for maven
```suggestion
implementation 'org.springframework:spring-web:SPRING_VERSION'
```

PR #915, opened 2020-08-06
Add spring webmvc documentation

mabdinur:
1. Add readme to document instrumentation-core:spring-webmvc-3.1 
2. Rename all references WebMVCFilter to WebMvcFilter

### DIFF
--- a/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/webmvc/WebMvcFilterAutoConfiguration.java
+++ b/instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/webmvc/WebMvcFilterAutoConfiguration.java
@@ -16,7 +16,7 @@
 
 package io.opentelemetry.instrumentation.spring.autoconfigure.webmvc;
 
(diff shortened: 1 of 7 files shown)


instrumentation-core/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/webmvc/WebMvcFilterAutoConfiguration.java
```diff
@@ -30,10 +30,10 @@
 @EnableConfigurationProperties(WebMVCProperties.class)
```
anuraaga, line 30: Looks line we need to update other classes with the spelling change too

instrumentation-core/spring/spring-webmvc-3.1/README.md
```diff
@@ -0,0 +1,82 @@
+# Manual Instrumentation for Spring-WebMvc
```
anuraaga, line 1: Spring Web MVC

We should copy from upstream docs / config for names as much as possible.

https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/spring-webmvc.gradle#L1

anuraaga:
Thanks!

PR #916, opened 2020-08-06
Add spring webflux documentation

mabdinur:
1. Add README to document instrumentation-core:spring-webflux-5.0 
2.  Rename springwebflux package to spring.webflux


instrumentation/spring-webflux-5.0/src/main/java/io/opentelemetry/auto/instrumentation/spring/webflux/server/HandlerAdapterAdvice.java
```diff
@@ -14,9 +14,9 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.auto.instrumentation.springwebflux.server;
+package io.opentelemetry.auto.instrumentation.spring.webflux.server;
```
trask, line 17: our convention has been for the package name to be the module name minus the dashes and minus the version

i'm good with having a special case for spring (making this change), since we have many different spring modules (and growing), and this does look nicer

can you submit a follow-up PR to change this in the other spring auto modules so that we adopt this new spring module naming consistently?

instrumentation-core/spring/spring-webflux-5.0/README.md
```diff
@@ -0,0 +1,87 @@
...
+  <dependency>
+    <groupId>io.opentelemetry.instrumentation</groupId>
+    <artifactId>opentelemetry-spring-webflux-5.0</artifactId>
+    <version>OPENTELEMETRY_VERSION</version>
+  </dependency>
+
+   <!-- provides opentelemetry-sdk -->
+   <dependency>
+    <groupId>io.opentelemetry</groupId>
+    <artifactId>opentelemetry-exporters-logging</artifactId>
```
trask, line 28: maybe add a comment about alternatively using otlp, jaeger or zipkin exporter here (and in the other PRs), since that's what most people will be doing

instrumentation-core/spring/spring-webflux-5.0/README.md
```diff
@@ -0,0 +1,87 @@
...
+Replace `SPRING_VERSION` with the version of spring you're using.
+`Minimum version: 5.0`
+
+Replace `OPENTELEMETRY_VERSION` with the latest stable [release](https://mvnrepository.com/artifact/io.opentelemetry).
+`Minimum version: 0.8.0`
+
+For Maven add to your `pom.xml`:
+```xml
+<dependencies>
+  <!-- opentelemetry -->
```
trask, line 18:
```suggestion
  <!-- opentelemetry instrumentation -->
```

instrumentation-core/spring/spring-webflux-5.0/README.md
```diff
@@ -0,0 +1,87 @@
...
+  <dependency>
+    <groupId>io.opentelemetry.instrumentation</groupId>
+    <artifactId>opentelemetry-spring-webflux-5.0</artifactId>
+    <version>OPENTELEMETRY_VERSION</version>
+  </dependency>
+
+   <!-- provides opentelemetry-sdk -->
```
trask, line 25: i'm not sure we need to include note about providing the opentelemetry-sdk, i like the message of "use the OpenTelemetry API and/or use OpenTelemetry library instrumentation, then add an exporter, and you are good to go".

i think that's the message we agreed on in #846, but it would be good to get @anuraaga's thoughts here also

users don't need to care about the sdk, they just use API and add an exporter, and 
```suggestion
   <!-- opentelemetry exporter -->
```

instrumentation-core/spring/spring-webflux-5.0/README.md
```diff
@@ -0,0 +1,87 @@
...
+  <dependency>
+    <groupId>org.springframework</groupId>
+    <artifactId>spring-webflux</artifactId>
+    <version>SPRING_VERSION</version>
+  </dependency>
+
+</dependencies>
+```
+
+For Gradle add to your dependencies:
```
trask, line 43: i think copy same comments from pom above into gradle below, since gradle users will probably just skip the above section

trask:
hey @mabdinur! sorry I caused merge conflicts for you with #884 

merging upstream/master into your branch looks like there's only 1 merge conflict (even though github UI shows 7?), which is in the import section due to package renames

instrumentation/spring-webflux-5.0/src/main/java/io/opentelemetry/auto/instrumentation/spring/webflux/server/HandlerAdapterAdvice.java
mabdinur: I'll make this my next change

PR #917, opened 2020-08-07
Remove annotation instrumentation vertx dependency

trask:
Just some cleanup, thought I had removed annotation dependency from all the tests, but missed this


instrumentation/vertx-reactive-3.5/src/test/groovy/VertxReactiveWebServer.java
```diff
@@ -98,26 +103,34 @@ public void start(final io.vertx.core.Future<Void> startFuture) {
...
   }
 
-  @WithSpan
   private void handleListProducts(final RoutingContext routingContext) {
-    HttpServerResponse response = routingContext.response();
-    Single<JsonArray> jsonArraySingle = listProducts();
-
-    jsonArraySingle.subscribe(
-        arr -> response.putHeader("content-type", "application/json").end(arr.encode()));
+    Span span = tracer.spanBuilder("handleListProducts").startSpan();
```
anuraaga, line 107: Can we put this change in another PR? Is it related to annotations?

anuraaga: Oops missed that

trask:
oh, this is going to fix the current build failure in #929

Issue #903, opened 2020-08-06
Prevent duplicate telemetry when using both library and auto instrumentation


trask:
I think this is an important, e.g. you may start out using library instrumentation, and then decide to throw on the javaagent later (or your ops team decides to throw on the javaagent), and I think part of the goal of maintaining both library and auto instrumentation in the same repo is so we can make them play nicely together.

(this includes preventing duplicate `@WithSpan` telemetry when using both spring-aop instrumentation module and javaagent)

anuraaga:
For reference, I guess this is along the lines of the gRPC logic to prevent the interceptor from being added if it already exists

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/grpc-1.5/src/main/java/io/opentelemetry/auto/instrumentation/grpc/server/GrpcServerBuilderInstrumentation.java#L71

which interestingly seems extremely unlikely until we extract the library instrumentation :D

anuraaga:
Also I guess we can have a pattern of including a duplicate test in the `auto` project which configures the same way the `library` test does, and it should still pass (currently Armeria wouldn't I think, heh)

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/armeria-1.0/auto/src/test/groovy/io/opentelemetry/auto/instrumentation/armeria/v1_0/ArmeriaServerTest.groovy#L28

trask:
If possible, I think ideally the auto-instrumentation would take precedence over the library instrumentation (where they are both official packages from this repo). The reason for this is that it's generally easier to upgrade to the latest javaagent (e.g. done by ops to fix a bug or get a new feature) than to upgrade to the latest library instrumentation.

trask:
[adding more thoughts here]

see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/920#discussion_r467361890:

> another option for suppressing library instrumentation that might be easier to generalize, is to make the coordination more explicit, e.g. have a method in instrumentation-api like isAutoEnabled(instrumentationName), and the library instrumentation can explicitly check that to see if there is auto instrumentation with the same name enabled

anuraaga:
That's a nice idea. I'm a bit on the fence about introducing logic in library instrumentation to support auto - it's harmless but does seem like a break in separations of concerns.

[...]

Issue #892, opened 2020-08-04
Different approach for shading library instrumentation classes inside of auto instrumentation


trask:
Currently we are creating a shaded classifier for the library instrumentation, which is then used by auto instrumentation.

The issue I just ran across is that this makes it confusing when making changes across both library and auto instrumentation, because the changes you make in library instrumentation are not reflected in the auto instrumentation project until you rebuild the shaded classifier for the library instrumentation.

I'd like to propose that we change this and instead use regular dependency from auto instrumentation to library instrumentation, and shade the library instrumentation in the `opentelemetry-javaagent` build (which is where we shade some other stuff anyways).

PR #920, opened 2020-08-07
Add instrumentation for Armeria WebClient.

anuraaga:
Also updates the server to suppress user-provided decorator when auto instrumenting.

Went ahead and followed the pattern of `GrpcTest` by just testing server / client in the same test since it was simple. May extract as I add more sophisticated tests in a future PR (Armeria custom context storage class is coming!)


instrumentation/armeria-1.0/auto/src/main/java/io/opentelemetry/auto/instrumentation/armeria/v1_0/ArmeriaServerBuilderInstrumentation.java
```diff
@@ -0,0 +1,84 @@
...
+    public static boolean suppressDecorator(@Advice.Argument(0) Function<?, ?> decorator) {
+      return decorator
+          .getClass()
+          .getName()
+          .equals(
+              "io.opentelemetry.instrumentation.armeria.v1_0.server.OpenTelemetryService$Decorator");
+    }
+
+    @Advice.OnMethodExit
+    public static void handleSuppression(
```
anuraaga, line 75: This craziness is the only pattern I could find for suppressing a non-void method

instrumentation/armeria-1.0/testing/src/main/groovy/io/opentelemetry/instrumentation/armeria/v1_0/AbstractArmeriaTest.groovy
```diff
@@ -17,23 +17,25 @@
...
 import spock.lang.Shared
 
+import static io.opentelemetry.trace.Span.Kind.CLIENT
 import static io.opentelemetry.trace.Span.Kind.SERVER
 
-abstract class AbstractArmeriaServerTest extends InstrumentationSpecification {
+abstract class AbstractArmeriaTest extends InstrumentationSpecification {
 
-  abstract void configureServer(ServerBuilder sb)
+  abstract ServerBuilder configureServer(ServerBuilder serverBuilder)
```
anuraaga, line 36: Returning `ServerBuilder` here allows checking my suppression didn't return `null`

instrumentation/armeria-1.0/auto/src/main/java/io/opentelemetry/auto/instrumentation/armeria/v1_0/ArmeriaServerBuilderInstrumentation.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public static class SuppressDecoratorAdvice {
+    @Advice.OnMethodEnter(skipOn = Advice.OnNonDefaultValue.class)
+    public static boolean suppressDecorator(@Advice.Argument(0) Function<?, ?> decorator) {
+      return decorator
+          .getClass()
+          .getName()
+          .equals(
+              "io.opentelemetry.instrumentation.armeria.v1_0.server.OpenTelemetryService$Decorator");
```
trask, line 71: i think #892 will break this because it will shade this constant also. there are ways to prevent the string from getting shaded (e.g. make it a constant NAME = "_io.opentelemetry...".substring(1)".

another option for suppressing library instrumentation that might be easier to generalize, is to make the coordination more explicit, e.g. have a method in `instrumentation-api` like `isAutoEnabled(instrumentationName)`, and the library instrumentation can explicitly check that to see if there is auto instrumentation with the same name enabled

in this case i guess that would mean returning a no-op decorator, or changing the API to expose a method that registers a decorator on the `ServerBuilder` so we can no-op the registration in this case

I'll add this note to #903, no need to decide/address in this PR

instrumentation/armeria-1.0/auto/armeria-1.0-auto.gradle
```diff
@@ -20,5 +20,6 @@ dependencies {
 
   compileOnly group: 'com.linecorp.armeria', name: 'armeria', version: '0.99.8'
 
+  testImplementation project(path: ':instrumentation:armeria-1.0:library')
```
iNikem, line 23: `path:` is not needed here

instrumentation/armeria-1.0/auto/src/main/java/io/opentelemetry/auto/instrumentation/armeria/v1_0/AbstractArmeriaInstrumentation.java
```diff
@@ -29,7 +29,12 @@ public AbstractArmeriaInstrumentation() {
   @Override
   public String[] helperClassNames() {
     return new String[] {
-      packageName + ".shaded.internal.ContextUtil",
+      packageName + ".shaded.client.ArmeriaClientTracer",
+      packageName + ".shaded.client.ArmeriaClientTracer$ArmeriaSetter",
+      packageName + ".shaded.client.OpenTelemetryClient",
+      packageName + ".shaded.client.OpenTelemetryClient$Decorator",
+      // .thenAccept(log -> lambda
```
iNikem, line 36: Not sure if this comment is helping or confusing :)

instrumentation/armeria-1.0/auto/src/main/java/io/opentelemetry/auto/instrumentation/armeria/v1_0/ArmeriaServerBuilderInstrumentation.java
iNikem: Please add some documentation about what you are trying to achieve here

instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/client/ArmeriaClientTracer.java
```diff
@@ -0,0 +1,78 @@
...
+  protected Setter<ClientRequestContext> getSetter() {
+    return ArmeriaSetter.INSTANCE;
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.armeria-1.0";
+  }
+
+  private enum ArmeriaSetter implements Setter<ClientRequestContext> {
```
iNikem, line 70: Although this is a nice way to ensure only one instance of the class exists in JVM, it differs from all other instrumentations. Unless there is some strong reason for such differences, I prefer to have uniform codebase.

instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/client/OpenTelemetryClient.java
```diff
@@ -0,0 +1,103 @@
...
+                if (log.responseCause() != null) {
+                  clientTracer.endExceptionally(
+                      span, log, log.responseCause(), requestEndTimeNanos);
+                } else {
+                  clientTracer.end(span, log, requestEndTimeNanos);
+                }
+              });
+    }
+
+    try (Scope ignored = clientTracer.startScope(span, ctx)) {
```
iNikem, line 85: Just to be sure: you start new scope even if span is not recording?

instrumentation/armeria-1.0/auto/src/main/java/io/opentelemetry/auto/instrumentation/armeria/v1_0/AbstractArmeriaInstrumentation.java
anuraaga: Tried a bit harder

instrumentation/armeria-1.0/auto/src/main/java/io/opentelemetry/auto/instrumentation/armeria/v1_0/ArmeriaServerBuilderInstrumentation.java
anuraaga: I ended up needing yet a different pattern because of the shading change making the classnames the same. I guess it's actually a bit more robust now and we'll revisit in #903 for an inversion of control.

instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/client/OpenTelemetryClient.java
anuraaga: Yeah - I think it's required for logs injection

trask: ah, interesting, hadn't thought of that 👍

Issue #926, opened 2020-08-08
New gradle retry doesn't seem to be working for AkkaHttpServerInstrumentationTestAsync


trask:
If my understanding is correct, this failure should have been retried:
https://15978-210933087-gh.circle-artifacts.com/0/reports/instrumentation/akka-http-10.0/tests/test/classes/AkkaHttpServerInstrumentationTestAsync.html

(this build was after merging in latest master so has the new gradle retry config)

trask:
`AkkaExecutorInstrumentationTest` is also not retrying.

To repro locally, add

```
ext {
  retryTests = true
}
```

to `akka-context-propagation.gradle`

and change something in `AkkaExecutorInstrumentationTest` to cause it to fail

and run:

```
./gradlew :instrumentation:akka-context-propagation:test --info --stacktrace
```

and it should fail with:

```
Execution failed for task ':instrumentation:akka-context-propagation:test'.
> org.gradle.test-retry was unable to retry the following test methods, which is unexpected. Please file a bug report at https://github.com/gradle/test-retry-gradle-plugin/issues
     AkkaExecutorInstrumentationTest#akka.dispatch.forkjoin.ForkJoinPool@139d3ed5[Running, parallelism = 12, size = 0, active = 0, running = 0, steals = 0, tasks = 0, submissions = 0] 'invoke ForkJoinTask' propagates
     AkkaExecutorInstrumentationTest#akka.dispatch.forkjoin.ForkJoinPool@3bed3ad7[Running, parallelism = 12, size = 0, active = 0, running = 0, steals = 0, tasks = 0, submissions = 0] 'execute ForkJoinTask' propagates
     AkkaExecutorInstrumentationTest#akka.dispatch.forkjoin.ForkJoinPool@34ef8759[Running, parallelism = 12, size = 0, active = 0, running = 0, steals = 0, tasks = 0, submissions = 0] 'submit Runnable' propagates
     AkkaExecutorInstrumentationTest#akka.dispatch.forkjoin.ForkJoinPool@221fe5a2[Running, parallelism = 12, size = 0, active = 0, running = 0, steals = 0, tasks = 0, submissions = 0] 'submit ForkJoinTask' propagates
[...]

PR #927, opened 2020-08-08
Add retry to akka tests

trask:
Not sure why, but this seems to be failing a lot since #911


trask:
Closing until we resolve #926

Issue #298, opened 2020-03-31
Smoke test improvements


trask:
Improve smoke test harness
* Test against different application servers and different JVMs

Add more smoke tests, starting with the most fragile instrumentation we have:
* ~Log4j 1 instrumentation~ (EDIT: removed until after GA)
* ~Log4j 2 instrumentation~ (EDIT: removed until after GA)
* ~java.util.logging instrumentation~ (EDIT: removed until after GA)
* ~Logback instrumentation~ (EDIT: removed until after GA)
  * ~Need to be careful not to double capture when one logging framework delegates to another~
  * ~JBoss/Wildfly has it's own unique java.util.logging implementation~
* gRPC instrumentation
  * gRPC instrumentation is fragile because of the shading conflicts involved with gRPC Context used internally by embedded OpenTelemetry API
* OpenTelemetry API instrumentation
  * OpenTelemetry API instrumentation is fragile because of the shading involved to bridge the user brought OpenTelemetry API and the embedded OpenTelemetry API

trask:
More things to smoke test:

* Exporter configuration (OTLP, Jaeger, Zipkin, Prometheus)
* B3 Propagator configuration

trask:
Another thing to improve:

* Don't use logging exporter / log parsing, which is unreliably under concurrent logging

iNikem:
Based [GA requirements](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/ga-requirements.md) I will narrow down the scope of this task.

Smoke test consists of running given application (Spring Boot, Wildlfy) with Otel java auto instrumentation agent attached.  Agent should be configured to run given exporter (OTLP, Jaeger, Zipkin) and to send data to locally run Otel Collector (also started by the test harness). Collector should be configured to use [file exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/master/exporter/fileexporter/README.md). Test harness should execute requests against the application and verify that all expected data was written by Collector to the export file.

Test harness should be run on all LTS versions of starting from 8 plus on the latest java version if not LTS.

iNikem:
I have a working prototype, which runs Spring Boot application with java agent attached and the Collector and then makes a request to the app. Collector is configured to use file exporter. Test then reads that file written by the Collector and verifies that it contains expected spans.

My prototype differs from current tests and I want to discuss it before making a PR. First, it uses testcontainers to run both the application and the Collector within Docker containers. 

[...]

PR #929, opened 2020-08-08
Fix WithSpan annotation instrumentation

### LINKED ISSUE #913: @WithSpan annotation not functioning with 0.7.0
A little bird told me that the shading might need to be updated for the new packaging of the extensions code.

trask:
Resolves #913


opentelemetry-auto-annotations-beta-shaded-for-instrumenting/opentelemetry-auto-annotations-beta-shaded-for-instrumenting.gradle
```diff
@@ -0,0 +1,17 @@
...
+}
+
+apply from: "$rootDir/gradle/java.gradle"
+
+dependencies {
+  implementation deps.opentelemetryAutoAnnotations
+}
+
+// OpenTelemetry Auto Annotations shaded so that it can be used in opentelemetry-api-beta
+// instrumentation, and then its usage can be unshaded after OpenTelemetry API is shaded
```
anuraaga, line 12: Just to confirm this means our unit tests can catch the issue in the future?

smoke-tests/springboot/src/main/java/io/opentelemetry/smoketest/springboot/controller/WebController.java
```diff
@@ -16,13 +16,19 @@
...
 
+import io.opentelemetry.extensions.auto.annotations.WithSpan;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
 public class WebController {
   @RequestMapping("/greeting")
   public String greeting() {
-    return "Sup Dawg";
```
anuraaga, line 26: Good change

opentelemetry-auto-annotations-beta-shaded-for-instrumenting/opentelemetry-auto-annotations-beta-shaded-for-instrumenting.gradle
trask: unfortunately, no, this was needed for the fix. currently at least we have to rely on smoke tests to catch shading-related problems. probably we should create a whole smoke test around instrumentation OpenTelemetry API, since that instrumentation is shading-fragile. i will open issue........ (a couple min later) oh ya, i have an issue already for that: #298 😄

iNikem:
I will continue spreading confusion :)

First, I don't understand what exactly was broken? Our instrumentation depended and referenced `WithSpan` class directly, meaning that it referenced the correct class in the correct package. It worked before without shading, why it stopped working?

trask:
> It worked before without shading, why it stopped working?

It stopped working as soon as `WithSpan.kind()` was added.

The reason it stopped working is because the type of `kind` is `io.opentelemetry.trace.Span.Kind`, and so during shading, our call to 

```
io/opentelemetry/extensions/auto/annotations/WithSpan.kind()Lio/opentelemetry/trace/Span$Kind;
```

was shaded to call 

```
io/opentelemetry/extensions/auto/annotations/WithSpan.kind()Lio/opentelemetry/auto/shaded/io/opentelemetry/trace/Span$Kind;
```

and then `muzzle` kicked in at runtime and said that method does exist, and so it disabled the `annotations` instrumentation

PR #931, opened 2020-08-09
Avoid ClassCastException on direct SDK access

### LINKED ISSUE #336: ClassCastException when using opentelemetry SDK with auto agent
I am experimenting the OT auto agent with OT SDK. My use case is rather simple:

- A simple java job that triggers some HTTP client calls to some services
- Since my java job itself would not be instrumented by the auto-agent, I added dependency to opentelemetry-sdk to start the scope and span
- Since the http client calls are made within the scope, they should get picked up by the auto agent

However, in the beginning of the code, when I tried to obtain the tracer from SDK
```
public static void main(String[] args) throws InterruptedException {
        // Get the tracer
        TracerSdkProvider tracerProvider = OpenTelemetrySdk.getTracerProvider();
...
```

Exception is triggered : 
```
Exception in thread "main" java.lang.ClassCastException: class io.opentelemetry.auto.instrumentation.opentelemetryapi.trace.UnshadedTracerProvider cannot be cast to class io.opentelemetry.sdk.trace.TracerSdkProvider (io.opentelemetry.auto.instrumentation.opentelemetryapi.trace.UnshadedTracerProvider and io.opentelemetry.sdk.trace.TracerSdkProvider are in unnamed module of loader 'app')
```

I am wondering if my usage is legit as in #10 ? Or have I misused the auto agent/sdk? 

Many thanks in advance!

trask:
Closes #336


instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/trace/UnshadedTracerProvider.java
```diff
@@ -42,4 +53,20 @@ public Tracer get(final String instrumentationName, final String instrumentation
...
         contextStore);
   }
+
+  // this is called by OpenTelemetrySdk, which expects to get back a real TracerProviderSdk
+  @Override
+  public Object unobfuscate() {
+    if (!messageAlreadyLogged.getAndSet(true)) {
+      String message =
+          "direct usage of the OpenTelemetry SDK is not supported when running agent"
+              + " (run with debug logging to see stack trace)";
```
anuraaga, line 63: I'd include the flag here if mentioning debug logging

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/trace/UnshadedTracerProvider.java
```diff
@@ -42,4 +53,20 @@ public Tracer get(final String instrumentationName, final String instrumentation
             .get(instrumentationName, instrumentationVersion),
         contextStore);
   }
+
+  // this is called by OpenTelemetrySdk, which expects to get back a real TracerProviderSdk
+  @Override
+  public Object unobfuscate() {
+    if (!messageAlreadyLogged.getAndSet(true)) {
+      String message =
+          "direct usage of the OpenTelemetry SDK is not supported when running agent"
```
anuraaga, line 62: Adding the concrete operation can provide good context

Direct usage of the OpenTelemetry SDK by calling OpenTelemetrySdk.getTracrProvider() is not...

anuraaga: Also maybe a link to how to configure the agent (presumably what they were trying to do).

iNikem:
These shaded and unshaded terms are very confusing to me and don't seem natural. Can we replace them with some words that reflect the problem domain, not the solution space? 

I mean we have one set of classes that we bring as dependency, right? And another set which application can bring, right? So maybe agent/application or internal/external or something?

anuraaga:
Yeah I think we can address it outside this PR but I agree with @iNikem. Actually, it seems to be the worst of all worlds right now, where unprefixed is shaded, and prefixed is unshaded which is highly unusual. So use-case specific package names like @iNikem suggests would be nice, if there's a technical problem with that, I think swapping so unprefixed is unshaded would be easier to make presentations about :P

trask:
> Can we replace them with some words that reflect the problem domain, not the solution space?

Great idea, created #935

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/trace/UnshadedTracerProvider.java
```diff
@@ -42,4 +53,20 @@ public Tracer get(final String instrumentationName, final String instrumentation
...
+              + " (run with debug logging to see stack trace)";
+      if (log.isDebugEnabled()) {
+        log.debug(message, new Exception("stack trace"));
+      } else {
+        log.info(message);
+      }
+    }
+    return ((Obfuscated<?>) originalTracerProvider).unobfuscate();
```
iNikem, line 70: Sorry, but I am totally confused. Javadoc for `Obfuscated` says: "because auto-instrumentation takes over the static global providers returned by the API and points them to it's embedded SDK". And also "Static global providers are obfuscated when they are returned from the API to prevent users from casting them to their SDK specific implementation."

But it seems to me that here we are doing exactly the same, no? We don't "take over the static global providers returned by the API and point them to it's embedded SDK", but we take global provider, unobfuskcate it and return to the end users.

Can we have a chat where @trask we explain to me (and @anuraaga if he wishes so) this whole business with obfuscation, shading, future support for multiple versions of API/SDK etc etc. Because I am totally confused.

iNikem:
A! I was too late! :(

trask:
Oh my bad!! I meant to address @anuraaga's comments. Will submit another PR and we can continue discussing there.

iNikem:
Can we discuss this during special meeting?

trask:
👍 added special topic meeting to public calendar for Tue Aug 11, 10pm Pacific Time

https://calendar.google.com/event?action=TEMPLATE&tmeid=MnFjajExYW5xdmtsdTRpM2dzZWNqajM0YWEgZ29vZ2xlLmNvbV9iNzllM2U5MGo3YmJzYTJuMnA1YW41bGY2MEBn&tmsrc=google.com_b79e3e90j7bbsa2n2p5an5lf60%40group.calendar.google.com

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/trace/UnshadedTracerProvider.java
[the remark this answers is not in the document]
trask: Not supported means that if you use `OpenTelemetrySdk.getTraceProvider()`, you won't get back the agent's `TraceProvider` bridge, and so any telemetry you emit won't get intercepted / correlated / exported by the agent the way we do if you use `OpenTelemetry.getTraceProvider()` from OpenTelemetry API

PR #934, opened 2020-08-09
Improved smoke tests

iNikem:
New smoke tests based on Testcontainers and OpenTelemetry Collector.

Target applications (Spring Boot, Wildfly and Play) are run inside Testcontainers with attached java agent. Agent is configured to send data to the Collector which uses `fileexporter` to write received spans to a json file. Tests then read that file and can assert anything they like about spans.

Partly addresses #298. What's left to be done there: write GitHub Action to build and publish target application distributions on their changes; support different java versions for target applications.


iNikem:
Build fails with "Previous attempts to find a Docker environment failed. Will not retry. Please see logs and check configuration". Hm... can we run Testcontainers in CircleCI at all??

trask:
> Hm... can we run Testcontainers in CircleCI at all??

We are still using `datadog/dd-trace-java-docker-build:latest` to run our builds, so maybe that image doesn't have docker installed?

iNikem:
But we do want to use testcontainers? Not only for smoke tests, right? Should we hurry up our migration to Github Actions?

trask:
> But we do want to use testcontainers? Not only for smoke tests, right? Should we hurry up our migration to Github Actions?

:thumbsup: from me

I'm feeling more confident about the move to Github Actions given https://github.com/open-telemetry/community/issues/429#issuecomment-669592921

iNikem:
But I don't think we can do it until we have working Gradle cache and that requires open-telemetry/community#392

trask:
> But I don't think we can do it until we have working Gradle cache and that requires [open-telemetry/community#392](https://github.com/open-telemetry/community/issues/392)

why not? just too slow without the cache?

iNikem:
Yes. Current nightly takes ~35 minutes. I think this is too slow for PR build

trask:
Can we use local build cache for CI (at least for now) and save/restore that cache as part of the build?

iNikem:
We can try. I will get back to this in couple of days.

iNikem:
This PR also adds workflows for PR checks and for building applications for smoke tests (for now only for java 11). I also disabled smoke tests in CircleCI as the latter cannot run testcontainers. Github Actions take a lot of time for now, but Gradle cache is in place, so I will monitor how well will it work.

All in all, please review, let merge this and see what will happen.

iNikem:
GH Actions fail because there is no such images in this repo's packages yet. Will be solved shortly after merged.

I totally failed to split this into working independent pieces :(

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,37 @@
...
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          job-id: jdk${{ matrix.java }}
+
+      - name: Test with Gradle
+        uses: nick-invision/retry@v1
+        with:
+          command: ./gradlew testJava${{ matrix.java }} --stacktrace
+          timeout_minutes: 60
+          max_attempts: 3
```
trask, line 37: just checking that we want to do this retry on PRs (not just nightly)?

.github/workflows/build-play-smoke-dist.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Cache gradle dependencies
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          job-id: jdk11
+
+      - name: Login to GitHub Package Registry
+        run: echo ${{ secrets.GITHUB_TOKEN }} | docker login docker.pkg.github.com -u ${{ github.repository }} --password-stdin
+
+      - name: Build Docker Image
+        run: cd smoke-tests/play && ../../gradlew jib
```
trask, line 28: now that this dir has gradle?
```suggestion
        run: cd smoke-tests/play && ./gradlew jib
```

.github/workflows/build-springboot-smoke-dist.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Cache gradle dependencies
+        uses: burrunan/gradle-cache-action@v1
+        with:
+          job-id: jdk11
+
+      - name: Login to GitHub Package Registry
+        run: echo ${{ secrets.GITHUB_TOKEN }} | docker login docker.pkg.github.com -u ${{ github.repository }} --password-stdin
+
+      - name: Build Docker Image
+        run: cd smoke-tests/springboot && ../../gradlew jib
```
trask, line 28: same
```suggestion
        run: cd smoke-tests/springboot && ./gradlew jib
```

smoke-tests/play/README.md
```diff
@@ -0,0 +1,6 @@
+# Play framework smoke test
+Play application used by smoke tests of OpenTelemetry java agent.
+Builds and publishes Docker image containing a trivial Play application.
+
+This is a separate gradle project, independent from the rest. This was done to allow
+to build and publish it only when actually needed and not on every project build.
```
trask, line 6: 👍

smoke-tests/springboot/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+  id 'org.springframework.boot' version '2.3.2.RELEASE'
+  id 'io.spring.dependency-management' version '1.0.9.RELEASE'
+  id 'java'
+  id 'com.google.cloud.tools.jib' version '2.4.0'
+}
+
+group = 'io.opentelemetry'
+version = '0.0.1-SNAPSHOT'
+sourceCompatibility = '11'
```
trask, line 10: i think nice to keep our smoke tests targeted to Java 8, even if we are only running them on Java 11 currently
```suggestion
sourceCompatibility = '8'
```

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -0,0 +1,142 @@
...
+  protected GenericContainer target = new GenericContainer<>(getTargetImage())
+    .withExposedPorts(8080)
+    .withNetwork(network)
+    .withLogConsumer(new Slf4jLogConsumer(logger))
+    .withFileSystemBind(agentPath, "/opentelemetry-javaagent-all.jar")
+    .withEnv("JAVA_TOOL_OPTIONS", "-javaagent:/opentelemetry-javaagent-all.jar")
+    .withEnv("OTEL_BSP_MAX_EXPORT_BATCH", "1")
+    .withEnv("OTEL_BSP_SCHEDULE_DELAY", "10")
+    .withEnv("OTEL_OTLP_ENDPOINT", "collector:55680")
+    .withEnv(extraEnv)
```
trask, line 56: nice 👍

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/WildflySmokeTest.groovy
```diff
@@ -0,0 +1,46 @@
...
+import okhttp3.Request
+
+class WildflySmokeTest extends SmokeTest {
+
+  protected String getTargetImage() {
+    "jboss/wildfly:latest"
+  }
+
+  //We don't have support for Wildfly Undertow server yet.
+  //So this test just verifies that Wildfly has come up.
```
trask, line 29: i think this should work now with real span verification, the reason it didn't before is because we were verifying from logging exporter and wildfly messed with stdout/stderr, i can try it out after the merge

iNikem:
Hm, indeed probably there is no need for several Gradle wrappers in one repo...

.github/workflows/pr.yaml
iNikem: I think we still want to protect PRs from flaky test.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/WildflySmokeTest.groovy
iNikem: I haven't seen any spans generated during this test

iNikem:
@anuraaga I still prefer to add Gradle wrappers into those projects. They are meant to be developed independently. You most probably will end up with a separate Idea project for them and it will be nice if they are self-contained. Gradle wrappers are small, less than 80K

.github/workflows/pr.yaml
trask: just realized, this will slow down feedback (3x?) on real failing builds, which are common in PRs

iNikem: Not exactly 3x, because already passed tests will not be re-run. But you are right. Will see how this goes...

PR #943, opened 2020-08-11
Use better gradle cache action for local Gradle cache

iNikem:


.github/workflows/nightly.yaml
```diff
@@ -6,34 +6,6 @@ on:
...
-          # TODO replace/augment with remote gradle cache
-          key: build-${{ env.cache-name }}-${{ env.TODAY }}
-          restore-keys: |
-            build-${{ env.cache-name }}-
-
-      - name: Build with Gradle
-        run: ./gradlew assemble testClasses --stacktrace
-
   test:
     needs: build
```
trask, line 10:
```suggestion
```

iNikem:
I talked to Vladimir (the author of the action) today about our setup. The way his action handles different jdk versions for the same job (as in our case) effectively precludes from having that separate "warm-up" step. So I had to remove it for now and see how will that go

Issue #764, opened 2020-07-23
Opentracing Shim Propagation Errors


FrankSpitulski:
The https://github.com/open-telemetry/opentelemetry-java/tree/master/opentracing_shim throws the following errors when it tries to inject context, but this is supposed to be a supported use case as per @trask's guidance. I invoked this by using the 0.6.0 agent, and shim and performing an okhttp opentracing request.

```
[opentelemetry.auto.trace 2020-07-22 17:29:07:517 -0700] [http-nio-8080-exec-2] DEBUG io.opentelemetry.auto.tooling.Instrumenter$Default - Applying instrumentation: opentelemetry-api -- io.opentelemetry.auto.instrumentation.opentelemetryapi.TracingContextUtilsInstrumentation on sun.misc.Launcher$AppClassLoader@18b4aac2
[...]

PR #947, opened 2020-08-12
Instrument DefaultSpan.create to return a bridged span. 

anuraaga:
Half of #764 - I realized that the issue actually shows two exceptions, Unexpected span and Unexpected context. Will continue to debug but figured it's worth getting this in first since it's ready.


instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/auto/instrumentation/opentelemetryapi/DefaultSpanInstrumentation.java
```diff
@@ -0,0 +1,56 @@
...
+    return Collections.singletonMap(
+        isMethod().and(isPublic()).and(isStatic()).and(named("create")),
+        DefaultSpanInstrumentation.class.getName() + "$CreateAdvice");
+  }
+
+  public static class CreateAdvice {
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    public static void methodExit(
+        @Advice.Return(readOnly = false) unshaded.io.opentelemetry.trace.Span span) {
+      span = Bridging.toUnshaded(DefaultSpan.create(Bridging.toShaded(span.getContext())));
```
iNikem, line 53: 😭

iNikem:
Don't wait for me, I cannot bring myself to approving this :)

PR #953, opened 2020-08-12
Better naming around around shaded/unshaded in OpenTelemetry API instrumentation

### LINKED ISSUE #935: Reduce confusion around the "unshaded" package name
We do a strange shading-unshading of OpenTelemetry API in order to instrument the "unshaded" (application's) OpenTelemetry API, while sending telemetry to the "shaded" (agent's) OpenTelemetry API.

But in the intermediate step, we shade OpenTelemetry API under "unshaded." prefix (which will later be unshaded), and we use an unshaded OpenTelemetry API (which will eventually be shaded).

Looking at the code, and talking about the code in this intermediate step is very confusing.

From https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/931#issuecomment-671015917:

> These shaded and unshaded terms are very confusing to me and don't seem natural. Can we replace them with some words that reflect the problem domain, not the solution space?

> I mean we have one set of classes that we bring as dependency, right? And another set which application can bring, right? So maybe agent/application or internal/external or something?

I like this idea of not using the package names "shaded" or "unshaded" anywhere.

trask:
Closes #935


trask:
> It indeed seems much better than before

👍 ya, I can revisit again, see what you think next time you're in this code

Issue #754, opened 2020-07-22
Package rename: io.opentelemetry.auto.instrumentation -> io.opentelemetry.instrumentation.auto


anuraaga:
Boring but I think we do need to make this change to match our artifacts and follow the general pattern of a common prefix for packages in a project.

trask:
this will give us something like:

* `io.opentelemetry.instrumentation.armeria`
* `io.opentelemetry.instrumentation.apachehttpclient`
* ...
* `io.opentelemetry.instrumentation.vertx`
* <b>`io.opentelemetry.instrumentation.api`</b>
* <b>`io.opentelemetry.instrumentation.auto`</b>
  * `io.opentelemetry.instrumentation.auto.armeria`
  * `io.opentelemetry.instrumentation.auto.apachehttpclient`
  * ...
  * `io.opentelemetry.instrumentation.auto.vertx`
  * <b>`io.opentelemetry.instrumentation.auto.api`</b>

i like this.

but i'm not quite sure how `auto-tooling`, `auto-bootstrap` fit in here best.

putting these under `io.opentelemetry.instrumentation.auto` pollutes that namespace which would be nice to be reserved for instrumentation ( + the `api` package).

so maybe:

* `io.opentelemetry.instrumentation.agent.tooling`
* `io.opentelemetry.instrumentation.agent.bootstrap`

or even better i think:

* `io.opentelemetry.javaagent.tooling`
* `io.opentelemetry.javaagent.bootstrap`

after #884, there won't be circular dependency between `io.opentelemetry.javaagent` and `io.opentelemetry.instrumentation`, which was something i didn't like about this idea earlier https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/597#issuecomment-651429565

iNikem:
If our build scripts and other tooling can cope with 2 namespaces, then I think io.opentelemetry.javaagent is a good one

trask:
While doing this, need a better name for `io.opentelemetry.auto.bootsrap.Agent`, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/955#discussion_r469707190

PR #955, opened 2020-08-13
Rename Pre-Main class

### LINKED ISSUE #924: Rename io.opentelemetry.auto.bootstrap.AgentBootstrap
Updated javaagent-jar-components.md in #884 makes it seem strange that this class named "bootstrap" is the 1 class that is not in the bootstrap class loader. It does "bootstrap" the agent, but I think better to rename and avoid this confusion. Since this is the `Premain-Class`, we should probably do this on the sooner side if we are going to do it.

As far as what to rename it to.... maybe `io.opentelemetry.auto.Agent`?

trask:
Closes #924


opentelemetry-javaagent/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.auto.bootstrap;
+package io.opentelemetry.javaagent;
```
iNikem, line 17: Are you sure we don't want to keep `auto` part? `io.opentelemetry.auto.javaagent`?

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/Agent.java
```diff
@@ -27,7 +27,7 @@
 /**
```
anuraaga, line 27: We probably need a better name for this file, `AgentBootstrap`, `AgentDelegate`, etc

opentelemetry-javaagent/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
trask: I did this to be forward compatible with the proposal in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/754#issuecomment-671011320, in particular this part at the end of that comment:

* io.opentelemetry.javaagent.tooling
* io.opentelemetry.javaagent.bootstrap

agent-bootstrap/src/main/java/io/opentelemetry/auto/bootstrap/Agent.java
trask: 👍 will do this as part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/754#issuecomment-673275367

PR #971, opened 2020-08-14
Use startScope() on all HttpClientTracers

### LINKED ISSUE #465: Generic mechanism for preventing multiple Server/Client spans
[Otel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spankind) says the following about SERVER and CLIENT spans:

- `SERVER` [...] span is the child of a remote CLIENT
- `CLIENT` [...] span is the parent of a remote SERVER

Together with the remaining section this brings me to the conclusion, that spec demands having at most 1 span of kind `SERVER` while a trace passes single JVM. The same for `CLIENT`. Partially this requirement is already encoded into instrumentations. E.g. incoming HTTP instrumentations attach entry-point span to a request and avoid creating new spans if they find one already attached to a request. But I propose to make this requirement more explicit.

Let us do the following:

1. Whenever `SERVER` or `CLIENT` span is created, associate it with the current Context under the corresponding key
2. Before attempting to create new `SERVER` or `CLIENT` span instrumentations should check if one is already present in the current context. If yes, then no new span should be created.
3. Until we have `semantic tracers`, these methods can leave in `BaseDecorator`.
4. Explicitly document that instrumentation may augment already existing `SERVER` or `CLIENT` span with additional information. E.g. to change span name or to add transport level attributes.

This will directly address #440 as well.

### LINKED ISSUE #964: Add a 3rd generic parameter to HttpClientTracer for HEADERS
`HttpClientTracer.startScope()` and `.getSetter()` are not working in some places because the object we want to inject the header into does not have the same type as the request, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/922#discussion_r470166242

If there are a few of these places, it may make sense to add a 3rd generic parameter to HttpClientTracer for HEADERS, to make `startScope()` and `getSetter()` more broadly applicable.

trask:
Closes #964

Also closes #465


instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/spring/webflux/client/WebClientTracingFilter.java
```diff
@@ -66,6 +62,7 @@ public static void addFilter(
               })
           .doOnCancel(
               () -> {
+                TRACER.onCancel(span);
```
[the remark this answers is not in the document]
anuraaga: Is it possible to add a unit test for it?

instrumentation/kubernetes-client-7.0/src/main/java/io/opentelemetry/instrumentation/auto/kubernetesclient/KubernetesClientTracer.java
```diff
@@ -55,6 +55,7 @@ protected String responseHeader(Response response, String name) {
 
   @Override
   protected Setter<Request> getSetter() {
+    // TODO (trask) no propagation implemented yet?
```
anuraaga, line 58: Kubernetes client is a wrapper in okhttp instrumentation I think. We probably want to refactor this to inherit from the okhttp tracer at some point, assuming it's technical feasible

instrumentation/netty/netty-3.8/src/main/java/io/opentelemetry/instrumentation/auto/netty/v3_8/client/NettyHttpClientTracer.java
```diff
@@ -22,19 +22,34 @@
...
+public class NettyHttpClientTracer
+    extends HttpClientTracer<HttpRequest, HttpHeaders, HttpResponse> {
   public static final NettyHttpClientTracer TRACER = new NettyHttpClientTracer();
 
+  @Override
+  public Scope startScope(Span span, HttpHeaders headers) {
+    if (!headers.contains("amz-sdk-invocation-id")) {
+      return super.startScope(span, headers);
+    } else {
+      // TODO (trask) if we move injection up to aws-sdk layer, and start suppressing nested netty
```
anuraaga, line 44: Hmm I remember hoping to have removed this check but bringing it back but not much else. Will check it out to try to remember

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/spring/webflux/client/WebClientTracingFilter.java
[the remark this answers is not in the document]
trask: yeah, good idea, i created #1011 to track this

instrumentation/kubernetes-client-7.0/src/main/java/io/opentelemetry/instrumentation/auto/kubernetesclient/KubernetesClientTracer.java
trask: Opened #1012

instrumentation/grizzly-client-1.9/src/main/java/io/opentelemetry/instrumentation/auto/grizzly/client/GrizzlyClientTracer.java
```diff
@@ -16,14 +16,16 @@
...
 import com.ning.http.client.Request;
 import com.ning.http.client.Response;
 import io.opentelemetry.context.propagation.HttpTextFormat.Setter;
 import io.opentelemetry.instrumentation.api.tracer.HttpClientTracer;
 import java.net.URI;
 import java.net.URISyntaxException;
 
-public class GrizzlyClientTracer extends HttpClientTracer<Request, Response> {
+public class GrizzlyClientTracer
+    extends HttpClientTracer<Request, FluentCaseInsensitiveStringsMap, Response> {
```
iNikem, line 28: Are you sure it wouldn't be better `HttpClientTracer<Request, Request, Response>`?

instrumentation/jaxrs-client/jaxrs-client-1.1/src/main/java/io/opentelemetry/instrumentation/auto/jaxrsclient/v1_1/InjectAdapter.java
```diff
@@ -16,15 +16,15 @@
...
 
-public final class InjectAdapter implements HttpTextFormat.Setter<ClientRequest> {
+public final class InjectAdapter implements HttpTextFormat.Setter<MultivaluedMap<String, Object>> {
 
   public static final InjectAdapter SETTER = new InjectAdapter();
 
   @Override
-  public void set(ClientRequest clientRequest, String key, String value) {
-    clientRequest.getHeaders().putSingle(key, value);
+  public void set(MultivaluedMap<String, Object> carrier, String key, String value) {
```
iNikem, line 27: I am not sure this change made things better. Why replaces `ClientRequest` with `MultivaluedMap<String, Object>` here?

instrumentation/grizzly-client-1.9/src/main/java/io/opentelemetry/instrumentation/auto/grizzly/client/GrizzlyClientTracer.java
trask: `FluentCaseInsensitiveStringsMap` most closely represents the `HEADERS` type, but if you still prefer `Request` I don't mind and will change it

instrumentation/jaxrs-client/jaxrs-client-1.1/src/main/java/io/opentelemetry/instrumentation/auto/jaxrsclient/v1_1/InjectAdapter.java
trask: same

instrumentation/grizzly-client-1.9/src/main/java/io/opentelemetry/instrumentation/auto/grizzly/client/GrizzlyClientTracer.java
iNikem: Yeah, my comments probably actually mean "are you sure that HEADERS is the best candidate for this new type parameter"? I have review around half of the changed files and it seems to me that Request is actually enough for like 90%+ cases. Even if actually need a separate type parameter for those 10%, it does not make sense to force that HEADERS name to the letter and use such strange types.

Rename HEADER type to [Carrier](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#carrier) and use the simplest/most natural type?

trask: oh yes I like renaming `HEADERS` type to `CARRIER` will do

trask: done

instrumentation/jaxrs-client/jaxrs-client-1.1/src/main/java/io/opentelemetry/instrumentation/auto/jaxrsclient/v1_1/InjectAdapter.java
trask: and done

instrumentation/jaxrs-client/jaxrs-client-2.0/src/main/java/io/opentelemetry/instrumentation/auto/jaxrsclient/v2_0/JaxRsClientTracer.java
```diff
@@ -16,14 +16,17 @@
...
 import io.opentelemetry.context.propagation.HttpTextFormat.Setter;
 import io.opentelemetry.instrumentation.api.tracer.HttpClientTracer;
 import java.net.URI;
 import javax.ws.rs.client.ClientRequestContext;
 import javax.ws.rs.client.ClientResponseContext;
+import javax.ws.rs.core.MultivaluedMap;
 
 public class JaxRsClientTracer
-    extends HttpClientTracer<ClientRequestContext, ClientResponseContext> {
+    extends HttpClientTracer<ClientRequestContext, MultivaluedMap, ClientResponseContext> {
```
iNikem, line 29: Carrier can be ClientRequestContext here as well

PR #977, opened 2020-08-14
Create abstraction for library dependencies for instrumentation.

anuraaga:
Adds three configurations `library`, `testLibrary`, `latestDepTest` library as described [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/977/files#diff-350ee91265ec7ec178354ff35e08b597)

When Gradle is run with `-PtestLatestDeps=true`, the standard `testImplementation` configuration is overridden with the latest dep test wildcards. Without the property, it's as if the wildcards don't exist. This means IntelliJ won't read the wildcard versions anymore. The bottom-left corner of IDE with all the external libraries is still huge, but far smaller than before, presuambly speeding up some IntelliJ indexing / refactoring operations. Most importantly, now IntelliJ won't ever freeze to refresh dynamic versions in the middle of development.


iNikem:
Have you tried it with some module where latestDeps differ significantly from `compileOnly`? If we have such.

If it works, it is a very good simplification :)

anuraaga:
I'm continuing to play with this PR and think I have the concepts mostly ready. I've ended up defining three configurations

`library` - added to `compileOnly` and `testImplementation` as is normally, and when `-PtestLatestDeps=true`, instead added to `testImplementation` with the version override of `+`

`testLibrary` - an additional library needed for testing (usually a runtime provider or framework of some sort). added to `testImpelementation` as is normally, and when `-PtestLatestDeps=true`, instead added to `testImplementation` with the version override of `+`

`latestDepTestLibrary` - essentially a version override for the latest dep test when `+` isn't appropriate. When `-PtestLatestDeps=true`, added to testImplementation as is without overriding version.

A couple of takeaways

- I didn't do an official count, but it seems like about half, maybe a bit more, of the current `latestDepTestImplementation` wildcards are `+`. These have a lot of boilerplate removed by just using `library`. But there is some cognitive load, an alternative is to remove `testLibrary` and always have the version override explicit. Different from our current solution is this would still only happen when `-PtestLatestDeps=true` and the win for the IDE development will still be there.

- Some tests have `testLatestDeps` source set. Some of these don't even have a wildcard for the `latestDepTestImplementation` version in which case I will rename them completely and let them run on the normal build (similar to my recent jms change). Some of them do have a wildcard, so these will still use the current solution and I'll add `enabled = findProperty(testLatestDeps)`, but the IDE benefit won't exist. There are very few of these, I may try to find a solution even for them at some point but am not too worried right now.

Let me know if you have any thoughts on approach, naming, etc.

anuraaga:
Hopefully this is ready.

We can see the difference in dependencies in build scans like these
https://scans.gradle.com/s/glng4kkzyzzw2/dependencies?toggled=W1szOF0sWzM4LDExXV0
https://scans.gradle.com/s/2ko2puzwhly24/dependencies?toggled=W1szOF0sWzM4LDExXV0

anuraaga:
One point is there are just a few tests left that have actually separate sources for `latestDepTest` - this is exactly what the test-sets plugin is for :) As a follow up I will think about them a little more to see if something can be simplified but will probably leave them as is.

instrumentation/servlet/servlet-common/servlet-common.gradle
```diff
@@ -22,7 +17,4 @@ dependencies {
   }
   testImplementation group: 'org.eclipse.jetty', name: 'jetty-server', version: '7.0.0.v20091005'
   testImplementation group: 'org.eclipse.jetty', name: 'jetty-servlet', version: '7.0.0.v20091005'
```
trask, line 19: i think?
```suggestion
  testLibrary group: 'org.eclipse.jetty', name: 'jetty-server', version: '7.0.0.v20091005'
  testLibrary group: 'org.eclipse.jetty', name: 'jetty-servlet', version: '7.0.0.v20091005'
```

anuraaga: Oops, no it wasn't - the previous `latestDepTest` was a ghost, it didn't specify the source set :P #1051

PR #979, opened 2020-08-14
Implement Armeria instrumentation context propagation.

anuraaga:
This is both to provide a good mechanism for tying our context to Armeria's for users, and also food for thought in case the implementation can inspire something in other libraries. I don't see any reason a similar concept wouldn't work with Akka, etc.


instrumentation/armeria-1.0/auto/armeria-1.0-auto.gradle
```diff
@@ -22,3 +22,9 @@ dependencies {
 
   testImplementation project(':instrumentation:armeria-1.0:testing')
 }
+
+tasks.withType(Test) {
+  // Armeria has its own mechanism for context propagation which we don't want our automatic context
+  // propagation mechanism to interfere with.
+  jvmArgs '-Dotel.integration.java_concurrent.enabled=false'
```
iNikem, line 29: How this will work in production where `java_concurrent` instrumentation IS enabled by default? Don't you want to test in realistic env?

instrumentation/armeria-1.0/library/src/main/java/io/grpc/override/ContextStorageOverride.java
```diff
@@ -0,0 +1,91 @@
...
+package io.grpc.override;
+
+import com.linecorp.armeria.common.RequestContext;
+import io.grpc.Context;
+import io.grpc.Context.Storage;
+import io.netty.util.AttributeKey;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ContextStorageOverride extends Storage {
```
iNikem, line 26: First, javadoc please :)

Second, having this class will override context storage for the whole JVM with our agent. Including all our instrumentations. And may change the before of the application, if it already has its own override.

instrumentation/armeria-1.0/auto/armeria-1.0-auto.gradle
anuraaga: Yeah I don't have a good story for this yet - I tried to inject the context storage override and play with this flag disabled but the bootstrap classloader issue I ran into is problematic. I may need to just instrument the context methods themselves. I will follow up for it.

Interestingly, even with concurrent instrumentation, my unit test fails - I don't do anything particularly weird in it I think, just call `CompletableFuture` :(

instrumentation/armeria-1.0/library/src/main/java/io/grpc/override/ContextStorageOverride.java
anuraaga: Yeah - I think this is better behavior for Armeria users than the default. When Armeria context isn't available, the behavior is the same as normal so I'm not terribly worried about affecting the JVM.

But this gRPC API is silly so is problematic... indeed most problematic is if the user wants to override this themselves. Best I can think of is to provide our implementation (I extracted into a separate class) so they can delegate to it.

iNikem (approved):

instrumentation/armeria-1.0/library/src/main/java/io/grpc/override/ContextStorageOverride.java
trask: Have you tested this with the full javaagent? I think (shaded) `io.grpc.override.ContextStorageOverride` will be in the AgentClassLoader, while (shaded) `io.grpc.Context` will be in the bootstrap class loader, so may not find the (shaded) `io.grpc.override.ContextStorageOverride`.

anuraaga: Yeah I think this is why the `auto` unit tests don't work. I'll continue to follow up on it, but any quick ideas on why the java_concurrent instrumentation wouldn't be enough anyways? https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/979#discussion_r472839958

Issue #882, opened 2020-08-03
Save "artifacts" at the end of GitHub Nightly build


trask:
Capturing the same things as we are capturing for CircleCI via `collect_libs.sh`, `collect_reports.sh` and `collect_results.sh`

From what I can tell this looks like standard action https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts, and we can only upload single files or zips. Which is too bad, I like being able to browse and link to the html artifacts in CircleCI, but not a deal breaker.

iNikem:
I have some thoughts about publishing test reports.

Is it enough to have jar files published to oss.jfrog.com or do you want them in GitHub UI as well?

trask:
I think they will be much more discoverable (e.g. new contributors looking at why their PR build failed) if they are stored in the normal GitHub actions artifacts, so would prefer that unless we can store them somewhere where you can navigate and link to the html reports, in which case those benefits may outweigh GitHub actions artifacts.

trask:
The build scans look really nice for investigating failures, e.g.

https://scans.gradle.com/s/a3rvevp7invbc/tests/failed

Can we make them more visible somehow, e.g. a link to it in the artifacts section?

trask:
E.g. this build report makes it look like `AkkaExecutorInstrumentationTest` caused the build to fail:

https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-instrumentation/2700/workflows/d458b21f-fb47-411d-9909-0a0c641ffd4a/jobs/16258/tests

but really it was retried and then succeeded, and it was another test that caused the build to fail (which is reported lower down in the above report).

While the build scan makes it clear that `AkkaExecutorInstrumentationTest` was FLAKY, and others really FAILED.

iNikem:
Visible links to Gradle build scans are in my pipeline

iNikem:
Nightly build runs have links (unfortunately not clickable) to all build scans: https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/253140819

PR #980, opened 2020-08-14
Try CiMate

iNikem:
This PR adds CiMate integration.

CiMate is an incubating project @bsideup providing test report aggregation. For now this will add a status link to PRs which points to easy to read test report UI. CiMate currently does not have over the time aggregation view which would help to detect flaky tests, but there are such plans.

Let us try it out if it will make our life easier :)

Partially addresses #882

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -35,3 +35,13 @@ jobs:
           command: ./gradlew testJava${{ matrix.java }} --stacktrace
           timeout_minutes: 60
           max_attempts: 3
+
+      - name: Aggregate test reports with ciMate
+        if: always()
+        continue-on-error: true
+        env:
+          CIMATE_PROJECT_ID: mz1jo49x
+        run: |
+          wget -q https://get.cimate.io/release/linux/cimate
+          chmod +x cimate
+          ./cimate -v "**/TEST-*.xml"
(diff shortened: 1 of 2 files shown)


bsideup:
Hey! One of ciMate's creators here. We're in contact with GitHub's support on why the check isn't appearing. It seems to be some exceptional case (on GH's side, since `check_suite` isn't being sent to ciMate's API), probably because @iNikem used to have an app installed on his fork.

But the service itself seems to be working, here is the aggregation for the last commit on this PR:
https://app.cimate.io/mz1jo49x/c6117a4 (the format in `https://app.cimate.io/$projectId/$commitPrefix`)

bsideup:
While waiting for GitHub's response, I will try to analyze why some test suites are marked as "failed".
As @iNikem suggested, it may be that they are being re-run.

Despite all of said issues, I hope ciMate will bring at least some help when it comes to analysing CI test failures, and we have a lot of ideas in our roadmap that should make it easier to pinpoint the issue and get more sense out of the test results.
I kinda like how it matches OT's goal of getting the most out of otherwise raw traces 😊

trask:
thx for the help @bsideup!

bsideup:
FTR as a workaround, I enabled the GitHub Check. It requires an additional click, but at least should work with PRs.

PR #982, opened 2020-08-14
Add S3 cache

### LINKED ISSUE #490: Use Gradle build cache
It takes a lot of time to build instrumentation project. But the majority of that time is spent building and testing tons of individual instrumentations. As developers often make changes to a specific instrumentation, there is no sense in building all other, unrelated, instrumentations. [Gradle build cache](https://docs.gradle.org/current/userguide/build_cache.html) can help a lot here. One easy way to start using it is via [this plugin](https://github.com/myniva/gradle-s3-build-cache)

iNikem:
Enables remote Gradle cache on S3, closes #490

First, S3 bucket permissions should be rechecked based on https://github.com/burrunan/gradle-s3-build-cache#s3-bucket-permissions-for-cache-population. Cache should be readable for everybody, without credentials. In order to write to cache, we have to provide credentials. Plugin does require its own pair of env variables (not to mess with default configs locally). GHA populate them from secrets, which are available only on master branch. This is automatically guarantee that only nighty build on master can populate cache. And all other builds, including local on developers machines, can read that cache.

This is complimentary to GHA Gradle cache that we already have. Both will do their part to, hopefully, reduce our build times .


.github/workflows/nightly.yaml
```diff
@@ -34,6 +34,9 @@ jobs:
 
       - name: Test with Gradle
         uses: nick-invision/retry@v1
+        env:
+          S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
```
anuraaga, line 38: Who's key is this? Hope it's CNCF's.

iNikem: right now - Trask :) We are still waiting on https://github.com/open-telemetry/community/issues/392

trask: yeah, we should get CNCF's soon, but wanted to unblock us in the short term

trask:
Policy for the `ACCESS_KEY_ID` / `SECRET_KEY`:

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:PutObject"
            ],
            "Resource": [
                "arn:aws:s3:::otel-gradle-cache-temp/*"
            ]
        }
    ]
}
```

And public access is not blocked, so I think that means anyone should be able to `GetObject`?

PR #996, opened 2020-08-15
Use opentelemetry-java-instrumentation-gradle-test1 for build cache.

anuraaga:


settings.gradle
```diff
@@ -22,7 +22,7 @@ apply plugin: 'com.github.burrunan.s3-build-cache'
 buildCache {
   remote(com.github.burrunan.s3cache.AwsS3BuildCache) {
     region = 'us-west-2'
-    bucket = 'otel-gradle-cache-temp'
+    bucket = 'opentelemetry-java-instrumentation-gradle-test1'
```
iNikem, line 25: Why `test1`? :) Can we have "proper" name? Like `...-gradle-cache`?

anuraaga: Somehow I read temp as test and added a 1 expecting to mess up a bucket or two (well technically I messed up one bucket which didn't have a 1 ;) Let's stick with this for now since we'll probably change buckets later anyways.

iNikem: How can you "mess the bucket"? :) You can delete and recreate one, no?

anuraaga: It takes a long time to be able to reuse the bucket name. Wasn't ready yet after about 30 min.

PR #997, opened 2020-08-15
Disable local build cache on CI

anuraaga:
With the addition of the remote build cache, we probably will save by not unnecessarily writing to local build cache / restoring it.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -39,7 +39,7 @@ jobs:
           S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
           S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
         with:
-          command: ./gradlew testJava${{ matrix.java }} --stacktrace
+          command: ./gradlew testJava${{ matrix.java }} --stacktrace -PpushBuildCache=true
           timeout_minutes: 60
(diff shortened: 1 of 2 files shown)


iNikem:
I am not sure at all about that. Remote cache is populate from master branch during nightly build only. So if one PR has to be build several times due to fixes, it will still benefit from local cache.

Also, remote case does not contain dependencies. Downloading them from local GHA cache as one blob is faster than usual remote downloading.

Remote and local cache are NOT "one or another". They nicely complement each other.

anuraaga:
I noticed on circleci building took just 4 or so minutes and it took an entire 5 min to save the cache. Not sure if build cache is a large part of that but figured it could be. Agree they work well together on local machine but on CI where it has to be saved and restored, I suspect it will generally be a loss.

Agree that we still need to use a cache for module dependencies this is just about the build cache. 

I think we can enable pushing build cache on master build as well instead of just nightly. I think even for PR rebuilds this should allow enough to come from cache in most cases, without having to save and restore the entire local cache every time.

iNikem:
CircleCI is totally different story that I don't care about at all. It will be deprecated soon anyway. I am speaking about GHA and how they use cache.

Please, let me take my work on caching to its conclusion. We still don't have reliably working S3 cache, so we just cannot make any informed decisions.

iNikem:
Build cache, as in "cached outputs of Gradle tasks" takes less than 100M. Compare that to 1.5G of dependencies.

anuraaga:
Ah that's a big difference between cache and dependencies - ok let's see how it goes

PR #998, opened 2020-08-15
Use property to control build cache push instead of CI env variable s…

anuraaga:
…ince many builds have that set.

I noticed that the builds for non-pushing ones, like the PR build, have an authentication failure at the top. I think it's better to more precisely control the check so we don't have the error message and possible trouble - I'm not sure about this S3 cache in particular, but I've used a cache that would be disabled completely including reads if there is ever an error during an operation.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -23,7 +23,7 @@ buildCache {
   remote(com.github.burrunan.s3cache.AwsS3BuildCache) {
     region = 'us-west-2'
     bucket = 'opentelemetry-java-instrumentation-gradle-test1'
-    push = 'true'.equals(gradle.startParameter.projectProperties.get('pushBuildCache'))
+    push = System.getenv('S3_BUILD_CACHE_ACCESS_KEY_ID') != null
   }
 }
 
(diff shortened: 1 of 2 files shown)


anuraaga:
Ah yeah found the error message in a previous build

"The remote build cache was disabled during the build due to errors."

so hoping this fixes it so remote cache is actually used.

settings.gradle
```diff
@@ -23,7 +23,7 @@ buildCache {
   remote(com.github.burrunan.s3cache.AwsS3BuildCache) {
     region = 'us-west-2'
     bucket = 'opentelemetry-java-instrumentation-gradle-test1'
-    push = isCI
+    push = 'true'.equals(gradle.startParameter.projectProperties.get('pushBuildCache'))
```
iNikem, line 26: Why not `push = project.hasProperty('pushBuildCache')`?

anuraaga: This is settings.gradle so there's no project here. But I changed to check env instead which I think is closer to what you had in mind in our chat conversation

PR #1013, opened 2020-08-16
Change groovy import order

### LINKED ISSUE #816: Change groovy import layout to match java import layout
Currently our docs say to use the default Intellij import layout for groovy files (https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/intellij-setup.md).

@iNikem would like the groovy import layout to match the java import layout https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/701#discussion_r456217872, which makes sense, we just need to update our docs and do a one-time reformat.

trask:
Closes #816


anuraaga:
Apparently global refactoring of Groovy is hard, who'd have thought? :P

trask:
it seems like it broke because an import was lost, but whenever i put that import back in it gets removed because Intellij thinks its unnecessary (which seems correct to me).

anyways, worked around it... (well, probably should speak too soon, until CI has it's say)

PR #1017, opened 2020-08-16
Remove thread utils module

### LINKED ISSUE #989: Get rid of separate thread-utils module
If (really) needed by instrumentation, merge into `instrumentation-api`, otherwise merge into `javaagent-tooling`

trask:
Closes #989

Also, related, removed thread pool used by `cassandra-3.0` instrumentation, since we don't do this for any other instrumentation it seems unnecessary here.


instrumentation/cassandra/cassandra-3.0/src/main/java/io/opentelemetry/instrumentation/auto/cassandra/v3_0/TracingSession.java
```diff
@@ -150,7 +144,7 @@ public ResultSetFuture executeAsync(String query) {
     Span span = TRACER.startSpan(session, query);
     try (Scope ignored = TRACER.startScope(span)) {
       ResultSetFuture future = session.executeAsync(query);
-      future.addListener(createListener(span, future), EXECUTOR_SERVICE);
+      future.addListener(createListener(span, future), SameThreadExecutor.INSTANCE);
```
anuraaga, line 147: Probably not for this PR but should use `Futures.addCallback` to avoid recording a wrapped exception.

PR #1028, opened 2020-08-17
Remove latestDepTest from rxjava-1.0 which doesn't have any test sour…

anuraaga:
…ces.


iNikem:
Probably more correct solution is to add tests? :)

anuraaga:
Yeah should have filed an issue about that! 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1033

I think this change still stands separately.

iNikem:
Sorry for nitpicking, but if we solve #1033, we will add `latestDepTest` back, no?

anuraaga:
Yeah, though it might be a different form, just `library` :) Let's solve that in #1033 not here, right?

PR #1029, opened 2020-08-17
Separate out shared test sources for couchbase/play-ws to another pro…

anuraaga:
…ject.

I noticed we use a pattern of copying `src/test` files for use in other tests, which relies on the test sets plugin for something called `testJar`. I couldn't find how this actually works (why the parent tests aren't run in the child project), but anyways it seems better to separate the tests into testing projects to use Gradle's normal functionality.


anuraaga:
GitHub's finding some file moves as full diffs, but the only change is the class name (added Abstract to prefix if couchbase tests)

settings.gradle
```diff
@@ -67,6 +67,7 @@ include ':instrumentation:cassandra:cassandra-4.0'
 include ':instrumentation:cdi-testing'
 include ':instrumentation:couchbase:couchbase-2.0'
 include ':instrumentation:couchbase:couchbase-2.6'
+include ':instrumentation:couchbase:testing'
```
trask, line 70:
```suggestion
include ':instrumentation:couchbase:couchbase-testing'
```

settings.gradle
```diff
@@ -137,6 +138,7 @@ include ':instrumentation:play-ws:play-ws-1.0'
 include ':instrumentation:play-ws:play-ws-2.0'
 include ':instrumentation:play-ws:play-ws-2.1'
 include ':instrumentation:play-ws:play-ws-common'
+include ':instrumentation:play-ws:testing'
```
trask, line 141:
```suggestion
include ':instrumentation:play-ws:play-ws-testing'
```

PR #1030, opened 2020-08-17
Some tweaks to improve gradle cacheability

iNikem:


iNikem:
It seems strange to me that `jar dependsOn shadowJar`. I would expect the reverse. I rather add `assemble dependsOn shadowJar`, wdyt?

anuraaga:
The problem I've found with only assemble depending on jar is dependent projects. When tests for another project run that require the shaded behavior, they run jar, not assemble. @trask's recent changes have stopped this being an issue, but this is why as a general solution I've generally used this trick to essentially "replace jar with shadowJar".

iNikem:
In my eyes it is a bug in dependent project, if it fails to say that it actually wants shadow jar.

Also, how can `jar dependsOn shadowJar` work when jar is disabled? I would assume that gradle will remove the whole section of tasks DAG, including `shadowJar`.

anuraaga:
It's a reasonable assumption and strange admittedly ;) At the same time, if the module is replacing the "empty classifier" with a shadow jar because shadow is an implementation detail of that module itself, it seems to make sense for dependent projects to not have to do anything special. FWIW I wish gradle shadow plugin made this simpler.

iNikem:
Looking at shadowJar issue tracker, it seems the author consciously does not want to do that.

PR #1034, opened 2020-08-17
Add maven/gradle documentation to manual instrumentation instructions.

johnbley:
I had a customer comment that the jump from this README to the Quick Start guide left them confused about what libraries to add (they wound up depending on the SDK and then got lost trying to work with SDK types).  It would be best to be redundant with the base Java README and explicitly spell out here that the `-api` library is how you get started, with the same maven/gradle examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,29 @@ only supports manual instrumentation using the `opentelemetry-api` version with 
 number as the Java agent you are using. Starting with 1.0.0, the Java agent will start supporting
 multiple (1.0.0+) versions of `opentelemetry-api`.
 
-You can use the OpenTelemetry `getTracer` or the `@WithSpan` annotation to
+You'll need to add a dependency on the `opentelemetry-api` library to get started.
+
+### Maven
+
+```xml
+  <dependencies>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -257,7 +257,29 @@ only supports manual instrumentation using the `opentelemetry-api` version with
...
+
+### Gradle
+
+```groovy
+dependencies {
+	compile('io.opentelemetry:opentelemetry-api:0.7.0')
+}
+```
+
+Now you can use the OpenTelemetry `getTracer` or the `@WithSpan` annotation to
```
anuraaga, line 282: Thanks @johnbley, this definitely seems like a nice improvement! The only point I have is this gets us `getTracer` but for `@WithSpan` I believe a different annotations dependency is required.

https://github.com/open-telemetry/opentelemetry-java/tree/master/extensions/auto_annotations

Do you think you can add that too? Or we can follow up in a separate PR, no worries.

johnbley: Nice catch; I'll add that now.

trask:
Thanks @johnbley!

PR #1037, opened 2020-08-17
Instrumentation prioritization

trask:
I find myself making this distinction sometimes on whether something should be considered before GA or after GA, so wanted to document what I thought was sort of "tier 1" instrumentation. Not sure if this is useful for others, or if we can even all agree on such a list, so no worries if this doesn't land.


iNikem:
I don't have strong feelings about that. IF we need such priorities, then this list seems Ok.

trask:
Ok, thanks for feedback, will go ahead and merge

PR #1054, opened 2020-08-19
Better local caching via burrunan/gradle-cache-action

iNikem:


.github/workflows/pr.yaml
```diff
@@ -24,17 +24,14 @@ jobs:
...
         with:
           job-id: jdk${{ matrix.java }}
-
-      - name: Test with Gradle
-        uses: nick-invision/retry@v1
-        with:
-          command: ./gradlew testJava${{ matrix.java }} --stacktrace
-          timeout_minutes: 60
-          max_attempts: 3
+        arguments: testJava${{ matrix.java }} --stacktrace
```
vlsi, line 34:
```suggestion
          arguments: testJava${{ matrix.java }} --stacktrace
```

.github/workflows/nightly.yaml
```diff
@@ -63,20 +57,14 @@ jobs:
...
-      - name: Test with Gradle
-        uses: nick-invision/retry@v1
         env:
           S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
           S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
-        with:
-          command: ./gradlew test -PtestLatestDeps=true --stacktrace
-          timeout_minutes: 60
-          max_attempts: 3
+        arguments: test -PtestLatestDeps=true --stacktrace
```
vlsi, line 67:
```suggestion
          arguments: test -PtestLatestDeps=true --stacktrace
```

.github/workflows/nightly.yaml
```diff
@@ -28,20 +28,14 @@ jobs:
...
-        uses: nick-invision/retry@v1
         env:
           S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
           S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
         with:
-          command: ./gradlew testJava${{ matrix.java }} --stacktrace
-          timeout_minutes: 60
-          max_attempts: 3
+          job-id: jdk${{ matrix.java }}
+        arguments: testJava${{ matrix.java }} --stacktrace -Dorg.gradle.caching.debug=true
```
vlsi, line 38:
```suggestion
          arguments: testJava${{ matrix.java }} --stacktrace -Dorg.gradle.caching.debug=true
```

.github/workflows/nightly.yaml
```diff
@@ -28,20 +28,14 @@ jobs:
         with:
           java-version: 11
 
-      - name: Cache gradle dependencies
+      - name: Test
         uses: burrunan/gradle-cache-action@v1
```
anuraaga, line 32: Looks like gradle-caching-runner or the like will be a good name. Not surprising to see the best caching comes with having full visibility to Gradle.

.github/workflows/nightly.yaml
```diff
@@ -28,20 +28,14 @@ jobs:
         with:
           java-version: 11
 
-      - name: Cache gradle dependencies
+      - name: Test
```
anuraaga, line 31: While we're here how about `Test on ${{matrix.java}}`?

.github/workflows/nightly.yaml
```diff
@@ -63,20 +57,14 @@ jobs:
...
```
anuraaga, line 60: Similarly how about `Test latest deps`?

.github/workflows/pr.yaml
```diff
@@ -24,17 +24,11 @@ jobs:
         with:
           java-version: 11
 
-      - name: Cache gradle dependencies
+      - name: Test
```
anuraaga, line 27: Ditto

anuraaga (approved):
Curious what changed in the action will have to check it out some time. Thanks @inikem @vlsi!

.github/workflows/nightly.yaml
vlsi: Naming is hard :)
Thanks for the suggestion. 

I see JUnit team uses the following, and it [can't be folded to a single command](https://github.com/junit-team/junit5/blob/4688072d1722f4906152351083e0133ae4e78feb/.github/workflows/main.yml#L44-L46), so I'm exploring ways to allow multiple (Gradle? Shell?) commands to be executed (to avoid copy-paste the boring `uses:...`)

```yaml
run: |
  ./gradlew --version
  ./gradlew build
```

.github/workflows/nightly.yaml
iNikem: We already have in job name. Isn't it enough?

trask:
Thanks @vlsi!

Issue #1047, opened 2020-08-18
Change play-2.4 tests to run against 2.4 instead of 2.5


trask:
This comment is no longer valid, as we have netty-3.x instrumentation now:

```
  // Before 2.5, play used netty 3.x which isn't supported, so for better test consistency, we test with just 2.5
  testImplementation group: 'com.typesafe.play', name: 'play-java_2.11', version: '2.5.0'
  testImplementation group: 'com.typesafe.play', name: 'play-java-ws_2.11', version: '2.5.0'
  testImplementation(group: 'com.typesafe.play', name: 'play-test_2.11', version: '2.5.0') {
    exclude group: 'org.eclipse.jetty.websocket', module: 'websocket-client'
  }
```

PR #1056, opened 2020-08-19
Change play-2.4 tests to run against 2.4 instead of 2.5.

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1047](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1047)


dengliming:
@trask It seem doesn't pass the `test(8)`, Can you help take a look?

trask:
hey @dengliming, can you reproduce locally? (either Intellij or command line)?

dengliming:
It's weird. I run testcase by using java8 it works. Can you re-run CI build again?

trask:
still failing 🙁

i forgot, to run locally on different Java version, you'll probably need to run from command line:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/running-tests.md#executing-tests-with-specific-java-version

trask:
Hey @dengliming! Thanks for giving this a try! I'm really not very adept with play framework either 😭.

@iNikem @anuraaga what do you think about re-baselining to Play 2.5 and dropping Play 2.3/2.4? Play 2.4 has been EOL for 3.5 years plus it's not a [P1 instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/ga-requirements.md#p1).

iNikem:
Did we actually support Play 2.4? If there is such huge API difference, does it mean that we always worked only with play 2.5+?

trask:
> Did we actually support Play 2.4? If there is such huge API difference, does it mean that we always worked only with play 2.5+?

Muzzle passes against Play 2.4, which means the API that we instrument hasn't changed, so _probably_ the instrumentation works on play 2.4.

anuraaga:
I'm ok with dropping support, but maybe we can add to rationale doc about why we're dropping and that we would probably still be open to someone interested in play 2.4 contributing tests for it if they are still using it.

PR #1068, opened 2020-08-20
Rationale for not supporting Java 7

trask:
Summarizing decision and rationale from yesterday's meeting on this topic. Please review and provide feedback.

### DIFF
--- a/docs/java-7-rationale.md
+++ b/docs/java-7-rationale.md
@@ -1,0 +1,35 @@
+## Rationale for not supporting Java 7
+
+### Android support is no longer tied to Java 7
+
+Even for supporting old Android API levels:
+
+> If you're building your app using Android Gradle plugin 4.0.0 or higher, the plugin extends
+> support for using a number of Java 8 language APIs <b>without requiring a minimum API level for
+> your app</b>.
+
+(https://developer.android.com/studio/write/java8-support#library-desugaring)
+
+There are some Java 8 APIs that Android does not desugar, but we can use
+[animal sniffer plugin](https://github.com/xvik/gradle-animalsniffer-plugin) to ensure we don't use
+those particular Java 8 APIs that not available in whatever base Android level we decide to support,
+e.g. OkHttp takes this approach to [ensure compliance with Android API level 21](https://github.com/square/okhttp/blob/96a2118dd447ebc28a64d9b11a431ca642edc441/build.gradle#L144-L153)
+
+### Modern test tooling requires Java 8+
+
+Both JUnit 5 and Testcontainers both require Java 8+.
+
+### Auto-instrumentation (Javaagent)
+
+We could run tests against Java 8+ and ensure Java 7 compliance by using similar animal sniffer
+technique as above.
+
+But bytecode instrumentation tends to be much more sensitive to Java versions than normal code, and
+we would lose a lot of confidence in the quality of our Java 7 support without being able to run our
+standard tests against it.
+
+### Library (manual) instrumentation
+
+We believe that Java 7 users are primarily in maintenance mode and not interested in cracking open
+their code anymore and adding library (manual) instrumentation, so we don't believe there is much
+interest in library instrumentation targeting Java 7.


docs/java-7-rationale.md
```diff
@@ -0,0 +1,40 @@
+## Rationale for not supporting Java 7
```
anuraaga, line 1: I like Rationale docs hope we doc out more of them :)

docs/java-7-rationale.md
```diff
@@ -0,0 +1,40 @@
...
+e.g. OkHttp takes this approach to
+[ensure compliance with Android API level 21](https://github.com/square/okhttp/blob/96a2118dd447ebc28a64d9b11a431ca642edc441/build.gradle#L144-L153)
+
+We will use this approach for the `instrumentation-api` module and for any library (manual)
+instrumentation that would be useful to Android developers
+(e.g. library instrumentation for OkHttp).
+
+### Modern test tooling requires Java 8+
+
+Both JUnit 5 and Testcontainers both require Java 8+.
```
anuraaga, line 25:
```suggestion
Both JUnit 5 and Testcontainers require Java 8+.
```

anuraaga (approved):

docs/java-7-rationale.md
trask: I blame your and @adriancole's influence 😁

iNikem (approved):

adriancole:
rationale looks to cover the reasons why. if you want to trump our RATIONALEs, add what you are willing to not do.

ex. "modern" is fair enough, but this concedes implicitly that open-telemetry cannot be used for things not modern, so something else will.

Maybe @codingfabian or others know the fuller impact of how many sites from an existing vendor require Java 1.7 instrumentation.

CodingFabian:
Only commenting since @adriancole pinged me.
From a vendor perspective this makes me happy. We have 1.6 / 1.7 customers and we happily support them with our auto instrumentation.
From a technologist perspective, I would say that not being able to test against 1.7 is just an excuse. It is possible (as a vendor I do that!). One might argue that this is too much effort, and this could be true depending on many factors. I can not, and do not want to judge this.
The true question is what Java 8 library provided class is worth dropping the support. I doubt that any bytecode level or runtime behaviour differences would justify the decision. But I assume this has been clarified in the meeting.

Is that the input you were looking for, Adrian?

adriancole:
> Is that the input you were looking for, Adrian?

it seems github doesn't have a 🍿 reaction, so I need to make a comment instead

trask:
@adriancole @CodingFabian appreciate the feedback!

> how many sites from an existing vendor require Java 1.7 instrumentation.

Oh yes, I will add this to the doc. We know from New Relic's (very large) user base that 2.5% of apps are still on Java 7:

https://blog.newrelic.com/technology/state-of-java/

> From a vendor perspective this makes me happy. We have 1.6 / 1.7 customers and we happily support them with our auto instrumentation.

Yup, we actually discussed this, and it makes us feel less bad about not supporting Java 7 users, knowing that those users have other options for codeless monitoring. I will add this to the doc also.

> I would say that not being able to test against 1.7 is just an excuse

I'm not following this, are you saying it's ok to `--release 7` and only test against Java 8? Or are you saying we should stick to JUnit 4 and not use Testcontainers?

CodingFabian:
I am not proposing anything for OT. I just want to say that Instana runs Java 1.6 tests, and they do include testcontainers based tests. So I guess it really just depends on the effort you want to invest.

Are you saying NewRelic will recommend Instana? quite an honor.

trask:
> Are you saying NewRelic will recommend Instana? quite an honor.

You would have to ask them, but I doubt it 😁

jkwatson:
New Relic's own non-OpenTelemetry agent support java 7, so no. :)

trask:
> Instana runs Java 1.6 tests, and they do include testcontainers based tests. So I guess it really just depends on the effort you want to invest.

Yes, good point, I will add this option to the doc.

adriancole:
ps while I think this is fun banter, it is more important that people know
that in open source this limits the options.

It was fairly routine for folks to drive by recommend yanking out brave etc
because opentracing was "standard" I expect similar once otel goes into
that mode (hopefully it stays more in the here is an option we think is
good mode). anyway when people aspire to push low level deps it reduces the
instrumentation available in OSS when they do swap out something more
portable.

Similar to instana, Brave has been supporting JRE 1.6 for ages and we
currently compile with JDK 14 and have lambda tests in the test tree also
have JFR etc in certain modules with higher level requirements. what maybe
Fabian is saying is that it is not required to prevent Java 1.6 or 1.7 in
order to use later in tests.. it is just simpler to do it like that.

trask:
updated the rationale doc based on feedback

docs/java-7-rationale.md
```diff
@@ -33,8 +33,37 @@ But bytecode instrumentation tends to be much more sensitive to Java versions th
 we would lose a lot of confidence in the quality of our Java 7 support without being able to run our
 standard tests against it.
 
+Another option would be to run the "code under test" in a separate JVM from our test harness, which
```
adriancole, line 36: the option brave uses is retrolambda + animal sniffer until https://youtrack.jetbrains.com/issue/IDEA-85478

whether your tests execute in JRE 1.7 is less important than having code compile correctly. You can use a different process to smoke test a few paths as it is unlikely to break as long as you lock signatures and compilation levels.

CodingFabian: Instana has a test harness that forks test apps and runs test agents/instrumentations. The test harness uses testcontainers. It can compile and run the test apps in various JVM vendors, versions, languages (java, groovy, scala, kotlin..). There is an API that communicates between the app under test and the harness.

Its a lot of work. So in my opinion an honest answer to the "rationale" is stating the amount of work the project is willing to invest into these kind of compatibility tests. It also would be honest to state that the project does NOT do any kid of compatibility tests.

PR #1074, opened 2020-08-20
More attempts to fix cache misses

iNikem:


build.gradle
```diff
@@ -56,12 +56,6 @@ allprojects {
       downloadSources = false
     }
   }
-
-  tasks.withType(JavaForkOptions).configureEach {
-    maxHeapSize = System.properties["otel.forkedMaxHeapSize"]
-    minHeapSize = System.properties["otel.forkedMinHeapSize"]
-    jvmArgs "-XX:ErrorFile=/tmp/hs_err_pid%p.log"
```
anuraaga, line 63: Nice find!

anuraaga (approved):

build.gradle
iNikem: I only wondering if we may still need that heap config...

iNikem: a, well, let us see what will happen. Experiment FTW!

PR #1082, opened 2020-08-21
Support RESTeasy 4.0 version in unit tests

mateuszrzeszutek:
* Split jaxrs-2.0 into jaxrs-2.0-common and jaxrs-2.0-testing (similar to e.g. play)
* Support newest RESTEasy version in unit tests
* Remove some deprecated `SpanWithScope` usages

This PR can only be merged after https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1083 is solved (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1167 merged) -- before that happens `./gradlew :instrumentation:jaxrs:jaxrs-2.0:jaxrs-2.0-resteasy-3.1:test -PtestLatestDeps=true` will always fail.


iNikem:
> * Support newest RESTEasy version in unit tests

Can you explain this a little more? I haven't noticed 4.0 version in any changed files :)

mateuszrzeszutek:
> Can you explain this a little more? I haven't noticed 4.0 version in any changed files :)

Sure! RESTeasy changed their module naming scheme in 4.0: before it was always named `resteasy-jaxrs`, from version 4.0 the same module was called `resteasy-core`. The code that actually enables testing 4.0 is in the `jaxrs-2.0-resteasy-3.1.gradle` file:

```groovy
  latestDepTestLibrary group: 'org.jboss.resteasy', name: 'resteasy-core', version: '+'
}

if (findProperty('testLatestDeps')) {
  configurations {
    testImplementation.exclude group: 'org.jboss.resteasy', module: 'resteasy-jaxrs'
  }
}
```

where for the `latestDepTest` run `resteasy-jaxrs` is excluded and `resteasy-core` is added.

iNikem (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/instrumentation/auto/jaxrs/v2_0/AbstractRequestContextInstrumentation.java
```diff
@@ -74,39 +73,35 @@ public AbstractRequestContextInstrumentation() {
...
         context.setProperty(JaxRsAnnotationsTracer.ABORT_HANDLED, true);
-        Span parent = BaseTracer.getCurrentServerSpan();
-        Span span = TRACER.getCurrentSpan();
+        Span serverSpan = BaseTracer.getCurrentServerSpan();
+        Span currentSpan = TRACER.getCurrentSpan();
 
-        if (span == null) {
-          span = TRACER.startSpan(resourceClass, method);
-          return new SpanWithScope(span, currentContextWith(span));
+        if (currentSpan == null || currentSpan == serverSpan) {
```
trask, line 84: why checking `currentSpan == serverSpan`?

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.1/jaxrs-2.0-resteasy-3.1.gradle
```diff
@@ -22,7 +22,22 @@ muzzle {
...
+  implementation project(':instrumentation:jaxrs:jaxrs-2.0:jaxrs-2.0-common')
+
+  testImplementation project(':instrumentation:servlet:servlet-3.0')
+  testImplementation project(':instrumentation:jaxrs:jaxrs-2.0:jaxrs-2.0-testing')
+
+  testLibrary(group: 'org.jboss.resteasy', name: 'resteasy-undertow', version: '3.1.0.Final') {
+    exclude group: 'org.jboss.resteasy', module: 'resteasy-client'
+  }
+
+  latestDepTestLibrary group: 'org.jboss.resteasy', name: 'resteasy-core', version: '+'
```
trask, line 36:
```suggestion
  // artifact name changed from 'resteasy-jaxrs' to 'resteasy-core' starting from version 4.0.0
  latestDepTestLibrary group: 'org.jboss.resteasy', name: 'resteasy-core', version: '+'
```

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.1/jaxrs-2.0-resteasy-3.1.gradle
```diff
@@ -22,7 +22,22 @@ muzzle {
...
+}
+
+if (findProperty('testLatestDeps')) {
+  configurations {
+    testImplementation.exclude group: 'org.jboss.resteasy', module: 'resteasy-jaxrs'
```
trask, line 41:
```suggestion
    // artifact name changed from 'resteasy-jaxrs' to 'resteasy-core' starting from version 4.0.0
    testImplementation.exclude group: 'org.jboss.resteasy', module: 'resteasy-jaxrs'
```

trask (commented):
Nice 👍 One small question below.

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/instrumentation/auto/jaxrs/v2_0/AbstractRequestContextInstrumentation.java
mateuszrzeszutek: I should have added a comment here -- will fix it.

I've found out that `DefaultRequestContextInstrumentation` and implementation-specific `RequestContextInstrumentation`s (e.g. `Resteasy31RequestContextInstrumentation`) worked differently when executed in different order.
* In case of default -> impl order: the default instrumentation has always created a span, but without adding the `ABORT_HANDLED` marker, so the impl instrumentation also ran; `span != null` so only the span names were updated.
* In case of impl -> default: `span` was also not null (because the servlet span would be the current span), so no new span was started here; `ABORT_HANDLED` marker was added though. The default instrumentation did not run; in this case we were missing one span.

trask (approved):
Nice 👍

PR #1089, opened 2020-08-24
Merge changes from dd-trace-java 0.58.0, 0.59.0, 0.60.0 (Part 2)

trask:
https://github.com/DataDog/dd-trace-java/releases/tag/v0.58.0
https://github.com/DataDog/dd-trace-java/releases/tag/v0.59.0
https://github.com/DataDog/dd-trace-java/releases/tag/v0.60.0

Thanks @bantonsson @richardstartin @tylerbenson!

All of the original PRs are linked in the associated commit messages.

-----------------

See #1107 for Part 1


auto-api/src/main/java/io/opentelemetry/instrumentation/auto/api/WeakMap.java
```diff
@@ -87,10 +87,14 @@ public static boolean isProviderRegistered() {
   }
 
   class MapAdapter<K, V> implements WeakMap<K, V> {
+    private final Object[] locks = new Object[16];
```
anuraaga, line 90: Doesn't have to block this PR, but do we ever use the `synchronizedMap` fallback? If we're using `WeakConcurrentMap` as a general default, or a Guava concurrent map otherwise, then we could remove the locks and `computeIfAbsent` (only used once in Netty instrumentation, `putIfAbsent` which is present on `WeakConcurrentMap` should work ok for it) and simplify things a lot.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/FixedSizeCache.java
```diff
@@ -0,0 +1,127 @@
...
+  public FixedSizeCache(int capacity) {
+    if (capacity <= 0) {
+      throw new IllegalArgumentException("Cache capacity must be > 0");
+    }
+    if (capacity > MAXIMUM_CAPACITY) {
+      capacity = MAXIMUM_CAPACITY;
+    }
+    // compute a power of two size for the given capacity
+    int n = -1 >>> Integer.numberOfLeadingZeros(capacity - 1);
+    n = (n < 0) ? 1 : (n >= MAXIMUM_CAPACITY) ? MAXIMUM_CAPACITY : n + 1;
```
anuraaga, line 70: This line needs a comment since it's pretty hard to read. Though I'd generally recommend just copying from the leading expert on caches

https://github.com/ben-manes/caffeine/blob/37f6ad303eb4474cd8a644551d528dfda37c5bfc/caffeine/src/main/java/com/github/benmanes/caffeine/SingleConsumerQueue.java#L136

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/Functions.java
```diff
@@ -0,0 +1,169 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.api;
+
+public final class Functions {
```
anuraaga, line 19: I didn't do a thorough review but if this can be package private that'd be good, it's a lot of API to expose for seemingly limited use cases (just string manipulation)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/QualifiedClassNameCache.java
```diff
@@ -0,0 +1,85 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.api;
+
+public class QualifiedClassNameCache {
```
anuraaga, line 19: Package private

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
```diff
@@ -29,19 +32,26 @@
...
-      };
+  private static final QualifiedClassNameCache CLASS_NAMES =
+      new QualifiedClassNameCache(
+          new Function<Class<?>, String>() {
+            @Override
+            public String apply(Class<?> clazz) {
+              String simpleName = clazz.getSimpleName();
+              if (simpleName.isEmpty()) {
+                String name = clazz.getName();
+                int start = name.lastIndexOf('.');
```
anuraaga, line 48: Is this different from `getEnclosingClass().getSimpleName()` (looping up until hit)? If so, I think this has a good chance of beating the cache in cases except for extremely deeply nested anonymous classes.

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/spring/webflux/client/SpringWebfluxHttpClientTracer.java
```diff
@@ -49,6 +54,12 @@ protected URI url(ClientRequest httpRequest) {
 
   @Override
   protected Integer status(ClientResponse httpResponse) {
+    if (null != RAW_STATUS_CODE) {
```
anuraaga, line 57:
```suggestion
    if (RAW_STATUS_CODE != null) {
```

instrumentation/akka-context-propagation-2.5/src/main/java/io/opentelemetry/instrumentation/auto/akkaconcurrent/AkkaForkJoinPoolInstrumentation.java
```diff
@@ -28,27 +28,33 @@
...
   }
 
   @Override
   protected boolean defaultEnabled() {
     return false;
   }
 
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    // This might need to be an extendsClass matcher...
```
anuraaga, line 54: What's this comment referring to? Is this a safe change?

instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AwsSdkClientTracer.java
```diff
@@ -26,20 +26,35 @@
...
 
 public class AwsSdkClientTracer extends HttpClientTracer<Request<?>, Request<?>, Response<?>> {
 
   static final String COMPONENT_NAME = "java-aws-sdk";
 
   public static final AwsSdkClientTracer TRACER = new AwsSdkClientTracer();
 
-  private final Map<String, String> serviceNames = new ConcurrentHashMap<>();
-  private final Map<Class, String> operationNames = new ConcurrentHashMap<>();
+  private final QualifiedClassNameCache cache =
```
anuraaga, line 42: Ok so here's where it's used outside the package. I would stick to the old version, we could have it keyed on a struct of Class / String if it would help to bring down allocations. I don't know if I grokked the classnamecache correctly, but it seems to be a layer of fixedsize caches? Wouldn't expect it to win significantly to warrant this complexity.

instrumentation/finatra-2.9/finatra-2.9.gradle
```diff
@@ -36,5 +42,10 @@ dependencies {
   // Required for older versions of finatra on JDKs >= 11
   testImplementation group: 'com.sun.activation', name: 'javax.activation', version: '1.2.0'
 
-  latestDepTestLibrary group: 'com.twitter', name: 'finatra-http_2.11', version: '20.6.+'
+  // TODO latestDepTestLibrary doesn't work here
```
anuraaga, line 45: To confirm, this is because finatra 20.8.+ requires a new test class?

instrumentation/hystrix-1.4/src/main/java/io/opentelemetry/instrumentation/auto/hystrix/HystrixDecorator.java
```diff
@@ -17,24 +17,80 @@
...
 public class HystrixDecorator extends BaseDecorator {
   public static final HystrixDecorator DECORATE = new HystrixDecorator();
 
+  private final boolean extraTags;
+
+  private HystrixDecorator() {
+    extraTags = Config.get().isHystrixTagsEnabled();
+  }
+
+  private static final FixedSizeCache<ResourceNameCacheKey, String> RESOURCE_NAME_CACHE =
```
anuraaga, line 36: Do we expect enough cardinality to avoid just using a normal `ConcurrentHashMap` without fixed size?

instrumentation/kafka-clients-0.11/src/main/java/io/opentelemetry/instrumentation/auto/kafkaclients/Base64Decoder.java
```diff
@@ -0,0 +1,159 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.auto.kafkaclients;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+// TODO - can be removed when JDK7 is dropped (adapted from JDK8 source)
```
anuraaga, line 22: Hmm - the words `adapted from JDK8 source` seem to indicate a GPL violation. I wouldn't have noticed if we weren't honest about it ;) But now it's unrevertable.

Maybe we can go ahead and make this package Java 8+ now and use the standard lib, or punt on this commit for now.

anuraaga (commented):
Yeah multiple PRs would have been better, heh

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
richardstartin: Nope.

instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AwsSdkClientTracer.java
richardstartin: The motivation to use `ClassValue` is to handle class unloading properly.

instrumentation/hystrix-1.4/src/main/java/io/opentelemetry/instrumentation/auto/hystrix/HystrixDecorator.java
richardstartin: That's a call you need to make, but "expect" and "what will actually" can be two different things. `FixedSizeCache` degrades to churning rather than unbounded growth in degenerate cases. Attributing this class to me is a little unfair too, it was @bantonsson's work.

instrumentation/kafka-clients-0.11/src/main/java/io/opentelemetry/instrumentation/auto/kafkaclients/Base64Decoder.java
richardstartin: I recommend entirely removing this, and the feature. It was implemented to help a customer using Kafka mirrormaker, and there were probably better, infrastructural solutions.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
richardstartin: This produces names like `EnclosingClass$1` and does this to maintain behaviour. This code is not new and was moved here in a refactoring.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/FixedSizeCache.java
richardstartin: Hey @bantonsson, you should have just copied this ;)

javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/test/DefaultInstrumenterTest.groovy
```diff
@@ -93,9 +93,8 @@ class DefaultInstrumenterTest extends AgentSpecification {
 
   def "configure default sys prop as #value"() {
     setup:
-    ConfigUtils.updateConfig {
```
richardstartin, line 96: Is this attributed to me because I merged the PR? This was @tylerbenson's work.

instrumentation/kafka-clients-0.11/src/main/java/io/opentelemetry/instrumentation/auto/kafkaclients/Base64Decoder.java
tylerbenson: Yeah, @trask https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/46682e9b70cfa48b6951300a02ecbe72410eb0dc can be dropped.  I don't think we need it in this project.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -105,6 +105,7 @@ public static ResettableClassFileTransformer installBytebuddyAgent(
         new AgentBuilder.Default()
             .disableClassFormatChanges()
             .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
+            .with(AgentBuilder.RedefinitionStrategy.DiscoveryStrategy.Reiterating.INSTANCE)
```
tylerbenson, line 108: Interesting... why did you add this?

instrumentation-core/spring/spring-webflux-5.0/src/main/java/io/opentelemetry/instrumentation/spring/webflux/client/SpringWebfluxHttpClientTracer.java
tylerbenson: If this bothers you, I suggest adding a findbugs/checkstyle rule for it.

javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/test/DefaultInstrumenterTest.groovy
trask: Oh no, you're right! I cherry-picked the merge commits which is why they're attributed this way. I'll fix these up (we preserve these commits on the dd-trace-java merge PRs as opposed to our default squash merge).

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
trask: Oh yes, I had meant to point this out (nice find 😄), this fixed the guava addListener test that was left commented out in `CassandraClientTest`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/Functions.java
```diff
@@ -0,0 +1,169 @@
...
+
+  // TODO the majority of this can be removed/simplified if/when dropping Java 7
```
iNikem, line 21: Didn't we already agreed on dropping Java 7?

instrumentation/akka-context-propagation-2.5/src/main/java/io/opentelemetry/instrumentation/auto/akkaconcurrent/AkkaForkJoinPoolInstrumentation.java
trask: This class used to extend `AbstractExecutorInstrumentation` whose `typeMatcher()` matches all Executors in `WHITELISTED_EXECUTORS`.

But this instrumentation is specific to `akka.dispatch.forkjoin.ForkJoinPool`, so this is an optimization.

Probably it should use `extendsClass`, since the class it applies to is not final, in case someone subclasses it and overrides the method(s) that we instrument.

In general, I don't think we are super consistent on when we use `named(..)` vs `extends(named(..))`, I opened #1097 to tack this.

instrumentation/finatra-2.9/finatra-2.9.gradle
trask: i'm not sure why `latestDepTestLibrary` doesn't work here, just guessing it has something to do with your TODO markers above(?)

> TODO(anuraaga): Something about library configuration doesn't work well with scala compilation here.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
```diff
@@ -29,19 +32,26 @@
...
 public abstract class BaseDecorator {
 
-  private static final ClassValue<SpanNames> SPAN_NAMES =
-      new ClassValue<SpanNames>() {
-        @Override
-        protected SpanNames computeValue(Class<?> type) {
-          return new SpanNames(getClassName(type));
-        }
-      };
+  private static final QualifiedClassNameCache CLASS_NAMES =
```
anuraaga, line 40: Can we stick with weak map for this?

instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AwsSdkClientTracer.java
anuraaga: Can we stick with weak map if we need to handle class unload specially instead of the complex cache? Though since it's not static, not sure if there would be an unload issue but could be wrong.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/FixedSizeCache.java
```diff
@@ -0,0 +1,127 @@
...
+ * match or an unused slot.
+ *
+ * <p>The cache is thread safe, and assumes that the <code>Creator</code> passed into <code>
+ * computeIfAbsent</code> is idempotent or otherwise you might not get back the value you expect
+ * from a cache lookup.
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+public final class FixedSizeCache<K, V> {
```
anuraaga, line 33: If we can remove this from `BaseDecorator`, hopefully in favor of a weak cache, perhaps we can move this to agent API module to give time to think about exposing it as a public API?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
richardstartin: Why would you use `WeakMap` in preference to `ClassValue`?

iNikem, line 40: Why do we need this cache at all? Is it that expensive to call `class.getSimpleName()`?

richardstartin: Well, it does slightly more than that. No, it's not expensive to call `class.getSimpleName()` - but we do repeatedly concatenate class names with method names to create resource names, and it adds up. We found these strings tend to get promoted when spans linger. It may not be necessary in this fork.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/Functions.java
trask: Fixed comment, and added tracking for this TODO: #1151

auto-api/src/main/java/io/opentelemetry/instrumentation/auto/api/WeakMap.java
richardstartin: I recommend dropping this commit. The change here was made for symmetry with the other class Adapter and doesn’t really make much sense given that this one only wraps the synchronised wrapper. I also expect that we will simplify/refactor this soon, so if you keep syncing changes this would lead to some churn.

instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AwsSdkClientTracer.java
richardstartin: By unloading I’m referring to the classes themselves. ClassValue _is_ a weak map, just one optimised for class keys.

anuraaga: Thanks for the context, glad to have learned a new tool :) I sent https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1153 which I think handles this case well with less complexity

richardstartin: Fair enough, but the same pattern of constructing resource names with basic preprocessing and concatenation is all over the code base, and we'll be moving more of these over to use that class as they come up. IMHO the complexity you refer to comes from wanting to do in JDK7 what would have been simple with a JDK8 baseline.

richardstartin:
I've moved a lot of the caching stuff out of the bootstrap package and we won't put that kind of thing in there again which should make it easier for you to continue to sync changes. Dropping all of the caching related commits from this PR (including all the functions stuff - we'll still be on JDK7 for a while) should mean you can carry on as normal after we merge [this](https://github.com/DataDog/dd-trace-java/pull/1845)

trask:
closing for now, hope to come back to this later

PR #1090, opened 2020-08-24
Remove executor queue offer check

trask:
Since we only wrap lambdas, which can't be cast back to their original type, the executor queue offer check doesn't seem necessary.

Inspired by https://github.com/DataDog/dd-trace-java/pull/1752 and https://github.com/glowroot/glowroot/commit/725e7d0f3fe2c261f1e13fa8d31ccaa3e539e809.

cc: @tylerbenson


iNikem:
So https://github.com/raphw/byte-buddy/issues/558 is not an issue anymore?

anuraaga (commented):
TBH I don't know enough about this to review, though the code looks fine :-)

trask:
> So [raphw/byte-buddy#558](https://github.com/raphw/byte-buddy/issues/558) is not an issue anymore?

Lambda classes aren't generally instrumentable (unfortunately they aren't passed to `ClassFileTransformer`).

ByteBuddy has an interesting `LambdaInstrumentationStrategy` that attempts to work around this issue, but seems that issue you pointed to is why we aren't using it.

Instead, we manually wrap the lambdas, and the wrappers (which are not lambdas themselves) are instrumented.

trask:
Oh yeah, and the offer check I suspect was to verify that the executor doesn't try to re-cast the Runnable/Callable back to a known type.

This is what happened to me in https://github.com/glowroot/glowroot/issues/564.

Initially I wasn't propagating context to lambdas. Then when I found I had to wrap them, I figured I may as well wrap everything. And then I got that issue, and backed off and only wrapped lambdas. Which is what this codebase already does.

tylerbenson:
@devinsba do you want to weigh in here?

trask:
Thanks @tylerbenson @devinsba!

@anuraaga @iNikem I suggest that we go forward with this change, given that there are concerns about the queue offer check anyways https://github.com/DataDog/dd-trace-java/pull/1752#discussion_r469370585.

anuraaga (approved):
LGTM (? :P)

iNikem (approved):

Issue #1083, opened 2020-08-21
Servlet 3.0 instrumentation does not fully cover 3.1 and 4.0


mateuszrzeszutek:
Servlet 3.1 introduced a new method in `ServletOutputStream`: `setWriteListener(WriteListener)`. This method is not implemented in our current servlet 3.0 instrumentation, so when it's called an `AbstractMethodError` is thrown instead:

```
 org.jboss.resteasy.spi.UnhandledException: java.lang.AbstractMethodError: Receiver class io.opentelemetry.instrumentation.auto.servlet.v3_0.CountingHttpServletResponse$CountingServletOutputStream does not define or inherit an implementation of the resolved method 'abstract void setWriteListener(javax.servlet.WriteListener)' of abstract class javax.servlet.ServletOutputStream.
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:381)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:216)
	at org.jboss.resteasy.core.SynchronousDispatcher.asynchronousExceptionDelivery(SynchronousDispatcher.java:571)
...
Caused by: java.lang.AbstractMethodError: Receiver class io.opentelemetry.instrumentation.auto.servlet.v3_0.CountingHttpServletResponse$CountingServletOutputStream does not define or inherit an implementation of the resolved method 'abstract void setWriteListener(javax.servlet.WriteListener)' of abstract class javax.servlet.ServletOutputStream.
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$DeferredOutputStream.queue(HttpServletResponseWrapper.java:204)
	at org.jboss.resteasy.plugins.server.servlet.HttpServletResponseWrapper$DeferredOutputStream.asyncWrite(HttpServletResponseWrapper.java:179)
	at org.jboss.resteasy.util.CommitHeaderAsyncOutputStream.asyncWrite(CommitHeaderAsyncOutputStream.java:94)
	at org.jboss.resteasy.spi.AsyncOutputStream.asyncWrite(AsyncOutputStream.java:23)
...
```

[...]

Issue #890, opened 2020-08-04
Better hardcoded helper class names


trask:
Currently, the `helperClassNames` are written like this:

```
  @Override
  public String[] helperClassNames() {
    return new String[] {
      packageName + ".GoogleHttpClientDecorator",
      packageName + ".RequestState",
      packageName + ".HeadersInjectAdapter"
    };
  }
```

while using `packageName` for prefix seems like a nice idea, it also prevents Intellij refactoring from noticing and updating the helper class names when you rename one of the classes.

I think we should remove the `packageName` prefix and just hard code the fully qualified name so Intellij refactoring will work correctly.

I recommend we wait to implement this until after #828 because @heyams has work in progress that this change would cause conflicts with.

trask:
I did some testing, and it seems that even using the fully qualified class name, you still have to enable "Search in comments and strings" when refactoring:

![image](https://user-images.githubusercontent.com/218610/90195662-eb35a280-dd7e-11ea-9950-270e959d5047.png)

and in this case Intellij only looks for simple class name in comments and strings, so there's no difference between using fully qualified class name here or using `packageName` + ".<simpleName>"

In case someone finds some different behavior and we still want to do this, here's the script I used to do the mass change, before I did some testing and found that it didn't seem to help the refactoring situation:

```
for file in $(git ls-files *.java)
do
  PACKAGE=$(grep ^package $file | sed 's/package //' | sed 's/;//')
  sed -i "s/packageName + \"/\"$PACKAGE/g" $file
done
```

I tested with package renames also, and didn't help.

Not sure if this change makes sense if it doesn't help us with refactoring.

Closing, but feel free to post your experience if different and we can re-open / re-visit this.

PR #1104, opened 2020-08-26
Fix Servlet instrumentation for servlet-api 3.1+

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1083

* Separate Servlet 3.0 and Servlet 3.1 instrumentations
* Separate Jetty 8.0 (depends on Servlet 3.0) and Jetty 9.1 (depends on Servlet 3.1) instrumentations


instrumentation/jetty/jetty-9.1/jetty-9.1.gradle
```diff
@@ -4,22 +4,21 @@ muzzle {
...
-  library group: 'org.eclipse.jetty', name: 'jetty-server', version: '8.0.0.v20110901'
-  implementation project(':instrumentation:servlet:servlet-3.0')
+  implementation project(':instrumentation:servlet:servlet-3:servlet-3-common')
+  implementation project(':instrumentation:servlet:servlet-3:servlet-3.1')
 
-  // Don't want to conflict with jetty from the test server.
-  testImplementation(project(':testing-common')) {
-    exclude group: 'org.eclipse.jetty', module: 'jetty-server'
-  }
+  // 9.1 is the first Jetty version that introduces Servlet 3.1
```
iNikem, line 15: "supports", not "introduces" :)

instrumentation/jetty/jetty-9.1/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty91HandlerAdvice.java
```diff
@@ -0,0 +1,101 @@
...
+import io.opentelemetry.instrumentation.auto.servlet.v3.TagSettingAsyncListener;
+import io.opentelemetry.instrumentation.auto.servlet.v3_1.Servlet31CountingHttpServletResponse;
+import io.opentelemetry.trace.Span;
+import java.lang.reflect.Method;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import net.bytebuddy.asm.Advice;
+
+public class Jetty91HandlerAdvice {
```
iNikem, line 34: This is now identical to `Jetty8HandlerAdvice`, right? Can it be somehow deduplicated?

instrumentation/jetty/jetty-9.1/src/main/java/io/opentelemetry/instrumentation/auto/jetty/package-info.java
```diff
@@ -0,0 +1,16 @@
...
+ * This module provides support for creating server spans for Jetty Handlers.
+ *
+ * <p>It is possible to write web application running on <a
+ * href="https://www.eclipse.org/jetty/">Eclipse Jetty</a> without actually writing any servlets.
+ * Instead one can use <a
+ * href="https://www.eclipse.org/jetty/documentation//current/jetty-handlers.html">Jetty
+ * Handlers</a>.
+ *
+ * <p>As instrumentation points differ between servlet instrumentations and this one, this module
+ * has its own {@code JettyHandlerInstrumentation} and {@code JettyHandlerAdvice}. But this is the
```
iNikem, line 11: `JettyHandlerAdvice` -> `Jetty91HandlerAdvice`

instrumentation/jetty/jetty-testing/src/main/groovy/JettyContinuationHandlerTest.groovy
```diff
@@ -45,7 +45,7 @@ abstract class JettyContinuationHandlerTest extends JettyHandlerTest {
           continuation.resume()
         }
       } else {
-        handleRequest(baseRequest, response)
+        JettyHandlerTest.handleRequest(baseRequest, response)
```
iNikem, line 48: Why this change is needed?

instrumentation/jsp-2.3/jsp-2.3.gradle
```diff
@@ -15,7 +15,9 @@ dependencies {
   compileOnly group: 'javax.servlet.jsp', name: 'javax.servlet.jsp-api', version: '2.3.0'
   compileOnly group: 'javax.servlet', name: 'javax.servlet-api', version: '3.1.0'
 
-  testImplementation project(':instrumentation:servlet:servlet-3.0')
+  testImplementation project(':instrumentation:servlet:servlet-3:servlet-3.0')
+  testImplementation project(':instrumentation:servlet:servlet-3:servlet-3.1')
```
iNikem, line 19: Why both are needed here?

instrumentation/jetty/jetty-9.1/jetty-9.1.gradle
mateuszrzeszutek: Fixed

instrumentation/jetty/jetty-9.1/src/main/java/io/opentelemetry/instrumentation/auto/jetty/package-info.java
mateuszrzeszutek: Missed that - thanks!

instrumentation/jetty/jetty-testing/src/main/groovy/JettyContinuationHandlerTest.groovy
mateuszrzeszutek: It's not -- IntelliJ introduced it on its own when I was moving the file around and I must have missed it. Fixed.

instrumentation/jsp-2.3/jsp-2.3.gradle
mateuszrzeszutek: Tomcat 7.X uses servlet 3.0, Tomcat 9.X uses 4.0 (I think). I need to include both instrumentation so that both `test` and `test -PtestLatestDeps=true` pass.

instrumentation/jetty/jetty-9.1/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty91HandlerAdvice.java
mateuszrzeszutek: It's almost identical -- the only differences are `TRACER` and the response wrapper type. Since it's an advice class everything must be static, no dependency injection, no passing parameters -- I don't really know how to parametrize it. Maybe someone here will have a good idea, I didn't, so I just copied the least code that I absolutely had to to make it work.

By the way, servlet 3.0 vs 3.1 has a very similar situation: the only difference is the response wrapper type.

iNikem:
@mateuszrzeszutek  Can you please rebase this PR?

instrumentation/jetty/jetty-8.0/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty8HandlerInstrumentation.java
```diff
@@ -54,16 +56,19 @@ public JettyHandlerInstrumentation() {
...
       "io.opentelemetry.instrumentation.servlet.HttpServletRequestGetter",
       "io.opentelemetry.instrumentation.servlet.ServletHttpServerTracer",
-      "io.opentelemetry.instrumentation.auto.servlet.v3_0.Servlet3HttpServerTracer",
-      "io.opentelemetry.instrumentation.auto.servlet.v3_0.TagSettingAsyncListener",
-      "io.opentelemetry.instrumentation.auto.servlet.v3_0.CountingHttpServletRequest",
-      "io.opentelemetry.instrumentation.auto.servlet.v3_0.CountingHttpServletResponse",
-      "io.opentelemetry.instrumentation.auto.servlet.v3_0.CountingHttpServletResponse$CountingServletOutputStream",
-      "io.opentelemetry.instrumentation.auto.servlet.v3_0.CountingHttpServletResponse$CountingPrintWriter",
-      packageName + ".JettyHttpServerTracer",
+      commonPackageName + ".CountingHttpServletRequest",
```
anuraaga, line 64: @trask suggested we keep the full names so IDE can refactor them properly. What do you think?

instrumentation/jetty/jetty-9.1/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty91HandlerInstrumentation.java
```diff
@@ -0,0 +1,85 @@
...
+@AutoService(Instrumenter.class)
+public final class Jetty91HandlerInstrumentation extends Instrumenter.Default {
+
+  public Jetty91HandlerInstrumentation() {
+    super("jetty", "jetty-9.1");
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    // Jetty 9.0 uses servlet-api 3.1
```
anuraaga, line 43:
```suggestion
    // Jetty 9.1 uses servlet-api 3.1
```

I think

instrumentation/servlet/servlet-3/servlet-3.0/src/main/java/io/opentelemetry/instrumentation/auto/servlet/v3_0/Servlet30CountingHttpServletResponse.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return outputStream;
+  }
+
+  @Override
+  protected int getOutputStreamCounter() {
+    return outputStream == null ? 0 : outputStream.counter;
+  }
+
+  static class CountingServletOutputStream extends ServletOutputStream {
```
anuraaga, line 49: Can this class be shared in servlet-common? It seems to only depend on servlet (technically only JDK)

anuraaga (commented):
Can you start a section in the README about the differences in instrumentation of the servlet 3.0 versions you've made here?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/master/instrumentation/servlet

We should fill it in with all the differences and reference information (e.g., jetty versions), but for now just 3.x would be nice. Especially since most of the code looks so similar, it's really hard to see what's changed between versions. Ideally, we can cut down on duplication where possible, but if it's difficult, it's ok but I think we need some text to point people at the places that matter.

anuraaga:
Also really sorry to bring this up so late, I reread the issue #1083 - is it possible to just keep one instrumentation but raise the `compileOnly` dependency to include servlet 3.1? I think it's supposed to be safe for our class to define methods that may not be available at runtime if they're not called. An example of where we do this is Lettuce, we don't have special instrumentation for lettuce 5.2 but support its new API.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/lettuce/lettuce-5.1/src/main/java/io/opentelemetry/instrumentation/auto/lettuce/v5_1/OpenTelemetryTracing.java#L61

I don't know if it works here but wanted to point this out.

instrumentation/jetty/jetty-9.1/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty91HandlerInstrumentation.java
mateuszrzeszutek: 👍 , missed that one

instrumentation/servlet/servlet-3/servlet-3.0/src/main/java/io/opentelemetry/instrumentation/auto/servlet/v3_0/Servlet30CountingHttpServletResponse.java
mateuszrzeszutek: Actually this class is the reason why I had to split 3.0 and 3.1 instrumentations 😄 

3.1 introduced two new abstract methods and 1 new type reference (`WriteListener`) so I had to create 2 different `ServletOutputStream` implementations.

mateuszrzeszutek:
> is it possible to just keep one instrumentation but raise the compileOnly dependency to include servlet 3.1? I think it's supposed to be safe for our class to define methods that may not be available at runtime if they're not called. An example of where we do this is Lettuce, we don't have special instrumentation for lettuce 5.2 but support its new API.

Hmm, I'm not sure how it'll work since one of those methods (`setWriteListener`) references a type added in 3.1, but I'll test it -- if the instrumentation works and muzzle check passes then it would be a great way to reduce code duplication.

instrumentation/jetty/jetty-8.0/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty8HandlerInstrumentation.java
mateuszrzeszutek: Good point, I haven't thought about that.

mateuszrzeszutek:
@anuraaga I've prepared another PR based on your suggestion: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1149

I must say that despite it being a tad hacky I like the new approach more -- no need to split modules or copy-paste entire advice classes.

instrumentation/jetty/jetty-8.0/src/main/java/io/opentelemetry/instrumentation/auto/jetty/Jetty8HandlerInstrumentation.java
trask: just fyi: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/890#issuecomment-673754096

PR #1112, opened 2020-08-27
Telemetry auto version

### LINKED ISSUE #404: Add the agent version to the tracer name/version
Currently, the tracers are acquired with a name, but not a version. It would be very useful to have the agent version included as well. This could be used to identify the version of the agent when looking at spans, making doing customer support easier. We wouldn't need to ask the customer for what version of the agent they were running, but just look at the data directly.

### LINKED ISSUE #750: Remove opentelemetry-javaagent.version and use MANIFEST.MF Implementation-Version
See discussion #669 for background.

### LINKED ISSUE #1108: Printing version throws exception
Running with the latest agent throws an exception when it prints its version number; then the agent says it's unknown.

VersionLogger looks for a file named "opentelemetry-javaagent.version" but gradle produces one based on the project name, called javaagent.version.

Which is the best fix? Looking for the file that's created or changing the one that gets generated?

iNikem:
Implements semantic convention for `telemetry.auto.version`.

Reports instrumentation library version.

Closes #1108
Closes #750 
Closes #404


anuraaga (approved):
Nice

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/AgentVersion.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.api;
+
+public class AgentVersion {
```
trask, line 19: since this is used by library instrumentation also, maybe rename to `Version` or maybe `InstrumentationApiVersion`?

iNikem: `Version` is too generic for me. And I am not sure about `Instrumentation*Api*Version`. I see it more like implementation version, the id of the specific artifact.

trask: I'm good with `InstrumentationVersion` (just trying to get `Agent` out of the name)

iNikem: renamed

PR #1120, opened 2020-08-28
Merge exporter auto config into javaagent-tooling.

### LINKED ISSUE #762: TracerCustomizer packaging
io.opentelemetry.auto.bootstrap.spi.TracerCustomizer should be in the contrib.auto package like io.opentelemetry.sdk.contrib.auto.config.SpanExporterFactory so that it can be imported by other projects. I've worked around it by copying the file with the same package name into my own initializer jar.

anuraaga:
Fixes #762 (hopefully)

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/exporter/DefaultExporterConfig.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/exporter/DefaultExporterConfig.java
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.javaagent.tooling;
+package io.opentelemetry.javaagent.tooling.exporter;
 
 import io.opentelemetry.instrumentation.api.config.Config;
 
-public class DefaultExporterConfig implements io.opentelemetry.sdk.extensions.auto.config.Config {
+public class DefaultExporterConfig implements ExporterConfig {
   private final String prefix;
 
   public DefaultExporterConfig(String prefix) {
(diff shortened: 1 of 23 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/exporter/ExporterConfig.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * An interface used to provide configuration information to {@link SpanExporterFactory} and {@link
+ * MetricExporterFactory} implementations.
+ *
+ * <p>This interface is intentionally kept very simple since the underlying implementations may only
+ * have access to very basic configuration mechanisms such as system properties and environment
+ * variables.
+ */
+public interface ExporterConfig {
```
iNikem, line 27: What implements this interface?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/exporter/DefaultExporterConfig.java
```diff
@@ -14,11 +14,11 @@
  * limitations under the License.
  */
 
-package io.opentelemetry.javaagent.tooling;
+package io.opentelemetry.javaagent.tooling.exporter;
 
 import io.opentelemetry.instrumentation.api.config.Config;
 
-public class DefaultExporterConfig implements io.opentelemetry.sdk.extensions.auto.config.Config {
+public class DefaultExporterConfig implements ExporterConfig {
```
anuraaga, line 21: @iNikem Implemented here

PR #1127, opened 2020-08-29
Fix build due to snapshot issue

trask:


gradle/dependencies.gradle
```diff
@@ -9,24 +9,25 @@ configurations.all {
 ext {
   versions = [
     // Check https://oss.jfrog.org/libs-snapshot/io/opentelemetry/ for latest snapshot version.
-    opentelemetry     : '0.8.0-20200826.153459-48',
+    opentelemetry       : '0.8.0-20200826.153459-48',
     // Snapshot versions are often split into two suffixes based on jfrog's whims, best to keep
```
anuraaga, line 13: often split across versions

gradle/instrumentation.gradle
```diff
@@ -40,6 +40,12 @@ afterEvaluate {
       // OpenTelemetry SDK is not needed for compilation, and :opentelemetry-sdk-shaded-for-testing
       // is brought in for tests by project(:testing-common) below
       exclude group: 'io.opentelemetry', module: 'opentelemetry-sdk'
+      // need explicit excludes on transitive dependencies because they are included explicitly
```
anuraaga, line 43: Nice find - I think we should use dependency constraints instead of explicitly depending to make sure we don't forget adding this anywhere

https://docs.gradle.org/current/userguide/dependency_constraints.html#sec:adding-constraints-transitive-deps

anuraaga (approved):
But approve anyways since either approach is fine to fix the build!

iNikem (approved):

Issue #1061, opened 2020-08-19
Add support for redisson instrumentation.


dengliming:
see: [https://github.com/redisson/redisson](https://github.com/redisson/redisson)

This framework provides redis-based functions, especially distributed locks, which are mainly used in our current projects. Recently, I plan to submit a PR to support it. Does it make sense to support it?

iNikem:
If you use the library and want it being monitored, then of course :) Community contributions are welcomed!

PR #1132, opened 2020-08-30
Add support for redisson instrumentation

dengliming:
closes [https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1061](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1061)


trask:
Hey @dengliming, this looks great!

Can you implement `dbConnectionString(CONNECTION)` also?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f738a50323303cb17e198c519a2a31f7e3f4b208/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/DatabaseClientTracer.java#L142-L145

dengliming:
> Hey @dengliming, this looks great!
> 
> Can you implement `dbConnectionString(CONNECTION)` also?
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f738a50323303cb17e198c519a2a31f7e3f4b208/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/DatabaseClientTracer.java#L142-L145

Sure. thanks for quick reply : )

instrumentation/redisson-3.0/redisson-3.0.gradle
```diff
@@ -0,0 +1,23 @@
...
+    module = "redisson"
+    versions = "[3.0.0,)"
+    assertInverse = true
+  }
+}
+
+dependencies {
+  compileOnly group: 'org.redisson', name: 'redisson', version: '3.0.0'
+
+  testLibrary group: 'org.redisson', name: 'redisson', version: '3.0.0'
```
trask, line 20: i think this can be consolidated
```suggestion
  library group: 'org.redisson', name: 'redisson', version: '3.0.0'
```

instrumentation/redisson-3.0/redisson-3.0.gradle
```diff
@@ -0,0 +1,23 @@
...
+  testImplementation group: 'com.github.kstyrc', name: 'embedded-redis', version: '0.6'
+  latestDepTestLibrary group: 'org.redisson', name: 'redisson', version: '3.+'
```
trask, line 22: it looks like `3.+` resolves the same as `+`, so using the new `library` dependencies, you don't need this (it will default to `+`)
```suggestion
```

instrumentation/redisson-3.0/redisson-3.0.gradle
dengliming: Thanks. fixed at lastest commit.

instrumentation/redisson-3.0/redisson-3.0.gradle
dengliming: Thanks. fixed at lastest commit.

dengliming:
@trask Can you help to take a look on ci log. Seems many error logs is about `redisson-2.x,` but I already set minimum version `3.0.0`

instrumentation/redisson-3.0/src/main/java/io/opentelemetry/instrumentation/auto/redisson/RedissonClientTracer.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  protected String normalizeQuery(Object args) {
+    String commandName = "Redis Command";
+    // get command
+    if (args instanceof CommandsData) {
+      List<CommandData<?, ?>> commands = ((CommandsData) args).getCommands();
+      if (commands != null && !commands.isEmpty()) {
+        commandName = commands.get(0).getCommand().getName() + "... [bulk]";
```
anuraaga, line 44: I don't think we do any truncation in other places, like Lettuce or DB instrumentation. It probably is a good idea but we should do it consistently for all the instrumentation, how about storing everything for now?

instrumentation/redisson-3.0/src/test/groovy/RedissonClientTest.groovy
```diff
@@ -0,0 +1,303 @@
...
+            "${SemanticAttributes.NET_PEER_PORT.key()}" port
+            "${SemanticAttributes.DB_STATEMENT.key()}" "GET"
+          }
+        }
+      }
+    }
+  }
+
+  def "test batch command"() {
+    when:
```
anuraaga, line 107: Does it make sense to have a more complete async test with callbacks? I notice the return type is `RFuture` I wonder if our concurrent instrumentation will apply to it.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/lettuce/lettuce-5.1/src/test/groovy/io/opentelemetry/instrumentation/auto/lettuce/v5_1/LettuceAsyncClientTest.groovy#L327

We can focus on async in another PR though.

anuraaga (commented):
Thanks, just small and one big point that can probably be saved for later.

instrumentation/redisson-3.0/src/main/java/io/opentelemetry/instrumentation/auto/redisson/RedissonClientTracer.java
iNikem: Java SDK already supports truncating too large attributes. So this better be left to them :)

instrumentation/redisson-3.0/src/test/groovy/RedissonClientTest.groovy
dengliming: Nothing seems to have changed.

instrumentation/redisson-3.0/src/test/groovy/RedissonAsyncClientTest.groovy
```diff
@@ -0,0 +1,133 @@
...
+    result.whenComplete(new BiConsumer<Boolean, Throwable>() {
+      @Override
+      void accept(Boolean res, Throwable throwable) {
+        RList<String> strings = redisson.getList("list1")
+        strings.add("a")
+      }
+    })
+    then:
+    result.get(3, TimeUnit.SECONDS)
+    assertTraces(2) {
```
anuraaga, line 100: Thanks, this is a good test! I just noticed our async redis tests don't check context propagation, we can look at that separately #1139

anuraaga (approved):
Thanks

PR #1134, opened 2020-08-30
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


anuraaga (commented):
I am personally anti-dependabot, updating dependencies multiple times between releases instead of once before a release is unproductive and spammy (security alerts without automatic PR spam is great of course). Will leave to other maintainers in case they have a different preference :)

trask:
I have not used dependabot, so defer here to @anuraaga (and @iNikem if he has any preference).

anuraaga:
Actually I realized in this repository in particular, we probably can't use dependabot anyways. We have a pattern of using compile only dependencies with old versions to allow instrumentation to work with the user's dependency versions. We have a small set of "real" dependencies but I guess dependabot doesn't support this use case.

iNikem:
In general I am pro-Dependabot, but I also don't see how we can use it in this repo. We certainly cannot automatically upgrade library dependencies of the instrumentations.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
anuraaga, line 6:
```suggestion
```

How about let's just leave it to checking the github actions versions, that seems pretty nice actually.

sullis: > How about let's just leave it to checking the github actions versions, that seems pretty nice actually.

Done.

iNikem (approved):

anuraaga (approved):
Thanks!

PR #1136, opened 2020-08-31
Use a fake OTLP backend for storing smoke tests traces instead of fil…

anuraaga:
…e exporter.

By using the same Java proto for serializing and deserializing traces, no chance of protocol version incompatibility.

Also gives us some more control, e.g., we can clear requests using an endpoint instead of restarting the collector every test.

I don't have bintray creds so I haven't pushed the docker image, but I could verify everything worked by using local docker build. Since smoke tests are already broken, I guess we can merge this and let GitHub build instead of splitting into two PRs but let me know.


smoke-tests/fake-backend/build.gradle
```diff
@@ -0,0 +1,38 @@
...
+    }
+  }
+}
+
+compileJava {
+  options.release = 11
+}
+
+dependencies {
+  implementation("com.linecorp.armeria:armeria-grpc:1.0.0")
```
trask, line 26: 😄 👍

anuraaga:
Interesting question - the collector provides some logging, but I guess that's about it. If @iNikem agrees (including with this PR in general :) ), I'd do a follow up for that want to get the smoke tests green soon.

smoke-tests/fake-backend/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.6-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
iNikem, line 5: Please add checksum as in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/899263612d338a85310eec4e3c54aaac32aef327/gradle/wrapper/gradle-wrapper.properties#L6

iNikem (approved):

smoke-tests/fake-backend/gradle/wrapper/gradle-wrapper.properties
anuraaga: I see none of our smoke tests projects do so far. Do you mind if I go ahead and follow up with a PR to remove the "subrepos" to use our normal Gradle? While working on this I realized

1) It's a bit (not dealbreaker) annoying to start up another intellij to tweak each project
2) It'd be nice to apply our normal gradle scripts for spotless, license, etc
3) Worried about interaction with tooling, such as #1133 

I could either flag out these projects but since the normal build would only run `compileJava` essentially, I think it's ok to actually just have them enabled by default. What do you think?

iNikem: I have moved them outside our usual Gradle structure on purpose. E.g. `play` project contributes a lot of Gradle warnings during the build. I did not want them to pollute all our builds.

anuraaga: Hmm - ok I've added the distribution sums for now. But I think the tradeoffs of the separate projects warrant another look into what can be done about things like the deprecation warnings.

PR #1145, opened 2020-09-01
Set instrumentation groups to unique values when using auto/library/t…

anuraaga:
…esting pattern.

This allows us to keep task / folder names short without running into the Gradle's dedupe bug.

See https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1125#issuecomment-683515183

I realized that until fixing this, the agent is missing some instrumentation like aws-sdk-2.2

### DIFF
--- a/gradle/instrumentation-common.gradle
+++ b/gradle/instrumentation-common.gradle
@@ -74,3 +74,12 @@ if (testLatestDeps) {
     }
   }
 }
+
+if (['auto', 'library', 'testing'].contains(projectDir.name)) {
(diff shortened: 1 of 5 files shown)


iNikem (approved):
Honestly, this seems somewhat strange to me. But I am not ready to dive into it and offer informed alternative :) If this solves some of our immediate problems, then let it be so.

anuraaga:
Agree it's strange, I don't like it that much either :) But so is having a bug in Gradle for years for now I find it best, let's see if we find something better or Gradle fixes it.

PR #1149, opened 2020-09-01
Fix Servlet instrumentation for servlet-api 3.1+ - alternative take

### LINKED ISSUE #1083: Servlet 3.0 instrumentation does not fully cover 3.1 and 4.0

mateuszrzeszutek:
Fixes #1083

Alternative solution to  #1104 -- does not require as much copy-paste but it's a bit more hacky. No need to change any gradle modules as well.


instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/instrumentation/auto/servlet/v3_0/CountingHttpServletResponse.java
```diff
@@ -110,6 +113,30 @@ public void flush() throws IOException {
...
     }
+
+    // New abstract methods introduced in servlet-api 3.1, 3.0 does not have them.
+    // These methods purposefully use reflection to call the delegate object so that they compile
+    // against both 3.0 and 3.1+ servlet-api.
+
+    // @Override
+    public boolean isReady() {
+      try {
+        Method isReady = ServletOutputStream.class.getMethod("isReady");
```
iNikem, line 124: This certainly should be cached, no need to lookup on every method call. Also, probably `MethodHandle` will work faster than `Method`

instrumentation/servlet/servlet-3.0/src/main/java/javax/servlet/WriteListener.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.IOException;
+import java.util.EventListener;
+
+/**
+ * Backport of a callback interface introduced in Servlet 3.1 API. Having it here is required to
+ * make {@code Servlet3Instrumentation} work with Servlet 3.0 API, otherwise muzzle would disable it
+ * because of a missing class.
+ */
+public interface WriteListener extends EventListener {
```
iNikem, line 27: And what will happen if there is the original `WriteListener` on the classpath?

iNikem (commented):
This is certainly much nicer PR than the other one :) 4 classes vs 45 :)

instrumentation/servlet/servlet-3.0/src/main/java/javax/servlet/WriteListener.java
mateuszrzeszutek: First one found will be loaded, I guess. This class does not change in any [3.1, 5) version, so we shouldn't worry about any conflict or missing methods.

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/instrumentation/auto/servlet/v3_0/CountingHttpServletResponse.java
mateuszrzeszutek: Very good point, done. Thanks!

anuraaga:
Agreed :) Judging from the CI it seems muzzle may not be liking this yet but it's probably our job to fix muzzle for this sort of use case. Can't dig too deep today but thanks a lot for bearing with this!

mateuszrzeszutek:
Looks like I was too optimistic about this solution: `java.lang.SecurityException: class "javax.servlet.WriteListener"'s signer information does not match signer information of other classes in the same package` -- it seems that just copying a single class to our project is not viable. (It's a bit strange that muzzle check passed on my laptop, but I guess that it may depend on Java version that's being used)

I have several ideas on how to solve this:
1. Go back to the copy-paste solution;
2. Implement a way to make muzzle ignore a selected mismatch, perhaps annotation on method param?;
3. Create a dynamic proxy in runtime -- since we don't care about these new methods at all. This one could be significantly slower than other solutions;
4. Create a dynamic class extending `ServletOutputStream` when instrumentation is loaded -- this should have better performance than 3 bun can be tricky.

What do you all think? I'd be grateful for any other ideas/comments/feedback.

PR #1150, opened 2020-09-01
Fix redisson Nightly build

dengliming:
fixes [https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1143](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1143)

The lastest Redisson version only support completely scheme address like `redis://` or `rediss://`, try to update the testcase to fit it.

### DIFF
--- a/instrumentation/redisson-3.0/redisson-3.0.gradle
+++ b/instrumentation/redisson-3.0/redisson-3.0.gradle
@@ -15,6 +15,9 @@ muzzle {
 
 dependencies {
   library group: 'org.redisson', name: 'redisson', version: '3.0.0'
-  testImplementation group: 'org.redisson', name: 'redisson', version: '3.4.3'
   testImplementation group: 'com.github.kstyrc', name: 'embedded-redis', version: '0.6'
+}
+
+test {
+  systemProperty "testLatestDeps", testLatestDeps
 }
--- a/instrumentation/redisson-3.0/redisson-3.0.gradle
+++ b/instrumentation/redisson-3.0/redisson-3.0.gradle
@@ -15,5 +15,6 @@ muzzle {
 
 dependencies {
   library group: 'org.redisson', name: 'redisson', version: '3.0.0'
+  testImplementation group: 'org.redisson', name: 'redisson', version: '3.4.3'
   testImplementation group: 'com.github.kstyrc', name: 'embedded-redis', version: '0.6'
 }
--- a/instrumentation/redisson-3.0/src/test/groovy/RedissonAsyncClientTest.groovy
+++ b/instrumentation/redisson-3.0/src/test/groovy/RedissonAsyncClientTest.groovy
@@ -46,7 +46,7 @@ class RedissonAsyncClientTest extends AgentTestRunner {
   String address = "localhost:" + port
 
   def setupSpec() {
-    if ("true".equals(System.properties.getProperty("testLatestDeps"))) {
+    if (Boolean.getBoolean("testLatestDeps")) {
       // Newer versions of redisson require scheme, older versions forbid it
       address = "redis://" + address
     }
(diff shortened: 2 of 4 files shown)


instrumentation/redisson-3.0/redisson-3.0.gradle
```diff
@@ -15,5 +15,6 @@ muzzle {
 
 dependencies {
   library group: 'org.redisson', name: 'redisson', version: '3.0.0'
+  testImplementation group: 'org.redisson', name: 'redisson', version: '3.4.3'
```
anuraaga, line 18: We intentionally use latest dep test to check both old versions and new version. I don't think we need to change the test dependency version for standard build.

anuraaga (commented):
Thanks! Test change looks good but IIUC, we shouldn't need to change the version.

instrumentation/redisson-3.0/redisson-3.0.gradle
iNikem: Yeah, we certainly want to test with the same version that we claim we support :) Meaning 3.0.0 in this case

dengliming: What about config the latestDepTestLibrary like ` latestDepTestLibrary group: 'org.redisson', name: 'redisson', version: '(,3.4.3)'`
because `redis:// ` is a break change.

anuraaga: Oh shoot - so does that mean that redis:// doesn't work at all on the older versions? That's pretty unfortunate. I can't think of a simpler solution but the brute force is tweaking the Gradle build to set a property affecting the URL when testing latest deps. If this sounds too tricky, let's address it as a follow-up.

dengliming: Seems we can set the properties like this. 
```
test {
  if (testLatestDeps) {
    systemProperty "testLatestDeps", true
  }
}
```

instrumentation/redisson-3.0/src/test/groovy/RedissonAsyncClientTest.groovy
```diff
@@ -42,8 +42,13 @@ class RedissonAsyncClientTest extends AgentTestRunner {
     .port(port).build()
   @Shared
   RedissonClient redisson
+  @Shared
+  String address = "localhost:" + port
 
   def setupSpec() {
+    if ("true".equals(System.properties.getProperty("testLatestDeps"))) {
+      address = "redis://" + address
```
anuraaga, line 50:
```suggestion
      // Newer versions of redisson require scheme, older versions forbid it
      address = "redis://" + address
```

instrumentation/redisson-3.0/src/test/groovy/RedissonClientTest.groovy
```diff
@@ -47,8 +47,13 @@ class RedissonClientTest extends AgentTestRunner {
     .port(port).build()
   @Shared
   RedissonClient redisson
+  @Shared
+  String address = "localhost:" + port
 
   def setupSpec() {
+    if ("true".equals(System.properties.getProperty("testLatestDeps"))) {
+      address = "redis://" + address
```
anuraaga, line 55:
```suggestion
      // Newer versions of redisson require scheme, older versions forbid it
      address = "redis://" + address
```

anuraaga (approved):
Thanks!

instrumentation/redisson-3.0/src/test/groovy/RedissonClientTest.groovy
```diff
@@ -47,8 +47,14 @@ class RedissonClientTest extends AgentTestRunner {
     .port(port).build()
   @Shared
   RedissonClient redisson
+  @Shared
+  String address = "localhost:" + port
 
   def setupSpec() {
+    if ("true".equals(System.properties.getProperty("testLatestDeps"))) {
```
iNikem, line 54: You can use `java.lang.Boolean#getBoolean`

iNikem (approved):
Please rebase this PR. We now have working mainline, I hope.

instrumentation/redisson-3.0/src/test/groovy/RedissonClientTest.groovy
dengliming: 👍 thanks.

PR #1153, opened 2020-09-02
Use ClassValue and full name cache in AWS SDK 1.1

anuraaga:
Alternative to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1089/files#r475355242

I find the `QualifiedNameCache` to bee too complex for a public API, especially the joining functions. One of the reasons it needs to be public is its use in AWS SDK, so I've tried this simpler approach

- AWS SDK client cardinality is guaranteed to be low, `ConcurrentHashMap` is fine
- The actual transformation seems much more readable than having to grok the suffix join functionality IMO

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AWSClientInstrumentation.java
+++ b/instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AWSClientInstrumentation.java
@@ -54,6 +54,7 @@ public final class AWSClientInstrumentation extends Instrumenter.Default {
   public String[] helperClassNames() {
     return new String[] {
       packageName + ".AwsSdkClientTracer",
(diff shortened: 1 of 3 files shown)


instrumentation/aws-sdk/aws-sdk-1.11/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v1_11/AWSClientInstrumentation.java
```diff
@@ -54,6 +54,8 @@ public AWSClientInstrumentation() {
   public String[] helperClassNames() {
     return new String[] {
       packageName + ".AwsSdkClientTracer",
+      packageName + ".AwsSdkClientTracer$NamesCache",
+      packageName + ".AwsSdkClientTracer$1",
```
anuraaga, line 58: @trask Not sure what this corresponds to

iNikem (approved):

trask (approved):

PR #1157, opened 2020-09-02
Validate gradle wrapper only when changed

iNikem:
No need to run it on every PR

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
     paths:
       - 'gradle/wrapper/**'
+  push:
+    paths:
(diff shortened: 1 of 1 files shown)


iNikem:
I can add `push` with same path restriction. But I thought we have protected main branch?

jkwatson:
> I can add `push` with same path restriction. But I thought we have protected main branch?

Can't maintainers still push to the main branch?

Adding push seems reasonable, though. Is this a very time consuming check?

iNikem:
Nay, very fast. It just annoys my GitHub inbox :)

anuraaga (approved):
Change LGTM but

> Nay, very fast. It just annoys my GitHub inbox :)

Do you get GitHub notifications for successful workflows too? Seems like a lot of noise :)

trask (approved):

PR #1158, opened 2020-09-02
mapping redis (jedis) connection net.peer attributes

kubawach:
Changes to 
- set net.peer.* semantic attributes for jedis instrumentation.- both 1.4 and 3.0
- utilising existing helper utils 
- making extra effort to resolve address (if hostname, not IP)
- slight refactor of the utils for method delegation

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/utils/NetPeerUtils.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/utils/NetPeerUtils.java
@@ -30,20 +30,34 @@ public final class NetPeerUtils {
     if (remoteConnection != null) {
       InetAddress remoteAddress = remoteConnection.getAddress();
       if (remoteAddress != null) {
-        setNetPeer(span, remoteAddress);
+        setNetPeer(span, remoteAddress, remoteConnection.getPort());
       } else {
         // Failed DNS lookup, the host string is the name.
-        setNetPeer(span, remoteConnection.getHostString(), null);
+        setNetPeer(span, remoteConnection.getHostString(), null, remoteConnection.getPort());
       }
-      span.setAttribute(SemanticAttributes.NET_PEER_PORT.key(), remoteConnection.getPort());
     }
   }
 
-  public static void setNetPeer(Span span, InetAddress remoteAddress) {
-    setNetPeer(span, remoteAddress.getHostName(), remoteAddress.getHostAddress());
+  public static void setNetPeer(Span span, InetAddress remoteAddress, int port) {
+    setNetPeer(span, remoteAddress.getHostName(), remoteAddress.getHostAddress(), port);
+  }
+
+  public static void setNetPeer(Span span, String nameOrIp, int port) {
+    try {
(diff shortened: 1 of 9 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/utils/NetPeerUtils.java
```diff
@@ -26,24 +26,7 @@
...
-      span.setAttribute(SemanticAttributes.NET_PEER_PORT.key(), remoteConnection.getPort());
-    }
-  }
-
-  public static void setNetPeer(Span span, InetAddress remoteAddress) {
-    setNetPeer(span, remoteAddress.getHostName(), remoteAddress.getHostAddress());
-  }
-
-  public static void setNetPeer(Span span, String peerName, String peerIp) {
+  public static void setNetPeer(Span span, String peerName, String peerIp, int port) {
```
iNikem, line 29: Please review the order of method. Before this change it was somewhat logical if you look at parameter types. Now it seems to me as more chaotic.

kubawach: Goal of my refactoring was to make it in order - main method (setting attrs) first, then delegates starting from the simplest one. But I can bring it back as it was, just confirm that you'd like it that way :)

iNikem: I prefer the same order as with constructors: shorter before, longer after. If A delegates to B, then A is before B.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/utils/NetPeerUtils.java
```diff
@@ -26,24 +26,10 @@
...
-      }
-      span.setAttribute(SemanticAttributes.NET_PEER_PORT.key(), remoteConnection.getPort());
-    }
-  }
+  public static void setNetPeer(Span span, String peerName, String peerIp, int port) {
 
-  public static void setNetPeer(Span span, InetAddress remoteAddress) {
-    setNetPeer(span, remoteAddress.getHostName(), remoteAddress.getHostAddress());
-  }
+    System.out.println("NAME: "+peerName+", IP: "+peerIp+", PORT: "+port);
```
iNikem, line 31: Don't forget to remove :)

kubawach: You didn't see anything! ;-)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/utils/NetPeerUtils.java
kubawach: Changed as suggested :)

iNikem (approved):

anuraaga (approved):
Awesome, thanks!

PR #1164, opened 2020-09-03
Use Alex to prevent biased language during PRs

iNikem:
Trying https://github.com/reviewdog/action-alex


iNikem:
Nah, not as useful as I hoped.

trask:
nice idea

PR #1165, opened 2020-09-03
Add a logback appender that can be added to a config to export context IDs

anuraaga:
Heavily inspired by Armeria, even copied a file as-is

https://github.com/line/armeria/blob/master/logback/src/main/java/com/linecorp/armeria/common/logback/RequestContextExportingAppender.java

While MDC integration can cover non-logback / log4j, I think directly integrating with them like this and the log4j module work well. This is why I'm hoping to be able to inject `logback.xml` using auto instrumentation at some point though :)

/cc @trustin


instrumentation/logback/logback-1.0.0/library/NOTICE.txt
```diff
@@ -0,0 +1,19 @@
+This product contains a modified part of Armeria, distributed by LINE:
```
anuraaga, line 1: I would generally copy the license header into the file I copied as well, but spotless deletes the second header if I try. For now I've added this NOTICE but let me know if there's something I should do here.

iNikem:
Should you update some documentation as well? At least README.md or some sort of highly visible link to module specific readme

iNikem (approved):

anuraaga:
I was thinking of adding to top-level README.md but since this isn't supported by agent yet thought it might be confusing. How about I ~~struggle~~ try to set up auto instrumentation in a followup and otherwise will start a dedicated section for manual-only instrumentations we provide.

anuraaga:
@zenmoto FYI I've added this logback instrumentation very similar to the log4j instrumentation. Happy to update if there's any semantic changes :)

PR #1156, opened 2020-09-02
Don't wrap request/response in Jetty

iNikem:
Fixes #1096

Issue #1123, opened 2020-08-28
Muzzle should fail on unimplemented abstract methods


mateuszrzeszutek:
This is an idea for improvement that would have prevented something like https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1083 from happening.
In that issue, Servlet API 3.1 introduced two new abstract methods to `ServletOutputStream` that were unimplemented in our code -- and everything was OK until someone tried to call one of them.

I believe that muzzle should fail for an instrumentation if one of our classes that implements/extends some lib/framework class has missing methods.

Issue #1096, opened 2020-08-25
ClassCastException for v3_0.CountingHttpServletResponse


ishg:
Hello, running the Java agent (v0.7.1) with a Java 1.8 app results in the following error with `jetty.server.Response`: 

```
[2020-08-24 20:13:46,734] [qtp1474515510-53] WARN server.HttpChannel: Could not send response error 500: java.lang.ClassCastException: io.opentelemetry.auto.instrumentation.servlet.v3_0.CountingHttpServletResponse cannot be cast to org.eclipse.jetty.server.Response
[2020-08-24 20:13:48,671] [qtp1474515510-121] WARN server.HttpChannel: /sso/health
java.lang.ClassCastException: io.opentelemetry.auto.instrumentation.servlet.v3_0.CountingHttpServletResponse cannot be cast to org.eclipse.jetty.server.Response
    at org.eclipse.jetty.server.handler.RequestLogHandler.handle(RequestLogHandler.java:104) ~[jetty-server-9.1.2.v20140210.jar:9.1.2.v20140210]
    at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1114) ~[jetty-server-9.1.2.v20140210.jar:9.1.2.v20140210]
    at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:479) ~[jetty-servlet-9.1.2.v20140210.jar:9.1.2.v20140210]
    at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:183) ~[jetty-server-9.1.2.v20140210.jar:9.1.2.v20140210]
...
...
[REDACTED]
```

I can provide the relevant Pom files after scrubbing if you'd like to know versions for specific libraries, but Jetty seems to be `v9.1.2` which should be supported according to the list of frameworks in the main [README](https://github.com/open-telemetry/opentelemetry-java-instrumentation#supported-java-libraries-and-frameworks).

Is there any guidance on how best to debug this or does this seem like a potential bug in the intstrumentation ? 

**To Note:** This DOES break the application due to the ClassCastException and is not a graceful error (even though in the logs it says WARN)

iNikem:
Hello @ishg, thank you for this report. I believe that this information is enough for us to start our investigation. Sorry for this inconvenience.

ishg:
[...]

PR #1167, opened 2020-09-03
Fix Servlet instrumentation for servlet-api 3.1+ - remove output stream wrapper

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1083

The third time's the charm, I hope. This PR contains removal of the `ServletoutputStream` wrapper according to [this comment](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1083#issuecomment-685679223).

This might be not so bad anyway, because according to `ServletResponse#getOutputStream()` JavaDoc:
```
     * Either this method or {@link #getWriter} may 
     * be called to write the body, not both.
```
so only those implementations that use `getOutputStream()` will get inaccurate content sizes (always 0), while those using `getWriter()` will be perfectly fine.


instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
```diff
@@ -122,7 +122,7 @@ class DropwizardTest extends HttpServerTest<DropwizardTestSupport> {
         "${SemanticAttributes.HTTP_USER_AGENT.key()}" TEST_USER_AGENT
         "${SemanticAttributes.HTTP_CLIENT_IP.key()}" TEST_CLIENT_IP
         // exception bodies are not yet recorded
-        "${SemanticAttributes.HTTP_RESPONSE_CONTENT_LENGTH.key()}" { it == responseContentLength || /* async */it == null }
+        "${SemanticAttributes.HTTP_RESPONSE_CONTENT_LENGTH.key()}" { it == 0 || it == responseContentLength || /* async */it == null }
```
FrankSpitulski, line 125: allowing 0 indicates that the response was not counted. if it's not counted in the tests I wouldn't expect it to be counted elsewhere.

mateuszrzeszutek: Well, it's to be expected that all frameworks that use `getOutputStream()` will now get a 0.

FrankSpitulski: This means it misses spring boot's default configuration which I'd hazard to say is a large portion of applications being instrumented. Is there some alternative implementation or is this attribute just being dropped?

FrankSpitulski:
Can you add a unit test that succeeds now but if the wrapped response were to be used would fail so that there will be no regressions?

instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
mateuszrzeszutek: There were two:
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1104 contained a separate instrumentation for servlet 3.0 and 3.1; it had lots of copy-pasted code
* In https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1149 I initially tried to just copy the `WriteListener` class that was absent in 3.0, but that failed. Then I tried to generate a class extending `ServletOutputStream` in runtime, which probably should work -- but there's no way to load it correctly without modifying `Instrumenter.Default` heavily.

mateuszrzeszutek:
> Can you add a unit test that succeeds now but if the wrapped response were to be used would fail so that there will be no regressions?

I can do a simple `instanceof` check -- the same way as it was handled in `JettyHandlerTest` for a [similar issue](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1156).
Calling either `ServletOutputStream#isReady()` or ` ServletOutputStream#setWriteListener()` would be ideal, but unfortunately I cannot do that without bumping servlet api version.

instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
anuraaga: @FrankSpitulski We're leaning towards just using `content-length` header for servlet. The counting approach seems to have too many corner cases for us to deal with well unfortunately 😿 

@mateuszrzeszutek I don't think we should set `0` if we couldn't count, `0` means no content, not content couldn't be counted - can you update the implementation of `getContentLength` to avoid setting `0` when `printWriter` is `null`?

iNikem:
Taking into account https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1096#issuecomment-685450841 and https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1083#issuecomment-685679223 I propose to completely remove generic response counting support from servlet instrumentation. I feel bad about this, but if we are left with a partial solution (as @FrankSpitulski noticed above "it misses spring boot's default configuration"), then added complexity is not worth it. History has shown, that this is not as easy as we would like :(

Unless somebody has a good proposal how to solve this otherwise, I ask @mateuszrzeszutek to remove `CountingHttpServletRequest`, `CountingHttpServletResponse` and all related functionality for the time being.

mateuszrzeszutek:
OK, I've removed all `Counting*` classes.

anuraaga (approved):
Thanks @mateuszrzeszutek!

@FrankSpitulski Really sorry about this >< Thanks a lot for the help and giving us the opportunity to think through some of the problems regards to instrumentation here, we may not have noticed that Muzzle needs additional checks related to base-class matching without this exercise so will hopefully get better.

iNikem (approved):

instrumentation/dropwizard-testing/src/test/groovy/DropwizardTest.groovy
mateuszrzeszutek: I've removed all the counting stuff altogether, according to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1167#issuecomment-689345608

mateuszrzeszutek:
> we may not have noticed that Muzzle needs additional checks related to base-class matching without this exercise so will hopefully get better.

There's an issue for that now: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1123 . I have an idea how to do implement that, so hopefully this will get done soon and we'll avoid similar issues in the future.

PR #1168, opened 2020-09-03
Properly track causality of kotlin suspend functions

johnbley:
Fix propagation/causality when kotlin suspend functions are used, by adding instrumentation specific to the kotlin coroutine system.


instrumentation/java-concurrent/kotlin-testing/src/test/kotlin/KotlinCoroutineTests.kt
```diff
@@ -138,6 +140,55 @@ class KotlinCoroutineTests(private val dispatcher: CoroutineDispatcher) {
...
+      GlobalScope.launch {
+        a(i.toLong())
+      }
+      GlobalScope.launch {
+        b(i.toLong())
+      }
+    }
+    // A simple and generous timeout (as opposed to joining on all the generated Jobs) also helps
+    // fail the test if our overhead throws the cost of context switching 100s of times way out of normal
+    Thread.sleep(4000)
```
iNikem, line 154: You may want to use `runBlocking`: https://kotlinlang.org/docs/reference/coroutines/basics.html#structured-concurrency

johnbley: Yes, thanks!

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+import kotlin.jvm.functions.Function2;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@AutoService(Instrumenter.class)
+public class KotlinProbeInstrumentation extends Instrumenter.Default {
```
anuraaga, line 38: I think it'd make sense for us to rename `kotlin-testing` to `kotlin-coroutines` and add this instrumentation there too since this isn't instrumenting the JDK itself.

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+      return contextWrapper;
+    }
+
+    @Override
+    public void resumeWith(@NotNull Object o) {
+      proxy.resumeWith(o);
+    }
+  }
+
+  public static class CoroutineContextWrapper implements CoroutineContext {
```
anuraaga, line 151: Is it possible to use `ThreadContextElement` instead of hooking into the lifecycle ourselves? If not, can we add some docs on why that doesn't work? I haven't used kotlin much TBH, but my understanding is that's the standard way of syncing a threadlocal with coroutines, which is what we want to achieve here.

https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-thread-context-element/

anuraaga: Also, just brainstorming, but is there an approach that wraps dispatchers using something like https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html#currentContextExecutor-java.util.concurrent.Executor- ? I have no idea the pros/cons for kotlin :) But in Java, my goto for context propagation is usually a wrapped `Executor` so it just came to mind. I've used [this approach](https://github.com/line/armeria/blob/master/examples/context-propagation/kotlin/src/main/kotlin/example/armeria/contextpropagation/kotlin/MainService.kt#L43) in some of the only kotlin I've ever written outside of build scripts.

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+  overhead should be fine for production usage.
+  */
+
+  public KotlinProbeInstrumentation() {
+    super("kotlin-coroutines");
+  }
+
+  @Override
+  public ElementMatcher<? super TypeDescription> typeMatcher() {
+    return named("kotlin.coroutines.jvm.internal.DebugProbesKt");
```
anuraaga, line 69: Just to confirm, I guess there's no public entrypoint that is similar to this one? For example, we won't have any breakage across kotlin version if we can stick to public APIs than internal ones. One API that comes to mind is `startCoroutine`

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines/start-coroutine.html

Perhaps my other suggestions can bring to light an idea that can keep interception to the public API. But if not, it's ok too we have other instrumentation reliant on private APIs and use muzzle to bite us when something breaks.

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+@AutoService(Instrumenter.class)
+public class KotlinProbeInstrumentation extends Instrumenter.Default {
+  /*
+  Kotlin coroutines with suspend functions are a form of cooperative "userland" threading
+  (you might also know this pattern as "fibers" or "green threading", where the OS/kernel-level thread
+  has no idea of switching between tasks.  Fortunately kotlin exposes hooks for the key events: knowing when
+  coroutines are being created, when they are suspended (swapped out/inactive), and when they are resumed (about to
+  run again).
+
+  Without this instrumentation, heavy concurrency and usage of kotlin suspend functions will break causality
```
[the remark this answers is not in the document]
johnbley: Will do.

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,196 @@
...
+  1) Use the DebugProbes callbacks to learn about coroutine create, resume, and suspend operations
+  2) Wrap the creation Coroutine and its Context and use that wrapping to add an extra Context "key"
+  3) Use the callback for resume and suspend to manipulate our context "key" whereby an appropriate state
+     object can be found (tied to the chain of Continutations in the Coroutine).
+  4) Do our swapping-context dance with that appropriate state
+  5) Test with highly concurrent well-known span causality and ensure everything looks right.
+     Without this instrumentation, this test fails with concurrency=2; with this instrumentation,
+     it passes with concurrency=200.
+
+  I've run the basic shape of this by an expert in kotlin coroutines (Roman Elizarov) and he agreed that the
```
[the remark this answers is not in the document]
johnbley: 👍

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
johnbley: Are you proposing to move this instrumentation and test to a new <root>/instrumentation/kotlin-coroutines module?  That makes sense to me.  Please confirm my understanding and I'll work on that refactoring.

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
johnbley: So, yes, I have experimented with more "public" ways of doing this.  I had a version with the `ThreadContextElement` approach and it passed the tests, but it exposes the attached CoroutineContext in a way that is visible (and modifiable) to the app (in addition to being a bit more awkward to implement owing to the larger implementation surface of `CoroutineContext.Element` and writing this in Java rather than Kotlin).  I got worried about app code clearing or resetting context and accidentally dropping our bits.  I am not a Kotlin expert at all though, so perhaps there's a cleaner/more elegant way to use the `ThreadContextElement` approach that I'm not seeing.

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
johnbley: You raise some good points.  Yes, we're helped by muzzle here.  We're also helped in that kotlin has a debugging extension using this code (hence the name `DebugProbes`) for tracking coroutine causality, and that their implementation also uses bytecode instrumentation to swap in the debugging version.  Because of the friction introduced by that design, it's unlikely that they'll change this interface frequently 😄

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
anuraaga: Yup that's what I'm thinking

johnbley: OK, I'll work on that today.  Thanks!

instrumentation/java-concurrent/src/main/java/io/opentelemetry/instrumentation/auto/javaconcurrent/KotlinProbeInstrumentation.java
anuraaga: FWIW, if it helps, I don't see a reason to avoid writing the instrumentation in kotlin for kotlin if it helps - main overhead of kotlin is usually the size of the runtime which is no an issue when the instrumentation already requires it.

But given it seems we can expect relative stability from the debug probes API, I wouldn't worry too much easier it's a point in case anything was obvious. We can always investigate more if something breaks in the future instead of now.

instrumentation/kotlin-coroutines/kotlin-testing/kotlin-testing.gradle
```diff
@@ -3,6 +3,8 @@ apply from: "$rootDir/gradle/instrumentation.gradle"
 apply from: "$rootDir/gradle/test-with-kotlin.gradle"
```
anuraaga, line 3: Do we need a separate testing project? Can we just merge this into kotlin-coroutines since these are tests for that instrumentation?

instrumentation/kotlin-coroutines/kotlin-coroutines.gradle
```diff
@@ -0,0 +1,21 @@
...
+  minJavaVersionForTests = JavaVersion.VERSION_1_8
+}
+
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    extraDependency('org.jetbrains.kotlin:kotlin-stdlib-common:1.3.72')
+    extraDependency('org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.7')
+    coreJdk()
```
anuraaga, line 13: Since we're not instrumenting the JDK we shouldn't need this. We want muzzle to run against kotlin-coroutines versions so that also should be the target rather than an extra dependency. The normal muzzle pattern looks something like this, which is verifying the implementation applies to the range of versions of the library (we're interested in kotlin-coroutine-core).

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/jetty-8.0/jetty-8.0.gradle#L5

johnbley: I took a first pass at this; I'd appreciate feedback on how far back to go in the testing - I set it as `[1.0.0,)` since that's what made sense to me.

instrumentation/kotlin-coroutines/kotlin-testing/kotlin-testing.gradle
johnbley: Good call, this was just an artifact of how things were when this lived in java-concurrent.  Fixed and simplified.

instrumentation/kotlin-coroutines/kotlin-coroutines.gradle
johnbley: Looking into muzzle failures here it appears that recent versions of kotlin-core offer a large number of variants.  I can't figure out how to get variant selection to work with muzzle; I've tried adding attributes to the `DefaultArtifact`'s `Properties`-based constructor, but that didn't seem to help.  To unblock the build I'm adding a range limit that doesn't include latest (1.3.9) and will circle back to this later.

anuraaga (approved):
Thanks!

instrumentation/kotlin-coroutines/kotlin-coroutines.gradle
```diff
@@ -0,0 +1,25 @@
...
+  minJavaVersionForTests = JavaVersion.VERSION_1_8
+}
+
+apply from: "$rootDir/gradle/instrumentation.gradle"
+apply from: "$rootDir/gradle/test-with-kotlin.gradle"
+muzzle {
+  pass {
+    group = 'org.jetbrains.kotlinx'
+    module = 'kotlinx-coroutines-core'
+    versions = "[1.0.0,1.3.8)"
```
iNikem, line 13: 1.4 is not supported?

instrumentation/kotlin-coroutines/src/main/java/io/opentelemetry/instrumentation/auto/kotlincoroutines/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,193 @@
...
+    @Override
+    public void resumeWith(@NotNull Object o) {
+      proxy.resumeWith(o);
+    }
+  }
+
+  public static class CoroutineContextWrapper implements CoroutineContext {
+    private final CoroutineContext proxy;
+    private Context myContext;
+    private Context prevContext;
```
iNikem, line 153: As there are a lot of contexts flying around here, I think it would decrease confusion, if these names somehow referenced "tracing" or similar context. To distinguish them from coroutine context.

instrumentation/kotlin-coroutines/src/main/java/io/opentelemetry/instrumentation/auto/kotlincoroutines/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,193 @@
...
+        prevContext = myContext.attach();
+      }
+      return proxy.get(key);
+    }
+
+    @NotNull
+    @Override
+    public CoroutineContext minusKey(@NotNull Key<?> key) {
+      if (key == TraceScopeKey.INSTANCE) {
+        myContext = prevContext.attach();
```
iNikem, line 178: Reading javadoc for `io.grpc.Context#attach`, I think you better doing `myContext.detach(prevContext)` here. Otherwise you just piling contexts on top of each other...

instrumentation/kotlin-coroutines/src/main/java/io/opentelemetry/instrumentation/auto/kotlincoroutines/KotlinProbeInstrumentation.java
```diff
@@ -0,0 +1,193 @@
...
+    @Override
+    public <R> R fold(R r, @NotNull Function2<? super R, ? super Element, ? extends R> function2) {
+      return proxy.fold(r, function2);
+    }
+
+    @Nullable
+    @Override
+    public <E extends Element> E get(@NotNull Key<E> key) {
+      if (key == TraceScopeKey.INSTANCE) {
+        prevContext = myContext.attach();
```
iNikem, line 169: Are you sure that `get` with side effect is a good idea? `CoroutineResumedAdvice.enter` looks strange as well, when you `get` a value and then ignore it. May be you can extract context manipulation from `get` and `minusKey` and do that directly in advice methods?

instrumentation/kotlin-coroutines/src/test/kotlin/KotlinCoroutineTests.kt
```diff
@@ -138,6 +140,54 @@ class KotlinCoroutineTests(private val dispatcher: CoroutineDispatcher) {
...
+    delay(10)
+    a2(iter)
+    scope.close()
+    span.end()
+  }
+  suspend fun a2(iter: Long) {
+    var span = tracer.spanBuilder("a2").startSpan()
+    span.setAttribute("iter", iter)
+    var scope = currentContextWith(span)
+    delay(10)
```
iNikem, line 169: Why do you need scope and delay here?

johnbley: The delay is what forces the coroutine concurrency to be "pessimal" from a tracing perspective, switching frequently on the same Java thread among multiple coroutines and/or passing coroutines to other threads.  The scopes around delay are also there to stress test the system - before this instrumentation, we could easily see nonsensical traces like a2->b.

instrumentation/kotlin-coroutines/src/main/java/io/opentelemetry/instrumentation/auto/kotlincoroutines/KotlinProbeInstrumentation.java
johnbley: Good call; I found a cleaner way to do this (get() the context holder as its own value and call `tracingSuspend/tracingResume`) in a way that keeps the kotlin types happy.

instrumentation/kotlin-coroutines/kotlin-coroutines.gradle
johnbley: I can't get muzzle to work properly with variants:
```
Execution failed for task ':instrumentation:kotlin-coroutines:muzzle-AssertPass-org.jetbrains.kotlinx-kotlinx-coroutines-core-1.3.9'.
> Could not resolve all files for configuration ':instrumentation:kotlin-coroutines:muzzle-AssertPass-org.jetbrains.kotlinx-kotlinx-coroutines-core-1.3.9'.
   > Could not resolve org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9.
     Required by:
         project :instrumentation:kotlin-coroutines
      > Cannot choose between the following variants of org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.9:
```
I've tried diving into the muzzle plugin and playing with setting various properties on the `DefaultArtifact` constructor to have it select the right one, but nothing has worked so far...

iNikem (approved):

trask:
❤️

PR #1172, opened 2020-09-04
Support injecting resources into classloader and use it in aws-sdk-2.…

anuraaga:
…2 instrumentation.

This adds support for instrumentation to "inject" resource files into classloaders. It does this by intercepting the `getResource` method on `ClassLoader` and have it read from the injected files, which are stored into the bootstrap classloader when applying instrumentation.


anuraaga:
Guess muzzle issue just required a clean :D

instrumentation/java-classloader/src/main/java/io/opentelemetry/instrumentation/auto/javaclassloader/ResourceInjectionInstrumentation.java
```diff
@@ -0,0 +1,150 @@
...
+        return;
+      }
+
+      if (!resources.hasMoreElements()) {
+        resources = Collections.enumeration(Collections.singleton(helper));
+        return;
+      }
+
+      // TODO(anuraaga): Don't inline BytesUtil.toByteArray after understanding AccessError when
+      // trying to use the class from bootstrap classloader.
```
trask, line 110: `BytesUtil` class and its `toByteArray` method both need to be `public` b/c the advice is inlined, so it's really in a different package

instrumentation/aws-sdk/aws-sdk-2.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v2_2/AwsClientInstrumentation.java
```diff
@@ -17,72 +17,46 @@
...
+ * Injects resource file with reference to our {@link TracingExecutionInterceptor} to allow SDK's
+ * service loading mechanism to pick it up.
+ */
 @AutoService(Instrumenter.class)
-public final class AwsClientInstrumentation extends AbstractAwsClientInstrumentation {
+public class AwsClientInstrumentation extends AbstractAwsClientInstrumentation {
+  @Override
+  public String[] helperResourceNames() {
+    return new String[] {
+      "software/amazon/awssdk/global/handlers/execution.interceptors",
```
anuraaga, line 39: Without a smoke test, I'm actually a little worried this won't get shaded

iNikem (approved):

iNikem:
Probably needs some documentation as well, so that developers of new instrumentation know about this option.

instrumentation/aws-sdk/aws-sdk-2.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v2_2/AwsClientInstrumentation.java
trask: I'm not following, what do you want shaded here?

anuraaga: Sorry nevermind :)

anuraaga:
@iNikem Hmm - agree that more documentation is always better than less. But we don't have any docs right now for even normal instrumentation I think - I don't want to start documentation with a corner case :)

instrumentation/aws-sdk/aws-sdk-2.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/awssdk/v2_2/AwsClientInstrumentation.java
trask: 😅

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/HelperResources.java
```diff
@@ -0,0 +1,55 @@
...
+import io.opentelemetry.instrumentation.auto.api.WeakMap;
+import java.net.URL;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A holder of resources needed by instrumentation. We store them in the bootstrap classloader so
+ * instrumentation can store from the agent classloader and apps can retrieve from the app
+ * classloader.
+ */
```
trask, line 30: 👍

instrumentation/java-classloader/src/main/java/io/opentelemetry/instrumentation/auto/javaclassloader/ResourceInjectionInstrumentation.java
```diff
@@ -0,0 +1,113 @@
...
+        @Advice.Argument(0) String name,
+        @Advice.Return(readOnly = false) URL resourceUrl) {
+      if (resourceUrl != null) {
+        // Give their classloader precedence.
+        return;
+      }
+
+      resourceUrl = HelperResources.load(classLoader, name);
+    }
+  }
```
trask, line 80: do we have a use case for this? this doesn't seem very auto-instrumentation-y to me (backing off if app already has a particular resource present)

anuraaga: Hmm - don't think we have a use case. `getResource` won't help much in practice anyways since we're doing this mostly to interact with service loading mechanisms. It's here mostly for symmetry - on the flip side, lacking a use case, I lean towards being less intrusive into the app but what do you think?

trask: Without a use case, I would suggest removing instrumentation of `getResource`

iNikem: Now I am confused: how this resource injection works without this instrumentation?

anuraaga: Service loading mechanisms call `getResources`, not `getResource`, since they need to find all the classes of an interface usually. So `getResource` isn't really used in practice. But @trask does this potential confusion make the case that we should stick to symmetry here?

iNikem: That should definitely be documented :)

trask: I'm not sure if that's the right implementation for `getResource` (singular) instrumentation, e.g. we may want to merge contents in some way, but it's hard to know without a use case, so would suggest we remove it until we need it

anuraaga: Makes sense - deleted

trask (approved):

anuraaga:
Hmm having more flakiness on github actions than before I think. Will go ahead and merge this since 8 unflaked and it's our most representative build

PR #1173, opened 2020-09-04
Fix groups / base name for javaagent artifacts

anuraaga:
From what I understand, the intention was

- All javaagent artifacts in group `io.opentelemetry.instrumentation.auto` (our default, so no need to override group in their build files)

- All archives that start with javaagent, including our tooling and exporter modules, should start with `opentelemetry-javaagent` artifact name

While the timing of our snapshots getting moved to `io.opentelemetry` group corresponds to #1145 I can't find what I did there to affect this 🤷

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -10,7 +10,8 @@ apply from: "$rootDir/gradle/codenarc.gradle"
 apply from: "$rootDir/gradle/spotbugs.gradle"
 
 afterEvaluate {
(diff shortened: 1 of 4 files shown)


iNikem:
Can all current rules about group and artefact ids be explained in a couple of sentences? If not, I am afraid they are unnecessary complicated.

iNikem:
Ok, fair enough. Does that mean `io.opentelemetry.instrumentation.auto:opentelemetry-javaagent-exporters-otlp`?

anuraaga:
Yeah right now it's `opentelemetry-javaagent-exporter-otlp` (not plural)

iNikem:
Sorry, @anuraaga for my silence :( I will get to it today.

PR #1177, opened 2020-09-07
Fix a bug when FieldBackedProvider uses a map and value is set to null

mateuszrzeszutek:
This fixes a small bug that caused `contextStore.put(key, null);` calls to throw NPE when a map-backed context store implementation was used.


anuraaga (approved):
Thanks!

iNikem (approved):

Issue #983, opened 2020-08-14
Update message queue (PRODUCER/CONSUMER) span names


trask:
As pointed out in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/963#issuecomment-673871290, at least for JMS, we are not currently creating span name according to spec
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#span-name, and we should review / update other message queue instrumentation as needed also (Kafka, RabbitMQ)

PR #1179, opened 2020-09-07
add operation name to jms span name

malafeev:
part of #983


instrumentation/jms-1.1/src/main/java/io/opentelemetry/instrumentation/auto/jms/JMSDecorator.java
```diff
@@ -38,20 +38,20 @@
   public static final Tracer TRACER = OpenTelemetry.getTracer("io.opentelemetry.auto.jms-1.1");
 
   public String spanNameForReceive(Message message) {
```
iNikem, line 40: Do we need both this method and `spanNameForConsumer` below?

iNikem (approved):

instrumentation/jms-1.1/src/main/java/io/opentelemetry/instrumentation/auto/jms/JMSDecorator.java
malafeev: removed

anuraaga (approved):
Thanks!

PR #1182, opened 2020-09-09
Move AgentClassLoader out of public API.

anuraaga:
Currently we have agent classloader, an implementation detail of our agent, in the auto instrumentation public API. This moves it into the bootstrap. I think that the reason to not have done this before was the usage of `Pair` - presumably `Pair` is too small of a class to have in our API anyways, and for now it was easy to replace its usage in the classloader.

~~Now since bootstrap doesn't depend on the auto-api, there's a lot less surface on the bootstrap classloader to boot (XD).~~


anuraaga:
Realized that without quite a bit of refactoring, at least, auto-api still needs to be in bootstrap

trask (approved):
Nice 🎉

iNikem (approved):

PR #1121, opened 2020-08-28
JDK 11 HttpClient instrumentation

malafeev:

PR #1183, opened 2020-09-09
Use platformClassLoader as muzzle parent instead of bootstrap.

anuraaga:
Java 9+ modules are in the platform classloader, not bootstrap. As we only support building on Java 11, there is no need to have a check here I just always use platform classloader.

I confirmed this fixes muzzle on #1121


iNikem (approved):

iNikem:
Thank you!

PR #1184, opened 2020-09-09
Implement some support for JAX-RS 2.1 additions

mateuszrzeszutek:
* HTTP method `PATCH`
* Async resource methods returning a `CompletionStage`

### DIFF
--- a/instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/instrumentation/auto/jaxrs/v2_0/JaxRsAnnotationsInstrumentation.java
+++ b/instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/instrumentation/auto/jaxrs/v2_0/JaxRsAnnotationsInstrumentation.java
@@ -36,18 +36,17 @@ import io.opentelemetry.javaagent.tooling.Instrumenter;
 import io.opentelemetry.trace.Span;
 import java.lang.reflect.Method;
 import java.util.Map;
+import java.util.concurrent.CompletionStage;
 import javax.ws.rs.Path;
 import javax.ws.rs.container.AsyncResponse;
 import net.bytebuddy.asm.Advice;
 import net.bytebuddy.description.method.MethodDescription;
 import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.implementation.bytecode.assign.Assigner.Typing;
 import net.bytebuddy.matcher.ElementMatcher;
 
 @AutoService(Instrumenter.class)
 public final class JaxRsAnnotationsInstrumentation extends Instrumenter.Default {
-
-  private static final String JAX_ENDPOINT_OPERATION_NAME = "jax-rs.request";
-
   public JaxRsAnnotationsInstrumentation() {
     super("jax-rs", "jaxrs", "jax-rs-annotations");
   }
@@ -76,6 +75,7 @@ public final class JaxRsAnnotationsInstrumentation extends Instrumenter.Default 
       "io.opentelemetry.javaagent.tooling.ClassHierarchyIterable",
       "io.opentelemetry.javaagent.tooling.ClassHierarchyIterable$ClassIterator",
       packageName + ".JaxRsAnnotationsTracer",
+      packageName + ".CompletionStageFinishCallback"
     };
   }
 
@@ -92,6 +92,7 @@ public final class JaxRsAnnotationsInstrumentation extends Instrumenter.Default 
                             "javax.ws.rs.GET",
                             "javax.ws.rs.HEAD",
                             "javax.ws.rs.OPTIONS",
+                            "javax.ws.rs.PATCH",
                             "javax.ws.rs.POST",
                             "javax.ws.rs.PUT")))),
         JaxRsAnnotationsInstrumentation.class.getName() + "$JaxRsAnnotationsAdvice");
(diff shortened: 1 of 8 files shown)


instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/instrumentation/auto/jaxrs/v2_0/JaxRsAnnotationsInstrumentation.java
```diff
@@ -155,15 +157,23 @@ public static void stopSpan(
...
+          returnValue instanceof CompletionStage ? (CompletionStage<?>) returnValue : null;
+
       if (asyncResponse != null && !asyncResponse.isSuspended()) {
         // Clear span from the asyncResponse. Logically this should never happen. Added to be safe.
         InstrumentationContext.get(AsyncResponse.class, Span.class).put(asyncResponse, null);
       }
-      if (asyncResponse == null || !asyncResponse.isSuspended()) {
+      if (asyncReturnValue != null) {
+        // span finished by CompletionStageFinishCallback
+        asyncReturnValue = asyncReturnValue.whenComplete(new CompletionStageFinishCallback<>(span));
```
anuraaga, line 169: I don't think we consistently do so in this codebase, but FWIU `handle` has much better performance than `whenComplete` (due to a JDK bug?)

https://github.com/line/armeria/issues/1426

Can we use handle here?

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsHttpServerTest.groovy
```diff
@@ -73,6 +104,15 @@ abstract class JaxRsHttpServerTest<S> extends HttpServerTest<S> {
     true
   }
 
+  private static boolean supportsJaxRs21() {
```
anuraaga, line 107: If this is false normally and true when running latest deps, recommend being more explicit about it like our redisson instrumentation

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/redisson-3.0/redisson-3.0.gradle#L21

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsHttpServerTest.groovy
```diff
@@ -58,6 +59,36 @@ abstract class JaxRsHttpServerTest<S> extends HttpServerTest<S> {
...
+    assumeTrue(supportsJaxRs21())
+
+    given:
+    def url = HttpUrl.get(address.resolve("/async-completion-stage")).newBuilder()
+      .addQueryParameter("action", action)
+      .build()
+    def request = request(url, "GET", null).build()
+
+    when:
+    def response = client.newCall(request).execute()
```
anuraaga, line 73: IIUC, it's important our test makes sure a span hasn't been ended before the future is completed. How about adding some sleep in the API method, and separate out making the call, checking there are no traces (with previous code I think there would have been), than joining on the future, then checking there are traces?

anuraaga (commented):
Cool, thanks

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/instrumentation/auto/jaxrs/v2_0/JaxRsAnnotationsInstrumentation.java
mateuszrzeszutek: > Can we use handle here?

Of course -- changed. Thanks!

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsHttpServerTest.groovy
mateuszrzeszutek: 👍 
That's much better, done.

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsHttpServerTest.groovy
mateuszrzeszutek: Instead of adding sleep (which would make this test at least a bit flaky) I've added synchronisation using a `CyclicBarrier` -- but the principle is the same: no traces until the future is complete.

anuraaga (approved):
Thanks!

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-jersey-2.0/jaxrs-2.0-jersey-2.0.gradle
```diff
@@ -24,3 +24,7 @@ dependencies {
   testImplementation group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.3'
   testImplementation group: 'com.fasterxml.jackson.module', name: 'jackson-module-afterburner', version: '2.9.10'
 }
+
+test {
+  systemProperty 'testLatestDeps', testLatestDeps
```
iNikem, line 29: Why is this needed? @anuraaga is it?

iNikem: A, got it...

iNikem (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-jersey-2.0/jaxrs-2.0-jersey-2.0.gradle
anuraaga: A, but this is a great point out. @mateuszrzeszutek if you get a chance to add a one line comment here sometime to avoid future confusion would be great!

Issue #1129, opened 2020-08-29
Decide on what modules contain public APIs


trask:
A couple of options to start discussion:

**Option 1:**
Merge `auto-api` module (back) into `javaagent-bootstrap`

This would leave us with public API in 2 modules:
* `javaagent-tooling`
* `javaagent-bootstrap`

**Option 2:**
Rename `auto-api` to `auto-runtime-api`, and split out `auto-matcher-api` from `javaagent-tooling`, and introduce `javaagent-spi`(?) for `TracerCustomizer`, `ExporterConfig`, `SpanExporterFactory`, etc

This would remove all public API from `javaagent-tooling` and `javaagent-bootstrap`, and leave us with public API in 3 modules:
* `auto-runtime-api`
* `auto-matcher-api`
* `javaagent-spi`(?)

The `auto-*-api` modules would be public API for creating auto instrumentation.

The `javaagent-spi`(?) module would be for creating a custom Javaagent distribution.

anuraaga:
Yeah option 2 feel natural to me. Right now we indeed have a tricky split between classes in the tooling / bootstrap modules that isn't intuitive.

iNikem:
I am still somewhat confused about what classes should go where, but `auto-runtime-api` seems like a strange name to me :)

anuraaga:
Heh yeah need a reminder of what auto runtime API is. SPI I think used by initializer or exporter, matcher used by instrumentation. Is the rest agent initialization code?

anuraaga:
Want to knock this out soon to get closer to our final artifacts sooner than later. I looked again and am wondering

- Do we need to split auto-matcher-api / auto-runtime-api? What if we move `matcher` package and `Instrumenter` interface to `auto-api`. IIUC, it should be possible to have a single artifact that auto-instrumentation depends on instead of two. We would probably also move some of `auto-api` back into bootstrap, `AgentClassLoader`, `InternalJarURLHandler` don't seem like they would ever be needed in auto-instrumentation modules. 

- We do need to move our SPI classes somewhere - there are too many public classes in `javaagent-tooling` for us to let users depend on it as a standard practice (e.g., `Constants.java`). I think these could also just be stuffed into `auto-api` - it's all the public API exposed by our agent, which includes SPI.

Thoughts? Will start with PRs with any vote in any direction :)

anuraaga:
[...]

PR #1187, opened 2020-09-10
Split out artifact for javaagent SPI.

anuraaga:
This seemed like the least controversial and easiest of the artifact refactors so figured I'd send a PR

For #1129


trask (approved):
👍

cc: @jkwatson - external exporter jar will need to be re-built again. sorry.

iNikem (approved):

Issue #1103, opened 2020-08-26
java.lang.StackOverflowError during Reactor instrumentation


iNikem:
I have tried to run a moderate load test on https://github.com/spring-petclinic/spring-petclinic-microservices with Otel 0.7.0 attached. It runs nicely on light load, but with a little bit more load it gave me this error:

```
java.lang.StackOverflowError: null
	at java.base/java.util.HashMap.hash(Unknown Source)
	at java.base/java.util.HashMap.get(Unknown Source)
	at java.base/java.util.Collections$SynchronizedMap.get(Unknown Source)
	at ch.qos.logback.classic.util.LogbackMDCAdapter.get(LogbackMDCAdapter.java:151)
	at org.slf4j.MDC.get(MDC.java:203)
	at brave.context.slf4j.MDCScopeDecorator$MDCContext.getValue(MDCScopeDecorator.java:92)
	at brave.baggage.CorrelationScopeDecorator$Multiple.decorateScope(CorrelationScopeDecorator.java:194)
	at org.springframework.cloud.sleuth.log.Slf4jScopeDecorator.decorateScope(Slf4jScopeDecorator.java:95)
	at brave.propagation.CurrentTraceContext.decorateScope(CurrentTraceContext.java:108)
	at brave.propagation.CurrentTraceContext.maybeScope(CurrentTraceContext.java:141)
	at org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.request(ScopePassingSpanSubscriber.java:75)
	at reactor.core.publisher.MonoPeekTerminal$MonoTerminalPeekSubscriber.request(MonoPeekTerminal.java:132)
	at io.opentelemetry.auto.instrumentation.reactor.shaded.TracingSubscriber.request(TracingSubscriber.java:98)
	at org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.request(ScopePassingSpanSubscriber.java:76)
	at reactor.core.publisher.FluxPeekFuseable$PeekFuseableSubscriber.request(FluxPeekFuseable.java:137)
	at io.opentelemetry.auto.instrumentation.reactor.shaded.TracingSubscriber.request(TracingSubscriber.java:98)
	at org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.request(ScopePassingSpanSubscriber.java:76)
[...]

PR #1189, opened 2020-09-10
Reactor bugs

### LINKED ISSUE #1138: Bugs in Project Reactor integration
I have two new tests in my branch: https://github.com/open-telemetry/opentelemetry-java-instrumentation/compare/master...iNikem:reactor-bugs?expand=1. Both of them currently fails.

One of them demonstrates that `Mono.subscribe` does not propagate context.

Another demonstrates that Reactor Netty client can attach spans to wrong traces.

I have no idea atm how to solve them.

iNikem:
Fixes #1138 

The main difference in this PR is what context gets captured and propagated with Project Reactor operators. Before context was captured during pipeline assembly. This PR changes it to capture context during subscription to the pipeline. I believe this is the right way to do. Also tests confirm that :) Those which failed, now are green.

Spring reactive WebClient integration was fixed in the similar way. NB! WebClient integration borrows most of its code from Spring Sleuth project. I don't know how to properly attribute that code to the original source, please advise. For now I just add it to javadoc.

I still have weird issues with some version of Spring Boot, but I believe it will not affect the majority of this PR and its new approach to context capture. Therefore, please review and comment.

Also there is more work to do to polish tests and project structure. E.g. change `reactor-3.1` modules into new structure. That will be a follow up work.


devinsba (approved):
Looks good to me, assuming the lettuce 5 tests also pass

iNikem:
And why all this works on my machine? :(

trask:
It's failing for me on your branch, let me know if you still can't repro it locally and I can look more.

```
$ java -version
openjdk version "11.0.3" 2019-04-16 LTS
OpenJDK Runtime Environment Zulu11.31+11-CA (build 11.0.3+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.31+11-CA (build 11.0.3+7-LTS, mixed mode)

$ ./gradlew :instrumentation:spring:spring-webflux-5.0:auto:test

> Configure project :
Inferred project: opentelemetry-java-instrumentation, version: 0.8.0-SNAPSHOT

> Task :instrumentation:spring:spring-webflux-5.0:auto:test
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/C:/git/opentelemetry-java-instrumentation/instrumentation/spring/spring-webflux-5.0/auto/build/libs/test-bootstrap.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/C:/Users/trstalna/.m2/repository/ch/qos/logback/logback-classic/1.2.3/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/C:/Users/trstalna/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-simple/1.7.30/e606eac955f55ecf1d8edcccba04eb8ac98088dd/slf4j-simple-1.7.30.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
17:23:35,791 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
17:23:35,792 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback.groovy]
17:23:35,793 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Found resource [logback.xml] at [file:/C:/git/opentelemetry-java-instrumentation/instrumentation/spring/spring-webflux-5.0/auto/build/resources/test/logback.xml]
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
17:23:35,856 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - debug attribute not set
17:23:35,859 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - About to instantiate appender of type [ch.qos.logback.core.ConsoleAppender]
17:23:35,861 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - Naming appender as [console]
17:23:35,900 |-WARN in ch.qos.logback.core.ConsoleAppender[console] - This appender no longer admits a layout as a sub-component, set an encoder instead.
17:23:35,900 |-WARN in ch.qos.logback.core.ConsoleAppender[console] - To ensure compatibility, wrapping your layout in LayoutWrappingEncoder.
17:23:35,900 |-WARN in ch.qos.logback.core.ConsoleAppender[console] - See also http://logback.qos.ch/codes.html#layoutInsteadOfEncoder for details
17:23:35,901 |-INFO in ch.qos.logback.classic.joran.action.RootLoggerAction - Setting level of ROOT logger to WARN
17:23:35,901 |-INFO in ch.qos.logback.core.joran.action.AppenderRefAction - Attaching appender named [console] to Logger[ROOT]
17:23:35,901 |-INFO in ch.qos.logback.classic.joran.action.LoggerAction - Setting level of logger [io.opentelemetry.auto] to DEBUG
17:23:35,901 |-INFO in ch.qos.logback.classic.joran.action.ConfigurationAction - End of configuration.
17:23:35,902 |-INFO in ch.qos.logback.classic.joran.JoranConfigurator@30b7c004 - Registering current configuration as safe fallback point

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/C:/Users/trstalna/.gradle/caches/modules-2/files-2.1/org.codehaus.groovy/groovy/2.5.11/4d8f25c5da08af46fb204fd493ec894699a3f4e8/groovy-2.5.11.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

SingleThreadedSpringWebfluxTest > GET test with async response #testName FAILED
    org.gradle.internal.exceptions.DefaultMultiCauseException at TestRun.java:196
        Caused by: org.spockframework.runtime.ConditionNotSatisfiedError at InMemoryExporterAssert.groovy:71
            Caused by: org.codehaus.groovy.runtime.powerassert.PowerAssertionError at SpanAssert.groovy:103

...
```

iNikem:
WebFlux tests I can repro now, don't know why they were green during my pre-PR run :( They are easy to fix.

What puzzles me is failure of Cassandra tests on java 8. Don't see yet how this PR changes that.

anuraaga (commented):
Do we have any test of concatenated fluxes? Conceptually it seems like only subscriber should need to provide context but in practice I've found publisher does too (IIUC this PR removes that), though this was a while ago perhaps there were reactor bugs at the time.

https://github.com/line/armeria/blob/master/examples/context-propagation/reactor/src/main/java/example/armeria/contextpropagation/reactor/MainService.java#L91

anuraaga:
More concretely, from my debugging of this back in the day, I found no consistency in whether an operator runs on the subscriber thread of publisher thread - so even if subscriber has context I found cases where operators missed them.

iNikem:
> Do we have any test of concatenated fluxes?

I don't. Can you supply one? And if it fails, I would really love to address that separately. I think this PR is a step in right direction, even if it does not cover all the cases atm.

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -0,0 +1,63 @@
...
+    Hooks.onEachOperator(tracingLift());
+  }
+
+  /** Unregisters the hook registered by {@link #registerOnEachOperator()}. */
+  public static void resetOnEachOperator() {
+    Hooks.onEachOperator(tracingLift());
+  }
+
+  private static <T> Function<? super Publisher<T>, ? extends Publisher<T>> tracingLift() {
+    return Operators.lift(new Lifter<>());
```
anuraaga, line 48: I think this needs to return the same reference for `reset` to work

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.function.Function;
+import org.reactivestreams.Publisher;
+import reactor.core.CoreSubscriber;
+import reactor.core.Fuseable;
+import reactor.core.Scannable;
+import reactor.core.publisher.Hooks;
+import reactor.core.publisher.Operators;
+
+/** @author Based on Spring Sleuth's Reactor instrumentation. */
+public class TracingOperator {
```
anuraaga, line 30: /cc @minwoox do you mind providing any thoughts on this instrumentation? I guess you've already experienced all the corner cases since LINE uses reactor a lot. It will affect Armeria users too if they start using OpenTelemetry :)

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingSubscriber.java
```diff
@@ -18,132 +18,65 @@
...
-public class TracingSubscriber<T>
-    implements Subscription, CoreSubscriber<T>, Fuseable.QueueSubscription<T>, Scannable {
-
-  private final io.grpc.Context upstreamContext;
-  private final CoreSubscriber<T> delegate;
+/**
+ * Based on Spring Sleuth's Reactor instrumentation. A trace representation of the {@link
+ * Subscriber}
+ *
+ * @author Jose Montoya
```
anuraaga, line 30: Since we don't have `@author` anywhere else I would recommend not including it here too even for copied files.

Ideally we fix spotless to allow us to have two copyright headers on a file. In the meantime, I took a NOTICE approach to referencing copied code which I think is the bare minimum legally (`@author` doesn't help at all legally)

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/logback/logback-1.0.0/library/NOTICE.txt

instrumentation/netty/netty-4.1/src/test/resources/logback.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
anuraaga, line 1: I think with the `AgentTestRunner` fix we shouldn't need to copy logback.xml around

instrumentation/netty/netty-4.1/src/test/resources/logback.xml
```diff
@@ -0,0 +1,18 @@
...
+        %d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n
+      </Pattern>
+    </layout>
+  </appender>
+
+  <root level="WARN">
+    <appender-ref ref="console"/>
+  </root>
+
+  <logger name="io.opentelemetry.auto" level="debug"/>
```
anuraaga, line 16: And this logback.xml like many of ours is wrong :-D /cc @trask

instrumentation/spring/spring-webflux-5.0/auto/src/test/groovy/SpringWebfluxTest.groovy
```diff
@@ -34,6 +31,9 @@ import server.FooModel
 import server.SpringWebFluxTestApplication
 import server.TestController
 
+import static io.opentelemetry.trace.Span.Kind.INTERNAL
```
anuraaga, line 34: Believe IntelliJ isn't set up correctly - https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/intellij-setup.md

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingSubscriber.java
```diff
@@ -18,132 +18,65 @@
...
-import reactor.core.Scannable;
 import reactor.util.context.Context;
 
-public class TracingSubscriber<T>
-    implements Subscription, CoreSubscriber<T>, Fuseable.QueueSubscription<T>, Scannable {
-
-  private final io.grpc.Context upstreamContext;
-  private final CoreSubscriber<T> delegate;
+/**
+ * Based on Spring Sleuth's Reactor instrumentation. A trace representation of the {@link
```
anuraaga, line 27: Can you add permalinks to original of all copied classes?

anuraaga (commented):
Yeah will try to add a test for concatenated publishers based on the link later.

iNikem:
I had to disable one test: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1189/files#diff-b13ec53a5a6d43118f8232d493c27a12R126. Will add a separate issue about this

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
iNikem: I don't quite understand. You mean cache the result of `lift`? This is copy-pasted from OpenTracing: https://github.com/opentracing-contrib/java-reactor/blob/master/src/main/java/io/opentracing/contrib/reactor/TracedSubscriber.java

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
minwoox: Well, LINE uses RxJava a lot. 🤣 
Recently we got a few customers who use Reactor with Armeria, so I guess I didn't experience all the corner cases yet.
This implementation looks good and if I found something I will leave comments though. 😄

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
anuraaga: Yeah like `private static final Function<? super Publisher<T>, ? extends Publisher<T>>  TRACING_LIFTER = Operators.lift(new Lifter<>());` and `return TRACING_LIFTER;`. I didn't look at the implementation of reset in detail but I can't conceive it's possible without it being the same reference.

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -0,0 +1,63 @@
...
+   * reactive stream. This should generally be called in a static initializer block in your
+   * application.
+   */
+  public static void registerOnEachOperator() {
+    Hooks.onEachOperator(tracingLift());
+  }
+
+  /** Unregisters the hook registered by {@link #registerOnEachOperator()}. */
+  public static void resetOnEachOperator() {
+    Hooks.onEachOperator(tracingLift());
```
anuraaga, line 44:
```suggestion
    Hooks.resetOnEachOperator(tracingLift());
```

Can you add a test for reset?

instrumentation-core/reactor-3.1/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
iNikem: That was actual bug, thank you :) Solved now

anuraaga (approved):
Thanks!

iNikem:
Damn! I managed to change my test so that it fails again :(

instrumentation/netty/netty-4.1/src/test/resources/logback.xml
iNikem: Btw, it was not wrong! `InMemoryExporter` still lives in `io.opentelemetry.auto.test` package!

trask: well, it's not _totally_ wrong then 😂

fixed in #1194

instrumentation/cassandra/cassandra-3.0/src/test/groovy/CassandraClientTest.groovy
```diff
@@ -128,10 +122,42 @@ class CassandraClientTest extends AgentTestRunner {
     "SELECT * FROM users where name = 'alice' ALLOW FILTERING"                                         | "async_test"
   }
 
+  //TODO fix this test
+  @Ignore("""
```
trask, line 126: i'm getting this same failure in #1194 😭

it's _somehow_ due to fixing the logger name in `AgentTestRunner`, so if you revert that change you should be able to bring back this test, and i'll investigate as part of #1194

iNikem:
Although I now at least one more bug in our Reactor instrumentation (more precisely in the interplay between Project Reactor, Reactor Netty and Netty), I propose to merge this PR. It fixes some bugs and uses "better" integration points (subscription as opposed to assembly).

iNikem:
Last commit slow fixes #1103

trask:
🎉

iNikem:
No reason to celebrate :( Reactor instrumentation still does not work well. It is such a mess :( I will continue looking for help understanding how to properly propagate context in Project Reactor and friends.

instrumentation/reactor-3.1/src/test/groovy/ReactorCoreTest.groovy
```diff
@@ -242,117 +241,59 @@ class ReactorCoreTest extends AgentTestRunner {
...
     runUnderTrace {
-      // The operations in the publisher created here all end up children of the publisher-parent
+      // The "add one" operations in the publisher created here should be children of the publisher-parent
       Publisher<Integer> publisher = publisherSupplier()
 
-      // After this activation, all additions to the assembly will create new traces
       def tracer = OpenTelemetry.getTracer("test")
-      def scope = tracer.withSpan(DefaultSpan.getInvalid())
+      def intermediate = tracer.spanBuilder("intermediate").startSpan()
+      // After this activation, the "add two" operations below should be children of this span
```
devinsba, line 272: This comment doesn't match the assertions below. Not sure that is intended or not

instrumentation/spring/spring-webflux-5.0/auto/src/test/groovy/SpringWebfluxTest.groovy
```diff
@@ -173,7 +172,7 @@ class SpringWebfluxTest extends AgentTestRunner {
         }
         span(2) {
           operationName "tracedMethod"
-          childOf span(1)
+          childOf span(0)
```
devinsba, line 175: I think this change was wrong. Now the modeling isn't correct:

BEFORE:

```
netty handler
-> spring controller
---> traced method
```

AFTER
```
netty handler
-> spring controller
-> traced method
```

The traced method ends up a child of the netty span instead of the spring controller span. **This was the reason that I had captured the context at assembly time**

iNikem: This is not so obvious for me. If we run such application under profiler, then I expect that there will NOT be a single callstack with "netty -> spring controler -> traced method". Because traced method was actually executed AFTER returning from spring controller.

In addition I definitely expect that `[start:end]` of traced method IS NOT contained in `[start:end]` of spring controller. Currently it IS and I think this is a bug. Again, due to profiler/callstack argument above.

For these reasons I think that capturing context during assembly is wrong.

devinsba: I follow what you are saying but from a user point of view, I would consider that traced method to be a part of my controller call. I understand that it runs after the controller returns but my intent as the user would be that the traced method is part of the controller

devinsba:
```
  @GetMapping("/foo-mono-from-callable/{id}")
  Mono<FooModel> getMonoFromCallable(@PathVariable("id") long id) {
    return Mono.fromCallable { return tracedMethod(id) }
  }

  @Trace()
  private FooModel tracedMethod(long id) {
    return new FooModel(id, "tracedMethod")
  }
```

I think this is an async modeling question but from a user point of view the model makes most sense to me to have `tracedMethod` as a child of `getMonoFromCallable`

iNikem: Yeah, that is kinda true. But there are examples of wrong traces in both cases: capture on assembly and capture on subscribe. I think the whole reactive story is currently very weak.

devinsba:
Due to the project being compiled with JDK 11 now, there are a couple tests that I am seeing fail with JDK 8 in the DD project that I can't prove fail in this project due to the fact I can't even run them with JDK 8

```
> Task :dd-java-agent:instrumentation:lettuce-5:test

Lettuce5ReactiveClientTest > async subscriber FAILED
    org.spockframework.runtime.ConditionNotSatisfiedError at Lettuce5ReactiveClientTest.groovy:508
        Caused by: org.codehaus.groovy.runtime.powerassert.PowerAssertionError at Lettuce5ReactiveClientTest.groovy:394

Lettuce5ReactiveClientTest > async subscriber with specific thread pool FAILED
    org.spockframework.runtime.ConditionNotSatisfiedError at Lettuce5ReactiveClientTest.groovy:508
        Caused by: org.codehaus.groovy.runtime.powerassert.PowerAssertionError at Lettuce5ReactiveClientTest.groovy:449
```

Both caused by:

```
assert trace.size() == expectedSize
       |     |      |  |
       |     2      |  3
       |            false
```

Seems like this is handled differently in JDK 8. Perhaps the thread pool implementation changed 🤷 

```
    runUnderTrace("test-parent") {
      reactiveCommands.set("a", "1")
        .then(reactiveCommands.get("a")) // The get here is ending up in another trace
        .subscribeOn(Schedulers.elastic())
        .subscribe()
    }
```

trask:
hey @devinsba! Check out [executing tests with specific java version](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/running-tests.md#executing-tests-with-specific-java-version)

Also here's the Java 8 buildscan for this PR: https://scans.gradle.com/s/iv2z7ue655wis/tests?search=LettuceReactiveClientTest

instrumentation/spring/spring-webflux-5.0/auto/src/test/groovy/SpringWebfluxTest.groovy
trask: This is an interesting example, and similar to:

```
  @GetMapping("/foo-callable/{id}")
  Callable<FooModel> getCallable(@PathVariable("id") long id) {
    return () -> { return tracedMethod(id) }
  }

  @Trace()
  private FooModel tracedMethod(long id) {
    return new FooModel(id, "tracedMethod")
  }
```

(https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-async)

We wouldn't capture `traced method` under `controller` in this example either (since we capture context when passing the callable to the executor, and not when constructing the callable).

devinsba:
That buildscan shows it was run with a Java 11 VM: https://scans.gradle.com/s/iv2z7ue655wis#infrastructure

instrumentation/spring/spring-webflux-5.0/auto/src/test/groovy/SpringWebfluxTest.groovy
trask: What do you think of propagating the context from the controller to the return value when the return value represents an asynchronously executed operation? This seems similar to how we propagate the context across other async boundaries.

anuraaga: I don't grok reactor very well but @trask's example is pretty easy to understand. By the way, do we handle the span correctly? If the span ends when `getCallable` returns, it seems broken even outside of just context propagation. This seems like a very special case.

My understanding here is that spring seems to offer a very specific execution model - the user "registers a task", but the actual execution happens by the framework. Because this task corresponds to a request, we want it to also correspond to a Span. This seems like it requires very specific handling - I think it's what you're saying Trask and I agree with it.

My feeling is this is a spring (or any framework that provides a similar task registration execution model) specific aspect, I don't think it should leak into Reactor. It does seem true that we can't really guarantee when subscription happens, maybe it happens in the wrong context. But the "reactive model" I think is - register many callbacks and then execute by calling subscribe, as opposed to completablefuture where you register callbacks while executing. But in user code I would expect the vast majority of usages to expect that subscription to be the parent - it is similar to our concept of an "invocation context" which we often remount, and it should apply to reactor as well, the subscriber is the invoker. If that doesn't fit with a separate abstraction, like webflux, we could handle that separately.

Unrelated to our instrumentation, but my understanding is that the challenge we have with reactive is they don't seem to define the concept of a scope but probably should - is the scope assembly, subscription, or something else, it's something that affects how we propagate and needs to be known. I personally use Dagger producers and never have context issues since it has a concept of a `ProductionComponent` which acts like a scope (similar to DI scopes). So all callbacks defined in that component run with the same context, the context of that scope. Works great, and no code required by the user as long as they follow the one requirement that the Dagger entrypoint needs to be executed in the request handler

`@Production` defines the executor used when invoking a component - so just set it to the request executor and problem solved :)
https://github.com/curioswitch/curiostack/blob/master/common/server/framework/src/main/java/org/curioswitch/common/server/framework/ServerModule.java#L250

Invocation looks like this. Yes if that invocation happened after being delegated to a threadpool it wouldn't work, but my framework just doesn't allow that (would throw guaranteed runtime exception), knowing how important context is.
https://github.com/curioswitch/curiostack/blob/master/cafe-map/server/src/main/java/org/curioswitch/cafemap/server/CafeMapService.java#L63

Unlike my framework, reactor does not have such a requirement - but I think we can still expect common usage by users to be that the stream is "scoped" to the subscription point.

anuraaga: @iNikem @trask Just random update on this point, @kojilin came up with an interesting example which further seems to make context at subscription time, not assembly time, make sense

https://github.com/line/armeria/issues/3051#issuecomment-752413533

If contexting at assembly time, `cache` will generally cause disasters.

PR #1190, opened 2020-09-10
Lettuce 5.1 - Remove redis args from span attributes for command such as AUTH

RashmiRam:
1. Added `name != "AUTH"` check so that args of AUTH command won't be added in `db.statement` attribute value.
2. Added tests only in `LettuceSyncClientTest.groovy`


iNikem (approved):

trask (commented):
Thanks @RashmiRam!

Can you extract the new test out into a separate class, e.g. `LettuceSyncClientAuthTest`? I think that will be cleaner than spinning up two redis servers in this test, and hopefully we can consolidate into a base class at some point in the future to share similar code across all the `Lettuce*ClientTest` classes.

anuraaga (approved):
Thanks!

PR #499, opened 2020-06-11
Changed instrumentation tracer" TRACER.startSpan() to never return `null`

trask:
Changed `TRACER.startSpan()` to never return `null`, I think this is less surprising.

PR #1191, opened 2020-09-10
Enhance AWS DynamoDB instrumentation

### LINKED ISSUE #1181: Enhance AWS DynamoDB instrumentation with db.* attributes
Current implementation of AWS SDK instrumentation treats all types of requests (S3, SQS, Kinesis, DynamoDB) as network ones, setting net.* and http.* attributes. DynamoDB however is more of a DB kind. Therefore, in order to provide high quality tracing data, DynamoDB instrumentation should be enhanced to set db.* attributes (when appropriate). 

There is, however, one outstanding question - what should be done with network (net.*) and HTTP (http.*) attributes currently set also for DynamoDB requests? Shall we keep adding all of them or leave network only (as the call is done via network communication) ?

kubawach:
Resolves #1181 

- enhanced DynamoDB instrumentation for AWS SDK 2.2 (added DB attributed)
- added tests to cover DynamoDB requests separately
- a bit of refactoring

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/RequestType.java
+++ b/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/RequestType.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.awssdk.v2_2;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.awssdk.core.SdkRequest;
+
+enum RequestType {
+  S3("S3Request", "Bucket"),
+  SQS("SqsRequest", "QueueUrl", "QueueName"),
+  Kinesis("KinesisRequest", "StreamName"),
+  DynamoDB("DynamoDbRequest", "TableName");
+
+  private final String requestClass;
+  private final String[] fields;
+
+  RequestType(String requestClass, String... fields) {
+    this.requestClass = requestClass;
+    this.fields = fields;
+  }
+
+  String[] getFields() {
+    return fields;
+  }
+
+  @Nullable
+  static RequestType ofSdkRequest(SdkRequest request) {
+    // exact request class should be 1st level child of request type
+    String typeName =
+        (request.getClass().getSuperclass() == null
+                ? request.getClass()
+                : request.getClass().getSuperclass())
+            .getSimpleName();
+    for (RequestType type : values()) {
+      if (type.requestClass.equals(typeName)) {
+        return type;
+      }
+    }
+    return null;
+  }
+}
--- a/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
+++ b/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
@@ -16,10 +16,16 @@
 
 package io.opentelemetry.instrumentation.awssdk.v2_2;
 
-import static io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkClientTracer.TRACER;
+import static io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdk.getSpanFromAttributes;
+import static io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkHttpClientTracer.TRACER;
+import static io.opentelemetry.instrumentation.awssdk.v2_2.RequestType.ofSdkRequest;
 
 import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.Span.Kind;
+import java.util.EnumMap;
+import java.util.HashMap;
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import software.amazon.awssdk.awscore.AwsResponse;
 import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SdkResponse;
@@ -35,7 +41,29 @@ final class TracingExecutionInterceptor implements ExecutionInterceptor {
   static final ExecutionAttribute<Span> SPAN_ATTRIBUTE =
       new ExecutionAttribute<>("io.opentelemetry.auto.Span");
 
+  static final ExecutionAttribute<RequestType> REQUEST_TYPE_ATTRIBUTE =
+      new ExecutionAttribute<>("io.opentelemetry.auto.aws.RequestType");
+
   static final String COMPONENT_NAME = "java-aws-sdk";
+
+  private static final Map<RequestType, SdkRequestDecorator> TYPE_TO_DECORATOR = mapDecorators();
+  private static final Map<String, String> FIELD_TO_ATTRIBUTE = mapFieldToAttribute();
+
+  private static Map<RequestType, SdkRequestDecorator> mapDecorators() {
+    Map<RequestType, SdkRequestDecorator> result = new EnumMap<>(RequestType.class);
+    result.put(RequestType.DynamoDB, new DbRequestDecorator());
+    return result;
+  }
+
+  private static Map<String, String> mapFieldToAttribute() {
+    Map<String, String> result = new HashMap<>();
+    result.put("QueueUrl", "aws.queue.url");
+    result.put("Bucket", "aws.bucket.name");
+    result.put("QueueName", "aws.queue.name");
+    result.put("StreamName", "aws.stream.name");
+    result.put("TableName", "aws.table.name");
+    return result;
+  }
 
   private final Kind kind;
 
@@ -43,77 +71,55 @@ final class TracingExecutionInterceptor implements ExecutionInterceptor {
     this.kind = kind;
   }
 
+  @Nullable
+  private SdkRequestDecorator decorator(ExecutionAttributes executionAttributes) {
+    RequestType type = getTypeFromAttributes(executionAttributes);
+    return TYPE_TO_DECORATOR.get(type);
+  }
+
+  private RequestType getTypeFromAttributes(ExecutionAttributes executionAttributes) {
+    return executionAttributes.getAttribute(REQUEST_TYPE_ATTRIBUTE);
+  }
+
   @Override
   public void beforeExecution(
       Context.BeforeExecution context, ExecutionAttributes executionAttributes) {
-    Span span = TRACER.getOrCreateSpan(spanName(executionAttributes), AwsSdk.tracer());
+    Span span = TRACER.getOrCreateSpan(spanName(executionAttributes), AwsSdk.tracer(), kind);
     executionAttributes.putAttribute(SPAN_ATTRIBUTE, span);
+    RequestType type = ofSdkRequest(context.request());
+    if (type != null) {
+      executionAttributes.putAttribute(REQUEST_TYPE_ATTRIBUTE, type);
+    }
   }
 
   @Override
   public void afterMarshalling(
       Context.AfterMarshalling context, ExecutionAttributes executionAttributes) {
-    Span span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
+    Span span = getSpanFromAttributes(executionAttributes);
     if (span != null) {
       TRACER.onRequest(span, context.httpRequest());
-      onSdkRequest(span, context.request());
-      onAttributes(span, executionAttributes);
+      SdkRequestDecorator decorator = decorator(executionAttributes);
+      if (decorator != null) {
+        decorator.decorate(span, context.request(), executionAttributes);
+      }
+      decorateWithGenericRequestData(span, context.request());
+      decorateWithExAttributesData(span, executionAttributes);
     }
   }
 
-  @Override
-  public void afterExecution(
-      Context.AfterExecution context, ExecutionAttributes executionAttributes) {
-    Span span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
-    if (span != null) {
-      executionAttributes.putAttribute(SPAN_ATTRIBUTE, null);
-      TRACER.afterExecution(span, context.httpRequest());
-      onSdkResponse(span, context.response());
-      TRACER.end(span, context.httpResponse());
+  private void decorateWithGenericRequestData(Span span, SdkRequest request) {
+
+    RequestType type = ofSdkRequest(request);
+    if (type != null) {
+      for (String field : type.getFields()) {
+        request
+            .getValueForField(field, String.class)
+            .ifPresent(val -> span.setAttribute(FIELD_TO_ATTRIBUTE.get(field), val));
+      }
     }
   }
 
-  @Override
-  public void onExecutionFailure(
-      Context.FailedExecution context, ExecutionAttributes executionAttributes) {
-    Span span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
-    if (span != null) {
-      executionAttributes.putAttribute(SPAN_ATTRIBUTE, null);
-      TRACER.endExceptionally(span, context.exception());
-    }
-  }
-
-  Span onSdkRequest(Span span, SdkRequest request) {
-    // S3
-    request
-        .getValueForField("Bucket", String.class)
-        .ifPresent(name -> span.setAttribute("aws.bucket.name", name));
-    // SQS
-    request
-        .getValueForField("QueueUrl", String.class)
-        .ifPresent(name -> span.setAttribute("aws.queue.url", name));
-    request
-        .getValueForField("QueueName", String.class)
-        .ifPresent(name -> span.setAttribute("aws.queue.name", name));
-    // Kinesis
-    request
-        .getValueForField("StreamName", String.class)
-        .ifPresent(name -> span.setAttribute("aws.stream.name", name));
-    // DynamoDB
-    request
-        .getValueForField("TableName", String.class)
-        .ifPresent(name -> span.setAttribute("aws.table.name", name));
-    return span;
-  }
-
-  String spanName(ExecutionAttributes attributes) {
-    String awsServiceName = attributes.getAttribute(SdkExecutionAttribute.SERVICE_NAME);
-    String awsOperation = attributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
-
-    return awsServiceName + "." + awsOperation;
-  }
-
-  Span onAttributes(Span span, ExecutionAttributes attributes) {
+  private void decorateWithExAttributesData(Span span, ExecutionAttributes attributes) {
 
     String awsServiceName = attributes.getAttribute(SdkExecutionAttribute.SERVICE_NAME);
     String awsOperation = attributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
@@ -121,15 +127,45 @@ final class TracingExecutionInterceptor implements ExecutionInterceptor {
     span.setAttribute("aws.agent", COMPONENT_NAME);
     span.setAttribute("aws.service", awsServiceName);
     span.setAttribute("aws.operation", awsOperation);
-
-    return span;
   }
 
-  // Not overriding the super.  Should call both with each type of response.
-  Span onSdkResponse(Span span, SdkResponse response) {
+  @Override
+  public void afterExecution(
+      Context.AfterExecution context, ExecutionAttributes executionAttributes) {
+    Span span = getSpanFromAttributes(executionAttributes);
+    if (span != null) {
+      clearAttributes(executionAttributes);
+      TRACER.afterExecution(span, context.httpRequest());
+      onSdkResponse(span, context.response());
+      TRACER.end(span, context.httpResponse());
+    }
+  }
+
+  private void onSdkResponse(Span span, SdkResponse response) {
     if (response instanceof AwsResponse) {
       span.setAttribute("aws.requestId", ((AwsResponse) response).responseMetadata().requestId());
     }
-    return span;
+  }
+
+  @Override
+  public void onExecutionFailure(
+      Context.FailedExecution context, ExecutionAttributes executionAttributes) {
+    Span span = getSpanFromAttributes(executionAttributes);
+    if (span != null) {
+      clearAttributes(executionAttributes);
+      TRACER.endExceptionally(span, context.exception());
+    }
+  }
+
+  private void clearAttributes(ExecutionAttributes executionAttributes) {
+    executionAttributes.putAttribute(SPAN_ATTRIBUTE, null);
+    executionAttributes.putAttribute(REQUEST_TYPE_ATTRIBUTE, null);
+  }
+
+  private String spanName(ExecutionAttributes attributes) {
+    String awsServiceName = attributes.getAttribute(SdkExecutionAttribute.SERVICE_NAME);
+    String awsOperation = attributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
+
+    return awsServiceName + "." + awsOperation;
   }
 }
--- a/instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
+++ b/instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
@@ -16,6 +16,7 @@
 
 package io.opentelemetry.instrumentation.awssdk.v2_2
 
+import static com.google.common.collect.ImmutableMap.of
 import static io.opentelemetry.auto.test.server.http.TestHttpServer.httpServer
 import static io.opentelemetry.trace.Span.Kind.CLIENT
 
@@ -34,7 +35,14 @@ import software.amazon.awssdk.http.apache.ApacheHttpClient
 import software.amazon.awssdk.regions.Region
 import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue
 import software.amazon.awssdk.services.dynamodb.model.CreateTableRequest
+import software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest
+import software.amazon.awssdk.services.dynamodb.model.DeleteTableRequest
+import software.amazon.awssdk.services.dynamodb.model.GetItemRequest
+import software.amazon.awssdk.services.dynamodb.model.PutItemRequest
+import software.amazon.awssdk.services.dynamodb.model.QueryRequest
+import software.amazon.awssdk.services.dynamodb.model.UpdateItemRequest
 import software.amazon.awssdk.services.ec2.Ec2AsyncClient
 import software.amazon.awssdk.services.ec2.Ec2Client
 import software.amazon.awssdk.services.kinesis.KinesisClient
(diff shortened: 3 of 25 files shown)


anuraaga:
This turned out much larger than I was anticipating :-)

One thing I should have mentioned earlier is it's on my plate to refactor the handling of the "custom attributes" in a generic way. AWS has essentially semantic conventions defined for the AWS SDK, which we implement in xray and think otel should as well since they're benifecial for any tracing vendor. They are a generic transformation of the JSON itself like here

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/java/com/amazonaws/xray/interceptors/TracingInterceptor.java#L187
https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/resources/com/amazonaws/xray/interceptors/DefaultOperationParameterWhitelist.json

So it'd be great if we can hold off on adding refactoring to make a generic way of adding custom attributes since I think much of the logic we have right now will actually go away.

Also, I don't think the intention of the DB conventions is to normalize DB statements into something SQL-like. DynamoDB is essentially the same type of DB as MongoDB, but the example recommends not setting statement probably because it's JSON, not a statement

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#mongodb

So I don't think we should fill statement, which makes a lot of the complexity go away I think, the PR could probably just be something like

```java
if (awsServiceName.equals("DynamoDB")) {
  db.system = "dynamodb"
  db.operation = awsOperationName
}
```

Sorry for not being stronger about not including statement in the issue itself ><

anuraaga:
Alternatively, if we do really want to include statement, I would just put the `SdkResult` JSON as-is, without any sort of transformation logic on our side. But given the example for MongoDB in the spec, I wouldn't recommend it.

iNikem:
@anuraaga can you elaborate on `the handling of the "custom attributes" in a generic way`? I did not quite get it, what attributes are you talking about :)

iNikem:
I can see the benefit of having some sort of statement-like info on DynomoDB span. I am not very familiar with AWS SDK and I don't see what json are we talking about? :) But I do see that subclasses of `DynamoDbRequest` seem to have nice `toString` method. May be that will be enough?

anuraaga:
> can you elaborate on the handling of the "custom attributes" in a generic way? I did not quite get it, what attributes are you talking about :)

Yeah basically we have a mapping from JSON fields to semantic convention attributes in [this](https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/resources/com/amazonaws/xray/interceptors/DefaultOperationParameterWhitelist.json) data file. So we generically map SDK requests to richer information using it. One of my "job milestone tasks" is to port those conventions to OTel ;)

> I can see the benefit of having some sort of statement-like info on DynomoDB span.

Yeah I think `.toString()` is probably essentially the JSON that would have been sent, though I haven't checked. I'd be ok with that - but it might be worth confirming that we should be doing similar for mongodb too in the spec itself, I don't see a reason for dynamodb and mongodb to differ in terms of semantic conventions since they're so similar.

trask:
> it might be worth confirming that we should be doing similar for mongodb too in the spec itself, I don't see a reason for dynamodb and mongodb to differ in terms of semantic conventions since they're so similar.

I think [the spec section on mongodb](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#mongodb) is just an example, e.g. it doesn't show `db.connection_string` being captured either.

Our MongoDB instrumentation captures both `db.connection_string` and `db.statement`.

Capturing `db.statement` does bring PII problems though, e.g. for MongoDB we keep the json keys and scrub most of the json values.

anuraaga:
I should have looked at the mongodb code :D Seems like we record a scrubbed JSON object there. 

@kubawach Do you think it's possible to implement similar scrubbing in a generic way instead of having to manually serialize each different type of request? I think for the most part, any `SdkField` that ends up as a type of `AttributeValue` we can scrub and otherwise write them as is looking at the `Marshalling` information in the field. It would also be great to reduce the scope of the PR, it's hard to judge the refactoring along with the logic changes. Our current instrumentation is very simple logic I think, it's easier to understand refactoring after things get complex not before :) I might recommend

- PR to add db attributes except for statement
- PR to add statement attribute using generic serialization mechanism
- PR to refactor if things crossed a complexity line after those

What do you think?

iNikem:
I like the idea of smaller PRs

iNikem:
> * PR to add db attributes except for statement
> * PR to add statement attribute using generic serialization mechanism
> * PR to refactor if things crossed a complexity line after those

@kubawach ^^

kubawach:
Thanks for the comments, I appreciate all of them. I'm really happy to see that the proposal sparked a discussion regarding handling of the DynamoDB.

Before I attend the comments and proposals, let me give you some background my motivations of particular design choices. 

1. I come from business application development (over 10 years). I could say that I switched from consumer / user of telemetry solutions to the programmer / developer side. Over lat years I appreciated a lot when my telemetry gave me as many data points as possible, especially when issues (ie of performance nature) occurred. In such cases attributes like "statement" come __very handy__ tracing root cause. Thus a decision to have that included. 

2. DynamoDB domain-specific language is quite hard to read. It evolved over the years, for example switching from "Expected" to "ConditionExpression". However the first attribute is still present in the API, cluttering serialised form. Therefore I decided to include only relevant attributes and introduce a "normalisation" of the requests, making it akin to well-known SQL and easier to read by most developers. All above - in order to maintain readability. 

Regarding ideas that were mentioned in the course of the discussion (thanks once again):

@iNikem I believe that regular toString() serialisation of requests won't do the trick, as for exemplary request it looks like that: "CreateTableRequest(AttributeDefinitions=software.amazon.awssdk.core.util.DefaultSdkAutoConstructList@103f73a, TableName=sometable, KeySchema=software.amazon.awssdk.core.util.DefaultSdkAutoConstructList@103f73a, LocalSecondaryIndexes=software.amazon.awssdk.core.util.DefaultSdkAutoConstructList@103f73a, GlobalSecondaryIndexes=software.amazon.awssdk.core.util.DefaultSdkAutoConstructList@103f73a)"

@anuraaga regarding "generic" scrubbing instead of custom-per-request one - happy to do so, but we need to mind two issues here:
1. PII as pointed out by @trask 
2. Readability (limiting number of attributes serialized)
Therefore if we decide, as a community, to resign from custom ("normalised") form and include attributes, I believe we should scrub keep a kind of "allowed attributes" list and include only these. 

@anuraaga regarding splitting the PR - happy to do so as well, once we agree on the final approach (mainly attributes scrubbing). I'd then go for 2 separate PRs:
1. Adding generic DynamoDB attributes + general refactoring (those are really tightly coupled - adding a way to decorate spans per AwsSdkRequest type)
2. Generating statements for DynamoDbRequests.

Let me know what you think :)

iNikem:
I understand your point of view, @kubawach. I think we have a contention around `db.statement`, but we generally agree that DynamoDB spans should have DB semantic convention. Therefore, I would propose to make first PR which adds those and to use the simplest thing possible for `db.statement`: just simple class name.

Then, as a follow up task, let us discuss our options how to add proper db statement. That will be tougher task, so let us address it separately.

kubawach:
@iNikem great idea, splitting the MR then.

kubawach:
Dear all (@iNikem and @anuraaga in particular) - I've just split the MR to cover base refactoring / DB attributes here and statement generation in #1199 Please move with discussion there or to appropriate issue #1198 :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
```diff
@@ -0,0 +1,59 @@
...
+import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import io.opentelemetry.trace.DefaultSpan;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.Tracer;
+import io.opentelemetry.trace.TracingContextUtils;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.SdkExecutionAttribute;
+
+public class SpanFactory {
```
iNikem, line 28: Why this was extracted from `AwsSdkClientTracer`?

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
```diff
@@ -73,6 +80,106 @@ abstract class AbstractAws2ClientTest extends InstrumentationSpecification {
...
+    expect:
+    response != null
+    response.class.simpleName.startsWith(operation)
+    assertDynamoDbRequest(service, operation, path, method, requestId)
+
+    where:
+    [service, operation, method, path, requestId, builder, call] << dynamoDbRequestDataTable()
+  }
+
+  def "send DynamoDB #operation async request with builder #builder.class.getName() mocked response"() {
```
iNikem, line 107: I don't see how this test differs from the previous one

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
```diff
@@ -73,6 +80,106 @@ abstract class AbstractAws2ClientTest extends InstrumentationSpecification {
...
+            "${SemanticAttributes.HTTP_STATUS_CODE.key()}" 200
+            "${SemanticAttributes.HTTP_USER_AGENT.key()}" { it.startsWith("aws-sdk-java/") }
+            "aws.service" "$service"
+            "aws.operation" "${operation}"
+            "aws.agent" "java-aws-sdk"
+            "aws.requestId" "$requestId"
+            "aws.table.name" "sometable"
+            "${SemanticAttributes.DB_SYSTEM.key()}" "dynamodb"
+            "${SemanticAttributes.DB_NAME.key()}" "sometable"
+            "${SemanticAttributes.DB_OPERATION.key()}" "${operation}"
```
iNikem, line 152: can we get `db.user`?

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
```diff
@@ -73,6 +80,106 @@ abstract class AbstractAws2ClientTest extends InstrumentationSpecification {
...
+      trace(0, 1) {
+        span(0) {
+          operationName "$service.$operation"
+          spanKind CLIENT
+          errored false
+          parent()
+          attributes {
+            "${SemanticAttributes.NET_PEER_NAME.key()}" "localhost"
+            "${SemanticAttributes.NET_PEER_PORT.key()}" server.address.port
+            "${SemanticAttributes.HTTP_URL.key()}" { it.startsWith("${server.address}${path}") }
```
iNikem, line 141: Can we set https://github.com/open-telemetry/opentelemetry-specification/blob/bc272ab52869df2145065d55300d721ff008e34b/specification/trace/semantic_conventions/span-general.md#nettransport-attribute ?

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/HttpAwsSdkClientTracer.java
```diff
@@ -18,20 +18,17 @@
...
-import io.opentelemetry.trace.Tracer;
-import io.opentelemetry.trace.TracingContextUtils;
 import io.opentelemetry.trace.attributes.SemanticAttributes;
 import java.net.URI;
 import software.amazon.awssdk.http.SdkHttpHeaders;
 import software.amazon.awssdk.http.SdkHttpRequest;
 import software.amazon.awssdk.http.SdkHttpResponse;
 
-final class AwsSdkClientTracer
+final class HttpAwsSdkClientTracer
```
iNikem, line 28: IF you want to rename it, this `AwsSdkHttpClientTracer` will be better name. It follows the general (loose) guideline of class name being `<Library>HttpClientTracer extends HttpClientTracer`

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -35,101 +42,121 @@
...
+              }
+            });
+  }
+
+  private void decorateWithExAttributesData(Span span, ExecutionAttributes attributes) {
+
+    String awsServiceName = attributes.getAttribute(SdkExecutionAttribute.SERVICE_NAME);
+    String awsOperation = attributes.getAttribute(SdkExecutionAttribute.OPERATION_NAME);
+
+    span.setAttribute("aws.agent", COMPONENT_NAME);
```
iNikem, line 124: @anuraaga do you (at AWS I mean) have plans for formalising AWS SDK semantic convention in Otel spec?

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
kubawach: Separation of concerns / testability. Although I have not added a unit test, I'd be much easier now.

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: Good catch, forgot to propagate non-statement changes from the "original" PR.

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: Good question! What do you think about adding it to (generic HTTP tracer base class) - HttpClientTracer ? This way all HTTP based tracers would set "TCP.IP".

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: I don't think so. Client is (usually) configured with AWS access key (id / value) that does not say anything about a particular user.

iNikem: can access key id be used as username? I am not sure how sensitive this info is.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
iNikem: All other Tracers do create spans. If you extract that from them, what will remain? :)

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
iNikem: That is a good idea

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
kubawach: Pure tracing code ;) I don't have any other explanation for this refactoring apart from object oriented design principles :)

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: Will do :)

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: In general it can - the AWS AccessKeyId is used in Presigned URLs, API requests etc. Therefore it's usually AFAIK not considered as a secret / sensitive piece of data. @anuraaga any word on this?

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
anuraaga: Yeah I've started writing a markdown doc for it but need to finish it 😅 This one in particular should probably be deleted since it overlaps with the `InstrumentationLibraryInfo` but that's for a separate PR

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -35,101 +42,121 @@
...
+    result.put("TableName", "aws.table.name");
+    return result;
+  }
 
   TracingExecutionInterceptor(Kind kind) {
-    this.kind = kind;
+    this.spanFactory = new SpanFactory(kind);
+  }
+
+  private Optional<SdkRequestDecorator> decorator(ExecutionAttributes executionAttributes) {
```
anuraaga, line 75: Let's just return the nullable value instead of `Optional`, we don't really use that pattern in this or the Java SDK repo. It could be a good time to add a dependency on checker framework annotations which we haven't done here yet for `@Nullable`

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -35,101 +42,121 @@
...
+  static final ExecutionAttribute<RequestType> REQUEST_TYPE_ATTRIBUTE =
+      new ExecutionAttribute<>("io.opentelemetry.auto.aws.RequestType");
+
   static final String COMPONENT_NAME = "java-aws-sdk";
 
-  private final Kind kind;
+  private final SpanFactory spanFactory;
+
+  private final Map<RequestType, SdkRequestDecorator> decoratorsMap = mapDecorators();
+  private final Map<String, String> fieldToAttribute = mapFieldToAttribute();
```
anuraaga, line 53: Let's make the constant-type things static

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -35,101 +42,121 @@
...
+
+  private Map<RequestType, SdkRequestDecorator> mapDecorators() {
+    Map<RequestType, SdkRequestDecorator> result = new EnumMap<>(RequestType.class);
+    result.put(RequestType.DynamoDB, new DbRequestDecorator());
+    return result;
```
anuraaga, line 58: `unmodifiableMap` for both

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
anuraaga: I sort of agree that extracting this class doesn't seem to have huge bang for buck and makes this instrumentation significantly different from many of our other tracers. But it's ok too. Let's make sure all the new classes are package private instead of public

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/db/DbRequestDecorator.java
```diff
@@ -0,0 +1,49 @@
...
+    sdkRequest
+        .getValueForField("TableName", String.class)
+        .ifPresent(val -> span.setAttribute(SemanticAttributes.DB_NAME.key(), val));
+
+    operation(attributes)
+        .ifPresent(
+            operation -> span.setAttribute(SemanticAttributes.DB_OPERATION.key(), operation));
+  }
+
+  private Optional<String> operation(ExecutionAttributes attributes) {
```
anuraaga, line 45: Ditto

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
anuraaga: The access key ID is an ID of the key itself, but not of the IAM role which is closer to what a user is in a MySQL DB. While it could be mapped to a user by resolving the key ID through IAM API or something, I think in general the key ID isn't really useful for tracing and would leave it out.

anuraaga (commented):
@kubawach Thanks a lot for the summary of your reasoning it helps me a lot :) I agree that we want to add statement in some form, and thanks for separating it out since it's definitely a much harder topic than the standard DB attributes so glad to focus on those here.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
iNikem: I don't know if there is any performance difference, but other than that I would prefer `Optional`. But indeed, it should then be a global decision for the sake of consistency.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
iNikem: I still think I disagree with this refactoring atm. We are still in the process of migrating from old Decorators to new Tracers and of unifying their usage. Introducing one more pattern of structuring and using tracers will only impede that unification.

Sorry to be a party pooper, but I would ask to move it back :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
anuraaga: There's a small performance difference - I think `Optional` can still make sense for public APIs where users may not have tooling to deal with `@Nullable` (though `Optional` works poorly with other JVM languages like Kotlin / Scala so still not great...), for our private methods, we should just do what we can at compile time at compile time :) https://github.com/uber/NullAway

trask: fwiw, I also prefer `null` + `@Nullable` + compile-time tooling https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/499#issuecomment-643560951

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/SpanFactory.java
kubawach: No problem with that.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
kubawach: Before we decide against Optionals everywhere besides module / component public API, there are few things we need to consider:
- it's already, at least in the AWS SDK interfacing part - eg `getValueForField` returns Optionals - ie inside AWS SDK tracing code it's consistent with the SDK itself 
- there is possible performance difference (invokedynamic for functional calls vs cmpz for plain-old NULL check) but IMO it's offset greatly by readability (no explicit branch) and additional safeguard (optional clearly says - beware, this might be null)
- I mostly agree with @iNikem comments regarding introduction of static type check - results might get ignored easily (if not failing the build upon a problem) - or will take a lot of effort to get the build running after introduction (if failing the build)
- out of my experience Optionals work better than static code checkers - for example if we break the checker (remove it accidentally, lib is broken etc) the code will become vulnerable, language construct will stay till it's manually removed / chaged
- removing Optionals now will be unsafe as null check is not implemented in the project yet

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
kubawach: Done (with relevant name change).

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
kubawach: Not sure about that. Maps are private, not exposed to external (not leaking), unmod view will definitely add some CPU cycles (delegating methods). Don't see benefit here (it would be different if the map was exposed or leaking).

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
kubawach: It has already been here (not my change) so I'll resolve the conversation until we have a semantic convention in place :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/db/DbRequestDecorator.java
kubawach: Yep, perfectly possible to get a null here ;)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
anuraaga: We surely shouldn't add nullness checking without having it fail the build. I went ahead and sent #1203 finally to allow adding nullaway to modules.

> it's already, at least in the AWS SDK interfacing part - eg getValueForField returns Optionals - ie inside AWS SDK tracing code it's consistent with the SDK itself

It's more important for us to have consistency within our codebase than following the library patterns, unless there is a very specific use case. For example, I think our codebase becomes less consistent / maintainable for its authors if we prefer `Optional` in aws sdk instrumentation and `Nullable` in armeria instrumentation because of the library choices

> it's offset greatly by readability (no explicit branch) and additional safeguard (optional clearly says - beware, this might be null)

Readability argument is personal opinion I guess, but I find a small if statement to generally be at least as readable as a lambda. I don't think there's a difference in safeguardness when we have the checker in place (IntelliJ is excellent at highlighting usages of Nullable without check too)

> for example if we break the checker 

I feel this is true of all our tooling - if we break shadowJar configuration, agent doesn't work, etc. I've had great experience with static checking myself. So I'd like to be confident in our tooling as a whole :-D

anuraaga: I merged #1203 in case you want to use it :) Otherwise I'm happy to add the checking later. But would be nice to stick with nullable here, I think that PR also shows the overwhelming use of it already in our codebase

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -35,101 +42,131 @@
...
     if (span != null) {
-      executionAttributes.putAttribute(SPAN_ATTRIBUTE, null);
+      clearAttributes(executionAttributes);
       TRACER.afterExecution(span, context.httpRequest());
       onSdkResponse(span, context.response());
       TRACER.end(span, context.httpResponse());
     }
   }
 
+  // Not overriding the super.  Should call both with each type of response.
```
iNikem, line 144: This comment currently does not make sense here.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/db/DbRequestDecorator.java
```diff
@@ -0,0 +1,42 @@
...
+import software.amazon.awssdk.core.SdkRequest;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.SdkExecutionAttribute;
+
+public final class DbRequestDecorator implements SdkRequestDecorator {
+
+  @Override
+  public void decorate(Span span, SdkRequest sdkRequest, ExecutionAttributes attributes) {
+
+    span.setAttribute(SemanticAttributes.DB_SYSTEM.key(), "dynamodb");
```
iNikem, line 31: `SemanticAttributes.DB_SYSTEM.set(span, "dynamodb")` makes it easier to find non-semantic attributes in the future

iNikem (approved):

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/RequestType.java
```diff
@@ -0,0 +1,54 @@
...
+  public static Optional<RequestType> ofSdkRequest(SdkRequest request) {
+    // exact request class should be 1st level child of request type
+    String typeName =
+        (request.getClass().getSuperclass() == null
+                ? request.getClass()
+                : request.getClass().getSuperclass())
+            .getSimpleName();
+    for (RequestType type : values()) {
+      if (type.requestClass.equals(typeName)) {
+        return Optional.of(type);
```
anuraaga, line 49: Let's switch to `Nullable` here too

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/RequestType.java
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.awssdk.v2_2;
+
+import java.util.Optional;
+import software.amazon.awssdk.core.SdkRequest;
+
+public enum RequestType {
```
anuraaga, line 22: Can you make all the new classes package-private instead of public?

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/RequestType.java
```diff
@@ -0,0 +1,55 @@
...
+    this.requestClass = requestClass;
+    this.fields = fields;
+  }
+
+  public String[] getFields() {
+    return fields;
+  }
+
+  @Nullable
+  public static RequestType ofSdkRequest(SdkRequest request) {
```
anuraaga, line 41:
```suggestion
  static RequestType ofSdkRequest(SdkRequest request) {
```

anuraaga (approved):
Just one small nit, and addressing it will kick the CI so hopefully it doesn't flake

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/RequestType.java
```diff
@@ -0,0 +1,55 @@
...
+
+  private final String requestClass;
+  private final String[] fields;
+
+  RequestType(String requestClass, String... fields) {
+    this.requestClass = requestClass;
+    this.fields = fields;
+  }
+
+  public String[] getFields() {
```
anuraaga, line 36:
```suggestion
  String[] getFields() {
```

PR #1192, opened 2020-09-11
Add initial baseline instrumentation for AWS lambda runtime.

anuraaga:
This is very barebones, notable missing pieces are

1) Need to handle `RequestStreamHandler` as well
2) Need to take into account the function could be HTTP or messaging, which even means a different SpanKind

But having this baseline would help us implement more of these pieces (while also trying out the instrumentation for real and seeing how badly lambda's cold start / freezing behavior effect trace export).


iNikem:
@kubawach did an extensive research about AWS Lambda, including the performance perspective. IIRC, one of the conclusions was that having full-blown bytecode manipulation agent in Lambda is "very bad idea". Performance is just dismal.

anuraaga:
Having auto instrumentation of Java Lambda is a very high priority for my team :) So we are going to invest in making it work, including things on the lambda side itself if that is needed. @kubawach if it's possible for you to share findings (can email me to the address on my profile if confidential), that would be great in informing us of directions we need to take.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public final O handleRequest(I input, Context context) {
+    Span span = tracer.startSpan(context);
+    Throwable error = null;
+    try (Scope unused = TracingContextUtils.currentContextWith(span)) {
+      return doHandleRequest(input, context);
+    } catch (Throwable t) {
+      error = t;
```
trask, line 61: 👍

i wish there were better way to do this

in other places we are doing the simpler, but less correct 🙁

```
      try (Scope ignored = TRACER.startScope(span)) {
        ...
        TRACER.end(span, response);
        return response;
      } catch (Throwable t) {
        TRACER.endExceptionally(span, t);
        throw t;
      }
```

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
```diff
@@ -0,0 +1,73 @@
...
+   */
+  protected TracingRequestHandler(AwsLambdaTracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Override
+  public final O handleRequest(I input, Context context) {
+    Span span = tracer.startSpan(context);
+    Throwable error = null;
+    try (Scope unused = TracingContextUtils.currentContextWith(span)) {
```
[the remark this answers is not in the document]
anuraaga: Ah didn't know intellij treated that differently

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AwsLambdaInstrumentationHelper.java
```diff
@@ -0,0 +1,30 @@
...
+
+import io.opentelemetry.instrumentation.awslambda.v1_0.AwsLambdaTracer;
+
+public final class AwsLambdaInstrumentationHelper {
+
+  private static final AwsLambdaTracer TRACER = new AwsLambdaTracer();
+
+  public static AwsLambdaTracer tracer() {
+    return TRACER;
+  }
```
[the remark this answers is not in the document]
anuraaga: Yeah good point, went back to old pattern for now and will file an issue to discuss.

trask (approved):
👍

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
iNikem: Why is it less correct?

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AwsLambdaRequestHandlerInstrumentation.java
```diff
@@ -0,0 +1,86 @@
...
+        AwsLambdaRequestHandlerInstrumentation.class.getName() + "$HandleRequestAdvice");
+  }
+
+  public static class HandleRequestAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.Argument(1) Context context,
+        @Advice.Local("otelSpan") Span span,
+        @Advice.Local("otelScope") Scope scope) {
+      span = TRACER.startSpan(context);
```
iNikem, line 68: Auto instrumentation does not reuse `TracingRequestHandler` from library instrumentation, is it expected?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Tracer;
+
+/**
+ * A base class similar to {@link RequestHandler} but will automatically trace invocations of {@link
+ * #doHandleRequest(Object, Context)}.
+ */
+public abstract class TracingRequestHandler<I, O> implements RequestHandler<I, O> {
```
iNikem, line 29: Should there be a documentation or an example how to use it?

iNikem (commented):
Should update the instrumentations list in README?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
trask: it's only less correct if there's an internal exception inside `TRACER.end(span, response)` (at which point maybe we don't care?)

fwiw, the way it's written in this PR maps to how our `@OnMethodEnter`/`@OnMethodExit` behave, which I think makes it nice, if we're going to fail, we fail in a more consistent way

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AwsLambdaRequestHandlerInstrumentation.java
anuraaga: Yeah since you define the lambda function as the string of the class name, there isn't really any way to use an `@Advice` to apply the manual instrumentation I think (some very advanced bytebuddy maybe :) ). Currently all the logic is in `TRACER` so it's ok, I'll move things around if that stops being the case.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
anuraaga: Yup good call - added some

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
anuraaga: Hmm, thinking more on this I think both patterns are wrong. @trask's example seems especially wrong since we shouldn't have one codepath that ends a span with the scope active and another that doesn't since it's inconsistent. Let's imagine `SimpleSpanProcessor`, it may execute an RPC inline with the call to `end`, but will have a different parent for the two cases.

So the question becomes whether it's correct to have the parent in both cases or none of the cases. I guess having the parent is better - `SimpleSpanProcessor` is effectively running span processing, maybe even export, in the context of the request, possibly slowing it down a lot, so it should be traced together with the request.

I'll leave my code as is though since I don't think we have that behavior anywhere in our codebase right now :)

anuraaga: @trask Yeah the exception problem means we're breaking user code anyways so much bigger problem :) I'm less concerned about the exception handling though ^^

anuraaga: Well, I think I realized that can't work, if `SimpleSpanProcessor` was allowed to emit spans, it would emit as an infinite loop inline with the request :O

kubawach:
@anuraaga great to see some AWS lambda instrumentation work, I'll contact you sharing the details of some ground works we did in the same area :)

PR #1193, opened 2020-09-11
Muzzle should fail on unimplemented abstract methods

### LINKED ISSUE #1123: Muzzle should fail on unimplemented abstract methods

mateuszrzeszutek:
Resolves #1123

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
@@ -68,6 +68,10 @@ public interface HelperReferenceWrapper {
       return isAbstract;
     }
 
+    public String getDeclaringClass() {
+      return declaringClass;
+    }
+
     public String getName() {
       return name;
     }
@@ -88,11 +92,6 @@ public interface HelperReferenceWrapper {
     public int hashCode() {
       return Objects.hashCode(name, descriptor);
     }
-
-    @Override
-    public String toString() {
-      return declaringClass + "#" + name + descriptor;
-    }
   }
 
   class Factory {
@@ -109,12 +108,12 @@ public interface HelperReferenceWrapper {
     }
 
     private HelperReferenceWrapper create(String className) {
-      if (helperReferences.containsKey(className)) {
-        return new ReferenceType(helperReferences.get(className));
-      }
       Resolution resolution = classpathPool.describe(className);
       if (resolution.isResolved()) {
         return new ClasspathType(resolution.resolve());
+      }
+      if (helperReferences.containsKey(className)) {
+        return new ReferenceType(helperReferences.get(className));
       }
       throw new IllegalStateException("Missing class " + className);
     }
@@ -151,8 +150,7 @@ public interface HelperReferenceWrapper {
       }
 
       private boolean hasActualSuperType() {
-        return reference.getSuperName() != null
-            && !reference.getSuperName().equals(Object.class.getName());
+        return reference.getSuperName() != null;
       }
 
       private Function<String, HelperReferenceWrapper> toWrapper() {
@@ -215,8 +213,7 @@ public interface HelperReferenceWrapper {
       }
 
       private boolean hasActualSuperType() {
-        return type.getSuperClass() != null
-            && !type.getSuperClass().asErasure().equals(TypeDescription.OBJECT);
+        return type.getSuperClass() != null;
       }
 
       // Uses guava iterables to avoid unnecessary collection copying
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
@@ -177,7 +177,7 @@ public interface HelperReferenceWrapper {
           @Override
           public boolean apply(Reference.Method input) {
             return !(input.getFlags().contains(Flag.STATIC)
-                || input.getFlags().contains(Flag.PRIVATE_OR_HIGHER)
+                || input.getFlags().contains(Flag.PRIVATE)
                 || CONSTRUCTOR_INTERNAL_NAME.equals(input.getName()));
           }
         };
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
@@ -1,0 +1,278 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.javaagent.tooling.muzzle;
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singleton;
+import static net.bytebuddy.description.method.MethodDescription.CONSTRUCTOR_INTERNAL_NAME;
+
+import com.google.common.base.Function;
+import com.google.common.base.Objects;
+import com.google.common.base.Predicate;
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.Iterables;
+import io.opentelemetry.javaagent.tooling.muzzle.Reference.Flag;
+import java.util.Map;
+import net.bytebuddy.description.method.MethodDescription.InDefinedShape;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.pool.TypePool;
+import net.bytebuddy.pool.TypePool.Resolution;
+
+/** This class provides a common interface for {@link Reference} and {@link TypeDescription}. */
+public interface HelperReferenceWrapper {
+  boolean isAbstract();
+
+  /**
+   * @return true if the wrapped type extends any class other than {@link Object} or implements any
+   *     interface.
+   */
+  boolean hasSuperTypes();
+
+  /**
+   * @return An iterable containing the wrapped type's super class (if exists) and implemented
+   *     interfaces.
+   */
+  Iterable<HelperReferenceWrapper> getSuperTypes();
+
+  /** @return An iterable with all non-private, non-static methods declared in the wrapped type. */
+  Iterable<Method> getMethods();
+
+  final class Method {
+    private final boolean isAbstract;
+    private final String declaringClass;
+    private final String name;
+    private final String descriptor;
+
+    public Method(boolean isAbstract, String declaringClass, String name, String descriptor) {
+      this.isAbstract = isAbstract;
+      this.declaringClass = declaringClass;
+      this.name = name;
+      this.descriptor = descriptor;
+    }
+
+    public boolean isAbstract() {
+      return isAbstract;
+    }
+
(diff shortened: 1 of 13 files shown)


trask:
Thank you for tackling this ❤️👍

iNikem:
Hei, @mateuszrzeszutek, maybe you can help @trask with #536 ;)

trask:
> Hei, @mateuszrzeszutek, maybe you can help @trask with #536 ;)

💯

@mateuszrzeszutek I will review this PR tomorrow 👍

mateuszrzeszutek:
> Hei, @mateuszrzeszutek, maybe you can help @trask with #536 ;)

mateuszrzeszutek:
> Hei, @mateuszrzeszutek, maybe you can help @trask with #536 ;)

Sure - after implementing this change I think I can answer at least half of these questions. I'll put it on my TODO list and push a PR soon.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
```diff
@@ -0,0 +1,278 @@
...
+            .filter(isOverrideable())
+            .transform(toMethod());
+      }
+
+      private Predicate<Reference.Method> isOverrideable() {
+        return new Predicate<Reference.Method>() {
+          @Override
+          public boolean apply(Reference.Method input) {
+            return !(input.getFlags().contains(Flag.STATIC)
+                || input.getFlags().contains(Flag.PRIVATE_OR_HIGHER)
```
iNikem, line 180: Is private method overridable?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleVisitor.java
```diff
@@ -133,28 +133,18 @@ public MethodVisitor visitMethod(
     }
 
     public Reference[] generateReferences() {
-      // track sources we've generated references from to avoid recursion
```
iNikem, line 136: Why this is not required any more?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/Reference.java
```diff
@@ -363,25 +364,36 @@ public boolean matches(int asmFlags) {
...
-        return anotherFlag == NON_FINAL;
+        return anotherFlag == NON_FINAL || anotherFlag == ABSTRACT;
       }
 
       @Override
       public boolean matches(int asmFlags) {
         return (Opcodes.ACC_FINAL & asmFlags) != 0;
       }
     },
+    ABSTRACT {
```
iNikem, line 386: Can you add javadoc to methods of `Flag` enum?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -147,7 +157,64 @@ private boolean doesMatch(ClassLoader loader) {
...
 
-  public static List<Reference.Mismatch> checkMatch(
+  // for helper classes we make sure that all abstract methods from super classes and interfaces are
+  // implemented
+  private List<Reference.Mismatch> checkHelperClassMatch(Reference helperClass, TypePool typePool) {
+    List<Mismatch> mismatches = Collections.emptyList();
+
+    HelperReferenceWrapper helperWrapper = new Factory(typePool, references).create(helperClass);
+
+    if (!helperWrapper.hasSuperTypes() || helperWrapper.isAbstract()) {
```
iNikem, line 167: Do we support abstract helpers?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -147,7 +157,64 @@ private boolean doesMatch(ClassLoader loader) {
...
+    }
+  }
+
+  // ignore Object methods - sometimes defined by interfaces which causes them to be treated as
+  // abstract
+  private static boolean shouldIgnore(HelperReferenceWrapper.Method superMethod) {
+    return "hashCode".equals(superMethod.getName()) && "()I".equals(superMethod.getDescriptor())
+        || "equals".equals(superMethod.getName())
+            && "(Ljava/lang/Object;)Z".equals(superMethod.getDescriptor())
+        || "toString".equals(superMethod.getName())
```
iNikem, line 213: `clone` and `finalize`?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
mateuszrzeszutek: We do -- we just skip the abstract method check for them, because it's perfectly fine for them to contain abstract methods.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
mateuszrzeszutek: Good catch, will add.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleVisitor.java
mateuszrzeszutek: Because it served no purpose from the start: because `adviceClassNames` always contained unique elements `referenceSources.contains(adviceClass)` always returned false.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
mateuszrzeszutek: Nope, they're not - so we can skip checking them.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/Reference.java
mateuszrzeszutek: I'll add javadoc to `matches()` method -- I've checked the usages of all of them and it turned out that `contradicts()` and `supersedes()` are never used. I guess that I'll be doing a general code clean-up together with muzzle documentation 😄

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
iNikem: Now I am even more confused. We create a predicate for `isOverrideable` and we say it is `true` when method 

- is not static AND
- is not private or higher AND
- is not a constructor

aren't ALL methods "private or higher"? Which makes this predicate constantly `false`?

mateuszrzeszutek: Ah, that's what you meant. The problem here is that `PRIVATE_OR_HIGHER` (and others) is used in two ways in muzzle:
1) it just simply marks that a method/class is private which is the case here, or
2) it describes a minimum access level required between a type and a method/field/class.

I'll split those two types of usage into separate flags to make it a bit more clear.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceCreator.java
```diff
@@ -244,12 +253,68 @@ public FieldVisitor visitField(
   @Override
   public MethodVisitor visitMethod(
       int access, String name, String descriptor, String signature, String[] exceptions) {
+
+    // declared method references are not generated for advice classes, only for helper classes
```
trask, line 257: maybe as an optimization here, don't generate method references for private methods, <init> or <clinit> since they can't implement an abstract method

trask (commented):
have you tried to see if this would have caught the ServletOutputStream issue? i gave it a try, but muzzle still didn't fail for me on unimplemented ServletOutputStream methods

mateuszrzeszutek:
> have you tried to see if this would have caught the ServletOutputStream issue? i gave it a try, but muzzle still didn't fail for me on unimplemented ServletOutputStream methods

You're right, I unknowingly introduced a bug when I was refactoring the first working version. Thanks for noticing it! I'll add a unit test that covers this specific scenario.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/Reference.java
```diff
@@ -295,62 +296,63 @@ String getMismatchDetails() {
...
       String getMismatchDetails() {
-        return "Missing method " + className + "#" + method;
+        return "Missing method " + className + "#" + methodSignature;
       }
     }
   }
 
   /** Expected flag (or lack of flag) on a class, method, or field reference. */
   public enum Flag {
+    // The following constants represent the exact visibility of a referenced class/method
```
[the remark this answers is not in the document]
mateuszrzeszutek: I thought about extracting a common `Flag` interface and making all these new enums implement it, but leaving them all in a single enum set. I like the separate fields idea because of how easy it'd be to detect merge conflicts (you can't have a field that is both static and non-static).

I'll do that in a separate PR - I'll refactor, clean up and document muzzle someday soon.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -147,7 +157,67 @@ private boolean doesMatch(ClassLoader loader) {
...
+      (method.isAbstract() ? abstractMethods : plainMethods).add(method);
+    }
+
+    for (HelperReferenceWrapper superType : type.getSuperTypes()) {
+      collectMethodsFromTypeHierarchy(superType, abstractMethods, plainMethods);
+    }
+  }
+
+  // ignore Object methods - sometimes defined by interfaces which causes them to be treated as
+  // abstract
```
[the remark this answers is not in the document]
mateuszrzeszutek: Honestly I can't remember now why I stopped recursion before `Object` - I must have run into some issue with that, but I can't remember what it was. I've removed both this method and the `Object` condition in `HelperReferenceWrapper` and everything works as expected, so I'll remove this code altogether. The simpler it is the better.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceCreator.java
mateuszrzeszutek: That's a very good idea, I'll do that. Thanks!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -147,7 +157,67 @@ private boolean doesMatch(ClassLoader loader) {
...
+    Set<HelperReferenceWrapper.Method> unimplementedMethods =
+        Sets.difference(abstractMethods, plainMethods);
+    for (HelperReferenceWrapper.Method unimplementedMethod : unimplementedMethods) {
+      mismatches =
+          lazyAdd(
+              mismatches,
+              new Reference.Mismatch.MissingMethod(
+                  helperClass.getSources().toArray(new Reference.Source[0]),
+                  helperClass.getClassName(),
+                  unimplementedMethod.toString()));
```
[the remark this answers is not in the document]
mateuszrzeszutek: Good point, changed.

iNikem (approved):

trask (approved):
🎉🎉

PR #1194, opened 2020-09-12
Fix test logging

trask:
to reflect new package names

### DIFF
--- a/instrumentation/spring/spring-webflux-5.0/auto/src/test/resources/logback.xml
+++ b/instrumentation/spring/spring-webflux-5.0/auto/src/test/resources/logback.xml
@@ -13,6 +13,6 @@
     <appender-ref ref="console"/>
   </root>
 
-  <logger name="io.opentelemetry.auto" level="debug"/>
+  <logger name="io.opentelemetry" level="debug"/>
 
 </configuration>
--- a/instrumentation/spring/spring-webflux-5.0/auto/src/test/resources/logback.xml
+++ b/instrumentation/spring/spring-webflux-5.0/auto/src/test/resources/logback.xml
@@ -13,6 +13,8 @@
     <appender-ref ref="console"/>
   </root>
 
+  <!-- this is needed because when Spring Boot starts it overrides the debug log level that was
+        configured in AgentTestRunner -->
   <logger name="io.opentelemetry" level="debug"/>
 
 </configuration>
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -125,6 +125,7 @@ public class AgentInstaller {
       agentBuilder =
           agentBuilder
               .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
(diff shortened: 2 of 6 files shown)


instrumentation/spring/spring-webflux-5.0/auto/src/test/resources/logback.xml
```diff
@@ -13,6 +13,6 @@
     <appender-ref ref="console"/>
   </root>
 
-  <logger name="io.opentelemetry.auto" level="debug"/>
+  <logger name="io.opentelemetry" level="debug"/>
```
iNikem, line 16: Do we need these individual configuration files then?

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -125,6 +125,7 @@ public static ResettableClassFileTransformer installBytebuddyAgent(
       agentBuilder =
           agentBuilder
               .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
+              .with(AgentBuilder.RedefinitionStrategy.DiscoveryStrategy.Reiterating.INSTANCE)
```
trask, line 128: cc: @tylerbenson, I think this will fix that Cassandra test for you (https://github.com/DataDog/dd-trace-java/pull/1802)

instrumentation/spring/spring-webflux-5.0/auto/src/test/resources/logback.xml
trask: ya, they are needed because these tests run Spring Boot, which overrides the logging level we set in AgentTestRunner. I added a comment to clarify.

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
anuraaga: Can you add to the PR description what happened?

trask: 👍 done, it's a weird one, i can add more details if it is unclear

anuraaga: Hmm - not too clear to me still but not sure if it ever would. Main question for me is why we didn't need that before all of our package renaming. Is it because instead of `io.opentelemetry.javaagent` we switched to the more generic `io.opentelemetry`?

trask: oh ya, scroll up just a bit, this code is inside of a conditional `if (log.isDebugEnabled()) {` which wasn't triggered until I fixed the debug logging in `AgentTestRunner`

trask: as far as why we didn't need `RedefinitionStrategy.DiscoveryStrategy.Reiterating`, it's probably still not needed in prod situations because there aren't many classes already loaded when we install the agent, but it's different in tests where we apply instrumentation to some already loaded classes. in particular we already load guava classes, which is what caused the test issue with the Cassandra ListenableFuture test

PR #1199, opened 2020-09-14
Aws instrumentation - dynamo db statements

### LINKED ISSUE #1198: AWS instrumentation - DynamoDB statements
Follow-up for #1181

The main goal is to add statement support for DynamoDB AWS SDK requests.

kubawach:
Resolves #1198 
Split from #1191

Current assumptions:
- statement should be captured 
- PII should be maintained
- AWS SDK contains multiple legacy attributes that would clutter the output if captured
- Java toString() should not be used (not useful)


anuraaga:
> AWS SDK contains multiple legacy attributes that would clutter the output if captured

Could you give some examples of these? I'm still leaning towards generic serialization of scrubbed json because

- Most complete
- Users that care about statement are likely more familiar with DDB json than a custom string format we create?
- less hard coded implementation

This would have a disadvantage of logging legacy attributes that have no value but not clear to me how big of a deal it is.

trask:
If it makes sense, would be nice for statement scrubbing and formatting to be similar to mongodb instrumentation since those are also json-based queries

anuraaga:
I have finally composed the list of AWS SDK semantic conventions that AWS has defined

https://github.com/anuraaga/aws-opentelemetry-java-instrumentation/blob/master/docs/aws-sdk-semantic-conventions.md

These have been collected over the years while communicating with customers to find what can be useful for them. So hopefully they're a solid set of attributes and we do hope that all OTel AWS SDK client instrumentation implements these so all tracing users can take advantage of them.

These conventions capture interesting pieces of most DynamoDB requests - do you think we still need statement if we populate these? BTW, I hadn't looked well at the conventions until now finally writing this file up or I may have noticed the stronger duplication between statements and them earlier.

We use this JSON to define the mapping for use from SDKs (will need to be tweaked for some naming differences in my OTel version)

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/resources/com/amazonaws/xray/interceptors/DefaultOperationParameterWhitelist.json

and that allows all the languages to use similar, generic mapping logic to fill them in.

kubawach:
FYI - updated this PR to newest master (base for DynamoDB instrumentation). This is still "old" version with manually populated statement. I'll have a look at the mongo impl as pointed out by @trask and sem conventions as mentioned by @anuraaga

iNikem:
Closing this PR as inactive. @kubawach feel free to reopen if you return to this :)

PR #1200, opened 2020-09-14
Implement MDC auto-instrumentation for log4j2

### LINKED ISSUE #1196: Implement MDC auto-instrumentation for log4j2
The aim of this issue is to implement auto-instrumentation that sets `traceId`, `spanId` and `traceFlags` for each log produced by log4j2. Ideally `OpenTelemetryContextDataProvider` should be used.

mateuszrzeszutek:
Resolves #1196


README.md
```diff
@@ -204,6 +204,7 @@ provide the path to a JAR file including an SPI implementation using the system
 | [khttp](https://khttp.readthedocs.io)                                                                                                 | 0.1+                           |
 | [Kubernetes Client](https://github.com/kubernetes-client/java)                                                                        | 7.0+                           |
 | [Lettuce](https://github.com/lettuce-io/lettuce-core)                                                                                 | 4.0+                           |
+| [Log4j 2](https://logging.apache.org/log4j/2.x/)                                                                                      | 2.13.2+                        |
```
iNikem, line 207: We generally try to support as old version as possible. Are there some hard reasons why we cannot decrease this version?

iNikem: A, I see. Our current manual instrumentation uses a recent SPI. But if we go to bytecode level, can we actually provide support for older versions?

mateuszrzeszutek: > But if we go to bytecode level, can we actually provide support for older versions?

Yes, we can. I can add a separate instrumentation and instrument all `ContecxtDataInjector` implementations and support as low as 2.7 (which is 4 years old). If we want to go lower it'd probably turn out to be more complicated.

mateuszrzeszutek: Done.

instrumentation/log4j/log4j-2.7/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_7/Log4j27Instrumentation.java
```diff
@@ -0,0 +1,86 @@
...
+
+      if (contextData.containsKey("traceId")) {
+        // Assume already instrumented event if traceId is present.
+        return;
+      }
+
+      StringMap newContextData = new SortedArrayStringMap(contextData);
+      newContextData.putValue("traceId", currentContext.getTraceId().toLowerBase16());
+      newContextData.putValue("spanId", currentContext.getSpanId().toLowerBase16());
+      newContextData.putValue("traceFlags", currentContext.getTraceFlags().toLowerBase16());
```
mateuszrzeszutek, line 82: By the way, what do you think about adding some logging constant class in `auto-api`?

iNikem: If they are used in multiple places, why not

trask:
Hey @mateuszrzeszutek, have you seen the log4j2 instrumentation we used to have (#803)?

It was capturing the logging data instead of injecting MDC values, but we will be bringing it back once we have OpenTelemetry logging API.

The reason I mention, is we could potentially use the same instrumentation points to inject the MDC values.

This is what @anuraaga did in #1180 for logback.

A couple advantages to this approach would be to support older log4j2 versions, and potentially share instrumentation points for both MDC and logging API capture in the future.

anuraaga:
With logback, the need to modify the XML to reuse library instrumentation made it impractical to reuse in the auto. I'm hoping for log4j though we actually can just use the library as is.


I was thinking if doing it using the technique in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1172 (actually log4j is the reason I created that PR but am using it first in AWS sdk since closer to home :) ).

To support older versions, I would add library instrumentation that uses the old context SPI of log4j and use that in auto for the older versions.

Does this make sense?

iNikem:
Is there an older SPI for old Log4J versions? Including Log4J 1.x?

anuraaga:
Yeah there's this old one

https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/ContextDataInjector.html

It's still only 2.7+ so indeed it still doesn't provide full coverage so we may not like this option anyways.

I don't think any approach can apply to both 2.x and 1.x directly, they're very different APIs FWIU.

iNikem:
I did not mean common SPI for 2.x and 1.x :) I was wondering if there is _any_ SPI for 1.x.

For 2.x I am comfortable atm to only support 2.7+

anuraaga:
Ah got it - yeah not aware of any SPI for 1.x unfortunately

anuraaga:
Merged #1172 so if we want to use the SPI directly in auto instrumentation, we can :)

mateuszrzeszutek:
> Merged #1172 so if we want to use the SPI directly in auto instrumentation, we can :)

Thanks! Just injecting the service file looks much better.

> Hey @mateuszrzeszutek, have you seen the log4j2 instrumentation we used to have (#803)?
> It was capturing the logging data instead of injecting MDC values, but we will be bringing it back once we have OpenTelemetry logging API.
> The reason I mention, is we could potentially use the same instrumentation points to inject the MDC values.

I thought about it but I prefer using log4j's built in SPI capabilities in this case - we're using log4j types that were introduced exactly for the purpose of context data injection, so the instrumentation code looks quite simple and clean (esp. in case of 2.13.2+). Compared to that, previous `AbstractLogger`-using one looks much more complicated and fickle - how many `log` and `logMessage` does a logger have and which one is the correct one?

I believe that it'd probably be cleaner to have a separate instrumentation for purposes of OTel logging API - and if not let's refactor it then. We don't really know the final shape of it yet anyway.

instrumentation/log4j/log4j-2.7/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_7/Log4j27Instrumentation.java
mateuszrzeszutek: Done.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/log/MdcConstants.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.instrumentation.api.log;
+
+/** This class contains several constants used in logging libraries' MDC instrumentations. */
+public final class MdcConstants {
```
anuraaga, line 20: MDC is a name used by slf4j but these are more generic, how about `LoggingContextConstants` or `LoggingTags`?

instrumentation/log4j/log4j-2.7/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_7/Log4j27MdcInstrumentation.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    return not(hasClassesNamed("org.apache.logging.log4j.core.util.ContextDataProvider"));
+  }
+
+  @Override
+  public ElementMatcher<? super TypeDescription> typeMatcher() {
+    return implementsInterface(named("org.apache.logging.log4j.core.ContextDataInjector"));
```
anuraaga, line 56: Generally looking good - is there a reason to use bytecode instrumentation here instead of the SPI?

anuraaga: Oops - now I remember this wasn't even SPI >< https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/impl/ContextDataInjectorFactory.html

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/log/MdcConstants.java
mateuszrzeszutek: Good point, done.

instrumentation/log4j/log4j-2.7/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_7/Log4j27MdcInstrumentation.java
mateuszrzeszutek: Yeah, unfortunately it's not an SPI - but it's still a nice, single responsibility interface for context data injection.

instrumentation/log4j/log4j-2-testing/src/main/groovy/Log4j2Test.groovy
```diff
@@ -14,24 +14,22 @@
...
 import io.opentelemetry.trace.TracingContextUtils
 import org.apache.logging.log4j.LogManager
-import org.apache.logging.log4j.Logger
-import spock.lang.Specification
-
-class Log4j2Test extends Specification {
-
-  private static final Logger logger = LogManager.getLogger("TestLogger")
 
+abstract class Log4j2Test extends AgentTestRunner {
```
anuraaga, line 24: If I'm reading correctly, this is the base for both library and auto instrumentation. In that case it can't extend `AgentTestRunner` which executes auto instrumentation. We `AgentTestTrait` and `InstrumentationTestTrait` instead, do those work?

anuraaga (commented):
Looks swell, thanks!

instrumentation/log4j/log4j-2-testing/src/main/groovy/Log4j2Test.groovy
mateuszrzeszutek: It shouldn't execute auto instrumentation because it's not on classpath, but using those traits is a good idea anyway -- no need to run the agent if it's unnecessary. I'll fix that.

anuraaga (approved):
Thanks!

instrumentation/log4j/log4j-2.7/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_7/Log4j27MdcInstrumentation.java
```diff
@@ -0,0 +1,79 @@
...
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      "io.opentelemetry.instrumentation.auto.log4j.v2_7.SpanDecoratingContextDataInjector"
+    };
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    return not(hasClassesNamed("org.apache.logging.log4j.core.util.ContextDataProvider"));
```
iNikem, line 53: This is too broad matcher. The goal of `classLoaderMatcher` is to filter out as much classloaders as possible and not to match every class in it.

Add `hasClasssName("org.apache.logging.log4j.core.impl.ContextDataInjectorFactory")` or similar.

iNikem:
`log4j/log4j-2.13.2/library/README.md` needs update, et least `traceFlags` -> `sampled`

instrumentation/log4j/log4j-2.13.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_13_2/Log4j2MdcInstrumentation.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    return hasClassesNamed("org.apache.logging.log4j.core.util.ContextDataProvider");
+  }
+
+  @Override
+  public ElementMatcher<? super TypeDescription> typeMatcher() {
+    return named("org.apache.logging.log4j.core.impl.ThreadContextDataInjector");
```
iNikem, line 43: Change to `org.apache.logging.log4j.core.util.ContextDataProvider` for consistency?

iNikem (commented):
Why do we need both instrumentation for 2.7  and for 2.13.2? Cannot we just use one for all version 2.7+?

instrumentation/log4j/log4j-2.7/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_7/Log4j27MdcInstrumentation.java
mateuszrzeszutek: Done.

instrumentation/log4j/log4j-2.13.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_13_2/Log4j2MdcInstrumentation.java
mateuszrzeszutek: Unfortunately the instrumentation does not work if I use `ContextDataProvider` here: I suspect that it's because it's an interface.

mateuszrzeszutek:
> log4j/log4j-2.13.2/library/README.md needs update, et least traceFlags -> sampled

True, I missed that. Thanks!

> Why do we need both instrumentation for 2.7 and for 2.13.2? Cannot we just use one for all version 2.7+?

We could use a single instrumentation for all 2.7+ versions - I just thought that it would be a bit more elegant to use built-in SPI if it's possible.

instrumentation/log4j/log4j-2.13.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_13_2/Log4j2MdcInstrumentation.java
iNikem: Then at least add the comment

iNikem (approved):

anuraaga:
I don't know if there is a performance difference between the two, if not it could make sense for auto instrumentation to target 2.7.

But for library instrumentation, the 2.7 approach is even worse than gRPC SPI by requiring a system property. I have personally felt the annoyance before and appreciated they fixed the SPI - I think we do need the library instrumentation for 2.13 for those users.

iNikem:
Agree that library instrumentation can stay as it is, for 2.13.2+ only. And now thinking about it... both ways for auto instrumentation are Ok: 1 or 2 instrumentations. @mateuszrzeszutek what do you prefer?

instrumentation/log4j/log4j-2.13.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_13_2/Log4j2MdcInstrumentation.java
```diff
@@ -40,6 +40,8 @@ public Log4j2MdcInstrumentation() {
 
   @Override
   public ElementMatcher<? super TypeDescription> typeMatcher() {
+    // need to instrument a class, not an interface (ContextDataProvider) - otherwise helper classes
```
iNikem, line 43: "match", not "instrument" :) As you say yourself in `transformers`: "Nothing to instrument"

mateuszrzeszutek:
> Agree that library instrumentation can stay as it is, for 2.13.2+ only. And now thinking about it... both ways for auto instrumentation are Ok: 1 or 2 instrumentations. @mateuszrzeszutek what do you prefer?

I'm fine with both solutions actually. If anybody here has a strong preference towards one of them I'll change this PR accordingly.

instrumentation/log4j/log4j-2.13.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_13_2/Log4j2MdcInstrumentation.java
mateuszrzeszutek: That's true 🤦 

Little things like that are the easiest to miss 😄

instrumentation/log4j/log4j-2.13.2/auto/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v2_13_2/Log4j2MdcInstrumentation.java
trask: Interfaces are ok.

The problem with using `org.apache.logging.log4j.core.util.ContextDataProvider` is that then helper _class_ injection is performed during the loading (and transformation) of `ContextDataProvider`, which is a problem because one of the classes that we inject implements this same interface, causing the interface to be loaded while it's being transformed, which leads to duplicate class definition error after the interface is transformed and the triggering class loader tries to load it.

It seems this will be a common problem with injecting SPI resources, which is too bad, we'll need to be careful to pick type for the `typeMatcher` in these cases that we know will get loaded before SPI is triggered

mateuszrzeszutek: 😮 Thanks for the explanation, it makes much more sense now!

I've replaced my comment with your explanation.

anuraaga:
I'm fine with either too :) I'll let @iNikem merge

PR #1203, opened 2020-09-15
Add configuration for error prone / nullaway to allow opting in packa…

anuraaga:
…ge by package. Apply to armeria-library

I talk about nullaway, so I should go ahead and actually add it :P

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -323,7 +323,7 @@ plugins.withId('net.ltgt.errorprone') {
         // Doesn't work well with Java 8
         disable("FutureReturnValueIgnored")
 
-        option("NullAway:AnnotatedPackages", "io.opentelemetry,com.linecorp.armeria")
+        option("NullAway:AnnotatedPackages", "io.opentelemetry,com.linecorp.armeria,com.google.common")
       }
     }
   }
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -100,6 +100,8 @@ repositories {
 }
 
 dependencies {
+  compileOnly group: 'org.checkerframework', name: 'checker-qual', version: versions.checkerFramework
+
   testImplementation enforcedPlatform(group: 'org.junit', name: 'junit-bom', version: versions.junit5)
   testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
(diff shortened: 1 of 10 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -40,6 +40,7 @@
 import java.net.URISyntaxException;
 import java.util.Iterator;
 import java.util.List;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
iNikem, line 43: Does IDEA recognise this annotation?

gradle/java.gradle
```diff
@@ -305,4 +307,24 @@ tasks.withType(Test).configureEach {
...
+
+plugins.withId('net.ltgt.errorprone') {
+  dependencies {
+    annotationProcessor group: "com.uber.nullaway", name: "nullaway", version: versions.nullaway
+    errorprone group: "com.google.errorprone", name: "error_prone_core", version: versions.errorprone
+  }
+
+  tasks.withType(JavaCompile) {
+    if (!name.toLowerCase().contains("test")) {
+      options.errorprone {
```
iNikem, line 320: My experience with errorprone from Java repo is very negative. I hope it will not became such an annoying asshole as it is over there.

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
anuraaga: Yeah - IDEA (and every tool I've seen) recognizes any annotation named `Nullable` regardless of package :)

gradle/java.gradle
anuraaga: You'll see I'm already fairly liberal in disabling checks that are annoying :) There are some checks of questionable value that we can disable when we run into them, on the flip side I have had a few bad bugs caught by it before so do like it in general. Worst case, we'll disableAllChecks to only have nullness checking activated.

iNikem (approved):

PR #1204, opened 2020-09-15
Force flush after finishing lambda function.

anuraaga:
Currently, there is no way to export traces reliably from a lambda function short of flushing before returning because the VM freezes as soon as the function returns. For example, if there is only a single request made, that trace will basically never be visible, until another request comes in to unfreeze the VM, possibly hours later. This makes tracing mostly impossible without flushing. If I `wrk` against a lambda endpoint sending constant traffic, I see traces reliably, but otherwise, the gRPC export just never gets a chance to finish since it only has the function time slice to be able to make any progress.

However, for VMs that don't freeze much because they serve high traffic consistently, this flushing is wasteful. @wangzlei will be working on making this flush smarter, probably keeping track of invocation times and disabling autoflush when we notice there is not much freezing happening. Initially I'd like to add this simple force flush so OTel is actually usable with lambda, even if it adds significant overhead per invocation.


iNikem:
I am **very** concerned about performance implications of that. Will _anybody_ use otel with lambda given such overhead?

anuraaga:
I am hopeful that we will find a good adaptive flush based on freeze time that has low overhead on the case where performance matters more, low freeze time, and otherwise overhead could be acceptable for a low QPS function. This workaround is only until a better mechanism for async tasks happens in lambda.

Anyways the only real way we can know if anyone will use it is to try, and get input from users. With traces not showing up we can't even try and see how it goes.

anuraaga:
I'm more concerned about muzzle :) I don't know if it's safe to inject OTel SDK classes into the app. But if not, the only way to flush from auto instrumentation may be to add forceFlush to the API? 😨

anuraaga:
For a bit more color, I also had the thought that a big use case for lambda is processing message queue messages. In this case, since no one is waiting for a response, there is no visible overhead even with force flush. We get tons of requests for async so I'm actually aiming to add Instrumentation for that case before http even.

iNikem (approved):

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
```diff
@@ -53,4 +55,8 @@ public Scope startScope(Span span) {
   protected String getInstrumentationName() {
     return "io.opentelemetry.aws-lambda-1.0";
   }
+
+  public void forceFlush() {
+    OpenTelemetrySdk.getTracerProvider().forceFlush().join(1, TimeUnit.SECONDS);
```
trask, line 60: The OpenTelemetry SDK is in the agent class loader, so won't be accessible directly from auto-instrumentation.

Unless we want to shade and put it in the bootstrap class loader.

I wonder if we will have other use cases like this.

For now(?), maybe add a `forceFlush()` method in `auto-api` (soon to be `javaagent-instrumentation-api`), and register the underlying implementation during agent startup (the implementation needs to live in the agent class loader so it can call the SDK directly).

anuraaga:
@trask Thanks for the hint - added `OpenTelemetrySdkAccess` in the bootstrap classloader to be our bridge.

trask (approved):
👍

Issue #1201, opened 2020-09-14
Logback MDC auto-instrumentation gets muzzled and does not work


mateuszrzeszutek:
I've prepared a simple console app that helps reproduce this issue: https://github.com/mateuszrzeszutek/test-console-app
I tested both 0.8.0 and master branch -- auto-instrumentation does not work on both of them.

When you run it with debug logs you'll see the following message:

```
[opentelemetry.auto.trace 2020-09-14 16:26:09:885 +0200] [main] DEBUG io.opentelemetry.javaagent.tooling.Instrumenter$Default - Instrumentation muzzled: [logback] -- io.opentelemetry.instrumentation.auto.logback.v1_0_0.LogbackInstrumentation on jdk.internal.loader.ClassLoaders$AppClassLoader@2c13da15
[opentelemetry.auto.trace 2020-09-14 16:26:09:885 +0200] [main] DEBUG io.opentelemetry.javaagent.tooling.Instrumenter$Default - -- io.opentelemetry.javaagent.shaded.instrumentation.logback.v1_0_0.LoggingEventWrapper:100 Missing method getMarker#()Lio/opentelemetry/javaagent/slf4j/Marker;
```

and `traceId`, `spanId` and `traceFlags` won't be added to the MDC.

iNikem:
What happens if you remove sfl4j dependency and use the one brought by logback itself?

mateuszrzeszutek:
I think it still won't work - it shouldn't really matter which one you compile against as long as you relocate it later (shadowJar task):

```
  relocate 'org.slf4j', 'io.opentelemetry.javaagent.slf4j'
```

I think it probably needs the same treatment as OTel API -- i.e. prefixing and "unshadowing" packages:

```
  relocate "application.io.opentelemetry", "io.opentelemetry"
```

trask:
> I think it probably needs the same treatment as OTel API -- i.e. prefixing and "unshadowing" packages

👍

Another option (because I know @iNikem dislikes this option 😁), is to rework the logback instrumentation a bit to get rid of the `LoggingEventWrapper` class (which is what's pulling in the slf4j Marker dependency).

You could remove `LoggingEventWrapper`, and instead add a field to `ILoggingEvent` implementations in order to store the updated `mdcPropertyMap`:

  @Override
  public Map<String, String> contextStore() {
    return singletonMap("ch.qos.logback.classic.spi.ILoggingEvent", "java.util.Map");
  }

[...]

PR #1208, opened 2020-09-15
Logback MDC auto-instrumentation gets muzzled and does not work

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1201

### DIFF
--- a/instrumentation/logback/logback-1.0.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/logback/v1_0_0/LoggingEventInstrumentation.java
+++ b/instrumentation/logback/logback-1.0.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/logback/v1_0_0/LoggingEventInstrumentation.java
@@ -26,7 +26,7 @@ import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import com.google.auto.service.AutoService;
 import io.opentelemetry.instrumentation.auto.api.InstrumentationContext;
-import io.opentelemetry.instrumentation.logback.v1_0_0.UnionMap;
+import io.opentelemetry.instrumentation.logback.v1_0_0.internal.UnionMap;
 import io.opentelemetry.javaagent.tooling.Instrumenter;
 import io.opentelemetry.trace.Span;
 import io.opentelemetry.trace.SpanContext;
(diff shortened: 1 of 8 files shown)


instrumentation/logback/logback-1.0.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/logback/v1_0_0/LoggingEventInstrumentation.java
```diff
@@ -0,0 +1,106 @@
...
+            .and(takesArguments(0)),
+        LoggingEventInstrumentation.class.getName() + "$GetMdcAdvice");
+  }
+
+  public static class GetMdcAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(
+        @Advice.This ILoggingEvent event,
+        @Advice.Return(typing = Typing.DYNAMIC, readOnly = false) Map<String, String> contextData) {
+      if (contextData != null && contextData.containsKey("traceId")) {
```
mateuszrzeszutek, line 81: I'll replace `"traceId"` and others either here or in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1200 - depending on which one gets merged first.

instrumentation/logback/logback-1.0.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/logback/v1_0_0/LoggingEventInstrumentation.java
```diff
@@ -0,0 +1,106 @@
...
+        // Assume already instrumented event if traceId is present.
+        return;
+      }
+
+      Span currentSpan = InstrumentationContext.get(ILoggingEvent.class, Span.class).get(event);
+      if (currentSpan != null && !currentSpan.getContext().isValid()) {
```
trask, line 87: I think this:
```suggestion
      if (currentSpan == null || !currentSpan.getContext().isValid()) {
```

trask (commented):
Thanks 👍

iNikem (approved):

instrumentation/logback/logback-1.0.0/library/src/main/java/io/opentelemetry/instrumentation/logback/v1_0_0/UnionMap.java
```diff
@@ -29,14 +29,14 @@
  * An immutable view over two maps, with keys resolving from the first map first, or otherwise the
  * second if not present in the first.
  */
-final class UnionMap<K, V> extends AbstractMap<K, V> {
+public final class UnionMap<K, V> extends AbstractMap<K, V> {
```
anuraaga, line 32: Can you move this into a subpackage `internal`? We shouldn't add this to our public API

instrumentation/logback/logback-1.0.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/logback/v1_0_0/LoggingEventInstrumentation.java
mateuszrzeszutek: 🤦  - that's correct, that was a stupid mistake on my side.

instrumentation/logback/logback-1.0.0/library/src/main/java/io/opentelemetry/instrumentation/logback/v1_0_0/UnionMap.java
mateuszrzeszutek: Done, thanks!

anuraaga (approved):

PR #1210, opened 2020-09-16
Add instrumentation of SQS.

anuraaga:


anuraaga:
Sent this as a draft in advance to refer to from spec issue, no need to even start reviewing in detail yet (unless you're interested :) )

anuraaga:
Finished it up with my understanding of the current spec. I expect some changes in the spec as working through scenarios in https://github.com/open-telemetry/opentelemetry-specification/issues/963 but I think this is a pretty good first version.

instrumentation/aws-lambda-1.0/auto/aws-lambda-1.0-auto.gradle
```diff
@@ -21,6 +21,7 @@ muzzle {
     group = "com.amazonaws"
     module = "aws-lambda-java-core"
     versions = "[1.0.0,)"
+    extraDependency('com.amazonaws:aws-lambda-java-events:2.2.1')
```
anuraaga, line 24: I wonder if I should instead of a separate instrumentation module, `aws-lambda-sqs-2.2.1` due to this extra, optional dependency. Not quite sure where we want to set boundaries for highly related instrumentation like this case, let me know any thoughts.

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AwsLambdaRequestHandlerInstrumentation.java
```diff
@@ -53,12 +56,28 @@
...
+
+    // First instrument the function invocation itself.
+    transformers.put(
         isMethod()
             .and(isPublic())
             .and(named("handleRequest"))
             .and(takesArgument(1, named("com.amazonaws.services.lambda.runtime.Context"))),
         AwsLambdaRequestHandlerInstrumentation.class.getName() + "$HandleRequestAdvice");
+
+    // Then instrument message handling if appropriate.
```
anuraaga, line 71: If I separated out a different module, though, not sure it'd be even possible to do this pattern.

iNikem: I don't like this strict dependency between 2 instrumentations :(

instrumentation/aws-lambda-1.0/library/src/test/groovy/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaSqsMessageHandlerTest.groovy
```diff
@@ -0,0 +1,108 @@
...
+            "${SemanticAttributes.MESSAGING_SYSTEM}" "AmazonSQS"
+            "${SemanticAttributes.MESSAGING_OPERATION}" "process"
+            "${SemanticAttributes.MESSAGING_MESSAGE_ID}" "message1"
+            "${SemanticAttributes.MESSAGING_DESTINATION}" "queue1"
+          }
+          hasLink("5759e988bd862e3fe1be46a994272793", "53995c3f42cd8ad8")
+        }
+        span(3) {
+          operationName("queue1 process")
+          spanKind CONSUMER
```
iNikem, line 95: So, we can have nested CONSUMER spans? But not CLIENT?

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AwsLambdaRequestHandlerInstrumentation.java
anuraaga: I think it's similar to our servlet instrumentation - it seems to happen when there's an overall framework, with subframeworks built on top of it. Is it that bad?

instrumentation/aws-lambda-1.0/library/src/test/groovy/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaSqsMessageHandlerTest.groovy
anuraaga: Yeah our spec currently has an example of it

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-receiving

Since CONSUMER / PRODUCER aren't parent/child, I guess 1:1 doesn't matter as much. So in https://github.com/open-telemetry/opentelemetry-specification/issues/963 I'm also trying to clarify whether those kinds have any meaning other than INTERNAL anyways.

I'm happy to keep this instrumentation up to date as spec is updated, which I'm sure it will be :)

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AwsLambdaRequestHandlerInstrumentation.java
iNikem: In general inter-instrumentation dependencies are Ok, I guess. But in this case they apply to the same method and the order is important. This seems fragile

trask: i'm ok with current approach

the other option I see is to combine into a single advice, e.g.

```
  public static class HandleEventAdvice {
    public static void onEnter(
        @Advice.Argument(value = 0, typing = Typing.DYNAMIC) Object arg,
        @Advice.Argument(1) Context context,
        @Advice.Local("otelFunctionSpan") Span functionSpan,
        @Advice.Local("otelFunctionScope") Scope functionScope,
        @Advice.Local("otelMessageSpan") Span messageSpan,
        @Advice.Local("otelMessageScope") Scope messageScope) {
      functionSpan = FUNCTION_TRACER.startSpan(context, Kind.SERVER);
      functionScope = FUNCTION_TRACER.startScope(functionSpan);
      if (arg instanceof SQSEvent) {
        messageSpan = MESSAGE_TRACER.startSpan(context, (SQSEvent) arg);
        messageScope = MESSAGE_TRACER.startScope(messageSpan);
      }
    }

    public static void stopSpan(
        @Advice.Thrown Throwable throwable,
        @Advice.Local("otelFunctionSpan") Span functionSpan,
        @Advice.Local("otelFunctionScope") Scope functionScope,
        @Advice.Local("otelMessageSpan") Span messageSpan,
        @Advice.Local("otelMessageScope") Scope messageScope) {

      if (messageScope != null) {
        messageScope.close();
        if (throwable != null) {
          MESSAGE_TRACER.endExceptionally(messageSpan, throwable);
        } else {
          MESSAGE_TRACER.end(messageSpan);
        }
      }

      functionScope.close();
      if (throwable != null) {
        FUNCTION_TRACER.endExceptionally(functionSpan, throwable);
      } else {
        FUNCTION_TRACER.end(functionSpan);
      }
      OpenTelemetrySdkAccess.forceFlush(1, TimeUnit.SECONDS);
    }
  }
```

anuraaga: Thanks this is better!

instrumentation/aws-lambda-1.0/auto/src/main/java/io/opentelemetry/instrumentation/auto/awslambda/v1_0/AbstractAwsLambdaInstrumentation.java
```diff
@@ -28,7 +28,10 @@ public AbstractAwsLambdaInstrumentation() {
   public final String[] helperClassNames() {
     return new String[] {
       packageName + ".AwsLambdaInstrumentationHelper",
-      "io.opentelemetry.instrumentation.awslambda.v1_0.AwsLambdaTracer"
+      "io.opentelemetry.extensions.trace.propagation.AwsXRayPropagator",
```
trask, line 31: I think can remove this now:
```suggestion
```

trask (approved):
nice, I'm glad that worked out to require user to configure the xray propagator, makes this instrumentation feel a lot cleaner

instrumentation/aws-lambda-1.0/auto/aws-lambda-1.0-auto.gradle
trask: I think it makes sense here to add muzzle requirement on `aws-lambda-java-events` also, e.g.

```
  pass {
    group = "com.amazonaws"
    module = "aws-lambda-java-events"
    versions = "[2.2.1,)"
    extraDependency('com.amazonaws:aws-lambda-java-core:1.0.0')
  }
```

Issue #1197, opened 2020-09-14
Implement MDC auto-instrumentation for log4j1


mateuszrzeszutek:
The aim of this issue is to implement auto-instrumentation that sets `traceId`, `spanId` and `traceFlags` for each log produced by log4j1.

PR #1211, opened 2020-09-16
Implement MDC auto-instrumentation for log4j1

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1197


instrumentation/log4j/log4j-1.2/log4j-1.2.gradle
```diff
@@ -0,0 +1,24 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    group = "log4j"
+    module = "log4j"
+    versions = "[1.2,)"
```
iNikem, line 7: Should we cap that on 2.0? Or it has different module?

instrumentation/log4j/log4j-1.2/src/test/groovy/Log4j1MdcTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+    events[0].getMDC("sampled") == "true"
+
+    events[1].message == "log message 2"
+    events[1].getMDC("traceId") == null
+    events[1].getMDC("spanId") == null
+    events[1].getMDC("sampled") == null
+
+    events[2].message == "log message 3"
+    // this explicit getMDCCopy() call here is to make sure that whole instrumentation is tested
+    events[2].getMDCCopy()
```
iNikem, line 83: Why don't you assert the result of this method call?

trask (approved):

instrumentation/log4j/log4j-1.2/log4j-1.2.gradle
trask: ya it's a different module, we could cap, but nice to leave unbounded where possible

instrumentation/log4j/log4j-1.2/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v1_2/Log4j1Instrumentation.java
```diff
@@ -0,0 +1,78 @@
...
+
+@AutoService(Instrumenter.class)
+public class Log4j1Instrumentation extends Instrumenter.Default {
+  public Log4j1Instrumentation() {
+    super("log4j1", "log4j");
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    return hasClassesNamed("org.apache.log4j.MDC");
```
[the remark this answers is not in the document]
anuraaga: Also just to check, when the type matcher is a simple `named` we'd usually leave out `classLoaderMatcher` @trask?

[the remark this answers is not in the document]
trask: oh yes, you're right 👍

instrumentation/log4j/log4j-1.2/src/test/groovy/Log4j1MdcTest.groovy
mateuszrzeszutek: `getMDCCopy()`'s return type is `void`, contrary to its name it does not actually get anything. It just initialises the `mdcCopy` field.

instrumentation/log4j/log4j-1.2/src/main/java/io/opentelemetry/instrumentation/auto/log4j/v1_2/Log4j1Instrumentation.java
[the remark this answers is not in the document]
mateuszrzeszutek: OK, I've removed it altogether.

By the way, it would be great if `classLoaderMatcher()`'s Javadoc contained a short explanation of how, why and when we should use it; some common pattern etc. I'd add it but I don't feel confident about being able to explain it correctly 😅

iNikem (approved):

PR #1214, opened 2020-09-17
Review semantic convention for Http Client spans

iNikem:


instrumentation/google-http-client-1.19/src/test/groovy/AbstractGoogleHttpClientTest.groovy
```diff
@@ -83,6 +83,7 @@ abstract class AbstractGoogleHttpClientTest extends HttpClientTest {
             "${SemanticAttributes.HTTP_URL.key()}" "${uri}"
             "${SemanticAttributes.HTTP_METHOD.key()}" method
             "${SemanticAttributes.HTTP_STATUS_CODE.key()}" 500
+            "${SemanticAttributes.HTTP_FLAVOR.key()}" '1.1'
```
trask, line 86:
```suggestion
            "${SemanticAttributes.HTTP_FLAVOR.key()}" "1.1"
```

instrumentation/java-httpclient/src/test/groovy/JdkHttpClientTest.groovy
```diff
@@ -52,4 +55,45 @@ abstract class JdkHttpClientTest extends HttpClientTest {
   boolean testCircularRedirects() {
     return false
   }
+
+  //We override this test below because it produces somewhat different attributes
```
trask, line 59: is it only flavor that's different?

instrumentation/java-httpclient/src/main/java/io/opentelemetry/instrumentation/auto/httpclient/JdkHttpClientTracer.java
```diff
@@ -66,10 +70,22 @@ protected String requestHeader(HttpRequest httpRequest, String name) {
...
     return httpResponse.headers().firstValue(name).orElse(null);
   }
 
+  @Override
+  protected Span onResponse(Span span, HttpResponse<?> httpResponse) {
+    span = super.onResponse(span, httpResponse);
+
+    if (httpResponse != null) {
+      SemanticAttributes.HTTP_FLAVOR.set(
+          span, httpResponse.version() == Version.HTTP_1_1 ? "1.1" : "2.0");
```
trask, line 83: interesting, makes sense for a negotiated connection, where we don't know flavor ahead of time

instrumentation/apache-httpasyncclient-4.0/src/main/java/io/opentelemetry/instrumentation/auto/apachehttpasyncclient/ApacheHttpAsyncClientTracer.java
```diff
@@ -46,6 +47,11 @@ protected String method(HttpRequest request) {
     }
   }
 
+  @Override
+  protected @Nullable String flavor(HttpRequest httpRequest) {
```
trask, line 51: I like this placement (type annotation) 👍

trask (approved):

anuraaga (approved):

instrumentation/java-httpclient/src/test/groovy/JdkHttpClientTest.groovy
iNikem: Yes

instrumentation/apache-httpasyncclient-4.0/src/main/java/io/opentelemetry/instrumentation/auto/apachehttpasyncclient/ApacheHttpAsyncClientTracer.java
iNikem: I don't :) It is IDEA's doing, so I have to accept :(

anuraaga: Reformat code puts the annotation there? I haven't seen that before. I agree with @iNikem that for normal methods, annotation above looks a bit better mostly due to how annotations end up getting colored in IDE GitHub etc. @iNikem Is it not worth manually fixing these?

iNikem: Not reformat but "implement missing method", I think.

I don't want to fix this manually every time IDEA generates code for me. I can do it here, but it will not be sustainable

trask: ya, i like type annotations, but they are not common. i'm very surprised Intellij does the "right thing" 😉

Issue #717, opened 2020-07-17
Play 2.3 and 2.4 instrumentations are almost identical


iNikem:
Investigate if these two instrumentations can be merged into single one. An extra bonus if 2.6 can be merged as well.

PR #1215, opened 2020-09-17
[WIP] Play 2.8 Support 

easel:
This is an initial attempt at direct support for Play 2.8. As it stands, it compiles, but the tests fail due to some sort of incompatibility between the gradle proxies and the test routes. I'm pretty familiar with play, but much less so with gradle and groovy. 

I'd welcome any ideas as to how to get one of these tests to pass!

```
No signature of method: play.routing.RoutingDsl$PathPatternMatcher.routingAsync() is applicable for argument types: (com.sun.proxy.$Proxy40) values: [server.PlayAsyncServerTest$_startServer_closure1$_closure2@3bd2f258]
```


easel:
After seeing https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/717 I'm not entirely sure if this is necessary or not. It's definitely necessary in order to test against 2.8, but the existing 2.6 instrumentation seems like it might be working, although it seems to overlap with akka http a bit....

iNikem:
@easel Do you plan on continuing your work on this draft? Do you need any help from us? I am trying to clean up a list of open PR a little bit...

easel:
@iNikem I'm willing to continue working on it if it's generally considered to be valuable. The play2.6 instrumentation does seem to pick up my routes, although akka http also creates spans which is a little strange -- would we really want both, generally?

Some help would also be very welcome. Two things come to mind.

1. If there is anybody still active in the project that has a strong understanding of the existing play instrumentation, i'd definitely appreciate some more context on what might be broken with the 2.6 instrumentation and would need to be fixed.

2. I'm not very familiar with gradle and am struggling with figuring out how to adapt the test mocks for the routes under play 2.8. Some hints or feedback from somebody comfortable with the tests would also be very helpful.

Thanks!

trask:
hey @easel, sorry we let this drift, I'm going to close it, but if you want to pursue it just let us know and we'll figure out a way to help you forward.

> If there is anybody still active in the project that has a strong understanding of the existing play instrumentation

Unfortunately no, and we haven't seen much user demand for play support (yet), so we are sort of waiting for someone who wants to pick it up and dig into it 😉

> I'm not very familiar with gradle and am struggling with figuring out how to adapt the test mocks for the routes under play 2.8. Some hints or feedback from somebody comfortable with the tests would also be very helpful.

This we can definitely help out with. We have [office hours every Thu at 6p](https://calendar.google.com/event?action=TEMPLATE&tmeid=NmhybHF2djhtYnM4N2J1NGE4YWpjNTltdGhfMjAyMTA5MjRUMDEwMDAwWiBnb29nbGUuY29tX2I3OWUzZTkwajdiYnNhMm4ycDVhbjVsZjYwQGc&tmsrc=google.com_b79e3e90j7bbsa2n2p5an5lf60%40group.calendar.google.com&scp=ALL) or we can discuss async in the [slack channel](https://cloud-native.slack.com/archives/C0150QF88FL).

PR #1227, opened 2020-09-21
Add smoke test for Logback MDC instrumentation

### LINKED ISSUE #1209: Smoke test for Logback MDC instrumentation
The logback instrumentation is fragile (#1201) because logback depends on slf4j which we shade during our build, and our normal tests don't currently catch shading issues.

We are hoping to change how we run tests post-GA, so that they run using a real (and shaded) javaagent, which would catch these issues.

For now though, it maybe worth adding a smoke test to make sure this doesn't accidentally break again without us knowing about it.

mateuszrzeszutek:
Resolves #1209

Needs to be merged after changes from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1228 are deployed

### DIFF
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
@@ -16,6 +16,8 @@
 
 package io.opentelemetry.smoketest
 
+import static java.util.stream.Collectors.toSet
+
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.google.protobuf.util.JsonFormat
 import io.opentelemetry.auto.test.utils.OkHttpUtils
@@ -23,6 +25,7 @@ import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest
 import io.opentelemetry.proto.common.v1.AnyValue
 import io.opentelemetry.proto.trace.v1.Span
 import java.util.concurrent.TimeUnit
+import java.util.regex.Pattern
 import java.util.stream.Stream
 import okhttp3.OkHttpClient
 import okhttp3.Request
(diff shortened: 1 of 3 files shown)


smoke-tests/springboot/src/main/java/io/opentelemetry/smoketest/springboot/controller/WebController.java
```diff
@@ -17,13 +17,18 @@
...
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
 public class WebController {
+  private static final Logger LOGGER = LoggerFactory.getLogger(WebController.class);
+
   @RequestMapping("/greeting")
   public String greeting() {
+    LOGGER.info("HTTP request received");
```
iNikem, line 31: Tests fail because they don't build this application as part of the test. If you are sure that test work, you may submit two separate PRs: one for this app and another for the actual test

mateuszrzeszutek: I've tested it locally and it worked -- split into two PRs.
Thanks!

iNikem:
please rebase

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -172,4 +180,15 @@ abstract class SmokeTest extends Specification {
 
     return content
   }
+
+  protected static Set<String> getLoggedTraceIds(ToStringConsumer output) {
+    output.toUtf8String().lines()
+      .flatMap(SmokeTest.&findTraceId)
```
trask, line 186: oh groovy 😭

trask (approved):

PR #1231, opened 2020-09-22
Add otel prefix to excluded classes properties

jamal:
The documentation doesn't include the property prefix that is required to use this. The change just adds `otel.` and `OTEL_` prefix to the property table.


anuraaga (approved):
Thanks, sorry about that!

trask (approved):
👍

Issue #749, opened 2020-07-22
Review all latestDepTestLibrary versions that have an upper limit


trask:
`latestDepTestLibrary` implicitly uses the latest version available in maven central, e.g.

But sometimes we have to apply an upper limit for one reason or another.

E.g.

```
latestDepTestLibrary group: 'org.elasticsearch.client', name: 'transport', version: '5.+'
```

where the upper limit is needed because we have separate instrumentation for `6.0`.

And

```
latestDepTestLibrary group: 'io.lettuce', name: 'lettuce-core', version: '5.+'
```

where the upper limit is needed in order to avoid an alpha version `6.0.0.M1`.

And sometimes the upper limit is needed because we haven't fixed the tests or the instrumentation to run on the latest version.

It would be good to review all of these upper limits, and try to categorize these upper limits:

* Upper limit due to multi-versioned instrumentation
* Upper limit due to latest version being alpha (and don't want to deal with alpha API churn)
* Upper limit that should to be lifted

and decide on a plan for how to manage these going forward so the upper limit are not forgotten about.

PR #1236, opened 2020-09-22
Convert all remaining database client decorators to tracers

iNikem:


iNikem:
Hm, @trask is it normal and expected that `elasticsearch-transport-6.0.gradle` limits testing to versions <6.5, but muzzle checks (and fails in this PR) even 7+?

instrumentation/rxjava-1.0/src/main/java/io/opentelemetry/instrumentation/auto/rxjava/TracedOnSubscribe.java
```diff
@@ -16,56 +16,44 @@
...
-    if (parentSpan != null) {
-      spanBuilder.setParent(parentSpan);
-    }
-    Span span = spanBuilder.startSpan();
-
-    afterStart(span);
-
-    try (Scope scope = currentContextWith(span)) {
-      delegate.call(new TracedSubscriber(span, subscriber, decorator));
+    try (Scope ignored = ContextUtils.withScopedContext(parentContext)) {
```
trask, line 47: i think good to prefer not creating unnecessary scopes

though it would require a `BaseTracer.startSpan` overload that takes a parent Context (or Span)

instrumentation/rxjava-1.0/src/main/java/io/opentelemetry/instrumentation/auto/rxjava/TracedSubscriber.java
```diff
@@ -92,15 +92,12 @@ public void onCompleted() {
   public void onError(Throwable e) {
     Span span = spanRef.getAndSet(null);
     if (span != null) {
-      try (Scope scope = currentContextWith(span)) {
-        decorator.onError(span, e);
+      try (Scope ignored = currentContextWith(span)) {
         delegate.onError(e);
+        tracer.endExceptionally(span, e);
```
trask, line 97: before callback?

instrumentation/rxjava-1.0/src/main/java/io/opentelemetry/instrumentation/auto/rxjava/TracedSubscriber.java
```diff
@@ -92,15 +92,12 @@ public void onCompleted() {
...
       } catch (Throwable e2) {
-        decorator.onError(span, e2);
+        tracer.endExceptionally(span, e2);
```
trask, line 99: and remove this?

trask (commented):
sorry, lots of comments about pre-existing behavior, you can tell me to open issue(s) 😂

trask:
> Hm, @trask is it normal and expected that `elasticsearch-transport-6.0.gradle` limits testing to versions <6.5, but muzzle checks (and fails in this PR) even 7+?

this is not unheard of 😂

a lot of times it's the test code that needs reworking from version to version, e.g. harness stuff as opposed to the underlying instrumentation

also see #749:

> And sometimes the cap is needed because we haven't fixed the tests or the instrumentation to run on the latest version.

instrumentation/elasticsearch/elasticsearch-rest-5.0/src/latestDepTest/groovy/Elasticsearch6RestClientTest.groovy
```diff
@@ -96,14 +95,15 @@ class Elasticsearch6RestClientTest extends AgentTestRunner {
       trace(0, 2) {
         span(0) {
           operationName "GET _cluster/health"
-          spanKind INTERNAL
+          spanKind CLIENT
```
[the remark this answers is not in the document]
iNikem: It is. I think we should review that as part of #465

instrumentation/elasticsearch/elasticsearch-rest-5.0/src/main/java/io/opentelemetry/instrumentation/auto/elasticsearch/rest/v5_0/RestResponseListener.java
```diff
@@ -35,26 +35,22 @@ public RestResponseListener(ResponseListener listener, Span span) {
...
-      DECORATE.onResponse(span, response);
+      TRACER.onResponse(span, response);
     }
 
     try {
       listener.onSuccess(response);
     } finally {
-      DECORATE.beforeFinish(span);
-      span.end();
+      TRACER.end(span);
```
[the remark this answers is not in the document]
iNikem: I am not sure about that. We have an operation which finishes with `onSuccess` callback. Should span end before callback or after? I tend to think _after_

instrumentation/rxjava-1.0/src/main/java/io/opentelemetry/instrumentation/auto/rxjava/TracedOnSubscribe.java
iNikem: Yeah, exactly. I will just a TODO here for now and in the future I will review all `startSpan` methods anyway. They are kinda chaotic right now.

iNikem:
But if muzzle fails on later versions, should I then limit muzzle check range as well? And say that we don't actually support later versions?

anuraaga (approved):

trask (approved):
👍

Issue #1213, opened 2020-09-17
Allow vendor distribution to provide default configuration


iNikem:
Part of #566

As part of Splunk distribution of OpenTelemetry Java instrumentation we want to provide default values for some configuration options. Currently configuration can be provided in 3 ways:

- Env variable
- System property
- Via configuration file

Vendor distro has to provide default value only if none of the above provides it. As `Config` instance cannot be changed (setting reflection aside), the only way vendor can provide defaults now is to check all three of the above and, if value is not found, set e.g. system property. And to be careful to do all that before Config class has a chance to load. This is cumbersome.

Probably the most consistent way to provide better experience is to introduce yet another SPI ( :) ) for additional configuration sources. The order then should be "system properties -> environment -> config file -> SPI -> default values".

anuraaga:
SPI++

mateuszrzeszutek:
Can you assign this ticket to me?

jkwatson:
Is this something that belongs in the core SDK?

iNikem:
I am not sure. With manual instrumentation you can always configure stuff to your liking, no?

jkwatson:
> I am not sure. With manual instrumentation you can always configure stuff to your liking, no?

For sure. And, I'm happy with this starting in the auto-instrumentation project, and later migrating it to the core SDK if it seems to be more generally useful. Just wanted to make sure we'd thought about it. :)

iNikem:
Honestly, I haven't :) But I will tomorrow, when my brain is fresh

iNikem:
At last I had chance to think about it and yes, let us do it in instrumentation repo now.

PR #1243, opened 2020-09-23
Allow vendor distribution to provide default configuration

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1213

As an additional benefit, now all SDK `ConfigBuilder`s can use properties from both SPI and the `otel.trace.config` file (which they did not use before).

Needs to be merged after https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254

### DIFF
--- a/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/config/PropertySource.java
+++ b/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/config/PropertySource.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.opentelemetry.javaagent.spi.config;
+
+import java.util.Map;
+
+/**
+ * A service provider that allows to override default OTel agent configuration. Properties returned
+ * by implementations of this interface will be used after the following methods fail to find a
+ * non-empty property value: system properties, environment variables, properties configuration
+ * file.
+ */
+public interface PropertySource {
+  /**
+   * @return all properties whose default values are overridden by this property source. Key of the
+   *     map is the propertyName (same as system property name, e.g. {@code otel.exporter}), value
+   *     is the property value.
+   */
+  Map<String, String> getProperties();
+}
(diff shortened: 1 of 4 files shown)


trask (commented):
Hey @mateuszrzeszutek!

What do you think of turning `Config` into just a POJO (e.g. `@AutoValue`)? (e.g. still with static `get()` method to get the singleton instance)

And moving the construction of the `Config` object into `javaagent-tooling`?

Then wouldn't need to reference agent class loader from bootstrap class loader, which is where the complexity enters.

One other thought, it looks like the `ExporterConfig` arg to `SpanExporterFactory.fromConfig(ExporterConfig)` is basically unused, and if we replaced that with a `Properties` object (we can combine `propertiesFromConfigFile` and `propertiesFromSpi` before passing in), then the exporter modules wouldn't need to depend on `instrumentation-api` and `javaagent-tooling` which would be nice.

iNikem:
> One other thought, it looks like the `ExporterConfig` arg to `SpanExporterFactory.fromConfig(ExporterConfig)` is basically unused, and if we replaced that with a `Properties` object (we can combine `propertiesFromConfigFile` and `propertiesFromSpi` before passing in), then the exporter modules wouldn't need to depend on `instrumentation-api` and `javaagent-tooling` which would be nice.

If we pass just `Properties` into `SpanExporterFactory`, then we have to normalise keys there, right?

mateuszrzeszutek:
> What do you think of turning Config into just a POJO (e.g. @AutoValue)? (e.g. still with static get() method to get the singleton instance)
> And moving the construction of the Config object into javaagent-tooling?

That's a great idea!

> One other thought, it looks like the ExporterConfig arg to SpanExporterFactory.fromConfig(ExporterConfig) is basically unused, and if we replaced that with a Properties object (we can combine propertiesFromConfigFile and propertiesFromSpi before passing in), then the exporter modules wouldn't need to depend on instrumentation-api and javaagent-tooling which would be nice.

Hmm, one downside of this approach would be duplicating the configuration loading order in each exporter: merged-spi-and-file-props -> env variables -> system properties.
Alternatively we could probably convert env var names to system property names and merge them all together into a single `Properties` instance.

Anyway, that seems like a lot of changes -- do you think that we should do it in this PR?

iNikem:
A, `io.opentelemetry.sdk.common.export.ConfigBuilder` already does that... Yes, I think it is a very good idea, to move actual compilation of configuration from many difference sources into a separate class in `tooling` module and to expose simple Config/Properties to other components.

iNikem:
> Hmm, one downside of this approach would be duplicating the configuration loading order in each exporter: merged-spi-and-file-props -> env variables -> system properties.
> Alternatively we could probably convert env var names (as @iNikem mentioned) to system property names and merge them all together into a single `Properties` instance.

You read configuration once, normalize it (or more precise, you let `ConfigBuilder` do it for you) and then pass single instance of already normalised `Properties` to all other components.

mateuszrzeszutek:
Nice, reusing SDK `ConfigBuilder` for this is a neat idea.

mateuszrzeszutek:
I've pushed a PR that contains all those refactorings: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254
I'll rebase this one with it.

mateuszrzeszutek:
Rebased after merging the refactoring PR.

@trask @iNikem I'm just letting you know that this one is ready for review.

anuraaga (approved):
Sweet!

iNikem (approved):

trask (approved):

PR #1246, opened 2020-09-23
Add traceId to logging exporter

### LINKED ISSUE #1245: Log exporter should log traceId
Log exporter is logging only spanIds at the moment https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/912a159c19822b17ef5d1d356900d05d3d3f790a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java#L36.

It should also log traceId to allow log correlation.

pavolloffay:

Resolves #1245

### DIFF
--- a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
+++ b/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
@@ -33,7 +33,7 @@ public class LoggingExporter implements SpanExporter {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
-      System.out.print(prefix + " " + span.getName() + " " + span.getSpanId() + " ");
+      System.out.print(String.format("%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId()));
       span.getAttributes()
           .forEach(
               new KeyValueConsumer<String, AttributeValue>() {
--- a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
+++ b/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
@@ -33,7 +33,9 @@ public class LoggingExporter implements SpanExporter {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
-      System.out.print(String.format("%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId()));
+      System.out.printf(
+          String.format(
+              "%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId()));
       span.getAttributes()
           .forEach(
               new KeyValueConsumer<String, AttributeValue>() {
--- a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
+++ b/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
@@ -33,9 +33,7 @@ public class LoggingExporter implements SpanExporter {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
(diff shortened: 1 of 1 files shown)


javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
```diff
@@ -33,7 +33,7 @@ public LoggingExporter(String prefix) {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
-      System.out.print(prefix + " " + span.getName() + " " + span.getSpanId() + " ");
+      System.out.print(String.format("%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId()));
```
iNikem, line 36: Use `System.out.printf`

iNikem (approved):

javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
pavolloffay: oh my bad thanks for the catch!

javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
```diff
@@ -33,7 +33,9 @@ public LoggingExporter(String prefix) {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
-      System.out.print(String.format("%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId()));
+      System.out.printf(
+          String.format(
+              "%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId()));
```
trask, line 38:
```suggestion
      System.out.printf("%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId());
```

pavolloffay: done

trask (approved):
👍

javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
```diff
@@ -33,7 +33,7 @@ public LoggingExporter(String prefix) {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
-      System.out.print(prefix + " " + span.getName() + " " + span.getSpanId() + " ");
+      System.out.printf("%s %s %s %s", prefix, span.getName(), span.getTraceId(), span.getSpanId());
```
anuraaga, line 36: `printf` is sloooow - can we switch back to appending?

anuraaga: Also looks like we're missing space between span ID and attributes now probably good to have that (or maybe enclose attributes in `{}`

trask: No objection to switching back to appending. Good catch on the missing space 👍

iNikem: @anuraaga Any numbers to back that claim? :) And it shit important enough for us?

anuraaga: Have benchmarked similar now and again, generally 10-20x slower. The readability difference is small enough to get the free speed, and I always recommend people to just never use printf in Java, unless formatting numbers, since it's not worth the cognitive load of deciding between one or the other the performance difference is too great. So I'd consider it important for us.

pavolloffay: I will submit a PR to use the appending. 

Would it be better to use string builder

PR #1251, opened 2020-09-24
Use + instead of printf in logging exporter

pavolloffay:

Created from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1246#discussion_r494045579

@trask @iNikem is there any reason why the agent is not using the logging exporter from the SDK? The only difference is that SDK uses JUL and agent directly stdout.

cc) @anuraaga

### DIFF
--- a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
+++ b/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
@@ -33,7 +33,8 @@ public class LoggingExporter implements SpanExporter {
   @Override
   public CompletableResultCode export(Collection<SpanData> list) {
     for (SpanData span : list) {
-      System.out.print(prefix + " " + span.getName() + " " + span.getTraceId() + " " + span.getSpanId() + " ");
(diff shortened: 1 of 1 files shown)


anuraaga (approved):
If I recall correctly it was for some Agent-y reason @trask mentioned once but can't remember exactly.

anuraaga:
Ah also FYI re: StringBuilder. Concatenation and SB are currently equivalent because we compile to Java 8. But Java 9+ have optimized it further so I generally recommend sticking with append so (hopefully in another 15 years or so at the current rate of JVM upgrades...) when compiling against a new Java we get the perf for free.

iNikem:
JUL is complicated for javaagent, especially on WildFly and other JBosses.

trask (approved):

trask:
hey @pavolloffay, looks like this needs a `spotlessApply`

pavolloffay:
done

pavolloffay:
Can somebody restart the `build_test_deploy` job?

trask:
hey @pavolloffay, it's failing due to #1258, i'll ping you on this PR once that's fixed and then you can rebase and it should pass

trask:
master is fixed now 👍

PR #1253, opened 2020-09-24
Support otlp_span and otlp_metric exporter config names

### LINKED ISSUE #1250: Add support for otlp_span and otlp_metric values of OTEL_EXPORTER
as defined in https://github.com/open-telemetry/opentelemetry-specification/blob/2088381f57004be1de4fa16e0c2adeea04cf2ce2/specification/sdk-environment-variables.md#exporter-selection

pavolloffay:

Resolves #1250 

cc) @malafeev

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
@@ -91,6 +91,7 @@ public class TracerInstaller {
       if (metricExporterFactory
           .getClass()
           .getSimpleName()
+          .replace("_", "")
           .toLowerCase()
           .startsWith(exporterName.toLowerCase())) {
         return metricExporterFactory;
@@ -107,6 +108,7 @@ public class TracerInstaller {
       if (spanExporterFactory
           .getClass()
           .getSimpleName()
+          .replace("_", "")
           .toLowerCase()
           .startsWith(exporterName.toLowerCase())) {
         return spanExporterFactory;


iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -91,6 +91,7 @@ private static MetricExporterFactory findMetricExporterFactory(String exporterNa
       if (metricExporterFactory
           .getClass()
           .getSimpleName()
+          .replace("_", "")
```
anuraaga, line 94: Hmm - feels a bit magic but I guess it's fine 😃

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
pavolloffay: yeah, maybe the factories could have a method that returns a name and do the comparison based on that.

malafeev (approved):

anuraaga:
Thanks!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
trask: I like the idea of SpanExporterFactory/MetricExporterFactory having a method `Set<String> getNames()` (Set since e.g. OtlpSpanExporterFactory can be enabled via both `otlp` and `otlp_span`)

trask: I guess this is only applicable to our built-in exporters though, as opposed to someone building their own Exporter and using `EXPORTER_JAR`, so maybe not. Also, I forget, now that we are Java 8, we can always add interface methods after GA 😁

anuraaga: Yeah I think my idea before was to have a map of wellknown names, and only use the classname for custom exporters

PR #1254, opened 2020-09-24
Refactoring Config into @AutoValue POJO

mateuszrzeszutek:
Refactored `Config` according to this comment: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1243#pullrequestreview-495224970


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -54,189 +43,82 @@
...
+  public String getProperty(String propertyName) {
+    return getAllProperties().get(normalizePropertyName(propertyName));
+  }
 
-    log.debug("New instance: {}", this);
+  // some integrations have '-' or '_' character in their names -- this does not work well with
+  // environment variables (where we replace every non-alphanumeric character with '.'), so we're
+  // replacing those with a dot
+  public static String normalizePropertyName(String propertyName) {
+    return PROPERTY_NAME_REPLACEMENTS.matcher(propertyName.toLowerCase()).replaceAll(".");
```
trask, line 108: I think this is only needed for integrationNames
```suggestion
  private static String normalizeIntegrationName(String integrationName) {
    return INTEGRATION_NAME_REPLACEMENTS.matcher(integrationName.toLowerCase()).replaceAll(".");
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -54,189 +43,82 @@
...
-    hystrixTagsEnabled =
-        getBooleanSettingFromEnvironment(HYSTRIX_TAGS_ENABLED, parent.hystrixTagsEnabled);
+  public abstract Map<String, String> getAllProperties();
 
-    endpointPeerServiceMapping =
-        getPropertyMapValue(
-            properties, ENDPOINT_PEER_SERVICE_MAPPING, parent.endpointPeerServiceMapping);
+  @Nullable
+  public String getProperty(String propertyName) {
+    return getAllProperties().get(normalizePropertyName(propertyName));
```
trask, line 101: i think it's only integration names that need to be normalized
```suggestion
    return getAllProperties().get(propertyName);
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/ConfigBuilder.java
```diff
@@ -0,0 +1,182 @@
...
+  public static final String KAFKA_CLIENT_PROPAGATION_ENABLED =
+      "otel.kafka.client.propagation.enabled";
+  public static final String HYSTRIX_TAGS_ENABLED = "otel.hystrix.tags.enabled";
+  public static final String ENDPOINT_PEER_SERVICE_MAPPING = "otel.endpoint.peer.service.mapping";
+
+  private final Map<String, String> allProperties = new HashMap<>();
+
+  @Override
+  public ConfigBuilder readProperties(Properties properties) {
+    return this.fromConfigMap(normalizedProperties(properties), NamingConvention.DOT);
```
trask, line 72: i think only integrationNames need this normalization, and can be done special for them in `isIntegrationEnabled`
```suggestion
    return this.fromConfigMap(properties, NamingConvention.DOT);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -246,351 +128,113 @@ public boolean isIntegrationEnabled(SortedSet<String> integrationNames, boolean
...
   }
 
-  /**
-   * Calls {@link #getSettingFromEnvironment(String, String)} and converts the result to a list by
-   * splitting on `,`.
-   */
-  private static List<String> getListSettingFromEnvironment(String name, String defaultValue) {
-    return parseList(getSettingFromEnvironment(name, defaultValue));
-  }
+  public abstract Optional<String> getExporterJar();
```
trask, line 137: in general we have been using `@Nullable` over Optional

trask (commented):
Hey @mateuszrzeszutek!

Reviewing this gave me a further idea.

I was going to suggest not exposing the `Config.getProperty` method publicly (since it's not used outside of `Config` currently).

But it seems like there's a pretty good use case for custom instrumentation needing to call `Config.getProperty`.

And then I realized why shouldn't "bundled" instrumentation do the same?

(talking to myself) And anyways why should `instrumentation-api` know about instrumentation-specific configuration properties?

If this makes sense to you and @iNikem, I'll open an issue for it.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -54,189 +43,82 @@
...
+          .setTraceAnnotations(Optional.empty())
+          .setTraceMethods("")
+          .setTraceAnnotatedMethodsExclude("")
+          .setTraceExecutorsAll(DEFAULT_TRACE_EXECUTORS_ALL)
+          .setTraceExecutors(Collections.emptyList())
+          .setSqlNormalizerEnabled(DEFAULT_SQL_NORMALIZER_ENABLED)
+          .setKafkaClientPropagationEnabled(DEFAULT_KAFKA_CLIENT_PROPAGATION_ENABLED)
+          .setHystrixTagsEnabled(DEFAULT_HYSTRIX_TAGS_ENABLED)
+          .setEndpointPeerServiceMapping(Collections.emptyMap())
+          .build();
```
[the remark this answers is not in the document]
trask: I guess no need to do this if we are going to remove most/all of these attributes from Config and primarily just store the properties map in Config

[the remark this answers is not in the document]
trask: the other comments (besides this one) still make sense for this PR i think, even if we are going to go forward with above idea in the future

iNikem:
Assorted thoughts follow.

If I create new instrumentation which needs a configuration option, should I go and change common `Config` class to add named setter for that? Seems like unnecessary coupling. Therefore, generic `Config.getProperty` makes sense and all instrumentations may/should use it and keep property names to themselves.

The only downside of the above: we still have to keep the common configuration page or to create "documentation per instrumentation". The latter may be a good idea, albeit work intensive.

We should re-use and not duplicate normalization logic from `ConfigBuilder` if possible.

The precedence of configuration, system properties -> env -> file -> SPI -> default, should ideally be in one place.

So, if I understand it correctly, some class from `tooling` module should collect all configuration from all known sources in the correct order, normalise it to a single instance of `Properties` or other map-like structure and create single `Config` instance (which lives in bootstrap module) from that. All instrumentations then can use it via single `getProperty` (or typed `getLongProperty` etc) method.

It seems I agree with @trask  :)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
mateuszrzeszutek: Yes, only integration names -- but we have to normalize them both when building the config (since someone can use dash in e.g. system property `-Dotel.integration.some-library.enabled=false`) and when `isIntegrationEnabled()` is called.
So I opted for adding the normalisation everywhere to be 100% consistent.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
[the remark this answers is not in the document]
mateuszrzeszutek: Hm, for now I've used the defaults pattern suggested by AutoValue.

mateuszrzeszutek:
> I guess no need to do this if we are going to remove most/all of these attributes from Config and primarily just store the properties map in Config

Do you suggest removing all those fields and just using `getProperty()` internally in `Config` getters?

mateuszrzeszutek:
> If I create new instrumentation which needs a configuration option, should I go and change common Config class to add named setter for that? Seems like unnecessary coupling. Therefore, generic Config.getProperty makes sense and all instrumentations may/should use it and keep property names to themselves.

There already are some properties like that: `getKafkaClientPropagationEnabled()` or `getHystrixTagsEnabled()`. I believe that we should remove them and make those instrumentations call `getProperty()` (typed `getBooleanProperty()`?) instead.
This way `Config` fields would only contain things that relate to the agent as a whole or are common to all instrumentations.
What do you think about it?

trask:
> > I guess no need to do this if we are going to remove most/all of these attributes from Config and primarily just store the properties map in Config
> 
> Do you suggest removing all those fields and just using `getProperty()` internally in `Config` getters?

I suggest removing both the fields and the getters for instrumentation-specific config properties and also for config properties that are specific to `javaagent-tooling` (since no need to expose `javaagent-tooling` properties explicitly in `instrumentation-api`).

> > If I create new instrumentation which needs a configuration option, should I go and change common Config class to add named setter for that? Seems like unnecessary coupling. Therefore, generic Config.getProperty makes sense and all instrumentations may/should use it and keep property names to themselves.
> 
> There already are some properties like that: `getKafkaClientPropagationEnabled()` or `getHystrixTagsEnabled()`. I believe that we should remove them and make those instrumentations call `getProperty()` (typed `getBooleanProperty()`?) instead.
> This way `Config` fields would only contain things that relate to the agent as a whole or are common to all instrumentations.
> What do you think about it?

👍 (and same for `javaagent-tooling` which can call `getProperty()` or typed variants also)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -201,7 +196,7 @@ private static void configure() {
     /* Update trace config from env vars or sys props */
     TraceConfig activeTraceConfig = tracerSdkProvider.getActiveTraceConfig();
     tracerSdkProvider.updateActiveTraceConfig(
-        activeTraceConfig.toBuilder().readEnvironmentVariables().readSystemProperties().build());
+        activeTraceConfig.toBuilder().readProperties(Config.get().asJavaProperties()).build());
```
anuraaga, line 199: Is it possible to only call `asJavaProperties` once?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/ConfigBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  private static Map<String, String> normalizedProperties(Properties properties) {
+    Map<String, String> configMap = new HashMap<>(properties.size());
+    for (Enumeration<?> e = properties.propertyNames(); e.hasMoreElements(); ) {
```
anuraaga, line 76: Never been a fan of `Properties` class 😅 Just wondering, is it a bit simpler to iterate over `keySet`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -54,189 +42,63 @@
...
-            TRACE_ANNOTATED_METHODS_EXCLUDE, DEFAULT_TRACE_ANNOTATED_METHODS_EXCLUDE);
-
-    traceExecutorsAll =
-        getBooleanSettingFromEnvironment(TRACE_EXECUTORS_ALL, DEFAULT_TRACE_EXECUTORS_ALL);
+  private static final Config DEFAULT = Config.newBuilder().build();
 
-    traceExecutors = getListSettingFromEnvironment(TRACE_EXECUTORS, DEFAULT_TRACE_EXECUTORS);
+  // INSTANCE can never be null - muzzle instantiates instrumenters when it generates
+  // getInstrumentationMuzzle() and the Instrumenter.Default constructor uses Config
+  private static volatile Config INSTANCE = DEFAULT;
```
anuraaga, line 58: `Config` becoming `volatile` is a surprisingly big change. It's going to make unit tests much easier where we currently use bytecode manipulation to turn `Config` from final to non-final so I'm generally happy about it though :)

We should confirm that we only call `Config.get` on cold paths, which I think should generally be the case as it's basically only  to initialize instrumentation. In Java, the performance of `Config.get().isFooEnabled()` is actually significantly different when Java can know a boolean is a constant and C1 compiler can just remove the branches with no de-opt mechanism at all. So when `.get` is volatile, we'd expect a `private static final boolean IS_FOO_ENABLED = Config.get().isFooEnabled()` to replace such code inline on a hot path where we would like fast enabling/disabling of code paths

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -54,189 +42,63 @@
...
-        getPropertyBooleanValue(properties, TRACE_EXECUTORS_ALL, parent.traceExecutorsAll);
-    traceExecutors = getPropertyListValue(properties, TRACE_EXECUTORS, parent.traceExecutors);
-
-    sqlNormalizerEnabled =
-        getPropertyBooleanValue(properties, SQL_NORMALIZER_ENABLED, parent.sqlNormalizerEnabled);
-
-    kafkaClientPropagationEnabled =
-        getPropertyBooleanValue(
-            properties, KAFKA_CLIENT_PROPAGATION_ENABLED, parent.kafkaClientPropagationEnabled);
+  public static Config get() {
```
anuraaga, line 73: Might add a recommendation to store config values as constants if they're referenced in performance-sensitive code.

anuraaga (approved):
Thanks a lot, just small comments. Great cleanup

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/ConfigBuilder.java
mateuszrzeszutek: It is, I kinda forgot that `Properties` are a `Map` 😅 

Done, though it's not ideal -- because unfortunately both key and value in `Properties` are `Object`s.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
mateuszrzeszutek: That's a good point -- I'll look through usages of `Config#get()`

mateuszrzeszutek:
> I suggest removing both the fields and the getters for instrumentation-specific config properties and also for config properties that are specific to javaagent-tooling (since no need to expose javaagent-tooling properties explicitly in instrumentation-api).

> > There already are some properties like that: getKafkaClientPropagationEnabled() or getHystrixTagsEnabled(). ...

>👍 (and same for javaagent-tooling which can call getProperty() or typed variants also)

> Might add a recommendation to store config values as constants if they're referenced in performance-sensitive code.

Great -- I'll create a separate GH issue for that. I can implement all those fixes right away, I just want to stop this PR from growing in scope (and get it merged, of course).

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
trask: makes sense 👍

trask (approved):

trask:
hey @mateuszrzeszutek, heads up, I merged master into your branch, but build not passing due to #1280, once that is merged you can merge master again and should be good to go 👍

trask:
@anuraaga @iNikem heads up there's a conflict between this PR (which replaces `ExporterConfig` with `Properties`) and #1262 (which uses `ExporterConfig`)

trask:
Thanks @mateuszrzeszutek!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -54,189 +42,63 @@
...
-
-    endpointPeerServiceMapping = getMapSettingFromEnvironment(ENDPOINT_PEER_SERVICE_MAPPING);
-
-    log.debug("New instance: {}", this);
+  /**
+   * Sets the agent configuration singleton. This method is only supposed to be called once, from
+   * the agent classloader just before the first instrumentation is loaded (and before {@link
+   * Config#get()} is used for the first time).
+   */
+  public static void internalInitializeConfig(Config config) {
```
iNikem, line 65: I usually prefer method names like `doSomethingInternal` vs `internalDoSomething`. What do others think?

anuraaga: I have slight preference for `internalDoSomething` - gets the point across that it's internal sooner. Another option could be to make the access `package-private` and move the method to an `InternalConfigAccess` type of class.

trask: interesting, I've typically done `doSomethingInternal`, because it is often a delegate of `doSomething`, but I like the reasoning of putting internal in front to "get the point across that it's internal sooner"

Issue #1043, opened 2020-08-18
Document how SNAPSHOT versions may or may not work with OpenTelemetry API manual instrumentation


trask:
Maybe doc how to find (and use) the right OpenTelemetry API SNAPSHOT if you do want to make it work.

trask:
Closing, this was created (and relevant) when we were using otel-java SNAPSHOTs

PR #1262, opened 2020-09-26
Add Prometheus metric exporter

### LINKED ISSUE #385: Add Prometheus auto-exporter
This involves a couple of hurdles

* We currently only support push-based metrics exporters
* We currently only support a single exporter jar, so how does someone use jaeger + prometheus?
  * EDIT: see https://github.com/open-telemetry/opentelemetry-specification/issues/710

pavolloffay:
Resolves #385

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -34,6 +34,7 @@ ext {
     nullaway            : "0.8.0",
     autoValue           : "1.7.4",
     systemLambda        : "1.1.0",
+    prometheus          : "0.9.0"
   ]
 
   deps = [
@@ -46,6 +47,7 @@ ext {
     opentelemetryJaeger         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-jaeger', version: versions.opentelemetry),
     opentelemetryOtlp           : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-otlp', version: versions.opentelemetryAnother),
     opentelemetryZipkin         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-zipkin', version: versions.opentelemetry),
+    opentelemetryPrometheus     : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-prometheus', version: versions.opentelemetry),
     opentelemetryLogging        : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-logging', version: versions.opentelemetry),
     opentelemetryProto          : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-proto', version: versions.opentelemetry),
 
@@ -63,6 +65,10 @@ ext {
     autoValueAnnotations        : "com.google.auto.value:auto-value-annotations:${versions.autoValue}",
     // annotation processor
     autoValue                   : "com.google.auto.value:auto-value:${versions.autoValue}",
+    prometheus                  : [
+      dependencies.create(group: 'io.prometheus', name: 'simpleclient', version: "${versions.prometheus}"),
+      dependencies.create(group: 'io.prometheus', name: 'simpleclient_httpserver', version: "${versions.prometheus}"),
+    ],
 
     // Testing
 
--- a/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
+++ b/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.spi.exporter;
+
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import java.util.Properties;
+
+/**
+ * A {@link MetricServer} acts as the bootstrap for metric exporters that use {@link MetricProducer}
+ * to consume the metrics.
+ *
+ * <p>Implementation of {@link MetricServer} must be registered through the Java SPI framework.
+ */
+public interface MetricServer {
+
+  /**
+   * Initialize the metric server that pull metric from the {@link MetricProducer}.
+   *
+   * @param producer The metric producer
+   * @param config The configuration
+   */
+  void configure(MetricProducer producer, Properties config);
+}
--- a/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
+++ b/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
@@ -17,10 +17,10 @@ import java.util.Properties;
 public interface MetricServer {
 
   /**
-   * Initialize the metric server that pull metric from the {@link MetricProducer}.
+   * Start the metric server that pulls metric from the {@link MetricProducer}.
    *
    * @param producer The metric producer
    * @param config The configuration
    */
-  void configure(MetricProducer producer, Properties config);
+  void start(MetricProducer producer, Properties config);
 }
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
@@ -80,7 +80,7 @@ public class TracerInstaller {
       if (metricServer != null) {
         installMetricServer(metricServer, config);
       } else {
-        log.debug("No {} metric producer customer found", exporterName);
+        log.debug("No {} metric server found", exporterName);
       }
     }
   }
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
@@ -8,10 +8,12 @@ package io.opentelemetry.javaagent.tooling;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.javaagent.spi.TracerCustomizer;
 import io.opentelemetry.javaagent.spi.exporter.MetricExporterFactory;
+import io.opentelemetry.javaagent.spi.exporter.MetricServer;
 import io.opentelemetry.javaagent.spi.exporter.SpanExporterFactory;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.metrics.export.IntervalMetricReader;
 import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
 import io.opentelemetry.sdk.trace.TracerSdkManagement;
 import io.opentelemetry.sdk.trace.config.TraceConfig;
 import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
@@ -73,6 +75,13 @@ public class TracerInstaller {
       } else {
         log.debug("No {} metric exporter found", exporterName);
       }
+
+      MetricServer metricServer = findMetricServer(exporterName);
+      if (metricServer != null) {
+        installMetricServer(metricServer, config);
+      } else {
+        log.debug("No {} metric producer customer found", exporterName);
+      }
     }
   }
 
@@ -88,6 +97,23 @@ public class TracerInstaller {
           .toLowerCase()
           .startsWith(exporterName.toLowerCase())) {
         return metricExporterFactory;
+      }
+    }
+    return null;
+  }
+
+  private static MetricServer findMetricServer(String exporterName) {
+    ServiceLoader<MetricServer> serviceLoader =
+        ServiceLoader.load(MetricServer.class, TracerInstaller.class.getClassLoader());
+
+    for (MetricServer metricServer : serviceLoader) {
+      if (metricServer
+          .getClass()
+          .getSimpleName()
+          .replace("_", "")
+          .toLowerCase()
+          .startsWith(exporterName.toLowerCase())) {
+        return metricServer;
       }
     }
     return null;
@@ -157,6 +183,12 @@ public class TracerInstaller {
         BatchSpanProcessor.newBuilder(spanExporter).readProperties(config).build();
     OpenTelemetrySdk.getTracerManagement().addSpanProcessor(spanProcessor);
     log.info("Installed span exporter: " + spanExporter.getClass().getName());
+  }
+
+  private static void installMetricServer(MetricServer metricServer, Properties config) {
+    MetricProducer metricProducer = OpenTelemetrySdk.getMeterProvider().getMetricProducer();
+    metricServer.configure(metricProducer, config);
+    log.info("Installed metric server: " + metricServer.getClass().getName());
   }
 
   private static <F> F getExporterFactory(Class<F> service, ExporterClassLoader exporterLoader) {
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
@@ -187,7 +187,7 @@ public class TracerInstaller {
 
   private static void installMetricServer(MetricServer metricServer, Properties config) {
     MetricProducer metricProducer = OpenTelemetrySdk.getMeterProvider().getMetricProducer();
-    metricServer.configure(metricProducer, config);
+    metricServer.start(metricProducer, config);
     log.info("Installed metric server: " + metricServer.getClass().getName());
   }
 
--- a/javaagent-exporters/javaagent-exporters.gradle
+++ b/javaagent-exporters/javaagent-exporters.gradle
@@ -12,6 +12,7 @@ dependencies {
   testImplementation project(':javaagent-exporters:jaeger')
   testImplementation project(':javaagent-exporters:logging')
   testImplementation project(':javaagent-exporters:zipkin')
+  testImplementation project(':javaagent-exporters:prometheus')
 }
 
 tasks.withType(Test).configureEach() {
@@ -19,6 +20,7 @@ tasks.withType(Test).configureEach() {
   dependsOn ':javaagent-exporters:jaeger:shadowJar'
   dependsOn ':javaagent-exporters:logging:shadowJar'
   dependsOn ':javaagent-exporters:zipkin:shadowJar'
+  dependsOn ':javaagent-exporters:prometheus:shadowJar'
   doFirst {
     systemProperty 'projectVersion', allprojects.version[0]
     systemProperty 'adapterRoot', "$rootDir/javaagent-exporters"
@@ -26,6 +28,7 @@ tasks.withType(Test).configureEach() {
     systemProperty 'jaegerExporterJar', project(':javaagent-exporters:jaeger').tasks.shadowJar.archivePath
     systemProperty 'loggingExporterJar', project(':javaagent-exporters:logging').tasks.shadowJar.archivePath
     systemProperty 'zipkinExporterJar', project(':javaagent-exporters:zipkin').tasks.shadowJar.archivePath
+    systemProperty 'prometheusExporterJar', project(':javaagent-exporters:prometheus').tasks.shadowJar.archivePath
   }
 }
 
@@ -38,6 +41,7 @@ dependencies {
   shadowInclude project(path: ':javaagent-exporters:otlp', configuration: 'shadow')
   shadowInclude project(path: ':javaagent-exporters:jaeger', configuration: 'shadow')
   shadowInclude project(path: ':javaagent-exporters:zipkin', configuration: 'shadow')
+  shadowInclude project(path: ':javaagent-exporters:prometheus', configuration: 'shadow')
 }
 
 jar.enabled = false
--- a/javaagent-exporters/prometheus/prometheus.gradle
+++ b/javaagent-exporters/prometheus/prometheus.gradle
@@ -1,0 +1,28 @@
+plugins {
+  id "com.github.johnrengelman.shadow"
+}
+
+apply from: "$rootDir/gradle/java.gradle"
+apply from: "$rootDir/gradle/publish.gradle"
+
+archivesBaseName = 'javaagent-exporters-prometheus'
+
+dependencies {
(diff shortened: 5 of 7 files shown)


pavolloffay:
Could have a look at this maintainer who has experience with metrics API?

I am not sure if this is the right approach to do this. I have tried to use this example in my demo spring boot APP but I didn't get any metrics exposed on the endpoint https://github.com/open-telemetry/opentelemetry-java/blob/master/examples/prometheus/src/main/java/io/opentelemetry/example/PrometheusExample.java#L31

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricExporterFactory.java
```diff
@@ -0,0 +1,60 @@
...
+  public MetricExporter fromConfig(ExporterConfig config) {
+    PrometheusCollector.newBuilder()
+        .setMetricProducer(meterSdkProvider.getMetricProducer())
+        .buildAndRegister();
+    try {
+      int port = config.getInt(PORT_CONF_PROP_NAME, DEFAULT_PORT);
+      String host = config.getString(HOST_CONF_PROP_NAME, DEFAULT_HOST);
+      log.info("Creating Prometheus exporter on host: '{}' and port: '{}'", host, port);
+      HTTPServer server = new HTTPServer(host, port);
+      return new PrometheusExporter(server);
```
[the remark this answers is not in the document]
pavolloffay: I have switch to this API, however I still don't get any metrics from my APP. 

An alternative solution might be to implement adapter for `MetricProducer` in `PrometheusMetricExporterFactory` that would cache the metrics returned by the `MetricExporter`.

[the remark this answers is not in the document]
pavolloffay: There is smth weird going on, I am not able to get metrics from the app to the agent. 

In the debugger I see the metrics being collected (`OpenTelemetrySdk.getMeterProvider().getMetricProducer().collectAllMetrics()`) but the exporter or collector do not get any metrcics.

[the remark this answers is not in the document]
pavolloffay: It seems that the app is using a different instance of `OpenTelemetrySdk.getMeterProvider`  than the agent.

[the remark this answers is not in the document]
pavolloffay: My issue is that the `opentelemetry-api-beta` does no kick so the calls to the OTel API in the app are noop. Since I have included only the API.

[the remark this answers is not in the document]
pavolloffay: Finally, I have managed to get this working. There are two things that I have learned by hard:
1. the app should use only the OTel API
2. I had to rebase and use OTEL 0.9.0-SNAPSHOT API in the app.

gradle/dependencies.gradle
```diff
@@ -43,6 +44,7 @@ ext {
     opentelemetryJaeger         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-jaeger', version: versions.opentelemetry),
     opentelemetryOtlp           : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-otlp', version: versions.opentelemetry),
     opentelemetryZipkin         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-zipkin', version: versions.opentelemetryOther),
+    opentelemetryPrometheus     : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporters-prometheus', version: versions.opentelemetry),
```
pavolloffay, line 47: @trask @anuraaga why there is `versions.opentelemetryOther` and `version.opentelemetry`?

anuraaga: Jfrog has been well behaved recently but often some of the snapshot artifacts have different version suffixes. We've found there to be cases of three different suffixes for the same version. An annoyance when dealing with snapshots that will go away with GA so currently living with it.

pavolloffay: Which one should I use? 

Currently all point to the same versions
```
    opentelemetry       : '0.9.0-20200925.025016-49',
    opentelemetryOther  : '0.9.0-20200925.025016-49',
    opentelemetryAnother: '0.9.0-20200925.025016-49',
```

iNikem: Does not matter, `opentelemetry` is fine if it works now

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricExporterFactory.java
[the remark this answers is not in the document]
trask: sorry about that ☹️

I logged #1279 for the first item, and bumped priority for the second item #1043, though that issue will magically go away once we GA since we will support all GA OTel API versions

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricCollector.java
```diff
@@ -39,22 +37,18 @@
...
+  public void init(MetricProducer producer, ExporterConfig config) {
+
+    PrometheusCollector.newBuilder().setMetricProducer(producer).buildAndRegister();
     try {
       int port = config.getInt(PORT_CONF_PROP_NAME, DEFAULT_PORT);
       String host = config.getString(HOST_CONF_PROP_NAME, DEFAULT_HOST);
       log.info("Creating Prometheus exporter on host: '{}' and port: '{}'", host, port);
-      HTTPServer server = new HTTPServer(host, port);
-      return new PrometheusExporter(server);
+      // TODO close the server
```
trask, line 47: nice thing about javaagent is we don't have to close, though we do have to worry about creating non-daemon threads, can you check that? e.g. use the javaagent w/ prometheus exporter from a command-line app and make sure we don't prevent the app from exiting and the end of `main`
```suggestion
```

trask:
@anuraaga @iNikem heads up there's a conflict between this PR (which uses `ExporterConfig`) and #1254 (which replaces `ExporterConfig` with `Properties`)

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricCollector.java
```diff
@@ -0,0 +1,36 @@
...
+
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
+
+/**
+ * A {@link MetricCollector} acts as the bootstrap for metric exporters that use {@link
+ * MetricProducer} to consume the metrics.
+ *
+ * <p>Implementation of {@link MetricCollector} must be registered through the Java SPI framework.
+ */
+public interface MetricCollector {
```
iNikem, line 27: Why is this `Collector` and not `Exporter`?

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricCollector.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private static final String EXPORTER_NAME = "prometheus";
+  private static final String PORT_CONF_PROP_NAME = EXPORTER_NAME + ".port";
+  private static final String HOST_CONF_PROP_NAME = EXPORTER_NAME + ".host";
+  private static final int DEFAULT_PORT = 9464;
+  // The empty address equals to any address
+  private static final String DEFAULT_HOST = "0.0.0.0";
+
+  @Override
+  public void init(MetricProducer producer, ExporterConfig config) {
```
iNikem, line 41: I think this method requires documentation to explain what is going on here. Just reading it I don't understand what is a connection between `PrometheusCollector` and we built and registered and `HTTPServer` that we started later. How that server gets metrics?

The absence of javadoc on `PrometheusCollector` does not help it neither :)

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricCollector.java
pavolloffay: The name was suggested by @trask. I think we can do better. 

I am more in favor of removing this interface and use https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1273

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricCollector.java
pavolloffay: +1 I will add some docs.

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricCollector.java
iNikem: You mean removing all individual SPIs and having one single SPI `OpenTelemetrySdkCustomizer`?

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricCollector.java
pavolloffay: Good catch, there is a `daemon` param on the constructor.

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricCollector.java
pavolloffay: Not really, The #1273 does not seem feasible since the static class, we could pass all the telemetry objects but that does not seem right to me.

I have renamed the class to 
```java
public interface MetricProducerCustomizer {

  /**
   * Initialize the metric collector.
   *
   * @param producer The metric producer
   */
  void configure(MetricProducer producer, Properties config);
}

the `configure` method matches the `TracerCustomizer`.
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -99,6 +108,23 @@ private static MetricExporterFactory findMetricExporterFactory(String exporterNa
...
+    ServiceLoader<MetricProducerCustomizer> serviceLoader =
+        ServiceLoader.load(MetricProducerCustomizer.class, TracerInstaller.class.getClassLoader());
+
+    for (MetricProducerCustomizer metricProducerCustomizer : serviceLoader) {
+      if (metricProducerCustomizer
+          .getClass()
+          .getSimpleName()
+          .replace("_", "")
+          .toLowerCase()
+          .startsWith(exporterName.toLowerCase())) {
```
mateuszrzeszutek, line 121: Instead of this undocumented naming convention what do you think about adding a `String getExporterName()` method to the `MetricProducerCustomizer` interface?

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricCollector.java
trask: @iNikem `MetricExporter` is for push-based metric exporters, there isn't a corresponding SDK interface for pull-based exporters

the name `MetricCollector` makes sense to me because they call `collectAllMetrics()` on the `MetricProducer`, as opposed to customizing the `MetricProducer`

but I'm also good with `MetricProducerCustomizer`

I think the important thing that we should only support one of these two configurations:

* A single `MetricProducerCustomizer` + zero `MetricExporterFactory`, or
* Zero `MetricProducerCustomizer` + any number of `MetricExporterFactory` (whose exporters we wrap in a single `IntervalMetricReader`)

(but this doesn't need to be in this PR, I created #1281 to track this)

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricProducerCustomizer.java
```diff
@@ -0,0 +1,58 @@
...
+ * Prometheus server scraps metric endpoint.
+ */
+@AutoService(MetricProducerCustomizer.class)
+public class PrometheusMetricProducerCustomizer implements MetricProducerCustomizer {
+  private static final Logger log =
+      LoggerFactory.getLogger(PrometheusMetricProducerCustomizer.class);
+
+  private static final String EXPORTER_NAME = "prometheus";
+  private static final String PORT_CONF_PROP_NAME = EXPORTER_NAME + ".port";
+  private static final String HOST_CONF_PROP_NAME = EXPORTER_NAME + ".host";
```
mateuszrzeszutek, line 41: Just wondering: shouldn't these settings be prefixed with `otel.exporter.` or at least `otel.`?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
trask: ya, this has come up for `SpanExporterFactory` and `MetricExporterFactory` also https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1253#discussion_r494221024 😄

I will create a tracking issue for this

trask: #1287

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricProducerCustomizer.java
trask: oh ya, this just changed in #1254, `ExporterConfig` used to automagically add the prefix

trask: @pavolloffay can you check this? then I think good to merge!

pavolloffay: sorry for the delay I have updated it.

mateuszrzeszutek (approved):

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricProducerCustomizer.java
```diff
@@ -0,0 +1,58 @@
...
+import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import io.prometheus.client.exporter.HTTPServer;
+import java.io.IOException;
+import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link PrometheusMetricProducerCustomizer} registers {@link MetricProducer} to the ${@link
+ * PrometheusMetricProducerCustomizer}. The collector pulls metrics from the producer when the
```
iNikem, line 31:
```suggestion
 * PrometheusCollector}. The collector pulls metrics from the producer when the
```

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricProducerCustomizer.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final String DEFAULT_HOST = "0.0.0.0";
+
+  @Override
+  public void configure(MetricProducer producer, Properties config) {
+    PrometheusCollector.newBuilder().setMetricProducer(producer).buildAndRegister();
+    try {
+      String portStr = config.getProperty(PORT_CONF_PROP_NAME, DEFAULT_PORT);
+      String host = config.getProperty(HOST_CONF_PROP_NAME, DEFAULT_HOST);
+      log.info("Creating Prometheus exporter on host: '{}' and port: '{}'", host, portStr);
+      new HTTPServer(host, Integer.parseInt(portStr), true);
```
iNikem, line 53: Sorry, but I still not quite understand how `HTTPServer` and `PrometheusCollector` are connected :(

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -165,6 +191,14 @@ private static void installExporter(SpanExporterFactory spanExporterFactory, Pro
     log.info("Installed span exporter: " + spanExporter.getClass().getName());
   }
 
+  private static void installCollector(
```
iNikem, line 194: What collector? I see `MetricProducerCustomizer` and `MetricProducer` in method body, what is collector? :)

iNikem (commented):
I am afraid that current proposal will be quite confusing in the future because of its vague use of terms. It seems that we miss some concept here. If before we had "push-based" exporters, then here we try to introduce "pull-based" on. So we have to find a way to expose a source of metrics (`MetricProducer`) to an outside scraper. `MetricEndpoint` may be? Something is missing in our current mental model and vocabulary.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
pavolloffay: This is a leftover from the previous naming. I will rename it to match the new name.

javaagent-exporters/prometheus/src/main/java/io/opentelemetry/javaagent/exporters/prometheus/PrometheusMetricProducerCustomizer.java
pavolloffay: Added a comment

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
pavolloffay: @iNikem the metrics can use the pull and push model. The p8s metric exporter uses pull model. It pulls metrics from OTEL once the metric endpoint is hit.

iNikem: But push model is supported by our Exporters API, they don't need any new SPI, right?

trask: @iNikem `MetricExporter` supports push model, but not pull model. Pull model has to use `MetricProducer` directly, e.g. https://github.com/open-telemetry/opentelemetry-java/blob/a08dd9fbe4a98f34aec3618475905fa55fece87e/examples/prometheus/src/main/java/io/opentelemetry/example/PrometheusExample.java#L46-L48

trask: > E.g. if we make `MetricProducerCustomizer.getExporterName` it will be confusing to me: "why customizer of MetricProducers is connected to an exporter"?

This is maybe a reason to rename `MetricProducerCustomizer`. The question is what to rename it to?

trask: Some options:
* MetricConsumer
* MetricCollector
* MetricEndpoint

pavolloffay: Maybe we could unify these interfaces and just simply add MetricProducer to the current interface? If null is returned it would not install the exporter to the SDK.


```java
public interface MetricExporterFactory {
  MetricExporter fromConfig(MetricProducer, Properties);
}
```

trask: there's some benefit to having different factories for push and pull cases, since we can support multiple push factories, but only a single pull factory (and not both): https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1281#issuecomment-700834529

trask: @pavolloffay @iNikem @anuraaga I think we just need a name for `MetricProducerCustomizer` and then good to merge. 

What do you think of:

```
public interface MetricExporterInitializer {
  void start(MetricProducer, Properties);
}
```

Though maybe that's too close to `MetricExporterFactory`, and people might use `MetricExporterInitializer` accidentally when it would be better to use `MetricExporterFactory` (which can support multiple exporters, and can be configured using `IntervalMetricReader`).

Another option is to go all-in on the one use case we have so far (exposing a metric server):

```
public interface MetricServer {
  void start(MetricProducer, Properties);
}
```

I'm struggling to come up with a name that's both general and not confusing, so would probably opt for the narrower and less confusing `MetricServer`. (credit to @iNikem for this name btw)

Thoughts?

pavolloffay:
PR updated and rebased.

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public interface MetricServer {
+
+  /**
+   * Initialize the metric server that pull metric from the {@link MetricProducer}.
+   *
+   * @param producer The metric producer
+   * @param config The configuration
+   */
+  void configure(MetricProducer producer, Properties config);
```
iNikem, line 25: Do you think `confugure` is a better name than `start` as Trask proposed?

pavolloffay: my bad I didn't pay attention. Either one if fine with me, I will rename it to start.

trask (approved):
Thanks @pavolloffay!

Sorry for the delays around naming 😁.

🎉 for being able to test metrics locally via prometheus endpoint now

Issue #482, opened 2020-06-09
Publish a debug jar without the classdata obfuscation


anuraaga:
I noticed that I can't debug classes in the agent in IntelliJ because it doesn't recognize the files named `.classdata`. Let me know if there's a way that I'm missing :) But if not, it would be helpful to have a `-debug` jar or something that has normal `.class` files I believe that would allow the debugger to work.

devinsba:
Adding `disableShadowRelocate=true` to your `~/.gradle/gradle.properties` should disable the renaming

anuraaga:
Hmm, I'm not sure but that seems like it would disable shading, but not the file renaming? I don't see the flag used here

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/java-agent/java-agent.gradle#L15

But assuming a flag does exist or is created, that requires building locally which takes a long time for this monolith so publishing such a build to download would be great :)

FrankSpitulski:
Just tried disableShadowRelocate=true, it does not appear to restore the .classdata files to .class.

pavolloffay:
@anuraaga I am able to use breakpoint in the instrumentation classes e.g. `Servlet3HttpServerTracer`  by running `java -javaagent:opentelemetry-javaagent-<version>-all.jar -jar -agentlib:jdwp="transport=dt_socket,server=y,suspend=y,address=5000" app.jar`

jbaris:
Hi @pavolloffay 
Where the -all.jar comes from?

Thanks in advance

mateuszrzeszutek:
We used to publish an `opentelemetry-javaaget-all.jar` on release - we've renamed it to `opentelemetry-javaagent.jar` later on because it was confusing.

PR #1263, opened 2020-09-26
Add debugging docs

### LINKED ISSUE #1241: Add instructions on how to debug agent
I could not find any instructions on how to debug the agent with a running application. 

By running the following code I was able to debug the app, but the breakpoints in the agent were never triggered e.g. in `TracerInstaller`.

```bash
OTLP_EXPORTER=otlp java -javaagent:/home/ploffay/tmp/opentelemetry-javaagent-0.8.0-all.jar -jar -agentlib:jdwp="transport=dt_socket,server=y,suspend=y,address=5000" app.jar
```

pavolloffay:
Resolves #1241 

Related to #482 cc) @anuraaga

### DIFF
--- a/docs/contributing/debugging.md
+++ b/docs/contributing/debugging.md
@@ -3,20 +3,21 @@
 Debugging java agent can be a challenging task since some instrumentation
 code is directly inlined into target classes and debugger is
 usually not attached early enough to activate breakpoints
-in agent initialization code (`premain`/`agentmain`, `agent-tooling` artifact etc.).
+in agent initialization code (`OpenTelemetryAgent`, `AgentInitializer`, `AgentInstaller`,
+`TracerInstaller`, etc.).
 
-#### Template methods and agent initialization
+#### Advice methods and agent initialization
 
-Breakpoints do not work in template methods, because their code is directly inlined
+Breakpoints do not work in advice methods, because their code is directly inlined
 by ByteBuddy into the target class. It is good to keep these methods as small as possible.
-The template methods are annotated with:
+The advice methods are annotated with:
 
 ```java
 @net.bytebuddy.asm.Advice.OnMethodEnter
 @net.bytebuddy.asm.Advice.OnMethodExit
 ```
 
-The best approach to debug template methods and agent initialization is to use the following statements:
+The best approach to debug advice methods and agent initialization is to use the following statements:
 
 ```java
 System.out.println()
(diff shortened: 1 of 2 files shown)


docs/contributing/debugging.md
```diff
@@ -0,0 +1,33 @@
+### Debugging
+
+Debugging java agent can be a challenging task since some instrumentation
+code is directly inlined into target classes and debugger is
+usually not attached early enough to activate breakpoints
+in agent initialization code (`premain`/`agentmain`, `agent-tooling` artifact etc.).
```
trask, line 6: `agentmain` is for "hot attach" and I think breakpoints may work there (though we don't use/test/doc hot attach at this time).

naming specific classes as examples I think is good, since lots of `agent-tooling` classes will work under debugger, just not the initialization code

```suggestion
in agent initialization code (`OpenTelemetryAgent`, `AgentInitializer`, `AgentInstaller`, `TracerInstaller`, etc.).
```

docs/contributing/debugging.md
```diff
@@ -0,0 +1,33 @@
...
+
+#### Template methods and agent initialization
```
trask, line 8: I don't think we use the term "template methods" elsewhere. I think "advice methods" may be better.

trask (commented):
Thanks, this in really nice, addresses an important gap in our contributing docs!

docs/contributing/debugging.md
```diff
@@ -0,0 +1,33 @@
...
+The best approach to debug template methods and agent initialization is to use the following statements:
+
+```java
+System.out.println()
+Thread.dumpStack()
+```
+
+#### Enable debugger
+
+The following example shows remote debugger configuration. The breakpoints
```
anuraaga, line 28: IIUC, this requires remote debugging from this repo since it has the code, but not a users if they are using the agent. Since this is CONTRIBUTING.md that's totally fine :) With the linked issue I'm hoping we can provide that jar so it's easier to debug for a user though - add compileOnly dependency on obfuscated jar and use normal IntelliJ local debug, it's worked ok for me in the past.

anuraaga (approved):
LGTM with trask's comments

pavolloffay:
I have fixed the review comments and updated the PR.

trask:
Thanks @pavolloffay 🎉

PR #1266, opened 2020-09-26
Update to the latest opentelemetry-java 0.9.0-SNAPSHOT

### LINKED ISSUE #1258: Build failing because opentelemetry-java SNAPSHOTs are being pruned from https://oss.jfrog.org/artifactory/oss-snapshot-local
This is happening again (see #1102 for background).

I will fix tomorrow (or hopefully today if I can get to it).

trask:
Resolves #1258


trask:
Having to use `SpanBuilder.setParent(Context)` now turned out to be very good, forcing us to propagate the full Context in remaining places where we were still only propagating Span, cc: @Oberon00

The new Attribute API is a great improvement also, cc: @jkwatson

anuraaga (approved):

anuraaga:
That was intense

Issue #1233, opened 2020-09-22
Muzzle runtime check optimization


trask:
I don't think we need to create (and check) references in `getInstrumentationMuzzle` for OpenTelemetry API classes, since those are shaded and we know those shaded classes are always available.

Same for classes in `instrumentation-api` and `javaagent-instrumentation-api` (currently named `auto-api`).

Here's an example `getInstrumentationMuzzle` from `io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientInstrumentation`:

```
  @Override
  protected synchronized ReferenceMatcher getInstrumentationMuzzle() {
    if (null == this.instrumentationMuzzle) {
      this.instrumentationMuzzle =
          new ReferenceMatcher(
              this.helperClassNames(),
              new Reference[] {
                (new Builder("io.opentelemetry.trace.Span"))
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientHelper",
                        56)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientHelper",
                        33)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientHelper",
                        34)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientTracer",
                        84)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientHelper",
                        32)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.WrappingStatusSettingResponseHandler",
                        32)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientTracer",
                        31)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.WrappingStatusSettingResponseHandler",
                        38)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.ApacheHttpClientHelper",
                        54)
                    .withSource(
                        "io.opentelemetry.instrumentation.auto.apachehttpclient.v4_0.WrappingStatusSettingResponseHandler",
                        39)
[...]

PR #1270, opened 2020-09-27
Add high level Muzzle docs

pavolloffay:
Updates #536

### DIFF
--- a/buildSrc/muzzle.md
+++ b/buildSrc/muzzle.md
@@ -1,0 +1,27 @@
+# Muzzle
+
+Muzzle is a feature of the Java agent that ensures API compatibility
+between libraries/symbols on the application classpath and APIs of instrumented
+3rd party libraries used by the Agent. In other words the Muzzle ensures
+that the API symbols used by the Agent are compatible with API symbols
+on the application classpath. The Muzzle can prevent loading an instrumentation
+if the APIs do not match.
+
+## How does it work
+
+At build time the Muzzle gradle plugin generates matcher for 3rd party APIs used by the agent.
+The matcher is created in the instrumentation class in `ReferenceMatcher getInstrumentationMuzzle()`.
+
+At runtime the Muzzle checks API compatibility between symbols used by the Agent
+and symbols at the application classpath. Before inspecting the classpath
+the Muzzle plugin checks symbols defined in `SomeInstrumentation.typeMatcher()`
+as a performance optimization. If the symbols do not match the instrumentation
+is not loaded.
+
+## Muzzle gradle plugin
+
+The `printReferences` task prints 3rd party API references in a given module.
+
+```bash
+./gradlew :instrumentation:google-http-client-1.19:printReferences
+```
--- a/buildSrc/muzzle.md
+++ b/buildSrc/muzzle.md
@@ -10,7 +10,7 @@ if the APIs do not match.
 ## How does it work
 
 At build time the Muzzle gradle plugin generates matcher for 3rd party APIs used by the agent.
-The matcher is created in the instrumentation class in `ReferenceMatcher getInstrumentationMuzzle()`.
+The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
 
 At runtime the Muzzle checks API compatibility between symbols used by the Agent
 and symbols at the application classpath. Before inspecting the classpath
--- a/docs/contributing/muzzle.md
+++ b/docs/contributing/muzzle.md
@@ -4,16 +4,16 @@ Muzzle is a feature of the Java agent that ensures API compatibility
 between libraries/symbols on the application classpath and APIs of instrumented
 3rd party libraries used by the Agent. In other words the Muzzle ensures
 that the API symbols used by the Agent are compatible with API symbols
-on the application classpath. The Muzzle can prevent loading an instrumentation
+on the application classpath. The Muzzle will prevent loading an instrumentation
 if the APIs do not match.
 
 ## How does it work
 
-At build time the Muzzle gradle plugin generates matcher for 3rd party APIs used by the agent.
+At build time the Muzzle gradle plugin generates a matcher for 3rd party APIs used by the agent.
 The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
 
 At runtime the Muzzle checks API compatibility between symbols used by the Agent
-and symbols at the application classpath. Before inspecting the classpath
+and symbols in the application class loader. Before inspecting the class loader
 the Muzzle plugin checks symbols defined in `SomeInstrumentation.typeMatcher()`
 as a performance optimization. If the symbols do not match the instrumentation
 is not loaded.
@@ -25,3 +25,8 @@ The `printReferences` task prints 3rd party API references in a given module.
 ```bash
 ./gradlew :instrumentation:google-http-client-1.19:printReferences
 ```
+
+## Muzzle location
+
+* `buildSrc` - Muzzle Gradle plugin
+* `agent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle` - Muzzle ByteBuddy plugin
--- a/docs/contributing/muzzle.md
+++ b/docs/contributing/muzzle.md
@@ -9,21 +9,31 @@ if the APIs do not match.
 
 ## How does it work
 
-At build time the Muzzle gradle plugin generates a matcher for 3rd party APIs used by the agent.
+At build time, for each instrumentation the Muzzle ByteBuddy plugin collects symbols referring to both internal
+and 3rd party APIs used by the currently processed instrumentation. The reference collection process starts
+from advice classes - values of the map returned by the `Instrumenter.Default#transformers()` method.
+
+All those references are then used to create a `ReferenceMatcher` instance.
 The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
 
 At runtime the Muzzle checks API compatibility between symbols used by the Agent
-and symbols in the application class loader. Before inspecting the class loader
-the Muzzle plugin checks symbols defined in `SomeInstrumentation.typeMatcher()`
-as a performance optimization. If the symbols do not match the instrumentation
-is not loaded.
+and symbols in the application class loader. If the symbols do not match the instrumentation is not loaded.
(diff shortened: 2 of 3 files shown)


buildSrc/muzzle.md
```diff
@@ -0,0 +1,27 @@
+# Muzzle
+
+Muzzle is a feature of the Java agent that ensures API compatibility
+between libraries/symbols on the application classpath and APIs of instrumented
+3rd party libraries used by the Agent. In other words the Muzzle ensures
+that the API symbols used by the Agent are compatible with API symbols
+on the application classpath. The Muzzle can prevent loading an instrumentation
```
trask, line 7:
```suggestion
on the application classpath. The Muzzle will prevent loading an instrumentation
```

buildSrc/muzzle.md
```diff
@@ -0,0 +1,27 @@
...
+if the APIs do not match.
+
+## How does it work
+
+At build time the Muzzle gradle plugin generates matcher for 3rd party APIs used by the agent.
```
trask, line 12:
```suggestion
At build time the Muzzle gradle plugin generates a matcher for 3rd party APIs used by the agent.
```

buildSrc/muzzle.md
```diff
@@ -0,0 +1,27 @@
...
+The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
+
+At runtime the Muzzle checks API compatibility between symbols used by the Agent
+and symbols at the application classpath. Before inspecting the classpath
```
trask, line 16:
```suggestion
and symbols in the application class loader. Before inspecting the class loader
```

trask (approved):
Our first muzzle doc!!! 🥳

I think muzzle is pervasive enough that this doc deserves to be under `docs/contributing`, with a new section/link from `CONTRIBUTING.md`

anuraaga (approved):
Nice, thanks!

docs/contributing/muzzle.md
```diff
@@ -0,0 +1,32 @@
...
+The `printReferences` task prints 3rd party API references in a given module.
+
+```bash
+./gradlew :instrumentation:google-http-client-1.19:printReferences
+```
+
+## Muzzle location
+
+* `buildSrc` - Muzzle Gradle plugin
+* `agent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle` - Muzzle ByteBuddy plugin
```
pavolloffay, line 32: Is this correct? In this location there is `MuzzleGradlePlugin` but it more seems like ByteBuddy plugin as it implements bytebuddy plugin interface.

docs/contributing/muzzle.md
```diff
@@ -0,0 +1,32 @@
...
+if the APIs do not match.
+
+## How does it work
+
+At build time the Muzzle gradle plugin generates a matcher for 3rd party APIs used by the agent.
+The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
+
+At runtime the Muzzle checks API compatibility between symbols used by the Agent
+and symbols in the application class loader. Before inspecting the class loader
+the Muzzle plugin checks symbols defined in `SomeInstrumentation.typeMatcher()`
```
iNikem, line 17: Do you mean `classLoaderMatcher`?

pavolloffay: No I meant this https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/5d2ae07797069106a1f455f722bafece8c238118/instrumentation/google-http-client-1.19/src/main/java/io/opentelemetry/instrumentation/auto/googlehttpclient/GoogleHttpClientInstrumentation.java#L57

docs/contributing/muzzle.md
```diff
@@ -0,0 +1,32 @@
...
+
+The `printReferences` task prints 3rd party API references in a given module.
+
+```bash
+./gradlew :instrumentation:google-http-client-1.19:printReferences
+```
+
+## Muzzle location
+
+* `buildSrc` - Muzzle Gradle plugin
```
tylerbenson, line 31: This part of muzzle isn't really documented above.  This plugin is responsible for downloading the range of libraries from Maven Central and verifying expected compatibility.  If a new incompatible version is published, this is what will break the build.

pavolloffay: Is there also a task to run it explicitly?

docs/contributing/muzzle.md
pavolloffay: cc) @tylerbenson

docs/contributing/muzzle.md
tylerbenson: Yeah, this is what actually runs when you call the `muzzle` task.
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/.circleci/config.yml#L132

That can be called at the top level, or for an individual project.
As an optimization, these tasks are only added to gradle if `muzzle` is explicitly listed as a task, otherwise it slows down the config stage substantially.

docs/contributing/muzzle.md
tylerbenson: The ByteBuddy plugin is what does the source graph traversal to find all the references for purposes of matching at run time as well as build time in the `muzzle` task.

mateuszrzeszutek: Yeah, it's a ByteBuddy plugin. The name's a bit misleading and it should be changed in the future -- though to be honest I don't have any great idea right now.

docs/contributing/muzzle.md
```diff
@@ -0,0 +1,32 @@
...
+
+At runtime the Muzzle checks API compatibility between symbols used by the Agent
+and symbols in the application class loader. Before inspecting the class loader
+the Muzzle plugin checks symbols defined in `SomeInstrumentation.typeMatcher()`
+as a performance optimization. If the symbols do not match the instrumentation
+is not loaded.
+
+## Muzzle gradle plugin
+
+The `printReferences` task prints 3rd party API references in a given module.
```
mateuszrzeszutek, line 23: It prints all references that are validated by muzzle -- not only the 3rd party ones.

docs/contributing/muzzle.md
mateuszrzeszutek, line 17: Muzzle actually collects symbols from advice classes -- values of the map returned by `transformers()` method. Muzzle by itself calls neither `classLoaderMatcher()` nor `typeMatcher()`.

docs/contributing/muzzle.md
```diff
@@ -0,0 +1,32 @@
...
+between libraries/symbols on the application classpath and APIs of instrumented
+3rd party libraries used by the Agent. In other words the Muzzle ensures
+that the API symbols used by the Agent are compatible with API symbols
+on the application classpath. The Muzzle will prevent loading an instrumentation
+if the APIs do not match.
+
+## How does it work
+
+At build time the Muzzle gradle plugin generates a matcher for 3rd party APIs used by the agent.
+The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
```
mateuszrzeszutek, line 13: Can you add a sentence explaining where do the symbols come from?

```suggestion
At build time, for each instrumentation the Muzzle ByteBuddy plugin collects symbols referring to both internal and 3rd party APIs used by the currently processed instrumentation. The reference collection process starts from advice classes - values of the map returned by the `Instrumenter.Default#transformers()` method.
All those references are then used to create a `ReferenceMatcher` instance.
The matcher is stored in the instrumentation class in method `ReferenceMatcher getInstrumentationMuzzle()`.
```

mateuszrzeszutek (approved):

docs/contributing/muzzle.md
trask: this is nice, i'm not sure about "internal" APIs though, we do currently, but hopefully we won't after #1233?

docs/contributing/muzzle.md
```diff
@@ -0,0 +1,32 @@
...
+
+At runtime the Muzzle checks API compatibility between symbols used by the Agent
+and symbols in the application class loader. Before inspecting the class loader
+the Muzzle plugin checks symbols defined in `SomeInstrumentation.typeMatcher()`
+as a performance optimization. If the symbols do not match the instrumentation
+is not loaded.
```
trask, line 19: another suggestion 😄

```suggestion
and symbols in the application class loader. If the symbols do not match the instrumentation is not loaded.
Because the muzzle matcher is expensive, it is only performed after a match has been made by the `classLoaderMatcher()` and `typeMatcher()` matchers. 
```

docs/contributing/muzzle.md
trask: I'm hoping we can optimize it to only check 3rd party ones

docs/contributing/muzzle.md
pavolloffay: We could rename it to `MuzzleByteBuddyPlugin`

docs/contributing/muzzle.md
pavolloffay: ack, changing it to all APIs for now.

docs/contributing/muzzle.md
pavolloffay: @trask mentioned that muzzle first checks API referenced in `typeMatcher()` for optimization purposes

docs/contributing/muzzle.md
pavolloffay: +1. The match for `classLoaderMatcher()` and `typeMatcher()` is done by ByteBuddy right?

pavolloffay:
@tylerbenson @iNikem @mateuszrzeszutek @trask @anuraaga thanks for the feedback. The PR is updated with your suggestions.

docs/contributing/muzzle.md
mateuszrzeszutek: Hopefully we'll get rid of at least most of these

docs/contributing/muzzle.md
mateuszrzeszutek: Yeah, it's done a few steps before the `MuzzleMatcher` runs.

mateuszrzeszutek (approved):

trask:
@pavolloffay can you run `spotlessApply`, and merge master into your branch to get the build fix from #1280? then I think this is good to merge!

docs/contributing/muzzle.md
trask: I created #1288 for renaming it

trask:
Thanks again for this 💯

PR #1275, opened 2020-09-28
Convert more decorators to tracers

iNikem:


instrumentation/spring/spring-scheduling-3.1/src/main/java/io/opentelemetry/instrumentation/auto/spring/scheduling/SpringSchedulingRunnableWrapper.java
```diff
@@ -34,18 +35,14 @@ public void run() {
     if (runnable == null) {
       return;
     }
-    Span span =
-        TRACER.spanBuilder(SpringSchedulingDecorator.DECORATE.spanNameOnRun(runnable)).startSpan();
-    SpringSchedulingDecorator.DECORATE.afterStart(span);
+    Span span = TRACER.startSpan(TRACER.spanNameOnRun(runnable), Kind.INTERNAL);
```
trask, line 38: Should this be `Kind.SERVER` since it's a "root" span?

trask (approved):

instrumentation/spring/spring-scheduling-3.1/src/main/java/io/opentelemetry/instrumentation/auto/spring/scheduling/SpringSchedulingRunnableWrapper.java
iNikem: It was `INTERNAL` before, I haven't changed that. But you have a point... I don' know :)

anuraaga: I don't think scheduling tasks are really server (no chance of ever being connected to a client). Seems ok to me.

anuraaga (approved):

Issue #1277, opened 2020-09-28
Remove instrumentation (and tooling) specific properties from Config


mateuszrzeszutek:
The aim of this issue is to clean up `Config` by removing all property fields and getters that are used by specific instrumentation - or by the `java-tooling` module (as they shouldn't be exposed to the API). Also, we should include a recommendation to store the result of `Config.get().getProperty()` in a (static?) field if it's going to be used after the agent initialization.

See comments:
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254#issuecomment-698805524
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254#issuecomment-699109191
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254#discussion_r495769091
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254#discussion_r495774331

mateuszrzeszutek:
Can anybody assign this to me? (unfortunately I can't do it)
Thanks!

PR #1286, opened 2020-09-29
Remove instrumentation (and tooling) specific properties from Config

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1277

I like how the `Config` class looks like now 😄

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
@@ -19,12 +19,12 @@ package io.opentelemetry.instrumentation.api.config;
 import static java.util.Objects.requireNonNull;
 
 import com.google.auto.value.AutoValue;
-import io.grpc.Context;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.SortedSet;
+import java.util.function.Function;
 import java.util.regex.Pattern;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.slf4j.Logger;
@@ -35,23 +35,7 @@ public abstract class Config {
   private static final Logger log = LoggerFactory.getLogger(Config.class);
   private static final Pattern PROPERTY_NAME_REPLACEMENTS = Pattern.compile("[^a-zA-Z0-9.]");
 
-  // locations where the context was propagated to another thread (tracking multiple steps is
-  // helpful in akka where there is so much recursive async spawning of new work)
-  public static final Context.Key<List<StackTraceElement[]>> THREAD_PROPAGATION_LOCATIONS =
-      Context.key("thread-propagation-locations");
-  public static final boolean THREAD_PROPAGATION_DEBUGGER =
-      Boolean.getBoolean("otel.threadPropagationDebugger");
-
-  public static final String DEFAULT_EXPORTER = "otlp";
-  public static final boolean DEFAULT_TRACE_ENABLED = true;
-  public static final boolean DEFAULT_INTEGRATIONS_ENABLED = true;
-  public static final boolean DEFAULT_RUNTIME_CONTEXT_FIELD_INJECTION = true;
-  public static final boolean DEFAULT_TRACE_EXECUTORS_ALL = false;
-  public static final boolean DEFAULT_SQL_NORMALIZER_ENABLED = true;
-  public static final boolean DEFAULT_KAFKA_CLIENT_PROPAGATION_ENABLED = true;
-  public static final boolean DEFAULT_HYSTRIX_TAGS_ENABLED = false;
-
-  private static final Config DEFAULT = Config.newBuilder().build();
+  private static final Config DEFAULT = Config.create(Collections.emptyMap());
 
   // INSTANCE can never be null - muzzle instantiates instrumenters when it generates
   // getInstrumentationMuzzle() and the Instrumenter.Default constructor uses Config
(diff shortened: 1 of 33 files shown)


mateuszrzeszutek:
Regarding https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1254#discussion_r495769091: I've checked all usages and everything besides the Kafka one looks OK. In case of Kafka unit tests change the configuration over and over again during a single scenario which makes it impossible to store the boolean flag in a simple static final constant. I've tried several things but nothing worked out. I guess we're left with three options:
1) leave it like it is;
2) add a static getter method and use PowerMock;
3) rewrite the unit tests from scratch.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/ConfigBuilder.java
```diff
@@ -16,53 +16,15 @@
...
-import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Properties;
-import java.util.function.Function;
-import org.checkerframework.checker.nullness.qual.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public final class ConfigBuilder
```
trask, line 26: Is this class still needed/useful?

testing-common/src/main/groovy/io/opentelemetry/auto/test/AgentTestRunner.java
```diff
@@ -102,7 +102,7 @@
   private static volatile ClassFileTransformer activeTransformer = null;
 
   static {
-    INSTRUMENTATION = ByteBuddyAgent.getInstrumentation();
+    INSTRUMENTATION = ByteBuddyAgent.install();
```
trask, line 105: I'm curious the reason for this change?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.List;
+
+public final class ContextPropagationDebug {
+
+  // locations where the context was propagated to another thread (tracking multiple steps is
+  // helpful in akka where there is so much recursive async spawning of new work)
+  public static final Context.Key<List<StackTraceElement[]>> THREAD_PROPAGATION_LOCATIONS =
+      Context.key("thread-propagation-locations");
+  public static final boolean THREAD_PROPAGATION_DEBUGGER =
+      Boolean.getBoolean("otel.threadPropagationDebugger");
```
trask, line 29: maybe make fields private with static accessors, e.g. `isEnabled()`, `getLocation(Context)`, `withLocation(List<StackTraceElement[]>, Context)`

testing-common/src/main/groovy/io/opentelemetry/auto/test/AgentTestRunner.java
trask: oh nm, I see `install()` was moved here from `AgentSpecification` :thumbsup:

anuraaga (commented):
Thanks @mateuszrzeszutek - I think for Kafka, if the tests can be rewritten to only set config once per test, which I guess is a normal pattern, that would be nice, but can be another PR

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -35,23 +35,7 @@
...
-  public static final boolean DEFAULT_TRACE_ENABLED = true;
-  public static final boolean DEFAULT_INTEGRATIONS_ENABLED = true;
-  public static final boolean DEFAULT_RUNTIME_CONTEXT_FIELD_INJECTION = true;
-  public static final boolean DEFAULT_TRACE_EXECUTORS_ALL = false;
-  public static final boolean DEFAULT_SQL_NORMALIZER_ENABLED = true;
-  public static final boolean DEFAULT_KAFKA_CLIENT_PROPAGATION_ENABLED = true;
-  public static final boolean DEFAULT_HYSTRIX_TAGS_ENABLED = false;
-
-  private static final Config DEFAULT = Config.newBuilder().build();
+  private static final Config DEFAULT = Config.create(Collections.emptyMap());
```
iNikem, line 38: So default config is one without any configuration? Maybe it should use old default values in some way?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
iNikem: Specification explicitly advices against exposing context keys.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
```diff
@@ -17,10 +17,10 @@
 package io.opentelemetry.instrumentation.api.decorator;
 
 import static io.opentelemetry.OpenTelemetry.getPropagators;
+import static io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils.ENDPOINT_PEER_SERVICE_MAPPING;
```
iNikem, line 20: Looks like more than `Utils` now

instrumentation/jdbc/src/test/groovy/SqlNormalizerTest.groovy
```diff
@@ -139,19 +137,4 @@ class SqlNormalizerTest extends AgentSpecification {
       JDBCUtils.normalizeSql(sb.toString())
     }
   }
-
-  def "config can disable sql normalizer"() {
```
iNikem, line 143: Why removed?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Instrumenter.java
```diff
@@ -237,7 +237,7 @@ protected ReferenceMatcher getInstrumentationMuzzle() {
     }
 
     protected boolean defaultEnabled() {
-      return Config.get().isIntegrationsEnabled();
+      return Config.get().getBooleanProperty("otel.integrations.enabled", true);
```
iNikem, line 240: This should be cached, I think.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/ConfigBuilder.java
mateuszrzeszutek: I think it is: it allows us to reuse SDK `ConfigBuilder`'s environment variable and system property mechanism, it is a single place where we can specify the property reading order.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
mateuszrzeszutek: Very good point, fixed.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
mateuszrzeszutek: The default values are now specified by `Config` callers, so there's no need to keep anything here -- which makes sense, because we wouldn't want to keep a default value for some Kafka property here.

mateuszrzeszutek:
> f the tests can be rewritten to only set config once per test, which I guess is a normal pattern, that would be nice, but can be another PR

I agree -- after thinking about it I believe this is the way to go. Especially if we want to use full agent jar in tests -- in that case dynamically changing property value in the middle of the scenario would be pretty difficult.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/decorator/BaseDecorator.java
mateuszrzeszutek: Yes - but on the other hand, once we remove all `BaseDecorator` extensions this constant will go back to being private.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Instrumenter.java
mateuszrzeszutek: It probably should - but its value is modified in the unit test heavily, so I've decided to just leave it here. This method is only called during the instrumenter initialization, so calling the `Config` here adds no overhead once the agent starts.

instrumentation/jdbc/src/test/groovy/SqlNormalizerTest.groovy
mateuszrzeszutek: I had two options here:

1) store the property in a static final constant and remove this test (changing property value would require unloading & loading `JDBCUtils`);
2) leave this test and call `Config` on each `normalizeSql()` call.

I suspect that `normalizeSql()` can be called very often (each SQL query), so I've decided to choose the first option for performance reasons.

trask (approved):
👍

PR #1291, opened 2020-09-30
Remove AgentSpecification

anuraaga:
Config isn't `final` anymore, thanks @mateuszrzeszutek! And the system property we can set with Gradle to avoid ordering constraints.


iNikem (approved):

mateuszrzeszutek (approved):

trask:
Nice 👍

PR #1304, opened 2020-10-02
Update to latest snapshot.

anuraaga:
This is as far as I was able to get today. I added more version pinning than technically needed while troubleshooting baggage not class loadable issue, but I guess it's nice to have as much pinned properly as possible anyways.

Tests fail with more traces being reported than expected (e.g, both server and client span for client test). Not sure why this change would affect that - if anyone has time to run with this branch that's cool, otherwise I'll take another look later.

### DIFF
--- a/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
+++ b/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
@@ -60,8 +60,8 @@ public class AutoInstrumentationPlugin implements Plugin<Project> {
         "io.opentelemetry.instrumentation.api",
         "io.opentelemetry.OpenTelemetry", // OpenTelemetry API
         "io.opentelemetry.common", // OpenTelemetry API
+        "io.opentelemetry.baggage", // OpenTelemetry API
         "io.opentelemetry.context", // OpenTelemetry API (context prop)
-        "io.opentelemetry.correlationcontext", // OpenTelemetry API
         "io.opentelemetry.internal", // OpenTelemetry API
         "io.opentelemetry.metrics", // OpenTelemetry API
         "io.opentelemetry.trace", // OpenTelemetry API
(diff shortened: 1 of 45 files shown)


trask:
hey @anuraaga! I just opened https://github.com/open-telemetry/opentelemetry-java/issues/1742 for the reason why these tests are failing.

For now at least, looks like we need to configure w3c propagator in AgentTestRunner and TracerInstaller.

On another topic in this PR, my reading of https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#statuscanonicalcode is that we should be setting `ERROR` wherever we were previously setting `UNKNOWN` (or other error conditions), and we should only remove `setStatus` where we were previously setting it to `OK`.

anuraaga:
@trask Not sure, it says "The status code should only be set to ERROR according to the rules defined within the semantic conventions." The only convention I could find was HTTP. @iNikem Do you know how to record the new codes?

trask:
Agree I maybe read the spec with some bias 😂.

What I hope, is that we should set span status to ERROR if we think backends should display that span as an error (I think most backends have this concept, e.g. display the span in red).

Ya, @iNikem will know.

trask:
> if anyone has time to run with this branch that's cool

I have time, will see about finishing this up since master is failing due to jfrog snaphot pruning.

Not sure what to do about setting/not setting ERROR. I think setting ERROR will be closest to what we have currently, and will be easy to remove later as we get clarification (as opposed to removing it now and needing to add it back later), so will probably go that route for now.

anuraaga (commented):
Thanks for the help! LGTM.

I presume we'll revisit the status stuff anyways when Nikita creates a table for that one too :D

trask (approved):

iNikem:
@trask @anuraaga @jkwatson Can we talk about this error status point during our common meeting?

trask:
> @trask @anuraaga @jkwatson Can we talk about this error status point during our common meeting?

Added to [agenda](https://docs.google.com/document/d/1WK9h4p55p8ZjPkxO75-ApI9m0wfea6ENZmMoFRvXSCw) 👍

PR #1305, opened 2020-10-02
Compile error workaround

iNikem:


instrumentation/spring/spring-boot-autoconfigure/spring-boot-autoconfigure.gradle
```diff
@@ -17,6 +17,9 @@ dependencies {
   implementation project(':instrumentation:spring:spring-webmvc-3.1:library')
   implementation project(':instrumentation:spring:spring-webflux-5.0:library')
 
+  //This SDK dependency should not be here, but exporter deps below bring sdk-SNAPSHOT, not fixed-timed
```
anuraaga, line 20: Ah yeah I noticed this too

Can you update the constraints instead? That's how we're managing transitive dependency.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1304/files#diff-bcb9553c38af5f637c5a9f1c0b006f6f

anuraaga (approved):

instrumentation/spring/spring-boot-autoconfigure/spring-boot-autoconfigure.gradle
iNikem: done

PR #1309, opened 2020-10-02
Add system metrics

malafeev:
It is based on https://github.com/open-telemetry/opentelemetry-java/pull/1323
Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/system-metrics.md


malafeev:
cc @carlosalberto

iNikem:
@jkwatson do you want to review this? Metrics are your cup of tea, aren't they?

carlosalberto:
Looks good to me! I'm just wondering where/how should we initialize it, as I'd like this to be part of the rest of the instrumentation.

carlosalberto (approved):

trask:
hey @malafeev! sorry for the delay, will review this by end of the week

system-metrics/system-metrics.gradle
```diff
@@ -0,0 +1,11 @@
+apply from: "$rootDir/gradle/java.gradle"
+apply from: "$rootDir/gradle/publish.gradle"
+
+
+dependencies {
+  implementation deps.opentelemetryApi
+  implementation "com.github.oshi:oshi-core:5.2.5"
```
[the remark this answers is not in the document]
carlosalberto: An alternative to this would be to add this but:

1) Disable it by default AND
2) Make oshi a compile dependency.

This way, if users want to leverage it, they'd need to explicitly enable this instrumentation part, AND provide the Oshi dependency themselves.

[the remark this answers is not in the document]
anuraaga: I think it's a good idea to make oshi a compile dependency. How about we model this change exactly like any instrumentation, `instrumentation/oshi/{library,javaagent}` for example. This code goes in library, giving user a manual way of instrumenting (calling the static method) similar to our other library instrumentations, and auto can call it automatically if oshi is found on the classpath (not sure if we have a pattern for this yet but figure it's possible right @trask?). We can start with just moving this into library and add javaagent later.

For reference, https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/grpc-1.5/library/grpc-1.5-library.gradle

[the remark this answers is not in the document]
anuraaga: In which case we should split out the JVM metrics into a separate `instrumentation/jvm/library` too

iNikem:
Why this PR is blocked from merging? We do have one approval from @carlosalberto ... @trask do you still want to review this?

system-metrics/system-metrics.gradle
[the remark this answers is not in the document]
carlosalberto: > This code goes in library, giving user a manual way of instrumenting (calling the static method) similar to our other library instrumentations

Sounds good to me.

> In which case we should split out the JVM metrics into a separate instrumentation/jvm/library too

Same.

@malafeev any opinion on this? ;)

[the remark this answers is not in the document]
malafeev: I'm fine with it. any other opinions?

jkwatson:
> @jkwatson do you want to review this? Metrics are your cup of tea, aren't they?

Don't hold this up on my account. :)

system-metrics/system-metrics.gradle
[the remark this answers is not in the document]
trask: I like the idea of adding just library instrumentation in the first PR, and then work on how to integrate it into javaagent after that

trask:
> Why this PR is blocked from merging?

@iNikem it's blocked from merging because it hasn't been approved by someone in the CODEOWNERS file

system-metrics/system-metrics.gradle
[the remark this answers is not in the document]
malafeev: I moved code to `instrumentation/oshi/library` because both jvm and system metrics depends on oshi

malafeev:
@trask could you review this PR again?

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/JavaMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+import io.opentelemetry.metrics.AsynchronousInstrument.LongResult;
+import io.opentelemetry.metrics.Meter;
+import java.lang.management.GarbageCollectorMXBean;
+import java.lang.management.ManagementFactory;
+import oshi.SystemInfo;
+import oshi.software.os.OSProcess;
+import oshi.software.os.OperatingSystem;
+
+/** Java Runtime Metrics Utility */
+public class JavaMetrics {
```
trask, line 21: maybe rename to `ProcessMetrics` to make clear these are not just standard Java-provided metrics?

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/JavaMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+              public void update(LongResult r) {
+                long gcCount = 0;
+                for (final GarbageCollectorMXBean gcBean :
+                    ManagementFactory.getGarbageCollectorMXBeans()) {
+                  gcCount += gcBean.getCollectionCount();
+                }
+
+                r.observe(gcCount, Labels.of(TYPE_LABEL_KEY, "count"));
+              }
+            });
```
trask, line 80: can this be part of the other (future) module that only depends on JVM-provided metrics?

trask (commented):
thx @malafeev! sorry for the delayed review

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/JavaMetrics.java
malafeev: It depends on oshi.
If I will move it to separate module it will have the same dependency on oshi as system metrics.
Activation will be the same: presence of oshi dependency.
Therefore I doubt to split it to two modules.

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/JavaMetrics.java
malafeev: renamed

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/JavaMetrics.java
trask: `runtime.java.gc_count` metric has dependency on oshi?

malafeev: oh, you mean only this specific metric, I thought you are talking about all jvm metrics here.

malafeev: I can create separate `instrumentation/jvm/libarary` with only this metric.

trask: 👍 how about removing `runtime.java.gc_count` from this PR and then we'll merge this (sorry again it took so long!), and u can send separate PR with `instrumentation/jvm/library`?

malafeev: removed

trask (approved):
thx @malafeev 👍

trask:
@malafeev can you merge master into your branch? we updated to otel api 0.10.0-SNAPSHOT, which breaks a couple of metric APIs, just in case it affects this PR

malafeev:
@trask synced with master

trask:
@malafeev looks like the branch needs a `./gradlew spotlessApply`

malafeev:
@trask master needs a `./gradlew spotlessApply` :)

trask:
oh no! I think I messed up with the merge enforcement switching over from CircleCI to GHA, which probably led to that

anuraaga (approved):
Thanks!

Issue #1306, opened 2020-10-02
:instrumentation-api:test fail


iNikem:
After recent changes in how `Config` works I started to get test failures locally. E.g. for `NetPeerUtilsTest`:

```
Too few invocations for:

1 * span.setAttribute(SemanticAttributes.PEER_SERVICE, expectedPeerService)   (0 invocations)

Unmatched invocations (ordered by similarity):

1 * span.setAttribute(AttributeKeyImpl{getType=STRING, key=net.peer.ip}, '1.2.3.4')
One or more arguments(s) didn't match:
0: argument == expected
   |        |  |
   |        |  AttributeKeyImpl{getType=STRING, key=peer.service}
   |        false
   AttributeKeyImpl{getType=STRING, key=net.peer.ip}
1: argument == expected
   |        |  |
   1.2.3.4  |  catservice
            false
            10 differences (0% similarity)
            (1.2.3.4---)
            (catservice)
1 * span.setAttribute(AttributeKeyImpl{getType=LONG, key=net.peer.port}, 888)
One or more arguments(s) didn't match:
0: argument == expected
   |        |  |
   |        |  AttributeKeyImpl{getType=STRING, key=peer.service}
   |        false
   AttributeKeyImpl{getType=LONG, key=net.peer.port}
1: argument == expected
   |        |  |
   888      |  catservice
            false


	at org.spockframework.mock.runtime.InteractionScope.verifyInteractions(InteractionScope.java:98)
	at org.spockframework.mock.runtime.MockController.leaveScope(MockController.java:77)
	at io.opentelemetry.instrumentation.api.tracer.NetPeerUtilsTest.test setAttributes with mapped peer(NetPeerUtilsTest.groovy:58)
```

There are 3 tests in `instrumentation-api` module which all set one and the same system property to different values. It seems that during parallel/concurrent test invocation they step on each others toes. Interestingly enough, they seem to work in our CI, probably due to different level of parallelism.

Issue #1146, opened 2020-09-01
Run tests with full agent jar


anuraaga:
This could probably come after supporting running tests with a mini module-specific agent jar. We should also have a mode, probably built in the nightly build, to run tests with the full agent jar. Currently, we have no way of catching simple packaging issues like instrumentation not being bundled at all.

iNikem:
See #1422 . That bug would be found by our tests if we used full agent jar.

iNikem:
Btw, @anuraaga have you thought already how will we get hold of the traces produced by tests?

trask:
in-memory otlp listener launched by the test?

iNikem:
> in-memory otlp listener launched by the test?

You mean actually listening on a network port and accepting OTLP?

anuraaga:
@iNikem I missed replying to this issue - yeah I was thinking using network as last resort should always be possible. But we have some test-specific logic in the agent related too bootstrap classloader, so assuming that's an ok thing and not something to remove completely, another approach that came to mind is test-specific logic to have spans exported to a list in the bootstrap classloader and just read directly.

trask:
> have spans exported to a list in the bootstrap classloader and just read directly.

the only complication is that the SDK classes (e.g. `SpanData`) live in AgentClassLoader, so we can't use them directly during test verification

trask:
Adding a note here, when we do this, to make sure that the instrumentation classes (`src/main/java`) are not on the classpath when running the tests, in order to catch problems like https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1585#issuecomment-723670977

anuraaga:
> the only complication is that the SDK classes (e.g. SpanData) live in AgentClassLoader, so we can't use them directly during test verification

Thought a bit about this, and perhaps this isn't a big deal if the spans are exported to the bootstrap classloader as serialized protos - test would be able to deserialize them trivially without any shading issues.

anuraaga:
And another thought, not sure about it, but if this test-exporter used the exporter_jar mechanism, IIUC, getting the spans from tests could be as simple as the agent providing a reference to the exporter classloader through the bootstrap. We can make it package-private and still access it through reflection from tests I think. Then classload the test exporter, which would have methods for accessing the serialized protos.

anuraaga:
[...]

PR #1310, opened 2020-10-02
Refactor all tests that use Config so that they don't fail locally

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1306

> There are 3 tests in instrumentation-api module which all set one and the same system property to different values. It seems that during parallel/concurrent test invocation they step on each others toes. Interestingly enough, they seem to work in our CI, probably due to different level of parallelism.

Why did it work on the CI: we use retries on the CI, but not when running tests locally. I suspect that Gradle only re-ran the failing tests (since that would be the sane thing to do), so with each retry `NetPeerUtils` would be loaded again, but with different property value. And so on until it passed - we use 5 retries (`java.gradle`) so that should be enough.

I've decided to refactor all tests that modify `Config` to avoid the same situation in different places:
* for modules that did not need to modify the test configuration I've used the newly-added config SPI to provide static config;
* those that needed to modify the config still use `ConfigUtils`, but I've refactored it and added Javadocs. I hoped that I could get rid of `AgentTestRunner.resetInstrumentation()`, but not in this PR unfortunately.


trask (commented):
Using the new config SPI is nice, but I think will require separate test sets for each different configuration that we want to test with, which feels a little awkward to have to split out test sets for different configuration values.

Also, I kind of like keeping the configuration close to the test that uses it.

And it's nice to have one way to deal with test configuration instead of two.

So I wonder if it's better to use `ConfigUtils` everywhere for now, until we figure out our desired end state?

One option in the future (which would get rid of `resetInstrumentation`) is to have  `Config.getBooleanProperty()` return a `BooleanConfigValue` (etc) and by default `BooleanConfigValue` can just have a `final boolean value` in it, but when running tests we could sub out a different impl that always reads from `Config.allProperties`, so we can update freely at runtime. Instrumentations would have to cache the `BooleanConfigValue` instead of the raw `boolean`, but the default impl should perform well since it's just a holder of a constant (non-volatile) value.

anuraaga (commented):
Agree with @trask that it's probably simpler to stick with the same pattern for static config too. But in general, this looks like a great idea to me, closer to the usual setup/cleanup pattern of saving an old value before tests and restoring it after tests

anuraaga:
@trask `final boolean value` - unfortunately, Java won't do anything quite as cool if it's not `static` XD It's not the end of the world though, avoiding the volatile still helps enough and is a pretty common pattern in Java, just want to point it out. Not sure how much slower `resetInstrumentation` is since it's on the spec level rather than the test level.

iNikem:
> So I wonder if it's better to use `ConfigUtils` everywhere for now, until we figure out our desired end state?

Does anybody have an idea how to solve our current problem when every test can its own, conflicting, configuration?

iNikem:
My only idea is to have two different implementations of Config or Config storage: current one with `io.opentelemetry.instrumentation.api.config.Config#INSTANCE` for production, and another one with e.g. ThreadLocal storage for tests.

WDYT @mateuszrzeszutek @anuraaga @trask ?

mateuszrzeszutek:
> Does anybody have an idea how to solve our current problem when every test can its own, conflicting, configuration?

Exactly. My idea was to remove the conflicts altogether -- the easy fix for this issue was just changing `HttpClientTracerTest` to use the same values as `NetPeerUtilsTest` (and others), so the order in which those test execute wouldn't matter.

> So I wonder if it's better to use `ConfigUtils` everywhere for now

I can remove all SPI calls and use `ConfigUtils` everywhere with consistent configuration settings, but if someone changes `HttpClientTracerTest`'s configuration and some other seemingly unrelated tests start failing the same issue will resurface once again.

> My only idea is to have two different implementations of Config or Config storage: current one with io.opentelemetry.instrumentation.api.config.Config#INSTANCE for production, and another one with e.g. ThreadLocal storage for tests.

Hmm, but this doesn't actually solve the `static final` problem: once you get something from `Config` and set as constant it stays immutable.

I can think of two options here:
* the `BooleanConfigValue` idea is a brilliant one and will certainly work in this situation: `NetPeerUtils` could store sth. like `Supplier<Map<String, String>>` instead of a plain map and the "real" `Config` implementation would always return the same map, the test one would ask `Config` every time;
* in case of `instrumentation-api` tests I can just put all the shared config in one class (which will use `ConfigUtils`) to set them and call this class in `static{}` blocks of test classes that use configuration. I think that all other modules can call `ConfigUtils` directly without any problems.

mateuszrzeszutek:
Also, some explanation regarding `resetInstrumentation()`: I wanted to remove it not only because I don't like it (😄 ) but also because it should make implementing https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1146 significantly easier.

iNikem:
The more I think about it the more I wonder: maybe we indeed use statics too much?

trask:
> Also, some explanation regarding `resetInstrumentation()`: I wanted to remove it not only because I don't like it (😄 ) but also because it should make implementing https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1146 significantly easier. 

💯

on a different topic, an option (if it helps) for more complex config data types that need to parse the string:

```
ConfigValue<T> getConfigValue(Parser<T> parser)
```

this would allow one implementation that calls the parser once and returns an immutable `ConfigValue<T>`, and one implementation (for testing) that returns a `ConfigValue<T>` that internally calls the parser on each access using the latest property string (I don't think we care, but this could be optimized to cache the latest value and only re-parse if the underlying property string has changed)

mateuszrzeszutek:
I've removed all SPI usage and made all tests use `ConfigUtils` for configuration changes. I'd like to try the `ConfigValue` idea in a separate PR - this one is just a fix after all.

iNikem:
Does this fix the original issue? Or if somebody changes config in one of the api tests, it will fail again?

mateuszrzeszutek:
It'll fail if somebody changes it.

iNikem (approved):

instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/HttpClientTracerTest.groovy
```diff
@@ -5,28 +5,29 @@
...
-    updateConfig {
-      System.setProperty(
-        "otel.endpoint.peer.service.mapping",
-        "myhost-mapped=reservation-service")
-    }
+  static final PREVIOUS_CONFIG = ConfigUtils.updateConfig {
+    it.setProperty(
+      "otel.endpoint.peer.service.mapping",
+      "1.2.3.4=catservice,dogs.com=dogsservice,opentelemetry.io=specservice")
+  }
```
trask, line 20:
```suggestion
  static final PREVIOUS_CONFIG = ConfigUtils.updateConfig {
    it.setProperty(
      "otel.endpoint.peer.service.mapping",
      "dogs.com=dogsservice")
  }
```

instrumentation/jms-1.1/src/jms2Test/groovy/JMS2Test.groovy
```diff
@@ -34,6 +35,11 @@ import org.hornetq.jms.client.HornetQTextMessage
 import spock.lang.Shared
 
 class JMS2Test extends AgentTestRunner {
+  static final PREVIOUS_CONFIG = ConfigUtils.updateConfigAndResetInstrumentation {
+    it.setProperty("otel.trace.classes.exclude",
+      "org.springframework.jms.config.JmsListenerEndpointRegistry\$AggregatingCallback,"
+        + "org.springframework.context.support.DefaultLifecycleProcessor\$1")
+  }
```
trask, line 42: Is this needed?
```suggestion
```

instrumentation/jms-1.1/src/jms2Test/groovy/SpringListenerJMS2Test.groovy
```diff
@@ -15,6 +16,15 @@ import org.springframework.jms.core.JmsTemplate
...
 class SpringListenerJMS2Test extends AgentTestRunner {
+  static final PREVIOUS_CONFIG = ConfigUtils.updateConfigAndResetInstrumentation {
+    it.setProperty("otel.trace.classes.exclude",
+      "org.springframework.jms.config.JmsListenerEndpointRegistry\$AggregatingCallback,"
+        + "org.springframework.context.support.DefaultLifecycleProcessor\$1")
+  }
+
+  def cleanupSpec() {
+    ConfigUtils.setConfig(PREVIOUS_CONFIG)
+  }
```
trask, line 27: is this needed?
```suggestion
```

instrumentation/jms-1.1/src/jms2Test/groovy/SpringTemplateJMS2Test.groovy
```diff
@@ -28,6 +29,11 @@ import org.springframework.jms.core.JmsTemplate
 import spock.lang.Shared
 
 class SpringTemplateJMS2Test extends AgentTestRunner {
+  static final PREVIOUS_CONFIG = ConfigUtils.updateConfigAndResetInstrumentation {
+    it.setProperty("otel.trace.classes.exclude",
+      "org.springframework.jms.config.JmsListenerEndpointRegistry\$AggregatingCallback,"
+        + "org.springframework.context.support.DefaultLifecycleProcessor\$1")
+  }
```
trask, line 36: same
```suggestion
```

instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/HttpClientTracerTest.groovy
mateuszrzeszutek: I can only remove `specservice`, `catservice` is used by e.g. `NetPeerUtilsTest`. Will be fixed after next PR (`ConfigValue`!)

instrumentation/jms-1.1/src/jms2Test/groovy/SpringListenerJMS2Test.groovy
mateuszrzeszutek: Nope, not needed anymore.

mateuszrzeszutek:
I've created a new issue for the next `Config` refactoring: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1345

iNikem:
@anuraaga @trask do you approve?

trask (approved):

PR #1323, opened 2020-10-05
Allow parenting with the AWS Lambda span itself.

anuraaga:
AWS Lambda sets an environment variable with a trace header.

For HTTP requests, this is the header for a span which links back to an original HTTP request, for example from API gateway. For X-Ray users, parenting to this will connect back to the client through intermediate spans exported to X-Ray, while showing lambda processing in between. For non-X-ray users, these intermediate spans won't exist and thus the parent must be the client span itself, not the lambda span so the lambda span needs to be ignored.

For SQS requests, this is the header for a span which is a local root for message processing. It never hurts to parent to this, and links will be added which can be used by other tracing backends without problem.

While I considered adding a special configuration for whether to do this parenting or not, it seems that just configuring the propagator works fine so I documented that behavior instead. It means that if a user isn't using X-ray as a backend but wants to use X-Ray trace propagation format for their HTTP requests, it's not supported - I don't think such a user exists in practice though. Let me know if you have any thoughts.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -32,7 +32,8 @@ ext {
     checkerFramework    : "3.6.1",
     errorprone          : "2.4.0",
     nullaway            : "0.8.0",
-    autoValue           : "1.7.4"
+    autoValue           : "1.7.4",
+    systemLambda        : "1.1.0",
   ]
 
   deps = [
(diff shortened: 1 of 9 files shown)


trask (approved):
> It means that if a user isn't using X-ray as a backend but wants to use X-Ray trace propagation format for their HTTP requests, it's not supported - I don't think such a user exists in practice though

Makes sense to me. Thanks for the detailed explanations.

PR #1324, opened 2020-10-05
Run muzzle during PR

iNikem:
Every module as a separate job

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -72,7 +72,7 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.4
         with:
           job-id: muzzle${{ matrix.module }}
-          arguments: :instrumentation:${{ matrix.module }}:muzzle --no-daemon
+          arguments: ${{ matrix.module }}:muzzle --no-daemon
 
 #  muzzle:
 #    runs-on: ubuntu-latest
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,6 +1,6 @@
 name: PR build
 
-on: push
+on: pull_request
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


anuraaga:
I don't think our config for muzzle has enabled caching but I see `FROM-CACHE` in the logs. Do you know how this is happening?

anuraaga:
Do you mind doing a run with cache disabled to measure the time? Some changes, like updating Gradle, will invalidate all cache. That doesn't mean we have to stick with CircleCi but knowing how the worst case does will be helpful. For example, if worst case of CircleCi is 30min and GHA is 4 hours, I don't know what to do.

iNikem:
Buildscan claims that we saved 15s due to remote cache. It just needs to compile couple of modules. Probably tests will still be the longest checks

iNikem:
@anuraaga Tried. Tests are still the longest part.

anuraaga:
@iNikem Something magical is going on or I'm not being able to read the logs correctly - isn't https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1324/commits/ab6b2cab66dffb4357fe9570a89a1c8adccb6410 also printing a lot of FROM-CACHE?

iNikem:
Screenshot 2020-10-05 at 18 47 51

Don't see

anuraaga:
Perhaps it depends on the version. I think your icon on this screenshot makes it clear right? :-)

image

anuraaga:
Ah - nevermind, why am I not looking at muzzle. Got it

anuraaga:
Here's a muzzle with cache (fairly confident this time but still may be doing something silly :P). Any idea why it could be? That being said, of course if unmuzzled cache is still fast it's ok, but TBH I'm just a bit surprised since muzzle seems so intence.

image

anuraaga (approved):
Why is muzzle so fast even though it seems like it should be really expensive?

trask (commented):
Nice 😄
 
image

trask (approved):

Issue #891, opened 2020-08-04
Confusing how to run tests for modules with maxJavaVersionForTests = JavaVersion.VERSION_1_8


trask:
E.g. it's confusing how to run `netty-4.0` tests.

Running them in Intellij reports success but no tests are run.

Running them via normal command line:

```
./gradlew :instrumentation:netty:netty-4.0:test
```

also succeeds but no tests are run.

The "correct" way currently is:

```
export JAVA_8_HOME=...
./gradlew :instrumentation:netty:netty-4.0:testJava8
```

At least we can document/highlight this in [running-tests.md](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/running-tests.md).

But also, @anuraaga mentioned at one point about automatically downloading and configuring Java versions to run with. I think this could give us a really nice solution. Especially if we can automatically run every test using the `minJavaVersionForTests` when you run the normal `test` task.

anuraaga:
Gradle 6.7 will support automatically downloading Java for running tests, can't wait!

https://docs.gradle.org/6.7-rc-1/userguide/toolchains.html#specify_custom_toolchains_for_individual_tasks

iNikem:
That's cool :)

PR #1325, opened 2020-10-05
fix and/or ignore spotbugs errors

imavroukakis:
There are a couple of spotbugs complaints that I'd like some feedback on

```
H C NP: Null passed for non-null parameter of io.opentelemetry.trace.Span.setAttribute(AttributeKey, Object) in io.opentelemetry.instrumentation.auto.grpc.common.GrpcHelper.prepareSpan(Span, String, InetSocketAddress, boolean)  Method invoked at GrpcHelper.java:[line 62]
```
API for `setAttribute` strongly advises against `null`. If it's still ok to pass `null` here we can add a spotbugs exclusion otherwise, is an empty string an acceptable value?

---

```
H B Nm: The class name io.opentelemetry.javaagent.tooling.config.ConfigBuilder shadows the simple name of the superclass io.opentelemetry.sdk.common.export.ConfigBuilder  At ConfigBuilder.java:[lines 26-61]
```
I _think_ this can be safely ignored

---

```
H D UC: Method io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils.setNetPeer(Span, String, String, int) seems to be useless  At NetPeerUtils.java:[line 86]
```
This is probably a false positive as it is used in `HttpClientTracer`, do you agree?

### DIFF
--- a/gradle/spotbugs-exclude.xml
+++ b/gradle/spotbugs-exclude.xml
@@ -42,11 +42,4 @@
     </Not>
   </Match>
 
-  <Match>
-    <Class name="~TestServlets.*"/>
-    <Not>
-      <Bug code="UMAC_UNCALLABLE_METHOD_OF_ANONYMOUS_CLASS"/> <!-- can be ignored for servlet doCall() methods in testing -->
-    </Not>
-  </Match>
-
 </FindBugsFilter>
--- a/gradle/spotbugs-exclude.xml
+++ b/gradle/spotbugs-exclude.xml
@@ -1,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FindBugsFilter>
+  <Match>
+    <Class name="io.opentelemetry.instrumentation.auto.jdbc.normalizer.SimpleCharStream"/>
+    <Not>
+      <Bug code="DM_DEFAULT_ENCODING"/> <!-- ignore default encoding for auto-genned class -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="io.opentelemetry.auto.util.gc.GCUtils"/>
+    <Not>
+      <Bug code="DM_GC"/> <!-- forced GC only used in testing -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="~muzzle\.TestClasses.*"/>
+    <Not>
+      <Bug code="MS_SHOULD_BE_FINAL"/> <!-- final field warnings can be ignored for this test class -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="~HttpServletResponseTest\$.*"/>
+    <Not>
+      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="~HttpServletTest\$.*"/>
+    <Not>
+      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="~SpymemcachedTest\$.*"/>
+    <Not>
+      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="~TestServlets.*"/>
+    <Not>
+      <Bug code="UMAC_UNCALLABLE_METHOD_OF_ANONYMOUS_CLASS"/> <!-- can be ignored for servlet doCall() methods in testing -->
+    </Not>
+  </Match>
+
+</FindBugsFilter>
--- a/gradle/spotbugs-exclude.xml
+++ b/gradle/spotbugs-exclude.xml
@@ -2,44 +2,43 @@
 <FindBugsFilter>
   <Match>
     <Class name="io.opentelemetry.instrumentation.auto.jdbc.normalizer.SimpleCharStream"/>
-    <Not>
-      <Bug code="DM_DEFAULT_ENCODING"/> <!-- ignore default encoding for auto-genned class -->
-    </Not>
+    <Bug pattern="DM_DEFAULT_ENCODING"/> <!-- ignore default encoding for auto-genned class -->
   </Match>
 
   <Match>
     <Class name="io.opentelemetry.auto.util.gc.GCUtils"/>
-    <Not>
-      <Bug code="DM_GC"/> <!-- forced GC only used in testing -->
-    </Not>
+    <Bug pattern="DM_GC"/> <!-- forced GC only used in testing -->
   </Match>
 
   <Match>
     <Class name="~muzzle\.TestClasses.*"/>
-    <Not>
-      <Bug code="MS_SHOULD_BE_FINAL"/> <!-- final field warnings can be ignored for this test class -->
-    </Not>
+    <Bug
+      pattern="MS_SHOULD_BE_FINAL"/> <!-- final field warnings can be ignored for this test class -->
   </Match>
 
   <Match>
     <Class name="~HttpServletResponseTest\$.*"/>
-    <Not>
-      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
-    </Not>
+    <Bug
+      pattern="SE_BAD_FIELD_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
   </Match>
 
   <Match>
     <Class name="~HttpServletTest\$.*"/>
-    <Not>
-      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
-    </Not>
+    <Bug
+      pattern="SE_BAD_FIELD_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
   </Match>
 
   <Match>
     <Class name="~SpymemcachedTest\$.*"/>
-    <Not>
-      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
-    </Not>
+    <Bug
+      pattern="SE_BAD_FIELD_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
+  </Match>
+
+  <Match>
+    <Class name="io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils"/>
+    <Method name="setNetPeer"
+      params="io.opentelemetry.trace.Span,java.lang.String,java.lang.String,int" returns="void"/>
+    <Bug pattern="UC_USELESS_VOID_METHOD"/>
   </Match>
 
 </FindBugsFilter>
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/AgentConfigBuilder.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/AgentConfigBuilder.java
@@ -12,12 +12,12 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 
-public final class ConfigBuilder
-    extends io.opentelemetry.sdk.common.export.ConfigBuilder<ConfigBuilder> {
+public final class AgentConfigBuilder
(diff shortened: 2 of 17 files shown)


gradle/spotbugs-exclude.xml
```diff
@@ -0,0 +1,52 @@
...
+
+  <Match>
+    <Class name="~SpymemcachedTest\$.*"/>
+    <Not>
+      <Bug code="SE_INNER_CLASS"/> <!-- inner class serialization warning can be ignored for testing purposes -->
+    </Not>
+  </Match>
+
+  <Match>
+    <Class name="~TestServlets.*"/>
```
iNikem, line 46: We should fix it instead of silencing. The correct way is to pass Callable, as `HttpServerTest#controller` expects. Which now means just lambda.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -46,7 +46,7 @@ abstract class SmokeTest extends Specification {
 
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
 
-  protected static OkHttpClient client = OkHttpUtils.client()
+  protected static final OkHttpClient CLIENT = OkHttpUtils.client()
```
iNikem, line 49: Can we instead ignore this rule? I think @trask will object, but why static final variable MUST be in uppercase?

iNikem:
> There are a couple of spotbugs complaints that I'd like some feedback on
> 
> ```
> H C NP: Null passed for non-null parameter of io.opentelemetry.trace.Span.setAttribute(AttributeKey, Object) in io.opentelemetry.instrumentation.auto.grpc.common.GrpcHelper.prepareSpan(Span, String, InetSocketAddress, boolean)  Method invoked at GrpcHelper.java:[line 62]
> ```

Should be fixed: set attribute only if it is not null.

> 
> ```
> H B Nm: The class name io.opentelemetry.javaagent.tooling.config.ConfigBuilder shadows the simple name of the superclass io.opentelemetry.sdk.common.export.ConfigBuilder  At ConfigBuilder.java:[lines 26-61]
> ```

@mateuszrzeszutek please remind me, why do we need this class which largely duplicates superclass?

> 
> ```
> H D UC: Method io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils.setNetPeer(Span, String, String, int) seems to be useless  At NetPeerUtils.java:[line 86]
> ```

Main branch does not 86 lines in `NetPeerUtils` :) Can you rebase your branch and try again?

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
imavroukakis: spotBugs actually doesn't complain about the uppercase, just the `final` declaration. As for the uppercase, I believe it's  a JLS convention for constants, but given the naming exception done for loggers, I don't see why it cannot be the same here - just possibly, have to look out for other static analysis checkers complaining about it ;)

mateuszrzeszutek:
>>  H B Nm: The class name io.opentelemetry.javaagent.tooling.config.ConfigBuilder shadows the simple name of the superclass io.opentelemetry.sdk.common.export.ConfigBuilder  At ConfigBuilder.java:[lines 26-61]

>@mateuszrzeszutek please remind me, why do we need this class which largely duplicates superclass?

It doesn't really duplicate its superclass, it only extends those methods that it absolutely needs to. Though I agree that its name is somewhere between "uncreative" and "terrible", but I didn't have any better idea when I wrote it 😛

mateuszrzeszutek (approved):

imavroukakis:
> export.ConfigBuilder

how about `ExportConfigBuilder` ?

imavroukakis:
> Main branch does not 86 lines in `NetPeerUtils` :) Can you rebase your branch and try again?
Now it complains about line 75 😆 
```
H D UC: Method io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils.setNetPeer(Span, String, String, int) seems to be useless  At NetPeerUtils.java:[line 75]
```

mateuszrzeszutek:
> how about ExportConfigBuilder ?

I believe that `io.opentelemetry.sdk.common.export.ConfigBuilder` is the full name of the super class in the OTel SDK. Hmm, how about `AgentConfigBuilder`?

imavroukakis:
Odd failures for `:instrumentation:servlet:glassfish-testing:test` and `:instrumentation:jms-1.1:test` I can't replicate them locally.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
iNikem: I don't think JLS has any naming conventions. Has it?

iNikem:
jms is a known flaky test

imavroukakis:
> jms is a known flaky test

ah thanks, @iNikem , is there any way to manually kick the build again?

iNikem:
I will wait, maybe #1327 will fix this

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
imavroukakis: https://docs.oracle.com/javase/specs/jls/se11/html/jls-6.html look for "Constant Names" section.
Emphasis, mine.

>The names of constants in interface types should be, and final variables of class types **may** conventionally be, a sequence of one or more words, acronyms, or abbreviations, all uppercase


The capitalisation could have been a misrepresentation though, and the intent may have been to use them for primitives, but it somehow stuck for everything

anuraaga:
CircleCI is flaky lately (maybe running out of money with all the free alternatives? :P) Sorry if it looks a little weird @imavroukakis but we may close/open PRs to retrigger the build :)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/AgentConfigBuilder.java
```diff
@@ -12,12 +12,12 @@
 import java.util.Map;
 import java.util.Properties;
 
-public final class ConfigBuilder
-    extends io.opentelemetry.sdk.common.export.ConfigBuilder<ConfigBuilder> {
+public final class AgentConfigBuilder
+    extends io.opentelemetry.sdk.common.export.ConfigBuilder<AgentConfigBuilder> {
```
mateuszrzeszutek, line 16: Really minor: you can now import `ConfigBuilder`

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/config/AgentConfigBuilder.java
imavroukakis: good point, let's keep it clean.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
trask: > Can we instead ignore this rule? I think @trask will object, but why static final variable MUST be in uppercase?

I do not object 😂

static final variables should only be uppercase if they are constants, see https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names for "what is a constant"

imavroukakis:
I'm struggling to replicate this issue locally for the JDK8 build but mainly because I can't convince the build to run under JDK8 - any suggestions?

imavroukakis:
> I'm struggling to replicate this issue locally for the JDK8 build but mainly because I can't convince the build to run under JDK8 - any suggestions?

Ugh that was painful, I finally figured the command line I needed to put together
`JAVA_8_HOME=/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home/ gradle --no-daemon :instrumentation:servlet:glassfish-testing:testJava8 -x :smoke-tests:test`

It fails locally too, so I'll be trying to figure out why.

trask:
> Ugh that was painful

We are all looking forward to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/891#issuecomment-696640848

imavroukakis:
> > Ugh that was painful
> 
> We are all looking forward to [#891 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/891#issuecomment-696640848)

That's cool!
Figuring out why the tests fail in JDK8 _only_ is going to be interesting.

trask:
Ah, the reason it's not failing on Java 11, is because it's not really running on Java 11:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f2de47a150a4b0cb28ab7dca2a12230217ae2416/instrumentation/servlet/glassfish-testing/glassfish-testing.gradle#L2

imavroukakis:
Finally, I know _how_ it happens but not _why_

This works

```
new Callable<Object>() {
   public Object call() throws Exception {
      resp.sendRedirect(endpoint.getBody());
         return null;
      }
});
```

This doesn't:

```
() -> {
  resp.sendRedirect(endpoint.getBody());
  return null;
})
```

I don't quite know why the test chokes with the lambda, maybe this has to do something with how it's called from Groovy?

mateuszrzeszutek:
> I don't quite know why the test chokes with the lambda, maybe this has to do something with how it's called from Groovy?

That might be caused by the fact that `Callable`s are instrumented by the `java-concurrent` instrumentation and Groovy `Closure`s are not.

imavroukakis:
> > I don't quite know why the test chokes with the lambda, maybe this has to do something with how it's called from Groovy?
> 
> That might be caused by the fact that `Callable`s are instrumented by the `java-concurrent` instrumentation and Groovy `Closure`s are not.

Hmm, I don't understand your point. The issue here is that, whilst the Groovy `Closure` and Java `Callable` work fine, when the `Callable` is expressed as a lambda, it's as if the Glassfish endpoint is not registered (this is a guess as I'm seeing 404s)

mateuszrzeszutek:
> whilst the Groovy Closure and Java Callable work fine

Wait, Groovy `Closure` works fine? Then please disregard my previous theory 😄 
(`Callable` is instrumented to carry the context/scope with a current span that started it; I thought that groovy `Closure`s were losing this piece of data and that was causing those tests to fail)

iNikem:
Beats me why CircleCI does not kick in...

imavroukakis:
> > whilst the Groovy Closure and Java Callable work fine
> 
> Wait, Groovy `Closure` works fine? Then please disregard my previous theory 😄
> (`Callable` is instrumented to carry the context/scope with a current span that started it; I thought that groovy `Closure`s were losing this piece of data and that was causing those tests to fail)

 😄  it's just the `Callable` when expressed as a lambda that chokes

iNikem:
@imavroukakis I am still looking into why it fails to kick in, but please run `./gradlew codenarcTest` task and fix failures.

imavroukakis:
> ./gradlew codenarcTest

image

Guess we have to go full uppercase for this @iNikem after all.

iNikem:
@imavroukakis I don't know if this is important, but you may want to fix this:

Screenshot 2020-10-06 at 14 38 31

imavroukakis:
> @imavroukakis I don't know if this is important, but you may want to fix this:
> 
> Screenshot 2020-10-06 at 14 38 31

Ah thanks, it seems to be picking up the wrong key in GH.

imavroukakis:
circleCI exploded
```
* Exception is:
java.lang.OutOfMemoryError: Java heap space
	at org.gradle.api.internal.file.archive.impl.FileZipInput.<init>(FileZipInput.java:71)
	at org.gradle.api.internal.file.archive.impl.FileZipInput.create(FileZipInput.java:43)
	at org.gradle.api.internal.changedetection.state.Zip
```

imavroukakis:
And now it's run out of space.apparently.
```
> Task :smoke-tests:test
java.io.IOException: No space left on device
com.esotericsoftware.kryo.KryoException: java.io.IOException: No space left on device
	at com.esotericsoftware.kryo.io.Output.flush(Output.java:165)
	at com.esotericsoftware.kryo.io.Output.require(Output.java:142)
```

iNikem:
Yea, I am trying to understand that disk space issue

trask (approved):
Thanks @imavroukakis 👍

I'll submit a separate PR to update the codenarc configuration to avoid having to use uppercase for (non-constant) static final fields (in groovy)

PR #1327, opened 2020-10-05
Ported JMS 1.x tests to testcontainers

### LINKED ISSUE #1206: Flaky test - :instrumentation:jms-1.1:test
I've seen this test fail a lot lately - it does get retried but fails every one of them from what I can tell.

https://gradle.com/s/geylzhj6543j2

iNikem:
Fixes #1206

### DIFF
--- a/gradle/instrumentation.gradle
+++ b/gradle/instrumentation.gradle
@@ -61,5 +61,7 @@ afterEvaluate {
     testAnnotationProcessor deps.autoservice
     testImplementation deps.autoservice
     testImplementation project(':utils:test-utils')
+
+    testImplementation deps.testcontainers
   }
 }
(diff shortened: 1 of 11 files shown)


iNikem:
I have a suspicion that build fails due to `AdditionalLibraryIgnoresMatcher` and static configs. And I still don't understand why do we need `AdditionalLibraryIgnoresMatcher` :(

iNikem:
Yes, CiMate shows nicely that failure is due to `AdditionalLibraryIgnoresMatcher`. @trask can you remind me, why do we need this matcher and this strange check in our tests?

tylerbenson:
The `AdditionalLibraryIgnoresMatcher` is intended as an optimization for startup to try to bypass a significant amount of classes we know we won't instrument.  It's handled a bit differently from other ignores in the tests because if we do happen to instrument something that matches it in a test, then something changed and we need to update the matcher.

trask (approved):
Nice, let the testcontainers migration start 🎉

trask:
Interesting, the muzzle matrix was picked up in this PR even though we didn't merge that change in from master, is that normal?

anuraaga (approved):

Issue #1313, opened 2020-10-02
How to trace gRPC calls and log them?


asarkar:
I'm looking for a way to trace gRPC calls in a Kotlin + Spring Boot application, and for time being, only interested in finding the logs by a trace id. I started with Spring Cloud Sleuth, but immediately hit the bug https://github.com/spring-cloud/spring-cloud-sleuth/issues/1749.

> I've this application with a gRPC server (A) that calls a gRPC client (B) that calls an external gRPC service. It all starts with a ServerInterceptor injecting a user id in the Context that is required by the service A; since A doesn't have access to the headers, context is the only way to propagate metadata from the client.

gRPC has OOTB support for OpenCensus, so, for a new project, I'm inclined to use OpenTelemetry if possible. However, my Google searches have turned up absolutely nothing related to my simple question: "How to trace gRPC calls and have them show up in the logs?" At this time, I'm not interested in exporting them to another system like Jaeger.

I'd like to avoid the magic of a javaagent, and instead have the instrumentation supported in the code first hand, like @mabdinur did in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/661 for HTTP. This is a preference, and not a must.

Part of the problem might be that I've user-level knowledge of distributed tracing, but by no means I'm an expert. @trask and @anuraaga, from scouring gRPC related issues in this repo, you two seem to be the experts in this domain, so, I'll really appreciate if you can throw me a bone or two to get started. If something reusable comes up in my quest, like the PR 661, I'll be happy to contribute it back to this project.

Using [grpc-kotlin](https://github.com/grpc/grpc-kotlin) v0.1.5, Spring Boot 2.3.4.RELEASE, and Logback.

anuraaga:
Hi @asarkar - unfortunately our current instrumentation for gRPC relies on using the javaagent. We are in the process of splitting out library instrumentation, which can be used without the agent, and gRPC is next on the list. We would essentially provide you a library with a `ServerInterceptor` and `ClientInterceptor` implementation you register on a `ServerBuilder` or `ManagedChannelBuilder`. 

In the meantime, if avoiding the agent isn't a must, maybe you can try playing with it. If you set `OTEL_EXPORTER` environment variable to logging, it will just log spans.

https://github.com/open-telemetry/opentelemetry-java-instrumentation#logging-exporter

How does that sound? We'll also get to extracting gRPC instrumentation soon, thanks for pushing for it.

[...]

PR #1329, opened 2020-10-06
Split out gRPC library instrumentation.

anuraaga:
For #1313


instrumentation/grpc-1.5/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_5/AbstractGrpcTest.groovy
```diff
@@ -77,8 +85,6 @@ class GrpcTest extends AgentTestRunner {
             "${SemanticAttributes.RPC_SYSTEM.key()}" "grpc"
             "${SemanticAttributes.RPC_SERVICE.key()}" "example.Greeter"
             "${SemanticAttributes.RPC_METHOD.key()}" "SayHello"
-            "${SemanticAttributes.NET_PEER_NAME.key()}" "localhost"
```
anuraaga, line 80: This is a regression, but I want to fix forward. Previously auto instrumentation got the address from the channel builder, but it seems weird to force that into library instrumentation too. gRPC does have a remote transport address attribute with some behavior difference between versions. Want to dig into that more as a followup instead of having to add address as a parameter to the interceptor's factories. How does that sound?

asarkar:
@anuraaga The brave-grpc instrumentation crapped out for my simple use case. You might want to check if your impl works when the `Context` is modified using a `ServerInterceptor`.
https://github.com/spring-cloud/spring-cloud-sleuth/issues/1749

anuraaga:
@asarkar Thanks for the note, we'll definitely make sure that use case works well. I'm currently refactoring OTel's context propagation from the ground up and gRPC context interop is a main goal of the changes so hopefully it helps you too.

instrumentation/grpc-1.5/testing/grpc-1.5-testing.gradle
```diff
@@ -0,0 +1,39 @@
...
+  api project(':testing-common')
+
+  api group: 'io.grpc', name: 'grpc-core', version: grpcVersion
+  api group: 'io.grpc', name: 'grpc-protobuf', version: grpcVersion
+  api group: 'io.grpc', name: 'grpc-stub', version: grpcVersion
+
+  implementation group: 'javax.annotation', name: 'javax.annotation-api', version: '1.3.2'
+
+  implementation deps.guava
+
```
trask, line 35: are these dependencies needed?
```suggestion
```

instrumentation/grpc-1.5/testing/grpc-1.5-testing.gradle
```diff
@@ -0,0 +1,39 @@
...
+  plugins {
+    grpc { artifact = "io.grpc:protoc-gen-grpc-java:${grpcVersion}" }
+  }
+  generateProtoTasks {
+    all()*.plugins { grpc {} }
+  }
+}
+
+dependencies {
+  compileOnly(project(':instrumentation:grpc-1.5:library'))
```
trask, line 24: it's amazing to me that this does not create a cyclic dependency

the test that uses `GrpcHelper` does look a bit suspicious (using the same method to verify the status as was used to map the status), maybe better to add an extra spock data column to those test for otelStatus?

instrumentation/grpc-1.5/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_5/AbstractGrpcTest.groovy
trask: No objection from me, let's create a tracking issue for this if @iNikem does not object.

trask (approved):

instrumentation/grpc-1.5/testing/grpc-1.5-testing.gradle
anuraaga: Previously I was avoiding having to duplicate the statuscode mapping in the logic, its unit test, and here too. Now it's always one status though so no reason to use the helper :)

PR #1335, opened 2020-10-07
Update codenarc config

trask:
Since we are following Google Java Style Guide, `static final` fields should only be UPPER_CASE if they are constants (see https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names for "what is a constant?")

Note (because I forgot) codenarc only applies to groovy files.


gradle/enforcement/codenarc.groovy
```diff
@@ -308,7 +308,8 @@ ruleset {
   FieldName {
     regex = '^_?[a-z][a-zA-Z0-9]*$'
     finalRegex = '^_?[a-z][a-zA-Z0-9]*$'
-    staticFinalRegex = '^logger$|^[A-Z][A-Z_0-9]*$|^serialVersionUID$'
+    // can be either constant (ABC_XYZ) or non-constant (abcXyz)
+    staticFinalRegex = '^[A-Z][A-Z_0-9]*$|^_?[a-z][a-zA-Z0-9]*$'
```
iNikem, line 312: Isn't it equivalent to just `^_?[a-z][a-zA-Z0-9]*$`?

iNikem (approved):

gradle/enforcement/codenarc.groovy
mateuszrzeszutek: Hmm, this one won't fail something like `_ABC` - I don't know if that's an issue though

mateuszrzeszutek (approved):

gradle/enforcement/codenarc.groovy
trask: ya, i was being lazy and just copy-pasted and or-ed them and didn't apply any thinking 😄, i don't think we use leading underscores anywhere

Issue #1249, opened 2020-09-24
Document traceId/spanId injection into logs


iNikem:
Document functionality added by #1200 and other relevant PRs

PR #1339, opened 2020-10-07
Document traceId/spanId injection into logs

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1249

### DIFF
--- a/docs/logger-mdc-instrumentation.md
+++ b/docs/logger-mdc-instrumentation.md
@@ -1,0 +1,37 @@
+# Logger MDC auto-instrumentation
+
+The Mapped Diagnostic Context (MDC) is
+
+> an instrument for distinguishing interleaved log output from different sources.
+> &mdash; <cite> [log4j MDC documentation](http://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/MDC.html) </cite>
+
+It contains thread-local contextual information which is later copied to each logging event captured
+by a logging library.
+
+The OTel Java agent injects several pieces of information about the current span into each logging
+event's MDC copy:
+
+- `traceId` - the current trace id
+  (same as `TracingContextUtils.getCurrentSpan().getContext().getTraceIdAsHexString()`);
+- `spanId` - the current span id
+  (same as `TracingContextUtils.getCurrentSpan().getContext().getSpanIdAsHexString()`);
+- `sampled` - a boolean flag marking whether the current span is sampled or not
+  (same as `TracingContextUtils.getCurrentSpan().getContext().isSampled()`).
+  Only added when `sampled` is `true`.
+
+Those three pieces of information can be included in log statements produced by the logging library
+by specifying them in the pattern/format. Example for Spring Boot configuration (which uses logback):
+
+```properties
+logging.pattern.console = %d{yyyy-MM-dd HH:mm:ss} - %logger{36} - %msg t:%X{traceId} s:%X{spanId} %n
+```
+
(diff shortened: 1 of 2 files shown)


docs/logger-mdc-instrumentation.md
```diff
@@ -0,0 +1,37 @@
...
+The OTel Java agent injects several pieces of information about the current span into each logging
+event's MDC copy:
+
+- `traceId` - the current trace id
+  (same as `TracingContextUtils.getCurrentSpan().getContext().getTraceIdAsHexString()`);
+- `spanId` - the current span id
+  (same as `TracingContextUtils.getCurrentSpan().getContext().getSpanIdAsHexString()`);
+- `sampled` - a boolean flag marking whether the current span is sampled or not
+  (same as `TracingContextUtils.getCurrentSpan().getContext().isSampled()`).
+  Only added when `sampled` is `true`.
```
iNikem, line 20: If it is always true, why include it?

mateuszrzeszutek: I've added an explanation to the doc. Basically if it's there the span is sampled, if not it is not - I guess the reasoning behind it is that most of spans are not sampled and keeping sampled=false in every log line would be just a waste of space.

iNikem (approved):

docs/logger-mdc-instrumentation.md
iNikem: Aren't most span sampled? With `AlwaysOn` sampler, which is default

mateuszrzeszutek: 🤦  That's true...

I somehow thought that it worked completely the other way 😅 

Now I'm starting to wonder why it works that way... @anuraaga I think you've added it first in the logback instrumentation -- why do we only log `true` value?

trask (approved):
Nice doc 👍

Good question about whether we want to change `sampled` field behavior, it seems like it would be easier for log parsers if the field is not conditional(?)

anuraaga:
Yeah it's just a premature optimization, I don't think I taught too deeply when making that change. Always adding sampled seems nice to me too.

anuraaga (approved):

iNikem:
> Yeah it's just a premature optimization, I don't think I taught too deeply when making that change. Always adding sampled seems nice to me too.

Does this mean we have to make some code change in our MDC handling?

anuraaga:
@iNikem Yeah - so I'm good with merging this first and updating code / README in a followup.

mateuszrzeszutek:
Hmm, looks like random errors:

> Could not GET 'https://jcenter.bintray.com/org/springframework/spring-beans/5.0.9.RELEASE/spring-beans-5.0.9.RELEASE.pom'. Received status code 502 from server: Bad Gateway
> Could not GET 'https://jcenter.bintray.com/io/projectreactor/ipc/reactor-netty/0.7.8.RELEASE/reactor-netty-0.7.8.RELEASE.pom'. Received status code 502 from server: Bad Gateway
> Could not GET 'https://jcenter.bintray.com/org/springframework/spring-beans/5.2.1.RELEASE/spring-beans-5.2.1.RELEASE.pom'. Received status code 502 from server: Bad Gateway

mateuszrzeszutek:
Rebased to force CircleCI run

iNikem:
@mateuszrzeszutek will you do a follow-up fix right away or will you create an issue for that?

mateuszrzeszutek:
I'll do it right away.

PR #1343, opened 2020-10-08
Don't mount context in gRPC instrumentation since gRPC automatically …

anuraaga:
…does it, better.

The client seems to mount the context automatically with no work, and on server we can use `ServerCalls.intercept`. This makes sure we don't clobber context modifications that happen downstream from our interceptors.

/cc @asarkar Thanks for bring this to our attention!


anuraaga:
There's some workarounds for the bootstrap classpath, but it will become simpler after grpc context is gone.

anuraaga:
Strange can't repro the test failures locally which is surprising.

trask (approved):
Nice 👍

trask:
Just a note, since I didn't understand this before chatting with @anuraaga, the goal is that piggybacking on the "real" gRPC Context should continue to work even after we migrate to OTel Context because @anuraaga is planning to add interop between OTel Context / gRPC Context interop into the agent at that time.

PR #1346, opened 2020-10-08
spotbugs upgrade and fixes for JMS1 tests

imavroukakis:
My previous PR (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1325) missed a _spotBugs_ issue in `KotlinCoroutineTests`. This turned out into a minor change, in order to avoid adding an exclusion for `KotlinCoroutineTests` in _spotBugs_

* _jms-1.1:_ tests where tweaked to pass _spotBugs_ checks
* the _spotBugs_ gradle plugin and tool upgrade was applied.


imavroukakis:
I just realised I named the branch "spotless" instead of "spotBugs" 🤦 apologies.

gradle/spotbugs.gradle
```diff
@@ -7,6 +7,26 @@ allprojects {
     omitVisitors = ["FindDeadLocalStores"]
     effort = "max"
     excludeFilter = file("$rootDir/gradle/spotbugs-exclude.xml")
+    toolVersion = '4.1.3'
```
iNikem, line 10: Why older version here?

gradle/spotbugs.gradle
```diff
@@ -7,6 +7,26 @@ allprojects {
...
+        enabled = true
+        destination = file("$buildDir/reports/spotbugs/main/spotbugs.html")
+        stylesheet = 'fancy-hist.xsl'
+      }
+    }
+  }
+  spotbugsMain {
+    reports {
+      html {
+        enabled = false
```
iNikem, line 25: First, why html report for tests is enabled and for other code disabled? Second, if this report is disabled, why do you configure its destination?

instrumentation/jms-1.1/src/test/groovy/JMS1Test.groovy
```diff
@@ -33,7 +33,7 @@ class JMS1Test extends AgentTestRunner {
 
   private static final Logger logger = LoggerFactory.getLogger(JMS1Test)
 
-  public static GenericContainer activemq = new GenericContainer("rmohr/activemq")
+  private static final GenericContainer BROKER = new GenericContainer("rmohr/activemq")
```
iNikem, line 36: what check requires uppercase names?

instrumentation/kotlin-coroutines/src/test/kotlin/KotlinCoroutineTests.kt
```diff
@@ -105,27 +104,17 @@ class KotlinCoroutineTests(private val dispatcher: CoroutineDispatcher) {
...
-      }
-    )
-
     withTimeout(TimeUnit.SECONDS.toMillis(30)) {
-      select<Boolean> {
-        children.forEach { child ->
-          child.onAwait { it }
-        }
-      }
+      tracedChild("timeout1")
```
iNikem, line 108: Why this change was required?

gradle/spotbugs.gradle
imavroukakis: This is the latest spotbugs _tool_ version, the gradle plugin has a different version cadence.

instrumentation/jms-1.1/src/test/groovy/JMS1Test.groovy
imavroukakis: image
spotBugs recommends this to be private (although maybe sharing it between the tests might also make sense) and once you turn it `private final` _codeNarc_ complains about the naming - I believe @trask was going to look into this.

iNikem: I believe he already did: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1335. Is your baseline fresh enough?

gradle/spotbugs.gradle
iNikem: And do you _have_ to specify it here? If you don't will anything break?

instrumentation/kotlin-coroutines/src/test/kotlin/KotlinCoroutineTests.kt
imavroukakis: The way I understand what this test does is, that it launches these coroutines and then `onAwait`s for the results from each with a timeout. `spotBugs` complains about passing null into https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1346/files/a6525b40e700bf52a8dcf707cdad68bc0fa9bb91#diff-ca9202f7cbd3f2a56910810901063998L126 - it's obviously confused! Instead of adding the class to the exclusion, rewriting it this way keeps the original intent, makes the code somewhat simpler and satisfies spotBugs.

gradle/spotbugs.gradle
imavroukakis: Good catch, it shouldn't have been. Ideally, what I would like to do is, conditionally run HTML reports if not under a CI environment and XML otherwise for the CI server to process -  but I'm not sure if I can do this within the context of the script. I did notice some tests that picked up a system environment variable, maybe the same would work here?

gradle/spotbugs.gradle
imavroukakis: no, nothing breaks; if left unspecified it will revert to whatever version the plugin defines. IMO it is probably better to peg the tool version instead of leaving it undefined.

instrumentation/jms-1.1/src/test/groovy/JMS1Test.groovy
imavroukakis: probably not. I will rebase and run the tasks again, thanks

gradle/spotbugs.gradle
imavroukakis: Ah I found it in the root gradle file! `def isCI = System.getenv("CI") != null`

gradle/spotbugs.gradle
iNikem: Extra unnecessary configuration is extra code to maintain. If we already peg the version of the plugin, it should enough, no?

gradle/spotbugs.gradle
```diff
@@ -2,11 +2,35 @@ allprojects {
...
+    ignoreFailures = false
     reportLevel = "high"
     omitVisitors = ["FindDeadLocalStores"]
     effort = "max"
     excludeFilter = file("$rootDir/gradle/spotbugs-exclude.xml")
+    toolVersion = '4.1.3'
+  }
+
+  if (!isCI) {
+    spotbugsTest {
```
iNikem, line 14: What of these are default configuration anyway and what specifically you want to override?

gradle/spotbugs.gradle
imavroukakis: If you upgrade the gradle plugin, you get additional plugin fixes and/or configuration but you _might_ also get a new version of the _spotBugs_ tool. If you're explicit about the tool version, you can be reasonably confident of it not changing underneath your feet - having said that I don't feel strongly about it, so I'm happy to go with whatever direction you decide.

gradle/spotbugs.gradle
imavroukakis: Default config is to generate an XML file for machine consumption, makes sense on a CI environment, not as much sense on a local machine.

instrumentation/kotlin-coroutines/src/test/kotlin/KotlinCoroutineTests.kt
iNikem: I am not sure these are equivalent. First, old test waited 30s max in total, new test waits 30 seconds on each child.

And reading https://github.com/spotbugs/spotbugs/issues/573 I wonder if we should run spotbugs on Kotlin code at all :)

gradle/spotbugs.gradle
iNikem: Then I believe this whole `if` block can be replaced with 
```
  spotbugsTest {
    reports {
      html.enabled = !isCI
    }
  }
```

gradle/spotbugs.gradle
iNikem: I prefer not to fix tool versions unless there is an actual reason for that :) Upgrade all the things! :)

gradle/spotbugs.gradle
```diff
@@ -10,27 +10,24 @@ allprojects {
...
-        html {
-          destination = file("$buildDir/reports/spotbugs/main.html")
-          stylesheet = 'fancy-hist.xsl'
-        }
+  }
+  spotbugsMain {
+    reports {
+      html.enabled = !isCI
+      xml.enabled = isCI
+      html {
```
iNikem, line 27: If these are default values anyway, then remove them

imavroukakis: Default is `true` for `xml.enabled`. However, to enable HTML reports, you have to explicitly disable the XML ones so both need to be present and set. Does this make sense or have I missed your point entirely?

instrumentation/kotlin-coroutines/src/test/kotlin/KotlinCoroutineTests.kt
imavroukakis: `detekt` is definitely more appropriate.

gradle/spotbugs.gradle
iNikem: I was talking about `html` configuration for stylesheet and report location. Why default values for them are not enough?

imavroukakis: Oh I follow. Yeah we can remove the report location, but maybe keep the stylesheet, as it's not the default?

iNikem (approved):
Thank you for bearing with me! :) Good work. Seems it is so good that it breaks the build https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-instrumentation/3689/workflows/0f9faf9b-26b9-4ad5-98ca-3c9c402fb8ad/jobs/18734

imavroukakis:
> Thank you for bearing with me! :) Good work. Seems it is so good that it breaks the build [app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-instrumentation/3689/workflows/0f9faf9b-26b9-4ad5-98ca-3c9c402fb8ad/jobs/18734](https://app.circleci.com/pipelines/github/open-telemetry/opentelemetry-java-instrumentation/3689/workflows/0f9faf9b-26b9-4ad5-98ca-3c9c402fb8ad/jobs/18734)

Spotted it 😆 I'll get that fixed up

PR #1348, opened 2020-10-08
Instrument Netty addTask to ensure complete coverage of async Runnables 

johnbley:
We found that certain code paths in RxNetty through vanilla Netty exposed an issue where Netty WriteTasks weren't capturing the correct/any context during submission.  Turns out that newer 4.1 variants of Netty have a backchannel "lazy" execution that uses `execute(Runnable, boolean)`, bypassing the standard `execute(Runnable)`.  Fortunately, all the variants flow through to `addTask(Runnable)`.  

Without the addTask instrumentation addition, the added unit test fails with 3 traces (the second client request shows up under a brand new 3rd trace without any inherited context).

### DIFF
--- a/instrumentation/netty/netty-4.1/netty-4.1.gradle
+++ b/instrumentation/netty/netty-4.1/netty-4.1.gradle
@@ -27,7 +27,7 @@ muzzle {
 }
 
 dependencies {
-  library group: 'io.netty', name: 'netty-codec-http', version: '4.1.52.Final'
+  library group: 'io.netty', name: 'netty-codec-http', version: '4.1.0.Final'
 
   testImplementation project(':instrumentation:reactor-3.1')
 
@@ -45,7 +45,7 @@ configurations.testImplementation {
     eachDependency { DependencyResolveDetails details ->
       //specifying a fixed version for all libraries with io.netty' group
       if (details.requested.group == 'io.netty') {
-        details.useVersion "4.1.52.Final"
+        details.useVersion "4.1.0.Final"
       }
     }
   }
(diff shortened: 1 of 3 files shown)


instrumentation/netty/netty-4.1/netty-4.1.gradle
```diff
@@ -27,7 +27,7 @@ muzzle {
 }
 
 dependencies {
-  library group: 'io.netty', name: 'netty-codec-http', version: '4.1.0.Final'
+  library group: 'io.netty', name: 'netty-codec-http', version: '4.1.52.Final'
```
iNikem, line 30: Wouldn't it be enough to have later versions in `latestDepTest`? I think we still want to test with the oldest supported version.

trask (commented):
looks good, thanks 👍

agree with @iNikem's comment about test version

anuraaga:
For reference, we have this pattern if you need to be able to branch in the test based on whether it's normal or latest dep

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/e74ffe2e52cb30f2ca350da2e154a082bdd0dc08/instrumentation/redisson-3.0/redisson-3.0.gradle#L17

instrumentation/netty/netty-4.1/netty-4.1.gradle
johnbley: Yep, thanks for that.  Confirmed that with `-PtestLatestDeps=true` and the `addTask` instrumentation commented out the test fails, so reverting the gradle file.

trask (approved):
Thanks, nice find!

trask (approved):

PR #1351, opened 2020-10-09
Remove unnecessary scopes

### LINKED ISSUE #898: Remove scopes that only exist for internal agent log correlation
There are many places that create scopes who's purpose is unclear, e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/227bde87ce7aa8f9ae725f2daaf7b0c312117b7e/instrumentation/rabbitmq-2.7/src/main/java/io/opentelemetry/auto/instrumentation/rabbitmq/amqp/RabbitChannelInstrumentation.java#L270-L278

The original purpose of this was so that internal logging within the agent that occurred within this block of code could include the active span info, but that purpose is very hidden and not clear to anyone reading the code.

I suggest we remove them.

An alternative, if we want to keep them, would be to apply this pattern everywhere, and use a wrapper named something like `startLoggingScope` to make it's purpose clear.

trask:
Closes #898


anuraaga (approved):

instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/instrumentation/auto/servlet/dispatcher/RequestDispatcherInstrumentation.java
```diff
@@ -115,18 +114,16 @@ public static void start(
         parent = servletContext;
       }
 
-      try (Scope ignored = ContextUtils.withScopedContext(parent)) {
-        span = TRACER.startSpan(method);
+      span = TRACER.startSpan(method);
```
iNikem, line 117: Here the correct parent span is important. I think this change breaks it.

instrumentation/spymemcached-2.12/src/main/java/io/opentelemetry/instrumentation/auto/spymemcached/CompletionListener.java
```diff
@@ -28,29 +26,27 @@ public CompletionListener(MemcachedConnection connection, String methodName) {
   }
 
   protected void closeAsyncSpan(T future) {
-    try (Scope ignored = currentContextWith(span)) {
-      try {
-        processResult(span, future);
-      } catch (CancellationException e) {
+    try {
+      processResult(span, future);
```
iNikem, line 30: What if `processResult` creates new spans? Will their parent change now?

iNikem (changes requested):

instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/instrumentation/auto/servlet/dispatcher/RequestDispatcherInstrumentation.java
trask: Eeek, thanks for catching this!

instrumentation/spymemcached-2.12/src/main/java/io/opentelemetry/instrumentation/auto/spymemcached/CompletionListener.java
trask: `processResult` is only used to populate span attributes, it doesn't make callbacks into the user code, so I think this is ok

iNikem (approved):

instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/instrumentation/auto/servlet/dispatcher/RequestDispatcherInstrumentation.java
anuraaga: Can we add a test for it?

iNikem:
@trask it seems you have to run spotless on this.

instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/instrumentation/auto/servlet/dispatcher/RequestDispatcherInstrumentation.java
trask: done

PR #1366, opened 2020-10-09
Add org.gradle.workers.max to gradle.properties

### LINKED ISSUE #740: Remove org.gradle.parallel=true from gradle.properties
Currently we have `org.gradle.parallel=true` specified in our `gradle.properties`. This makes my machine grind to a halt during the builds. I can fix this by adding `org.gradle.workers.max` to my `~/.gradle/gradle.properties`, but I'm not sure that `org.gradle.parallel=true` provides the best new contributor experience.

trask:
Closes #740

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,9 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 
+# limiting the number of parallel workers helps to keep your machine responsive during builds
+# see https://github.com/gradle/gradle/issues/14224
+org.gradle.workers.max=2
+
 # Gradle default is 256m which causes issues with our build - https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -1,5 +1,9 @@
 org.gradle.parallel=true
 org.gradle.caching=true
 
+# limiting the number of parallel workers helps to keep your machine responsive during builds
+# see https://github.com/gradle/gradle/issues/14224
+org.gradle.workers.max=2
```
anuraaga, line 6: Just for reference, I have 4 cores with hyperthreading, and max-workers of 4 works well since it ends up with 8. This will end up with 6 workers if I'm not mistaken (build always adds 4 extra workers) - wonder if it works on a 4-core without hyperthreading machine.

iNikem (approved):

trask:
Merging, but very open to revisit (further restrict, e.g. remove parallel completely) if we find contributors continue to run into this issue.

Issue #1186, opened 2020-09-10
Maven groupIds and artifactIds revisited


trask:
Based on discussion from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1173#issuecomment-687203429 😸:

#### io.opentelemetry.instrumentation

* Library (manual) instrumentation that can be used by a user without the Java agent to instrument `cats`
  * ~`opentelemetry-cats`~ 🡆 `opentelemetry-instrumentation-cats` (for symmetry, though will create some long names)
* Library (manual) instrumentation API
  * `opentelemetry-instrumentation-api`

#### ~io.opentelemetry.instrumentation.javaagent~ 🡆 io.opentelemetry.javaagent

* Java agent (auto) instrumentation modules, loaded by the Java agent to interact with `cats`
  * `opentelemetry-javaagent-cats`
* Java agent (auto) instrumentation API (TBD #1129)
  * `opentelemetry-javaagent-api`
* The Java agent itself
  * `opentelemetry-javaagent`
* The Java agent SPI (TBD #1129)
  * `opentelemetry-javaagent-spi`
* Embedded exporters for the javaagent
  * `opentelemetry-javaagent-exporters-otlp`
  * `opentelemetry-javaagent-exporters-jaeger`
  * `opentelemetry-javaagent-exporters-zipkin`
  * `opentelemetry-javaagent-exporters-logging`
* Internal components of the javaagent
  * `opentelemetry-javaagent-bootstrap`
  * `opentelemetry-javaagent-tooling`

The above naming also explores how things would look if we changed terminology from `auto` to `javaagent`.

trask:
I'm a little torn between `auto` and `javaagent`.

I agree `auto` is too generic.

On the other hand, `javaagent` is maybe too restrictive. We have already had discussions about re-using the bytecode instrumentation outside of a javaagent (e.g. bootstrapping during `main()` for aws lambda, and compile-time instrumentation for android).

OTOH, it makes sense to name things for their primary use cases (in our case, using `javaagent` instead of `auto`). I think it would be ok if we reuse those artifacts for other use cases (e.g. using `javaagent` artifacts for compile time android instrumentation).

iNikem:
[...]

PR #1370, opened 2020-10-12
Rename "auto" to "javaagent" in modules

trask:
Another part of #1186

This includes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1186#issuecomment-706770335:

> Trying out shortening the name `opentelemetry-javaagent-instrumentation-api` --> `opentelemetry-javaagent-api`

But not the shortened package rename yet, because still thinking about https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1186#issuecomment-706824508.


anuraaga (approved):
SGTM

iNikem (approved):
LGTM but one question: why we have to repeat `JAVAagent`? Maybe just `agent`?

trask:
> LGTM but one question: why we have to repeat `JAVAagent`? Maybe just `agent`?

I like changing it here:

* armeria-1.0
  * javaagent -> agent
  * library
  * testing

since `library` in this context doesn't really correspond to package name / artifact name either.

For other places, I'm torn:

#### io.opentelemetry.agent

  * `opentelemetry-agent-cats`
  * `opentelemetry-agent-api`
  * `opentelemetry-agent`
  * `opentelemetry-agent-spi`
  * `opentelemetry-agent-exporters-otlp`
  * `opentelemetry-agent-exporters-jaeger`
  * `opentelemetry-agent-exporters-zipkin`
  * `opentelemetry-agent-exporters-logging`
* Internal components of the agent (maybe we don't need to publish?)
  * `opentelemetry-agent-internal-bootstrap`
  * `opentelemetry-agent-internal-tooling`

I like the shorter `agent`, but also we were trying to avoid confusion with external agents, e.g. OpenTelemetry collector agent.

iNikem:
Does anybody call Otel Collector an "agent"?

trask:
> Does anybody call Otel Collector an "agent"?

Yeah, https://opentelemetry.io/docs/collector/about/

Also, when most ops people hear "agent", they think of out-of-process agent (like the collector agent), as opposed to in-process agent (like a javaagent). Some discussion about this occurred here: https://github.com/open-telemetry/oteps/blob/master/text/0001-telemetry-without-manual-instrumentation.md

iNikem:
Let it be "javaagent" then

docs/contributing/javaagent-jar-components.md
```diff
@@ -82,7 +82,7 @@ module
 Available in the bootstrap class loader:
 
 - `io/opentelemetry/javaagent/bootstrap/` - contains the `javaagent-bootstrap` module
-- `io/opentelemetry/instrumentation/auto/api/` - contains the `auto-api` module
+- `io/opentelemetry/javaagent/api/` - contains the `javaagent-api` module
```
mateuszrzeszutek, line 85: Currently it's `io/opentelemetry/javaagent/instrumentation/api/`

trask: thanks! fixed

mateuszrzeszutek (approved):

PR #1380, opened 2020-10-13
Add SPI to configure additional bootstrap package prefixes

pavolloffay:
This feature is useful when a large set of custom instrumentations is
using common classes from a custom package.

cc) @trask 

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/BootstrapPackagePrefixesHolder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/BootstrapPackagePrefixesHolder.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.config;
+
+/**
+ * {@link BootstrapPackagePrefixesHolder} in an utility class that holds package prefixes. The
+ * classes from these packages are pushed to the bootstrap classloader.
+ *
+ * <p>The prefixes are loaded by {@code AgentInstaller} and consumed by classloader instrumentation.
+ * The instrumentation does not have access to the installer, therefore this utility class is used
+ * to share package prefixes.
+ */
+public class BootstrapPackagePrefixesHolder {
+
+  private static String[] bootstrapPrefixes;
+
+  public static String[] getBootstrapPrefixes() {
+    return bootstrapPrefixes;
+  }
+
+  public static void setBootstrapPrefixes(String[] prefixes) {
+    bootstrapPrefixes = prefixes;
+  }
+}
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
@@ -5,6 +5,9 @@
 
 package io.opentelemetry.instrumentation.api.internal;
 
+import java.util.Collections;
+import java.util.List;
+
 /**
  * {@link BootstrapPackagePrefixesHolder} is an utility class that holds package prefixes. The
  * classes from these packages are pushed to the bootstrap classloader.
@@ -15,13 +18,17 @@ package io.opentelemetry.instrumentation.api.internal;
  */
 public class BootstrapPackagePrefixesHolder {
 
-  private static String[] bootstrapPrefixes;
+  private static volatile List<String> BOOSTRAP_PACKAGE_PREFIXES;
 
-  public static String[] getBootstrapPrefixes() {
-    return bootstrapPrefixes;
+  public static List<String> getBoostrapPackagePrefixes() {
+    return BOOSTRAP_PACKAGE_PREFIXES;
   }
 
-  public static void setBootstrapPrefixes(String[] prefixes) {
-    bootstrapPrefixes = prefixes;
+  public static void setBoostrapPackagePrefixes(List<String> prefixes) {
+    if (BOOSTRAP_PACKAGE_PREFIXES != null) {
+      // Only possible by misuse of this API, just ignore.
+      return;
+    }
+    BOOSTRAP_PACKAGE_PREFIXES = Collections.unmodifiableList(prefixes);
   }
 }
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
@@ -3,10 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.instrumentation.api.config;
+package io.opentelemetry.instrumentation.api.internal;
 
 /**
- * {@link BootstrapPackagePrefixesHolder} in an utility class that holds package prefixes. The
+ * {@link BootstrapPackagePrefixesHolder} is an utility class that holds package prefixes. The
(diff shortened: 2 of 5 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/BootstrapPackagePrefixesHolder.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * {@link BootstrapPackagePrefixesHolder} in an utility class that holds package prefixes. The
+ * classes from these packages are pushed to the bootstrap classloader.
+ *
+ * <p>The prefixes are loaded by {@code AgentInstaller} and consumed by classloader instrumentation.
+ * The instrumentation does not have access to the installer, therefore this utility class is used
+ * to share package prefixes.
+ */
+public class BootstrapPackagePrefixesHolder {
```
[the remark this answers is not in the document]
trask: can this go in `javaagent-bootstrap` near `HelperResources` which serves somewhat similar purpose?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/BootstrapPackagePrefixesHolder.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.config;
+
+/**
+ * {@link BootstrapPackagePrefixesHolder} in an utility class that holds package prefixes. The
```
trask, line 9:
```suggestion
 * {@link BootstrapPackagePrefixesHolder} is an utility class that holds package prefixes. The
```

instrumentation/java-classloader/src/main/java/io/opentelemetry/javaagent/instrumentation/javaclassloader/ClassLoaderInstrumentation.java
```diff
@@ -88,7 +89,7 @@ public ClassLoaderInstrumentation() {
         return null;
       }
       try {
-        for (String prefix : Constants.BOOTSTRAP_PACKAGE_PREFIXES) {
+        for (String prefix : BootstrapPackagePrefixesHolder.getBootstrapPrefixes()) {
```
anuraaga, line 92: Is Config not available here? If it is, we can just use a config property instead of a separate SPI?

trask: See @pavolloffay's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1380#discussion_r503974593

> We could also remove the SPI so vendors would set the packages in the config - I don't like this since it easily opens the config to the consumers.

I think I agree, I think nice to encapsulate vendor extension points in SPI, even if we end up with lots of them(?)

anuraaga: Sorry for missing the comment! SGTM

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/BootstrapPackagePrefixesHolder.java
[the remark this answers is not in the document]
pavolloffay: When the class is in `javaagent-bootstrap` (and added to helper class names in the classloader instrumentation) the set prefixes set by the `AgentInstaller` are always null - the agent installer has a different instance as the instrumentation.  

Loading the providers from the instrumentation fails at build time on the muzzle size limit.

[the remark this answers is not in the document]
pavolloffay: I have moved it to `internal` package in the same module.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -207,6 +211,23 @@ private static void addByteBuddyRawSetting() {
...
+            BootstrapPackagesProvider.class, AgentInstaller.class.getClassLoader());
+    for (BootstrapPackagesProvider provider : bootstrapPackagesProviders) {
+      List<String> packagePrefixes = provider.getPackagePrefixes();
+      log.debug(
+          "Loaded bootstrap package prefixes from {}: {}",
+          provider.getClass().getName(),
+          packagePrefixes);
+      bootstrapPackages.addAll(packagePrefixes);
+    }
+    return bootstrapPackages.toArray(new String[] {});
```
iNikem, line 228:
```suggestion
    return bootstrapPackages.toArray(new String[0]);
```

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+public class BootstrapPackagePrefixesHolder {
+
+  private static String[] bootstrapPrefixes;
+
+  public static String[] getBootstrapPrefixes() {
+    return bootstrapPrefixes;
+  }
+
+  public static void setBootstrapPrefixes(String[] prefixes) {
```
mateuszrzeszutek, line 24: Can you make this class more similar to `Config`/`OpenTelemetrySdkAccess`? I.e. `volatile` static field, `internal...` setter method name, checking that you can only set the value once.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
```diff
@@ -0,0 +1,27 @@
...
+ * <p>The prefixes are loaded by {@code AgentInstaller} and consumed by classloader instrumentation.
+ * The instrumentation does not have access to the installer, therefore this utility class is used
+ * to share package prefixes.
+ */
+public class BootstrapPackagePrefixesHolder {
+
+  private static String[] bootstrapPrefixes;
+
+  public static String[] getBootstrapPrefixes() {
+    return bootstrapPrefixes;
```
mateuszrzeszutek, line 21: If it's supposed to be available in `instrumentation-api` consider using a `List` (and `Collections.unmodifiableList()` wrapper): arrays are mutable and can be easily changed by anybody.

pavolloffay:
@mateuszrzeszutek I have updated the PR

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/BootstrapPackagePrefixesHolder.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public class BootstrapPackagePrefixesHolder {
+
+  private static volatile List<String> BOOSTRAP_PACKAGE_PREFIXES;
+
+  public static List<String> getBoostrapPackagePrefixes() {
+    return BOOSTRAP_PACKAGE_PREFIXES;
+  }
+
+  public static void setBoostrapPackagePrefixes(List<String> prefixes) {
```
mateuszrzeszutek, line 27: Just a minor comment: how about `internalSetBootstrapPackagePrefixes`?

mateuszrzeszutek (approved):

pavolloffay:
@trask @anuraaga could we get this in?

anuraaga (approved):

trask (approved):
👍

trask:
Closing and re-opening to re-trigger the stuck CLA check

PR #1383, opened 2020-10-14
Add server interceptor to end in javaagent because they are run in re…

anuraaga:
…verse order.

/cc @asarkar This fixes the javaagent which also made the wrong assumption that server interceptors are run in order. But you may be able to get some inspiration from the unit test changes.

### DIFF
--- a/instrumentation/grpc-1.5/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_5/AbstractGrpcTest.groovy
+++ b/instrumentation/grpc-1.5/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_5/AbstractGrpcTest.groovy
@@ -18,6 +18,7 @@ import io.grpc.Channel
 import io.grpc.ClientCall
 import io.grpc.ClientInterceptor
 import io.grpc.Context
+import io.grpc.Contexts
 import io.grpc.ManagedChannel
 import io.grpc.ManagedChannelBuilder
 import io.grpc.Metadata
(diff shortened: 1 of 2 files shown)


instrumentation/grpc-1.5/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_5/AbstractGrpcTest.groovy
```diff
@@ -323,25 +324,26 @@ abstract class AbstractGrpcTest extends InstrumentationSpecification {
     def port = PortUtils.randomOpenPort()
     Server server
     server = configureServer(ServerBuilder.forPort(port)
-      .addService(greeter))
+      .addService(greeter)
       .intercept(new ServerInterceptor() {
         @Override
         <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
-          def ctx = Context.current().withValue(key, "meow")
-          def oldCtx = ctx.attach()
```
anuraaga, line 331: This was wrong in my previous change, naturally just like TracingServerInterceptor, any interceptor that modifies context must use `Contexts.interceptCall` to make sure it's propagated.

anuraaga: Incidentally this is much easier when context is mutable :)

iNikem (approved):

mateuszrzeszutek (approved):

trask (approved):

PR #1387, opened 2020-10-14
Run build without warnings

### LINKED ISSUE #984: gradle build and gradle check are failing
This can be confusing for contributors, we should fix, and have them run on CI so they don't break again.

iNikem:
Fixes #984

Required test coverage levels are very relaxed atm. We need to address them separately.


instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JDBCConnectionUrlParser.java
```diff
@@ -68,7 +70,7 @@
    */
   JTDS_URL_LIKE() {
     @Override
-    DBInfo.Builder doParse(String jdbcUrl, DBInfo.Builder builder) {
+    Builder doParse(String jdbcUrl, Builder builder) {
```
anuraaga, line 73: Is the rename to generic name Builder intended?

iNikem: It's not rename, it's re-import! No, it was not intended.

iNikem: Reverted

trask (approved):
the reason for the strange `io.opentelemetry.javaagent.instrumentation.api.jdbc` package was as an optimization https://github.com/DataDog/dd-trace-java/pull/1208.

i'm good with moving it back to jdbc instrumentation module.

you'll need to add those classes back into `helperClassNames` to make muzzle happy.

makes sense to do the same with `io.opentelemetry.javaagent.instrumentation.api.rmi`

anuraaga (approved):

instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JDBCConnectionUrlParser.java
anuraaga: @jkwatson I just talked about this issue at the last meeting and we run into it in the same week :P It does make me consider biting the bullet and replacing AutoValue with top-level builders if it can prevent this sort of trouble for end users.

trask: I like `DBInfo.Builder` better than `DBInfoBuilder`, but ya, I see what you're talking about now, and would not disagree with such a change

iNikem:
> the reason for the strange `io.opentelemetry.javaagent.instrumentation.api.jdbc` package was as an optimization [DataDog/dd-trace-java#1208](https://github.com/DataDog/dd-trace-java/pull/1208).

Is this one more piece of tribal knowledge about muzzle? I was not aware that muzzle's performance may depend on our own class location.

mateuszrzeszutek:
> i'm good with moving it back to jdbc instrumentation module.

👍 

I think that we'll gain more performance in muzzle by implementing https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1233 anyway.

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

trask:
> Is this one more piece of tribal knowledge about muzzle? I was not aware that muzzle's performance may depend on our own class location.

everything in `helperClassNames()` is injected into the user's class loader, which is not free

PR #1397, opened 2020-10-15
apache camel 2.20.x instrumentation

### LINKED ISSUE #1356: Apache Camel 2.20.x instrumentation
Apache Camel team provides OpenTelemetry instrumentation for Camel > 3.5 (https://camel.apache.org/components/latest/others/opentelemetry.html). In order to support customers using older versions (especially 2.x due to a number of changes between 2.x and 3.x) we shall provide a skinny instrumentation.

kubawach:
Fixes #1356 

Instrumentation for Apache Camel 2.20.x
Apache Camel provides OTEL instrumentation (https://camel.apache.org/components/latest/others/opentelemetry.html) but for Camel >= 3.5. For customers using older versions that don't want / can't move to 3.5 or newer, a separate instrumentation library is provided.

*NOTE*
Large portions of the code has been adapted from existing Apache Camel OpenTracing library. In order to honour the license and work of the original Authors, I've added following header:
```
/*
 * Includes work from:
 * Copyright Apache Camel Authors
 */
```
to all files consisting at least some code from the library. 
The header was put under "package" in order to pass spotless:license plugin check.
I'm not entirely sure that the header is 100% compatible with Apache License 2.0 of the original code (see https://github.com/apache/camel/tree/camel-2.20.0/components/camel-opentracing). Therefore during review please have a look at this as well.


instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/decorators/HttpSpanDecorator.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    // 5. GET otherwise.
+    return GET_METHOD;
+  }
+
+  @Override
+  public String getOperationName(Exchange exchange, Endpoint endpoint) {
+    // Based on HTTP component documentation:
+    return getHttpMethod(exchange, endpoint);
```
mateuszrzeszutek, line 51: Is it possible to use the route for operation name? (ass suggested in [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#name))

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
```diff
@@ -0,0 +1,174 @@
...
+          attributes {
+            "$SemanticAttributes.HTTP_METHOD.key" "POST"
+            "$SemanticAttributes.HTTP_URL.key" "http://localhost:$portOne/serviceOne"
+            "$SemanticAttributes.HTTP_STATUS_CODE.key" 200
+            "camel.uri" "http://localhost:$portOne/serviceOne"
+          }
+        }
+        it.span(3) {
+          name "HTTP POST"
+          kind CLIENT
```
mateuszrzeszutek, line 102: Having two `CLIENT` spans in a row looks wrong -- [semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spankind) says that `CLIENT` must be a parent of `SERVER` span. 
Can you merge those two spans into one?

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
```diff
@@ -0,0 +1,174 @@
...
+            "$SemanticAttributes.HTTP_STATUS_CODE.key" 200
+            "$SemanticAttributes.NET_PEER_NAME.key" "0.0.0.0"
+            "$SemanticAttributes.NET_PEER_PORT.key" portTwo
+            "$SemanticAttributes.NET_TRANSPORT.key" "IP.TCP"
+            "$SemanticAttributes.HTTP_FLAVOR.key" "1.1"
+            "$SemanticAttributes.HTTP_USER_AGENT.key" "Jakarta Commons-HttpClient/3.1"
+          }
+        }
+        it.span(7) {
+          name "CachedChain.doFilter"
```
mateuszrzeszutek, line 148: Hmm, most HTTP server frameworks (JAX-RS, Spring, ...) usually rename their `SERVER` span so that it has the actual route as its name. IS it possible to do in camel?

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
```diff
@@ -0,0 +1,174 @@
...
+
+    when:
+    template.sendBody("direct:input", "Example request")
+
+    then:
+    assertTraces(1) {
+      trace(0, 9) {
+        it.span(0) {
+          name "input"
+          kind CLIENT
```
mateuszrzeszutek, line 78: This is wrong too, there's no `SERVER` child for that span. `direct` camel component isn't really a server by the way, so maybe this trace should start with an `INTERNAL` span? It's not really a real-life scenario anyway, since you can only call `direct:*` from inside.

instrumentation/apache-camel-2.20/apache-camel-2.20.gradle
```diff
@@ -0,0 +1,37 @@
...
+  testImplementation group: 'org.apache.camel', name: 'camel-spring-boot-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-spring-boot-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-jetty-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-http-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-jaxb-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-undertow', version: '2.20.1'
+
+  testImplementation 'javax.xml.bind:jaxb-api:2.3.1'
+
+  latestDepTestLibrary group: 'org.apache.camel', name: 'camel-core', version: '2.20.3'
```
mateuszrzeszutek, line 36: Why stop at 2.20.3? Won't thiswork for later 2.x versions?

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/CamelContextInstrumentation.java
```diff
@@ -0,0 +1,95 @@
...
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+
+    return not(isAbstract()).and(implementsInterface(named("org.apache.camel.CamelContext")));
+  }
+
+  @Override
+  public String[] helperClassNames() {
+    return new String[] {
+      getClass().getName() + "$ContextAdvice",
```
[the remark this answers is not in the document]
kubawach: 🙆‍♂️

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/CamelContextInstrumentation.java
```diff
@@ -0,0 +1,95 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.apache.camel.CamelContext;
+
+@AutoService(Instrumenter.class)
+public class CamelContextInstrumentation extends Instrumenter.Default {
+
+  public CamelContextInstrumentation() {
+    super("apachecamel", "apache-camel");
```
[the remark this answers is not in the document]
kubawach: In source lib authors always used "apache camel" not just "camel". I guess it's a matter of full, proper naming - ther eare many camels, but only one by apache ;)

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/CamelContextInstrumentation.java
```diff
@@ -0,0 +1,95 @@
...
+      "io.opentelemetry.instrumentation.auto.apachecamel.CamelEventNotifier",
+      "io.opentelemetry.instrumentation.auto.apachecamel.CamelRoutePolicy",
+      "io.opentelemetry.instrumentation.auto.apachecamel.CamelTracingService",
+      "io.opentelemetry.instrumentation.auto.apachecamel.CamelContextInstrumentation"
+    };
+  }
+
+  @Override
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    final Map<ElementMatcher<? super MethodDescription>, String> transformers = new HashMap<>();
```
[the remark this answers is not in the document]
kubawach: That's actually not that NITpicking - done.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/CamelTracingService.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.auto.apachecamel;
```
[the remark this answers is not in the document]
kubawach: It had to be really recently 😃

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/ActiveSpanManager.java
```diff
@@ -0,0 +1,105 @@
...
+      spanWithScope.deactivate();
+      exchange.setProperty(ACTIVE_SPAN_PROPERTY, spanWithScope.getParent());
+      if (LOG.isTraceEnabled()) {
+        LOG.trace("Deactivated span: " + spanWithScope);
+      }
+    }
+  }
+
+  public static class SpanWithScope {
+    private SpanWithScope parent;
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: It's not Camel, it's a servlet instrumentation ;)))

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: It's not something that Camel instrumentation can control. The second CLIENT span is generated by "commons-http-client" instrumentation that kicks in the middle of execution (ie is called by camel implementation).

instrumentation/apache-camel-2.20/apache-camel-2.20.gradle
kubawach: Haven't tested - can try latest 2.x (2.25.2)

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
mateuszrzeszutek: Then camel instrumentation should not generate the first span in the first place, I believe

instrumentation/apache-camel-2.20/apache-camel-2.20.gradle
mateuszrzeszutek: You can just remove this line if the latest one works

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
mateuszrzeszutek: I know, bu the point stays: the "web framework" usually changes the server/servlet span name. Can it be done in camel?

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: That's a really good question. 
IMO setting "direct" as CLIENT is OK, as "direct" is a way to call a route (with remote services in line). We may model this as a Camel equivalent of a HttpClient. However as you pointed out - this does not go in line with SemConv recommendation. There may be some client side processing - producing INTERNAL spans before equivalent SERVER span is created. Heck, the route can be entirely local (without any SERVER spans). Same would apply to eg "timer" spans. 
Anyone else would like to comment on that matter? @iNikem ? @trask ? :)

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/decorators/HttpSpanDecorator.java
kubawach: Theoretically - yes, obviously. Practically- not that much.
1. Http decorator serves multiple components - ahc, http, jetty, undertow.
2. Component route can include all GET parameters (query) - would yield high cardinality
3. Same decorator is used for both sides (SERVER, CLIENT) - so being conservative (always use HTTP method) won

At the same time if there is a consensus that using method in all cases is wrong, I'm open to discuss alternatives, like:
- adding more http decorators to handle different cases separately
- removing query
This however won't allow us to eg sanitize path fragments that are values and inevitably yields high cardinality.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/CamelPropagationUtil.java
```diff
@@ -0,0 +1,50 @@
...
+
+  static void injectParent(Context context, final Map<String, Object> exchangeHeaders) {
+    OpenTelemetry.getPropagators()
+        .getTextMapPropagator()
+        .inject(context, exchangeHeaders, MapSetter.INSTANCE);
+  }
+
+  private static class MapGetter implements Getter<Map<String, Object>> {
+
+    private static final MapGetter INSTANCE = new MapGetter();
```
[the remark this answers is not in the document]
kubawach: I believe that in OTEL such instances are always static finals (multiple places, really).

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/decorators/BaseSpanDecorator.java
```diff
@@ -0,0 +1,109 @@
...
+  public void post(Span span, Exchange exchange, Endpoint endpoint) {
+    if (exchange.isFailed()) {
+      span.setAttribute("error", true);
+      if (exchange.getException() != null) {
+        Attributes logEvent =
+            Attributes.newBuilder()
+                .setAttribute("error.kind", "Exception")
+                .setAttribute("message", exchange.getException().getMessage())
+                .build();
+        span.addEvent("error", logEvent);
```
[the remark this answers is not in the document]
kubawach: Didn't know about that feature, nice!

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/CamelPropagationUtil.java
```diff
@@ -0,0 +1,50 @@
...
+
+package io.opentelemetry.instrumentation.auto.apachecamel;
+
+import io.grpc.Context;
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.context.propagation.TextMapPropagator.Getter;
+import io.opentelemetry.context.propagation.TextMapPropagator.Setter;
+import java.util.Map;
+
+public class CamelPropagationUtil {
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/decorators/RestSpanDecorator.java
```diff
@@ -0,0 +1,89 @@
...
+    String path = null;
+    int index = uri.indexOf(':');
+    if (index != -1) {
+      index = uri.indexOf(':', index + 1);
+      if (index != -1) {
+        path = uri.substring(index + 1);
+        index = path.indexOf('?');
+        if (index != -1) {
+          path = path.substring(0, index);
+        }
```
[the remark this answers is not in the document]
kubawach: Programmer had an issue. He decided to fix it with regexp. Now programmer had two issues.

instrumentation/apache-camel-2.20/src/test/groovy/test/SingleServiceCamelSpringBootBasedTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+import spock.lang.Shared
+
+class SingleServiceCamelSpringBootBasedTest extends AgentTestRunner {
+
+  @Shared
+  ConfigurableApplicationContext server
+  @Shared
+  OkHttpClient client = OkHttpUtils.client()
+  @Shared
+  int port = PortUtils.randomOpenPort()
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: Then you're loosing data that can be valuable for customers. Also, Camel instr does not know if any internals used to execute the call are instrumented and will generate the span or not. How would you approach that?

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: If I understand correctly - you're asking if Camel can set the name of the servlet that is being exposed (by Jetty component in this case) and instrumented by servlet instrumentation, currently generating span with "CachedChain.doFilter" name? IF this is the question the answer is - I don't think so (see https://camel.apache.org/components/latest/jetty-component.html). Besides it's not something that Camel instrumentation would do.

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: As discussed offline, two ideas:
- create Camel spans as INTERNAL, counting on another instrumentation creating CLIENT just before sending a message 
- change some of our "last stage before sending" instrumentations (in this case - http client) to check if parent span was CLIENT - if so, change it's type to become INTERNAL

@trask @iNikem any thoughts on the above?

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: Also discussed offline. Will propse something, although this seems tricky, due to Camel nature.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/ActiveSpanManager.java
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * This method activates the supplied span for the supplied exchange. If an existing span is found
+   * for the exchange, this will be pushed onto a stack.
+   *
+   * @param exchange The exchange
+   * @param span The span
+   */
+  public static void activate(Exchange exchange, Span span) {
+
+    SpanWithScope parent = (SpanWithScope) exchange.getProperty(ACTIVE_SPAN_PROPERTY);
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/decorators/DbSpanDecorator.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+    return null;
+  }
+
+  private String getDbName(Endpoint endpoint) {
+    switch (component) {
+      case "mongodb":
+        {
+          Map<String, String> queryParameters = toQueryParameters(endpoint.getEndpointUri());
+          return queryParameters.toString();
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/instrumentation/auto/apachecamel/decorators/KafkaSpanDecorator.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Map;
+import org.apache.camel.Endpoint;
+import org.apache.camel.Exchange;
+
+class KafkaSpanDecorator extends MessagingSpanDecorator {
+
+  public static final String KAFKA_PARTITION_TAG = "kafka.partition";
+  public static final String KAFKA_PARTITION_KEY_TAG = "kafka.partition.key";
+  public static final String KAFKA_KEY_TAG = "kafka.key";
+  public static final String KAFKA_OFFSET_TAG = "kafka.offset";
```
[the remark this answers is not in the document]
kubawach: Well, the instrumentation for Kafka was changed 3 days ago, it was hard 7 days ago to reference it :D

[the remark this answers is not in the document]
kubawach: But yeah, let's align.

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: Done for HTTP components, although I'm not a huge fan of the implementation.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/ActiveSpanManager.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.apachecamel;
+/*
+ * Includes work from:
```
anuraaga, line 8: I'm trying this pattern and it seems to work well to keep the copyrights together, what do you think?

https://github.com/open-telemetry/opentelemetry-java/blob/master/build.gradle#L331

Also if the original isn't shortened I don't know if we're allowed to - IANAL but my understanding is we need to preserve the original header without modification.

kubawach: Excellent idea! Reformatted to put "includes" header right after our license. Also, unwinded the original one per suggestion.

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: Decided to reimplement internal components as "INTERNAL" spans. Data produced looks much better now.

instrumentation/apache-camel-2.20/src/test/groovy/test/TwoServicesCamelSpringBootBasedTest.groovy
kubawach: I believe there is a separate issue to handle such cases (https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/465). Will leave this as it is then.

kubawach:
@mateuszrzeszutek added new tests (multithreaded and simple internal direct). Also - added logic to prevent wrong CLIENT spas with all internal components. Please have another look :)

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/ActiveSpanManager.java
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * This method activates the supplied span for the supplied exchange. If an existing span is found
+   * for the exchange, this will be pushed onto a stack.
+   *
+   * @param exchange The exchange
+   * @param span The span
+   */
+  public static void activate(Exchange exchange, Span span) {
+
```
jkwatson, line 58: super minor...but why the extra CR here at the beginning of the method? I see this on some methods, but not all of them.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelEventNotifier.java
```diff
@@ -0,0 +1,95 @@
...
+          sd.post(span, ese.getExchange(), ese.getEndpoint());
+          ActiveSpanManager.deactivate(ese.getExchange());
+        } else {
+          LOG.warn("Could not find managed span for exchange " + ese.getExchange());
+        }
+      }
+    } catch (Throwable t) {
+      LOG.warn("Failed to capture tracing data", t);
+    }
+  }
```
jkwatson, line 84: this method is really big. What do you think about extracting methods for each of the two branches?

jkwatson:
I'm not a primary reviewer for this, but I wonder how it is possible to effectively review a 2222 line PR. Can this be broken up into a series of smaller PRs to make it easier for maintainers to review?

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelEventNotifier.java
kubawach: Done.

kubawach:
@iNikem @anuraaga @trask @tylerbenson please have a look, looking into closing this PR this week. Thanks in advance!

iNikem:
Honestly, I am waiting for @mateuszrzeszutek approval :) He has done such thorough first review.

kubawach:
@iNikem yep, I'm nagging him on a daily basis ;)

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/HttpSpanDecorator.java
```diff
@@ -0,0 +1,134 @@
...
+    return (serverSpan != null && CamelDirection.INBOUND.equals(camelDirection));
+  }
+
+  private void updateServerSpanName(Span serverSpan, String httpUrl) {
+    try {
+      URL url = new URL(httpUrl);
+      serverSpan.updateName(url.getPath());
+    } catch (MalformedURLException e) {
+      // ignored
+    }
```
mateuszrzeszutek, line 101: I guess that for server components (`jetty`, `servlet`, ...) it's fine to just use the path, there _shouldn't_ be any path parameters there. But for `rest` component we should use the route - can you make sure that `RestSpanDecorator` uses the correct value here?

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/HttpSpanDecorator.java
```diff
@@ -0,0 +1,134 @@
...
+
+    // 5. GET otherwise.
+    return GET_METHOD;
+  }
+
+  @Override
+  public String getOperationName(Exchange exchange, Endpoint endpoint) {
+    // Based on HTTP component documentation:
+    return getHttpMethod(exchange, endpoint);
```
mateuszrzeszutek, line 70: HTTP clients should use `HTTP METHOD` span name:

> HTTP client spans SHOULD be using conservative, low cardinality names formed from the available parameters of an HTTP request, such as "HTTP {METHOD_NAME}".

In case of inbound spans I think using the path would be fine here.

instrumentation/apache-camel-2.20/src/test/groovy/test/MulticastDirectCamelTest.groovy
```diff
@@ -0,0 +1,73 @@
...
+        it.span(0) {
+          name "input"
+          kind INTERNAL
+          hasNoParent()
+          attributes {
+            "camel.uri" "direct://input"
+          }
+        }
+        it.span(1) {
+          name "second"
```
mateuszrzeszutek, line 55: Are you sure that this is 100% deterministic?

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/ActiveSpanManager.java
```diff
@@ -0,0 +1,120 @@
...
+      if (LOG.isTraceEnabled()) {
+        LOG.trace("Deactivated span: " + spanWithScope);
+      }
+    }
+  }
+
+  public static class SpanWithScope {
+    @Nullable private SpanWithScope parent;
+    private Span span;
+    private Scope scope;
```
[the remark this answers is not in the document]
kubawach: +3

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelEventNotifier.java
```diff
@@ -0,0 +1,102 @@
...
+        onExchangeSent((ExchangeSentEvent) event);
+      }
+    } catch (Throwable t) {
+      LOG.warn("Failed to capture tracing data", t);
+    }
+  }
+
+  /** Camel about to send (outbound). */
+  private void onExchangeSending(ExchangeSendingEvent event) {
+    ExchangeSendingEvent ese = event;
```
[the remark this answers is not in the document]
kubawach: Nope.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/BaseSpanDecorator.java
```diff
@@ -0,0 +1,117 @@
...
+    return true;
+  }
+
+  @Override
+  public String getOperationName(Exchange exchange, Endpoint endpoint) {
+    String[] splitURI = StringHelper.splitOnCharacter(endpoint.getEndpointUri(), ":", 2);
+    if (splitURI.length > 0) {
+      return splitURI[0];
+    } else {
+      return null;
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/KafkaSpanDecorator.java
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Extracts header value from the exchange for given header
+   *
+   * @param exchange the {@link Exchange}
+   * @param header the header name
+   * @param type the class type of the exchange header
+   * @return
+   */
+  private <T> String getValue(final Exchange exchange, final String header, Class<T> type) {
+    T partition = exchange.getIn().getHeader(header, type);
```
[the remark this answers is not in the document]
kubawach: +1

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/HttpSpanDecorator.java
kubawach: It's for inbound only - so jetty, servlet, undertow, restlet server. After previous code review i decided to use path (instead of operation name). What do you mean by "route" then ?

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/RestSpanDecorator.java
```diff
@@ -0,0 +1,104 @@
...
+    super.pre(span, exchange, endpoint, camelDirection);
+
+    getParameters(getPath(endpoint.getEndpointUri()))
+        .forEach(
+            param -> {
+              Object value = exchange.getIn().getHeader(param);
+              if (value != null) {
+                span.setAttribute(param, (String) value);
+              }
+            });
```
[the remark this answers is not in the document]
kubawach: This was Camel's authors decision to store all REST parameters as separate attributes. The idea behind that is a mystery for me. The question is - does this fit into our (OTEL) spec? It's clearly not "semantic" but may be of some benefit for customers.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/BaseSpanDecorator.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Override
+  public void pre(Span span, Exchange exchange, Endpoint endpoint, CamelDirection camelDirection) {
+    span.setAttribute("camel.uri", URISupport.sanitizeUri(endpoint.getEndpointUri()));
+  }
+
+  @Override
+  public void post(Span span, Exchange exchange, Endpoint endpoint) {
+    if (exchange.isFailed()) {
+      span.setAttribute("error", true);
```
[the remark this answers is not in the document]
kubawach: Digging deeper - I'll change the attribute to status (span.setStatus(StatusCanonicalCode.ERROR). Adding an exception however IMO goes against the spec that specifically describes exception (not error in general) handling. Also looking into eg BaseDecorator class we see that there is no "empty exception" added if throwable passed is null.

instrumentation/apache-camel-2.20/src/test/groovy/test/MulticastDirectCamelTest.groovy
kubawach: In all of the tests I performed it was, although I could not find a way to prove this synthetically.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/HttpSpanDecorator.java
kubawach: Yep, we can do that, since we already update server span.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/RestSpanDecorator.java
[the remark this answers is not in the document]
mateuszrzeszutek: All other web frameworks do not log params. Besides, the full URI is still available as `http.url`, right?

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/HttpSpanDecorator.java
mateuszrzeszutek: > What do you mean by "route" then ?

The URI template, like `/resource/{id}/something`

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/RestSpanDecorator.java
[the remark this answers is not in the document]
kubawach: It is - but as `camel.uri`. I can remove the params, the question is if a customer already using Camel instrumentation won't miss these.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/HttpSpanDecorator.java
kubawach: But it is - have a look at the `RestSpanDecorator` - it logs "path" of the rest component.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/DbSpanDecorator.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+    return null;
+  }
+
+  @Override
+  public void pre(Span span, Exchange exchange, Endpoint endpoint, CamelDirection camelDirection) {
+    super.pre(span, exchange, endpoint, camelDirection);
+
+    span.setAttribute(SemanticAttributes.DB_SYSTEM, system);
+    String statement = getStatement(exchange, endpoint);
```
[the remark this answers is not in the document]
kubawach: Yep. Good idea.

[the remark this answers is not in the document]
kubawach: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1499

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/DbSpanDecorator.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+    return null;
+  }
+
+  private String getDbName(Endpoint endpoint) {
+    switch (component) {
+      case "mongodb":
+        {
+          Map<String, String> queryParameters = toQueryParameters(endpoint.getEndpointUri());
+          return queryParameters.get("database");
```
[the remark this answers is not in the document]
kubawach: Good catch, should serialize all params for statement.

instrumentation/apache-camel-2.20/apache-camel-2.20.gradle
```diff
@@ -0,0 +1,37 @@
...
+  testImplementation group: 'org.apache.camel', name: 'camel-spring-boot-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-spring-boot-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-jetty-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-http-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-jaxb-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-undertow', version: '2.20.1'
+
+  testImplementation 'javax.xml.bind:jaxb-api:2.3.1'
+
+  latestDepTestLibrary group: 'org.apache.camel', name: 'camel-core', version: '2.25.2'
```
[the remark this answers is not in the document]
kubawach: I don't really believe that there will be more releases of 2.x stream, but yeah, why not.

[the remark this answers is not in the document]
mateuszrzeszutek: Yeah, it's more for the readability - when you see `2.+` you're immediately sure that it covers all versions.

instrumentation/apache-camel-2.20/src/test/groovy/test/RestCamelTest.groovy
```diff
@@ -0,0 +1,117 @@
...
+          kind CLIENT
+          attributes {
+            "$SemanticAttributes.HTTP_METHOD.key" "GET"
+            "$SemanticAttributes.HTTP_STATUS_CODE.key" 200
+            "camel.uri" "rest://get:api/%7Bmodule%7D/unit/%7BunitId%7D"
+          }
+        }
+        it.span(2) {
+          name "/api/firstModule/unit/unitOne"
+          kind SERVER
```
mateuszrzeszutek, line 86: Would it be possible to make the method that generates parent span name protected and override it in rest decorator? Ideally the span name would be `/api/{module}/unit/{unitId}`

kubawach: Done, now also server span update for REST follows desired convention.

kubawach:
@mateuszrzeszutek pls have another look ;)

mateuszrzeszutek (approved):

iNikem (approved):

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/ActiveSpanManager.java
kubawach: Usually I add extra CR when I believe it increases readability of the code (due to long method signature). Rule of thumb frankly.

instrumentation/apache-camel-2.20/apache-camel-2.20.gradle
```diff
@@ -0,0 +1,37 @@
...
+  testImplementation project(':instrumentation:servlet:servlet-2.2')
+  testImplementation project(':instrumentation:servlet:servlet-3.0')
+
+  testImplementation group: 'org.spockframework', name: 'spock-spring', version: "$versions.spock"
+
+  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test', version: '1.5.17.RELEASE'
+  testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter', version: '1.5.17.RELEASE'
+
+  testImplementation group: 'org.apache.camel', name: 'camel-spring-boot-starter', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-spring-boot-starter', version: '2.20.1'
```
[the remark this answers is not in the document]
kubawach: Yes, removed.

instrumentation/apache-camel-2.20/apache-camel-2.20.gradle
```diff
@@ -0,0 +1,37 @@
...
+    versions = "[2.20.1,3)"
+  }
+}
+
+dependencies {
+  library group: 'org.apache.camel', name: 'camel-core', version: '2.20.1'
+
+  testImplementation project(':instrumentation:apache-httpclient:apache-httpclient-2.0')
+  testImplementation project(':instrumentation:servlet:servlet-2.2')
+  testImplementation project(':instrumentation:servlet:servlet-3.0')
```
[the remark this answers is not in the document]
kubawach: I was under the impression that both are needed for tests (different camel components using different servlet impl under the hood). After retest - only 3.0 is needed, so removing the other one.

instrumentation/apache-camel-2.20/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/DbSpanDecorator.java
```diff
@@ -0,0 +1,138 @@
...
+
+import io.opentelemetry.javaagent.instrumentation.apachecamel.CamelDirection;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.attributes.SemanticAttributes;
+import java.net.URI;
+import java.util.Map;
+import org.apache.camel.Endpoint;
+import org.apache.camel.Exchange;
+
+class DbSpanDecorator extends BaseSpanDecorator {
```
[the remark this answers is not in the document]
kubawach: Following are public:

PR #1398, opened 2020-10-15
Add operation and collection attributes for MongoDB spans

### LINKED ISSUE #1278: Implement db.mongodb.collection semantic attribute
It is [required by specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#call-level-attributes-for-specific-technologies).

jamal:
This change introduces `db.operation` and `db.mongodb.collection` attributes to the mongo client instrumentation. Because there is no common functionality for `db.operation` in `DatabaseClientTracer` we are just overriding `onConnect` to add mongo specific attributes.

In the future, this should be refactored when `db.operation` is added across all db instrumetnation.

Fixes #1278

### DIFF
--- a/instrumentation/mongo/mongo-3.1/src/test/groovy/MongoClientTest.groovy
+++ b/instrumentation/mongo/mongo-3.1/src/test/groovy/MongoClientTest.groovy
@@ -3,12 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import io.opentelemetry.javaagent.instrumentation.mongo.MongoClientTracer
-
-import static io.opentelemetry.instrumentation.test.utils.PortUtils.UNUSABLE_PORT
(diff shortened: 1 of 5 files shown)


trask (commented):
Thanks @jamal!

Looks like CI is failing due to formatting, if you run `./gradlew spotlessApply` that will fix it up.

jamal:
Thanks for the feedback @trask! I've updated the tests based on your recommendation (updated the other ones as well to be consistent). I don't really understand what happened with spotlessApply. When I ran it in Windows it didn't find any issues, but running it on Linux worked. Hopefully this should do it but I'll keep an eye on the build status and will fix any more issues if they come up. Thanks so much!

trask (approved):
Thanks @jamal 👍

iNikem (approved):

Issue #1223, opened 2020-09-19
JDBCConnectionUrlParser parse sqlserver IPv6 addresses error


dengliming:
20:18:29.120 [Test worker] DEBUG io.opentelemetry.instrumentation.auto.api.jdbc.JDBCConnectionUrlParser - Error parsing URL
```
java.lang.NumberFormatException: For input string: "8311:eeee:f70f:0:5eae:10.203.31.9\instance1"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at io.opentelemetry.instrumentation.auto.api.jdbc.JDBCConnectionUrlParser$3.doParse(JDBCConnectionUrlParser.java:175)
	at io.opentelemetry.instrumentation.auto.api.jdbc.JDBCConnectionUrlParser$9.doParse(JDBCConnectionUrlParser.java:382)
	at io.opentelemetry.instrumentation.auto.api.jdbc.JDBCConnectionUrlParser.parse(JDBCConnectionUrlParser.java:802)
	at io.opentelemetry.instrumentation.auto.api.jdbc.JDBCConnectionUrlParser$parse.callStatic(Unknown Source)
	at JDBCConnectionUrlParserTest.$spock_feature_0_1(JDBCConnectionUrlParserTest.groovy:56)
	at jdk.internal.reflect.GeneratedMethodAccessor11.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.spockframework.util.ReflectionUtil.invokeMethod(ReflectionUtil.java:200)
	at org.spockframework.runtime.model.MethodInfo.invoke(MethodInfo.java:113)
	at org.spockframework.runtime.BaseSpecRunner.invokeRaw(BaseSpecRunner.java:484)
	at org.spockframework.runtime.BaseSpecRunner.invoke(BaseSpecRunner.java:467)
	at org.spockframework.runtime.BaseSpecRunner.runFeatureMethod(BaseSpecRunner.java:408)
	at org.spockframework.runtime.BaseSpecRunner.doRunIteration(BaseSpecRunner.java:322)
[...]

PR #1403, opened 2020-10-16
ipv6 support for modified-URL-like

imavroukakis:
Closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1223


trask (approved):
Nice 👍

instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JDBCConnectionUrlParser.java
```diff
@@ -159,7 +188,19 @@
...
-      int portLoc = serverName.indexOf(":");
+      Matcher ipv6Matcher = IPv6.matcher(serverName);
+      boolean isIpv6 = ipv6Matcher.find();
+
+      int portLoc = -1;
+      if (isIpv6) {
+        if (serverName.startsWith("[")) {
+          portLoc = serverName.indexOf("]:") + 1;
+        } else {
+          serverName = "[" + serverName + "]";
```
iNikem, line 199: There cannot be port in this case?

iNikem (approved):

instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JDBCConnectionUrlParser.java
imavroukakis: There isn't one, but because the `shortUrl` contains the port number, it needs to be formatted that way
`shortUrl` example `sqlserver://[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:1433`. I'm now thinking though, whether this is not appropriate for the `host` attribute i.e. `[2001:0db8:85a3:0000:0000:8a2e:0370:7334]`. What do you think?

By the way, the `mariadb` example is partially incorrect because it includes the port as part of the `shortUrl`

```
"jdbc:mariadb:loadbalance://[2001:0660:7401:0200:0000:0000:0edf:bdd7]:33,mdb.host/mdbdb"|null|"mariadb:loadbalance://2001:0660:7401:0200:0000:0000:0edf:bdd7:33"|"mariadb"|"loadbalance"|null|"2001:0660:7401:0200:0000:0000:0edf:bdd7"|33|null|"mdbdb"
```

iNikem: You lost me :) I am not proficient with ip6 formats. I just see:
```
if (serverName.startsWith("[")) {
   portLoc = XXX
} else {
  serverName = "[" + serverName + "]";
}
```
and wonder, why if `serverName` does not start with `[`, then it means there is no port?

imavroukakis: Apologies, I went neck-deep into IPv6 land. Yes, since `:` is a delimiter for the IPv6 parts, the only way to define a port in a URL/URI-like manner, is to wrap the address in `[`-`]` and append the port with a `:`.

Does this make sense, especially with regards to my point about the MariaDB tests?

imavroukakis: Clarification, when I say 

>the only way to define a port in a URL/URI-like manner

, I should have said instead, "the only _unambigious_ way to define a port in a URL/URI-like manner"

iNikem: I see. I have left a comment in #1421, but that is a separate issue.

Issue #1287, opened 2020-09-29
Reduce magic of selecting SpanExporterFactory and friends


trask:
See https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1253#discussion_r494221024 and https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1262#discussion_r496878453 (and I think it was brought up once before that too 😄)

anuraaga:
My suggestion before was to have a hard-coded map of well-known exporters (what's in spec) and use fully-qualified class name for non-well-known exporters. This is mainly to prevent having collisions of names between two exporters - but maybe UX for a custom exporter like stackdriver isn't very good with it so defining the name as an API seems ok too.

iNikem:
We have to prevent a situation when some external exporter claims to have a name of some built-in one.

trask:
I thought `OTEL_EXPORTER` only applies to built-in exporters?

I was imagining vendor distributions would just have their one exporter hard-coded (no need for `OTEL_EXPORTER`).

And external exporter jar files are specified using `OTEL_EXPORTER_JAR`.

iNikem:
Don't think about 1 exporter, think about 3 exporters: for traces, metrics and (in the future) logs. Vendor can hard-code some of them, external jar can contain some of them and some may be built-in.

pavolloffay:
> I thought OTEL_EXPORTER only applies to built-in exporters?

Is there anything blocking us from loading vendor exporters via this setting? From the user-experience it would be better to keep the consistency and allow loading custom exporters.

iNikem:
The question is: where to load it from? If it is vendor distribution which packed its export in - sure, we should to that. Otherwise - how?

trask:
Does this proposal address the above concerns?
* `OTEL_EXPORTER` only applies to exporters that are bundled into the (otel or vendor) distribution
* add `List<String> getExporterNames()` to `SpanExporterFactory` and `MetricServer`
  * returning `List` e.g. allows OtelSpanExporterFactory to be enabled by both `otel_span` and `otel`

iNikem:
How can we/should we prevent vendor exporter to provide the same exporter name as one of the Otel's built in?

trask:
> How can we/should we prevent vendor exporter to provide the same exporter name as one of the Otel's built in?

Vendor distributions can pretty much do whatever they want, and it's not our responsibility(?)

[...]

PR #1404, opened 2020-10-16
fix finding SpanExporterFactory by exporterName with underscore e.g. "otlp_span"

### LINKED ISSUE #1401: No otlp_span span exporter found
```
java -javaagent:opentelemetry-javaagent-0.9.0-all.jar \
   -Dotel.exporter=otlp_span \
   -jar target/opentelemetry-agent-example.jar 
[opentelemetry.auto.trace 2020-10-16 15:03:48:051 +0800] [main] WARN io.opentelemetry.javaagent.tooling.TracerInstaller - No otlp_span span exporter found
[opentelemetry.auto.trace 2020-10-16 15:03:48:070 +0800] [main] INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent - version: 0.9.0
```

In 0.9.0 release otlp_span exporter should be supported: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1253

cc @pavolloffay

malafeev:
fix #1401


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -127,9 +127,8 @@ private static SpanExporterFactory findSpanExporterFactory(String exporterName)
       if (spanExporterFactory
           .getClass()
           .getSimpleName()
-          .replace("_", "")
           .toLowerCase()
-          .startsWith(exporterName.toLowerCase())) {
+          .startsWith(exporterName.replace("_", "").toLowerCase())) {
```
mateuszrzeszutek, line 131: We need to fix that black magic exporter matching soon...

Can you add the same fix to `MetricExporterFactory` and `MetricServer` methods?

Thanks!

trask: can we make it so that `otlpspan` does not work? I think better to be strict about only supporting the values in the spec https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md#exporter-selection

malafeev: done

trask: i think `otlpspan` will still work with latest change?

malafeev: @trask, yes, `otlpspan` still works. If I will check for strict list of supported values then how user will be able to specify custom exporter?

trask: ok, makes sense, we can tackle this later in #1287

trask (approved):

iNikem (approved):

Issue #1395, opened 2020-10-15
Make muzzle reference creation package(s) configurable


pavolloffay:
The `instrumentationMuzzle` is created only for instrumenters in `io.opentelemetry.instrumentation.`

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/dadaac9a8cba35e3054fea7e62996e9b6c3e1693/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceCreationPredicate.java#L15

In my custom java agent build I am adding instrumentations in a different package and these instrumentations should have also `instrumentationMuzzle` properly created.

The package prefixes could be configured via a system property which would be set at build time. I don't think we need an SPI for it.

mateuszrzeszutek:
It needs to be available both in compile-time and runtime - `ReferenceMatcher` is also using this predicate.

Are these the same packages as in `BootstrapPackagePrefixesHolder`?

By the way, I'm thinking that maybe we need to have some sort of `AgentInternalConfig` class that groups all such properties together, and a single SPI for changing that config (similar to `TracerCustomizer`). This way a vendor distribution could configure all its needs in a single place.

pavolloffay:
> Are these the same packages as in BootstrapPackagePrefixesHolder?

no, the packages in `BootstrapPackagePrefixesHolder` are different (not instrumentation) - these are classes that provide common functionality for instrumentations.

mateuszrzeszutek:
Another idea based on: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2524#discussion_r589388031

Right now if we want to include a 3rd party library instrumentation in the javaagent (e.g. [couchbase 3.1 instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2524)) we're forced to add all classes used in the library instrumentation as `additionalHelperClassNames()`.
That's because muzzle collects neither helper classes nor references (used symbols) from 3rd party packages/classes.

To fix that a new customization point should be introduced to the `InstrumentationModule`: a method that returns a `Predicate<String>` (or similar) that returns a predicate determining whether a passed class name refers to an instrumentation helper class. The default behavior would be the current one e.g. `InstrumentationClassPredicate`.

To sum it up, instead of using SPI (as proposed above) we should allow customizing this behavior per instrumentation module.

Issue #1225, opened 2020-09-20
Build and publish muzzle check plugin separately


iNikem:
When trying to run muzzle in parallel in GHA, building projects `buildSrc` takes 2/3 of each task execution. This is total waste of time. Should build it once, publish and consume externally.

iNikem:
Damn :( muzzle plugin, living inside `buildSrc`, reaches to `javaagent-bootstrap` and `javaagent-tooling` to grab their runtime classpaths :( This makes extracting plugin almost impossible :(

pavolloffay:
Is there any way how it could be done? I am happy to help with this.

iNikem:
I don't know one right now.

pavolloffay:
@iNikem Is this the problem?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/4624584979c87dde9e90c8fc2275201601d4d798/buildSrc/src/main/groovy/MuzzlePlugin.groovy#L62-L63

I would like to use the muzzle plugin in a custom build with added instrumentations. The muzzle would be very helpful there.

iNikem:
Yes, this is it.

pavolloffay:
Commenting from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1408

The muzzle task could expose a property to configure classpath with `javaagent-tooling` and `javaagent-bootstrap` artifacts. E.g something similar to what ByteBuddy plugin does
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/gradle/instrumentation.gradle#L36

```groovy
  byteBuddy {
    transformation {
      tasks = ['compileJava', 'compileScala', 'compileKotlin']
      plugin = 'io.opentelemetry.javaagent.tooling.muzzle.collector.MuzzleCodeGenerationPlugin'
      classPath = project(':javaagent-tooling').configurations.instrumentationMuzzle + configurations.runtimeClasspath + sourceSets.main.output
    }
  }
```

mateuszrzeszutek:
A bit of clarification: this issue concerns the muzzle check plugin (`MuzzlePlugin` currently), not the muzzle compile one - there's a separate issue for the muzzle compile plugin extraction.

iNikem:
Done

PR #1408, opened 2020-10-16
Make muzzle task depend on instrumentation project runtimeclasspath

pavolloffay:

This is related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1225. See the hacks what I did in regards to runtime classpath of the plugin. Maybe the plugin could expose a setting to configure javaagent-tooling and javaagent-bootstrap classpath (a similar what bytebuddy plugin does https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/ee0e0a96974fac8eba802b400201ff2a5ee9da75/gradle/instrumentation.gradle#L36)

I am using muzzle plugin in my custom OTEL agent build. There are a couple of hacks that I had to make:
* copy buildSrc to my project (obvious and ugly!)
* define `javaagent-tooling` and `javaagent-bootstrap` subprojects that pull in appropriate OTEL artifacts
* move my instrumentation classes to `io.opentelemetry.instrumentation.`  https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1395
* set correctly classpath in bytebuddy plugin (`instrumentationMuzzle`)

With all this I was able to run muzzle task and properly generate `instrumentationMuzzle`. However the build was passing only after `./gradlew assemble`. Running `./gradlew clean muzzle` was always failing. The following fixed the build on a clean project.

### DIFF
--- a/buildSrc/src/main/groovy/MuzzlePlugin.groovy
+++ b/buildSrc/src/main/groovy/MuzzlePlugin.groovy
@@ -327,6 +327,7 @@ class MuzzlePlugin implements Plugin<Project> {
     }
 
     def muzzleTask = instrumentationProject.task(taskName) {
+      dependsOn(instrumentationProject.configurations.named("runtimeClasspath"))
       doLast {
         ClassLoader instrumentationCL = createInstrumentationClassloader(instrumentationProject, toolingProject)
         def ccl = Thread.currentThread().contextClassLoader


iNikem:
> Maybe the plugin could expose a setting to configure javaagent-tooling and javaagent-bootstrap classpath

This is the great idea! If we set required classpaths externally, then we can easily do #1225

iNikem (approved):

PR #1417, opened 2020-10-19
Remove MoreAttributes class

trask:


iNikem (approved):

anuraaga (approved):
That was easier than I was anticipating :-)

PR #1418, opened 2020-10-19
Add context root to the front of the route in the span name

### LINKED ISSUE #1299: Add context root to the front of the route in the span name
Currently if you deploy a war file with context root "test", we will capture a span name similar to

```
GET /controller
```

but I think we should capture

```
GET /test/controller
```

trask:
Resolves #1299

This PR does a few things, please see commit breakout:

* Add application root to span name
  * This commit keeps track of the context path, and prepends it to the route, e.g. old span name was '/mycontroller', new span name is '/mycontext/mycontroller'. If there is no context path (app is deployed to root) then there is no change
* Reduce duplicate context lookups
  * Since the prior commit added extra context lookups, this commit consolidated them where there was already a context lookup being done
* User servlet path as route
  * Previously, we had span names like `ApplicationFilterChain.doFilter` and `HttpServlet.service`. Now it uses `request.getServletPath()` for the span name instead.
* Test against jetty instrumentation
  * The jetty instrumentation occurs before getContextPath() and getServletPath() are available, so a few changes were needed to handle this better (see comments in this commit for alternative options)

Currently missing tests for non-root context path in
* `jaxrs-1.0`
* `webflux`


trask:
this is ready for review

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -30,6 +30,11 @@
   public static final ContextKey<Span> CONTEXT_CLIENT_SPAN_KEY =
       ContextKey.named("opentelemetry-trace-auto-client-span-key");
 
+  // Keeps track of the application root (e.g. servlet context path) that needs to be prepended
```
anuraaga, line 33: Isn't this too specific a use case to put in BaseTracer?

anuraaga: Also is context the only way? I thought servlet has `getServletPath` and `getRequestUri` we'd be able to compute the context path based on it?

https://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html

trask: > Isn't this too specific a use case to put in BaseTracer?

That's a good idea, I can move this to something like an `ApplicationRoot` utility class in `instrumentation-api`

> Also is context the only way? I thought servlet has `getServletPath` and `getRequestUri` we'd be able to compute the context path based on it?

I'm not following this suggestion, can you expand?

anuraaga: > I'm not following this suggestion, can you expand?

I also might not be following the problem :) But if the problem is the URL we populate is incomplete, I would assume something on HttpServlerRequest would be sufficient to provide the complete URL?

trask: Got it, ya, that's what this PR does. When servlet starts, it captures `request.getContextPath()` and stores in the context, so that later on, when routes are updated, they can pre-pend the context path in front of the route.

anuraaga: I see I think I follow - there are places we compute the route and update name, but have no access to the HttpServletRequest. It feels a bit weird to put a string into Context like this, I'd probably put HttpServletRequest itself to indicate we are propagating information about the request. 

Also instead of `ApplicationRoot` utility, I was hoping this could go in HttpServletTracer since it's specific to servlet.

trask: > I see I think I follow - there are places we compute the route and update name, but have no access to the HttpServletRequest. It feels a bit weird to put a string into Context like this, I'd probably put HttpServletRequest itself to indicate we are propagating information about the request.

We can't have a `ContextKey<HttpServletRequest>` in `instrumentation-api`, since `instrumentation-api` doesn't have access to `HttpServletRequest`.

We could do `ContextKey<Object>` and cast it when we pull it out, but that doesn't feel great either.

> Also instead of `ApplicationRoot` utility, I was hoping this could go in HttpServletTracer since it's specific to servlet.

It needs to be accessed from Tracers which don't extend HttpServletTracer.

We could still put it in HttpServletTracer, and access it from those Tracers, do u think that's better? I don't have strong opinion.

trask:
See the updated the description for explanation of the different things in this PR.

I can submit the commits one-by-one if that's preferred.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
anuraaga: Perhaps we could have a servlet-utils with a static helper then - actually I had no idea webflux works with servlet since our instrumentation doesn't have a dependency on it! So it seems to make sense that we could have a package dedicated to helping out servlet implementations. Having it outside the main instrumentation API would make me a lot less anxious.

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
```diff
@@ -26,23 +27,24 @@
...
+    Span serverSpan = BaseTracer.getCurrentServerSpan(context);
 
     // When jax-rs is the root, we want to name using the path, otherwise use the class/method.
     String spanName;
     if (serverSpan == null) {
       spanName = pathBasedSpanName;
     } else {
       spanName = spanNameForMethod(target, method);
-      updateServerSpanName(serverSpan, pathBasedSpanName);
+      updateServerSpanName(context, serverSpan, pathBasedSpanName);
```
anuraaga, line 39: I suspect we don't need this since we calculate the server span name correctly in the tracer IIUC. The fact that no tests changed means either this update isn't necessary or we're missing a test :)

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v2_0/JaxRsAnnotationsTracer.java
```diff
@@ -32,14 +33,16 @@ public Span startSpan(Class<?> target, Method method) {
...
+    updateSpanNames(context, span, BaseTracer.getCurrentServerSpan(context), target, method);
     return span;
   }
 
-  public void updateSpanNames(Span span, Span serverSpan, Class<?> target, Method method) {
-    // When jax-rs is the root, we want to name using the path, otherwise use the class/method.
-    String pathBasedSpanName = getPathSpanName(target, method);
+  public void updateSpanNames(
+      Context context, Span span, Span serverSpan, Class<?> target, Method method) {
+    String pathBasedSpanName =
```
anuraaga, line 44: Here also I suspect `updateSpanName(serverSpan)` isn't doing anything, and for the non-serverspan case we don't need the `getApplicationRoot` call

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/HandlerAdapterAdvice.java
```diff
@@ -51,7 +51,8 @@ public static SpanWithScope methodEnter(
       PathPattern bestPattern =
           exchange.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
       if (serverSpan != null && bestPattern != null) {
-        serverSpan.updateName(bestPattern.getPatternString());
+        serverSpan.updateName(
```
anuraaga, line 54: I think for webflux we can access the `ServletHttpRequest` directly with an instanceof for the `ServerRequest`

https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/server/ServletServerHttpRequest.java#L57

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringWebMvcTracer.java
```diff
@@ -31,12 +32,12 @@ public Span startSpan(ModelAndView mv) {
...
   }
 
-  public void onRequest(Span span, HttpServletRequest request) {
+  public void onRequest(Context context, Span span, HttpServletRequest request) {
     if (request != null) {
       Object bestMatchingPattern =
           request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
       if (bestMatchingPattern != null) {
-        span.updateName(bestMatchingPattern.toString());
+        span.updateName(BaseTracer.getApplicationRoot(context) + bestMatchingPattern.toString());
```
anuraaga, line 40: Let's just use the `HttpServletRequest`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
anuraaga: That being said if my below suggestions make sense I may have been able to avoid the propagation requirement 🤞

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
trask: we didn't have any JAX-RS test that used a non-root context path, I added one now

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v2_0/JaxRsAnnotationsTracer.java
trask: I think this is needed, I added a test for this now

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/RouteOnSuccessOrError.java
```diff
@@ -41,7 +41,8 @@ public void accept(HandlerFunction<?> handler, Throwable throwable) {
 
           Span serverSpan = context.get(BaseTracer.CONTEXT_SERVER_SPAN_KEY);
           if (serverSpan != null) {
-            serverSpan.updateName(parseRoute(predicateString));
+            serverSpan.updateName(
+                BaseTracer.getApplicationRoot(context) + parseRoute(predicateString));
```
trask, line 45: I didn't see how to get context path directly here, but will look at it closer tomorrow

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
trask: I think we need propagation at least for JAX-RS.

I agree there's danger of making it look too general, and people thinking that they need to always use it for all instrumentation.

A new module `:instrumentation:servlet:servlet-context-path` that can be imported into other instrumentation as needed seems nice.

Though if the initial filter/servlet isn't in the same class loader as the jax-rs resource, that may not work, because we will inject the class w/ context key into each place, and they won't share.

What do you think of putting the context key in `io.opentelemetry.instrumentation.api.servlet.ContextPath`, to make it look less general?

Or create `servlet-context-path` as a secondary "instrumentation-api" that we include in the bootstrap class loader?

Or...

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
```diff
@@ -26,23 +27,24 @@
...
     }
 
     return tracer.spanBuilder(spanName).startSpan();
   }
 
-  private void updateServerSpanName(Span span, String spanName) {
+  private void updateServerSpanName(Context context, Span span, String spanName) {
     if (!spanName.isEmpty()) {
-      span.updateName(spanName);
+      span.updateName(BaseTracer.getApplicationRoot(context) + spanName);
```
iNikem, line 47: How can this work? I see in JAX-RS 1.0 tests that we expect span name to be `POST /test/hello/{name}`. And now you just prepend application root to it?

iNikem, line 47: Also I don't like this scattered non-obvious logic :( Several places which are very far from each other and not really related now have to implement the same rule for span name. If/when we change them again, it is certain we don't find all these places.

At the very least we should encapsulate this name rule and application root access into a separate class. E.g. `ServletSpanName` or something.

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
iNikem: But you did not for JAX-RS 1.0 :)

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringWebMvcTracer.java
```diff
@@ -36,7 +36,12 @@ public void onRequest(Span span, HttpServletRequest request) {
       Object bestMatchingPattern =
           request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
       if (bestMatchingPattern != null) {
-        span.updateName(bestMatchingPattern.toString());
+        String contextPath = request.getContextPath();
+        if (contextPath != null && !contextPath.isEmpty() && !contextPath.equals("/")) {
+          span.updateName(contextPath + bestMatchingPattern.toString());
```
iNikem, line 41: This is essentially one more way to calculate span name. Why we don't use `BaseTracer.getApplicationRoot` here? We have similar but something different rule scattered across the code base.

iNikem (changes requested):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
anuraaga: Yeah I'm trying to find a way to avoid extra API for a seemingly very specific use case.

I have a theory that any request handled by servlet always has access to the `HttpServletRequest` somehow :) It's one of the reasons I've been persistent here.

Adding a break point at `updateSpanNames` and see what's available when running the tests has helped find some options I think.

I found for resteasy, we can get `HttpServletRequest` anywhere with this static call `ResteasyProviderFactory.getContextData(HttpServletRequest.class)`

For jersey, it's not `HttpServletRequest` but we seem to have a similar static call `ContextHandler.getCurrentContext().getContextPath()`.

So is it possible for us to use these to solve the problem at the bottom layer? I guess it means introducing e.g., `ResteasyAnnotationsTracer extends JaxRsAnnotationsTracer` for the implementation-specific extraction but doesn't seem too bad.

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/RouteOnSuccessOrError.java
anuraaga: It looks like this will work on all webflux, as suspicious as it may appear

`((RequestPath) serverRequest.pathContainer()).contextPath()`

trask:
Added note to description:

> Currently missing tests for non-root context path in
> * `jaxrs-1.0`
> * `webflux`

The tests may take some work, so would prefer to create separate issue to track.

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
trask: Oops! I fixed the JAX-RS 1.0 instrumentation. Still no test for JAX-RS 1.0 + non-root context path though, I added comment about this below.

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/RouteOnSuccessOrError.java
trask: updated

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
trask: I added comment below about missing tests for jaxrs-1.0 and webflux

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
trask: @anuraaga It looks like `ContextHandler.getCurrentContext().getContextPath()` is for jetty not jersey (well, would work for jersey running in jetty).

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -23,6 +24,22 @@
 
   private static final Logger log = LoggerFactory.getLogger(ServletHttpServerTracer.class);
 
+  @Override
+  public Scope startScope(Span span, HttpServletRequest request) {
```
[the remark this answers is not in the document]
trask: this is a really good question :monocle_face:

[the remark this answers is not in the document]
trask: fixed 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
trask: > Is this whole problem specific to servlets? Or for other HTTP Server spans as well? E.g. Ratpack or Ktor or I-dont-know-what-else?

It's specific to any routing framework that runs inside of a servlet container. I think I was trying to generalize it too much to "application root", but better to keep it specific to "servlet context path".

trask:
I'm not sure we can always get the servlet request statically (e.g. jersey).

If we can't get the servlet context path everywhere, then I think probably better to use the `ServletContextPath` context key everywhere for consistency (since at that point we will need to propagate it anyways). The latest commit re-consolidates to this (now better formed) pattern.

Let me know your thoughts.

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringWebMvcTracer.java
trask: the latest commit consolidates everything to single approach

instrumentation/jaxrs/jaxrs-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java
trask: I think this is addressed now, let me know what u think

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServletContextPath.java
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.servlet;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+
+public class ServletContextPath {
```
anuraaga, line 11: ServletContextContextUtils? Just kidding :)

anuraaga: Let's add some class javadoc to explain what's going on, including why this needs to go in instrumentation-api

anuraaga (approved):
Had no idea I was using jetty, not jersey. Servlet's very complicated 😅

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServletContextPath.java
```diff
@@ -0,0 +1,28 @@
...
+
+  // Keeps track of the servlet context path that needs to be prepended to the route when updating
+  // the span name
+  public static final ContextKey<String> CONTEXT_KEY =
+      ContextKey.named("opentelemetry-servlet-context-path-key");
+
+  /** Returns the servlet context path from the given context or <code>""</code> if not found. */
```
iNikem, line 18: This javadoc is obsolete

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/RouteOnSuccessOrError.java
```diff
@@ -69,4 +72,13 @@ private String parseRoute(String routerString) {
                 .trim())
         .replaceAll("");
   }
+
+  private static String getContextPath(ServerRequest serverRequest) {
```
iNikem, line 76: Is this method used?

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -107,4 +125,13 @@ protected String flavor(HttpServletRequest connection, HttpServletRequest reques
   protected String requestHeader(HttpServletRequest httpServletRequest, String name) {
     return httpServletRequest.getHeader(name);
   }
+
+  private static String getSpanName(HttpServletRequest request) {
+    String spanName = request.getServletPath();
+    String contextPath = request.getContextPath();
+    if (contextPath != null && !contextPath.isEmpty() && !contextPath.equals("/")) {
+      spanName = contextPath + spanName;
```
iNikem, line 133: Technically, this is a duplication with `ServletContextPath.prepend`

iNikem (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/RouteOnSuccessOrError.java
trask: nope, gone 👍

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
trask: ya, I could add `public static String prepend(String contextPath, String spanName)` to `ServletContextPath`, but I'm thinking better to live with the duplication instead of adding API surface to an already confusing class `ServletContextPath`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServletContextPath.java
trask: 👍 will do tomorrow before merging

Issue #1376, opened 2020-10-12
Mongo db.name attribute is incorrect


trask:
Currently Mongo `db.name` is being captured as the `description` of the server, as opposed to the database name that the query is executed against.

imavroukakis:
I can take this on if you'd like.

trask:
that would be awesome 👍

imavroukakis:
I assume that there are no tests that catch this, am I correct? Looking at the existing ones, it looks like `db.name` is set to the correct value

![image](https://user-images.githubusercontent.com/1764981/96367309-fae9b280-1144-11eb-8c4e-32b8ee9535ed.png)

imavroukakis:
Ah whoops, I see it
![image](https://user-images.githubusercontent.com/1764981/96368613-929ecf00-114c-11eb-88bc-1746cb3a1fbc.png)

PR #1419, opened 2020-10-19
capture mongodb database name as mongo-db.name attribute instead of client description

imavroukakis:
Closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1376

### DIFF
--- a/instrumentation/mongo/mongo-3.1/src/test/groovy/MongoClientTest.groovy
+++ b/instrumentation/mongo/mongo-3.1/src/test/groovy/MongoClientTest.groovy
@@ -9,9 +9,6 @@ import com.mongodb.MongoTimeoutException
 import com.mongodb.ServerAddress
 import com.mongodb.client.MongoCollection
 import com.mongodb.client.MongoDatabase
-import io.opentelemetry.instrumentation.test.asserts.TraceAssert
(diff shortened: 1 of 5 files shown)


imavroukakis:
It should probably be discussed what is to be done with the client description and a sensible fallback when a dbName doesn't exist. 

Client Description: Should it be added to `SemanticAttributes`?
Empty/Null database name: Client description might not be a sensible fallback, is returning `null` acceptable?

iNikem (approved):
If `db.name` is not available, then it is totally Ok to return `null`. What else can we do? I even think we can remove the current fallback to description.

imavroukakis:
> If `db.name` is not available, then it is totally Ok to return `null`. What else can we do? I even think we can remove the current fallback to description.

Makes sense. Should client description also be captured if available? That would be outside the scope of this PR, but would be good to get an opinion.

iNikem:
> Should client description also be captured if available?

If some MongoDB user thinks that this information is useful, then yes. _After_ the corresponding PR to the spec :)

mateuszrzeszutek (approved):

trask (approved):
Thanks @imavroukakis!

PR #1425, opened 2020-10-20
Fix span name populated by JAX-RS instrumentation

trask:
Fixes two issues:

* removes `httpMethod` prefix from span name
* fixes double slash when classPath.value() is empty string

### DIFF
--- a/instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsAnnotationsInstrumentationTest.groovy
+++ b/instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsAnnotationsInstrumentationTest.groovy
@@ -16,6 +16,7 @@ import javax.ws.rs.OPTIONS
 import javax.ws.rs.POST
 import javax.ws.rs.PUT
 import javax.ws.rs.Path
+import spock.lang.Unroll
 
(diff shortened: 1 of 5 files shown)


instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsAnnotationsInstrumentationTest.groovy
```diff
@@ -32,14 +33,15 @@ abstract class JaxRsAnnotationsInstrumentationTest extends AgentTestRunner {
...
-          name "POST /a"
+          name "/a"
           attributes {
           }
         }
       }
     }
   }
 
+  @Unroll
```
iNikem, line 44: I think we have global unroll, so this is not needed

iNikem (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-testing/src/main/groovy/JaxRsAnnotationsInstrumentationTest.groovy
mateuszrzeszutek: AFAIR only HTTP client & server base test classes have `@Unroll` on them, every test that does not extend one of them still needs to use `@Unroll`

mateuszrzeszutek (approved):

PR #1384, opened 2020-10-14
Update to latest SDK snapshot

anuraaga:
WIP for after the next release. gRPC library instrumentation went swell, still debugging an issue with auto instrumentation of all libraries, some tests fail, others give a `VerifyError`

PR #1426, opened 2020-10-20
Remove usage of Contexts from grpc instrumentation.

anuraaga:
And add a smoke test.

I originally was just trying to add a smoke test and found our agent is broken :) `Contexts.interceptCall` seems to be hard for us to use in instrumentation since `Contexts` needs to be unshaded, but `Context` is shaded. This problem will go away in #1384 when we don't use gRPC context.

As a quick fix, I have restored manual propagation of span - we should be able to cherrypick this into the release branch if we want to fix grpc support in the agent. Note, I am propagating span, not context, which we don't like to do, but I think clobbering user context is worse than losing baggage or whatever may be lost - and it's just temporary.


anuraaga:
I wasn't able to get the smoke test to pick up a new version of the javaagent without disabling gradle cache to rebuild it - not sure but a bit worried the build failure is the same cache issue.

.github/workflows/build-grpc-smoke-dist.yaml
```diff
@@ -0,0 +1,33 @@
+name: Build Spring Boot smoke test distribution
```
iNikem, line 1: That's not Spring Boot, right?

.gitignore
```diff
@@ -58,7 +57,4 @@ hs_err_pid*
 replay_pid*
 
 !java-agent/benchmark/releases/*.jar
-/smoke-tests/play/.gradle/
```
iNikem, line 61: Why?

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/GrpcSmokeTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+    [currentAgentVersion] as Set == findResourceAttribute(traces, "telemetry.auto.version")
+      .map { it.stringValue }
+      .collect(toSet())
+
+    cleanup:
+    stopTarget()
+
+    where:
+    jdk << [11]
+    // jdk << [8, 11, 15]
```
iNikem, line 52: Wanna change back?

smoke-tests/grpc/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+  implementation "io.grpc:grpc-stub:1.32.1"
+  implementation "io.opentelemetry:opentelemetry-proto:0.9.1"
+  implementation "io.opentelemetry:opentelemetry-extension-auto-annotations:0.9.1"
+  implementation "org.apache.logging.log4j:log4j-core"
+
+  runtimeOnly("org.apache.logging.log4j:log4j-slf4j-impl")
+}
+
+compileJava {
+  options.release = 11
```
mateuszrzeszutek, line 27: Shouldn't it be 8 here? If you want it to run on Java 8

.gitignore
anuraaga: I made .gradle a non-absolute path above which seems like low chance of collision :)

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/GrpcSmokeTest.groovy
anuraaga: Doh yeah

trask (approved):
🎉 new smoke test 🎉

Issue #1407, opened 2020-10-16
Cassandra instrumentations should store normalised CQL queries as db.statement


mateuszrzeszutek:
Currently both Cassandra 3.0 and 4.0 instrumentations do not normalise/mask the CQL query in any way -- it's possible that sensitive data may be stored in `db.statement` span attribute.

The aim of this issue is to fix that and normalise the query before it's added to span. This can be achieved by reusing the `SqlNormalizer` class - CQL has very similar syntax and I believe that the generic SQL normaliser would work perfectly fine without any need for modifications or improvements. We could move the normaliser class to a common module (`javaagent-api`?)and reuse it both for SQL and CQL (and possibly other SQL-like languages in the future).

PR #1427, opened 2020-10-20
Cassandra instrumentations should store normalised CQL queries as db.statement

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1407

* Move `DbSystem` to package `...instrumentation.api.db`
* Move `SqlNormalizer` to `javaagent-api`
* Refactor Cassandra 3.0 tests so that they use testcontainers (and run on Java 11)
* Implement Cassandra statement normalization


instrumentation/cassandra/cassandra-3.0/src/test/groovy/CassandraClientTest.groovy
```diff
@@ -13,39 +13,45 @@ import io.opentelemetry.instrumentation.test.AgentTestRunner
...
-    /*
-    Looks like sometimes our requests fail because Cassandra takes to long to respond,
-    Increase this timeout as well to try to cope with this.
-     */
-    cluster.getConfiguration().getSocketOptions().setReadTimeoutMillis(120000)
+    cassandra = new GenericContainer("cassandra:3")
+      .withExposedPorts(9042)
+      .withLogConsumer(new Slf4jLogConsumer(log))
+      .withStartupTimeout(Duration.ofSeconds(120))
+    cassandra.start()
```
mateuszrzeszutek, line 43: Should I change Cassandra 4.0 tests to use testcontainers as well? WDYT?

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/db/cassandra/CassandraQueryNormalizer.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.api.db.cassandra;
+
+import io.opentelemetry.javaagent.instrumentation.api.db.sql.normalizer.ParseException;
+import io.opentelemetry.javaagent.instrumentation.api.db.sql.normalizer.SqlNormalizer;
```
iNikem, line 9: I don't like this new package structure. How many classes in total are there under `io.opentelemetry.javaagent.instrumentation.api.db` package? Do we really need to split them even more into several subpackages?

instrumentation/cassandra/cassandra-3.0/src/test/groovy/CassandraClientTest.groovy
iNikem: Yes, please. Although in the future I personally would prefer to have these logically separate changes in separate PRs :)

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/db/cassandra/CassandraQueryNormalizer.java
mateuszrzeszutek: Honestly I don't really like it as well, it's too long and there's too much nesting. How about `...api.db.normalizer`? The auto-generated code probably needs its own package, because that package needs to be excluded from spotbugs/checkstyle.

instrumentation/cassandra/cassandra-3.0/src/test/groovy/CassandraClientTest.groovy
mateuszrzeszutek: I agree, I probably should have split this PR... but once I started working on this there were so many things that needed to be fixed/improved that I went with the flow and put them all here 😅

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/db/cassandra/CassandraQueryNormalizer.java
iNikem: Makes sense

mateuszrzeszutek: Done

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/db/CassandraQueryNormalizer.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package io.opentelemetry.javaagent.instrumentation.api.db;
+
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.ParseException;
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.SqlNormalizer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class CassandraQueryNormalizer {
```
trask, line 13: this feels cassandra-specific enough, and small enough, to not include in the api

javaagent-api/src/main/javacc/SqlNormalizer.jj
```diff
@@ -22,7 +11,7 @@ options {
 
 PARSER_BEGIN(SqlNormalizer)
 
-package io.opentelemetry.javaagent.instrumentation.jdbc.normalizer;
+package io.opentelemetry.javaagent.instrumentation.api.db.normalizer;
```
trask, line 14: how about `io.opentelemetry.javaagent.instrumentation.api.normalizer`?

i'm hoping `io.opentelemetry.javaagent.instrumentation.api.db.DbSystem` may go away https://github.com/open-telemetry/opentelemetry-java/issues/1841, which would then leave `.db` empty

also, maybe it should go in `instrumentation-api`? (we can also decide this later)

trask (commented):
❤️ not capturing sensitive data

javaagent-api/src/main/javacc/SqlNormalizer.jj
iNikem: `api.normalizer` is strange. We don't normalize API, do we? I think we need some sort of qualification here.

iNikem (approved):

javaagent-api/src/main/javacc/SqlNormalizer.jj
mateuszrzeszutek: `api.querynormalizer`?

> also, maybe it should go in instrumentation-api? (we can also decide this later)

Hmm, there's no need for this class to be accessible in manual instrumentations. Also, shouldn't we treat `instrumentation-api` kind of like a public API?

iNikem: There is no need, but there can be a "want". Manual instrumentations may want to have access to a standard normalization logic.

iNikem:
@mateuszrzeszutek you want to exclude `CassandraClientTest` from running in CircleCI as it does not allow running docker containers. E.g. by adding `@Requires({ "true" != System.getenv("CIRCLECI") })` to test class. See `SpringTemplateJMS1Test`

javaagent-api/src/main/javacc/SqlNormalizer.jj
trask: How about `api.normalize`? or `api.normalization`?

trask (approved):

mateuszrzeszutek:
(Rebased to make CircleCI run again)

javaagent-api/src/main/javacc/SqlNormalizer.jj
mateuszrzeszutek: Hmm, that still sounds kind of like we're normalizing API.

How about leaving it like it is for now? We can come back to this issue when `DbSystem` gets deleted

Issue #1260, opened 2020-09-25
Should instrumentation version for instrumentations be changed to something more descriptive?


jamesylgan:
What is the motivation behind including the instrumentation version [here](
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8ed4ef55c87c8259434dac15531030a2f32667f8/instrumentation/apache-httpclient/apache-httpclient-4.0/src/main/java/io/opentelemetry/instrumentation/auto/apachehttpclient/v4_0/ApacheHttpClientTracer.java#L78)?

Perhaps naively, I feel that providing the version the instrumentation was created for has less value than it could: users will observe the instrumentation version, e.g. `4.0.0` showing up for any versions 4.0.0+, and may mistake that for the library version being impacted by the instrumentation.

Alternatively, would it be better to store instrumentation versions as `4.0.0+` explicitly, so users can know by default that the instrumentation is for version `4.0.0` or above? If there's some reason for it to be `4.0.0` I'd love to read up on how the decision was made. Thanks!

trask:
Yeah, this does seem like it will cause confusion.

There isn't too much reasoning behind the current naming other than it matches the instrumentation module name (in this case `apache-httpclient-4.0`), so people can tell which instrumentation module produced the span.

jamesylgan:
That's sort of what I figured - would definitely propose changing it (pretty simply) from `<instrumentation_version>` to `<instrumentation_version>+`. I think that would help clarify things

iNikem:
E.g. we have instrumentation `jedis-1.4` and `jedis-3.0`. Where effectively first one is for versions 1.4-3.0. Having `1.4+` for it still may be confusing, because it seems to be unbounded.

I think it is better to remove those versions from `InstrumentationName` completely.

jamesylgan:
That may break some things @iNikem :)

It could be split into `InstrumentationVersion` and `InstrumentationName` though

trask:
> That may break some things @iNikem :)

Curious what you are thinking of that may break?

> It could be split into `InstrumentationVersion` and `InstrumentationName` though

I think `InstrumentationVersion` is supposed to be the version of the instrumentation, e.g. `0.8.0`, and get bumped with each new release of the instrumentation.

jamesylgan:
[...]

PR #1432, opened 2020-10-20
Remove version from `apache-httpclient` instrumentation

vadeg:
Remove version from `apache-httpclient` instrumentation

Fixes: #1260

### DIFF
--- a/instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientTracer.java
+++ b/instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientTracer.java
@@ -64,6 +64,6 @@ public class AkkaHttpClientTracer
 
   @Override
   protected String getInstrumentationName() {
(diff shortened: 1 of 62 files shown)


trask:
Hi @vadeg, thanks for the PR!

Are you able to sign the CLA (above)?

And could you make this change for all instrumentation in this PR, so we don't end up with some instrumentation names with the version and some without the version?

vadeg:
Hi @trask ,

Definitely. I have one question. There is [SpringWebMvcServerTracer](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcServerTracer.java#L76-L77) in the `spring-webmvc-3.1` project and it has no name:
```
  @Override
    return null;
  }
```
Is it intentional?

trask:
That does not look intentional.

I realized I have one doubt about the instrumentation names (I opened #1436 to discuss), but I think it's good to go ahead and remove the versions in this PR 👍.

vadeg:
I will reopen PR. I can not accept CLA. The links redirects me to the blank page.

PR #1439, opened 2020-10-21
Fix HttpURLConnection instrumentation on WebLogic

agoallikmaa:
In WebLogic, `URL.openConnection()` returns its own internal implementation of `HttpURLConnection`, which does not delegate the methods that have to be instrumented to the JDK superclass.

Instrumenting it directly resolves the issue of outgoing requests made with `URL.openConnection()` not being detected. It has been manually tested with WLS 12.2.1.4 and 14.1.1.0.


iNikem (approved):

trask (approved):
Thanks @agoallikmaa!

PR #1440, opened 2020-10-21
Prefer mavenCentral over jcenter

iNikem:
The latter was too unstable lately :(


anuraaga (approved):
Heh I think I see this change around once a year to various repos. Never saw it reverted.

mateuszrzeszutek (approved):

Issue #1406, opened 2020-10-16
Log normalised full statement in Redis instrumentations


mateuszrzeszutek:
To be done after https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1405 gets merged

Currently Lettuce 5.1 is the only instrumentation that logs the full `db.statement`, all others are just logging the command. [DB semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#call-level-attributes) says that we should log full statement if possible:

> For Redis, the value provided for db.statement SHOULD correspond to the syntax of the Redis CLI. If, for example, the HMSET command is invoked, "HMSET myhash field1 'Hello' field2 'World'" would be a suitable value for db.statement.

The aim of this issue is to straighten this out and use `RedisCommandNormalizer` in all Redis instrumentations (where it's possible to do so).

PR #1441, opened 2020-10-21
Log normalised full statement in Redis instrumentations

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1406

Implemented for: Lettuce 5.0, Jedis and Redisson.
Lettuce 4.0 was skipped because args are not accessible, they're written straight into the command's byte array.
Rediscala was skipped because it has a separate class for every command, and every class has separate fields for the command's args. It pretty much requires writing a separate normalizer.


instrumentation/jedis/jedis-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisInstrumentation.java
```diff
@@ -59,14 +63,10 @@ public JedisInstrumentation() {
     public static void onEnter(
         @Advice.This Connection connection,
         @Advice.Argument(0) ProtocolCommand command,
+        @Advice.Argument(1) byte[][] args,
         @Advice.Local("otelSpan") Span span,
         @Advice.Local("otelScope") Scope scope) {
-      int callDepth = CallDepthThreadLocalMap.incrementCallDepth(Connection.class);
```
iNikem, line 64: Why this is removed?

instrumentation/jedis/jedis-3.0/src/test/groovy/Jedis30ClientTest.groovy
```diff
@@ -76,29 +76,29 @@ class Jedis30ClientTest extends AgentTestRunner {
...
+            "$SemanticAttributes.NET_PEER_IP.key" "127.0.0.1"
+            "$SemanticAttributes.NET_PEER_NAME.key" "localhost"
+            "$SemanticAttributes.NET_PEER_PORT.key" port
           }
         }
       }
       trace(1, 1) {
         span(0) {
-          name "GET"
+          name "GET foo"
```
iNikem, line 93: I am not sure about this. `foo` here is cache key, right? Thus its cardinality is high. Thousands and millions potentially. Semantic convention says that span name should be low-cardinality value. `select * from table_name` is Ok span name, but here in Redis having cache key as part of the name - I am not sure.

instrumentation/redisson-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonClientTracer.java
```diff
@@ -24,24 +28,33 @@ protected String getInstrumentationName() {
...
     }
     return "Redis Command";
   }
 
+  private static String normalizeSingleCommand(CommandData<?, ?> command) {
+    List<String> args = new ArrayList<>();
+    if (command.getCommand().getSubName() != null) {
+      args.add(command.getCommand().getSubName());
+    }
+    args.addAll(Stream.of(command.getParams()).map(String::valueOf).collect(Collectors.toList()));
```
iNikem, line 54: Can `args` be `List<Object>`? This would eliminate this stream and thus extra allocations.

instrumentation/jedis/jedis-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisInstrumentation.java
mateuszrzeszutek: It's no longer needed - I changed the method matcher so that exactly one method matches; previously there were 2 or 3 methods that got matched.

instrumentation/jedis/jedis-3.0/src/test/groovy/Jedis30ClientTest.groovy
mateuszrzeszutek: Hmm, that's true. Lettuce 5.1 uses just the command name as span name, how about making all other Redis instrumentations do the same? I think it matches the spec:

> The span name SHOULD be set to a low cardinality value representing the statement executed on the database. It may be a stored procedure name (without arguments), SQL statement without variable arguments, operation name, etc.

iNikem: Yes, just command name seems good.

mateuszrzeszutek: Done.

instrumentation/redisson-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonClientTracer.java
mateuszrzeszutek: Done.

iNikem (approved):

iNikem:
@open-telemetry/java-instrumentation-approvers this PR could use another reviewer, not from Splunk

instrumentation/jedis/jedis-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v1_4/JedisInstrumentation.java
```diff
@@ -47,34 +50,81 @@ public JedisInstrumentation() {
...
+        @Advice.This Connection connection,
+        @Advice.Argument(0) Command command,
+        @Advice.Local("otelSpan") Span span,
+        @Advice.Local("otelScope") Scope scope) {
+      int callDepth = CallDepthThreadLocalMap.incrementCallDepth(Connection.class);
+      if (callDepth > 0) {
+        return;
+      }
+
+      span = TRACER.startSpan(connection, new CommandWithArgs(command, new byte[0][]));
```
trask, line 90: can we extract constant for `new byte[0][]` to avoid allocating each time?

trask (approved):
Nice 👍

anuraaga (approved):

PR #1442, opened 2020-10-21
Proper scope handling for Resteasy instrumentations.

### LINKED ISSUE #1422: Resteasy client leads to duplicate spanId injections into http headers
I have a bug report from the client who sees multiple spanIds injected into outgoing http requests:
```
x-b3-spanid : [7700a7830ef64c36, 7a2b99e5202c3214]
x-b3-traceid :[00000000000000006d776b6577656321,00000000000000006d776b6577656321]
x-b3-sampled : [1, 1]
```

They use Resteasy-jaxrs 3.1.4.Final. By default Resteasy uses Apache Http Client (AHC) under the hood. When I add our `apache-httpclient-4.0` instrumentation as test dependency to `jaxrs-client-2.0`, all Resteasy tests started to fail:

```
17:34:48.058 [Test worker] DEBUG io.opentelemetry.instrumentation.test.InMemoryExporter - >>>1603118088058001959 SPAN START: HTTP GET id=b80393f07d812347 traceid=f88b3a57c349e7ed43385ab3f1383d0f parent=0000000000000000, library=InstrumentationLibraryInfo{name=io.opentelemetry.auto.jaxrs-client-2.0, version=null}
17:34:48.059 [Test worker] DEBUG io.opentelemetry.instrumentation.test.InMemoryExporter - >>>1603118088059001301 SPAN START: HTTP GET id=5ceedbb271029c7a traceid=3eb4fd91c8698af8048ba8b8ed711355 parent=0000000000000000, library=InstrumentationLibraryInfo{name=io.opentelemetry.auto.apache-httpclient-4.0, version=null}
```

The reason for this is two-fold. First, AHC instrumentation does not check if `CLIENT` span already exists and it should not create a new one. Second, even if AHC did that, it would not help, because jaxrs-client instrumentaiton does NOT install itself as the active span for the duration of the request. See `io.opentelemetry.javaagent.instrumentation.jaxrsclient.v2_0.ClientTracingFilter#filter(javax.ws.rs.client.ClientRequestContext)`

iNikem:
Fixes #1422. This PR solves the original problem with Resteasy. If other JAX-RS client implementations have the same problem, remains to be seen.

### DIFF
--- a/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-resteasy-2.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ResteasyClientConnectionErrorInstrumentation.java
+++ b/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-resteasy-2.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ResteasyClientConnectionErrorInstrumentation.java
@@ -25,11 +25,11 @@ import net.bytebuddy.matcher.ElementMatcher;
 import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
 
 /**
- * Unlike other supported JAX-RS Client implementations, Resteasy's one is very simple and
- * passes all requests through single point. Both sync ADN async! This allows for easy
- * instrumentation and proper scope handling.
+ * Unlike other supported JAX-RS Client implementations, Resteasy's one is very simple and passes
+ * all requests through single point. Both sync ADN async! This allows for easy instrumentation and
+ * proper scope handling.
  *
- * This specific instrumentation will not conflict with {@link JaxRsClientInstrumentation},
+ * <p>This specific instrumentation will not conflict with {@link JaxRsClientInstrumentation},
  * because {@link JaxRsClientTracer} used by the latter checks against double client spans.
  */
 @AutoService(Instrumenter.class)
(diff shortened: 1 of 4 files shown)


instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-resteasy-2.0/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ResteasyClientConnectionErrorInstrumentation.java
```diff
@@ -5,30 +5,32 @@
...
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
-import org.jboss.resteasy.client.jaxrs.internal.ClientConfiguration;
+import org.jboss.resteasy.client.jaxrs.internal.ClientInvocation;
 
 /**
- * JAX-RS Client API doesn't define a good point where we can handle connection failures, so we must
- * handle these errors at the implementation level.
+ * Unlike other supported JAX-RS Client implementations, Resteasy's one is very simple and passes
+ * all requests through single point. Both sync ADN async! This allows for easy instrumentation and
```
trask, line 29:
```suggestion
 * all requests through single point. Both sync AND async! This allows for easy instrumentation and
```

trask (approved):
Nice!

anuraaga (approved):

PR #1446, opened 2020-10-21
Remove apache httpclient version

### LINKED ISSUE #1260: Should instrumentation version for instrumentations be changed to something more descriptive?

vadeg:
This is reponed PR #1432. There was problems with CLA.

What is in this PR?

1. Versions are removed from instrumentation names
2. Add instrumentation name to `spring-web` (was `null` before)

Fixes #1260

### DIFF
--- a/instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
+++ b/instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
@@ -24,7 +24,6 @@ import org.bson.BsonValue;
 public class MongoClientTracer extends DatabaseClientTracer<CommandStartedEvent, BsonDocument> {
   public static final MongoClientTracer TRACER = new MongoClientTracer();
 
-  // TODO use tracer names *.mongo-3.1, *.mongo-3.7, *.mongo-async-3.3 respectively in each module
   @Override
   protected String getInstrumentationName() {
     return "io.opentelemetry.auto.mongo";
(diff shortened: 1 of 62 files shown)


instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -24,7 +24,6 @@
 public class MongoClientTracer extends DatabaseClientTracer<CommandStartedEvent, BsonDocument> {
   public static final MongoClientTracer TRACER = new MongoClientTracer();
 
-  // TODO use tracer names *.mongo-3.1, *.mongo-3.7, *.mongo-async-3.3 respectively in each module
```
trask, line 27: 👍

trask (approved):
Thanks @vadeg!

anuraaga (approved):
Thanks!

Issue #1328, opened 2020-10-05
Change logging exporter from System.out to SLF4J


trask:
Tracking proposal from @anuraaga https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1313#issuecomment-703035396

Makes sense to me, logging at INFO level, which we enable by default.

Anisha-28:
Hi .. can. i work on this issue 
Would appreciate some clarification please

iNikem:
@Anisha-28  There is [LoggingExporter](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java) which just prints all created spans. Right now it is using `System.out.println`. The goal of this is to replace `System.out` with `org.slf4j.Logger` logger, as many other classes in this repo.

iNikem:
If you wish to take this issue, I can assign it to you.

Anisha-28:
yes @iNikem i will do it .. Thanks for explaining though

Anisha-28:
@iNikem i have committed a change if you can see it .. and let me know if i did it right

nitram509:
@Anisha-28 once you have committed, don't forget to push it too and create a pull request.
(if you just did commit, no one can see it)

Cheers 
Martin

PR #1447, opened 2020-10-21
Replace `print` with log messages.

vadeg:
Replace `print` with log messages.

* Add `slf4j` compile time dependency to the `javaagent-exporters.logging` module
* Replace `System.out.print` with `log.debug` messages

Concerns:
1. I am not sure about `log.debug`. I can replace it if you think that it should be in other level.
2. `AttributeKey` ls logged using `toString()` method from its implementation. The only implementation right now is `AutoValue_AttributeKeyImpl` which prints the key like `"AttributeKeyImpl{getType=" + this.getType + ", key=" + this.key + "}"`. I think it is not a good idea to use `toString` from objects for logging because anyone can change the method's implementation or provide implementation without `toString` (with default) and it will affect logging quality. I would change it not rely on `toString` output, but I don't know what would the appropriate format for this. If you have ideas please share.

Fixes: #1328

### DIFF
--- a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
+++ b/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
@@ -44,7 +44,7 @@ public class LoggingExporter implements SpanExporter {
                 @Override
                 public <T> void consume(AttributeKey<T> key, T value) {
 
-                  stringBuilder.append(key).append('=');
+                  stringBuilder.append(key.getKey()).append('=');
 
                   if (key.getType() == AttributeType.STRING) {
                     stringBuilder.append('"').append(value).append('"');
--- a/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
+++ b/javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
@@ -55,7 +55,7 @@ public class LoggingExporter implements SpanExporter {
                 }
               });
     }
-    log.debug(stringBuilder.toString());
+    log.info(stringBuilder.toString());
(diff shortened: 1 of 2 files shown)


javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
```diff
@@ -21,28 +25,37 @@ public LoggingExporter(String prefix) {
...
-                  System.out.print(key + "=");
-                  switch (key.getType()) {
-                    case STRING:
-                      System.out.print('"' + String.valueOf(value) + '"');
-                      break;
-                    default:
-                      System.out.print(value);
-                      break;
+
+                  stringBuilder.append(key).append('=');
```
anuraaga, line 47:
```suggestion
                  stringBuilder.append(key.getKey()).append('=');
```

Thanks for bringing up the issue about toString, I think getKey should solve that

javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
```diff
@@ -21,28 +25,37 @@ public LoggingExporter(String prefix) {
...
+                  } else {
+                    stringBuilder.append(value);
                   }
-                  System.out.print(" ");
+                  stringBuilder.append(' ');
                 }
               });
     }
-    System.out.println();
+    log.debug(stringBuilder.toString());
```
anuraaga, line 58:
```suggestion
    log.info(stringBuilder.toString());
```

This exporter needs to be explicitly enabled so I think by default it should log something :)

anuraaga (commented):
Thanks for the help!

iNikem (approved):
What Anuraag said

vadeg:
Updated with suggestions from @anuraaga

trask (approved):
👍

PR #1450, opened 2020-10-22
Copy in release build workflow from opentelemetry-java

anuraaga:
Almost the same as opentelemetry-java except without README update steps.

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -49,7 +49,6 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: ${{ needs.prepare-release-branch.outputs.release-branch-name }}
-          submodules: true
       - uses: actions/setup-java@v1
         with:
           java-version: 11
@@ -70,8 +69,6 @@ jobs:
           job-id: jdk11
           remote-build-cache-proxy-enabled: false
           arguments: build --stacktrace -Prelease.version=${{ github.event.inputs.version }}
-          properties: |
-            enable.docker.tests=true
       - name: Publish artifacts
         uses: burrunan/gradle-cache-action@v1.5
         with:
@@ -83,13 +80,3 @@ jobs:
           BINTRAY_KEY: ${{ secrets.BINTRAY_KEY }}
           GRGIT_USER: ${{ github.actor }}
           GRGIT_PASS: ${{ secrets.GITHUB_TOKEN }}
-      - run: git push
-      - name: Checkout master
-        uses: actions/checkout@v2
-        with:
-           submodules: true
-      - uses: burrunan/gradle-cache-action@v1.5
-        with:
-          job-id: jdk11
-          remote-build-cache-proxy-enabled: false
-          arguments: --stacktrace -Prelease.version=${{ github.event.inputs.version }}
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -1,0 +1,95 @@
+# Releases a patch by cherrypicking commits into a release branch based on the previous
+# release tag.
+name: Patch Release Build
+on:
+  workflow_dispatch:
+    inputs:
+      version:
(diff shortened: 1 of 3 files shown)


RELEASING.md
```diff
@@ -13,22 +13,69 @@ This signals Nebula plugin to build and publish to
...
-That build, after usual building and checking, will run `./gradlew -Prelease.scope=patch final`.
-This will signal Nebula plugin to build a new version `vX.Y.(Z+1)` and publish it to Bintray repo.
+## Starting the Release
+
+Open the release build workflow in your browser [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions?query=workflow%3A%22Release+Build%22).
+
+You will see a button that says "Run workflow". Press the button, enter the version number you want
+to release in the input field that pops up, and then press "Run workflow".
+
+This triggers the release process, which builds the artifacts, updates the README with the new
```
iNikem, line 23: We don't update README in this repo, right?

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,32 @@
...
+        required: true
+
+jobs:
+  build:
+    name: Build and release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
```
iNikem, line 17: This repo does not have submodules

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          job-id: jdk11
+          remote-build-cache-proxy-enabled: false
+          arguments: build --stacktrace -Prelease.version=${{ github.event.inputs.version }}
+      - name: Publish artifacts
+        run: ./gradlew final --stacktrace -Prelease.version=${{ github.event.inputs.version }}
+        env:
+          BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
+          BINTRAY_KEY: ${{ secrets.BINTRAY_KEY }}
+          GRGIT_USER: ${{ github.actor }}
```
iNikem, line 31: We should not need `GRGIT` or?

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,95 @@
...
+              # Trim whitespaces and cherrypick
+              echo $word | sed 's/ *$//g' | sed 's/^ *//g' | git cherry-pick --stdin
+          done
+      - uses: burrunan/gradle-cache-action@v1.5
+        with:
+          job-id: jdk11
+          remote-build-cache-proxy-enabled: false
+          arguments: build --stacktrace -Prelease.version=${{ github.event.inputs.version }}
+          properties: |
+            enable.docker.tests=true
```
iNikem, line 74: We don't need this, do we?

.github/workflows/patch-release-build.yml
```diff
@@ -0,0 +1,95 @@
...
+      - run: git push
+      - name: Checkout master
+        uses: actions/checkout@v2
+        with:
+           submodules: true
+      - uses: burrunan/gradle-cache-action@v1.5
+        with:
+          job-id: jdk11
+          remote-build-cache-proxy-enabled: false
+          arguments: --stacktrace -Prelease.version=${{ github.event.inputs.version }}
```
iNikem, line 95: What does this do?

.github/workflows/release-build.yml
anuraaga: nebula will create and push a tag so GRGIT credentials allows it to do that.

anuraaga (commented):
Thanks

.github/workflows/release-build.yml
iNikem: But `prepare-release-branch` job above creates and pushes branch too. And without GRGIT?

anuraaga: Oops forgot to respond - yeah normal `git` commands authenticate automatically. I first tried without GRGIT hoping it'd behave the same in Gradle but it doesn't :(

iNikem (approved):

Issue #1443, opened 2020-10-21
Apache Geode instrumentation should store normalized queries as db.statement


mateuszrzeszutek:
The aim of this issue is to fix that and normalise the Geode query before it's added to span. Right now full query is used, without any sanitization at all. Geode uses SQL-like language for querying, so we can use the good old `SqlNormalizer` here - same as in the [Cassandra normalization issue](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1407).

Issue #1431, opened 2020-10-20
Standardize DB query normalizer property names


mateuszrzeszutek:
Right now we only have one property for SQL (JDBC): `sql.normalizer.enabled`. How about renaming it to something like `otel.query.normalizer.<db>.enabled`, `otel.<db>.query.normalizer.enabled`? We could use separate config properties for each query normalizer that's used in the agent (enabled by default).

Regarding the `<db>` placeholder: should we use the `DbSystem` name or the query language name (sql, cql, n1ql, ...)? I think it makes more sense to disable per query language, but on the other hand it introduces a new concept/dimension (you have to know which instrumentation/`DbSystem` uses which language).

trask:
Does it work to conform to `otel.instrumentation.<instrumentationid>.*` proposed in #1414?

E.g. `otel.instrumentation.jdbc.query.normalizer.enabled`

mateuszrzeszutek:
Hmm, that sounds good. The only "problem" that I can think of is with Redis, which has multiple instrumentations - but I believe that if for some reason you want to disable sanitization then you probably know which Redis library you're using.

PR #1455, opened 2020-10-22
Apache Geode instrumentation should store normalized queries as db.statement

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1443

### DIFF
--- a/instrumentation/geode-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/geode/GeodeQueryNormalizer.java
+++ b/instrumentation/geode-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/geode/GeodeQueryNormalizer.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.geode;
+
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.ParseException;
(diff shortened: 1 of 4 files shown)


instrumentation/geode-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/geode/GeodeQueryNormalizer.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.ParseException;
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.SqlNormalizer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class GeodeQueryNormalizer {
+  private static final Logger log = LoggerFactory.getLogger(GeodeQueryNormalizer.class);
+
+  public static String normalize(String query) {
```
iNikem, line 16: If this is just a simple wrapper, why do we need it?

mateuszrzeszutek: Just for the try-catch right now. In the (close) future we'll also check if the normalization is enabled (I'll add appropriate config flags in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1431)

iNikem (approved):

trask (approved):
👍

Issue #1444, opened 2020-10-21
Couchbase instrumentation should store normalized queries as db.statement


mateuszrzeszutek:
The aim of this issue is to fix that and normalise the Couchbase N1QL query before it's added to span. Currently Couchbase instrumentation stores the result of `toString()` in `db.statement`: e.g. `SimpleN1qlQuery{statement=SELECT mockrow}`. Putting the total lack of sanitization aside, I don't think the query class name is really needed here.
N1QL is another SQL-like query language and `SqlNormalizer` should be able to mask the query.

PR #1462, opened 2020-10-23
Couchbase instrumentation should store normalized queries as db.statement

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1444

### DIFF
--- a/instrumentation/couchbase/couchbase-2.0/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_0/CouchbaseQueryNormalizer.java
+++ b/instrumentation/couchbase/couchbase-2.0/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_0/CouchbaseQueryNormalizer.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.couchbase.v2_0;
+
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.ParseException;
+import io.opentelemetry.javaagent.instrumentation.api.db.normalizer.SqlNormalizer;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class CouchbaseQueryNormalizer {
+  private static final Logger log = LoggerFactory.getLogger(CouchbaseQueryNormalizer.class);
+
+  private static final Class<?> QUERY_CLASS;
+  private static final Class<?> STATEMENT_CLASS;
+  private static final Class<?> N1QL_QUERY_CLASS;
+  private static final MethodHandle N1QL_GET_STATEMENT;
+  private static final Class<?> ANALYTICS_QUERY_CLASS;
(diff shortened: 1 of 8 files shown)


iNikem (approved):

instrumentation/couchbase/couchbase-2.0/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_0/CouchbaseQueryNormalizer.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+    ANALYTICS_QUERY_CLASS = analyticsQueryClass;
+    ANALYTICS_GET_STATEMENT = analyticsGetStatement;
+  }
+
+  public static String normalize(Object query) {
+    if (query instanceof String) {
+      return normalizeString((String) query);
+    }
+    // Couchbase 2.0 uses Query, Couchbase 2.5+ uses Statement
```
trask, line 80: does it make sense to use muzzle and split out separate instrumentation module for couchbase 2.5?

iNikem: Just a general remark: I think the fewer modules we have, the better :)

trask: I think it's a trade-off of how much complexity we need to add in order to handle multiple incompatible versions in a single module

spring-webflux is the only other module that uses MethodHandle, and in that case adds much less complexity compared to this

all other modules currently use muzzle to handle incompatible versions

mateuszrzeszutek: It turns out the situation is much more complicated: I was kinda lazy last time (😅 ) and did not check in which version which class appears, but I did it right now and it looks like we'd need 4 different instrumentations if we wanted to avoid using `MethodHandle`s at all: 2.0.0 - 2.1.0, 2.1.0 - 2.2.0, 2.2.0 - 2.4.3, 2.4.3 - latest.

trask: yikes that's a lot of version dependency!

if we keep it in a single module, how do we get test coverage of these different versions? maybe we need a way to specify multiple versions for `testLatestDeps`?

mateuszrzeszutek: Right now all tests use Couchbase 2.5.0 (common test code in `couchbase-testing` for both 2.0.0 and 2.6.0 instrumentations) and every query type except the 2.0.0-2.2.0 `Query` interface is covered. If we wanted to test that we'd probably need to refactor the testing module.

trask: thx for explanations, approach / testing coverage seems reasonable to me

trask (approved):
👍

PR #1466, opened 2020-10-23
Stop being parallel by default

iNikem:
One more attempt of #740

Current tests are too flaky, let us try to fix them this way

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,3 @@
-org.gradle.parallel=true
 org.gradle.caching=true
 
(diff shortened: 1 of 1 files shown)


iNikem:
Build is green, but it took more than an hour. I don't like it :(

trask:
Yeah, not great, but better to have long builds than flaky builds 🤷‍♂️

Should we do `--no-parallel` instead of modifying `gradle.properties` so dev local builds will still use parallel?

iNikem:
Let's see if #1467 helped. And I will devote some attention to build speed and caching etc

Issue #1458, opened 2020-10-22
Replace hardcoded strings with API constants


vadeg:
This is a follow up ticket after making #1375. There are a lot of hardcoded strings, e.g. `"net.peer.ip"`, `"net.transport"` used in different places. It would be nice to use constants instead of them.

## Tasks

- [ ] Replace all `net.*` constants (#1459)
- [ ] Replace all `enduser.*` constants (#1459)
- [ ] Replace all `http.*` constants 
- [ ] Replace all `peer.*` constants 
- [ ] Replace all `rpc.*` constants (#1459)
- [ ] Replace all `message.*` constants
- [ ] Replace all `db.*` constants
- [ ] Replace all `messaging.*` constants
- [ ] Replace all `exception.*` constants
- [ ] Replace all `faas.*` constants

iNikem:
Thank you for taking this! 🙏

PR #1459, opened 2020-10-22
Replace `net.*`, `rpc.*`, `enduser.*` with constants

vadeg:
Replace `net.*`, `rpc.*`, `enduser.*` with constants.

Fixes: #1458

PR #1468, opened 2020-10-23
Replace string values with constants.

vadeg:
This is a second PR after #1459 to replace string values with respective constants.

This PR covers the following groups: 
- `http.*`
- `peer.*`
- `message.*`
- `db.*`
- `messaging.*`
- `exception.*`
- `faas.*`

Fixes: #1458


trask (approved):
👍

iNikem:
If I am not mistaken, `typedSpans` package contains auto-generated files. Am I correct @thisthat ?

thisthat:
> If I am not mistaken, `typedSpans` package contains auto-generated files. Am I correct @thisthat ?

You are indeed correct! The code that generates those is available at https://gist.github.com/thisthat/7e34742f4a7f1b5df57118f859a19c3b

thisthat:
Would you like a PR (similar to https://github.com/open-telemetry/opentelemetry-java/pull/1846) that introduces a script to generate them from the YAML definitions?

trask:
thanks for the PR @vadeg! we ended up removing these files since we aren't using them currently #1696

PR #1471, opened 2020-10-23
Wrappers for AWS lambda tracing

### LINKED ISSUE #1469: Implement AWS lambda tracing wrapper
Currently we provide a nice way for AWS lambda users that would like to instrument their code via base abstract TracingRequestHandler class. In order to support also users that don't want to change their code, we could also add a wrapper that would get an env configuration property delegating to final lambda handler.

kubawach:
Resolves #1469 

Adds:
- request streams handler (sibling to already existing request handler)
- wrappers for handlers, allowing tracing without code modifications
- tests

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestStreamHandler;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.Tracer;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A base class similar to {@link RequestStreamHandler} but will automatically trace invocations of
+ * {@link #doHandleRequest(InputStream input, OutputStream output, Context)}.
+ */
+public abstract class TracingRequestStreamHandler implements RequestStreamHandler {
+
+  private final AwsLambdaTracer tracer;
+
+  /**
+   * Creates a new {@link TracingRequestStreamHandler} which traces using the default {@link
+   * Tracer}.
+   */
+  protected TracingRequestStreamHandler() {
+    this.tracer = new AwsLambdaTracer();
+  }
+
+  /**
+   * Creates a new {@link TracingRequestStreamHandler} which traces using the specified {@link
+   * Tracer}.
+   */
+  protected TracingRequestStreamHandler(Tracer tracer) {
+    this.tracer = new AwsLambdaTracer(tracer);
+  }
+
+  /**
+   * Creates a new {@link TracingRequestStreamHandler} which traces using the specified {@link
+   * AwsLambdaTracer}.
+   */
+  protected TracingRequestStreamHandler(AwsLambdaTracer tracer) {
+    this.tracer = tracer;
+  }
+
+  @Override
+  public final void handleRequest(InputStream input, OutputStream output, Context context)
+      throws IOException {
+    Span span = tracer.startSpan(context, Kind.SERVER);
+    Throwable error = null;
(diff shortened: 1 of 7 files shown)


kubawach:
@anuraaga you may want to have a look at this :)

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A base class similar to {@link RequestStreamHandler} but will automatically trace invocations of
+ * {@link #doHandleRequest(InputStream input, OutputStream output, Context)}.
+ */
+public abstract class TracingRequestStreamHandler implements RequestStreamHandler {
```
anuraaga, line 24: Thanks for handling the stream handlers :)

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
```diff
@@ -0,0 +1,74 @@
...
+    this.tracer = tracer;
+  }
+
+  @Override
+  public final void handleRequest(InputStream input, OutputStream output, Context context)
+      throws IOException {
+    Span span = tracer.startSpan(context, Kind.SERVER);
+    Throwable error = null;
+    try (Scope ignored = tracer.startScope(span)) {
+      doHandleRequest(input, output, context);
```
anuraaga, line 58: I think we need to wrap `OutputStream` so the span is closed on `OutputStream.close` which could be asynchronous?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamWrapper.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class TracingRequestStreamWrapper extends TracingRequestStreamHandler {
+
+  private WrappedLambda wrappedLambda;
+
+  @Override
+  protected void doHandleRequest(InputStream input, OutputStream output, Context context)
+      throws IOException {
+
+    if (wrappedLambda == null) {
```
anuraaga, line 28: I wonder if we should just make this static since the configuration is?

anuraaga (commented):
Thanks this is quite cool

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamWrapper.java
kubawach: Good idea, changed.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
kubawach: It's an option I considered. However now, with span issued on lambda exit, stream handler behaves exactly as "no stream" counterpart. In particular, the span gets send even if underlying output stream flush fails - I understand that in the case of RequestHandler implementations serialization / writing happens after the lambda exits. So with this implementation we have similar behaviour in both cases. What do you think?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
kubawach: :)

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
anuraaga: I don't know the details of lambda well but does it not allow something like

```
CompletableFuture<Foo> response = backend.getFoo();

response.thenAccept(writeToOutputStream);
```

If this is allowed than I think we need to make sure the span survives until the output is closed since that's the length of the request and how we generally model async requests.

kubawach: With this in mind, the idea totally makes sense. Will wrap OOS then :)

kubawach:
@anuraaga please have a second look :)

anuraaga (approved):
Thanks!

PR #1477, opened 2020-10-26
Remove unnecessary overrides

trask:


instrumentation/servlet/glassfish-testing/src/test/groovy/GlassFishServerTest.groovy
```diff
@@ -69,39 +63,11 @@ class GlassFishServerTest extends HttpServerTest<GlassFish> {
     true
   }
 
+  // Simple class name plus method name of the entry point of the given servlet container.
```
anuraaga, line 66: Hmm - actually I don't see this in the semantic conventions, I see examples like `/webshop/articles/:article_id`. Do we need to add a servlet semantic convention if we want to use class/method name for the server span name?

anuraaga: Though I mainly ask here just to make sure the method doesn't appear in the span name at some point :) But I guess either way it wouldn't.

anuraaga (approved):

instrumentation/servlet/glassfish-testing/src/test/groovy/GlassFishServerTest.groovy
trask: This comment was just moved, not new in this PR.

I'm glad you feel that way though, see latest commits in #1418 😄

PR #1482, opened 2020-10-26
Fix net.peer.* setting for Cassandra 4.+

### LINKED ISSUE #1460: Nightly build failed
Nightly job failed. Please take a look and fix it ASAP

### LINKED ISSUE #1461: Nightly build failed
Nightly job failed. Please take a look and fix it ASAP

### LINKED ISSUE #1472: Nightly build failed
Nightly job failed. Please take a look and fix it ASAP

### LINKED ISSUE #1473: Nightly build failed
Nightly job failed. Please take a look and fix it ASAP

### LINKED ISSUE #1478: Nightly build failed
Nightly job failed. Please take a look and fix it ASAP

iNikem:
#1427 removed max java version to run Cassandra 4 tests. This allowed `testLatestDep` job to actually run them (as it uses java 11 to run). This exposed that starting from some version of Cassandra client driver the value returned by `node.getBroadcastRpcAddress()` changed to `0.0.0.0` in our tests. This broke tests.

But this also exposed that we used wrong address for `net.peer.*` attributes in the first place. `BroadcastRpcAddress` means the address that Cassandra server accepts RPC connections. In case of containers/proxy this may be completely different address from the one used by client. But `node.getEndPoint()` returns "The information that the driver uses to connect to the node". Which seems like the correct one.

Closes #1478
Closes #1473 
Closes #1472 
Closes #1461 
Closes #1460

### DIFF
--- a/instrumentation/cassandra/cassandra-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v4_0/CassandraDatabaseClientTracer.java
+++ b/instrumentation/cassandra/cassandra-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v4_0/CassandraDatabaseClientTracer.java
@@ -14,7 +14,7 @@ import io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils;
 import io.opentelemetry.javaagent.instrumentation.api.db.DbSystem;
 import io.opentelemetry.trace.Span;
 import java.net.InetSocketAddress;
-import java.util.Optional;
+import java.net.SocketAddress;
 
 public class CassandraDatabaseClientTracer extends DatabaseClientTracer<CqlSession, String> {
   public static final CassandraDatabaseClientTracer TRACER = new CassandraDatabaseClientTracer();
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

instrumentation/cassandra/cassandra-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v4_0/CassandraDatabaseClientTracer.java
```diff
@@ -47,8 +47,10 @@ protected InetSocketAddress peerAddress(CqlSession cqlSession) {
   public void onResponse(Span span, ExecutionInfo executionInfo) {
     Node coordinator = executionInfo.getCoordinator();
     if (coordinator != null) {
-      Optional<InetSocketAddress> address = coordinator.getBroadcastRpcAddress();
-      address.ifPresent(inetSocketAddress -> NetPeerUtils.setNetPeer(span, inetSocketAddress));
+      SocketAddress socketAddress = coordinator.getEndPoint().resolve();
```
anuraaga, line 50: I don't think it's related to this PR since we have many instrumentations with the same pattern - I often worry about blocking an event loop with a DNS lookup just to set a span attribute. I want (someone) to remove the spec's language encouraging a DNS lookup just to resolve an attribute.

Just the word resolve makes it more obvious than Java where a seemingly normal getter has the chance to block and brought this point up for me :)

anuraaga (approved):

instrumentation/cassandra/cassandra-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v4_0/CassandraDatabaseClientTracer.java
iNikem: Good point. But this may be related to the general question of bringing performance considerations into semantic conventions...

trask: I'd be interested if we force DNS resolution anywhere, I know we've tried to avoid that in the past.

In this case, I'm not sure why the method is named `resolve()`, at least the implementation I'm seeing when I look doesn't seem to resolve anything.

PR #1489, opened 2020-10-27
Remove inner class imports; avoid future imports by disabling in editorconfig

bogdandrutu:
Signed-off-by: Bogdan Drutu <bogdandrutu@gmail.com>


trask (approved):
Thanks!

bogdandrutu:
@trask I also tested, by simply removing the import and press "option + enter" on MacOs which instead of importing the inner class (as it does without the change) was simply adding the "x.".

trask:
👍 even if we decide to go with `XBuilder`, this will still help with guava imports (and `Map.Entry`)

anuraaga (approved):
Nice, thanks!

bogdandrutu:
@trask that is the official jdk not guava that has Map.
Entry

PR #1493, opened 2020-10-27
Replace ContextStore for context bridge with a custom context storage…

### LINKED ISSUE #976: Need more complete instrumentation of Context
In particular, if you call `Context.withValues()` directly, and then pass that resulting `Context` back over the bridge to the agent, then you'll get:

```
java.lang.Exception: unexpected context
```

because we are currently expecting all `Context` instances to have been created via the bridge.

anuraaga:
… provider. And handles "span always operates on agent context" logic in the storage layer.

Fixes #976


anuraaga:
Came up with an approach for removing `TracingContextUtils` instrumentation. It's different - don't know if it's better 😅 . Instead of instrumenting the utils, it uses reflection to get the keys and then diverts to application or agent context based on the key at the storage layer. There's less bytecode manipulation (LoC of this change is -110 or so) and will be easily extensible to baggage. So I guess I like it a little, but don't love it. Let me know what your thoughts are.

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/ContextInstrumentation.java
```diff
@@ -6,65 +6,48 @@
...
+ * sure there is no dependency on a system property or possibility of a user overriding this since
+ * it's required for instrumentation in the agent to work properly.
+ */
 @AutoService(Instrumenter.class)
 public class ContextInstrumentation extends AbstractInstrumentation {
 
   @Override
   public ElementMatcher<? super TypeDescription> typeMatcher() {
-    return named("application.io.opentelemetry.context.Context");
+    return named("application.io.opentelemetry.context.DefaultContext");
```
anuraaga, line 35: We used to instrument the package-private class `TracingContextUtils`, now we do still end up with one instrumentation of `DefaultContext`

instrumentation/opentelemetry-api-beta/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/context/AgentContextStorage.java
```diff
@@ -0,0 +1,166 @@
...
+  static final io.opentelemetry.context.ContextKey<io.opentelemetry.trace.Span>
+      AGENT_SPAN_CONTEXT_KEY;
+  static final ContextKey<Span> APPLICATION_SPAN_CONTEXT_KEY;
+
+  static {
+    io.opentelemetry.context.ContextKey<io.opentelemetry.trace.Span> agentSpanContextKey;
+    try {
+      Class<?> tracingContextUtils = Class.forName("io.opentelemetry.trace.TracingContextUtils");
+      Field contextSpanKeyField = tracingContextUtils.getDeclaredField("CONTEXT_SPAN_KEY");
+      contextSpanKeyField.setAccessible(true);
```
anuraaga, line 61: Since an agent can take over apps, I'm assuming this is allowed? Not sure

anuraaga:
@trask Added a pretty detailed bridging test based on our SDK ones, looking pretty good

@iNikem I think I got your test cases but let me know if there's anything more to add!

instrumentation/opentelemetry-api-beta/src/test/groovy/ContextBridgeTest.groovy
```diff
@@ -0,0 +1,149 @@
...
+
+  def setupSpec() {
+    otherThread = Executors.newSingleThreadExecutor()
+  }
+
+  def cleanupSpec() {
+    otherThread.shutdown()
+  }
+
+  def "agent and application mix"() {
```
trask, line 36: does this test show interop, or just side-by-side, since everything is in the same thread?

would it be better to `Context.root().makeCurrent().withCloseable {}` (for both agent and non-agent) around each existing `makeCurrent().withCloseable {}` (except for the first) below?

instrumentation/opentelemetry-api-beta/src/test/groovy/ContextBridgeTest.groovy
```diff
@@ -0,0 +1,149 @@
...
+          Span.current().spanContext.traceIdAsHexString == applicationSpan.spanContext.spanIdAsHexString
+          io.opentelemetry.trace.Span.current().spanContext.spanIdAsHexString == applicationSpan2.spanContext.spanIdAsHexString
+          io.opentelemetry.trace.Span.current().spanContext.traceIdAsHexString == applicationSpan2.spanContext.traceIdAsHexString
+          io.opentelemetry.trace.Span.current().spanContext.traceIdAsHexString == applicationSpan.spanContext.traceIdAsHexString
+        }
+      }
+    }
+  }
+
+  def "agent wraps"() {
```
trask, line 97: these tests run with our java-concurrent, so I think that will auto-propagate agentContext(?) which makes it hard to reason about this test

maybe cleaner to do the `Context.root().makeCurrent().withScoped {}` on both, then call `context.wrap(runnable).run()` inside of that, and not mess with threads?

trask (approved):
left a couple suggestions on the tests, but this is great!

and closes #976 👍

instrumentation/opentelemetry-api-beta/src/test/groovy/ContextBridgeTest.groovy
anuraaga: I don't think this one shows interop indeed, just the context parenting. The below tests have "another thread" for propagation test.

Not sure I understand `would it be better to Context.root().makeCurrent().withCloseable {}` - if we mount root in the middle of the tree it's like resetting context, right?

instrumentation/opentelemetry-api-beta/src/test/groovy/ContextBridgeTest.groovy
anuraaga: Heh forgot about that. Yeah that seems like it will test it well.

instrumentation/opentelemetry-api-beta/src/test/groovy/ContextBridgeTest.groovy
trask: ya, you're right, the test below addresses propagating across threads, so no need for that in this test

PR #1495, opened 2020-10-27
More test fixes

iNikem:


anuraaga (approved):

anuraaga:
By the way, any ideas to make `PortUtils.randomOpenPort` more robust with parallel tests? I guess a lot of flakiness is parallel tests using the same value. Perhaps we can have Gradle inject a system property with a random port number so it can make sure each test gets a unique number - but I don't know if it's possible to do that without breaking caching... Just random brainstorming :)

iNikem:
Honestly, I haven't noticed any test failures recently due to clashing ports.

anuraaga:
I see 'Unable to bind' locally a couple times a day I guess while running tests during my SDK snapshot upgrades, it's why it came to mind. Perhaps it's more likely to happen on mac, not sure.

mateuszrzeszutek (approved):

Issue #1274, opened 2020-09-28
Muzzle not working for opentelemetry-api-beta


pavolloffay:
I got the following exception when the app code created a tracer. The agent is using OTel API 0.9.0-SNAPSHOT and the app   0.8.0.

The app didn't crash only the exception was logged. Did the Muzzle prevent the crash? There is no Muzzle mentioned in the logs. So either muzzle was not configured correctly or it could print more understandable output. 

```java
    Tracer foo = OpenTelemetry.getTracer("foo");
    foo.spanBuilder("foo").startSpan().end();
```

```bash
[opentelemetry.auto.trace 2020-09-28 12:04:29:526 +0200] [http-nio-8080-exec-1] ERROR io.opentelemetry.javaagent.tooling.HelperInjector - Error preparing helpers while processing class io.grpc.Context for GrpcContextInstrumentation. Failed to inject helper classes into instance org.springframework.boot.loader.LaunchedURLClassLoader@f446158
java.lang.IllegalStateException: Error invoking (accessor)::defineClass
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection$Dispatcher$UsingUnsafeInjection.defineClass(ClassInjector.java:922)
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection.injectRaw(ClassInjector.java:234)
	at io.opentelemetry.javaagent.tooling.HelperInjector.injectClassLoader(HelperInjector.java:208)
	at io.opentelemetry.javaagent.tooling.HelperInjector.transform(HelperInjector.java:143)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.doTransform(AgentBuilder.java:10325)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.transform(AgentBuilder.java:10263)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.access$1600(AgentBuilder.java:10029)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer$Java9CapableVmDispatcher.run(AgentBuilder.java:10722)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer$Java9CapableVmDispatcher.run(AgentBuilder.java:10660)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
[...]

PR #1509, opened 2020-10-28
Standardize DB query normalizer property names

mateuszrzeszutek:
And add normalizer configuration for Redis, Couchbase, Cassandra and Geode

Resolves: #1431

### DIFF
--- a/instrumentation/lettuce/lettuce-5.1/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v5_1/OpenTelemetryTracing.java
+++ b/instrumentation/lettuce/lettuce-5.1/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v5_1/OpenTelemetryTracing.java
@@ -36,6 +36,9 @@ public enum OpenTelemetryTracing implements Tracing {
   public static final io.opentelemetry.trace.Tracer TRACER =
       OpenTelemetry.getGlobalTracer("io.opentelemetry.auto.lettuce-5.1");
 
+  private static final RedisCommandNormalizer COMMAND_NORMALIZER =
+      new RedisCommandNormalizer("lettuce", "lettuce-5", "lettuce-5.1");
+
   @Override
   public TracerProvider getTracerProvider() {
     return OpenTelemetryTracerProvider.INSTANCE;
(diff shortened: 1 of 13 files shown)


instrumentation/lettuce/lettuce-5.1/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v5_1/OpenTelemetryTracing.java
```diff
@@ -36,6 +36,9 @@
   public static final io.opentelemetry.trace.Tracer TRACER =
       OpenTelemetry.getGlobalTracer("io.opentelemetry.auto.lettuce-5.1");
 
+  private static final RedisCommandNormalizer COMMAND_NORMALIZER =
+      new RedisCommandNormalizer("lettuce", "lettuce-5", "lettuce-5.1");
```
mateuszrzeszutek, line 40: I don't like repeating the instrumentation names here, but you can't exactly access the `Instrumenter` from the advice. Maybe it'll be possible to fix that once https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1274#issuecomment-707860035 is implemented.

trask: yes, that would be very nice to tie this to the existing instrumentation names 👍

trask (approved):
👍

instrumentation/lettuce/lettuce-5.1/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v5_1/OpenTelemetryTracing.java
```diff
@@ -36,6 +36,9 @@
   public static final io.opentelemetry.trace.Tracer TRACER =
       OpenTelemetry.getGlobalTracer("io.opentelemetry.auto.lettuce-5.1");
 
+  private static final RedisCommandNormalizer COMMAND_NORMALIZER =
```
iNikem, line 39: In all other cases you have `commandNormalizer`. Please make the same here as well

iNikem (approved):

iNikem:
@mateuszrzeszutek also please rebase

instrumentation/lettuce/lettuce-5.1/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v5_1/OpenTelemetryTracing.java
mateuszrzeszutek: Done

PR #1513, opened 2020-10-28
Add explicit names to metrics/trace SPI factories

### LINKED ISSUE #1287: Reduce magic of selecting SpanExporterFactory and friends

pavolloffay:

Resolves #1287 

See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md

### DIFF
--- a/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
+++ b/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
@@ -7,10 +7,14 @@ package io.opentelemetry.javaagent.spi.exporter;
 
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
 import java.util.Properties;
+import java.util.Set;
 
 /**
  * A {@link MetricServer} acts as the bootstrap for metric exporters that use {@link MetricProducer}
  * to consume the metrics.
+ *
+ * <p>Multiple names are useful for enabling a pair of span and metric exporters using the same
+ * name, while still having separate names for enabling them individually.
  *
  * <p>Implementation of {@link MetricServer} must be registered through the Java SPI framework.
  */
@@ -23,4 +27,11 @@ public interface MetricServer {
    * @param config The configuration
    */
   void start(MetricProducer producer, Properties config);
+
+  /**
+   * Returns names of metric servers supported by this factory.
+   *
+   * @return The metric server names supported by this factory
+   */
+  Set<String> getNames();
 }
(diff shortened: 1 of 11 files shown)


javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
```diff
@@ -23,4 +24,11 @@
    * @param config The configuration
    */
   void start(MetricProducer producer, Properties config);
+
+  /**
+   * Returns names of metric servers supported by this factory.
```
trask, line 29: i think it's a little confusing why we support multiple names without a bit of explanation:

```suggestion
   * Returns names of metric servers supported by this factory.
   *
   * <p>Multiple names are useful for enabling a pair of span and metric exporters using the same
   * name, while still having separate names for enabling them individually.
```

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/SpanExporterFactory.java
```diff
@@ -21,4 +22,11 @@
    * @return An implementation of a {@link SpanExporter}
    */
   SpanExporter fromConfig(Properties config);
+
+  /**
+   * Returns names of span exporters supported by this factory.
```
trask, line 27: same as above

```suggestion
   * Returns names of span exporters supported by this factory.
   *
   * <p>Multiple names are useful for enabling a pair of span and metric exporters using the same
   * name, while still having separate names for enabling them individually.
```

trask (approved):
Thanks @pavolloffay!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java
```diff
@@ -122,21 +114,11 @@ private static SpanExporterFactory findSpanExporterFactory(String exporterName)
...
 
     for (SpanExporterFactory spanExporterFactory : serviceLoader) {
-      if (spanExporterFactory
-          .getClass()
-          .getSimpleName()
-          .toLowerCase()
-          .startsWith(exporterName(exporterName).toLowerCase())) {
-        return spanExporterFactory;
-      }
+      if (spanExporterFactory.getNames().contains(exporterName)) return spanExporterFactory;
```
iNikem, line 117:
```suggestion
      if (spanExporterFactory.getNames().contains(exporterName)) {
        return spanExporterFactory;
      }
```

iNikem (approved):

PR #1523, opened 2020-10-29
Migrates to using a context when span is started

### LINKED ISSUE #1481: startSpan needs to return Context instead of Span
@marcingrzejszczak is probably one of our first serious users of baggage :) I was helping him debug an issue and noticed that currently it seems impossible to use baggage with our instrumentation because `startSpan` extracts from the request, but only returns a `Span`. The baggage information that was extracted is therefore lost (except for as a private field of `Span`, it's parent).

We need `startSpan` to always return a `Context` to make sure that span is together with its baggage. I think this means that we will remove all usage of `currentContextWith` type of pattern in our repo, since it's really the job of the tracer's to produce the context, not to use the current context directly IIUC. Calling `Span.fromContext` everywhere will be tedious, but it's what the spec wants :)

marcingrzejszczak:
fixes #1481

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
@@ -44,19 +44,20 @@ public abstract class HttpServerTracer<REQUEST, RESPONSE, CONNECTION, STORAGE> e
     super(tracer);
   }
 
-  public Span startSpan(REQUEST request, CONNECTION connection, Method origin) {
+  public Context startSpan(REQUEST request, CONNECTION connection, Method origin) {
     String spanName = spanNameForMethod(origin);
     return startSpan(request, connection, spanName);
   }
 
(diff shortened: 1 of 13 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -44,19 +44,20 @@ public HttpServerTracer(Tracer tracer) {
...
 
-  public Span startSpan(REQUEST request, CONNECTION connection, String spanName) {
+  public Context startSpan(REQUEST request, CONNECTION connection, String spanName) {
     return startSpan(request, connection, spanName, -1);
   }
 
-  public Span startSpan(
+  public Context startSpan(
       REQUEST request, CONNECTION connection, String spanName, long startTimestamp) {
+    Context extract = extract(request, getGetter());
```
anuraaga, line 58: `parentContext`

anuraaga (approved):
Thanks! just a small comment

PR #1525, opened 2020-10-29
Migrates to using a context when span is started

anuraaga:
@marcingrzejszczak asked me to take over this change while he sorts out CLA stuff.

### DIFF
--- a/instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/WebMvcTracingFilter.java
+++ b/instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/WebMvcTracingFilter.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.instrumentation.spring.webmvc;
 
+import io.opentelemetry.context.Context;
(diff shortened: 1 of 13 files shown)


instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/WebMvcTracingFilter.java
```diff
@@ -30,8 +31,8 @@ public WebMvcTracingFilter(Tracer tracer) {
   public void doFilterInternal(
       HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
       throws ServletException, IOException {
-    Span serverSpan = tracer.startSpan(request, request, FILTER_CLASS + "." + FILTER_METHOD);
-
+    Context ctx = tracer.startSpan(request, request, FILTER_CLASS + "." + FILTER_METHOD);
+    Span serverSpan = Span.fromContext(ctx);
```
iNikem, line 35: I think we can optimise this pattern, without taking span from context and then putting it back in `startScope`. But as @marcingrzejszczak  really wants it now, let's do it in follow up

iNikem (approved):

PR #1528, opened 2020-10-29
Run muzzle once per instrumented library - introducing InstrumentationModule

mateuszrzeszutek:
Implementation of this concept: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1274#issuecomment-707860035

Pros:
* shared muzzle matcher, result for CL is cached on the first run, so it's effectively run once for module;
* same with class loader matcher;
* and helper class injector. 

Spring MVC was chosen as the example because it's used in the smoke tests - this way you can see that modules work E2E.

I intentionally made it possible to have both `Instrumenter`s and `InstrumentationModule`s to avoid giant monster PR - if you like this concept I'll replace all `Instrumenter`s in batches.

### DIFF
--- a/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerAdapterInstrumentation.java
+++ b/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerAdapterInstrumentation.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.javaagent.instrumentation.springwebmvc;
 
 import static io.opentelemetry.javaagent.instrumentation.springwebmvc.SpringWebMvcTracer.tracer;
+import static io.opentelemetry.javaagent.tooling.ClassLoaderMatcher.hasClassesNamed;
 import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.implementsInterface;
 import static java.util.Collections.singletonMap;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
@@ -28,7 +29,13 @@ import net.bytebuddy.description.method.MethodDescription;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 
-public final class HandlerAdapterInstrumentation implements TypeInstrumentation {
+final class HandlerAdapterInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    // Optimization for expensive typeMatcher.
+    return hasClassesNamed("org.springframework.web.servlet.HandlerAdapter");
+  }
 
   @Override
   public ElementMatcher<TypeDescription> typeMatcher() {
--- a/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerAdapterInstrumentation.java
+++ b/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerAdapterInstrumentation.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.javaagent.instrumentation.springwebmvc;
 
 import static io.opentelemetry.javaagent.instrumentation.springwebmvc.SpringWebMvcTracer.tracer;
-import static io.opentelemetry.javaagent.tooling.ClassLoaderMatcher.hasClassesNamed;
 import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.implementsInterface;
 import static java.util.Collections.singletonMap;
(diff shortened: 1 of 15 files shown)


iNikem:
@mateuszrzeszutek Can you explicitly describe, what problem do you try to solve here? 

And/or when a developer starts a new instrumentation, how should they decide between using just `Instrumenter` or start new `InstrumentationModule`?

anuraaga:
I guess every module would have one `InstrumentationModule` and then 1+ `TypeInstrumentation`. I like the idea as compared to our current shared base-class approach that has always felt like duplicating something that it shouldn't :D

trask (approved):
👍👍👍

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerAdapterInstrumentation.java
```diff
@@ -16,43 +15,26 @@
...
-  public HandlerAdapterInstrumentation() {
-    super("spring-web");
-  }
-
-  @Override
-  public ElementMatcher<ClassLoader> classLoaderMatcher() {
-    // Optimization for expensive typeMatcher.
-    return hasClassesNamed("org.springframework.web.servlet.HandlerAdapter");
-  }
+public final class HandlerAdapterInstrumentation implements TypeInstrumentation {
```
anuraaga, line 31: Now that it's not service loader, just wondering do these need to be public?

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringMvcInstrumentationModule.java
```diff
@@ -0,0 +1,48 @@
...
+public final class SpringMvcInstrumentationModule extends InstrumentationModule {
+  public SpringMvcInstrumentationModule() {
+    super("spring-mvc");
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+    // Optimization for expensive typeMatchers.
+    return hasClassesNamed(
+        "org.springframework.context.support.AbstractApplicationContext",
```
anuraaga, line 27: The only unfortunate change is that these optimizations for expensive type matcher aren't next to the optimized type matcher anymore. I don't know if that means it's good to accept classloaderMatcher on instrumentation as well or not, wonder what others think.

anuraaga (approved):

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringMvcInstrumentationModule.java
trask: 👍 that would be nice to keep them next to each other somehow

mateuszrzeszutek: Hmm, that is true. I think that we have 2 pattern of `classLoaderMatcher()` usage (correct me if I'm wrong):
1. Optimization for `extends()`/`implements()` `typeMatcher()`
2. library version check, e.g. requiring a class that is present in a particular instrumented library version.

I believe that 2. should stay in the `InstrumentationModule` - 1. should be placed in `TypeInstrumentation` though, as it 's and optimization for a single type instrumentation. I'll add another `classLoaderMatcher()` method to `TypeInstrumentation` and just `and()` those two.

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerAdapterInstrumentation.java
mateuszrzeszutek: Nope, do not need. Thanks for catching this!

mateuszrzeszutek:
> @mateuszrzeszutek Can you explicitly describe, what problem do you try to solve here?

I believe that the core problem here was the inaccurate modelling - we want to instrument a **library**, not a single class; and currently the `Instrumenter` only can instrument single type (or all subtypes of a single class/interface). That's why we have issues like https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1274 where individual type instrumentations may or may not be applied, thus breaking the whole library instrumentation.
With this change it's all or nothing - either we instrument the entire library consistently or we do not instrument anything. 

> I guess every module would have one InstrumentationModule and then 1+ TypeInstrumentation.

Exactly that - ideally 1 library = 1 `InstrumentationModule`, 1+ `TypeInstrumentation`s.

> And/or when a developer starts a new instrumentation, how should they decide between using just Instrumenter or start new InstrumentationModule?

I'll add a deprecation note to `Instrumenter` - the idea is, when all instrumentations are refactored there will be no `Instrumenter`s at all, the only possible way will be to implement an `InstrumentationModule`.

PR #1530, opened 2020-10-30
Activate SystemMetrics

### LINKED ISSUE #1508: Activate SystemMetrics
- `SystemMetrics.registerObservers()` should be called if `oshi-core` dependency is present.
- `IntervalMetricReader` should be configured to send metrics, if not done already.
- Because of shading looks like there is no was to call `SystemMetrics.registerObservers()` manually

malafeev:
fix #1508


instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/AbstractOshiInstrumentation.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.oshi;
+
+import io.opentelemetry.javaagent.tooling.Instrumenter;
+
+public abstract class AbstractOshiInstrumentation extends Instrumenter.Default {
```
trask, line 10: since only one Instrumentation class, don't need to break out abstract base class

instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/OshiInstrumentation.java
```diff
@@ -0,0 +1,49 @@
...
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    return Collections.singletonMap(
+        isMethod().and(isPublic()).and(isStatic()).and(named("getCurrentPlatformEnum")),
+        OshiInstrumentation.class.getName() + "$OshiInstrumentationAdvice");
+  }
+
+  public static class OshiInstrumentationAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter() {
+      OshiInstrumentationHelper.registerObservers();
```
trask, line 46: I don't think u need helper class, can call `SystemMetrics.registerObservers()` directly from here

javaagent/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
```diff
@@ -67,6 +67,17 @@ public static void agentmain(String agentArgs, Instrumentation inst) {
...
+    try {
+      // Call oshi.SystemInfo.getCurrentPlatformEnum() to activate SystemMetrics.
+      // Oshi instrumentation will intercept this call and enable SystemMetrics.
+      Class<?> oshiSystemInfoClass =
+          ClassLoader.getSystemClassLoader().loadClass("oshi.SystemInfo");
+      Method getCurrentPlatformEnumMethod = oshiSystemInfoClass.getMethod("getCurrentPlatformEnum");
+      getCurrentPlatformEnumMethod.invoke(null);
+    } catch (Throwable ex) {
+      // OK
+    }
```
trask, line 80: I'm ok with this approach for now, and we can log an issue to improve this in the future (ideally the agent wouldn't know about the instrumentation), @iNikem @anuraaga?

instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/OshiInstrumentation.java
malafeev: removed helper

instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/AbstractOshiInstrumentation.java
malafeev: removed

trask (approved):

javaagent/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
anuraaga: @trask Do we have any gotchas for using system classloader here? It'd be good to understand well if there are cases we don't support right now.

Following up with the classloader callback approach sounds fine though.

anuraaga (approved):

javaagent/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
trask: Added #1565 to track this

trask:
@malafeev can you follow up with doc PR? I created #1566 to track. thx!

Issue #1409, opened 2020-10-17
SQL Normalizer should allow normalizing IN queries 


billoneil:
When viewing SQL queries analytics/metrics I would like to treat queries containing an `IN` clause as equivalent regardless of the number of parameters passed in.

## Query 1
```sql
select * from table where id in (?)
```

## Query 2
```sql
select * from table where id in (?, ?, ?, ?)
```

These two queries are identical functionally and just have a variable number of parameters passed in. It would be nice if these were automatically normalized to always have only a single `in (?)` or at least have an option to enable this functionality. The current implementation treats these as separate operations.

I would be happy to look into implementing this if I was pointed in the right direction and there was agreement on this behavior or if it should be a configuration option.

trask:
This proposed behavior makes sense to me, I'd like to hear from others (cc: @johnbley)

The current implementation is in [SqlNormalizer.jj](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/jdbc/src/main/javacc/SqlNormalizer.jj), and see [SqlNormalizerTest](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/jdbc/src/test/groovy/SqlNormalizerTest.groovy)

I'm not sure if this can be done using the existing token processor, may need post-process `(?, ?, ..., ?)` --> `(?)`

iNikem:
I think makes sense in the case described by topic starter. But what if there is an inner query? `select * from table where col in (select c from table2)`

billoneil:
@iNikem You are correct this normalization function should not modify subqueries only remove excessive `?`'s. If this is implemented as a post-process step that also adds a few edge cases such as when a query is truncated to its max length we could have an incomplete query.

My initial thoughts on possible edge case input.
| Post-process Input  | Output |
| --------------------|--------|
|`select * from table where id in (?)` | `select * from table where id in (?)`|
|`select * from table where id in (?,?,?)` // Multiple ? no spacing | `select * from table where id in (?)`|
|`select * from table where id in (?, ?, ?, ?)` // Multiple ? with spaces | `select * from table where id in (?)`|
|`select * from table where id in (?, ?,?)` // Multiple ? with mixed spacing | `select * from table where id in (?)`|
|`... id in (?,?,` // Long query truncated trailing comma | `select * from table where id in (?` // Should we close the paren or do we not have enough info?|
[...]

PR #1533, opened 2020-10-30
Simplify JDBC span names

mateuszrzeszutek:
Adding the `<db.operation> <db.name>.<db.sql.table>` idea from https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1409

I think it covers most of the common scenarios (and it's not that complicated).

Spec PR: https://github.com/open-telemetry/opentelemetry-specification/pull/1219


trask (commented):
👍 for moving this POC forward

iNikem (approved):

iNikem:
I would like to get approval either from @trask or @johnbley

javaagent-api/src/test/groovy/io/opentelemetry/javaagent/instrumentation/api/db/normalizer/SqlStatementInfoExtractorTest.groovy
```diff
@@ -0,0 +1,84 @@
...
+    'update /*table'                                                  | new SqlStatementInfo(sql, 'UPDATE', null)
+    // Merge
+    'merge into table'                                                | new SqlStatementInfo(sql, 'MERGE', 'table')
+    'merge table (into is optional in some dbs)'                      | new SqlStatementInfo(sql, 'MERGE', 'table')
+    'merge (into )))'                                                 | new SqlStatementInfo(sql, 'MERGE', null)
+    // Unknown operation
+    'and now for something completely different'                      | new SqlStatementInfo(sql, null, null)
+    ''                                                                | new SqlStatementInfo(sql, null, null)
+    null                                                              | new SqlStatementInfo(sql, null, null)
+  }
```
johnbley, line 59: Maybe add a comment one just to document the known edge case (unless I'm misreading the code and this actually works)?
```
'/*comment*/ select * from table1' -> null/null
```

johnbley (approved):

javaagent-api/src/test/groovy/io/opentelemetry/javaagent/instrumentation/api/db/normalizer/SqlStatementInfoExtractorTest.groovy
mateuszrzeszutek: This actually works, but I'll add a test case for that anyway.

instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcTracer.java
```diff
@@ -69,19 +71,44 @@ public Span startSpan(PreparedStatement statement) {
...
-  protected String normalizeQuery(String query) {
-    return JDBCUtils.normalizeSql(query);
+  protected String spanName(DBInfo connection, SqlStatementInfo query, String normalizedQuery) {
+    String dbName = dbName(connection);
+    if (query.getOperation() == null && query.getTable() == null) {
+      return dbName == null ? DB_QUERY : dbName;
+    }
+
+    StringBuilder name = new StringBuilder();
+    name.append(query.getOperation()).append(' ');
```
trask, line 101: can `query.getOperation()` be `null` here?

maybe a unit test for this method to test the different combinations?

trask (approved):

instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcTracer.java
mateuszrzeszutek: It can't - the only possible options are `null, null`, `operation, null`, `operation, table`.

I'll add more unit tests anyway.

instrumentation/jdbc/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcTracer.java
```diff
@@ -69,19 +71,44 @@ public Span startSpan(PreparedStatement statement) {
...
+  protected String normalizeQuery(SqlStatementInfo query) {
+    return query.getFullStatement();
   }
 
   @Override
-  protected String normalizeQuery(String query) {
-    return JDBCUtils.normalizeSql(query);
+  protected String spanName(DBInfo connection, SqlStatementInfo query, String normalizedQuery) {
+    String dbName = dbName(connection);
+    if (query.getOperation() == null && query.getTable() == null) {
```
trask, line 96: I think this makes more clear that query.getOperation() can't be null below
```suggestion
    if (query.getOperation() == null) {
```

trask:
Nice 🎉

PR #1536, opened 2020-10-30
AWS lambda instrumentation - HTTP context propagation

### LINKED ISSUE #1535: Enable HTTP trace propagation for AWS lambda instrumentation
Currently AWS lambda supports trace propagation only with X-Ray.

User should be able to configure any HTTP headers-based trace propagation for requests coming via API Gateway.

None

**Additional context**
None

kubawach:
Resolves #1535 

Sketch of an idea - for streaming handler / wrapper only. If the idea holds, will provide also for instrumentation part and regular handler / wrapper (in applicable cases).


kubawach:
@anuraaga please have a look at this - HTTP trace propagation sketch / idea

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final Logger log = LoggerFactory.getLogger(ApiGatewayProxyRequest.class);
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  static {
+    OBJECT_MAPPER.registerModule(new AfterburnerModule());
+  }
+
+  private final Headers headers;
+
+  private ApiGatewayProxyRequest(Headers headers) {
```
anuraaga, line 29: This is unfortunate but seems to be the only way to support streams, oh well.

For non-streams, will we use the event types?
https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
```diff
@@ -50,11 +51,16 @@ protected TracingRequestStreamHandler(AwsLambdaTracer tracer) {
   }
 
   @Override
-  public final void handleRequest(InputStream input, OutputStream output, Context context)
+  public void handleRequest(InputStream input, OutputStream output, Context context)
       throws IOException {
-    Span span = tracer.startSpan(context, Kind.SERVER);
+
+    InputStreamFactory streamFactory = InputStreamFactory.forStream(input);
```
anuraaga, line 57: Do you think it's possible to do this only if the propagator is configured?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java
anuraaga: Though I wonder if supporting stream is actually important. Looking at this, it seems like the body is encoded into JSON anyways so there is no benefit in the stream handler - so maybe it's a reasonable and practical limitation for us to just detect that event type?

anuraaga:
Supporting HTTP propagation is important so generally looks nice, thanks :)

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
kubawach: Great idea, will implement a switch,

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java
kubawach: Yes, events for non-streams should be doable, at least at the wrapper level. Handler gets argument type from a child class so in this case afair it's not possible to get the headers if arg is not the event type (correct me if I'm wrong). 

With regards to streams, now we have safeguards (only if propagators configured) and current implementation of the AWS proxy uses ButeArrayInputStream which is markable, so I'd say it's safe to leave it (as a drop-in method for customers already suing streams handler). 
If you agree, we can merge this PR and I'll file another one for non-streams wrapper.

kubawach:
@anuraaga pls have another look :)

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
```diff
@@ -17,8 +17,12 @@ apply from: "$rootDir/gradle/instrumentation-library.gradle"
 
 dependencies {
   library group: 'com.amazonaws', name: 'aws-lambda-java-core', version: '1.0.0'
+  library group: 'com.amazonaws.serverless', name: 'aws-serverless-java-container-core', version: '1.5.2'
+
   compileOnly deps.opentelemetrySdk
 
+  compile('com.fasterxml.jackson.core:jackson-databind:2.10.3', 'com.fasterxml.jackson.module:jackson-module-afterburner:2.9.10')
```
anuraaga, line 24: I don't think we need afterburner since we're not mapping to POJO, just reading the tree.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java
```diff
@@ -0,0 +1,120 @@
...
+  private static final Logger log = LoggerFactory.getLogger(ApiGatewayProxyRequest.class);
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  static {
+    OBJECT_MAPPER.registerModule(new AfterburnerModule());
+  }
+
+  private static boolean noHttpPropagationNeeded() {
+    return OpenTelemetry.getGlobalPropagators().getTextMapPropagator().fields().isEmpty();
```
anuraaga, line 33: IIUC, if the x-ray propagator is configured, fields wouldn't be empty but in that case we want to read the env var. Should we compare to its fields?

anuraaga (commented):
Thanks, just a couple of points

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java
kubawach: Hmmm yes, sound idea. I forgot about non-http propagator :)

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
kubawach: There is a mapping call fetching Headers class instance - `return OBJECT_MAPPER.readValue(jParser, Headers.class);` I believe that's optimised with afterburner (although I'm not expert here ;) ).

kubawach:
@anuraaga added support for XRay propagator and added some tests, pls have another look

kubawach:
Hi @anuraaga please have another look, resolved / responded to your previous comments. Thanks!

mateuszrzeszutek (approved):

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
anuraaga: Ah I missed that. That being said, since we're telemetry I don't think we should try for "more than the base". Do you know if aws SDK uses afterburner? If they do no problem adding it :) But if not we should try to stick to their dependence as much as possible and I would avoid it.

anuraaga (approved):
Thanks!

iNikem:
Trying to tigger EasyCLA

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
kubawach: Noted for future assignments, thanks :) Here it's ok as the afterburner is references by the SDK (in particular by 'aws-serverless-java-container-core').

PR #1540, opened 2020-11-02
Rename TRACER to tracer()

### LINKED ISSUE #1195: TRACER or tracer()
We currently have a pattern of using a public static final constant for an instrumentation tracer. We could provide a static method instead

Pros

- More inline with JDK which stopped using the pattern of public constants I believe

- Less all-caps yelling in code

Cons

- Annoying to write an accessor in addition to a private constant. When using enum singleton pattern generally requires an inner enum class

- More characters at call site (parens).

While I tend to be more used to accessor method pattern, writing this out not sure it's all that much better but worth a small thought 😅

anuraaga:
We don't have to go with this but figured it's worth taking a look at.

Fixes #1195

### DIFF
--- a/benchmark-integration/jetty-perftest/src/main/java/io/opentelemetry/perftest/Worker.java
+++ b/benchmark-integration/jetty-perftest/src/main/java/io/opentelemetry/perftest/Worker.java
@@ -13,11 +13,11 @@ import java.util.concurrent.TimeUnit;
 
 public class Worker {
 
-  private static final Tracer TRACER = OpenTelemetry.getGlobalTracer("io.opentelemetry.auto");
+  private static final Tracer tracer = OpenTelemetry.getGlobalTracer("io.opentelemetry.auto");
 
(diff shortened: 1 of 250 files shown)


iNikem (approved):
I certainly approve the removal of all caps name. Do we need an accessor method or can we just use `public static final Tracer tracer`? :)

anuraaga:
I'm not sure - lately I learned something new that methods are way easier to instrument with the agent than fields which is interesting. But obviously that doesn't matter in our agent itself. Methods I guess do give some flexibility, maybe branching return value based on some config even at runtime? With fields locked into the public API, I guess this becomes impossible.

Downside is it takes more effort for hotspot to warmup with additional inlining of methods which probably does matter somewhat. What do folks think?

trask:
As a default, I tend to only make constants (i.e. UPPER_CASE) public

trask (approved):

PR #1543, opened 2020-11-02
Smoke tests using Jaeger and Zipkin exporters

### LINKED ISSUE #1541: Smoke tests with Jaeger and Zipkin
Our Roadmap says that we want "End-to-end tests with OTLP, Jaeger and Zipkin". Currently only OTLP is implemented, Jaeger and Zipkin should be added.

@anuraaga this means either reverting #1136 or introducing Collector between tests and fake backend. Any suggestions or preferences?

iNikem:
Closes #1541


trask (approved):
Nice

anuraaga (approved):
I don't think there's a problem using Zipkin or Jaeger here - testing against every compatible backend is a bit different than testing against the real deals, and presumably it's why the exporters exist anyway ;) That being said, since our tests are very basic smoke tests, not really validating the server is sent data it expects, I guess it doesn't matter too much.

PR #1545, opened 2020-11-03
add baggage propagator to default propagators

malafeev:
cc @carlosalberto


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -49,7 +52,11 @@ public static void initializePropagators(List<String> propagators) {
       //  https://github.com/open-telemetry/opentelemetry-java/issues/1742
       OpenTelemetry.setGlobalPropagators(
           DefaultContextPropagators.builder()
-              .addTextMapPropagator(HttpTraceContext.getInstance())
+              .addTextMapPropagator(
+                  TraceMultiPropagator.builder()
+                      .addPropagator(HttpTraceContext.getInstance())
+                      .addPropagator(W3CBaggagePropagator.getInstance())
```
carlosalberto, line 58: W3CBaggagePropagator shouldn't be registered as part of `TraceMultiPropagator`, i.e. it should be added directly through `addTextMapPropagator()` ;)

malafeev: ups

malafeev: fixed

carlosalberto:
IIRC there's a spot where, if the user specified values (`b3`, `jaeger`, etc), we put all the propagators in a `TraceMultiPropagator`, right? If that's the case, we need to update that as well, to prevent adding the Baggage propagator there ;)

malafeev:
@carlosalberto if user specifies propagators then Baggage propagator will not be added.
Baggage propagator added only if user didn't specify any propagator.

trask:
> Am I right that `TraceMultiPropagator` should never contain Baggage propagator even if user specified multiple propagators including baggage?

Yes, I hadn't thought about this, but it makes sense, as `TraceMultiPropagator` stops extracting once it extracts a span context successfully.

malafeev:
@trask @carlosalberto fixed  adding Baggage propagator to TraceMultiPropagator

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -50,6 +53,7 @@ public static void initializePropagators(List<String> propagators) {
       OpenTelemetry.setGlobalPropagators(
           DefaultContextPropagators.builder()
               .addTextMapPropagator(HttpTraceContext.getInstance())
+              .addTextMapPropagator(W3CBaggagePropagator.getInstance())
```
malafeev, line 56: @carlosalberto 
`W3CBaggagePropagator` should be first one or last one? 
it doesn't matter for inject but for extract we return context from last propagator.

carlosalberto: It does't matter for the default configuration case 😄

trask: @carlosalberto does it matter the order we add `W3CBaggagePropagator` for the non-default configuration case?

carlosalberto: @trask So `Baggage` propagators work just like the Trace ones (B3, TraceContext, etc), regarding the fact that, upon successful extraction, any previous value will be overridden (i.e. that last one wins).

Currently only the Jaeger propagator does `Baggage` extraction besides `W3CBaggage`, so the order _may_ matter in such case, e.g. if the user explicitly specifies `OTEL_PROPAGATORS=jaeger,baggage`, then `W3CBaggage` will have higher priority, given the case that `Baggage` was propagated suing **both** `Jaeger` and `W3CBaggage` formats.

Then again, when users specify this, they should know what they are doing, IMHO 😉 (although maybe we should add a note in OTel Java regarding this, etc).

trask: @carlosalberto do u know what we should do if the user specifies:

`OTEL_PROPAGATORS=jaeger,baggage,tracecontext`

Should we add both TraceMultiPropagator(jaeger,tracecontext) and W3BaggagePropagator? And if so, do u know whether we should add TraceMultiPropagator first or last?

carlosalberto: After thinking a little bit more about this, I think we should leave `JaegerPropagator` out of the `TraceMultiPropagator` given that it **not only** extracts `SpanContext`. For example, `TraceMultiPropagator` with "tracecontext,jaeger" will stop if trace-context succeeded, without giving `JaegerPropagator` a chance to extract any existing `Baggage` ;)

It would mean, given your example, that you would end up with Jaeger + Baggage + TraceContext (no TraceMultiPropagator needed here).

Other than that, the order doesn't matter much - although `W3CBaggagePropagator` should probably come _after_ `JaegerPropagator`, as it can have more complex/complete values that Jaeger baggage lacks, e.g. metadata.

(We could split `JaegerPropagator` into a trace AND a baggage propagators but that slightly seems like an overkill)

carlosalberto: @malafeev ^ ;)

malafeev: so `JaegerPropagator` always should be added via `addTextMapPropagator`?
And next example is valid?:
```
DefaultContextPropagators.builder()
    .addTextMapPropagator(JaegerPropagator.getInstance())
    .addTextMapPropagator(W3CBaggagePropagator.getInstance()) 
    .addTextMapPropagator(TraceMultiPropagator.builder()
        .addPropagator(HttpTraceContext.getInstance())
        .addPropagator(AwsXRayPropagator.getInstance()).build())
    ...
```
?

carlosalberto: @malafeev  Yes ;) - and put a comment in the code on why we are doing it (i.e. Jaeger handles BOTH tracing and baggage, etc).

malafeev: done

iNikem: @malafeev I think both users and developers will have a hard time understanding how `otel.propagators` works now :) Please document the final result both in `PropagatorsInitializer.java` and in README. The latter, for example, currently does very poor job explaining the difference between trace context propagators and baggage ones.

malafeev: @iNikem I added javadoc, is it clear enough?

iNikem: Sorry, but no :) I will add comments to javadoc. And we still have update README

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -39,9 +42,25 @@
...
 
-  /** Initialize OpenTelemetry global Propagators with propagator list, if any. */
+  /**
+   * Initialize OpenTelemetry global Propagators with propagator list, if any.
+   *
+   * <p>Because TraceMultiPropagator returns first successful extracted Context and stops further
+   * extraction next rules applied:
+   *
+   * <ul>
+   *   <li>W3CBaggagePropagator is added to DefaultContextPropagators to allow another propagator
```
iNikem, line 55: Is it always added? Or only if `baggage` value is provided?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -39,9 +42,25 @@
...
+   *       extract Context.
+   *   <li>JaegerPropagator is added to DefaultContextPropagators because it extracts both Context
```
iNikem, line 57: Is it always added? Or only if `jaeger` value is provided?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -39,9 +42,25 @@
...
+   *       and Baggage. Otherwise in TraceMultiPropagator it may not get a chance to extract any
+   *       existing Baggage.
+   *   <li>W3CBaggagePropagator comes after JaegerPropagator, as it can have more complex/complete
```
iNikem, line 60: Above you say "returns first successful extracted Context", here you seem to say that later Baggage extraction can enrich the previous one. If context and baggage extractions behave differently in this regard, we have to say so explicitly.

iNikem (changes requested):
This PR introduces quiet complex rules about how propagators are configured and how they impact each other. These rules are not documented neither for developers in `PropagatorsInitializer` nor for end-users in README.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
malafeev: added `can be` ... `only`

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
malafeev: added `can be` ... `only`

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
malafeev: updated.

malafeev:
I didn't update README yet, let me finish with javadoc.
And I'm not sure that it should be in instrumentation README, maybe better place is otel-java.

malafeev:
updated README

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -50,15 +70,31 @@ public static void initializePropagators(List<String> propagators) {
...
     }
 
     DefaultContextPropagators.Builder propagatorsBuilder = DefaultContextPropagators.builder();
 
     List<TextMapPropagator> textPropagators = new ArrayList<>(propagators.size());
-    for (String propagatorId : propagators) {
-      TextMapPropagator textPropagator = TEXTMAP_PROPAGATORS.get(propagatorId.trim().toLowerCase());
+    List<String> propagatorIds =
+        propagators.stream()
+            .map(propagator -> propagator.trim().toLowerCase())
```
trask, line 83: I don't see mention of supporting case-insensitivity in [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md), I would suggest leaving this out?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -50,15 +70,31 @@ public static void initializePropagators(List<String> propagators) {
...
+            .collect(Collectors.toList());
+
+    if (propagatorIds.remove(JAEGER)) {
+      // Jaeger handles both tracing and baggage
+      propagatorsBuilder.addTextMapPropagator(JaegerPropagator.getInstance());
+      log.info("Added " + JaegerPropagator.getInstance() + " propagator");
```
trask, line 89: doesn't seem important enough to log by default (default is info)
```suggestion
      log.debug("Added " + JaegerPropagator.getInstance() + " propagator");
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -50,15 +70,31 @@ public static void initializePropagators(List<String> propagators) {
...
+    }
+    if (propagatorIds.remove(BAGGAGE)) {
+      propagatorsBuilder.addTextMapPropagator(W3CBaggagePropagator.getInstance());
+      log.info("Added " + W3CBaggagePropagator.getInstance() + " propagator");
```
trask, line 93:
```suggestion
      log.debug("Added " + W3CBaggagePropagator.getInstance() + " propagator");
```

README.md
```diff
@@ -150,14 +150,17 @@ attributes to stdout. It's mainly used for testing and debugging.
...
+to extract any existing Baggage, instead they can be added only to DefaultContextPropagators. 
+The multi-propagator attempts to extract the context from incoming requests using each of the 
+configured propagator formats (in order), stopping after the first successful context extraction. 
+The multi-propagator injects the context into outgoing requests using all the configured propagator 
+formats.
 
 | System property  | Environment variable | Description                                                                                                     |
 |------------------|----------------------|-----------------------------------------------------------------------------------------------------------------|
-| otel.propagators | OTEL_PROPAGATORS     | Default is `tracecontext` (W3C). Other supported values are `b3`, `b3single`, `jaeger`, `ottracer`, and `xray`. |
+| otel.propagators | OTEL_PROPAGATORS     | Default is `tracecontext,baggage` (W3C). Other supported values are `b3`, `b3single`, `jaeger`, `ottracer`, and `xray`. |
```
trask, line 163:
```suggestion
| otel.propagators | OTEL_PROPAGATORS     | The propagators to be used. Use a comma-separated list for multiple propagators. Supported propagators are `tracecontext`, `baggage`, `b3`, `b3single`, `jaeger`, `ottracer`, and `xray`. Default is `tracecontext,baggage` (W3C). |
```

README.md
```diff
@@ -150,14 +150,17 @@ attributes to stdout. It's mainly used for testing and debugging.
...
-to extract the context from incoming requests using each of the configured propagator formats (in order),
-stopping after the first successful context extraction. The multi-propagator injects the context into
-outgoing requests using all the configured propagator formats.
+If set to a comma-separated list of the values, the multi-propagator is used. Special case is `baggage` 
+and `jaeger` propagators: they cannot be added to multi-propagator because they may not get a chance 
+to extract any existing Baggage, instead they can be added only to DefaultContextPropagators. 
+The multi-propagator attempts to extract the context from incoming requests using each of the 
+configured propagator formats (in order), stopping after the first successful context extraction. 
+The multi-propagator injects the context into outgoing requests using all the configured propagator 
+formats.
```
trask, line 159: I think this is going to be really confusing for users, suggest we just leave it out, and if anyone asks, refer them to the spec. 

```suggestion
```

README.md
```diff
@@ -150,14 +150,17 @@ attributes to stdout. It's mainly used for testing and debugging.
 
 The propagator controls which distributed tracing header format is used.
```
trask, line 151: also change section title to "Propagators"
```suggestion
The propagators determine which distributed tracing header formats are used, and which baggage propagation header formats are used.
```

trask (commented):
ok, I think I understand propagators now 😅 

of course, that's what I thought before too 😁

last request is that I think this behavior of `OTEL_PROPAGATORS` should be documented in [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-environment-variables.md)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
malafeev: it was from beginning I suppose, didn't want change behavior

malafeev:
@trask I applied your suggested changes except one about case sensitivity.

trask:
> last request is that I think this behavior of OTEL_PROPAGATORS should be documented in the spec

@carlosalberto can you take this?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
trask: ok, I created #1632 to track this

carlosalberto:
@trask This part is not documented at all in the Spec (although there's a ticket to standardize `TraceMultiPropagator`, which only exists for Java at the moment, as an improvement in order to not extract `SpanContext` more than once), so no need to do it there for now.

However, we should, for sanity purposes, document this in `TraceMultiPropagator` in OTel Java, mentioning it should **only** contain `SpanContext`-only `Propagators`, in order to avoid unexpected, incorrect scenarios. Does that makes sense?

(We **could** also document in the Spec that the `OTEL_PROPAGATOR` values, and probably all env-var values, should be lower-case. Sounds about right?)

trask:
thanks @carlosalberto 

my concern with this not being defined in the spec, is that once it is defined, if it varies from our implementation, it would be a breaking change for our users

carlosalberto:
Hey @trask 

Sorry for the delay.

> my concern with this not being defined in the spec, is that once it is defined, if it varies from our implementation, it would be a breaking change for our users

So the idea of `TraceMultiPropagator` is simply that we don't extract `SpanContext` more than once (as an optimization), and MUST NOT change the propagation semantics at all. Logically speaking, the only thing to be careful about these days, is that **only** trace-only Propagators are added to it.

So, when we standardize this in the Spec, semantics MUST NOT change, and we will simply offer this as a optimization.

trask:
@carlosalberto thanks for that explanation, I think I understand now(!?).

@malafeev I sent a PR to your PR that I think would clarify a lot of questions/troubles that reviewers had understanding this.

malafeev:
@trask I merged your PR

iNikem:
@trask @malafeev @carlosalberto Can you please update me on this PR? I am still concerned that just from README nobody will understand how to configure propagators. Especially when interested in baggage as well.

malafeev:
For user it's simple like `OTEL_PROPAGATORS=b3,baggage`. Internally it's complicated but not for end user.

iNikem:
> For user it's simple like `OTEL_PROPAGATORS=b3,baggage`. Internally it's complicated but not for end user.

So if I specify `b3multi`, I will not get baggage? So I have to always specify `baggage` if I want to propagated? What about `jaeger,baggage`?

iNikem:
How should user know at all that some of the "Supported propagators are `tracecontext`, `baggage`, `b3`, `b3multi`, `jaeger`, `ottracer`, and `xray`" are trace only (which ones?), some are baggage only (which ones?) and some are both?

malafeev:
I don't thinks these questions should be answered here. Explanation of how propagators work is the otel-java or spec job.

iNikem:
> I don't think these questions should be answered here. Explanation of how propagators work is the otel-java or spec job.

You may be right, I will not argue. Then we should link to that explanation from our README.

carlosalberto:
@iNikem Hah! Yes, the user doesn't really need to worry about the propagators, at least through `OTEL_PROPAGATORS` - actual users of `TraceMultiPropagator` *do* however, as it helps to prevent double extraction of `SpanContext`, with the restriction that if another concern is extracted as part of it, such as baggage, it MUST NOT be included there (Jaeger is the prime example).

We should definitely document that (specially in OTel Java itself).

trask:
@iNikem is it ok to dismiss your request for changes, or is there anything else you'd like in this PR?

trask:
Closes #1632 (does it work to put that here?)

iNikem (approved):
I have created #1716 for proper documentation.

iNikem:
> @iNikem is it ok to dismiss your request for changes, or is there anything else you'd like in this PR?

why do you ask when you haven't yet approved it yourself? :)

trask:
> why do you ask when you haven't yet approved it yourself? :)

oops 😅

trask (approved):

PR #1549, opened 2020-11-04
Upgrade to latest SDK snapshot

anuraaga:
Confirmed smoke tests built locally pass, they won't pass on CI for this PR.


iNikem:
If smoke tests will not pass for this PR we cannot merge this PR. Is it possible to make a separate change to prepare smoke tests apps for this?

anuraaga:
Do you mean versioning the smoke test images? If so I could do that yeah.

iNikem:
Versioning is already being done in #1552

I meant that if code changes require new smoke test images, then it makes sense first to make a PR for those images and then a PR which introduce code changes.

anuraaga:
But it's so far a chicken egg issue - rebuilt smoke test images would fail until the new code is in. So with the versioning it will be ok.

trask (approved):

anuraaga:
Noticed that the recent updates to smoke tests building updated them to a version that passes this build (and fails the rest). So going to go ahead and merge if this goes green but accept-pr might have flaked.

Issue #1546, opened 2020-11-03
Tag smoke test docker images with proper versions


vovencij:
Currently docker images for smoke-test test apps are built and published without version tags. Given that these apps may use annotations or other dependencies from this or other openetelemetry project it is possible that other branches or forks (or even projects that happen to use these demo apps in their own tests) may break, because incompatible changes are published. 

Having test app images published with the same version as other artefacts would eliminate the problem.

PR #1552, opened 2020-11-04
Tag smoke tests

vovencij:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1546

### DIFF
--- a/smoke-tests/fake-backend/build.gradle
+++ b/smoke-tests/fake-backend/build.gradle
@@ -30,8 +30,8 @@ dependencies {
 }
 
 jib {
-  container.creationTime = 'USE_CURRENT_TIMESTAMP'
   from.image = "gcr.io/distroless/java-debian10:11"
   to.image = "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend"
   to.tags = [new Date().format("yyyyMMdd.HHmmSS")]
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
 }
--- a/smoke-tests/fake-backend/build.gradle
+++ b/smoke-tests/fake-backend/build.gradle
@@ -30,10 +30,8 @@ dependencies {
 }
 
 jib {
-  to {
-    image = "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend"
-  }
-  from {
-    image = "gcr.io/distroless/java-debian10:11"
-  }
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
+  from.image = "gcr.io/distroless/java-debian10:11"
+  to.image = "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend"
+  to.tags = [new Date().format("yyyyMMdd.HHmmSS")]
 }
--- a/smoke-tests/play/build.gradle
+++ b/smoke-tests/play/build.gradle
@@ -40,5 +40,7 @@ def targetJDK = project.hasProperty("targetJDK") ? project.targetJDK : 11
 jib {
   from.image = "bellsoft/liberica-openjdk-alpine:$targetJDK"
   to.image = "open-telemetry-docker-dev.bintray.io/java/smoke-play-jdk$targetJDK"
+  to.tags = [new Date().format("yyyyMMdd.HHmmSS")]
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
   container.mainClass = "play.core.server.ProdServerStart"
 }
--- a/smoke-tests/grpc/build.gradle
+++ b/smoke-tests/grpc/build.gradle
@@ -38,6 +38,8 @@ compileJava {
 def targetJDK = project.hasProperty("targetJDK") ? project.targetJDK : 11
 
 jib {
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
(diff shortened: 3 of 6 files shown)


smoke-tests/fake-backend/build.gradle
```diff
@@ -30,10 +30,8 @@ dependencies {
...
-  to {
-    image = "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend"
-  }
-  from {
-    image = "gcr.io/distroless/java-debian10:11"
-  }
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
+  from.image = "gcr.io/distroless/java-debian10:11"
+  to.image = "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend"
+  to.tags = [new Date().format("yyyyMMdd.HHmmSS")]
```
iNikem, line 36: Will that add `latest` tag as well?

smoke-tests/fake-backend/build.gradle
```diff
@@ -30,10 +30,8 @@ dependencies {
 }
 
 jib {
-  to {
-    image = "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend"
-  }
-  from {
-    image = "gcr.io/distroless/java-debian10:11"
-  }
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
```
iNikem, line 33: Is there a specific need for this change?

smoke-tests/play/build.gradle
```diff
@@ -40,5 +40,7 @@ def targetJDK = project.hasProperty("targetJDK") ? project.targetJDK : 11
 jib {
   from.image = "bellsoft/liberica-openjdk-alpine:$targetJDK"
   to.image = "open-telemetry-docker-dev.bintray.io/java/smoke-play-jdk$targetJDK"
+  to.tags = [new Date().format("yyyyMMdd.HHmmSS")]
+  container.creationTime = 'USE_CURRENT_TIMESTAMP'
```
iNikem, line 44: Please use consistent order of configuration

smoke-tests/fake-backend/build.gradle
vovencij: According to specification, `latest` tag is pointing at the latest specifically untagged image (which is `to.image`). This is what my docker images shows me:
```
REPOSITORY                                                              TAG                     IMAGE ID            CREATED             SIZE
open-telemetry-docker-dev.bintray.io/java/smoke-springboot-jdk11        20201104.1241567        c9cef0f61231        14 seconds ago      147MB
open-telemetry-docker-dev.bintray.io/java/smoke-springboot-jdk11        latest                  c9cef0f61231        14 seconds ago      147MB
open-telemetry-docker-dev.bintray.io/java/smoke-grpc-jdk11              20201104.1227972        bdff883aafb7        14 minutes ago      149MB
open-telemetry-docker-dev.bintray.io/java/smoke-grpc-jdk11              latest                  bdff883aafb7        14 minutes ago      149MB
open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend            20201104.1225631        de3c1e136bb9        15 minutes ago      231MB
open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend            latest                  de3c1e136bb9        15 minutes ago      231MB
open-telemetry-docker-dev.bintray.io/java/smoke-springboot-jdk11        20201104.1222643        ce4f7244e018        19 minutes ago      147MB
```

smoke-tests/fake-backend/build.gradle
vovencij: It is super inconvenient, that by default images built by jib will have 0 as a creation timestamp, which breaks ordering in `docker images`, because all such images are created `50 years ago`. Issue is tagged `contributor experience`, so I find this change is needed.

iNikem: So you don't think Jib's desire for reproducible builds as convincing? :) Why ordering of `docker images` is important/inconvenient?

vovencij: Achieving the desire for reproducible builds by lying about image creation time is not really convincing, yes. On multiple occasions debugging/troubleshooting either in development or in production I have benefitted from correct creation time that `docker images` show, because then I can easily spot, that some build step somewhere has failed silently and I'm using old images, that explain why something is not working as expected.

smoke-tests/play/build.gradle
vovencij: All `container` blocks are moved after `to`/`from` settings.

smoke-tests/fake-backend/build.gradle
iNikem: But now you have build time in image tag. Isn't it enough?

What happens when you run `./gradlew jibDockerBuild` two times in a row with `USE_CURRENT_TIMESTAMP` and without? Will it create two docker images in both cases or not?

vovencij: If you don't change anything, and will just rebuild, it will not build a new image in either case. You will have two tags pointing at the same image ID. It only matters, if you do a change in some source file like A -> B -> A. Then in case of creation timestamp zero you will get 2 image ID-s, one for A and one for B. But in case of `USE_CURRENT_TIMESTAMP` you will get 3 distinct image IDs.

PR #1559, opened 2020-11-05
Tag smoke test images with date and run ID from workflow.

anuraaga:
I think it's too confusing to have three different tags (based on the seconds value) for the same image but different JRE. We have a perfectly good identifier that's consistent and back-referencable, the github sha :)

Changelog:

Tooling:
- Smoke test images now are tagged with a consistent tag per build of "DATE.TIME.GITHUB_RUN_ID"


trask (approved):

iNikem:
sha does not have order (older-younger). Stable tags are good, sha is bad :)

vovencij:
I would suggest to use some kind of stable run number from github action (which doesn't change when action is re-run on error). And reduce the tag to `yyyyMMdd.<buildNumber>` then we will not produce new images on re-running the action.

.github/workflows/build-grpc-smoke-dist.yaml
```diff
@@ -28,7 +28,9 @@ jobs:
 
       - name: Build Docker Image
         run: |
-          ./gradlew jib -PtargetJDK=8 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace
-          ./gradlew jib -PtargetJDK=11 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace
-          ./gradlew jib -PtargetJDK=15 -Djib.httpTimeout=120000 -Djib.console=plain --stacktrace
+          TAG=$(date '+%Y%m%d.%H%M%S')
```
iNikem, line 31: Please think about @vovencij suggestion about using date+number as more stable tag.

anuraaga:
Thanks - used the format.

anuraaga:
Oops - didn't notice the removal of seconds but sounds nice!

anuraaga:
Confirmed logic here - https://github.com/anuraaga/opentelemetry-java-instrumentation/runs/1357364079?check_suite_focus=true

iNikem (approved):

mateuszrzeszutek (approved):

PR #1593, opened 2020-11-09
Fix ClassNotFoundException on Wildfly

### LINKED ISSUE #1591: Wildfly apps fail when running javaagent
I'm not sure yet why our wildfly smoke test didn't catch this, but deploying a simple hello world war to wildfly and then hitting the url results in this error:

```
Caused by: java.lang.ClassNotFoundException: io.opentelemetry.javaagent.shaded.instrumentation.api.internal.BootstrapPackagePrefixesHolder from [Module "io.undertow.jsp" version 2.0.9.Final from local module loader @421d54b3 (finder: local module finder @63f40ca0 (roots: C:\temp\wildfly-21.0.0.Final\modules,C:\temp\wildfly-21.0.0.Final\modules\system\layers\base))]
        at org.jboss.modules.ModuleClassLoader.findClass(ModuleClassLoader.java:255)
        at org.jboss.modules.ConcurrentClassLoader.performLoadClassUnchecked(ConcurrentClassLoader.java:410)
        at org.jboss.modules.ConcurrentClassLoader.performLoadClass(ConcurrentClassLoader.java:398)
        at org.jboss.modules.ConcurrentClassLoader.loadClass(ConcurrentClassLoader.java:116)
        ... 52 more
```

The problem appears to be our usage of `BootstrapPackagePrefixesHolder` inside of `ClassLoaderInstrumentation`:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/71f91877abf7456cc40013a78dc5207622aee365/instrumentation/java-classloader/src/main/java/io/opentelemetry/javaagent/instrumentation/javaclassloader/ClassLoaderInstrumentation.java#L93

The `ClassLoaderInstrumentation` is supposed to force class loaders to use the bootstrap class loader for our classes, so (in hindsight) it makes sense that we shouldn't reference one of our classes directly that has to be found in the bootstrap class loader.

This worked previously because it was using `Constants` class to get the list of bootstrap package prefixes, and the `Constants` class was registered as a helper class so that it would be injected into the class loader (instead of having to be resolved from the bootstrap class loader).

trask:
Resolves #1591.

I will open a new issue for fixing `BootstrapPackagesProvider` which this PR now breaks (see TODO comment in PR).

### DIFF
--- a/instrumentation/java-classloader/src/main/java/io/opentelemetry/javaagent/instrumentation/javaclassloader/ClassLoaderInstrumentation.java
+++ b/instrumentation/java-classloader/src/main/java/io/opentelemetry/javaagent/instrumentation/javaclassloader/ClassLoaderInstrumentation.java
@@ -18,7 +18,6 @@ import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
 
 import com.google.auto.service.AutoService;
-import io.opentelemetry.instrumentation.api.internal.BootstrapPackagePrefixesHolder;
 import io.opentelemetry.javaagent.instrumentation.api.CallDepthThreadLocalMap;
 import io.opentelemetry.javaagent.tooling.Constants;
 import io.opentelemetry.javaagent.tooling.Instrumenter;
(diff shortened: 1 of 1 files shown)


trask:
...and another issue to track down why our wildfly smoke test didn't catch this.

trask:
cc: @pavolloffay heads up

anuraaga (approved):

pavolloffay:
Yay, can we get a fix for this?

trask:
@pavolloffay just confirming, u were asking for (now released) https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v0.10.1?

Issue #1579, opened 2020-11-06
Document and enforce naming convention for instrumentation names


mateuszrzeszutek:
(The ones passed in `Instrumenter`/`InstrumentationModule` constructor)

Sometimes those names include library versions (e.g. netty), sometimes not (cassandra), sometimes the same instrumentation name is repeated in several modules (sometimes unrelated, e.g. apache-httpclient and java-httpclient have the same name `httpclient`), sometimes it's the name of our module, sometimes it's the name of the library jar. `scala-concurrent` instrumentation is named `java_concurrent`, some Redis clients use `redis` (but not all of them), etc.

iNikem:
We don't have anything formal yet. But you are right, we should have.

That name is used for instrumentation library identification and for enabling/disabling it. So probably it should be the same as Gradle module's name. This address the question of "with of without version": if module is unique for a given library, it does not have a version in its name. If module depends on library version, its name does have version.

mateuszrzeszutek:
That's a good start. What do you think about the following naming rules:

* Instrumentation names should consist of hyphen-separated words, e.g. `instrumented-library`
* Instrumentation names should be the same as the gradle module
  * Except for instrumentations that are optional, disabled by default e.g. `servlet-filter`, `jdbc-datasource`
* library + version can be the secondary/additional instrumentation name.

What do we do about e.g. `redisson-3.0` module though? This is the only Redisson module, but it has version in its name. Should we name the instrumentation `redisson` or `redisson-3.0`?

Similar question about spring: `spring-webmvc-3.1` is the only module for that spring library, what kind of names should it have?

trask:
👍

I would suggest change from:
* library + version can be the secondary/additional instrumentation name

to:
* library without version should be the second/additional instrumentation name if the gradle module name has a version suffix

iNikem:
I think that in any case the main name of the instrumentation library SHOULD NOT contain version. We expect that only one of e.g. lettuce instrumentation modules will be applied to any given runtime instance. And users can expect that they can just use `otel.instrumentation.lettuce.enabled=false` regardless of the version used.

At the same time, if module does have a version in its name, then this SHOULD be the additional name for that instrumentation library.

Other than that, instrumentation name should be as close to module name as possible, ideally identical (minus versions as described above)

trask:
Yes to what @iNikem said 👍

iNikem:
[...]

Issue #1374, opened 2020-10-12
Make muzzle generate helperClassNames() method


mateuszrzeszutek:
I have an idea for improvement in muzzle & `Instrumenter.Default`: muzzle already collects all references to our internal helper classes when the ByteBuddy plugin executes - and it knows which classes are "our" classes and which ones are external (see `ReferenceCreationPredicate`). I believe that we can make use of this:
* we can gather all helper classes in `ReferenceCreator` and references to other helper classes - just a simple graph, even `Map<String, Set<String>>` would be fine;
* [sort the graph vertices](https://en.wikipedia.org/wiki/Topological_sorting) so that the classes can be injected in the correct order;
* make `MuzzleVisitor` generate one more method that just returns an array of strings (`getMuzzleHelperClassNames`?).

This way we won't have to write `helperClassNames()` manually for most of our instrumentations, since we usually inject classes that are in the conventional packages (i.e. `io.opentelemetry.javaagent.instrumentation.*`). There are two exceptions that I know of that inject `rx.__OpenTelemetryTracingUtil` and a shaded guava class (`KubernetesClientInstrumentation`), but for those we can just leave an `additionalHelperClasses()` method that those instrumentations can override and supply those unusual helpers.
For any other instrumentation the whole helper class injection process will be 100% automatic (same as in "normal Java") and will make using lambda functions viable (since muzzle will pick up the auto-generated `$1` classes by itself).
(Also it will make errors in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1274 disappear - although it's worth noting that the underlying problem does not get fixed this way)

WDYT?

iNikem:
I like it 👍

trask:
👍 I like that idea of stopping at some package name boundary in order to protect against accidentally (and silently) pulling in too much, and using `additionalHelperClasses()` for those exceptional cases

anuraaga:
I remember @trask's idea of something similar to `spotlessApply` to automatically sync the helper class names mainly to prevent the accidental bloating. But if we can do that reliably automatically, that's even better :D

pavolloffay:
What package names will be accepted as helper classes? This might be related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1395 both issues are important for anybody extending the agent and adding instrumentations. Could we make both of them required for GA?

mateuszrzeszutek:
[...]

PR #1601, opened 2020-11-09
Add support for Reactor Netty's HttpClient

iNikem:
Newly added test fails without this instrumentation.

@mateuszrzeszutek should I re-do this as `InstrumentationModule`? Can I somehow avoid duplicating all Netty's helper classes?


mateuszrzeszutek:
> should I re-do this as `InstrumentationModule`?

Yes, definitely 😛 

> Can I somehow avoid duplicating all Netty's helper classes?

Without refactoring `AttributeKeys` - no, not right now. You won't have to list them all after https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1374 is implemented

instrumentation/reactor-netty-0.9/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor_netty/ReactorHttpClientInstrumentation.java
```diff
@@ -0,0 +1,102 @@
...
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientRequestTracingHandler",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientResponseTracingHandler",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientTracingHandler",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.NettyHttpServerTracer",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.NettyRequestExtractAdapter",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerRequestTracingHandler",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerResponseTracingHandler",
+      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerTracingHandler"
+    };
+  }
```
mateuszrzeszutek, line 68: You basically just need to move `typeMatcher()` and `transformers()` into the new `TypeInstrumentation` implementation:

```suggestion
@AutoService(InstrumentationModule.class)
public final class ReactorHttpClientInstrumentationModule extends InstrumentationModule {

  public ReactorHttpClientInstrumentationModule() {
    super("reactor-httpclient");
  }

  @Override
  public String[] helperClassNames() {
    return new String[] {
      ReactorHttpClientInstrumentation.class.getName() + "$MapConnect",
      ReactorHttpClientInstrumentation.class.getName() + "$OnRequest",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.AttributeKeys",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.AttributeKeys$1",
      // these below a transitive dependencies of AttributeKeys from above
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.NettyHttpClientTracer",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.NettyResponseInjectAdapter",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientRequestTracingHandler",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientResponseTracingHandler",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientTracingHandler",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.NettyHttpServerTracer",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.NettyRequestExtractAdapter",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerRequestTracingHandler",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerResponseTracingHandler",
      "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerTracingHandler"
    };
  }
  
  @Override
  public List<TypeInstrumentation> typeInstrumentations() {
    return singletonList(new HttpClientInstrumentation());
  }
  
  private static final class HttpClientInstrumentation implements TypeInstrumentation {
    @Override
    public ElementMatcher<TypeDescription> typeMatcher() {
      return named("reactor.netty.http.client.HttpClient");
    }

    @Override
    public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
      return singletonMap(
          isStatic().and(named("create")),
          ReactorHttpClientInstrumentation.class.getName() + "$CreateAdvice");
    }
  }
```

instrumentation/reactor-netty-0.9/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor_netty/ReactorHttpClientInstrumentationModule.java
```diff
@@ -0,0 +1,112 @@
...
+import reactor.core.publisher.Mono;
+import reactor.netty.Connection;
+import reactor.netty.http.client.HttpClient;
+import reactor.netty.http.client.HttpClientRequest;
+
+@AutoService(InstrumentationModule.class)
+public final class ReactorHttpClientInstrumentationModule extends InstrumentationModule {
+
+  public ReactorHttpClientInstrumentationModule() {
+    super("reactor-httpclient");
```
mateuszrzeszutek, line 37: How about naming it `ReactorNettyInstrumentationModule`? And using the following instrumentation names: `"reactor-netty", "reactor-netty-0.9"` (naming in accordance with our discussion in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1579)

instrumentation/reactor-netty-0.9/src/test/groovy/HttpClientTest.groovy
```diff
@@ -38,26 +40,34 @@ class ReactorNettyTest extends AgentTestRunner {
...
+      }
     }
 
+    def pool = Executors.newFixedThreadPool(4)
+
+    when:
+    pool.submit(job)
+    pool.submit(job)
+    pool.submit(job)
+    pool.submit(job)
```
mateuszrzeszutek, line 58: Really minor suggestion:

```suggestion
    4.times {
      pool.submit(job)
    }
```

mateuszrzeszutek (approved):

instrumentation/reactor-netty-0.9/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor_netty/ReactorHttpClientInstrumentationModule.java
iNikem: Yeah, it does seem logical, but... :) This is Reactor's HttpClient instrumentation, not reactor-netty-core. I think I would prefer to change module name to `reactor_httpclient` :)

iNikem: @mateuszrzeszutek @trask please help me with names :)

[Reactor Netty](https://projectreactor.io/docs/netty/release/reference/index.html) provides separate clients and servers for TCP, UDP and HTTP. Right now we instrument only HttpClient. I think we should leave module name as `reactor-netty`, because it may in the future collect instrumentation for all parts of it. But I don't know if this should be one module (and its name then should be `reactor-netty`) or it is better to have separate modules per client/server.

mateuszrzeszutek: > I think we should leave module name as `reactor-netty`

Definitely, I agree 👍 

> I don't know if this should be one module (and its name then should be reactor-netty) or it is better to have separate modules per client/server.

I think I would take the incremental approach: name the class `ReactorNettyInstrumentationModule` even if we only instrument HTTP right now; when we instrument TCP/UDP put them all in a single module; and if somebody wants to have them enabled/disabled separately then I'd split them into several `InstrumentationModule`s. 
Right now we're instrumenting a single jar/library so I think it makes sense to keep it in once module.

mateuszrzeszutek (approved):

instrumentation/reactor-netty-0.9/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor_netty/ReactorNettyInstrumentationModule.java
```diff
@@ -0,0 +1,112 @@
...
+        client = client.doOnRequest(new OnRequest()).mapConnect(new MapConnect());
+      }
+    }
+  }
+
+  public static class MapConnect
+      implements BiFunction<Mono<? extends Connection>, Bootstrap, Mono<? extends Connection>> {
+    @Override
+    public Mono<? extends Connection> apply(Mono<? extends Connection> m, Bootstrap b) {
+      return m.subscriberContext(s -> s.put("otel_context", Context.current()));
```
trask, line 101: is this attribute name shared with any other instrumentation? if not, maybe name it `reactor_netty_otel_context`?

trask (approved):

instrumentation/reactor-netty-0.9/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor_netty/ReactorNettyInstrumentationModule.java
```diff
@@ -0,0 +1,112 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import reactor.core.publisher.Mono;
+import reactor.netty.Connection;
+import reactor.netty.http.client.HttpClient;
+import reactor.netty.http.client.HttpClientRequest;
+
+@AutoService(InstrumentationModule.class)
+public final class ReactorNettyInstrumentationModule extends InstrumentationModule {
```
anuraaga, line 34: Can we add at least a small comment on what this instrumentation does? Especially since it doesn't seem to be about creating spans, but only context propagation.

instrumentation/reactor-netty-0.9/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor_netty/ReactorNettyInstrumentationModule.java
iNikem: Why, what difference does it make apart having longer key? :)

trask: when I saw this, the general name gave me the impression that this (reactor) context key was being used to propagate the (otel) context more generally for all reactor instrumentation

iNikem: Weeeeellll. That may be totally not a bad idea...

anuraaga (approved):

Issue #1578, opened 2020-11-06
Servlet Instrumentation does not cover scenarios like Spark-Jetty due to the missing instrumentation hook at Filter.doFilter()


singhalprerana:
**Is your feature request related to a problem? Please describe.**
Servlet instrumentation currently has hooks at `javax.servlet.FilterChain.doFIlter()` and `javax.servlet.http.HttpServlet.service()`. However, there might be code paths which do not go through either of these hooks -- they might involve `javax.servlet.Filter.doFIlter()` path instead -- for example, Embedded-Jetty-Spark applications.

**Describe the solution you'd like**
Add the servlet instrumentation hook at `javax.servlet.Filter.doFIlter()` too.

**Describe alternatives you've considered**
One alternative is to extend and add instrumentation hook at spark-web specific class (e.g. `spark.http.matching.MatcherFilter`) implementing `javax.servlet.Filter.doFIlter()`. However, there might be other such frameworks too and the entire instrumentation logic is exactly the same as servlet. So unless there is some potential issue which this might cause, a generic instrumentation hook might be a good idea.

mateuszrzeszutek:
There's a `Filter` instrumentation that's turned off by default: https://github.com/open-telemetry/opentelemetry-java-instrumentation#disabled-instrumentations . Have you tried turning it on (`-Dotel.integration.servlet-filter.enabled=true`)?

trask:
I'm not sure why we instrument `FilterChain` instead of `Filter`, sent draft PR #1606 to see what tests break if we change this

trask:
Should be resolved by #1606

PR #1606, opened 2020-11-09
Instrument java.servlet.Filter directly

trask:
I'm not sure why we instrument `java.servlet.FilterChain.doFilter()` instead of the more general `java.servlet.Filter.doFilter()`.

Let's see if any tests fail with this change 😂.

See #1578

### DIFF
--- a/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerMappingResourceNameFilter.java
+++ b/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerMappingResourceNameFilter.java
@@ -12,33 +12,44 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
 import java.io.IOException;
 import java.util.List;
+import javax.servlet.Filter;
 import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
 import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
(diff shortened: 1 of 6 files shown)


iNikem:
@trask what are your plans for making this real PR?

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/HandlerMappingResourceNameFilter.java
```diff
@@ -12,33 +12,44 @@
...
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.support.GenericBeanDefinition;
 import org.springframework.core.Ordered;
-import org.springframework.web.filter.OncePerRequestFilter;
 import org.springframework.web.servlet.HandlerExecutionChain;
 import org.springframework.web.servlet.HandlerMapping;
 
-public class HandlerMappingResourceNameFilter extends OncePerRequestFilter implements Ordered {
+public class HandlerMappingResourceNameFilter implements Filter, Ordered {
```
trask, line 28: can't have helper class extend `OncePerRequestFilter`, because then `OncePerRequestFilter` won't be instrumented itself

anuraaga (approved):

mateuszrzeszutek:
How does it compare to `FilterInstrumentationModule` in `servlet-common`? Can we get rid of that one?

trask:
`FilterInstrumentationModule` captures a span for every filter in the chain. I'd be happy to remove it, not sure why someone would want that. I'll submit a PR for that and see if anyone objects 😄.

PR #1585, opened 2020-11-08
Fix muzzle and add missing helper classes

trask:
Fixes #1584

PR #1557, opened 2020-11-04
Refactor Instrumenters into InstrumentationModules - D, E, F, G, H

mateuszrzeszutek:
Another part of #1274

Refactored all instrumentations whose names begin with D, E, F, G or H

PR #1607, opened 2020-11-10
Fix dropwizard instrumentation

trask:
This broke in #1557 due to removing the `helperClassNames` which (oddly but usefully) pointed to the Advice class.

But the tests didn't catch it because of #1585.

And so after both of those got merged, now master is failing (which is good).


anuraaga (approved):
Thanks!

PR #1610, opened 2020-11-10
Remove Instrumenter.Default

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1274


docs/contributing/muzzle.md
```diff
@@ -20,30 +20,32 @@ Muzzle has two phases:
...
 to verify that the class loader of the instrumented type has all necessary symbols (classes,
 methods, fields). If the `ReferenceMatcher` finds any mismatch between collected references and the
 actual application classpath types the whole instrumentation is discarded.
 
 It is worth noting that because the muzzle check is expensive, it is only performed after a match
-has been made by the `InstrumenterDefault#classLoaderMatcher()` and `Instrumenter.Default#typeMatcher()`
-matchers.
+has been made by the `InstrumentationModule#classLoaderMatcher()` and `TypeInstrumentation#typeMatcher()`
+matchers. The result of muzzle matcher is cached per each classloader, so that it is only executed
+once fo the whole instrumentation module.
```
trask, line 48:
```suggestion
matchers. The result of muzzle matcher is cached per classloader, so that it is only executed
once for the whole instrumentation module.
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -135,55 +120,37 @@ public static void assertInstrumentationMuzzled(
...
-          throw new RuntimeException(e);
+          getMuzzleMethod.setAccessible(true);
+          muzzle = (ReferenceMatcher) getMuzzleMethod.invoke(instrumentationModule);
+        } finally {
+          getMuzzleMethod.setAccessible(false);
+        }
+        System.out.println(instrumentationModule.getClass().getName());
+        for (Reference ref : muzzle.getReferences()) {
+          System.out.println(prettyPrint("  ", ref));
         }
```
trask, line 144: just checking if intentional

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -135,55 +120,37 @@ public static void assertInstrumentationMuzzled(
...
-      } else {
-        throw new RuntimeException(
-            "class "
-                + instrumenter.getClass().getName()
-                + " is not a default instrumenter. No refs to print.");
+      } catch (Exception e) {
+        System.out.println(
+            "Unexpected exception printing references for "
+                + instrumentationModule.getClass().getName());
+        throw new RuntimeException(e);
```
trask, line 149: include message in the re-thrown exception instead of sysout?

testing-common/src/test/groovy/context/FieldBackedProviderTest.groovy
```diff
@@ -5,11 +5,11 @@
...
-import static context.ContextTestInstrumentation.IncorrectCallUsageKeyClass
-import static context.ContextTestInstrumentation.IncorrectContextClassUsageKeyClass
-import static context.ContextTestInstrumentation.IncorrectKeyClassUsageKeyClass
-import static context.ContextTestInstrumentation.KeyClass
-import static context.ContextTestInstrumentation.UntransformableKeyClass
+import static ContextTestInstrumentationModule.IncorrectCallUsageKeyClass
+import static ContextTestInstrumentationModule.IncorrectContextClassUsageKeyClass
+import static ContextTestInstrumentationModule.IncorrectKeyClassUsageKeyClass
+import static ContextTestInstrumentationModule.KeyClass
+import static ContextTestInstrumentationModule.UntransformableKeyClass
```
trask, line 12: this class doesn't look like it changed package?

trask (approved):

testing-common/src/test/groovy/context/FieldBackedProviderTest.groovy
mateuszrzeszutek: 😮 
That surprised me, I don't know why would IntelliJ do this - reverted.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
mateuszrzeszutek: Sysout? Yes, this class is just a helper that's only run from the Gradle muzzle plugin.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
mateuszrzeszutek: I'll do both - as I mentioned in another comment, this class is supposed to print out what it's doing since it's a part of the Gradle plugin.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
trask: oh sorry! i missed that this is inside of `printMuzzleReferences` 😁

PR #1611, opened 2020-11-10
Update publish.gradle

iNikem:
Do not automatically publish to Bintray after upload.

Recent releases showed that `bintrayPublish` often times silently fails. Let us disable it and resort to manually clicking "Publish" link in Bintray UI.

Closes #1608


anuraaga (approved):
Seems reasonable, we should also see in parallel if there's any stabler and also automated option, e.g. paid bintray, sonatype oss, etc.

PR #1612, opened 2020-11-10
Add classloading diagram to contributing docs

pavolloffay:

cc) @trask


trask:
thanks @pavolloffay!

a couple ideas

* split the diagram into two diagrams (top half / bottom half)
* give them headings, e.g. "Initialization sequence" and "Class loader state"
* rename "Application/System classloader" to "User classloader(s)" just to be extra clear that those helpers are (always/only) injected into the system classloader

trask (approved):
Thanks @pavolloffay!

Issue #1577, opened 2020-11-06
Dynamic unload/load of instrumentations


pavolloffay:
**Is your feature request related to a problem? Please describe.**

Add/replace instrumentations at runtime.

**Describe the solution you'd like**

Have an SPI to hook into  `AgentInstaller` and get the `ResettableClassFileTransformer` but also be able to change the `AgentBuilder`.

**Describe alternatives you've considered**

None.

**Additional context**

Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/566 and slightly with related with https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1534 if we provide SPI to access ByteBuddy.

iNikem:
@pavolloffay Can you describe your use-case which requires unloading of the instrumentation? Or even loading it during runtime?

pavolloffay:
For instance provide patched instrumentations without requiring JVM to restart. Or provide new instrumentations without requiring JVM to restart.

iNikem:
Do you have an actual business need for that? I hardly can imagine a situation when I would want to dynamically load new instrumentation into running JVM... I would recommend against it :)

pavolloffay:
yes, in some enterprises it's hard to upgrade the agent once it's deployed. A capability to dynamically change instrumentations solves that.

pavolloffay:
IIRC Instana agent is able to update instrumentations at runtime. They are also using ByteBuddy internally.

iNikem:
I would like to note one thing. In our various discussions about different design or functionality decisions we often rely on our understanding of potential operational concerns. E.g. @trask bases some of this decisions on his opinion that it is easier for admins to update javaagent than for developers to change and release new code version. I usually think the opposite: code changes are cheaper and a good way to solve various problems.

So with this issue @pavolloffay brings yet another opinion on the table: javaagent updates are _very_ hard and we need more dynamic behaviour in the agent :)

I feel that we have to made our expectations about agent operational model more explicit and maybe even document them as the basis for decisions we make.

pavolloffay:
From the internal discussion with senior people (ex. AppD) I hear that updating an agent in bigger enterprises takes a lot of time. It usually has to go through multiple teams and approvals, whereas submitting a patch on the wire does not have to. It's not an ideal state but has been the experience from the field. 

[...]

Issue #1534, opened 2020-10-30
Run agent alongside other java agents - DataDog, AppDynamics, NewRelic...


pavolloffay:
**Is your feature request related to a problem? Please describe.**

I would like to exclude classes from other agents. E.g. custom build of OTEL agent should run well alongside DataDog/AppDynamics/NewRelic agents.  I have noticed that OTEL agent is instrumenting DataDog classes e.g. the Okhttp sender that reports data to the platform.

**Describe the solution you'd like**

Exclude classes e.g. `datadog.*`/`com.newrelic` from being instrumented. The exclusion pattern should be configurable.

**Describe alternatives you've considered**

`OTEL_TRACE_CLASSES_EXCLUDE` didn't work in my case. So I am not sure if this is a bug or a different feature.

```
 OTEL_EXPORTER=jaeger OTEL_TRACE_CLASSES_EXCLUDE="datadog.*"  java -javaagent:${HOME}/projects/opentelemetry/opentelemetry-java-instrumentation/javaagent/build/libs/opentelemetry-javaagent-0.10.0-SNAPSHOT-all.jar -javaagent:${HOME}/tmp/dd-java-agent.jar -jar target/demo-0.0.1-SNAPSHOT.jar 
```

I could see the DD `PUT` to the platform span

![Screenshot of Jaeger UI (1)](https://user-images.githubusercontent.com/5618424/97733214-d47a2e80-1ad7-11eb-8bc2-334a5801e875.jpg)

**Additional context**

The following code disables instrumenting of other agents.

```diff
diff --git a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
index b7ef96a73..021a7a6f7 100644
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -23,6 +23,7 @@ import io.opentelemetry.javaagent.tooling.config.ConfigInitializer;
 import io.opentelemetry.javaagent.tooling.context.FieldBackedProvider;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import java.lang.instrument.Instrumentation;
+import java.security.ProtectionDomain;
[...]

PR #1613, opened 2020-11-10
Add Bytebuddy agent builder customizer SPI

pavolloffay:
Related to #1534 
Slightly related to #1577 - it will need a different SPI to get the result of `agentBuilder.installOn`.

### DIFF
--- a/javaagent-spi/javaagent-spi.gradle
+++ b/javaagent-spi/javaagent-spi.gradle
@@ -5,4 +5,5 @@ apply from: "$rootDir/gradle/publish.gradle"
 
 dependencies {
   compileOnly deps.opentelemetrySdk
+  compileOnly deps.bytebuddy
 }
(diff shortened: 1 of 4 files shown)


iNikem:
@pavolloffay can you demonstrate your use-case of actually using this new SPI?

pavolloffay:
As mentioned in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1534 I want to exclude additional classes. We are building an agent based on OTEL and this SPI gives us control over the ByteBuddy which is an important feature to have.

pavolloffay:
@iNikem @trask I have renamed the SPI as we have discussed in the meeting.

trask:
👍

do u mind waiting on #1621 and then adding example as part of this PR?

pavolloffay:
I would prefer to get it merged and do an example separately once the example PR is in.

trask (approved):

trask:
> I would prefer to get it merged and do an example separately once the example PR is in.

created #1626 and assigned to you 😉

iNikem (approved):

Issue #1411, opened 2020-10-18
KotlinProbeInstrumentation does precisely what gRPC Javadoc says not to


asarkar:
[CoroutineContextWrapper.tracingSuspend()](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/kotlin-coroutines/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlincoroutines/KotlinProbeInstrumentation.java#L194):
```
public void tracingSuspend() {
    myTracingContext = Context.current();
    myTracingContext.detach(prevTracingContext);
}
```
Apart from saving the current context, the above code is equivalent to `Context.current().detach()`, which gRPC [Context.detach()](https://grpc.github.io/grpc-java/javadoc/io/grpc/Context.html#detach-io.grpc.Context-) asks not to do:
> **Never** use `Context.current().detach()`, as this will compromise this error-detecting mechanism.

On an aside, I noticed that gRPC context is used everywhere for propagation. This is very odd; OTEL ought to make their own. Most people wouldn't expect or want a tracing library to quietly pull in gRPC and its dependencies (Guava etc.).

anuraaga:
Thanks for reporting - I'm currently working on upgrading our sdk dependency to remove the grpc dep.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1384

I also noticed the exact same issue about kotlon as you did while doing so. This context current detach seems strange to me but currently our unit test fails without it for some reason. I've left a TODO for now to explore more why that could be happening.

asarkar:
@anuraaga I'm playing around with the Kotlin instrumentation, if I find something useful, I'll post here. Can you provide a link to the unit test so that I can see what it's doing?

anuraaga:
Thanks - the stress test "test concurrent suspend functions" fails if detaching myContext.

https://www.github.com/open-telemetry/opentelemetry-java-instrumentation/tree/master/instrumentation%2Fkotlin-coroutines%2Fsrc%2Ftest%2Fgroovy%2FKotlinCoroutineInstrumentationTest.groovy

asarkar:
[...]

PR #1617, opened 2020-11-11
Add a library module for kotlin coroutines which provides helper meth…

anuraaga:
…ods for managing context using kotlin's mechanisms.

I wonder if it's possible to replace our kotlin javaagent instrumentation too with just intercepting `launch` and adding the contextelement automatically. Though while writing this, I felt like given how native context is to kotlin, maybe we just shouldn't mess too much with it?

Related: #1411


anuraaga:
Realized in #1618 that there is a big problem in this PR that I have to implement the context element in Java instead of Kotlin to be able to inject it in the agent due to https://github.com/raphw/byte-buddy/issues/350#issuecomment-336045447

#1618 has the Java version in it. If any of this makes sense I'd probably just close this PR and focus on that one.

anuraaga:
This is in the SDK now so closing

PR #1618, opened 2020-11-11
Use Kotlin context element in agent instrumentation

### LINKED ISSUE #1411: KotlinProbeInstrumentation does precisely what gRPC Javadoc says not to

anuraaga:
Follow-on to #1617 

@johnbley @iNikem mentioned he would chat with you about the approach of using context element, including the notion that `makeCurrent` in kotlin code is illegal and shouldn't really be supported by us. Figured having this in front of your eyes will help the discussion :)

Fixes #1411

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -42,6 +42,7 @@ ext {
     opentelemetryApi            : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-api', version: versions.opentelemetry),
     opentelemetryAutoAnnotations: dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-extension-annotations', version: versions.opentelemetryAnother),
     opentelemetryContext        : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-context', version: versions.opentelemetryContext),
+    opentelemetryKotlin         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-extension-kotlin', version: versions.opentelemetry),
     opentelemetryTraceProps     : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-extension-trace-propagators', version: versions.opentelemetry),
     opentelemetrySdk            : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-sdk', version: versions.opentelemetryAnother),
     opentelemetryJaeger         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporter-jaeger', version: versions.opentelemetryOther),
--- a/instrumentation/kotlinx-coroutines/kotlinx-coroutines.gradle
+++ b/instrumentation/kotlinx-coroutines/kotlinx-coroutines.gradle
@@ -16,7 +16,7 @@ muzzle {
   }
 }
 dependencies {
-  implementation project(':instrumentation:executors')
+  implementation deps.opentelemetryKotlin
 
   compileOnly 'org.jetbrains.kotlin:kotlin-stdlib-common:1.3.72'
   compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.7'
--- a/instrumentation/kotlinx-coroutines/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlinxcoroutines/KotlinCoroutinesInstrumentationHelper.java
+++ b/instrumentation/kotlinx-coroutines/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlinxcoroutines/KotlinCoroutinesInstrumentationHelper.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.kotlinxcoroutines;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.extension.kotlin.ContextExtensionsKt;
+import kotlin.coroutines.CoroutineContext;
+
+public final class KotlinCoroutinesInstrumentationHelper {
+
+  public static CoroutineContext addOpenTelemetryContext(CoroutineContext coroutineContext) {
+    Context current = Context.current();
+    Context inCoroutine = ContextExtensionsKt.getOpenTelemetryContext(coroutineContext);
+    if (current == inCoroutine) {
+      return coroutineContext;
+    }
+    return coroutineContext.plus(ContextExtensionsKt.asContextElement(current));
+  }
+
+  private KotlinCoroutinesInstrumentationHelper() {}
+}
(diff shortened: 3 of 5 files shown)


instrumentation/kotlin-coroutines/javaagent/src/test/kotlin/KotlinCoroutineTests.kt
```diff
@@ -145,35 +147,35 @@ class KotlinCoroutineTests(private val dispatcher: CoroutineDispatcher) {
   suspend fun a(iter: Long) {
     var span = tracer.spanBuilder("a").startSpan()
     span.setAttribute("iter", iter)
-    var scope = span.makeCurrent()
-    delay(10)
-    a2(iter)
-    scope.close()
+    withContext(span.asContextElement()) {
```
anuraaga, line 150: These changes are the key to the proposal. My theory is `span.makeCurrent` is illegal code - the only reason it works here is because the agent is present, but if the agent was not present, resumed coroutines could see completely unrelated contexts, which is a disaster. The idea is that our agent is here to enable tracing, not to allow illegal code. How to "codify" that it's illegal is a tricky point though but something we will want to think about. Current idea is that https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/testing/src/main/java/io/opentelemetry/sdk/testing/context/StrictContextStorage.java#L99 can also check the stacktrace to see if it is being called from a coroutine, and not from the context element. Dunno if it'll work but if it does, I think we end up in a nice place :)

anuraaga: Also kudos to @adriancole for giving the idea of stack-walking as a potential mechanism for checking

johnbley: Does this require application authors who use auto-instrumentation and simply use (e.g.) kotlin suspend functions in their logic (without any manual instrumentation) to explicitly use (kotlin) `withContext` to pass our (otel) context?  That is, if app logic is written in kotlin and taking advantage of auto-instrumentation of inbound and outbound connections, will this still work without the app author having to take extra steps?

anuraaga: Don't think so - the agent with this code is making sure to automatically propagate the parent at the time the coroutine is launched into its context, so that parent is always available. If all they're doing is using libraries which are auto-instrumented by the agent, it should work fine without calling `withContext`. It's only when doing manual instrumentation within Kotlin that they need to use `withContext` instead of `makeCurrent`. How does that sound?

trask: I can see the confusion of this particular test, since it's a javaagent test, but also uses the library instrumentation directly.

do you think we need the nested spans to validate `launchConcurrentSuspendFunctions`? if not, we could remove `a2` and `b2` and not create any nested scopes at all, and not need to use the library instrumentation in the test.

if you think the nested scopes is a good test showing interop, maybe we can move it out into a separate test class `...InteropTest`.

trask (approved):
nice library instrumentation breakout 👍 and great point about the problem of `makeCurrent` under kotlin coroutines 😬

instrumentation/kotlin-coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/kotlincoroutines/ContextExtensions.kt
```diff
@@ -0,0 +1,38 @@
...
+import io.opentelemetry.api.trace.Span
+import io.opentelemetry.context.Context
+import io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Returns a [CoroutineContext] which will make this [Context] current when resuming a coroutine
+ * and restores the previous [Context] on suspension.
+ */
+fun Context.asContextElement(): CoroutineContext {
```
anuraaga, line 17: By the way, I'm wondering whether these should be in the SDK instead of instrumentation. Mainly since I was imagining adding to the javadoc for `makeCurrent` to not use it in Kotlin coroutines and use this instead. But that doesn't make as much sense when crossing repos.

trask: language-specific extensions to the API, e.g. `opentelemetry-extension-kotlin`? I like it

anuraaga: Sent https://github.com/open-telemetry/opentelemetry-java/pull/2066 for the library aspect of this PR

instrumentation/kotlin-coroutines/javaagent/src/test/kotlin/KotlinCoroutineTests.kt
johnbley: The nesting spans are crucial to the test - without any form of instrumentation, we were previously reporting nonsensical traces like "a causes b causes a2" and "b2 causes a".

johnbley: > . It's only when doing manual instrumentation within Kotlin that they need to use `withContext` instead of `makeCurrent`. How does that sound?

That sounds great; I have no problem making kotlin instrumentation authors work a little to preserve (otel) context.  I actually originally tried something like this (using the kotlin coroutine context system more "naturally) but was unable to make it work.  I'm glad you were able to figure it out!  l'm going to try to find some time to manually validate with a "real" kotlin app and see how this change works out in practice.

instrumentation/kotlin-coroutines/library/src/main/kotlin/io/opentelemetry/instrumentation/kotlincoroutines/ContextExtensions.kt
johnbley: I agree - this is a great approach.

johnbley (approved):

instrumentation/kotlin-coroutines/javaagent/src/test/kotlin/KotlinCoroutineTests.kt
holyjak: @johnbley #1620 links to a real, though simple kotlin app that exhibits problems with span propagation

holyjak:
Great work! 
I would like to use OT in a Kotlin app now(ish). Is there a workaround I can apply or is it best to wait a few days for this to be merged and then build the agent from `master`? Thank you!

iNikem:
@anuraaga What are your plans for making this real PR?

anuraaga:
@iNikem Going to do it after updating to snapshots :)

anuraaga:
Updated to use the SDK's kotlin extension

mateuszrzeszutek (approved):

PR #1619, opened 2020-11-11
Add AgentInstaller extensions SPI

pavolloffay:
This SPI allows to run any code in the agent's classloader. This is useful for loading (SPI) implementations (in agent classloader) of API (in bootstrap classloader) that is commonly used by instrumentation. In our case we the API is for filtering/blocking requests.

### DIFF
--- a/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/AgentInstallerExtension.java
+++ b/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/AgentInstallerExtension.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.spi;
+
+/**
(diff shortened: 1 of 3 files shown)


trask:
Discussed in SIG meeting. @pavolloffay is going to explore an idea of renaming to ComponentInstaller, with before and after hooks that get called before and after bytebuddy agent is installed, and then make TracerInstaller into a ComponentInstaller.

The before hook is needed because of:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/995e2caee0e34ab51af9dd1d4cd631b02843af75/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java#L86-L96

While most of the work should be in after hook, because we want bytebuddy agent installed as early as possible so it doesn't need to retransform (many) classes (which has performance impact, due to the retransform itself, as well as due to not being able to use FieldBackedProvider)

pavolloffay:
will be done in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1848

PR #1621, opened 2020-11-11
Add example of creating custom vendor distribution

iNikem:


examples/distro/custom/src/main/java/io/opentelemetry/demo/DemoIdGenerator.java
```diff
@@ -0,0 +1,25 @@
...
+  private static final AtomicLong spanId = new AtomicLong(0);
+
+  @Override
+  public String generateSpanId() {
+    return String.valueOf(spanId.incrementAndGet());
+  }
+
+  @Override
+  public String generateTraceId() {
+    return String.valueOf(traceId.incrementAndGet());
```
trask, line 23: same here
```suggestion
    return String.format("%032d", spanId.incrementAndGet());
```

examples/distro/custom/src/main/java/io/opentelemetry/demo/DemoIdGenerator.java
```diff
@@ -0,0 +1,25 @@
+package io.opentelemetry.demo;
```
trask, line 1: what do u think of putting everything under `com.example.javaagent` instead of `io.opentelemetry.demo`?

trask (approved):
This is really nice!! Will be great to have people add to this for any additional SPIs that are added.

examples/distro/custom/src/main/java/io/opentelemetry/demo/instrumentation/DemoServlet3Instrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+            .and(takesArgument(0, named("javax.servlet.ServletRequest")))
+            .and(takesArgument(1, named("javax.servlet.ServletResponse")))
+            .and(isPublic()),
+        DemoServlet3Instrumentation.class.getName() + "$DemoServlet3Advice");
+  }
+
+  @SuppressWarnings("unused")
+  public static class DemoServlet3Advice {
+
+    public static final String X_SERVER_ID = "X-server-id";
```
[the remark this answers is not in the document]
pavolloffay: Muzzle is not configured in this example. Since there are compile-time dependencies and can it can result in runtime errors I would document this in a readme and add links to issues.

examples/distro/build.gradle
```diff
@@ -0,0 +1,57 @@
+group 'io.opentelemetry.example'
+version '1.0-SNAPSHOT'
```
anuraaga, line 2: This might be too much personal preference, but I wonder if it's worth demonstrating a versioning pattern where the distro name is suffixed? e.g., `1.0.0-aws.1`, `1.0.0-demo.1`. It's a free way of getting the `telemetry.sdk` stuff to identify the distro and might be nice for vendors (I'm not doing it yet but am planning on it).

examples/distro/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+subprojects {
+    version = rootProject.version
+
+    apply plugin: "java"
+
+    ext {
+        versions = [
+                opentelemetry            : "0.10.0",
+                opentelemetryJavaagent   : "0.10.0",
+                opentelemetryJavaagentSpi: "0.10.0",
```
anuraaga, line 13:
```suggestion
                opentelemetryJavaagent   : "0.10.1",
                opentelemetryJavaagentSpi: "0.10.1",
```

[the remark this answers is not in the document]
anuraaga: I guess best practice is for us to publish a BOM :)

examples/distro/smoke-tests/src/test/java/com/splunk/opentelemetry/OkHttpUtils.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright Splunk Inc.
```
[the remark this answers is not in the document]
anuraaga: Don't think this example has it :) We should remove copyright headers from sample code.

anuraaga (approved):

examples/distro/build.gradle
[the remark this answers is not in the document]
iNikem: This remains from my first attempt which used snapshots. Will remove.

examples/distro/custom/src/main/java/io/opentelemetry/demo/instrumentation/DemoServlet3Instrumentation.java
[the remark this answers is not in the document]
iNikem: I don't understand the concern, tbh :) Do you mean static field in Advice?

examples/distro/build.gradle
iNikem: Take a look at `Implementation-Version` in `agent/build.gradle`. I totally agree with you on `telemetry.sdk`, but I think that jar built from the project should have just "local" version. Vendor name and otel version can go to manifest.

examples/distro/custom/src/main/java/io/opentelemetry/demo/instrumentation/DemoServlet3Instrumentation.java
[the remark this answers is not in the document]
trask: Oh good point about no muzzle here.

I think the only reason this example works is because javac is inlining the string constant.

If doesn't inline the string constant, this should fail, because the advice class itself lives in the AgentClassLoader, while the bytecode inside of the advice methods is injected into the instrumented class, which lives in the user class loader, and so the bytecode inside of the advice methods should not be allowed to reference a static field in the advice class, and you should get a runtime error.

I'd suggest inlining the string here, otherwise it may give the impression that this is a valid pattern (e.g. for Loggers, Tracers, other static fields on Advice).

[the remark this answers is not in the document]
iNikem: Done

pavolloffay:
a couple of suggestions for this example if we want to advertise it as a supported way of creating custom build:
* the instrumentation module should live in a separate __instrumentation__ module like in the upstream and it should have unit tests to test the custom fuctionality
* there should be ByteBuddy plugin configured to generate muzzle references 
* there should be muzzle configured to verify API compatibility for the instrumentation

iNikem:
> * the instrumentation module should live in a separate **instrumentation** module like in the upstream

Why is this important?

iNikem:
Also, @pavolloffay, what do you think about addressing your points in future PRs? At least we cannot do proper muzzle support without #1225, IMO

pavolloffay:
> Why is this important?

It's important from the testing perspective, I think users would like to unit test instrumentations like we do right now.

> Also, @pavolloffay, what do you think about addressing your points in future PRs? At least we cannot do proper muzzle support without #1225, IMO

We should at leat document this in the example readme and point to the issue.

iNikem:
Unit tests for instrumentations are certainly important. I am wondering why "a separate instrumentation module" is important

pavolloffay:
In case you want to have multiple instrumentations and test them in isolation? There might be multiple ways to do it, but following the core repo might be good as well.

breedx-nr:
> This is really nice!! Will be great to have people add to this for any additional SPIs that are added.

Would you see additions like `SpanExporterFactory` and `MetricExporterFactory` examples in a follow-up PR?  Should I make an issue to track that?

trask:
> > This is really nice!! Will be great to have people add to this for any additional SPIs that are added.
> 
> Would you see additions like `SpanExporterFactory` and `MetricExporterFactory` examples in a follow-up PR? Should I make an issue to track that?

Yes that would be great 👍

trask:
> I am wondering why "a separate instrumentation module" is important

I think we'll want this once we have muzzle added to the example, since muzzle needs each instrumentation in its own module, so makes sense to me to break it out now, but also happy to merge this PR as-is so that others can start contributing to this effort

examples/distro/custom/build.gradle
```diff
@@ -1,37 +1,30 @@
...
+    relocate("io.opentelemetry.instrumentation.api", "io.opentelemetry.javaagent.shaded.instrumentation.api")
 
-        // relocate OpenTelemetry API
-        relocate("io.opentelemetry.api", "io.opentelemetry.javaagent.shaded.io.opentelemetry.api")
-        relocate("io.opentelemetry.context", "io.opentelemetry.javaagent.shaded.io.opentelemetry.context")
-    }
+    // relocate OpenTelemetry API
+    relocate("io.opentelemetry.api", "io.opentelemetry.javaagent.shaded.io.opentelemetry.api")
+    relocate("io.opentelemetry.context", "io.opentelemetry.javaagent.shaded.io.opentelemetry.context")
+  }
```
iNikem: I don't like at all that this config is currently copy-pasted around. In the future we HAVE to address this.

iNikem:
@pavolloffay example of unit-testing should come in a follow-up, as it depends on #1644

examples/distro/README.md
```diff
@@ -0,0 +1,25 @@
+## Introduction
+
+This repository serves as a collection of examples of extending functionality of OpenTelemetry Java instrumentation agent.
+It demonstrates how to repackage the aforementioned agent adding custom functionality.
+For every extension point provided by OpenTelemetry Java instrumentation, this repository contains an example of
+its usage.
+
+## General structure
+
+This repository has three main submodules:
```
mateuszrzeszutek, line 10: Four submodules, now that we have `instrumentation`?

examples/distro/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Extensions examples
+
+* [DemoIdGenerator](custom/src/main/java/io/opentelemetry/demo/DemoIdGenerator.java) - custom `IdGenerator`
+* [DemoPropagator](custom/src/main/java/io/opentelemetry/demo/DemoPropagator.java) - custom `TextMapPropagator`
+* [DemoPropertySource](custom/src/main/java/io/opentelemetry/demo/DemoPropertySource.java) - default configuration
+* [DemoSampler](custom/src/main/java/io/opentelemetry/demo/DemoSampler.java) - custom `Sampler`
+* [DemoSpanProcessor](custom/src/main/java/io/opentelemetry/demo/DemoSpanProcessor.java) - custom `SpanProcessor`
+* [DemoSpanExporter](custom/src/main/java/io/opentelemetry/demo/DemoSpanExporter.java) - custom `SpanExporter`
+* [DemoServlet3Instrumentation](custom/src/main/java/io/opentelemetry/demo/instrumentation/DemoServlet3Instrumentation.java) - additional instrumentation
```
mateuszrzeszutek, line 25:
```suggestion
* [DemoServlet3Instrumentation](instrumentation/servlet-3/src/main/java/io/opentelemetry/demo/instrumentation/DemoServlet3Instrumentation.java) - additional instrumentation
```

examples/distro/smoke-tests/src/test/resources/logback.xml
```diff
@@ -0,0 +1,17 @@
...
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <root level="INFO">
+    <appender-ref ref="console"/>
+  </root>
+
+  <logger name="io.opentelemetry" level="debug"/>
+  <logger name="com.splunk" level="debug"/>
```
mateuszrzeszutek, line 15:
```suggestion
  <logger name="com.example.javaagent" level="debug"/>
```

😄

examples/distro/smoke-tests/src/test/java/com/example/javaagent/smoketest/SpringBootSmokeTest.java
```diff
@@ -0,0 +1,50 @@
...
+    String currentAgentVersion =
+        (String) new JarFile(agentPath)
+            .getManifest()
+            .getMainAttributes()
+            .get(Attributes.Name.IMPLEMENTATION_VERSION);
+
+    Response response = client.newCall(request).execute();
+    System.out.println(response.headers().toString());
+
+    Collection<ExportTraceServiceRequest> traces = waitForTraces();
```
mateuszrzeszutek, line 35: What do you think about adding `// given // when // then` comments?

malafeev:
it could be a  challenge for users (for me as well :) to translate gradle scripts to maven.

trask:
> it could be a challenge for users (for me as well :) to translate gradle scripts to maven.

I highly recommend using gradle for custom vendor distributions, at least until we have a brave contributor who documents the equivalent setup for maven. (and I ❤️ maven)

pavolloffay:
I am also not sure how our gradle plugins (muzzle, bytecode instrumentation) would work in maven. It wouldn't probably.

mateuszrzeszutek (approved):

iNikem:
I propose to merge this as a "good enough for now" and to continue improving these examples in the follow-ups

trask:
agree, this is awesome 👍

PR #1627, opened 2020-11-13
Update to Gradle 6.7 and use new toolchains feature for testing on Ja…

### LINKED ISSUE #891: Confusing how to run tests for modules with maxJavaVersionForTests = JavaVersion.VERSION_1_8

anuraaga:
…va versions.

Also fixes compatibility with codeNarc and Java 14+

Fixes #891


gradle/java.gradle
```diff
@@ -218,62 +190,18 @@ def isJavaVersionAllowed(JavaVersion version) {
...
-        }
-      }
 
-      task(taskName) {
-        if (project.tasks.findByName(testTaskName) != null) {
-          dependsOn testTaskName
-        }
-      }
-    }
+  if (rootProject.findProperty('testAdditionalJavaVersions') == 'true') {
```
anuraaga, line 200: I added this like the `-java` repo has since it doesn't hurt. Here, the tradeoff of test vs build time is so high that it probably doesn't make sense to take advantage of "build once, test many" that we do there, and instead continue to matrix in GitHub resulting in "build + test many" which is fine.

build.gradle
```diff
@@ -48,10 +48,6 @@ repositories {
 
 description = 'OpenTelemetry instrumentations for Java'
 
-wrapper {
-  distributionType = Wrapper.DistributionType.ALL
```
anuraaga, line 52: We're using `-bin` in `-java` per recommendation from Gradle team I believe

gradle/java.gradle
iNikem: I don't see where do we specify this property?

gradle/java.gradle
```diff
@@ -297,6 +224,12 @@ tasks.withType(Test).configureEach {
   testLogging {
     exceptionFormat = 'full'
   }
+
+  if (!isJavaVersionAllowed(JavaVersion.toVersion(11))) {
```
iNikem, line 228: Can you add comment explaining this?

gradle/java.gradle
anuraaga: Thought it could be useful from command line but on second thought guess not so much.

gradle/java.gradle
anuraaga: Yeah - also made sure it only applies to test task which I guess is more appropriate.

testing-common/testing-common.gradle
```diff
@@ -67,6 +67,8 @@ tasks.register("testDisabledFieldInjection", Test) {
   includes = ["context/FieldBackedProviderFieldInjectionDisabledTest.class"]
 }
 test.dependsOn(testDisabledFieldInjection)
-test.forkEvery 1
+tasks.withType(Test) {
```
anuraaga, line 70: Noticed that since this approach doesn't override test tasks like the previous one did, some configs get more complicated. I think it's rare so it might be ok but I can go back to the approach of overriding test instead of defining new tasks if it makes more sense.

anuraaga: Found another file that was affected and realized it's probably hard to reason about right now, went back to the approach of using a rule.

.github/workflows/nightly.yaml
```diff
@@ -51,15 +51,13 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - name: Set up JDK ${{ matrix.java }} for running tests
+      - id: setup-test-java
+        name: Set up JDK ${{ matrix.java }} for running tests
         uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
-      - name: Set JDK ${{ matrix.java }} home
-        run: echo JAVA_${{ matrix.java }}_HOME=${{ env.JAVA_HOME }} >> $GITHUB_ENV
```
trask, line 59: no more `JAVA_X_HOME`, yay!

docs/contributing/running-tests.md
```diff
@@ -22,14 +22,9 @@ instrumented library.
...
-testJava7` (or `testJava11` etc). Then Gradle task
-rule will kick in and do the following:
-
-- check, if Gradle already runs on a java with required version
-- if not, look for an environment variable named `JAVA_N_HOME`, where `N` is the requested java version
-- if Gradle could not found requested java version, then build will fail
-- Gradle will now find all corresponding test tasks and configure them to use java executable of the requested version.
+We run all tests on Java 11 by default, along with Java 8 and 15. To run on the later, use
+`./gradlew testJava8` or `./gradlew testJava15`. If you don't have a JDK of these versions
+installed, Gradle will automatically download it for you.
```
trask, line 27: this is so cool

gradle/java.gradle
```diff
@@ -218,28 +193,15 @@ def isJavaVersionAllowed(JavaVersion version) {
...
-/**
- * Returns java home for a given version or {@code null} if not found
- */
-String findJavaHome(JavaVersion version) {
-  def javaHome = System.getenv("JAVA_${version.majorVersion}_HOME")
-  if (javaHome == null) {
-    return null
+// We default to testing with Java 11 for most tests, but some tests don't support it, where we change
+// the default test task's version so commands like `./gradlew check` can test all projects regardless
+// of Java version.
```
trask, line 198: nice

trask (approved):
😎

gradle/java.gradle
iNikem: @anuraaga Have you tested that this interplays correctly with test rule below? I think that when java version is requested explicitly by "testJava15", then we should not silently decrease it to satisfy "maxJavaVersion".

anuraaga: Yeah that's the main reason I added it here, above the test rule. Though I don't think we have any situation to trigger ambiguity either because we have no case where minJavaVersionForTests is higher than 11. Even if we added it, it should work though.

iNikem (approved):

PR #1629, opened 2020-11-13
Add ComponentInstaller SPI and use it for OpenTelemetry SDK

pavolloffay:
Supersedes #1619 

@trask @iNikem  opening draft for now as it needs some cleanup. Let me know if it loos good and I will clean it up.

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
@@ -43,12 +43,19 @@ public class AgentInitializer {
     log = LoggerFactory.getLogger(AgentInitializer.class);
   }
 
+  // fields must be managed under class lock
+  private static ClassLoader AGENT_CLASSLOADER = null;
+
   // called via reflection from OpenTelemetryAgent
+  public static void initialize(Instrumentation inst, URL bootstrapURL) {
+    startAgent(inst, bootstrapURL);
+  }
+
   private static synchronized void startAgent(Instrumentation inst, URL bootstrapUrl) {
     try {
-      ClassLoader agentClassLoader = createAgentClassLoader("inst", bootstrapUrl);
+      AGENT_CLASSLOADER = createAgentClassLoader("inst", bootstrapUrl);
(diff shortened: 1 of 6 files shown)


pavolloffay:
@trask your comment from #1619 

> While most of the work should be in after hook, because we want bytebuddy agent installed as early as possible so it doesn't need to retransform (many) classes (which has performance impact, due to the retransform itself, as well as due to not being able to use FieldBackedProvider)

When are actually the instrumentations applied to the classes? Is it on https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java#L111-L130 or `AgentBuilder.installOn` https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java#L160 (I thought it is on the `installOn`)

trask:
> When are actually the instrumentations applied to the classes? (I thought it is on the `installOn`)

ya, inside of `installOn` it will iterate over the already loaded classes, and apply instrumentation (retransform) as needed

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -247,7 +144,7 @@ private static boolean isDebugMode() {
    *
    * @return true if we detect a custom log manager being used.
    */
-  private static boolean isAppUsingCustomLogManager() {
+  public static boolean isAppUsingCustomLogManager() {
```
trask, line 147: can this method be moved to `AgentInstaller` also?

trask (commented):
It's looking good to me.

I like the moving of code from AgentInitializer (bootstrap class loader) to AgentInstaller (agent class loader), great to get out of the bootstrap class loader and do as much as possible in the agent class loader 👍

pavolloffay:
There is smth wrong tests are failing but I am able to run the agent on my test app.

pavolloffay:
will be done in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1848

PR #1630, opened 2020-11-13
Collect WebLogic-specific span attributes

agoallikmaa:
This adds collection of WebLogic-specific attributes to servlet spans, along with `middleware.name` and `middleware.version` attributes that can be used for other application servers or servlet containers as well.

Sample output from the logging collector:
```
Attributes:
     -> middleware.name: STRING(WebLogic Server)
     -> middleware.version: STRING(12.2.1.4.0)
     -> middleware.weblogic.server: STRING(admin-server)
     -> middleware.weblogic.application: STRING(weblogic)
     -> middleware.weblogic.domain: STRING(domain1)
```

As there is no reason to create the spans at WebLogic instrumentation level (no custom logic required), this adds an instrumentation that only collects the attribute values, stores them as a map into `otel.middleware` `ServletRequest` attribute. The map contents are copied over to the span by `ServletHttpServerTracer` later when the generic servlet instrumentation is triggered.

WebLogic classes are not published as a public dependency, therefore its classes are accessed with `Class.forName` and `MethodHandle` instead of compiling against its classes directly. These are resolved by `WebLogicEntity` during its static initialization, which also provides wrapper classes for WebLogic classes. As an optimization, the full set of attributes is not collected on every request, but they are cached to a servlet context attribute.

### DIFF
--- a/instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
+++ b/instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
@@ -15,6 +15,7 @@ import io.opentelemetry.instrumentation.api.tracer.HttpServerTracer;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.security.Principal;
+import java.util.Map;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


devinsba:
I don't know that I would call WebLogic the middleware, wouldn't something like Spring MVC or Jersey be more of a middleware? I would consider this to be more like the servlet implementation name

trask:
hi @agoallikmaa! we are only adding semantic conventions that have been defined in the [specification repo](https://github.com/open-telemetry/opentelemetry-specification) first. can you open an issue in the spec repo to discuss adding these?

iNikem:
@trask I believe open-telemetry/opentelemetry-specification#1143 is the corresponding issue.

PR #1635, opened 2020-11-14
Prepare a single test bootstrap jar at top level.

### LINKED ISSUE #981: testBootstrapJar is not cacheable
Gradle says "Caching has not been enabled for the task." for `testBootstrapJar` . Although usually jar tasks are not cacheable in Gradle, we have a lot of them. Each takes couple of seconds, quite a lot of time in total. Can it be made cacheable?

anuraaga:
Currently there are only two types of bootstrap jars, one with servlet classes, one without, but we generate one for each project in the build. ~~It seems harmless to have just one jar which includes the servlet classes too.~~ Instead, we can exclude servlet-related classes from logback when creating a bootstrap jar and use the same one for all of them.

Fixes #981

### DIFF
--- a/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
+++ b/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
@@ -33,6 +33,7 @@ public class AutoInstrumentationPlugin implements Plugin<Project> {
               File testingBootstrapJar =
                   new File(
                       project.getRootDir(), "testing-bootstrap/build/libs/testing-bootstrap.jar");
+              // Make sure tests get rerun if the contents of the testing-bootstrap.jar change
               task.getInputs().property("testing-bootstrap-jar", testingBootstrapJar);
               task.getJvmArgumentProviders().add(new InstrumentationTestArgs(testingBootstrapJar));
             });
--- a/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
+++ b/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
@@ -5,21 +5,13 @@
 
 package io.opentelemetry.instrumentation.gradle;
 
-import static java.util.stream.Collectors.toList;
-
 import java.io.File;
 import java.util.Arrays;
-import java.util.concurrent.Callable;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
-import org.gradle.api.artifacts.Configuration;
-import org.gradle.api.file.DuplicatesStrategy;
 import org.gradle.api.plugins.JavaLibraryPlugin;
-import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.tasks.Internal;
-import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.testing.Test;
-import org.gradle.jvm.tasks.Jar;
 import org.gradle.process.CommandLineArgumentProvider;
 
 /**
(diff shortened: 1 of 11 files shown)


buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
```diff
@@ -29,106 +21,20 @@
 // TODO(anuraaga): Migrate more build logic into this plugin to avoid having two places for it.
```
trask, line 21: does this PR close this TODO?

testing-bootstrap/testing-bootstrap.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+  exclude {
+    // Tomcat tests require these classes from servlet API on the bootstrap classpath. They should
+    // be harmless for other tests. Note that we cannot include everything in javax.servlet because
+    // we expect to instrument them in the app classloader.
+    if (it.path.startsWith("javax/servlet") && !it.name.startsWith("ServletContext") && !it.name.startsWith("ServletContainerInitializer")) {
+      return true
+    }
+    return false
+  }
```
trask, line 33: oh, I know this problem(!)

```suggestion
  // need to exclude these logback classes from the bootstrap jar, otherwise tomcat will find them
  // and try to load them from the bootstrap class loader, which will fail with NoClassDefFoundError
  // since their super classes are servlet classes which are not in the bootstrap class loader
  exclude "ch/qos/logback/classic/servlet/LogbackServletContainerInitializer.class"
  exclude "ch/qos/logback/classic/servlet/LogbackServletContextListener.class"
  exclude "META-INF/services/javax.servlet.ServletContainerInitializer"
```

testing-bootstrap/testing-bootstrap.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+dependencies {
+  implementation project(":instrumentation-api")
+  implementation project(":javaagent-bootstrap")
+
+  implementation deps.opentelemetryApi
+  implementation deps.opentelemetryContext
+  implementation deps.slf4j
+  implementation "ch.qos.logback:logback-classic:${versions.logback}"
+  implementation "javax.servlet:javax.servlet-api:3.1.0"
```
trask, line 19: see below
```suggestion
```

trask (approved):
nice, this is so much simpler!

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
```diff
@@ -29,106 +21,20 @@
...
-              task.dependsOn(bootstrapJar);
-              task.getJvmArgumentProviders()
-                  .add(
-                      new InstrumentationTestArgs(
-                          new File(project.getBuildDir(), "libs/" + bootstrapJarName)));
+              task.dependsOn(":testing-bootstrap:shadowJar");
+              File testingBootstrapJar =
+                  new File(
+                      project.getRootDir(), "testing-bootstrap/build/libs/testing-bootstrap.jar");
+              task.getInputs().property("testing-bootstrap-jar", testingBootstrapJar);
```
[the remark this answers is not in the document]
anuraaga: Good callout, added a comment

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/AutoInstrumentationPlugin.java
anuraaga: Don't think so - I previously used a Java plugin for this since it was complicated, and wanted to share the `Constants.java` list of classes here with copy-paste. Now I'll probably move this into the `.gradle` file in a followup since it's barely any code.

testing-bootstrap/testing-bootstrap.gradle
anuraaga: Makes much more sense than my randomness!

anuraaga:
Unfortunately this change is failing reliably on CI in the akka tests but I can't repro locally, even with `clean` and `--no-build-cache` 😿

trask:
> Unfortunately this change is failing reliably on CI in the akka tests but I can't repro locally, even with `clean` and `--no-build-cache` 😿

I can submit that change in a separate PR after this, hopefully easier to troubleshoot it in isolation

anuraaga:
@trask Which change? 

By the way, while akka commonly causes grief, I think in this case it's just that they happen to be running first. So something wrong with this change only in CI. Currently I'm trying to run on `ubuntu` in docker to see if I can get anything similar locally.

trask:
oh ya, makes sense 🤞

testing-bootstrap/testing-bootstrap.gradle
```diff
@@ -0,0 +1,31 @@
...
+apply from: "$rootDir/gradle/java.gradle"
+
+// Depend on all libraries that are in the bootstrap classloader when running the agent. When
+// running tests, we simulate this by adding the jar produced by this project to the bootstrap
+// classpath.
+
+dependencies {
+  implementation project(":instrumentation-api")
+  implementation project(":javaagent-bootstrap")
+  implementation project(path: ":opentelemetry-sdk-shaded-for-testing", configuration: "shadow")
```
anuraaga, line 14: @trask Interestingly none of the instrumentation tests failed with this off the bootstrap, but `AgentTestRunner` which verifies the test classpath explicitly did. Do you know if anything we've done here has removed the requirement for this to be on the bootstrap?

trask: I don't think it needs to be on the bootstrap, could have been an earlier confusion of mine, check out commit I pushed to this PR below

Issue #1514, opened 2020-10-28
Enable checkstyle for google style guide


trask:
We follow Google Java Style Guide (https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/style-guideline.md), so we should add checkstyle enforcement for the same (e.g. https://checkstyle.sourceforge.io/google_style.html), which will be nice to avoid style nit reviews in PRs.

I'm guessing we may need to relax the base google java checkstyle in a few places due to weirdness that we need to do for bytecode instrumentation.

As a bonus goal, let's try to sync the checkstyle as much as possible between both Java repos.

iNikem:
But spotless already checks for code style and fails PR if violated

trask:
It didn't catch https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1513#discussion_r513743260

(https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used)

trask:
Having checkstyle enabled would also allow us to do other things: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1392#issuecomment-714289045

iNikem:
Ok, convinced 👍

trask:
Closed by #1641, #1656, and #1666

PR #1641, opened 2020-11-16
Enable checkstyle for google style guide

trask:
Most of #1514

I'll send another PR afterwards to enable `OverloadMethodsDeclarationOrder` as it requires moving a few methods around which makes for hard to review diffs.

EDIT: also will send separate PR to enable/discuss `ClassTypeParameterName` and `MethodTypeParameterName`


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/DatabaseClientTracer.java
```diff
@@ -18,7 +18,7 @@
 import java.net.InetSocketAddress;
 import java.util.concurrent.ExecutionException;
 
-public abstract class DatabaseClientTracer<CONNECTION, QUERY> extends BaseTracer {
+public abstract class DatabaseClientTracer<ConnectionT, QueryT> extends BaseTracer {
```
iNikem, line 21: Just a random pick: not sure I like this change :(

instrumentation/logback/logback-1.0/javaagent/logback-1.0.0-javaagent.gradle
```diff
@@ -9,11 +9,11 @@ muzzle {
 }
 
 dependencies {
-  implementation project(':instrumentation:logback:logback-1.0.0:library')
+  implementation project(':instrumentation:logback:logback-1.0:library')
```
mateuszrzeszutek, line 12: Hm, shouldn't you rename this file to `logback-1.0-javaagent.gradle` as well?

mateuszrzeszutek (approved):
Nice! 🎉  I love CamelCase consistency in all class/method names.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/DatabaseClientTracer.java
trask: I reverted the type variable names, and will submit a separate PR to discuss those changes (I had done that change as a separate commit suspecting potential objection 😁)

trask:
Going to merge before conflicts start building up. Not a problem to revisit any of the checkstyle rules if you find you don't like something later on.

PR #1645, opened 2020-11-16
Generate opentelemetry-javaagent "-all" artifact during "assemble"

### LINKED ISSUE #708: Generate opentelemetry-javaagent "-all" artifact during "assemble"
Currently

```
./gradlew assemble
```

does not generate the `-all` artifact.

Should we generate the `-all` artifact during `assemble`? (EDIT: based on @anuraaga's feedback below: yes!)

#707 is fixing `CONTRIBUTING.md` to accurately reflect the current state, but if we make this change then we can simplify `CONTRIBUTING.md` back to just running `assemble` in the future.

iNikem:
Fixes #708

```
./gradlew clean assemble
ls -l javaagent/build/libs 

total 62344
-rw-r--r--  1 nikem  staff    29M Nov 16 14:36 opentelemetry-javaagent-0.11.0-SNAPSHOT-all.jar
-rw-r--r--  1 nikem  staff   389K Nov 16 14:35 opentelemetry-javaagent-0.11.0-SNAPSHOT-javadoc.jar
-rw-r--r--  1 nikem  staff   3.4K Nov 16 14:35 opentelemetry-javaagent-0.11.0-SNAPSHOT-sources.jar
-rw-r--r--  1 nikem  staff   4.9K Nov 16 14:35 opentelemetry-javaagent-0.11.0-SNAPSHOT.jar
```

### DIFF
--- a/javaagent/javaagent.gradle
+++ b/javaagent/javaagent.gradle
@@ -123,3 +123,5 @@ tasks.withType(Test).configureEach {
 
   dependsOn shadowJar
 }
+
+assemble.dependsOn shadowJar
--- a/javaagent/javaagent.gradle
+++ b/javaagent/javaagent.gradle
@@ -123,5 +123,5 @@ tasks.withType(Test).configureEach {
 
   dependsOn shadowJar
 }
-
+assemble.dependsOn lightShadow
 assemble.dependsOn shadowJar
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,12 +30,6 @@ java -version
 ./gradlew assemble
 ```
 
-and then generate the -all artifact
-
-```bash
-./gradlew :javaagent:shadowJar
-```
-
 and then you can find the java agent artifact at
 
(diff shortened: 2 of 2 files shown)


mateuszrzeszutek (approved):

javaagent/javaagent.gradle
```diff
@@ -123,3 +123,5 @@ tasks.withType(Test).configureEach {
 
   dependsOn shadowJar
 }
+
+assemble.dependsOn shadowJar
```
trask, line 127: I think good to build both on assemble

interestingly, this is exactly what didn't work for me in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/708#issuecomment-674670104, but just tried and looks like working now, so yay?!

```suggestion
assemble.dependsOn lightShadow
assemble.dependsOn shadowJar
```

trask (approved):
Also can remove this section from `CONTRIBUTING.md` 👍

----------------

and then generate the -all artifact

```
./gradlew :javaagent:shadowJar
```

anuraaga (approved):

PR #1646, opened 2020-11-16
Prevent bootstrap classes from being included into agent classloader

### LINKED ISSUE #1547: agent jar has duplicated classes
```
nikem@Nikitas-iMac opentelemetry-java-instrumentation % jar tf javaagent/build/libs/opentelemetry-javaagent-0.10.0-SNAPSHOT-all.jar| grep SemanticAttributes.class
io/opentelemetry/javaagent/shaded/io/opentelemetry/api/trace/attributes/SemanticAttributes.class
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/trace/attributes/SemanticAttributes.classdata
```

iNikem:
Fixes #1547

### DIFF
--- a/instrumentation/instrumentation.gradle
+++ b/instrumentation/instrumentation.gradle
@@ -34,16 +34,13 @@ subprojects {
 }
 
 dependencies {
-  implementation project(':instrumentation-api')
-  implementation project(':javaagent-api')
-  implementation(project(':javaagent-tooling')) {
-    exclude module: ':javaagent-bootstrap'
-  }
+  compileOnly project(':instrumentation-api')
+  compileOnly project(':javaagent-api')
+  implementation(project(':javaagent-tooling'))
 }
 
 configurations {
   // exclude bootstrap dependencies from shadowJar
-  implementation.exclude module: deps.slf4j
   implementation.exclude group: 'org.slf4j'
   implementation.exclude group: 'io.opentelemetry', module: 'opentelemetry-api'
 }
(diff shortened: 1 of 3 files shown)


mateuszrzeszutek (approved):

instrumentation/instrumentation.gradle
```diff
@@ -53,6 +50,9 @@ shadowJar {
   mergeServiceFiles()
 
   exclude '**/module-info.class'
+  //These classes are added to bootstrap classloader by javaagent module
+  exclude 'io/opentelemetry/instrumentation/api/**/*.*'
+  exclude 'io/opentelemetry/javaagent/instrumentation/api/**/*.*'
```
trask, line 55: seems better (generally at least) to exclude at dependency level, can be done below:

```
  dependencies {
    exclude(project(':instrumentation-api'))
    exclude(project(':javaagent-api'))
    exclude(project(':javaagent-bootstrap'))
  }
```

anuraaga: With the dependencies changed to `compileOnly` do we need any change here?

trask: it's also getting these dependencies through `:javaagent-tooling`

instrumentation/instrumentation.gradle
```diff
@@ -34,16 +34,13 @@ subprojects {
...
 
 dependencies {
-  implementation project(':instrumentation-api')
-  implementation project(':javaagent-api')
-  implementation(project(':javaagent-tooling')) {
-    exclude module: ':javaagent-bootstrap'
-  }
+  compileOnly project(':instrumentation-api')
+  compileOnly project(':javaagent-api')
+  implementation(project(':javaagent-tooling'))
```
anuraaga, line 39: Should this also be `compileOnly`?

iNikem: `javaagent` module currently has only `:javaagent-bootstrap` in `shadowInclude` configuration. Which transitively depends on `:javaagent-api` and `:instrumentation-api`. All these three should be added to bootstrap classloader. This is done by `javaagent` module and thus these three modules should NOT come via `instrumentation` module. Everything that comes from the latter goes into `inst` folder inside jar and thus into agent classloader. `javaagent-tooling` should go there as well. Thus it is correct that `instrumentation` module includes it into its own shadowJar.

trask (approved):

PR #1648, opened 2020-11-16
Standardize instrumentation names in InstrumentationModules

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1579

* Rename gradle modules:
  * `vertx-3.0` to `vertx-web-3.0`
  * `spark-web-framework-2.3` to `spark-2.3`
* Add JavaDoc describing the naming rules to `InstrumentationModule`

I've decided to put all the rules from the #1579 discussion in the JavaDoc - `writing-instrumentation.md` file probably needs much more changes and improvements before we reference the code there.


settings.gradle
```diff
@@ -175,7 +175,7 @@ include ':instrumentation:servlet:glassfish-testing'
 include ':instrumentation:servlet:servlet-common'
 include ':instrumentation:servlet:servlet-2.2'
 include ':instrumentation:servlet:servlet-3.0'
-include ':instrumentation:spark-web-framework-2.3'
+include ':instrumentation:spark-2.3'
```
trask, line 178: `spark` makes me think Apache Spark. `spark-web-framework` name has it's problems too though. was thinking maybe `spark-web`, but that name has it's problems too. ok, I like `spark`. it seems popular enough for us to justify using that name, and Apache Spark would be `apache-spark` (like `apache-camel`) anyways if we even ever.

trask (approved):
thanks!

PR #1656, opened 2020-11-17
Enable checkstyle overload method ordering rule for Google Java Style

trask:
Enables checkstyle rule for https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents

> 3.4.2.1 Overloads: never split
When a class has multiple constructors, or multiple methods with the same name, these appear sequentially, with no other code in between (not even private members).

I don't feel strongly either way about this rule.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
@@ -143,50 +143,6 @@ public final class ReferenceMatcher {
     }
   }
 
-  // for helper classes we make sure that all abstract methods from super classes and interfaces are
-  // implemented
-  private List<Mismatch> checkHelperClassMatch(Reference helperClass, TypePool typePool) {
-    List<Mismatch> mismatches = Collections.emptyList();
-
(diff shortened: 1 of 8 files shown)


iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
```diff
@@ -143,50 +143,6 @@ private boolean doesMatch(ClassLoader loader) {
     }
   }
 
-  // for helper classes we make sure that all abstract methods from super classes and interfaces are
-  // implemented
-  private List<Mismatch> checkHelperClassMatch(Reference helperClass, TypePool typePool) {
```
mateuszrzeszutek, line 148: I think that the previous ordering made slightly more sense: methods were ordered by their usage. How about renaming the `checkMatch(Reference, TypeDescription)` method to `checkThirdPartyTypeMatch` (or something similar) without reordering them?

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
trask: 👍 that's better

PR #1657, opened 2020-11-17
Enable type var naming rule for Google Style

trask:
Enables checkstyle rule for https://google.github.io/styleguide/javaguide.html#s5.2.8-type-variable-names

> Each type variable is named in one of two styles:
> * A single capital letter, optionally followed by a single numeral (such as E, T, X, T2)
> * A name in the form used for classes (see Section 5.2.2, Class names), followed by the capital letter T (examples: RequestT, FooBarT).

Just trying this out, please provide feedback.

`Q` = query
`C` = connection
`S` = storage
`ReqT` = request (could also use `RequestT`)
`ResT` = response (could also use  `RespT` or `ResponseT`)
`CarT` = carrier (could also use `CarrierT`)

### DIFF
--- a/gradle/enforcement/checkstyle.xml
+++ b/gradle/enforcement/checkstyle.xml
@@ -216,25 +216,21 @@
       <message key="name.invalidPattern"
         value="Pattern variable name ''{0}'' must match pattern ''{1}''."/>
     </module>
-    <!--
     <module name="ClassTypeParameterName">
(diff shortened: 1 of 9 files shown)


iNikem:
Call be biased, but I like our current style much more. Current names are clearly distinguished and self-explanatory. Neither of these two proposed styles satisfy both these conditions.

iNikem (changes requested):
This is ugly style :)

trask:
Closing in favor of #1666

Issue #1625, opened 2020-11-12
MongoClientTracer can report incorrect collection name


jyemin:
**Describe the bug**
MongoClientTracer.collectionName assumes that 

* if the value of the command name field is a string, that string is always a collection name.  But there are administrative commands like [dropUser](https://docs.mongodb.com/manual/reference/command/dropUser/index.html) where this is not the case, and it may confuse users if those strings are reported as collection names.
* that the collection name is always the value of the command name field.  This is true for most commands, but in the case of [getMore](https://docs.mongodb.com/manual/reference/command/getMore/index.html) the collection name is contained in a different field.

**Steps to reproduce**
Found via code inspection

**What did you expect to see?**
I think it would be safer to uses an allowList of commands where we know for sure where the collection name is, rather than risk reporting an incorrect collection name.

**What did you see instead?**
The code assumes that if the value of the command name field is a string, that string is always the name of a collection.

**What version are you using?**
N/A

**Environment**
N/A

**Additional context**
N/A

trask:
Closed by #1662

PR #1662, opened 2020-11-17
Report only known collection names in MongoClientTracer (#1625)

jyemin:
* Use allow-list of commands that are known to use collection name as the value of the command
* Special-case getMore command, which uses a different field for the collection name


instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -130,10 +133,24 @@ private static BsonValue scrub(BsonValue origin) {
     return scrubbed;
   }
 
+  private static final Set<String> COMMANDS_WITH_COLLECTION_NAME_AS_VALUE =
+      new HashSet<>(asList(
+          "aggregate", "count", "distinct", "mapReduce", "geoSearch", "delete", "find", "killCursors",
```
jyemin, line 138: While a n allow-list is in some ways not future proof, in practice I think it's safe as the trend in MongoDB has been to use new initiating stages in aggregation pipelines to add new functionality (e.g. the `$changeStream` stage.

instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -130,10 +133,24 @@ private static BsonValue scrub(BsonValue origin) {
...
+          "findAndModify", "insert", "update", "listIndexes"));
+
   private static String collectionName(CommandStartedEvent event) {
-    BsonValue collectionValue = event.getCommand().get(event.getCommandName());
-    if (collectionValue != null && collectionValue.isString()) {
-      return collectionValue.asString().getValue();
+    if (event.getCommandName().equals("getMore")) {
+      if (event.getCommand().containsKey("collection")) {
```
jyemin, line 143: See https://docs.mongodb.com/manual/reference/command/getMore/

jyemin:
I signed the CLA just now, so I'm not sure why it's still saying that I'm not authorized

trask:
Ya, the bot doesn't update the comment, which is confusing 🙁

I can see the EasyCLA check passed tho, so u are good 👍

instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -130,10 +133,35 @@ private static BsonValue scrub(BsonValue origin) {
...
+              "distinct",
+              "mapReduce",
+              "geoSearch",
+              "delete",
+              "find",
+              "killCursors",
+              "findAndModify",
+              "insert",
+              "update",
+              "listIndexes"));
```
trask, line 150: does it make sense to include `create` in this list?

trask (commented):
thanks @jyemin!

instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
jyemin: I considered it.  It doesn't do any harm, but I left it out because in MongoDB AFAIK it's not so common to create collections explicitly, as MongoDB will create them implicitly on first insert.

But happy to add it (and `drop` as well, I guess) if there's support.

trask: sure, let's include them both 👍

jyemin: Added `create`, `drop`, and `createIndexes`

jyemin:
I can't tell why the build is failing.

trask (approved):
thx @jyemin! btw, ignore the build failures, we are having infra issues 🙁

Issue #1587, opened 2020-11-09
MongoClientTracer. UNSCRUBBED_FIELDS is incomplete


jyemin:
**Describe the bug**
MongoClientTracer. UNSCRUBBED_FIELDS is an incomplete list of commands and fields that need not be scrubbed out.  

**What did you expect to see?**
I didn't expect to see a hard-coded list of commands.   Instead, it would be both safe and more easily maintainable to simply not scrub the first key of each command, as the first key is by definition always the command name, and the value of the command name does not generally contain anything sensitive.

One exception is the "getMore" command: The value of the "getMore" field in that command is the cursor id (not all that sensitive, really).  The collection name is actually the value of the "collection" field.  

Another exception is the "dropUser" command: The value of the "dropUser" field in that command is the name of the user.  User names aren't the most sensitive data, but definitely more so than a lot of the data that _is_ being scrubbed.

**What did you see instead?**
The current list includes only "ordered", "insert", "count", "find",  and "create".  This leaves out important CRUD commands such as "update", "delete", "findAndModify", "aggregate", etc., and includes one non-command-name field  ("ordered") that is not significantly different than many others.

**What version are you using?**
I found this by code inspection of the master branch, not by using the tracer.

**Environment**
N/A

**Additional context**
I'm also not clear why it's necessary to preserve the collection name in the command, given that it's already pulled out into the `SemanticAttributes.DB_MONGODB_COLLECTION` attribute.

trask:
@jyemin thanks for reporting this, really appreciate your expertise here!

jyemin:
No worries.  I have worked with DataDog and AppDynamics in the past to make tracing work well with our drivers, and want to make sure that continues with this project.

I'm going to submit a PR to fix this.  There's a few other issues as well that I'd like to address.  At this stage of development, are you looking for small PRs, each associated with a GH issue, or is a single PR with several fixes ok?

iNikem:
If it is not overwhelmingly large, then single PR is Ok as well.

jyemin:
I just updated the description to include the example of the [dropUser](https://docs.mongodb.com/manual/reference/command/dropUser/index.html) command.  Are we ok not scrubbing user names?

trask:
Our jdbc instrumentation doesn't scrub usernames:

`ALTER USER xyz IDENTIFIED BY 'password'` ==> `ALTER USER xyz IDENTIFIED BY ?`

[...]

PR #1663, opened 2020-11-17
Improve command scrubber in MongoClientTracer (#1587)

jyemin:
* Don't scrub the command field value at all if it's a string
* Use JsonWriter to improve efficiency of the scrubber: this will be more efficient than copying to a new BsonDocument (which is a LinkedHashMap underneath), since JsonWriter will write directly to a StringWriter.  I considered doing this with ScrubbingBsonWriter class that implements org.bson.BsonWriter, but I was concerned about future proofing: while not common, MongoDB does occasionally add new BSON value types, and those do require new methods to be added to BsonWriter.  This actually happened in the 3.4 release with the addition of the Decimal128 type.  
* Truncate the query string.  If available, user JsonWriter.Builder.maxLength and JsonWriter.isTruncated to make the truncation more efficient.  Otherwise, just use substring.  Truncation is important because commands in MongoDB can be quite large: up to 48 MB for a bulk write

### DIFF
--- a/instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
+++ b/instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
@@ -22,7 +22,6 @@ import java.util.Optional;
 import org.bson.BsonArray;
 import org.bson.BsonDocument;
 import org.bson.BsonValue;
-import org.bson.json.JsonMode;
 import org.bson.json.JsonWriter;
 import org.bson.json.JsonWriterSettings;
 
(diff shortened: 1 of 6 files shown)


instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -12,18 +12,35 @@
...
+import org.bson.json.JsonWriterSettings;
 
 public class MongoClientTracer extends DatabaseClientTracer<CommandStartedEvent, BsonDocument> {
   private static final MongoClientTracer TRACER = new MongoClientTracer();
 
+  private final int maxNormalizedQueryLength;
+  private final JsonWriterSettings jsonWriterSettings;
+
+  public MongoClientTracer() {
+    this(32 * 1024);
```
jyemin, line 36: I got this default from the SqlNormalizer

instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -80,54 +97,111 @@ protected String dbConnectionString(CommandStartedEvent event) {
...
 
-  @Override
-  public String normalizeQuery(BsonDocument statement) {
-    // scrub the Mongo command so that parameters are removed from the string
-    BsonDocument scrubbed = scrub(statement);
-    return scrubbed.toString();
+  private static final Method IS_TRUNCATED_METHOD;
+
+  static {
+    IS_TRUNCATED_METHOD = Arrays.stream(JsonWriter.class.getMethods())
```
jyemin, line 103: This method was added in the 3.7 release of the driver so accessing it reflectively

instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -80,54 +97,111 @@ protected String dbConnectionString(CommandStartedEvent event) {
...
 
-  private static BsonDocument scrub(BsonDocument origin) {
-    BsonDocument scrub = new BsonDocument();
+  @Override
+  public String normalizeQuery(BsonDocument command) {
+    StringWriter stringWriter = new StringWriter(128);
+    writeScrubbed(command, new JsonWriter(stringWriter, jsonWriterSettings), true);
+    // If using MongoDB driver >= 3.7, the substring invocation will be a no-op due to use of
+    // JsonWriterSettings.Builder.maxLength in the static initializer for JSON_WRITER_SETTINGS
+    return stringWriter.getBuffer()
```
jyemin, line 147: Question for reviewers: should we append a "..." to the query or just cut it off?  Or do we need to make sure that even when truncated it remains valid JSON?

instrumentation/mongo/mongo-common/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -80,54 +97,111 @@ protected String dbConnectionString(CommandStartedEvent event) {
...
+  private static final String HIDDEN_CHAR = "?";
+
+  private static boolean writeScrubbed(BsonDocument origin, JsonWriter writer, boolean isRoot) {
+    writer.writeStartDocument();
+    boolean firstField = true;
     for (Map.Entry<String, BsonValue> entry : origin.entrySet()) {
-      if (UNSCRUBBED_FIELDS.contains(entry.getKey()) && entry.getValue().isString()) {
-        scrub.put(entry.getKey(), entry.getValue());
+      writer.writeName(entry.getKey());
+      // the first field of the root document is the command name, so we preserve its value
```
jyemin, line 158: Question for reviewers: since we already extract the collection name into the Span, is it even necessary to preserve it in the scrubbed command?

trask: preserving it seems good, e.g. table names are preserved when scrubbing sql queries

trask (approved):
👍

Issue #1436, opened 2020-10-21
Instrumentation names for javaagent vs library instrumentation


trask:
Should the instrumentation name be the same for both javaagent and library instrumentation?

E.g. `io.opentelemetry.grpc`

Or should they be different?

E.g. `io.opentelemetry.grpc` and `io.opentelemetry.javaagent.grpc`

We have the `telemetry.auto.version` resource attribute, but that will be applied to library instrumentation when running under the javaagent also, so I don't think we will be able to tell whether spans originated from javaagent instrumentation or library instrumentation in this case.

I'm not sure how important this is to know this difference though (and our goal is to suppress library instrumentation in this case #903).

If we want the instrumentation names to be different, one option would be for the javaagent to pass in named tracers when constructing the library tracers.

devinsba:
From a code convention point of view, if you were to use the same packages for both then you could use package private classes and methods where classes need to be accessed from the agent. This also has the benefit of decreasing bytecode size

mateuszrzeszutek:
Should we use the following pattern for all our instrumentation names:
```
io.opentelemetry.{instrumentedLibName}-{instrumentedLibMinVersion}
```
?

trask:
It would be nice to be able to tell the difference between library and javaagent instrumentation when they are both present (to see which one backed off for example).

mateuszrzeszutek:
Won't spans created by javaagent have the `telemetry.auto.version` resource attribute?

Also, we already have a couple of javaagent instrumentations that simply add the library interceptor/listener, the same way as how library instrumentation is supposed to be used. We don't change their instrumentation names, same value as library instrumentation is passed to the `TracerProvider`. If we want to distinguish between javaagent and library instrumentations, shouldn't we change that?

anuraaga:
Yeah I feel as if which code creates the spans is always deterministic - it might be a bit simpler to distinguish them if the names are different but don't think it's worth extra complexity (we can't expose instrumentation name as an option in library instrumentation, so any solution for sharing library instrumentation with javaagent and making the names different seems quite difficult).

trask:
> I feel as if which code creates the spans is always deterministic

I'm not clear what you mean, can you clarify?

trask:
@anuraaga @mateuszrzeszutek and I discussed this in meeting yesterday, summary:
[...]

PR #1671, opened 2020-11-18
Update instrumentation names to be consistent

trask:
* `auto` -> `javaagent` in instrumentation names
* removes version from instrumentation names where it snuck back - https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1260#issuecomment-699637803
* removes `amqp` package name from rabbitmq instrumentation

Still need to do #1436...

### DIFF
--- a/instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientTracer.java
+++ b/instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientTracer.java
@@ -68,6 +68,6 @@ public class AkkaHttpClientTracer
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.auto.akka-http";
+    return "io.opentelemetry.javaagent.akka-http";
   }
 }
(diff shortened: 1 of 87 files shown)


instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientTracer.java
```diff
@@ -68,6 +68,6 @@ protected String responseHeader(HttpResponse httpResponse, String name) {
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.auto.akka-http";
+    return "io.opentelemetry.javaagent.akka-http";
```
anuraaga, line 71: Should we drop javaagent from instrumentation names? Don't think we're using it in library-based agent instrumentation.

mateuszrzeszutek (approved):

iNikem (approved):

instrumentation/akka-http-10.0/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientTracer.java
trask: see #1436

PR #1679, opened 2020-11-18
Build and publish artifacts as a single step

iNikem:
This avoids restoring gradle cache twice, thus speeding up the workflow

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -63,17 +63,12 @@ jobs:
               # Trim whitespaces and cherrypick
               echo $word | sed 's/ *$//g' | sed 's/^ *//g' | git cherry-pick --stdin
           done
-      - uses: burrunan/gradle-cache-action@v1.5
-        with:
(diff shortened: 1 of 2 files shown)


anuraaga:
Does this make sure we don't publish some artifacts even if others fail?

iNikem:
I don't see how this changes could affect that. But I don't know, actually.

trask (approved):

anuraaga:
@iNikem Before we had a step in GitHub that built, then a step that published. So if build fails, publish doesn't happen. I think now, artifacts would get published as they're ready, even if other builds failed. Does it sound right?

anuraaga (approved):
Though I guess the key word is upload, it's not actually published. Since we have to do manual publishing right now it's not a disaster, but if we were to go back to auto publish, then it could potentially be a problem. Not sure if that's our long term direction or not.

iNikem:
Running `./gradlew build final -m` shows that build finishes before any upload is attempted.

PR #1684, opened 2020-11-19
Remove disabled-by-default servlet instrumentation

trask:
I don't see these being useful (even as opt-in). Sorry if we've discussed this before, you can reject this. (if it is rejected, there are a couple of misc cleanups caught up in this PR that I will break out)


anuraaga (approved):

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/ServletAndFilterInstrumentation.java
```diff
@@ -23,19 +23,14 @@
   @Override
   public ElementMatcher<ClassLoader> classLoaderMatcher() {
     // Optimization for expensive typeMatcher.
-    return hasClassesNamed("javax.servlet.http.HttpServlet");
+    return hasClassesNamed("javax.servlet.Filter");
```
iNikem, line 26: Why this change?

iNikem (approved):

mateuszrzeszutek (approved):

instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/ServletAndFilterInstrumentation.java
trask: mostly editorial, as it's the first one listed in the typeMatcher below

also, it's _slightly_ more specific, since Filter not introduced until servlet 2.3, though that doesn't really matter since we have `classLoaderMatcher` that already limits to 3.0+ in the InstrumentationModule

PR #1688, opened 2020-11-19
[DO NOT MERGE] Raise PR build timeout to try to get a build can of wh…

anuraaga:
…ole build


iNikem:
I don't know a single legit reason for smoke tests to take more than 30 minutes.

iNikem:
@anuraaga I think you want to increase timeout not on `smoke-test` job but on `test` job, where we have `testJava15` actually hitting this limit

iNikem:
should we close this?

PR #1689, opened 2020-11-19
Remove bootstrap jar from Test inputs

anuraaga:
I had issues while developing where updates to bootstrap jar didn't rerun tasks. But perhaps this input is too strict (maybe bootstrap jar has some minor change like a timestamp change). Since it changes so rarely it's probably ok not to add it as an input if it's not working correctly.

Just hypothesis, haven't confirmed in build scans or anything.


iNikem:
I don't understand this PR :) @anuraaga You said you had issues that tasks were not re-run. Now you want them to re-run even less?

anuraaga:
Since the bootstrap jar doesn't change much, if this input is causing cache to be broken I think we can live without it and be careful. But again haven't confirmed anything just seeing what it does. It's true that smoke tests have nothing to do with this so maybe unrelated.

anuraaga:
No effect it seems :/

PR #1687, opened 2020-11-19
Skip problematic akka-http version

trask:
Not sure if this akka-http release is still in-flight, but our build has been broken for 12+ hours, so let's skip this for now at least.

`10.1.13` is a maintenance release, so shouldn't affect `testLatestDeps`, just `muzzle`.

see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1683#issuecomment-730154798

PR #1691, opened 2020-11-19
Don't use JCenter for muzzle tasks

iNikem:
Alternative to #1687


iNikem:
Will not work :( Some libraries, e.g. khttp, are available on JCenter only

trask:
bummer

PR #1694, opened 2020-11-19
Remove scala from runtime dependencies

iNikem:
As a side effect of this, we had a whole scala language in our javaagent jar file.


mateuszrzeszutek (approved):

trask:
> As a side effect of this, we had a whole scala language in our javaagent jar file.

😱

trask (approved):

PR #1696, opened 2020-11-19
Remove typed spans

trask:
The motivation of including the auto-generated typed spans in this repo was that we thought we were going to use them. We still might, but in the meantime, it's a good chunk of classes that we need to keep updating with each SDK version.

I think a better location for typed spans is as an extension in otel-java repo (which was @thisthat's original proposal but we pushed to have them here so we could iterate on their initial development, but we just haven't had the time).


iNikem:
😿

iNikem (approved):

PR #1705, opened 2020-11-20
update SystemMetrics according to spec update

malafeev:
Updated to spec changes: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/semantic_conventions/system-metrics.md

### DIFF
--- a/instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
+++ b/instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
@@ -19,7 +19,6 @@ import oshi.hardware.NetworkIF;
 
 /** System Metrics Utility. */
 public class SystemMetrics {
-  private static final String TYPE_LABEL_KEY = "type";
 
   private SystemMetrics() {}
 
--- a/instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
+++ b/instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
@@ -32,15 +32,15 @@ public class SystemMetrics {
     meter
         .longUpDownSumObserverBuilder("system.memory.usage")
         .setDescription("System memory usage")
-        .setUnit("bytes")
+        .setUnit("By")
         .build()
         .setCallback(
             new Callback<LongResult>() {
               @Override
               public void update(LongResult r) {
                 GlobalMemory mem = hal.getMemory();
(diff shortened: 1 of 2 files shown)


trask (approved):
👍

trask:
looks like `SystemMetricsTest` needs to be updated

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
```diff
@@ -32,15 +31,15 @@ public static void registerObservers() {
...
+        .setUnit("By")
         .build()
         .setCallback(
             new Callback<LongResult>() {
               @Override
               public void update(LongResult r) {
                 GlobalMemory mem = hal.getMemory();
-                r.observe(mem.getTotal() - mem.getAvailable(), Labels.of(TYPE_LABEL_KEY, "used"));
-                r.observe(mem.getAvailable(), Labels.of(TYPE_LABEL_KEY, "free"));
+                r.observe(mem.getTotal() - mem.getAvailable(), Labels.of("state", "used"));
```
anuraaga, line 41: Should we create constants for all the labels?

anuraaga (approved):

malafeev:
test updated

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
malafeev: updated for label keys

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
```diff
@@ -19,7 +19,9 @@
 
 /** System Metrics Utility. */
 public class SystemMetrics {
-  private static final String TYPE_LABEL_KEY = "type";
+  private static final String DEVICE_LABEL_KEY = "device";
```
anuraaga, line 22: Shouldn't we do something like `private static final Labels STATE_USED = Labels.of(...)`? That way we don't have to create a labels object every time we record a value (at least I think that's what's happening right now)

malafeev: but it will not work for `Labels.of(DEVICE_LABEL_KEY, device, DIRECTION_LABEL_KEY, "transmit")`

malafeev: updated for state labels

PR #1709, opened 2020-11-20
Rename grizzly-client instrumentation

### LINKED ISSUE #1658: Rename grizzly-client instrumentation to async-http-client
It appears to just be instrumentation (and tests) of [AsyncHttpClient](https://github.com/AsyncHttpClient/async-http-client)

Also, I don't think we need to disable it by default (grizzly-server is a different story, and should be disabled by default).

trask:
Closes #1658


mateuszrzeszutek (approved):

iNikem:
I think there is some confusion here.

https://github.com/AsyncHttpClient/async-http-client uses artifactId `async-http-client` and package name `org.asynchttpclient`
https://github.com/eclipse-ee4j/grizzly-ahc uses artifactId `grizzly-http-client` and package name `com.ning.http.client`.

The instrumentation under review monitors the second one, not the first one.

iNikem (changes requested):
See my comment above

trask:
oh ya, `com.ning.http.client` was the package name prior to version `2.0`, I've updated the README to note that we only support `1.9+ (not including 2.x yet)`

iNikem:
Ok, I now understand better what is happening. But I am still not convinced this change is a good one. This instrumentation still was written for https://github.com/eclipse-ee4j/grizzly-ahc, judging from package and artifact names. Why do we want to instead say that this is about old version of https://github.com/AsyncHttpClient/async-http-client? Then we should change gradle build file and test with those artifacts, whose `artifactId=async-http-client-project`

trask:
> Then we should change gradle build file and test with those artifacts

👍 done

iNikem (approved):

Issue #1692, opened 2020-11-19
Dependencies resolution during configuration


iNikem:
Recent build scans report a warning "Dependencies were resolved during project configuration". This is a slight problem for CI, because e.g. `setup-muzzle-matrix` step, which should just print project names, takes several minutes.

By adding the following code into top-level `build.gradle` and running `./gradlew instrumentation:listInstrumentations -s` one can find that the culprit is `gradle/instrumentation.gradle` file

```
def afterEvaluation = false
gradle.projectsEvaluated {
  afterEvaluation = true
}

allprojects { project ->
  configurations.all { configuration ->
    configuration.incoming.beforeResolve {
      if (!afterEvaluation) {
        throw new Exception("Configuration $configuration.name of project $project.name is being resolved at configuration time.")
      }
    }
  }
}
```

```
  new ByteBuddyPluginConfigurator(project, sourceSets.main, pluginName, [
    project(':javaagent-tooling').configurations.instrumentationMuzzle.files,
    configurations.runtimeClasspath.files,
    sourceSets.main.output.files
  ]).configure()
```
This code above resolves all dependencies already during task configuration phase.

trask:
Closed by #1715

PR #1715, opened 2020-11-20
ByteBuddy tasks no longer resolve dependencies at config time (#1692).

agoallikmaa:
This fixes issue #1692 that caused `byteBuddy` task configuration to resolve dependencies. This was originally done to make it possible to store the classpath for that task in a task field with `Input` annotation as that requires serializable value, which an unresolved classpath is not.

Added a special implementation of `net.bytebuddy.build.gradle.Transformation` which resolves arguments for creating a `ClasspathByteBuddyPlugin`. This exposes the classpath to Gradle with `Classpath` annotation instead of `Input`, meaning that its value can be an unresolved classpath.


iNikem (approved):
Thanks!

trask (approved):
tricky stuff 😅 thanks 👍

PR #1720, opened 2020-11-20
Directly port the java-runtime-metrics module over from otel-java.

jkwatson:
Enjoy!


instrumentation/runtime-metrics/library/runtime-metrics-library.gradle
```diff
@@ -0,0 +1,7 @@
+apply from: "$rootDir/gradle/instrumentation-library.gradle"
+
+dependencies {
+  testImplementation project(':testing-common')
+  testImplementation group: 'org.assertj', name: 'assertj-core', version: '1.7.1'
```
anuraaga, line 5: Can we use newer versions of these?

anuraaga (approved):

instrumentation/runtime-metrics/library/runtime-metrics-library.gradle
jkwatson: heh. absolutely. I copied this from somewhere...so there might want to be a more wide upgrade.

trask: done

anuraaga (approved):

Issue #1728, opened 2020-11-21
Trace propagation is not working when a client span is created before HttpClient


jamal:
**Describe the bug**
Trace propagation is not working when a client span is created before the `HttpClientTracer` is invoked. This happens with instrumentations that create their own span but using an HTTP Client under the hood, such as the AWS SDK.

From reviewing the logic, this appears to be happening because `HttpClientTracer` is creating an invalid span when it detects a previously created client span in the current context. This span is then passed to the Trace Propagator which bails out when it detects that the span context is invalid. 

**Steps to reproduce**
This is not trivial to reproduce, I am working on a test case to provide in a bit.

The way I found this was by debugging the AWS SDK instrumentation to figure out why X-Ray was not capturing any data. After digging into this for a while, I found that `AwsXRayPropagator` was bailing before injecting the `X-Amzn-Trace-Id` header. But, other HTTP requests where showing this normally. I then realized that this was due to the behavior I explained above.

**What did you expect to see?**
I expected to see trace headers for the configured propagators to be included in these HTTP requests.

**What did you see instead?**
Headers are being passed in the AWS SDK requests for the standard things that the SDK populates, but the trace headers are all missing.

**What version are you using?**
`v0.10.1`

**Environment**
Compiler: OpenJDK 11
Runtime (if different from JDK above): `Oracle JRE 8`

jamal:
After digging into this issue, and my proposal solution a bit more, I found that some test cases were specifically checking against this behavior for the aws-sdk instrumentation.

The source of that is this commit: https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/ca708db5dae1e70a5909f043032bf4878348670f

This brings up a point I had not considered: adding headers to the HTTP request could break signing that some instrumentations rely on (such as the AWS SDK).

That being said, in this scenario, I think we would want to at the very least pass the `X-Amzn-Trace-Id` header with AWS SDK requests as this is the only way to enable the integration with X-Ray.

anuraaga:
Hi @jamal - thanks a lot for the investigation. I had thought we were injecting headers into the request in the AWS SDK instrumentation but I can't find it even in the git history. Indeed, especially because of the signing requirement, it's the AWS SDK instrumentation that needs to propagate, not HTTP client instrumentation. We can do so in the `modifyHttpRequest` callback of `TracingExecutionInterceptor`. Is this a change you can try making? Otherwise happy to take a stab.

[...]

PR #1729, opened 2020-11-21
Avoid sending invalid span to propagators when client span is created before HttpClientTracer

jamal:
This is a proposal for resolving #1728.

I was a bit conflicted on how this should be solved, so some feedback would be greatly appreciated. I was considering three options:

1. Update propagators to be aware of the client span: not possible since this concept only exists in the instrumentation
1. Return the client span instead of an invalid span in `HttpClientTracer#startSpan`: this would likely work but I'm afraid of what impact it could have to other parts of the instrumentation
1. Update `HttpClientTracer#startScope` to pass the expected context to propagators

I moved forward with the last approach, as it seemed to be the most isolated and least likely to negatively impact other parts of the system. With that, I simply check if a client span already exists, and set that as the active span key in the context before calling the propagators.

I also modified the logic to make sure we did not overwrite the client key with the invalid span. I imagine nothing else is interacting with this, but I imagine it could cause issues in the future if something expects this to be the first created client span in the context.

On that note, I am happy to change this approach entirely if there are better solutions that I did not consider or was not aware of. Or, ultimately, leaving this to someone else to implement if none of this makes sense.

Last note, before this I had not used Spock at all so please let me know if the tests need to be cleaned up at all.

Thanks!

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
@@ -77,8 +77,17 @@ public abstract class HttpClientTracer<REQUEST, CARRIER, RESPONSE> extends BaseT
       throw new IllegalStateException(
           "getSetter() not defined but calling startScope(), either getSetter must be implemented or the scope should be setup manually");
     }
+
+    // Avoid sending an invalid span to the propagators if a client span already exists.
+    Span clientSpan = context.get(CONTEXT_CLIENT_SPAN_KEY);
+    if (clientSpan != null) {
+      span = clientSpan;
+      context = context.with(span);
+    } else {
+      context = context.with(CONTEXT_CLIENT_SPAN_KEY, span);
+    }
+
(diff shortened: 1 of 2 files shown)


jamal:
Looks like some AWS SDK tests are failing, will have to dig into this later today.

jamal:
Actually, digging into this I found this commit:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/ca708db5dae1e70a5909f043032bf4878348670f

Which removed previous logic that injected trace headers to AWS SDK requests.

I didn't consider breaking signatures when adding headers to the HTTP request, so this is not a good solution and I will revert these changes when I get more feedback.

The commit also mentions that these are not useful, and I would like to question that. X-Ray will only record segments if the caller is instrumented (and passing `X-Amzn-Trace-Id` header). This is what I am trying to solve, so that we can correlate our instrumentation with what AWS is collecting in X-Ray. 

But, I'm not sure what the right approach here should be.
1. Would it make sense to revert the logic in that commit so that we run AWS requests through the registered propagators?
1. Or, should the AWS SDK have logic specific to only the AwsXRayPropagator (if it is enabled).

I will also update my issue report since this commit brings up a very good point I had not considered.

PR #1734, opened 2020-11-22
Add static keyword to make groovy inner classes nested

trask:


anuraaga (approved):
Just wondering is this mainly for performance?

trask:
nothing so important 😄, I just like the clue when I'm reading code, that it's not (intentionally) connected to outer class

iNikem (approved):

mateuszrzeszutek (approved):

PR #1736, opened 2020-11-23
Consistent equals implementations

trask:
Following this template:

```
  @Override
  public boolean equals(Object obj) {
    if (obj == this) {
      return true;
    }
    if (!(obj instanceof Customer)) {
      return false;
    }
    Customer other = (Customer) obj;
    return Objects.equals(id, other.id)
        && Objects.equals(firstName, other.firstName)
        && Objects.equals(lastName, other.lastName);
  }

  @Override
  public int hashCode() {
    return Objects.hash(id, firstName, lastName);
  }
```

### DIFF
--- a/instrumentation/opentelemetry-api-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/trace/ApplicationSpan.java
+++ b/instrumentation/opentelemetry-api-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/trace/ApplicationSpan.java
@@ -15,7 +15,6 @@ import application.io.opentelemetry.api.trace.SpanContext;
 import application.io.opentelemetry.api.trace.StatusCode;
 import application.io.opentelemetry.context.Context;
 import io.opentelemetry.javaagent.instrumentation.opentelemetryapi.context.AgentContextStorage;
-import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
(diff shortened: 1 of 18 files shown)


instrumentation/opentelemetry-api-1.0/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/trace/ApplicationSpan.java
```diff
@@ -140,11 +141,20 @@ public boolean isRecording() {
...
+  public boolean equals(Object obj) {
+    if (obj == this) {
+      return true;
+    }
+    if (!(obj instanceof ApplicationSpan)) {
       return false;
     }
-    return agentSpan.equals(((ApplicationSpan) other).agentSpan);
+    ApplicationSpan other = (ApplicationSpan) obj;
+    return Objects.equals(agentSpan, other.agentSpan);
```
anuraaga, line 152: I would usually not use Objects for non-nullable fields. Otherwise the non-nullness gets a bit obfuscated. What do you think?

trask: yeah, this makes sense, I'll update to use direct `equals` where that was the previous behavior at least

anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

PR #1739, opened 2020-11-23
Use namespace for attributes put into user classes

### LINKED ISSUE #1652: Use otel instrumentation namespace for string attribute keys that are visible to user applications
Inspired by https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1601#discussion_r521877777

Most of our attributes already have otel instrumentation prefix, e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6e9ccfc39dfe759338c3287bf1d749d83a729bdd/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java#L31

and

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6e9ccfc39dfe759338c3287bf1d749d83a729bdd/instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java#L37-L38

As an additional complication, we should also use different attribute keys for javaagent and library instrumentation if we are storing something that is not compatible between them (e.g. shaded/unshaded Context). Using class name (or package name), as in the second example, may be a good way to handle this, since when library instrumentation is reused in javaagent instrumentation it is shaded.

trask:
Closes #1652

### DIFF
--- a/instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
+++ b/instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
@@ -27,11 +27,7 @@ public class AttributeKeys {
   public static final AttributeKey<Context> CONNECT_CONTEXT_ATTRIBUTE_KEY =
       attributeKey(AttributeKeys.class.getName() + "connect.context");
 
-  /**
-   * This constant is copied over to
-   * io.opentelemetry.javaagent.instrumentation.ratpack.server.TracingHandler, so if this changes,
-   * that must also change.
-   */
+  // this attribute key is also used by ratpack instrumentation
   public static final AttributeKey<Context> SERVER_ATTRIBUTE_KEY =
       attributeKey(AttributeKeys.class.getName() + ".context");
 
--- a/instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
+++ b/instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
@@ -9,8 +9,6 @@ import io.netty.util.AttributeKey;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.javaagent.instrumentation.api.WeakMap;
-import io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.HttpClientTracingHandler;
-import io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerTracingHandler;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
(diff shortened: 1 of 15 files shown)


instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
```diff
@@ -26,23 +24,22 @@
...
    * io.opentelemetry.javaagent.instrumentation.ratpack.server.TracingHandler, so if this changes,
    * that must also change.
    */
   public static final AttributeKey<Context> SERVER_ATTRIBUTE_KEY =
-      attributeKey(HttpServerTracingHandler.class.getName() + ".context");
+      attributeKey(AttributeKeys.class.getName() + ".context");
 
-  // TODO understand and change to context
   public static final AttributeKey<Span> CLIENT_ATTRIBUTE_KEY =
-      attributeKey(HttpClientTracingHandler.class.getName() + ".span");
```
trask, line 42: removed `TODO` because I checked and this attribute is being used to pass Span and call end on it, no need to pass Context

instrumentation/ratpack-1.4/src/main/java/io/opentelemetry/javaagent/instrumentation/ratpack/TracingHandler.java
```diff
@@ -25,7 +25,7 @@
    */
   public static final AttributeKey<io.opentelemetry.context.Context> SERVER_ATTRIBUTE_KEY =
       AttributeKey.valueOf(
-          "io.opentelemetry.javaagent.instrumentation.netty.v4_1.server.HttpServerTracingHandler.context");
+          "io.opentelemetry.javaagent.instrumentation.netty.v4_1.AttributeKeys.context");
```
iNikem, line 28: Should this use class as well?

trask: done, this required adding a dependency on the netty instrumentation, which seems reasonable

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
```diff
@@ -26,23 +24,18 @@
             }
           };
 
-  public static final AttributeKey<Context> PARENT_CONNECT_CONTEXT_ATTRIBUTE_KEY =
-      attributeKey("io.opentelemetry.javaagent.instrumentation.netty.v4_1.parent.connect.context");
+  public static final AttributeKey<Context> CONNECT_CONTEXT_ATTRIBUTE_KEY =
```
anuraaga, line 27: For Netty 4.1, maybe we can just use it's native `AttributeKey.valueOf(Class<?>, String)`

anuraaga (approved):

iNikem (approved):

instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java
trask: 👍 I will send shortly in a follow-up PR, I think I've convinced myself we can remove the logic below at the same time, but want to propose in a separate PR for visibility

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/889bce09766d685dd7b3156eee6db46697c8b299/instrumentation/netty/netty-4.1/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AttributeKeys.java#L47-L53

trask: see #1757

PR #1741, opened 2020-11-23
Rename TypeInstrumentation#classLoaderMatcher() to classLoaderOptimization()

### LINKED ISSUE #1678: Regression in muzzle when switched to instrumentation modules
**Describe the bug**

When switching from Instrumenter to Instrumentation module in custom servlet instrumentation I had to change the muzzle fail gradle configuration - https://github.com/hypertrace/javaagent/pull/123#discussion_r526011640.

**Steps to reproduce**
Try https://github.com/hypertrace/javaagent/pull/123 with the original settings.

No changes to the muzzle when switching to instrumentation modules.

**What did you see instead?**

```
> Task :instrumentation:servlet:servlet-2.3:muzzle-AssertFail-javax.servlet-javax.servlet-api-3.0.1 FAILED
...
MUZZLE PASSED Servlet2BodyInstrumentationModule BUT FAILURE WAS EXPECTED
:instrumentation:servlet:servlet-2.3:muzzle-AssertFail-javax.servlet-javax.servlet-api-3.0.1 (Thread[Execution worker for ':' Thread 2,5,main]) completed. Took 0.033 secs.
:instrumentation:servlet:servlet-2.3:muzzle-AssertFail-javax.servlet-javax.servlet-api-3.1.0 (Thread[Execution worker for ':' Thread 2,5,main]) started.
```

0.10.0

mateuszrzeszutek:
According to this comment: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1678#issuecomment-730496363

Closes #1678

### DIFF
--- a/instrumentation/hibernate/hibernate-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/TransactionInstrumentation.java
+++ b/instrumentation/hibernate/hibernate-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/TransactionInstrumentation.java
@@ -28,7 +28,7 @@ import org.hibernate.Transaction;
 final class TransactionInstrumentation implements TypeInstrumentation {
 
   @Override
-  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
     // Optimization for expensive typeMatcher.
     return hasClassesNamed("org.hibernate.Transaction");
   }
--- a/instrumentation/hibernate/hibernate-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/TransactionInstrumentation.java
+++ b/instrumentation/hibernate/hibernate-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/TransactionInstrumentation.java
@@ -29,7 +29,6 @@ final class TransactionInstrumentation implements TypeInstrumentation {
 
   @Override
   public ElementMatcher<ClassLoader> classLoaderOptimization() {
-    // Optimization for expensive typeMatcher.
     return hasClassesNamed("org.hibernate.Transaction");
   }
 
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
@@ -19,11 +19,16 @@ import net.bytebuddy.matcher.ElementMatcher;
 public interface TypeInstrumentation {
   /**
    * A type instrumentation can implement this method to optimize an expensive {@link
-   * #typeMatcher()} - usually {@link AgentElementMatchers#implementsInterface(ElementMatcher)} or
-   * {@link AgentElementMatchers#extendsClass(ElementMatcher)}. In that case it's useful to check
-   * that the classloader contains the class/interface that is being extended.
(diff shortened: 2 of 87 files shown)


instrumentation/hibernate/hibernate-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/TransactionInstrumentation.java
```diff
@@ -28,7 +28,7 @@
 final class TransactionInstrumentation implements TypeInstrumentation {
 
   @Override
-  public ElementMatcher<ClassLoader> classLoaderMatcher() {
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
     // Optimization for expensive typeMatcher.
```
iNikem, line 32: Probably all these comments can be replaced by more verbose javadoc of the base method?

iNikem (approved):

instrumentation/hibernate/hibernate-4.0/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/TransactionInstrumentation.java
mateuszrzeszutek: Honestly I completely forgot about these comments. I'll remove them.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
```diff
@@ -19,13 +19,18 @@
...
-   * #typeMatcher()} - usually {@link AgentElementMatchers#implementsInterface(ElementMatcher)} or
-   * {@link AgentElementMatchers#extendsClass(ElementMatcher)}. In that case it's useful to check
-   * that the classloader contains the class/interface that is being extended.
+   * #typeMatcher()} - usually one that uses {@link AgentElementMatchers}, e.g. {@link
+   * AgentElementMatchers#implementsInterface(ElementMatcher)} or {@link
+   * AgentElementMatchers#extendsClass(ElementMatcher)}. Type matchers that check annotation
+   * presence or class inheritance are particularly expensive for classloaders that do not contain
+   * the base class/interface/annotation. To make this check significantly less expensive this
+   * method can be used to verify that the classloader contains the class/interface that is being
+   * extended.
```
trask, line 28: github is not letting me make this as a "suggestion":

```
   * An optimization to short circuit matching in the case where the instrumented library is not
   * even present on the class path.
   *
   * <p>Most applications have only a small subset of libraries on their class path, so this ends up
   * being a very useful optimization.
   *
   * <p>Some background on type matcher performance:
   *
   * <p>Type matchers that only match against the type name are fast, e.g. {@link
   * ElementMatchers#named(String)}.
   *
   * <p>All other type matchers require some level of bytecode inspection, e.g. {@link
   * ElementMatchers#isAnnotatedWith(ElementMatcher)}.
   *
   * <p>Type matchers that need to inspect the super class hierarchy are even more expensive, e.g.
   * {@link AgentElementMatchers#implementsInterface(ElementMatcher)}. This is because they require
   * inspecting multiple super classes/interfaces as well (which may not even be loaded yet in which
   * case their bytecode has to be read and inspected).
```

trask (approved):
👍

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
mateuszrzeszutek: This is great, thanks!

PR #1740, opened 2020-11-23
Rename otel trace properties

trask:
Closes #1697

* `otel.trace.enabled` --> `otel.javaagent.enabled`
* `otel.trace.debug` --> `otel.javaagent.debug`
* `otel.trace.config` --> `otel.javaagent.config`
* `otel.trace.runtime.context.field.injection` --> `otel.javaagent.runtime.context.field.injection` (this is surely not the last rename for this property)
* `otel.instrumentations.enabled` --> `otel.instrumentations.defaultEnabled`

Issue #1414, opened 2020-10-18
Standardize config property names


trask:
I'd like to propose that we standardize config property names:

`otel.instrumentation.<instrumentationId>.*`

where `<instrumentationId>` is the instrumentation module name minus version suffix

This will match `otel.instrumentation.<instrumentationId>.enabled` which is used to suppress the instrumentation (#1412).

mateuszrzeszutek:
I've got a few questions in regards to this:

* What about `java.concurrent` instrumentation? It uses `otel.trace.executors` and `otel.trace.executors.all` properties. It seems like it's a core/required instrumentation though so we may want to leave those names unchanged.
* There are also `otel.trace.methods`, `otel.trace.annotations` and `otel.trace.annotated.methods.exclude` that are used in external annotations instrumentation, but I suppose that we don't really want to change them.

trask:
> What about `java.concurrent` instrumentation? It uses `otel.trace.executors` and `otel.trace.executors.all` properties.

what do u think of
* `otel.instrumentation.executor.include`
* `otel.instrumentation.executor.include.all`

and renaming `java-concurrent` module to `executor`? (and we could rename `java-classloader` to simply `classloader`, I don't think either of those terms requires `java-` prefix for clarification)

> There are also `otel.trace.methods`, `otel.trace.annotations` and `otel.trace.annotated.methods.exclude` that are used in external annotations instrumentation

maybe
* `otel.instrumentation.methods.include`
* `otel.instrumentation.annotations.include`
* `otel.instrumentation.annotations.methods.exclude`

and splitting `external-annotations` into two modules, `annotations` and `methods`?

mateuszrzeszutek:
That's a great idea!

Two more things to add:
* Looks like `otel.trace.annotated.methods.exclude` is also used in the `methods` instrumentations - we should probably add a dedicated `otel.instrumentation.methods.exclude` instead. 
* This property is also used in the `@WithSpan` annotation instrumentation. I'm wondering whether we need this functionality there at all - it's a separate `InstrumentationModule`, you can turn the whole thing off if you don't need spans from the annotation. Well, if we do need it we can always rename the property.

iNikem:
> * This property is also used in the `@WithSpan` annotation instrumentation. I'm wondering whether we need this functionality there at all - it's a separate `InstrumentationModule`, you can turn the whole thing off if you don't need spans from the annotation. Well, if we do need it we can always rename the property.

[...]

PR #1747, opened 2020-11-24
Rename otel.trace.classes.excludes to otel.javaagent.exclude-classes

trask:
Missed this in #1740.

See latest note about `exclude-classes` vs `excludeClasses`: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1414#issuecomment-732632618, though happy to change to something else either in this PR or later on

### DIFF
--- a/docs/suppressing-instrumentation.md
+++ b/docs/suppressing-instrumentation.md
@@ -26,6 +26,6 @@ which could have unknown side-effects.
 If you find yourself needing to use this, it would be great if you could drop us an issue explaining why,
 so that we can try to come up with a better solution to address your need.
 
-| System property       | Environment variable  | Purpose                                                                                           |
-|-----------------------|-----------------------|---------------------------------------------------------------------------------------------------|
-| otel.trace.classes.exclude | OTEL_TRACE_CLASSES_EXCLUDE | Suppresses all instrumentation for specific classes, format is "my.package.MyClass,my.package2.*" |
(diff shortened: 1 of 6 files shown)


iNikem:
I think we should fix our property->env mapping to support `excludeAll`. This `exclude-all` is quite unnatural for Java I would say.

mateuszrzeszutek:
Hmm, we use dash-separated words in the instrumentation names though: e.g. `spring-webmvc`.
Still, I agree that `excludeAll` looks more Java-like in this case.

trask:
I have the same impression about camelCase being more Java-like, but I went looking at the [Spring Boot env var / system property mappings](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-relaxed-binding), and they support both camelCase and kebab-case, but recommend kebab-case:

> Property | Note
> -- | --
> acme.my-project.person.first-name | Kebab case, which is recommended for use in .properties and .yml files.

And also interestingly:

> The prefix value for the annotation must be in kebab case (lowercase and separated by -, such as acme.my-project.person).

So now I'm thinking kebab case is the way to go (and as benefit is consistent with our module names when used in the property names, e.g. `otel.instrumentation.external-annotations.exclude-methods`).

mateuszrzeszutek (approved):

iNikem (approved):

PR #1750, opened 2020-11-24
Don't apply unnecessary plugins at top level.

### LINKED ISSUE #1693: Strange jar file name
After recent nebula plugin version update I now see the following javaagent file name: `opentelemetry-javaagent-0.11.0-dev.48.uncommitted+80b068b-all.jar`. Can we have old name back? :)

anuraaga:
No clue how applying shadow plugin could cause a bug in nebula but they do seem incompatible right now. No worries though since we don't need shadow (and some other plugins) at the top level anyways.

There's non-zero chance this may fix our recent build slowness problem if that jar name was leaking into our up-to-date checks somehow :O

Fixes #1693

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,15 +9,15 @@ plugins {
 
   id 'org.gradle.test-retry' version '1.1.8' apply false
 
-  id 'org.unbroken-dome.test-sets' version '3.0.1'
+  id 'org.unbroken-dome.test-sets' version '3.0.1' apply false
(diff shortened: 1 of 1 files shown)


anuraaga:
```
> Configure project :benchmark
Inferred project: opentelemetry-java-instrumentation, version: 0.1.0-SNAPSHOT
```

iNikem (approved):
I would not get my hopes high regarding build speed...

PR #1752, opened 2020-11-24
Make instrumentations non final

### LINKED ISSUE #1582: Make instrumentations non final

I would like to extend instrumentations/instrumentation modules to add additional data capture and other functionality e.g. filtering/blocking requests. The instrumentations can be extended right now by writing additional instrumentation with the same matcher but it leads to a lot of code repetition and issues like https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1576 can arise. Also it might be hard to maintain since the upstream instrumentation can change.

The `instrumenter.getOrder` allows us to apply custom instrumentation after the "core" instrumentation. This is used to get current span started in the "core" instrumentation. 

Make instrumentation classes non `final` so they can be extended, then be able to load only extended instrumentations (avoid double loading). 

I can add instrumentation that copies helperclasses and other metadata from the core instrumentation. This can lead to issues like https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1576.

**Additional context**
None.

mateuszrzeszutek:
Fixes #1582

Adding a checkstyle rule to enforce this (https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1582#issuecomment-724430470) seems to be non-trivial: checkstyle does not have a "public non-final" check by default, and I'm not sure that it's possible to check "A extends B, B implements TypeInstr" situation reliably. I'll create a separate issue for that.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
@@ -16,6 +16,9 @@ import net.bytebuddy.matcher.ElementMatchers;
 
 /**
  * Interface representing a single type instrumentation. Part of an {@link InstrumentationModule}.
+ *
+ * <p>Classes implementing {@link TypeInstrumentation} should be public and non-final so that it's
+ * possible to extend and reuse them in vendor distributions.
  */
 public interface TypeInstrumentation {
   /**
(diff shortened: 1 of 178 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TypeInstrumentation.java
```diff
@@ -15,6 +15,9 @@
 
 /**
  * Interface representing a single type instrumentation. Part of an {@link InstrumentationModule}.
+ *
+ * <p>Classes implementing {@link TypeInstrumentation} should be public and non-final so that it's
+ * possible to extend and reuse them in vendor distributions.
```
trask, line 20: 👍

trask (approved):

Issue #534, opened 2020-06-16
Muzzle bytecode runtime logs could be warnings, at least for tests


anuraaga:
I found that when writing new instrumentation, getting the unit test to be instrumented at all requires a few struggles with the muzzle

```
14:29:16.770 [Test worker] DEBUG io.opentelemetry.auto.tooling.Instrumenter$Default - Instrumentation muzzled: [lettuce, lettuce-5, lettuce-5.2] -- io.opentelemetry.auto.instrumentation.lettuce.v5_2.LettuceClientResourcesInstrumentation on jdk.internal.loader.ClassLoaders$AppClassLoader@55054057
14:29:16.770 [Test worker] DEBUG io.opentelemetry.auto.tooling.Instrumenter$Default - -- io.opentelemetry.auto.instrumentation.lettuce.v5_2.OpenTelemetryTracing$OpenTelemetrySpan:274 io.opentelemetry.auto.instrumentation.lettuce.v5_2.OpenTelemetryTracing$OpenTelemetryEndpoint#portI requires flag PACKAGE_OR_HIGHER found 20
14:29:16.770 [Test worker] DEBUG io.opentelemetry.auto.tooling.Instrumenter$Default - -- io.opentelemetry.auto.instrumentation.lettuce.v5_2.OpenTelemetryTracing$OpenTelemetrySpan:273 io.opentelemetry.auto.instrumentation.lettuce.v5_2.OpenTelemetryTracing$OpenTelemetryEndpoint#ipjava/lang/String requires flag PACKAGE_OR_HIGHER found 20
14:29:16.770 [Test worker] DEBUG io.opentelemetry.auto.tooling.Instrumenter$Default - -- io.opentelemetry.auto.instrumentation.lettuce.v5_2.OpenTelemetryTracing$OpenTelemetryEndpoint:98 io.opentelemetry.auto.instrumentation.lettuce.v5_2.OpenTelemetryTracing$OpenTelemetryEndpoint#namejava/lang/String requires flag PACKAGE_OR_HIGHER found 20
```

The visibility requirements of the business logic isn't so clear to me, but that's less of an issue than that I think these are always an indication of programming bugs in instrumentation but it's very hard to find them among a lot of debug output. So could probably log them higher. If not for normal code, at least having higher level logging of these when running tests in this repo would be nice.

iNikem:
[...]

PR #1643, opened 2020-11-16
Run tests with javaagent.

anuraaga:
Don't get too excited, just first steps in this change for now :)

I wanted to get it out as a PoC just to see what you think of the general direction. Running a test with the agent active, having it load instrumentation for the project under test, and asserting on exported spans works. I'm hoping other steps aren't extremely hard but there are still a few big steps left and a chance of hitting a roadblock, but hope not.

I think there are these major components left

#### Allow registering multiple custom instrumentation like some of our tests require. 

I am using the `otel.initializer_jar` mechanism to load instrumentation. Some thoughts on how to proceed

- Allow `otel.initializer_jar` to accept a CSV of jars.
- Package multiple instrumentation into single JAR for test. Slows down build somewhat (more repackaging)
- Don't use `otel.initializer_jar` and generate an agent JAR for each project with everything bundled. Slows down build more (even more repackaging)

First option seems ok I guess? But maybe doesn't fit with our general customization plans. Second option doesn't seem that bad, especially if `java-concurrent` and `java-classloader` are packaged into the agent (always used), making the dual-packaging rare.

#### Allow registering transformation listeners

If I understand classloaders correctly, we can't simply pass a `AgentBuilder.Listener` from the test classloader to the agent classloader since the implementation would not be inheriting from the agent classloader's class (@trask correct me if I'm wrong, if so things are trivial). It would be very easy to define an interface similar to `AgentBuilder.Listener` in the bootstrap classloader for our use. Unfortunately, `TRANSFORMED_CLASSES_TYPES` depends on byte buddy's `TypeDescription` which puts a wrinkle in this. But instead of a mostly similar, mostly generic Listener, we could have the callback just accept a `boolean matchedGlobalLibrariesIgnoreMatcher`, which is only useful for our test, but guess it's ok.

Edit: Added `TransformationListener` as a POC for this.

#### Allowing updates to `Config`

~~This seems relatively easy, instead of modifying `INSTANCE` directly by initializing `Config` in the test classloader, we can just provide a private (accessed-through-reflection) method that accepts `Properties` and creates the `Config` in the agent classloader.~~  ~~Config's in bootstrap classloader :D~~ But it's shaded

Also feel free to bring up any potential other gotchas you see to come up with strategies to squash them :D

Issue #566, opened 2020-06-23
Provide hooks for vendor-specific customization


trask:
Some things vendors would like to be able to provide without having to fork this repo:

* ✅Custom configuration story
* Custom logging story
* ✅Custom IdsGenerator
* ✅Custom Resource
* ✅Custom Propagator

iNikem:
In addition:
* Add new instrumentation
* Override any existing instrumentation
* Add attributes to existing instrumentation's spans

trask:
Also:

* Custom sampler - https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/387#issuecomment-646354195

RashmiRam:
I am looking for this. @iNikem Do you want me to create a separate issue for this?
> In addition:
> 
> * Add new instrumentation
> * Override any existing instrumentation
> * Add attributes to existing instrumentation's spans

iNikem:
@RashmiRam I am afraid I don’t understand your question

RashmiRam:
Sorry If I am not being clear. I assumed that the above comment from you & question from me in Gitter https://gitter.im/open-telemetry/opentelemetry-java-instrumentation?at=5efefb93bb149531ede7d726 is similar. Is it not?

iNikem:
Ah :) If your need is for vendor-specific auto-instrumentation, then indeed this issue seems to cover that. I was under impression that you wanted for an option for manual instrumentation to add attributes to spans created by auto-instrumentation. I think they are somewhat different things.

RashmiRam:
Yes. You are right. I was specifically looking for an option for manual instrumentation to add attributes to spans created by auto-instrumentation. Can't it be achieved by vendor specific customization too?

iNikem:
Don't know yet :)

trask:
hey @RashmiRam, can you give an example that you are trying to solve, e.g. which instrumentation's spans, and what attributes you would like to add to those spans? i agree with @iNikem, i don't think we've thought about this use case yet, so would be great to have a concrete example to think about. thx!

RashmiRam:
Say, I have a multi-tenant application where each tenant id will be part of the incoming web request url. It would be useful to add the tenant-id as a custom attributes to all the child spans created for the request so that it could be seen if a particular tenant behaves erratically for these type of requests. 

Incoming Web request url `/api/<tenant-id>/resources` is coming to an app using jetty web server.
[...]

PR #1754, opened 2020-11-24
Muzzle mismatch logs should be warnings in tests

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/534

This change actually changes muzzle mismatch logs to warnings everywhere, not only in tests - but hides them behind the `otel.javaagent.debug` config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,7 +398,7 @@ Format is "my.package.MyClass1[method1,method2];my.package.MyClass2[method3]" |
 
 To turn on the agent's internal debug logging:
 
-`-Dio.opentelemetry.javaagent.slf4j.simpleLogger.defaultLogLevel=debug`
+`-Dotel.javaagent.debug=true`
 
 **Note**: These logs are extremely verbose. Enable debug logging only when needed.
 Debug logging negatively impacts the performance of your application.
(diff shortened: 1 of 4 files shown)


iNikem (approved):

iNikem:
If that actually works as intended, then 👍

trask (approved):
I don't totally love altering the log level based on the `debug` flag. If I see something logged at `warn` when the `debug` flag is on, I think I'd be confused not to see that also logged when the `debug` flag is off, and `info+` is (generally) being logged.

But I don't hate it either 😄, so good with it until (if) we think of something better.

iNikem:
But it is our original desire: to have muzzle logs on `WARN`, but do not log them by default. Which means that some `WARN` logs will be absent anyway when enabling only `info` or `debug` as default logging level.

mateuszrzeszutek:
> I don't totally love altering the log level based on the debug flag. If I see something logged at warn when the debug flag is on, I think I'd be confused not to see that also logged when the debug flag is off, and info+ is (generally) being logged.

Hmm, maybe it's just a matter of naming the debug flag more clearly: if we used something like `otel.javaagent.log-everything` instead of `otel.javaagent.debug` the intent would be more obvious (don't treat the log-everything name seriously, that's just an idea 😅 )

trask:
I think ideally we would do something like https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/566#issuecomment-727281792

> log at a higher level than DEBUG if all instrumentation for a particular library (e.g. both cassandra-3.0 and cassandra-4.0) are muzzled

also, our debug logging is very verbose, which makes things get lost easily, even info/warnings

we could consider not using the debug flag during tests (by default), especially after #1643, when we can have separate log level for testing code (e.g. nice to log full spans that are received from exporter)

just stuff to think about / come back to later

i'm good with this PR as-is, it does address a real need 👍

PR #1759, opened 2020-11-25
Add an AWS SDK type instrumentation on an 'entry point' type of class…

anuraaga:
… to ensure helper resources are injected before clients are made.

Working on #1643 brought to light that currently the AWS SDK type instrumentation runs too late since it only matches against `MakeHttpRequestStage` currently, something that happens during an SDK request. So the resource helper which registers our tracer only gets added after the first request is made, missing the first SDK client that is constructed. This currently isn't triggered by unit tests since the resource ends up on the classpath anyways without any injection.


trask (approved):
#1643 is paying off already and hasn't even landed yet ❗

mateuszrzeszutek (approved):

PR #1764, opened 2020-11-25
Enable caching of bytebuddy cache.

anuraaga:
I noticed bytebuddyjava task is never cached, and it seems like rerunning it causes downstream tasks like tests to always run. Allowing it to be cached fixes this - I could confirm that now `./gradlew test` `./gradlew clean` `./gradlew test` is fast, before it reran tests every single time.

That being said, I'm a bit nervous of this change since I would expect bytebuddyjava to be deterministic and not affect downstream tasks - @agoallikmaa can you doublecheck there isn't anything else wrong with the task to make sure us caching it won't cause any pollution?

### DIFF
--- a/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/bytebuddy/ByteBuddyPluginConfigurator.java
+++ b/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/bytebuddy/ByteBuddyPluginConfigurator.java
@@ -68,6 +68,7 @@ public class ByteBuddyPluginConfigurator {
   private Task createLanguageTask(AbstractCompile compileTask, String name) {
     ByteBuddySimpleTask task = project.getTasks().create(name, ByteBuddySimpleTask.class);
     task.setGroup("Byte Buddy");
+    task.getOutputs().cacheIf(unused -> true);
 
     File classesDirectory = compileTask.getDestinationDir();
     File rawClassesDirectory =


anuraaga:
Filed #1765 since probably need a deep dive to confirm bytebuddyjava is working as expected.

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/bytebuddy/ByteBuddyPluginConfigurator.java
```diff
@@ -68,6 +68,7 @@ public void configure() {
   private Task createLanguageTask(AbstractCompile compileTask, String name) {
     ByteBuddySimpleTask task = project.getTasks().create(name, ByteBuddySimpleTask.class);
     task.setGroup("Byte Buddy");
+    task.getOutputs().cacheIf(unused -> true);
```
iNikem, line 71: It is indeed suspicious. I would assume that better approach would be to have correctly configured inputs and outputs for this task

anuraaga: Do you mean to enable caching? The task is not a `CacheableTask` so even with correct input / outputs configured it can't be cached without opting in with this command.

I'm concerned about the determinism or the task but the outputs inputs seem sane from what I can tell.

anuraaga: To clarify, UP-TO-DATE seems to be working fine. What doesn't work

- After a clean, this task always reruns since it's not a `CacheableTask`. We can fix that by adding this line.

- If the task reruns after a clean, all downstream tasks also rerun even though we would expect deterministic output meaning this wouldn't happen. That needs more investigation in #1765.

I think this change is still worth it even if we fix the determinism issue, and maybe we can survive some risk of cache pollution to get the build sane again. Let me try rerunning the tests in this PR and see

anuraaga: Reran tests, cache worked fine-ish. Build / Java 8 3min, Java 15 25min - still something obviously strange about Java 15 but this seems to at least help us well.

trask (approved):
thanks @anuraaga @iNikem @agoallikmaa for understanding this stuff! still working on my gradle 😅, but I think mostly understand what you are saying here and in #1765

anuraaga:
@iNikem Let me get this in and we'll continue on #1765 since from what I see there, we understand the non-determinism but it shouldn't actually pollute our results.

PR #1769, opened 2020-11-26
Replace java test version rule with property.

anuraaga:
The rule has always been a bit fishy, e.g., `./gradlew testJava8 testJava15` isn't well defined (I guess it runs the build with Java 15). While debugging, I found myself scratching my head on how to do something like `./gradlew clean -PtestJavaVersion=15`, which may not actually be any different with or without the property but would make sure the configuration is identical when cleaning vs building.

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -194,43 +194,31 @@ def isJavaVersionAllowed(JavaVersion version) {
   return true
 }
 
-// We default to testing with Java 11 for most tests, but some tests don't support it, where we change
-// the default test task's version so commands like `./gradlew check` can test all projects regardless
-// of Java version.
-if (!isJavaVersionAllowed(JavaVersion.toVersion(DEFAULT_JAVA_VERSION))) {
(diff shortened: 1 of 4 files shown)


trask (approved):
makes sense, similar to `-PtestLatestDeps=true`

mateuszrzeszutek (approved):

gradle/java.gradle
```diff
@@ -240,7 +228,9 @@ tasks.withType(Test).configureEach {
   timeout = Duration.ofMinutes(15)
 
   // Disable all tests if skipTests property was specified
-  enabled = !project.rootProject.hasProperty("skipTests")
```
iNikem, line 243: I don't think we use this property anywhere. Can we delete it?

iNikem (approved):

gradle/java.gradle
anuraaga: Good point - unlike maven, we have all of assemble, check, build to replace the need for this sort of property.

PR #1770, opened 2020-11-26
Make ByteBuddy task deterministic by using linked sets/maps.

### LINKED ISSUE #1765: bytebuddyjava output not deterministic?
I noticed bytebuddyjava rerunning basically always causes tests to rerun, even with the same sources. This shouldn't happen as we would expect the output of the task to be deterministic. Even if it's not cached and gets rerun every single time, it should only contribute its own time to the build, not the invalidation of downstream tasks.

agoallikmaa:
Use `LinkedHashSet` and `LinkedHashMap` on collections that are later iterated to generate bytecode. This makes the `byteBuddy` tasks deterministic, as the order used for generating code is defined by the order they were discovered in the input class.

Tested by deleting the output of `byteBuddy` task between consecutive runs of `test` task in an instrumentation module. Before this, the tests always reran, but after these changes, even with `byteBuddy` task not being cached, only that task reruns and `test` task does not, as the input (output from `byteBuddy`) is always the same.

Fixes #1765.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/Reference.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/Reference.java
@@ -9,7 +9,7 @@ import io.opentelemetry.javaagent.tooling.Utils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
Thanks!

trask (approved):
can you add a class level comment to `Reference` and `ReferenceCollector` briefly explaining why `LinkedHashSet` and `LinkedHashMap` are used, since it's super non-obvious, and someone (me) might be tempted to switch them back at some time in the distant future?

mateuszrzeszutek (approved):

iNikem (approved):

agoallikmaa:
Added the class-level comments.

PR #1778, opened 2020-11-27
Sign artifacts in build instead of in bintray.

anuraaga:
We don't have to do this, but figured I'd whip it up as an idea.

This configures the build to sign artifacts instead of using Bintray's global key. This allows users to verify that the signing key is indeed owned by OTel (perhaps using Nexus Pro and a strict verification policy), possibly reduces some load on Bintray during publishing, and gives some flexibility if we want to try other publishing options.

For now, I just made a key for OpenTelemetry Java

```
pub   rsa2048 2020-11-27 [SC]
      3F05DDA9F317301E927136D417A27CE7A60FF5F0
uid           [ultimate] OpenTelemetry Java
sub   rsa2048 2020-11-27 [E]
```
`gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys  3F05DDA9F317301E927136D417A27CE7A60FF5F0`

We can change the key as we want to - not sure if it makes sense to have a global GPG key for OTel or stick with one for OTel Java. Otherwise, I'd need a way to share this key with other maintainers, especially the revocation key.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -151,6 +151,8 @@ jobs:
           S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
           BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
           BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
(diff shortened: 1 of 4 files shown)


trask (approved):
👍

iNikem:
I don't have any opinion about that :( But for some reason I don't believe it will help us with Bintray.

Issue #1773, opened 2020-11-26
Support more span attributes in the AWS lambda wrapper


mateuszrzeszutek:
The aim of this issue is to introduce support for most of the FAAS/cloud attributes defined in the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/faas.md). Currently only `faas.execution` is added to the span.

I think that the following attributes can be added pretty easily:
* `cloud.provider`
* `cloud.region`
* `faas.name`
* `faas.version`
* `faas.id` - this one is interesting, because it's marked as a resource attribute in the spec; in the context of AWS lambda functions ARN is used here and with the current implementation it's impossible to add it as a resource (because it can only be extracted from a lambda `Context` passed as a function parameter; the tracer is initialized before that happens)

Also, I'm planning to file a new spec issue so that the following things can be added (so this story will probably evolve in time):
* execution environment
* account id
* function qualifier

For `APIGatewayProxyRequestEvent`s I believe that we can implement most of the HTTP server spec - not sure about other event types though.

anuraaga:
For resource attributes, it probably makes sense to keep them together with other resource providers instead of instrumentation, so adding `LambdaResource` here - https://github.com/open-telemetry/opentelemetry-java/tree/master/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/resource

mateuszrzeszutek:
> I think this is cloud.account.id

Nice! It was absent from the trace semantic conventions doc - I should probably. add it there.

> For resource attributes, it probably makes sense to keep them together with other resource providers instead of instrumentation, so adding LambdaResource here

Yep, that was exactly my intention.

mateuszrzeszutek:
By the way @anuraaga : do you know if it's possible to get the account id from something other than the ARN? My idea was to just extract it from the ARN, but then again it cannot be a resource attribute.

mateuszrzeszutek:
Also, regarding the execution environment: I thought about it and decided to use more generic `process.runtime.*` attributes instead. I'll create an issue & PR in otel-java.

PR #1780, opened 2020-11-27
Add HTTP attributes support for AWS lambda function wrapper

mateuszrzeszutek:
Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1773

This PR implements HTTP server spec for lambdas that use `APIGatewayProxyRequestEvent` as input.

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/HttpSpanAttributes.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/HttpSpanAttributes.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import static io.opentelemetry.api.trace.attributes.SemanticAttributes.HTTP_METHOD;
+import static io.opentelemetry.api.trace.attributes.SemanticAttributes.HTTP_STATUS_CODE;
+import static io.opentelemetry.api.trace.attributes.SemanticAttributes.HTTP_URL;
+import static io.opentelemetry.api.trace.attributes.SemanticAttributes.HTTP_USER_AGENT;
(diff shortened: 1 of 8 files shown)


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/HttpSpanAttributes.java
```diff
@@ -0,0 +1,78 @@
...
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+
+final class HttpSpanAttributes {
```
anuraaga, line 24: Do you think it makes sense to have a `AwsLambdaGatewayProxyTracer extends HttpServerTracer`?

mateuszrzeszutek: I'd considered choosing different tracers based on the input class, but `AwsLambdaTracer` has some AWS specific logic (context propagation, `faas.*` attributes) that I'd need to duplicate in a new tracer (which would extend `HttpServerTracer`); I'd have to put the AWS logic in a separate helper class if I were to do that.
So out of the two choices: AWS logic in tracer vs HTTP logic in tracer I'd decided to put the HTTP things in the second place; it seemed to me that it'd be easier to add support for other event types this way (share the same tracer class with all the AWS things, separate strategy for additional attributes)

iNikem:
@anuraaga nobody review this but you :) Or should I just merge it?

anuraaga (approved):
Oops sorry for missing - I'm a bit worried if the maintenance or http conventions is harder with this approach, e.g., when adding new http sem conventions. But I think it's fine for now anyways

PR #1785, opened 2020-11-28
Convert the last two decorators to tracers and delete BaseDecorator

trask:


instrumentation/twilio-6.6/src/test/groovy/test/TwilioClientTest.groovy
```diff
@@ -144,7 +145,7 @@ class TwilioClientTest extends AgentTestRunner {
         }
         span(1) {
           name "MessageCreator.create"
-          kind CLIENT
+          kind INTERNAL
```
anuraaga, line 148: Is this change ok?

trask: well... it seemed better than having nested `CLIENT` spans. do u think better to suppress the nested `apache-httpclient` spans?

anuraaga: Ah yeah I think we'd normally suppress them.

trask: done

anuraaga (approved):

PR #1789, opened 2020-11-28
Configure maven-publish for uploading to bintray instead of bintray p…

anuraaga:
…lugin, and oss artifactory instead of artifactory plugin.

Let's try out the option of maven-publish to bintray instead of gradle bintray plugin. Even if we decide not to use bintray, it's simply just reconfiguring this repositories block.

https://jfrog.com/blog/publishing-your-maven-project-to-bintray/

This doesn't attempt to fully automate things yet (i.e., we would probably want to extend the workflows to call the bintray API to publish after uploading everything) - first let's see if it works more reliably (confirmed the artifacts get published by running locally, then deleting the uploaded files)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,6 @@ import nebula.plugin.release.git.opinion.Strategies
 plugins {
   id 'idea'
 
-  id "com.jfrog.artifactory" version "4.18.1" apply false
   id "nebula.release" version "15.3.0"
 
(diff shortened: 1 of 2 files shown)


trask (approved):

anuraaga:
Reconfigured snapshots too - seems no harm in avoiding extra plugins where we can.

PR #1790, opened 2020-11-28
Migrate test docker images to GHCR.

anuraaga:
With the hopes that it is less flaky than bintray's docker repo. Pushed the images manually from my computer to bootstrap.

Also only references versioned images to allow updating smoke-tests without disrupting other PRs.

### DIFF
--- a/.github/workflows/build-grpc-smoke-dist.yaml
+++ b/.github/workflows/build-grpc-smoke-dist.yaml
@@ -24,7 +24,11 @@ jobs:
           job-id: grpc-smoke
 
       - name: Login to GitHub Package Registry
-        run: docker login -u ${{ secrets.BINTRAY_USER }} -p ${{ secrets.BINTRAY_API_KEY }} open-telemetry-docker-dev.bintray.io
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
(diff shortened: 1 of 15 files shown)


trask (approved):
is there an advantage to embedding the version in the image name instead of using tag?

trask:
> is there an advantage to embedding the version in the image name instead of using tag?

nevermind me, just realized there's a single image name and everything got moved over into the tag 👍

iNikem:
> > is there an advantage to embedding the version in the image name instead of using tag?
> 
> nevermind me, just realized there's a single image name and everything got moved over into the tag 👍

Why this change? I think the old naming scheme is better.

anuraaga:
@iNikem It's mainly to be able to add new images without having to ask an org admin every time. Separate images is a bit better, but don't know if enough to deal with that, or failed bintray pushes.

iNikem:
Don't see how this can eliminate bintray pushes :)

But if we have to ask org admin every time we want to create a new image, that may be a reason enough. Unfortunate, but sufficient.

iNikem (approved):

anuraaga:
@iNikem I meant bintray pushes of the docker images, they fail a lot right?

iNikem:
Yeah, but do you think that "tag pushes" will be more stable than "name pushes"?

anuraaga:
No clue unless we try :) So far so good

PR #1791, opened 2020-11-28
Just have accept-pr be a noop, which will naturally not succeed if de…

anuraaga:
…pendencies fail.

Had `accept-pr` fail a perfectly good build on me. While I'm not sure, isn't it enough to have a no-op job instead of using the workflow-conclusion action?

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -138,17 +138,11 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 11
-      - name: Run  muzzle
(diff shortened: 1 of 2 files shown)


trask (commented):
I think this is what @iNikem originally tried (see #1498), but IIRC the github branch protection rule doesn't prevent merging on missing job, only on failed job.

trask (approved):
approving to see if merge button lights up even though `accept-pr` job was skipped

anuraaga:
Oh well - hopefully the accept-pr doesn't flake too much :)

PR #1792, opened 2020-11-28
Release build can rely on master build checks instead of testing again.

anuraaga:
Had this random idea.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -13,6 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Wait for checks
+        uses: jitterbit/await-check-suites@v1
+        with:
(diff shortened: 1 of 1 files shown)


trask:
should the tests run already if the corresponding commit has been published to the s3 build cache?

anuraaga:
I noticed release build always seems to rerun once - I think it's because the version bump invalidates the cache, presumably by changing manifests. Not sure if there's a way around it in the Gradle config.

iNikem:
I think this is a slippery path. We have not way to be sure that the same git commit, that we use to build published artifacts, was actually tested. If we don't want to re-test before publish/release, then we should just promote already published snapshots. As I proposed in Artifactory-way of releasing.

anuraaga:
This PR checks that the commit trying to be released has passing checks, that solves that issue right?

iNikem (approved):

PR #1668, opened 2020-11-18
Create javaagent dirs for all instrumentations

trask:
Closes #1638

```
#!/bin/bash -e

while read dir; do
  echo "$dir"
  name=$(basename $dir)
  module=$(echo $dir | sed 's|/|:|g')
  mkdir temp
  mv instrumentation/$dir/* temp
  mv temp instrumentation/$dir/javaagent
  mv instrumentation/$dir/javaagent/$name.gradle instrumentation/$dir/javaagent/$name-javaagent.gradle
  find -name '*.gradle' -type f | xargs sed -i "s/:instrumentation:$module/:instrumentation:$module:javaagent/"
done < list
```

list:

```
akka-actor-2.5
akka-http-10.0
apache-camel-2.20
apache-httpasyncclient-4.0
apache-httpclient/apache-httpclient-2.0
apache-httpclient/apache-httpclient-4.0
async-http-client-1.9
aws-sdk/aws-sdk-1.11
cassandra/cassandra-3.0
cassandra/cassandra-4.0
classloaders
couchbase/couchbase-2.0
couchbase/couchbase-2.6
dropwizard-views-0.7
elasticsearch/elasticsearch-rest-5.0
elasticsearch/elasticsearch-rest-6.4
elasticsearch/elasticsearch-rest-common
elasticsearch/elasticsearch-transport-5.0
elasticsearch/elasticsearch-transport-5.3
elasticsearch/elasticsearch-transport-6.0
elasticsearch/elasticsearch-transport-common
executors
external-annotations
finatra-2.9
geode-1.4
google-http-client-1.19
grizzly-2.0
guava-10.0
hibernate/hibernate-3.3
hibernate/hibernate-4.0
hibernate/hibernate-4.3
hibernate/hibernate-common
http-url-connection
hystrix-1.4
java-httpclient
jaxrs/jaxrs-1.0
jaxrs/jaxrs-2.0/jaxrs-2.0-common
jaxrs/jaxrs-2.0/jaxrs-2.0-jersey-2.0
jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.0
jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.1
jaxrs-client/jaxrs-client-1.1
jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common
jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0
[...]

PR #1794, opened 2020-11-29
Create javaagent dirs for all instrumentations, part 2

trask:
In #1668 (part 1) I re-ran the script and it failed near the end, and so I just ran the last few, but my script had a couple of git commands in it, that ended up clearing out the first part of the script, plus leaving behind a bunch of extra files.

Anyways, this fixes and completes that work.


trask:
The important file to review is `settings.gradle`. If I had re-reviewed it in #1668 after re-running the script I would have noticed the problem.

anuraaga (approved):

PR #1796, opened 2020-11-29
Don't use ConfigUtils outside of AgentTestRunner tests

trask:
Breaking this out from #1643, I think these are good changes on their own, and help reduce the size of that PR.

* `NetPeerUtilsTest` - refactor `NetPeerUtils` to be testable without relying on Config singleton
* `RedisCommandNormalizerTest` - refactor `RedisCommandNormalizer` to be testable without relying on Config singleton
* `InstrumentationModuleTest` - set `Config.INSTANCE` directly instead of going through `ConfigUtils`
* `HttpClientTracerTest` - set `Config.INSTANCE` directly instead of going through `ConfigUtils`


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -46,10 +47,17 @@ public static Config get() {
     return INSTANCE;
   }
 
-  public static Config create(Map<String, String> allProperties) {
+  public static Config createFromAlreadyNormalized(Map<String, String> allProperties) {
```
anuraaga, line 50: A bit obtuse name - maybe create with some javadoc is ok?

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
trask: ya, reworked it, moved `AgentConfigBuilder` to this module (renamed now as `ConfigBuilder`) so that tests can use it

mateuszrzeszutek (approved):
Nice!

PR #1801, opened 2020-11-29
Update WithSpan instrumentation to prevent multiple SERVER and CLIENT spans

trask:
Peeling out more things from #1643

### DIFF
--- a/instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanTracer.java
+++ b/instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanTracer.java
@@ -51,6 +51,10 @@ public class WithSpanTracer extends BaseTracer {
       io.opentelemetry.api.trace.Span span,
       io.opentelemetry.api.trace.Span.Kind kind) {
 
+    if (!span.getSpanContext().isValid()) {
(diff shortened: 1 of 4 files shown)


anuraaga (approved):

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanTracer.java
```diff
@@ -22,6 +24,42 @@ public static WithSpanTracer tracer() {
...
+  public Scope startScope(
+      Context context,
+      io.opentelemetry.api.trace.Span span,
+      io.opentelemetry.api.trace.Span.Kind kind) {
+
+    if (kind == io.opentelemetry.api.trace.Span.Kind.SERVER) {
+      return context.with(CONTEXT_SERVER_SPAN_KEY, span).with(span).makeCurrent();
+    }
+    if (kind == io.opentelemetry.api.trace.Span.Kind.CLIENT) {
+      return context.with(CONTEXT_CLIENT_SPAN_KEY, span).with(span).makeCurrent();
```
iNikem, line 58: See #1806

trask: oh this is a good point, I thought that issue was just about an optimization, but understand now 👍

PR #1803, opened 2020-11-29
Hardcode expected values in tests

trask:
I think this makes sense even outside of #1643 (where the agent tests no longer have access to the classes that the expected values are derived from).

I think it's good anyways to have the expected values be hardcoded instead of derived, so that if you change the underlying constants (or normalization algorithm) you are forced to also update the tests, which makes it clear that the change has end user impact.


anuraaga (approved):
Yup I tend to like this better too

iNikem (approved):

PR #1808, opened 2020-11-30
Test span context propagation

### LINKED ISSUE #1561: End to end tests for propagators
As part of our smoke tests we should test context and baggage propagators as well. 

Test controller can call another controller in the same JVM and return propagated context/baggage.

iNikem:
Closes #1561


mateuszrzeszutek (approved):

trask (approved):
👍

anuraaga (approved):

PR #1809, opened 2020-11-30
Use sdk-extensions/resources in the agent

### LINKED ISSUE #1779: Use sdk-extensions/resources in the agent
`OsResource`, `ProcessResource` (and soon `ProcessRuntimeResource`) add some useful attributes, I think that the agent could include those.
Also I think that even if the user included this jar on the classpath the resources would not register because the context classloader for the SDK is the agent classloader, and it can't load external classes.

mateuszrzeszutek:
Resolves #1779

### DIFF
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootSmokeTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootSmokeTest.groovy
@@ -42,6 +42,10 @@ class SpringBootSmokeTest extends SmokeTest {
     [currentAgentVersion] as Set == findResourceAttribute(traces, "telemetry.auto.version")
       .map { it.stringValue }
       .collect(toSet())
+    findResourceAttribute(traces, "os.name")
+      .map { it.stringValue }
+      .findAny()
+      .isPresent()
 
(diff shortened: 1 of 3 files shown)


iNikem:
That's it? Will those resource be automagically loaded? :)

mateuszrzeszutek:
> That's it? Will those resource be automagically loaded? :)

Yep, that's it. Believe me, I was surprised too.

By the way, now that I think of it, this can probably be tested in smoke tests - I'll add an assertion for those new attributes.

iNikem (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootSmokeTest.groovy
```diff
@@ -42,6 +42,10 @@ class SpringBootSmokeTest extends SmokeTest {
     [currentAgentVersion] as Set == findResourceAttribute(traces, "telemetry.auto.version")
       .map { it.stringValue }
       .collect(toSet())
+    findResourceAttribute(traces, "os.name")
+      .map { it.stringValue }
+      .findAny()
+      .isPresent()
```
trask, line 48: nice 👍

trask (approved):

PR #1810, opened 2020-12-01
Pass around context more instead of span (HttpServerTracer)

trask:
Follow-on to #1525

* now adds `CONTEXT_SERVER_SPAN_KEY` in `startSpan()` instead of `startScope()`
* `HttpServerTracer.startScope()` doesn't do anything after above change, so removed
* pass `Context` to `end()` and `endExceptionally()`, since now don't typically have Span reference handy to pass in

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
@@ -12,7 +12,6 @@ import io.opentelemetry.api.trace.SpanBuilder;
 import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.attributes.SemanticAttributes;
 import io.opentelemetry.context.Context;
-import io.opentelemetry.context.Scope;
 import io.opentelemetry.context.propagation.TextMapPropagator;
 import java.lang.reflect.Method;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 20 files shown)


instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/server/ArmeriaServerTracer.java
```diff
@@ -19,7 +19,8 @@
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class ArmeriaServerTracer
-    extends HttpServerTracer<HttpRequest, RequestLog, ServiceRequestContext, Void> {
+    extends HttpServerTracer<
+        HttpRequest, RequestLog, ServiceRequestContext, ServiceRequestContext> {
```
iNikem, line 23: Why this change?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -58,67 +69,51 @@ public Context startSpan(
...
-  /**
-   * Creates new scoped context, based on the given context, with the given span.
-   *
-   * <p>Attaches new context to the request to avoid creating duplicate server spans.
-   */
-  public Scope startScope(Span span, STORAGE storage, Context context) {
-    // TODO we could do this in one go, but TracingContextUtils.CONTEXT_SPAN_KEY is private
-    Context newContext = context.with(CONTEXT_SERVER_SPAN_KEY, span).with(span);
-    attachServerContext(newContext, storage);
-    return newContext.makeCurrent();
```
iNikem, line 82: I don't understand where this happens now. We have to active new context somewhere, haven't we?

iNikem: A, I see, every instrumentation now does that itself.

I have one problem with this and with the removal of `startScope`: inconsistency. Now instrumentations which use `HttpServerTracer` and those which use e.g. `HttpClientTracer` are different. They used to follow the same pattern: `startSpan+startScope`, now they are going to be different. Maybe it is worth leaving `startScope` method here with just `context.makeCurrent()` to preserve uniformity?

trask: #1811 😁

instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/server/ArmeriaServerTracer.java
trask: hm, I was going to try with `@Nullable Void`, but now can't repro at all, so reverting the change and let's see what CI says..

anuraaga (approved):
I've always found `startScope` to be a confusing method, so I like this

mateuszrzeszutek (approved):

instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/server/ArmeriaServerTracer.java
trask: wow, it passed, thx for asking this 👍👍

PR #1811, opened 2020-12-01
Use Context more in HttpClientTracer

### LINKED ISSUE #946: Don't use boxed numerics in tracer type parameters
I noticed that because of numeric expansion, it's actually very easy to call the wrong method when a subclass uses a boxed type like `Integer` in the tracer. It's probably best to avoid boxed types in type parameters because of this. The specific case I found is HTTP response code, and it could be worth adding a `HttpStatus` enum to our instrumentation API with a factory to convert from a response code, to avoid using integers more than needed in instrumentation.

### LINKED ISSUE #1806: Scopes with invalid spans
We have a check in place to prevent nested CLIENT spans:

```
  private Span internalStartSpan(REQUEST request, String name, long startTimeNanos) {
    Context context = Context.current();
    Span clientSpan = context.get(CONTEXT_CLIENT_SPAN_KEY);

    if (clientSpan != null) {
      // We don't want to create two client spans for a given client call, suppress inner spans.
      return Span.getInvalid();
    }
```

But we always call `tracer.startScope()` after `startSpan()` which results in the invalid span returned from the latter being put into the current context by the former. Which does not make sense. After CLIENT span was created it should always be the active span (until it ends of course).

trask:
* return `Context` from `HttpClientTracer.startSpan()`
  * add `CONTEXT_CLIENT_SPAN_KEY` at this time
  * inject context propagation headers at this time
  * (the above are both especially nice for async http clients, which otherwise had no need to call `startScope()`)
* no more need for `HttpClientTracer.startScope()`
* can no long decide to suppress span during `startSpan()` (before it returned `Span.getInvalid()`) because the caller won't know whether to call `end()` or not on the Span in the returned Context
  * so now introduces `HttpClientTracer.shouldStartSpan()` and calls that before calling `startSpan()`

EDIT: Closes #1806 
Closes #946

CHANGELOG:
* Improvements to the implementation and usage of HttpClientTracer classes


mateuszrzeszutek (commented):
I think that this PR also fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1806 - HTTP clients no longer store invalid spans in the context.

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
```diff
@@ -120,62 +125,30 @@ public static void methodExit(
...
+            state.finished = true;
           } else if ("getInputStream".equals(methodName)) {
-            state.finishSpan(responseCode);
+            // responseCode field is sometimes not populated.
+            // We can't call getResponseCode() due to some unwanted side-effects
+            // (e.g. breaks getOutputStream).
+            if (responseCode > 0) {
+              // Need to explicitly cast to boxed type to make sure correct method is called.
+              // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/946
+              tracer().end(state.context, responseCode);
```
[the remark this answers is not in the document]
trask: good point, it sort of got masked by my change, but underlying issue was still there, reworked and this PR resolves #946 now

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
```diff
@@ -147,127 +149,175 @@ public ApacheHttpClientInstrumentationModule() {
...
-        @Advice.Enter SpanWithScope spanWithScope,
         @Advice.Return Object result,
-        @Advice.Thrown Throwable throwable) {
-      ApacheHttpClientHelper.doMethodExitAndResetCallDepthThread(spanWithScope, result, throwable);
+        @Advice.Thrown Throwable throwable,
+        @Advice.Local("otelContext") Context context,
+        @Advice.Local("otelScope") Scope scope,
+        @Advice.Local("otelCallDepth") CallDepth callDepth) {
+      if (callDepth.decrementAndGet() == 0 && scope != null) {
+        scope.close();
```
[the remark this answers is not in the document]
trask: fixed (removed scope close from `onMethodExit()`)

instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/test/groovy/Elasticsearch5RestClientTest.groovy
```diff
@@ -98,18 +98,6 @@ class Elasticsearch5RestClientTest extends AgentTestRunner {
...
-        span(1) {
-          name expectedOperationName("GET")
-          kind CLIENT
-          childOf span(0)
-          attributes {
-            "${SemanticAttributes.NET_TRANSPORT.key()}" "IP.TCP"
-            "${SemanticAttributes.HTTP_URL.key()}" "_cluster/health"
-            "${SemanticAttributes.HTTP_METHOD.key()}" "GET"
-            "${SemanticAttributes.HTTP_STATUS_CODE.key()}" 200
-            "${SemanticAttributes.HTTP_FLAVOR.key()}" "1.1"
```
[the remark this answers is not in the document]
trask: I think `elasticsearch-rest` is the only instrumentation affected, and it only lost http semantic attributes, which I think is ok because it should be following database semantic conventions (same as `elasticsearch-transport` instrumentation).

I removed the remaining 2 http semantic attributes just now to make it more clear, let me know if this seems better.

instrumentation/apache-httpasyncclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientTracer.java
```diff
@@ -30,6 +32,17 @@ public static ApacheHttpAsyncClientTracer tracer() {
     return TRACER;
   }
 
+  public Context startSpan(Context parentContext) {
+    Span span = super.startSpan(DEFAULT_SPAN_NAME, Span.Kind.CLIENT);
+    return parentContext.with(span);
+  }
+
+  public void inject(Context context, HttpRequest request) {
+    OpenTelemetry.getGlobalPropagators()
```
anuraaga, line 41: This seems like a very boring method, do we need it?

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
```diff
@@ -147,127 +149,175 @@ public ApacheHttpClientInstrumentationModule() {
...
+      if (callDepth.getAndIncrement() != 0) {
+        return;
+      }
+      Context parentContext = currentContext();
+      if (!tracer().shouldStartSpan(parentContext)) {
+        return;
       }
 
-      return ApacheHttpClientHelper.doMethodEnter(request);
+      context = tracer().startSpan(parentContext, request, request);
```
anuraaga, line 166: I might be missing a detail, but instead of `shouldStartSpan` everywhere, can't `startSpan` just return `parentContext` if it shouldn't start a new span?

trask: the problem is how to know if the instrumentation should call end on the span in the returned context

instrumentation/apache-httpasyncclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientTracer.java
trask: removed 😂

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
anuraaga: Ah remember now - had the idea a loong time ago (one of my first issues in the repo :D) that for this case, we might wrap the span in something that would still allow propagation and setting attributes (e.g., to enrich in HTTP instrumentation) without forwarding `end` or any other dangerous methods (update span name also probably not forwarded). If this seems like a reasonable idea not sure if it's better to do it now or later.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -60,60 +59,67 @@ protected HttpClientTracer(Tracer tracer) {
...
   }
 
+  public void endExceptionally(Context context, Throwable throwable) {
+    Span span = Span.fromContext(context);
+    super.endExceptionally(span, throwable, -1);
+  }
+
   /**
    * Returns a new client {@link Span} if there is no client {@link Span} in the current {@link
    * Context}, or an invalid {@link Span} otherwise.
```
iNikem, line 117: This is outdated

instrumentation/apache-httpasyncclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientTracer.java
```diff
@@ -30,6 +31,11 @@ public static ApacheHttpAsyncClientTracer tracer() {
     return TRACER;
   }
 
+  public Context startSpan(Context parentContext) {
+    Span span = super.startSpan(DEFAULT_SPAN_NAME, Span.Kind.CLIENT);
+    return parentContext.with(span);
```
iNikem, line 36: Should this update client span in context as well?

instrumentation/armeria-1.0/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_0/client/OpenTelemetryClient.java
```diff
@@ -52,8 +53,10 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
     long requestStartTimeMicros =
         ctx.log().ensureAvailable(RequestLogProperty.REQUEST_START_TIME).requestStartTimeMicros();
     long requestStartTimeNanos = TimeUnit.MICROSECONDS.toNanos(requestStartTimeMicros);
-    Span span = clientTracer.startSpan(ctx, requestStartTimeNanos);
+    Context parentContext = Context.current();
```
iNikem, line 56: inline?

instrumentation/async-http-client-1.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/asynchttpclient/ResponseAdvice.java
```diff
@@ -26,14 +25,14 @@ public static Scope onEnter(
...
         InstrumentationContext.get(AsyncHandler.class, Pair.class);
-    Pair<Context, Span> spanWithParent = contextStore.get(handler);
-    if (null != spanWithParent) {
+    Pair<Context, Context> contextWithParent = contextStore.get(handler);
+    if (contextWithParent != null) {
       contextStore.put(handler, null);
     }
-    if (spanWithParent.hasRight()) {
-      AsyncHttpClientTracer.tracer().end(spanWithParent.getRight(), response);
+    if (contextWithParent.hasRight()) {
```
iNikem, line 32: Can `contextWithParent` be null or not? These two `ifs` clearly disagrees on this.

instrumentation/async-http-client-1.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/asynchttpclient/ResponseAdvice.java
```diff
@@ -26,14 +25,14 @@ public static Scope onEnter(
...
+      AsyncHttpClientTracer.tracer().end(contextWithParent.getRight(), response);
     }
-    return spanWithParent.hasLeft() ? spanWithParent.getLeft().makeCurrent() : null;
+    return contextWithParent.hasLeft() ? contextWithParent.getLeft().makeCurrent() : null;
```
iNikem, line 35: I think this is very suspicious. Why we make the context, which were active before http call on some other thread, active on response thread? Probably a separate issue from this PR but still

instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/rest/v5_0/Elasticsearch5RestClientInstrumentationModule.java
```diff
@@ -71,7 +71,6 @@ public static void onEnter(
       span = tracer().startSpan(null, method + " " + endpoint);
       scope = tracer().startScope(span);
 
-      tracer().onRequest(span, method, endpoint);
```
iNikem, line 74: Not sure about that. Seems like unrelated change and may be undesired

instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/test/groovy/Elasticsearch5RestClientTest.groovy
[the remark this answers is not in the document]
iNikem: I agree with Mateusz: this seems like undesired collateral damage

instrumentation/khttp-0.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/khttp/KHttpAdvice.java
```diff
@@ -22,33 +23,36 @@ public static void methodEnter(
...
+    if (callDepth.getAndIncrement() != 0) {
+      return;
     }
+    Context parentContext = currentContext();
+    if (!tracer().shouldStartSpan(parentContext)) {
+      return;
+    }
+
+    context = tracer().startSpan(parentContext, new RequestWrapper(method, uri, headers), headers);
+    headers = asWritable(headers);
```
iNikem, line 39: I think this was used to allow for context injection into outgoing headers. Are you sure this still works?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -60,60 +59,67 @@ protected HttpClientTracer(Tracer tracer) {
...
     Setter<CARRIER> setter = getSetter();
     if (setter == null) {
       throw new IllegalStateException(
           "getSetter() not defined but calling startScope(), either getSetter must be implemented or the scope should be setup manually");
     }
+    Context context = parentContext.with(span).with(CONTEXT_CLIENT_SPAN_KEY, span);
     OpenTelemetry.getGlobalPropagators().getTextMapPropagator().inject(context, carrier, setter);
-    context = context.with(CONTEXT_CLIENT_SPAN_KEY, span);
-    return context.makeCurrent();
+    return context;
```
iNikem, line 81: Is it a good idea to inject new context into outgoing request during `startSpan`? Technically, creating a new span does not make it "active" yet. It may never become one and then propagating its context is wrong. In this sense old `startScope` method was much better suited for context propagation.

instrumentation/async-http-client-1.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/asynchttpclient/ResponseAdvice.java
trask: this existing behavior seems good to me, so that any callbacks will get associated to the request instead of lost in the void

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
trask: I don't think injecting context into outgoing http requests is related to putting the context in scope, e.g. async http clients never put the context in scope

instrumentation/khttp-0.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/khttp/KHttpAdvice.java
trask: oh no, thanks for catching this, fixed

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
trask: I wrote up #1822 based on the discussion about this from yesterday

instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/test/groovy/Elasticsearch5RestClientTest.groovy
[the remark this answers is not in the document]
trask: does #1822 sound good?

I'm not too fond of mixing db and http semantic conventions on the same span (e.g. people writing span processors may then have to think through this odd case), but I do like the idea of capturing nested http requests as events on the db span.

iNikem:
@trask Can I ask you to put a separate changelog entry into the PR description? I personally will have a hard time formulating it :)

iNikem (approved):

PR #1812, opened 2020-12-01
Add newrelic and datadog package prefixes to global ignore matcher

pavolloffay:

Related to #1534


mateuszrzeszutek (approved):

tylerbenson (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/matcher/GlobalIgnoresMatcher.java
```diff
@@ -60,11 +60,14 @@ public boolean matches(T target) {
     if (name.startsWith("net.bytebuddy.")
         || name.startsWith("jdk.")
         || name.startsWith("org.aspectj.")
+        || name.startsWith("datadog.")
         || name.startsWith("com.intellij.rt.debugger.")
         || name.startsWith("com.p6spy.")
         || name.startsWith("com.dynatrace.")
         || name.startsWith("com.jloadtrace.")
         || name.startsWith("com.appdynamics.")
+        || name.startsWith("com.newrelic.")
```
jkwatson, line 69: IIRC, NR couldn't have this here because there are tons of non-agent classes around at NR that still need instrumentation.

pavolloffay: do you propose to remove `com.newrelic.` from here or are those classes used only internally at NR?

jkwatson: I don't have a proposal (I'm not at NR any more), but I was responsible for removing this originally, because NR wanted to be able to instrument their own code with the otel agent, and they couldn't because of this.

pavolloffay: Is there anybody from NewRelic to comment? 

cc) @breedx-nr

Maybe we could exclude some subpackages:
https://github.com/newrelic/newrelic-java-agent/tree/main/newrelic-agent/src/main/java/com/newrelic/agent
https://github.com/newrelic/newrelic-java-agent/tree/main/newrelic-api/src/main/java/com/newrelic/api/agent

jkwatson: heh. @breedx-nr is no longer at New Relic, either. I'm not sure who the right person would be.

pavolloffay: maybe @jasonjkeller?

iNikem:
Btw, @pavolloffay why do you want to do this? As I see both DataDog and NewRelic people don't recommend running several agents in parallel. As seen on referenced issue, at least some of Otel agent maintainers (myself :) ) support this sentiment.

What is your usecase for running several agents? 

If we fix `OTEL_TRACE_CLASSES_EXCLUDE`, will that be enough?

trask:
I think it's a lot of (painful) work to care about compatibility with other APM vendor java agents. I'm not sure we want to take that on.

I think `OTEL_TRACE_CLASSES_EXCLUDE` is a nice option for people who want to brave it on their own.

pavolloffay:
@iNikem @trask take into account https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1534#issuecomment-723203177 and @tylerbenson comment that even though they do not support this deployment model it is something that people run and we should do minimal work to make this work. We can document that we do not recommend running multiple agents in parallel but if you want to make it impossible that is another story.

This agent already has exclusions for other agents see the `GlobalIgnoresMatcher.java`. This is a minimal change that does not add any harm. 


When customers evaluate a different APM vendor they usually run multiple agents in parallel, this especially applies for Traceable since it is a security product.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/matcher/GlobalIgnoresMatcher.java
pavolloffay: I have changed `com.newrelic` to `com.newrelic.agent` and `com.newrelic.api.agent`. I doupt they use the OTEL agent internally though.

iNikem (approved):

anuraaga (approved):

trask (approved):
Btw, did you happen to notice what the underlying issue is with our side-by-side compatibility? E.g. I have run Glowroot and New Relic in parallel in many production deployments, and Glowroot did not need any special knowledge or exclusion of New Relic classes. In fact, it was very interesting to get observability of the internals of the New Relic agent 😁.

trask:
@jkwatson and everyone, based on the current discussion in https://github.com/open-telemetry/oteps/pull/143, if side-by-side compatibility is working for users in 1.0, but doesn't work in 1.1, do you think that would be considered a breaking change? if so, should we put this feature behind an "experimental" flag, since we don't test it and so we can't detect if we will be breaking it?

iNikem:
I am strongly convinced that our public message MUST be "we DO NOT support running several agents in parallel. It may or may not work, use at your own risk". Under the hood and _without_ any public communication we may try to be a good neighbour.

tylerbenson:
I agree with @iNikem's statement.  Compatibility should be "best effort, but not guaranteed". There are too many variables that may cause problems.

pavolloffay:
@trask can we move forward this PR and also https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1813

iNikem:
Taking into account all approves, I am going to merge it.

PR #1813, opened 2020-12-01
Add package prefix to classloader ignore matcher

pavolloffay:
Related to #1534 

OTEL agent already ignores other agent's classes by package prefix.
However agets also use non-shaded classes (usually present in agent classloader),
these classes are not excluded by package prefix and have to be excluded
by classloader. One example is Okhttp client used by DataDog agent.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ClassLoaderMatcher.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ClassLoaderMatcher.java
@@ -59,6 +59,9 @@ public final class ClassLoaderMatcher {
       if (canSkipClassLoaderByName(cl)) {
         return true;
       }
+      if (canSkipClassLoaderByPackagePrefix(cl)) {
+        return true;
+      }
       Boolean v = skipCache.getIfPresent(cl);
       if (v != null) {
         return v;
@@ -90,6 +93,18 @@ public final class ClassLoaderMatcher {
         default:
           return false;
       }
+    }
+
+    private static boolean canSkipClassLoaderByPackagePrefix(ClassLoader loader) {
+      String name = loader.getClass().getName();
+      if (name.startsWith("datadog.")
+          || name.startsWith("com.dynatrace.")
+          || name.startsWith("com.appdynamics.")
+          || name.startsWith("com.newrelic.")
+          || name.startsWith("com.nr.agent.")) {
+        return true;
+      }
+      return false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ClassLoaderMatcher.java
```diff
@@ -92,6 +95,18 @@ private static boolean canSkipClassLoaderByName(ClassLoader loader) {
       }
     }
 
+    private static boolean canSkipClassLoaderByPackagePrefix(ClassLoader loader) {
```
iNikem, line 98: I would just add this into `canSkipClassLoaderByName`

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ClassLoaderMatcher.java
pavolloffay: That function uses the FQCN

iNikem: Yes, but still: check if `loader.getClass().getName()` is in a set of names or starts with a given prefix. Still sounds like `canSkipClassLoaderByName` to me :) You make decision based on classloader class name, no?

trask:
I have similar concern as https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1812#issuecomment-738288396

Can we introduce something like `OTEL_CLASS_LOADERS_EXCLUDE` instead, so we don't give the impression that coexisting with other APM javaagents is something we support?

pavolloffay:
> Can we introduce something like OTEL_CLASS_LOADERS_EXCLUDE instead, so we don't give the impression that coexisting with other APM javaagents is something we support?

I would rather document that we do not recommend running multiple agents simultaneously. The experience from the field is however different, customers often run multiple agents in the evaluation stage and as @tylerbenson mentioned we should take the same position as DataDog and be a good neighbor to other agents.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ClassLoaderMatcher.java
pavolloffay: fixed

tylerbenson (approved):

pavolloffay:
@iNikem any idea why `
PR build / accept-pr (pull_request) Failing after 4s — accept-pr` failed? Could you please restart it?

iNikem:
@pavolloffay Because "PR build / muzzle (:instrumentation:jetty-8.0:javaagent) (pull_request)" failed. Please rebase, this should be fixed in #1849

pavolloffay:
rebased

Issue #1595, opened 2020-11-09
Improve wildfly smoke test


trask:
Need to improve the wildfly smoke test so that it would have caught #1591.

iNikem:
What version of WildFly did you use in your manual verification?

trask:
21.0.0.Final

iNikem:
@trask Can you please test this again? I cannot reproduce this issue with WildFly. But I can with GlassFish :)

iNikem:
I took 0.10.0 release which supposed to have this problem and tested wildfly-21.0.0.Final with it and a simple web app. Works without any problems.

iNikem:
A! Now I reproduced it! I needed JSP for that :) pure servlets work just fine

trask:
good find, I wonder if I was using some default app that had JSPs, no idea 😂

trask:
Closed by #1825

trask:
oh, just realized, the test matrix isn't running yet, probably why you didn't close this 😄

iNikem:
Will submit proper test and fix tomorrow

PR #1821, opened 2020-12-02
Introduce test matrix

iNikem:
Preparation for #1595


vovencij:
What's the reason for the matrix to have its own gradle wrapper and not be just a regular sub-project?

iNikem:
> What's the reason for the matrix to have its own gradle wrapper and not be just a regular sub-project?

As all other smoke-test image building subprojects: wish to avoid building them too often. They rarely change, no need to compile and build them every time

smoke-tests/matrix/src/payara-dockerfile.template
```diff
@@ -0,0 +1,5 @@
+FROM payara/server-full:@version@
```
anuraaga, line 1: Can we use docker build-arg instead of custom templating?

https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg

smoke-tests/matrix/build.gradle
```diff
@@ -0,0 +1,100 @@
...
+    "13.0.0.Final": ["8", "11", "15"],
+    "17.0.1.Final": ["8", "11", "15"],
+    "21.0.0.Final": ["8", "11", "15"]
+  ],
+  "liberty": [
+    "20.0.0.12": ["8", "11", "15", "8-jdk-openj9", "11-jdk-openj9", "15-jdk-openj9"]
+  ]
+]
+
+targets.each { server, data ->
```
anuraaga, line 65: I think this is worth some comments on what it's preparing

anuraaga: Nevermind I guess README is enough too

smoke-tests/matrix/src/payara-dockerfile.template
iNikem: Can you use them for FROM? I thought not

anuraaga: Think so, I've used it in Zipkin before https://github.com/openzipkin/zipkin/blob/master/docker/Dockerfile#L31

vovencij: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

iNikem: Still struggling to get it working... What's the big difference between gradle token expansion and docker build args?

iNikem: Fixed

anuraaga: Just that it can be more convenient to run docker build locally sometimes. But no big deal.

anuraaga (approved):

smoke-tests/matrix/src/payara-dockerfile.template
trask: oh that's cool, hadn't seen that before

trask:
> > What's the reason for the matrix to have its own gradle wrapper and not be just a regular sub-project?
> 
> As all other smoke-test image building subprojects: wish to avoid building them too often. They rarely change, no need to compile and build them every time

@iNikem the main inconvenience I find with this approach is needing multiple Intellij workspaces, do u know if there's a way to make the single Intellij understand this layout?

trask:
If not that's fine, as you mention, these projects shouldn't need to change much.

smoke-tests/matrix/build.gradle
```diff
@@ -0,0 +1,93 @@
...
+    "5.2020.6-jdk11": ["11"]
+  ],
+  "wildfly": [
+    "13.0.0.Final": ["8", "11", "15"],
+    "17.0.1.Final": ["8", "11", "15"],
+    "21.0.0.Final": ["8", "11", "15"]
+  ],
+  "liberty": [
+    "20.0.0.12": ["8", "11", "15", "8-jdk-openj9", "11-jdk-openj9", "15-jdk-openj9"]
+  ]
```
trask, line 55: this is some really great test coverage that we were missing!!!

iNikem:
> 
> @iNikem the main inconvenience I find with this approach is needing multiple Intellij workspaces, do u know if there's a way to make the single Intellij understand this layout?

Catch me tomorrow, I can share my screen and show to you. It is totally possible

PR #1823, opened 2020-12-03
Switch publishing release target to sonatype.

anuraaga:


gradle/publish.gradle
```diff
@@ -5,11 +5,11 @@ publishing {
   repositories {
     maven {
       url = version.toString().contains('SNAPSHOT')
-        ? "https://oss.jfrog.org/artifactory/oss-snapshot-local/"
-        : "https://api.bintray.com/maven/open-telemetry/maven/opentelemetry-java-instrumentation/"
+        ? "https://oss.sonatype.org/content/repositories/snapshots/"
```
iNikem, line 8: Why do we want to stop publishing snapshots to oss.jfrog? I am again concerned with users confusion and don't see actual need to change that.

anuraaga: Not a big reason - we talked at the meeting and I think we decided that being consistent in both being sonatype may be simpler for users. It's also less creds for us to deal with. Does it make more sense to do the switch after the next release?

iNikem: If we want to go all-in on Sonatype, then I think nice way to do it is the following:

- Release 0.12.0 to Sonatype and MavenCentral
- Announce that starting from this release we publish to these new repositories, including snapshots
- Start publishing 0.13.0-SNAPSHOT to new repos

anuraaga: Makes sense, thanks

iNikem (approved):

trask (approved):

PR #1830, opened 2020-12-04
Update to SDK 0.12.0

anuraaga:
Now that @jkwatson fixed the horrible [bug](https://github.com/open-telemetry/opentelemetry-java/pull/2176) where `OpenTelemetry` instances were not actually behaving as instances, I think our instrumentation approach of only instrumenting key global methods stopped working. I think we could instead instrument every single global static (which often have default methods implemented at the non-global level) but instead tried a different approach of implementing an `ApplicationOpenTelemetry`.

Also after some drift, this is checking `ObfuscatedTracerProvider` in the SDK, not the `Obfuscated` in the API

https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/all/src/main/java/io/opentelemetry/sdk/OpenTelemetrySdk.java#L51

While it could be changed to check for that API internal `Obfuscated` class, I figured we may as well just instrument the SDK with the agent instead of forcing that internal-sdk-agent dependency.


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequest.java
```diff
@@ -19,20 +20,24 @@
...
+    noHttpPropagationNeeded = fields.isEmpty() || xrayPropagationFieldsOnly(fields);
   }
 
-  private static boolean xrayPropagationFieldsOnly(List<String> fields) {
+  private static boolean xrayPropagationFieldsOnly(Collection<String> fields) {
     // ugly but faster than typical convert-to-set-and-check-contains-only
     return (fields.size() == 1)
-        && (ParentContextExtractor.AWS_TRACE_HEADER_PROPAGATOR_KEY.equalsIgnoreCase(fields.get(0)));
+        && (ParentContextExtractor.AWS_TRACE_HEADER_PROPAGATOR_KEY.equalsIgnoreCase(
+            fields.iterator().next()));
```
anuraaga, line 35: @jkwatson It pains me, but I'm leaning towards reverting the change from `List` to `Collection` - `List` is just easier to use

instrumentation/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/ApplicationDoubleSumObserver.java
```diff
@@ -76,6 +76,14 @@ protected Builder(io.opentelemetry.api.metrics.DoubleSumObserver.Builder agentBu
       return this;
     }
 
+    @Override
+    public DoubleSumObserver.Builder setCallback(Callback<DoubleResult> callback) {
+      agentBuilder.setCallback(
```
anuraaga, line 81: @jkwatson No clue what I'm doing here - running a callback with a lambda that itself runs a function seems really funky. I want my micrometer! 😿

instrumentation/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/ApplicationOpenTelemetry.java
```diff
@@ -0,0 +1,63 @@
...
+    applicationTracerProvider =
+        new ApplicationTracerProvider(agentOpenTelemetry.getTracerProvider());
+    applicationMeterProvider = new ApplicationMeterProvider(agentOpenTelemetry.getMeterProvider());
+    applicationContextPropagators =
+        new ApplicationContextPropagators(agentOpenTelemetry.getPropagators());
+  }
+
+  @Override
+  public void setPropagators(ContextPropagators contextPropagators) {
+    throw new UnsupportedOperationException("To be implemented?");
```
anuraaga, line 38: I don't think we supported this ever, it would just no-op into the application OpenTelemetry which is never used. I guess we would need an AgentContextPropatators(application.ContextPropagators) bridge class if we wanted to?

mateuszrzeszutek (approved):

instrumentation/opentelemetry-sdk-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetrysdk/OpenTelemetrySdkInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Advice.OnMethodEnter(skipOn = Advice.OnDefaultValue.class)
+    public static Object onEnter() {
+      return null;
+    }
+
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    public static void methodExit(@Advice.Return(readOnly = false) TracerSdkManagement management) {
+      NoopTracerManagement.logCannotUseTracerManagementWarning();
+      management = NoopTracerManagement.INSTANCE;
```
trask, line 50: do we need to return no-op?

instrumentation/opentelemetry-api-1.0/javaagent/opentelemetry-api-1.0-javaagent.gradle
```diff
@@ -20,6 +20,7 @@ dependencies {
   // and in the code "io.opentelemetry.*" refers to the (shaded) OpenTelemetry API that is used by
   // the agent (as those references will later be shaded)
   compileOnly project(path: ':opentelemetry-api-shaded-for-instrumenting', configuration: 'shadow')
+  compileOnly deps.opentelemetrySdk
```
trask, line 23: is this needed?

trask (approved):
nice, I like how the OpenTelemetry instance reduced instrumentation points

instrumentation/opentelemetry-sdk-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetrysdk/OpenTelemetrySdkInstrumentation.java
anuraaga: What do we return if not a no-op? It seems to also fit well with the "not supported" aspect, since we don't support it we make it a no-op

PR #1831, opened 2020-12-04
Support faas.id & cloud.account.id attributes in AWS Lambda

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1773 (this is the last thing that I wanted to support)

It probably needs to wait for the spec PR https://github.com/open-telemetry/opentelemetry-specification/pull/1265 to get merged.

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
@@ -5,7 +5,9 @@
 
 package io.opentelemetry.instrumentation.awslambda.v1_0;
 
+import static io.opentelemetry.api.trace.attributes.SemanticAttributes.CLOUD_ACCOUNT_ID;
 import static io.opentelemetry.api.trace.attributes.SemanticAttributes.FAAS_EXECUTION;
+import static io.opentelemetry.api.trace.attributes.SemanticAttributes.FAAS_ID;
 import static io.opentelemetry.api.trace.attributes.SemanticAttributes.FAAS_TRIGGER;
 
 import com.amazonaws.services.lambda.runtime.Context;
@@ -19,12 +21,30 @@ import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.api.trace.attributes.SemanticAttributes.FaasTriggerValues;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.util.Collections;
 import java.util.Map;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class AwsLambdaTracer extends BaseTracer {
 
   private static final String AWS_TRACE_HEADER_ENV_KEY = "_X_AMZN_TRACE_ID";
+  private static final MethodHandle GET_FUNCTION_ARN;
+
+  static {
+    MethodHandles.Lookup lookup = MethodHandles.publicLookup();
+    MethodHandle getFunctionArn;
+    try {
+      getFunctionArn =
+          lookup.findVirtual(
+              Context.class, "getInvokedFunctionArn", MethodType.methodType(String.class));
+    } catch (NoSuchMethodException | IllegalAccessException e) {
+      getFunctionArn = null;
+    }
+    GET_FUNCTION_ARN = getFunctionArn;
+  }
 
   private final HttpSpanAttributes httpSpanAttributes = new HttpSpanAttributes();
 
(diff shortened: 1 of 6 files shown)


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
```diff
@@ -19,12 +21,30 @@
...
 import java.util.Map;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class AwsLambdaTracer extends BaseTracer {
 
   private static final String AWS_TRACE_HEADER_ENV_KEY = "_X_AMZN_TRACE_ID";
+  private static final MethodHandle GET_FUNCTION_ARN;
+
+  static {
+    MethodHandles.Lookup lookup = MethodHandles.publicLookup();
```
anuraaga, line 37: I wonder if we can just set a boolean by checking with normal reflection and guard the call to the function, given Java doesn't link until call-time? I'm not aware of any advantage of method handles for guarding against different versions vs that (code is a bit simpler IMO) but maybe no one has ever told me them yet :P

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
```diff
@@ -69,13 +89,37 @@ private SpanBuilder createSpan(Context context, Object input, Map<String, String
...
   }
 
+  private void setCommonAttributes(SpanBuilder span, Context context) {
+    span.setAttribute(FAAS_EXECUTION, context.getAwsRequestId());
+    String arn = getFunctionArn(context);
+    if (arn != null) {
+      span.setAttribute(FAAS_ID, arn);
+      String[] arnParts = arn.split(":");
+      if (arnParts.length >= 5) {
+        span.setAttribute(CLOUD_ACCOUNT_ID, arnParts[4]);
```
anuraaga, line 106: Since account ID is constant, how about storing it into a lazy-inited field? We probably don't need to make it volatile since it'll converge eventually.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
mateuszrzeszutek: Yeah, we certainly can -- I used `MethodHandle` without thinking because that's kind of a default in the agent 😅

mateuszrzeszutek: Hmm, there's one issue that appears when I use the method directly: I need to bump the AWS lambda API to 1.2.0, which also requires bumping javaagent lib version to 1.2.0 because muzzle'll fail. That pretty much means dropping support for <1.2.0... @anuraaga are we OK with that?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
mateuszrzeszutek: I can store the value (whole ARN actually) in a lazy field, but is there any point in doing this? I mean, right now this code will only execute once per whole lambda so it won't be reused anywhere.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
anuraaga: Ah I guess muzzle is too smart for enough - in that case let's stick to the current approach!

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
anuraaga: Isn't it called once per span, meaning once per lambda invocation? And I think by caching we wouldn't need to recompute on every invocation.

mateuszrzeszutek: > You can add initialization code outside of your handler method to reuse resources across multiple invocations. When the runtime loads your handler, it runs static code and the class constructor. Resources that are created during initialization stay in memory between invocations, and can be reused by the handler thousands of times.

😮 
For some reason I was convinced that the whole JVM gets initialized for each lambda invocation. OK, now everything makes sense -- I'll cache it, that's a very good suggestion. Thanks!

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
```diff
@@ -69,13 +92,55 @@ private SpanBuilder createSpan(Context context, Object input, Map<String, String
...
+      return null;
+    }
+  }
+
+  @Nullable
+  private String getAccountId(@Nullable String arn) {
+    if (accountId == null) {
+      synchronized (this) {
+        if (accountId == null) {
+          accountId = new AtomicReference<>();
```
anuraaga, line 131: Since we're locking and volatile don't need to use AtomicReference, it just ends up being another volatile lookup inside it.

anuraaga (approved):
Just small point

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
anuraaga: That'd be a pretty crazy slow runtime ;)

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/AwsLambdaTracer.java
mateuszrzeszutek: Hmm, that is true. Okay, I went a little overboard with the thread safety 😛

PR #1833, opened 2020-12-04
Bootstrap prefix fix

### LINKED ISSUE #1594: Fix BootstrapPackagesProvider
Need to fix `BootstrapPackagesProvider` feature that was (intentionally) broken in PR #1593.

Need to figure out how to reference `BootstrapPackagesProvider` from inside of `ClassLoaderInstrumentation` without breaking restrictive wildfly/jboss modules class loaders.

### LINKED ISSUE #1595: Improve wildfly smoke test

iNikem:
This PR consists of two parts.

First, I introduced two new tests, `GlassFishSmokeTest` and `WildflySmokeTest`. They use new smoke tests docker images and verify that get expected spans on various combinations of app server/jdk versions. And that we don't crash the app server :). WildFly and GlassFish are notorious for their complicated classloading mechanisms, jboss modules and OSGI respectively. Thus these tests indirectly ensure our compatibility with those. This closes #1595

Second, I restored using `BootstrapPackagePrefixesHolder` in `ClassLoaderInstrumentation` which enables back `BootstrapPackagesProvider` SPI. This closes #1594


instrumentation/classloaders/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaclassloader/ClassLoaderInstrumentationModule.java
```diff
@@ -20,7 +20,10 @@ public ClassLoaderInstrumentationModule() {
 
   @Override
   protected String[] additionalHelperClassNames() {
-    return new String[] {"io.opentelemetry.javaagent.tooling.Constants"};
+    return new String[] {
+      "io.opentelemetry.javaagent.tooling.Constants",
+      "io.opentelemetry.javaagent.instrumentation.javaclassloader.ClassLoaderInstrumentation$Holder"
```
mateuszrzeszutek, line 25: This class should be picked automaticallty by muzzle, there's no need to specify it.

mateuszrzeszutek (approved):

instrumentation/classloaders/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaclassloader/ClassLoaderInstrumentationModule.java
iNikem: Why `io.opentelemetry.javaagent.tooling.Constants` is needed then?

mateuszrzeszutek: It's not in the `io.opentelemetry.javaagent.instrumentation` or `io.opentelemetry.instrumentation` (except `api`) package, so it does not get detected as a helper class.

trask (approved):
Nice fix 👍
And thanks for tracking down a way to repro 😅 ❤️

PR #1836, opened 2020-12-05
Use Context more in DatabaseClientTracer

trask:
* Pass `Context` in to `DatabaseClientTracer.startSpan()` and return new `Context`
* Add `CONTEXT_CLIENT_SPAN_KEY` in `startSpan()`
* Remove `DatabaseClientTracer.startScope()` and use `Context.makeCurrent()` directly instead
* Added `DatabaseClientTracer.shouldStartSpan()` but only used it to replace existing call depth tracking for now

### DIFF
--- a/instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/rest/v5_0/Elasticsearch5RestClientInstrumentationModule.java
+++ b/instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/rest/v5_0/Elasticsearch5RestClientInstrumentationModule.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.javaagent.instrumentation.elasticsearch.rest.v5_0;
 
+import static io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge.currentContext;
 import static io.opentelemetry.javaagent.instrumentation.elasticsearch.rest.ElasticsearchRestClientTracer.tracer;
 import static io.opentelemetry.javaagent.tooling.matcher.NameMatchers.namedOneOf;
 import static java.util.Collections.singletonList;
@@ -15,7 +16,7 @@ import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
 
 import com.google.auto.service.AutoService;
-import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.javaagent.tooling.InstrumentationModule;
 import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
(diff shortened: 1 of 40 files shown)


instrumentation/elasticsearch/elasticsearch-transport-5.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/transport/v5_3/Elasticsearch53TransportClientInstrumentationModule.java
```diff
@@ -67,27 +68,27 @@ public Elasticsearch53TransportClientInstrumentationModule() {
...
-        @Advice.Local("otelSpan") Span span,
+        @Advice.Local("otelContext") Context context,
         @Advice.Local("otelScope") Scope scope,
         @Advice.Argument(value = 2, readOnly = false)
             ActionListener<ActionResponse> actionListener) {
 
-      span = tracer().startSpan(null, action);
-      scope = tracer().startScope(span);
+      context = tracer().startSpan(currentContext(), null, action);
+      scope = context.makeCurrent();
```
anuraaga, line 77: We're really lucky that default interface methods work fine, heh

anuraaga (approved):

instrumentation/elasticsearch/elasticsearch-rest-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/rest/v5_0/Elasticsearch5RestClientInstrumentationModule.java
```diff
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.javaagent.instrumentation.elasticsearch.rest.v5_0;
 
+import static io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge.currentContext;
```
iNikem, line 8: In some places above you use `Context.current()` directly. What are the rules? How to choose when to use which? Are they written in CONTRIBUTING.md?

trask: See the Javadoc for `Java8BytecodeBridge`. This situation is not good though. #1840 looks promising, less code in advice, less (no?) need for `Java8BytecodeBridge`.

iNikem: If developers know about `Java8BytecodeBridge` they can read javadoc and understand what is going on. But how can they find it in the first place? :)

trask: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/writing-instrumentation.md#java-agent-instrumentation-gotchas

iNikem (approved):

PR #1837, opened 2020-12-06
Make sure AWS SDK spans suppress client spans all the time.

anuraaga:
Also explicitly populate aws header instead of global propagators.

A lot to process here.

Currently, AWS SDK instrumentation always breaks the signature because transport spans aren't suppressed.

When separating out library / javaagent instrumentation ages ago, I was still very green in this repo and don't think I understood the point of the `testImplementation` dependencies on downstream instrumentation and didn't port them over. So for a while we haven't verified suppression of transport spans.

Sync: This was working in 0.10.0 and stopped working in 0.11.0. I haven't dug into why but presume some refactoring around client tracers is related. We would have caught the issue if I hadn't broken the tests in the first place.

Async: I suspect this has been broken for a long time, but none of our (e.g., aws-otel-java) integration tests use async clients yet. The docs from age-old code in the instrumentation seemed to indicate context is propagated to Netty properly, though extremely hackily, so I removed the even more hacky `amz-sdk-invocation-id` header check that we used to have to prevent signature breakage. But after adding the downstream instrumentation to the tests, I found that this wasn't the case. Using an IntelliJ debugger, I could confirm `beforeTransmission` is run on the application thread, not Netty thread, so the complicated instrumentation hack was having no effect (our executor instrumentation does not apply to netty event loops so I wouldn't expect those to ever have been helping here. Anyways, it's too deep to rely on IMO.). I went ahead and removed the hacky instrumentation and moved the scope management to the interceptor for use in synchronous clients only. For the netty spans, I restored the fun `amz-sdk-invocation-id` as no alternative seems clearly better.

I also went ahead and had the SDK explicitly use aws propagation instead of the global ones. Unlike lambda, there is no use case for any other header, probably not for the next several years really, so this will allow a user's app to be more efficient, they don't have to enable aws globally (meaning on their own ingestion and extraction), and still get the benefit it provides on AWS SDK calls. This makes the integration test confirming that headers haven't been added (breaking signature, the worst thing our instrumentation could ever do) more robust in the process.


instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -41,10 +39,6 @@
  */
 public class TracingExecutionInterceptor implements ExecutionInterceptor {
 
-  public static class ScopeHolder {
-    public static final ThreadLocal<Scope> CURRENT = new ThreadLocal<>();
-  }
-
   private final ExecutionInterceptor delegate;
 
   public TracingExecutionInterceptor() {
```
anuraaga, line 44: I think after #1643 I may try to remove this class, instead just service loading the shaded library instrumentation.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/NettyHttpClientTracer.java
```diff
@@ -65,6 +66,21 @@ protected String responseHeader(HttpResponse httpResponse, String name) {
     return SETTER;
   }
 
+  public boolean shouldStartSpan(Context parentContext, HttpRequest request) {
+    if (!super.shouldStartSpan(parentContext)) {
+      return false;
+    }
+    // The AWS SDK uses Netty for asynchronous clients. We have not found how to propagate Context
+    // from the SDK instrumentation, which executes on an application thread, to Netty
+    // instrumentation, which executs on event loops. If it's possible, it may require instrumenting
```
trask, line 75: I think including a note here about the signature problem would be good too, since that's the biggest worry
```suggestion
    // instrumentation, which executes on event loops. If it's possible, it may require instrumenting
```

trask (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/NettyHttpClientTracer.java
```diff
@@ -70,11 +70,14 @@ public boolean shouldStartSpan(Context parentContext, HttpRequest request) {
...
-    // internal classes. Using a header which is more or less guaranteed to always exist is arguably
-    // more stable.
+    // The AWS SDK uses Netty for asynchronous clients but constructs a request signature before
+    // beginning transport. This means we MUST suppress Netty spans we would normally create or
+    // they will inject their own trace header, which does not match what was present when the
+    // signature was computed, breaking the SDK request completely We have not found how to
+    // cleanly propagate context from the SDK instrumentation, which executes on an application
+    // thread, to Netty instrumentation, which executes on event loops. If it's possible, it may
+    // require instrumenting internal classes. Using a header which is more or less guaranteed to
+    // always exist is arguably more stable.
```
trask, line 80: 👍

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -79,46 +80,55 @@ private RequestType getTypeFromAttributes(ExecutionAttributes executionAttribute
...
+    io.opentelemetry.context.Context otelContext = getContextFromAttributes(executionAttributes);
     if (otelContext == null) {
       return context.httpRequest();
     }
+
     SdkHttpRequest.Builder builder = context.httpRequest().toBuilder();
-    OpenTelemetry.getGlobalPropagators()
-        .getTextMapPropagator()
-        .inject(otelContext, builder, AwsSdkInjectAdapter.INSTANCE);
+    AwsXRayPropagator.getInstance().inject(otelContext, builder, AwsSdkInjectAdapter.INSTANCE);
```
iNikem, line 112: Even if you are sure that AWS SDK calls are always terminal and that nobody uses some alternative implementations of AWS services (like local S3 server?), we at least have to document this behaviour very clearly.

iNikem (approved):

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
anuraaga: Got a bit of extra README out of it too :)

PR #1839, opened 2020-12-07
Switch gRPC instrumentation to instrument public ServerBuilder class.

### LINKED ISSUE #1453: Fail to instrument gRPC server starting from version 1.33
Recently released [gRPC version 1.33](https://github.com/grpc/grpc-java/releases/tag/v1.33.0) broke our instrumentation point: gRPC server builders don't extend `io.grpc.internal.AbstractServerImplBuilder` that we instrument. It is fairly easy to migrate to instrumenting new superclass `ForwardingServerBuilder`, but it is hard to prevent double instrumentation.

anuraaga:
This would have been harder if tracing server interceptor had to be at the beginning of the list, but since it doesn't, we can just instrument the public type without problem.

Fixes #1453

### DIFF
--- a/instrumentation/grpc-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_5/GrpcServerBuilderInstrumentation.java
+++ b/instrumentation/grpc-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_5/GrpcServerBuilderInstrumentation.java
@@ -28,6 +28,11 @@ import net.bytebuddy.matcher.ElementMatcher;
 
 public class GrpcServerBuilderInstrumentation implements TypeInstrumentation {
 
+  // Computing the name at runtime is the simplest way to make sure the String doesn't get shaded.
+  private static final String LIBRARY_INSTRUMENTATION_INTERCEPTOR_NAME =
+      "library.io.opentelemetry.instrumentation.grpc.v1_5.server.TracingServerInterceptor"
+          .substring("library.".length());
+
   @Override
   public ElementMatcher<ClassLoader> classLoaderOptimization() {
     return hasClassesNamed("io.grpc.ServerBuilder");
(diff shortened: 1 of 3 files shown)


instrumentation/grpc-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_5/GrpcServerBuilderInstrumentation.java
```diff
@@ -21,33 +28,61 @@
...
-          break;
-        }
+    public static void onEnter(
+        @Advice.This ServerBuilder<?> serverBuilder,
+        @Advice.Argument(0) ServerInterceptor interceptor) {
+      // Check against unshaded name.
+      if (interceptor
+          .getClass()
+          .getName()
+          .equals("io.opentelemetry.instrumentation.grpc.v1_5.server.TracingServerInterceptor")) {
```
trask, line 66: this string will get shaded too 😄

anuraaga: Oops - removed the check for now so we can come up with a better strategy globally

trask (approved):

instrumentation/grpc-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_5/GrpcServerBuilderInstrumentation.java
iNikem: So, will we have multiple copies of `TracingServerInterceptor`?

trask: See #903

iNikem (approved):

Issue #1429, opened 2020-10-20
Introduce instrumenters (and move away from just tracers)


as-polyakov:
## Problem statement ##
we currently only do tracing in opentelemetry-java-instrumentaiton. We need to do metrics and potentially logs at some stage. For this we need another abstraction - "instrumented" to replace and extend whatever tracer is currently doing

## Proposed solution ##
Introduce Instrumenter instead of tracer, so for example this SpringWebMVC code changes from ([this place](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/WebMvcTracingFilter.java#L33))
```
public void doFilterInternal(
      HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
      throws ServletException, IOException {
    Span serverSpan = tracer.startSpan(request, request, FILTER_CLASS + "." + FILTER_METHOD);

    try (Scope ignored = tracer.startScope(serverSpan, request)) {
      filterChain.doFilter(request, response);
      tracer.end(serverSpan, response);
    } catch (Throwable t) {
      tracer.endExceptionally(serverSpan, t, response);
      throw t;
    }
  }
```

to
```
public void doFilterInternal(
      HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
      throws ServletException, IOException {
    Context operationCtx = instrumenter.startOperation(request, request, FILTER_CLASS + "." + FILTER_METHOD);

    try (Scope ignored = operationCtx.makeCurrent()) {
      filterChain.doFilter(request, response);
      instrumenter.end(operationCtx, response);
    } catch (Throwable t) {
      instrumenter.endExceptionally(serverSpan, t, response);
      throw t;
    }
  }
```

Idea is that a family of Instrumenter's can provide a unified abstraction for working with spans, metrics and other telemetry as well as set of well defined extension points for users to inject their logic. One specific example could be  - allow for customer hook to parse the request and extract needed stuff into context before instrumentation kicks in ([this code](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java#L59) is changed):

```
public Context startOperation(
      REQUEST request, CONNECTION connection, String spanName, long startTimestamp) {
    
[...]

PR #1840, opened 2020-12-07
Some cleaning up of the HttpClientTracer hierarchy

### LINKED ISSUE #1009: Optimize HttpClientTracer methods for invalid spans
In particular, `startScope(Span, ...)` and `end*(Span, ...)` methods can be optimized for invalid spans.

This will be important after removing `CallDepth` (#907) which will then make the invalid span route more common.

trask:
~Motivated by #1429, and by getting rid of `shouldStartSpan`.~

~This PR tackles `HttpClientTracer` subclasses, to get a reasonably broad view of whether this is a good idea or not.~

~Primarily, it converts `HttpClientTracer.startSpan()` to `HttpClientTracer.startOperation()`, returning a new concept Operation, instead of returning Context.~

~This Operation has the `end()` methods on it now, so that it can end spans, metrics, or do nothing if it's a no-op.~

~This eliminates the need for having to call the `shouldStartSpan()` methods before calling `startSpan()`.~

~And it generally reduces the amount of code in advice classes (e.g. way less `Java8BytecodeBridge`).~

~Btw, I think I like how it's the "Operation" that is now propagated via the library-specific attributes for the purpose of calling end somewhere else, just because it's a bit more concrete than "oh, we're propagating yet another Context".~

See the new scope of this PR outlined in [comment below](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1840#issuecomment-743718165).

Closes #1009


mateuszrzeszutek (approved):

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/DefaultHttpUrlConnectionOperation.java
```diff
@@ -0,0 +1,38 @@
...
+
+  @Override
+  public void endExceptionally(Throwable throwable) {
+    super.endExceptionally(throwable);
+    finished = true;
+  }
+
+  @Override
+  public boolean finished() {
+    return finished;
```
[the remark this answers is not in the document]
trask: oh, yes!

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
```diff
@@ -25,6 +34,36 @@ public static NettyHttpClientTracer tracer() {
     return TRACER;
   }
 
+  public HttpClientOperation<HttpResponse> startOperation(
+      ChannelHandlerContext ctx, MessageEvent msg, ChannelTraceContext channelTraceContext) {
+
+    if (!(msg.getMessage() instanceof HttpRequest)) {
+      ctx.sendDownstream(msg);
```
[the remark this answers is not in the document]
trask: oh that was a very bad copy-paste, thx

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/DefaultHttpClientOperation.java
```diff
@@ -0,0 +1,71 @@
...
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+
+public class DefaultHttpClientOperation<RESPONSE> implements HttpClientOperation<RESPONSE> {
+
+  private final Context context;
+  // TODO separate implementation when parentContext is not needed (memory optimization)?
+  private final Context parentContext;
+  // TODO (trask) move end/endExceptionally/etc to here and remove tracer reference
+  private final HttpClientTracer<?, ?, RESPONSE> tracer;
```
trask, line 19: I'm going to work a bit more on this PR. I'll move it back to draft status for now.

I think it will be useful to see how it looks once we start moving things over here from tracers, e.g.
* do tracers just become Operation factories, exposing a single public method `startOperation`?
* can we strictly limit the public surface of `HttpClientOperation`, so that we don't need a separate no-op implementation for each subclass

iNikem: I took only a cursory look at this and I think I like it. But why "moving things over here from tracers" is a good direction? Why is an `Operation` a better place for a code than a `Tracer`? Because it is not limited to tracing?

trask:
Some notes, now that I have better idea of how this is shaping up:
* Operation is not intended for subclassing for instrumentation-specific behavior
  * (though abstract Tracers that are part of `instrumentation-api` will likely have their own `Operation` implementations)
* Tracer is the place for instrumentation-specific behavior, but it only has one public method `startOperation(...)` and the rest of the instrumentation-specific behavior is implemented via protected method overrides
  * the rest of the communication from advice to Tracers is done via the public methods on Operations which call the methods in the abstract Tracer class (which they have access to since they are in same package)

iNikem:
I will probably make a lot of comments about specific code parts later, but let me start with "architectural discussion".

We are introducing new stateful "interface" called `Operation`, which mostly models the lifecycle of a `Span` that it wraps. This allows us to have `noop` operation and thus eliminate the need to call `shouldStartSpan` from all advices. And just call `operation.end()` without null-checking the `scope`.

`Tracers` continue to encapsulate semantic conventions. Operations now will communicate with Tracers instead of Advices, which allows us to reduce public surface of Tracers. I assume that in the future we will have Meters as well and Operations, not Advices, will talk to them too.

Tracers continue to serve as extension points. If somebody wants to alter attributes or modify produced telemetry, they are expected to extend Tracers, not Operations.

Is my understanding of the intent of this change correct?

If yes, then I have problems with `LazyHttpClientOperation` :) I try to understand which method should live where. If Tracers are for semantic convention and Operations are for lifecycle, then `LazyHttpClientOperation.onRequest()` is strange. How do we decide which methods should go to Operation and which to Tracer? Is it Ok for `Operation.onRequest` or similar method to add attributes to span or they must delegate to tracer? This seems like we have keep two classes in sync.

trask:
> Is my understanding of the intent of this change correct?

:thumbsup:

> How do we decide which methods should go to Operation and which to Tracer? Is it Ok for Operation.onRequest or similar method to add attributes to span or they must delegate to tracer? This seems like we have keep two classes in sync.

I'll try moving everything to Operation to see what that looks like, I think that might make things clearer.

iNikem:
>moving everything to Operation

You mean including semantic convention related stuff?

trask:
> > moving everything to Operation
> 
> You mean including semantic convention related stuff?

moving everything to Operation didn't work out well, so went in the other direction and moved those methods `end`, etc back into Tracer, and those methods now accept `Operation` instead of `Context`

trask (commented):
ready for review

iNikem:
> > > moving everything to Operation
> > 
> > 
> > You mean including semantic convention related stuff?
> 
> moving everything to Operation didn't work out well, so went in the other direction and moved those methods `end`, etc back into Tracer, and those methods now accept `Operation` instead of `Context`

Wait, if `end` is back in Tracer, then what is the purpose of Operation? And how this solves the initial problem of not ending wrong spans? Can you please provide a description of how the current state works?

trask:
> And how this solves the initial problem of not ending wrong spans?

If you only have Context, then you don't know if the Span in the Context should be ended or not.

If you have Operation, you can have no-op version, which returns no-op Span, or you can have real version, which returns real Span.

mateuszrzeszutek:
> moving everything to Operation didn't work out well, so went in the other direction and moved those methods end, etc back into Tracer, and those methods now accept Operation instead of Context

Just bring curious: what exactly didn't work out? On the paper having `end()` methods in `Operation` seemed like a good idea,

trask:
Moved the public `end()` methods back to `Operation`, while keeping all extension points inside of the Tracer.

Pros
* Feels natural to call end directly on the operation, instead of having to pass the operation to the tracer

Cons
* `Operation` class needs to be parameterized on `RESPONSE`
* `Operation` needs reference to the tracer
* Probably will want two Operation interfaces, one for operations that have a response object and one for those that don't

trask:
Moving back to Draft.

Discussed with @anuraaga and @bogdandrutu in SIG meeting. Going to try to eliminate the Operation concept, and add a `DelegatingSpan` to the Context in case of suppressed (e.g. nested CLIENT) Span. The `DelegatingSpan` will delegate everything to parent Span except for `end()`.

I'm not fully convinced this will work, because the suppressed instrumentation will end up setting a bunch of attributes on the parent span that may or may not make sense, but only one way to find out for sure, and maybe we'll see some other options emerge to deal with that problem once I prototype.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ParentPropagatingContext.java
```diff
@@ -0,0 +1,42 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.context;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.context.Scope;
+
+public class ParentPropagatingContext implements Context {
```
trask, line 12: this is what tracers can return from `startOperation` if they want to suppress the span

trask (commented):
`startOperation()` now always takes `parentContext` as the first arg, and returns `Context`, which is nice

still nervous about passing attributes from the suppressed child through to the parent, see `ParentPropagatingContext` as alternative to avoid that for now at least

trask:
Since no-ops have to perform some memory allocation with this approach (as opposed to `Operation.noop()`), it's not ideal for no-op'ing the instrumented overloaded methods case, and we may want to keep CallDepth optimization for those

trask:
maybe `shouldStartSpan()` is not such a bad idea, to avoid extra allocation in both op and no-op cases 😭

trask:
Renamed this PR to "Some cleaning up of the HttpClientTracer hierarchy".

It now tackles a lot less, keeping in place the existing `shouldStartOperation` for now, and we can still try other approaches to get rid of `shouldStartOperation` in the future.

Summary of what's in this PR:
* Rename shouldStartSpan to shouldStartOperation, and use it consistently everywhere
* Rename startSpan to startOperation
* Always pass in the parentContext to startOperation and return the new context
* Moved code from Advice into Tracers where it made sense
* Other related cleanup of the http client instrumentation and tracers

Please (re-)review, then we can move forward with #1871 from @parallelstream.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextWithParent.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.context;
+
+import io.opentelemetry.context.Context;
+
+public class ContextWithParent {
```
iNikem, line 10: Doesn't this look like we need `Context.getParent()` method? @anuraaga ?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -67,6 +67,16 @@ public Span startSpan(String spanName, Kind kind) {
     return tracer.spanBuilder(spanName).setSpanKind(kind).startSpan();
   }
 
+  // TODO (trask) make the key private
+  protected final boolean inClientSpan(Context parentContext) {
```
iNikem, line 71: Why `protected final`? It looks like default visibility to me.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -140,16 +140,14 @@ public void endExceptionally(Span span, Throwable throwable) {
...
-    onError(span, unwrapThrowable(throwable));
+    onException(span, throwable);
     end(span, endTimeNanos);
   }
 
-  protected void onError(Span span, Throwable throwable) {
-    addThrowable(span, throwable);
-  }
-
-  protected Throwable unwrapThrowable(Throwable throwable) {
```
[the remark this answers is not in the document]
iNikem: I am not sure this is a good idea

iNikem (commented):
Sorry @trask , but can I ask you to split this PR? It is huge and it does contain independent changes. Some of the them are mechanical, like method rename, and some are not. It is not easy to review that in one go.

trask:
sure, will split it up tomorrow

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextWithParent.java
anuraaga: Don't think so, there's no concept of parenting in `Context`, that's a tracing concept. e.g., `context.with(animal, "cat").with(car, "nissan")`, the intermediate contexts naturally aren't parents, but we have no way of knowing what the "logical parent" is at the context layer.

It's one reason how I've generally encouraged the word `invocationContext` instead of parent, but haven't gotten that to stick :P We need to keep a reference to the invocation context to remount it in asynchronous callbacks. 

Normally we would not need this sort of class since, e.g., both contexts would just get scoped into a lambda callback for an async handler. It seems here we're using it for `InstrumentationContext` - I suspect the reason we can't store them as two fields is because we only support a max of one fields for any given type. If this reasoning makes sense, we may want to move this class to the javaagent API instead of exposing it to library instrumentation which shouldn't really need it and at a glance it seems like a class specific to our javaagent needs.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdk.java
```diff
@@ -40,29 +38,21 @@ public static Tracer tracer() {
...
   }
 
   /**
-   * Returns an {@link ExecutionInterceptor} that can be used with an {@link
-   * software.amazon.awssdk.http.SdkHttpClient} to trace SDK requests. Spans are created with the
-   * provided {@link Kind}.
+   * Returns the {@link Operation} stored in the {@link ExecutionAttributes}, or {@code null} if
+   * there is no operation set.
    */
-  public static ExecutionInterceptor newInterceptor(Kind kind) {
```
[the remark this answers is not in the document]
anuraaga: Yup, thanks

trask:
I split out the more complex parts into #1907, #1908, #1909, #1910, #1911. Will do the more mechanical parts after.

iNikem:
Should we move this back to Draft and invalidate all reviews?

trask:
Closing, probably nothing in here that is still relevant 👍

Issue #1842, opened 2020-12-07
Instrument Spring Batch


mateuszrzeszutek:
The aim of this issue is to introduce a Spring Batch instrumentation into the Java agent.
It is intentionally very vague, as we don't even have a spec for batch jobs right now - it'll be filled with details over time.

anuraaga:
Hi @mateuszrzeszutek - can this issue be closed now?

mateuszrzeszutek:
Let's close it - when job semantic conventions (and/or spring batch specific conventions?) are in the spec we can open another issue for updating this instrumentation.

PR #1843, opened 2020-12-07
Instrument spring-batch

mateuszrzeszutek:
Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1842

This PR introduces just a "skeleton" instrumentation that does not log any span attributes, just creates the spans. Span names may change once a spec PR emerges.


jkwatson:
My personal opinion: We should do this in *much* smaller chunks. Keep PRs to a few hundred lines. It's simply not possible (for me) to do an effective review of 1300+ lines of code.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionTracer.java
```diff
@@ -0,0 +1,60 @@
...
+  public static ChunkExecutionTracer tracer() {
+    return TRACER;
+  }
+
+  public Context startSpan(ChunkContext chunkContext) {
+    String jobName = chunkContext.getStepContext().getJobName();
+    String stepName = chunkContext.getStepContext().getStepName();
+    SpanBuilder spanBuilder =
+        tracer
+            .spanBuilder(String.format("BatchJob %s.%s.Chunk", jobName, stepName))
```
trask, line 30: more efficient to concat

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobExecutionTracer.java
```diff
@@ -0,0 +1,36 @@
...
+    return Context.current().with(span);
+  }
+
+  public void end(Context context) {
+    end(Span.fromContext(context));
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.spring-batch";
```
trask, line 34:
```suggestion
    return "io.opentelemetry.javaagent.spring-batch";
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/step/StepExecutionTracer.java
```diff
@@ -0,0 +1,37 @@
...
+    return Context.current().with(span);
+  }
+
+  public void end(Context context) {
+    end(Span.fromContext(context));
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.spring-batch";
```
trask, line 35:
```suggestion
    return "io.opentelemetry.javaagent.spring-batch";
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -133,12 +133,17 @@ public boolean getBooleanProperty(String name, boolean defaultValue) {
 
   public boolean isInstrumentationEnabled(
       Iterable<String> instrumentationNames, boolean defaultEnabled) {
+    return getInstrumentationBooleanProperty(instrumentationNames, "enabled", defaultEnabled);
+  }
+
+  public boolean getInstrumentationBooleanProperty(
+      Iterable<String> instrumentationNames, String suffix, boolean defaultEnabled) {
```
trask, line 140:
```suggestion
      Iterable<String> instrumentationNames, String suffix, boolean defaultValue) {
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionTracer.java
```diff
@@ -0,0 +1,60 @@
...
+    end(Span.fromContext(context));
+  }
+
+  public void endExceptionally(Context context, Throwable throwable) {
+    endExceptionally(Span.fromContext(context), throwable);
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.spring-batch";
```
trask, line 58:
```suggestion
    return "io.opentelemetry.javaagent.spring-batch";
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobExecutionTracer.java
```diff
@@ -0,0 +1,36 @@
...
+public class JobExecutionTracer extends BaseTracer {
+  private static final JobExecutionTracer TRACER = new JobExecutionTracer();
+
+  public static JobExecutionTracer tracer() {
+    return TRACER;
+  }
+
+  public Context startSpan(JobExecution jobExecution) {
+    String jobName = jobExecution.getJobInstance().getJobName();
+    Span span = startSpan(String.format("BatchJob %s", jobName), INTERNAL);
```
trask, line 24: more efficient to concat

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/step/StepExecutionTracer.java
```diff
@@ -0,0 +1,37 @@
...
+  private static final StepExecutionTracer TRACER = new StepExecutionTracer();
+
+  public static StepExecutionTracer tracer() {
+    return TRACER;
+  }
+
+  public Context startSpan(StepExecution stepExecution) {
+    String jobName = stepExecution.getJobExecution().getJobInstance().getJobName();
+    String stepName = stepExecution.getStepName();
+    Span span = startSpan(String.format("BatchJob %s.%s", jobName, stepName), INTERNAL);
```
trask, line 25: more efficient to concat

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationModule.java
```diff
@@ -0,0 +1,62 @@
...
+
+@AutoService(InstrumentationModule.class)
+public class SpringBatchInstrumentationModule extends InstrumentationModule {
+  public SpringBatchInstrumentationModule() {
+    super(instrumentationNames());
+  }
+
+  public ElementMatcher.Junction<ClassLoader> classLoaderMatcher() {
+    // JSR-352 Batch API
+    return hasClassesNamed("javax.batch.api.Batchlet");
```
trask, line 33: would looking for a spring batch class here be more specific?

trask (commented):
Does it simplify things to instrument `Job.execute()` and `Step.execute()` directly instead of the registration and use of execution listeners? I'm not sure what the equivalent for Chunk would be..

anuraaga (commented):
It might help with @jkwatson's suggestion of splitting into steps (pun intended! :-) Would it help to start from library instrumentation? It reduces load to only be looking at Spring APIs first, and separately reviewing auto-instrumentation tricks.

mateuszrzeszutek:
> Would it help to start from library instrumentation?

That was my first approach actually 😅  But unfortunately I need to use `InstrumentationContext` in all listeners - Spring Batch's built in `ExecutionContext` is periodically stored in the DB, and trying to keep our context & scope there could be too intrusive (and there's no separate execution context for chunk).

> It might help with @jkwatson's suggestion of splitting into steps

👍  
I'll try splitting this PR into 2-3 smaller ones that will be easier to review

> Does it simplify things to instrument Job.execute() and Step.execute() directly instead of the registration and use of execution listeners? I'm not sure what the equivalent for Chunk would be..

There's no equivalent for Chunk unfortunately. For Job and Step I could instrument `execute()` directly and I believe that it would be perfectly fine, but I wanted to rely on the tools that the framework provides - we can be sure that `listener.after(Step|Job)` will be called after a particular unit of work no matter what kind of async/concurrent processing is being used.

mateuszrzeszutek:
@anuraaga @jkwatson @iNikem @trask Please review this PR if you find a bit of free time 🙏 
I've slimmed it down a bit, hope it's easier to read now. Thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -133,17 +133,22 @@ public boolean getBooleanProperty(String name, boolean defaultValue) {
 
   public boolean isInstrumentationEnabled(
       Iterable<String> instrumentationNames, boolean defaultEnabled) {
+    return getInstrumentationBooleanProperty(instrumentationNames, "enabled", defaultEnabled);
+  }
+
+  public boolean getInstrumentationBooleanProperty(
```
iNikem, line 139: Either this name or method's content is misleading. Name says about any boolean property. Content still talks about enablement and contains quite specific logic.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
```diff
@@ -0,0 +1,28 @@
...
+  private static final List<String> INSTRUMENTATION_NAMES =
+      unmodifiableList(asList("spring-batch", "spring-batch-3.0"));
+
+  public static List<String> instrumentationNames() {
+    return INSTRUMENTATION_NAMES;
+  }
+
+  public static boolean isTracingEnabled(String type) {
+    return Config.get()
+        .getInstrumentationBooleanProperty(instrumentationNames(), type + ".enabled", true);
```
iNikem, line 24: Looking at #1874, I think this should be disabled by default.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/GetObjectAdvice.java
```diff
@@ -0,0 +1,33 @@
...
+    ContextStore<JobExecution, ContextAndScope> executionContextStore =
+        InstrumentationContext.get(JobExecution.class, ContextAndScope.class);
+    JobExecutionListener tracingListener = new TracingJobExecutionListener(executionContextStore);
+    if (jobExecutionListeners == null) {
+      jobExecutionListeners = new JobExecutionListener[] {tracingListener};
+    } else {
+      JobExecutionListener[] newListeners =
+          new JobExecutionListener[jobExecutionListeners.length + 1];
+      newListeners[0] = tracingListener;
+      System.arraycopy(jobExecutionListeners, 0, newListeners, 1, jobExecutionListeners.length);
```
iNikem, line 29: Should we check if `jobExecutionListeners` already contains `TracingJobExecutionListener`?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobBuilderHelperInstrumentation.java
```diff
@@ -0,0 +1,52 @@
...
+        this.getClass().getName() + "$EnhanceAdvice");
+  }
+
+  public static class EnhanceAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.FieldValue("properties") JobBuilderHelper.CommonJobProperties properties) {
+      ContextStore<JobExecution, ContextAndScope> executionContextStore =
+          InstrumentationContext.get(JobExecution.class, ContextAndScope.class);
+      properties.addJobExecutionListener(new TracingJobExecutionListener(executionContextStore));
```
iNikem, line 49: Should be protect against double adding?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobFactoryBeanInstrumentation.java
```diff
@@ -0,0 +1,34 @@
...
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.Map;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class JobFactoryBeanInstrumentation implements TypeInstrumentation {
+  @Override
+  public ElementMatcher<? super TypeDescription> typeMatcher() {
+    // Spring Batch XML config
```
iNikem, line 22: What if factory created by JavaConfig?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
mateuszrzeszutek: Hm, I think that right now the whole instrumentations should be treated as experimental and turned off by default.
(And regarding the `isTracingEnabled()` method: honestly I'm not sure that they're needed at all, there's a good chance that I'll remove them in some later PR.)

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobFactoryBeanInstrumentation.java
mateuszrzeszutek: It still works - I just left that comment to signify that this instrumentation covers XML config use case.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobBuilderHelperInstrumentation.java
mateuszrzeszutek: Fortunately spring puts all listeners in a set sooner or later, so it's enough to just implement equals()/hashCode()

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
mateuszrzeszutek: True - since a boolean property is always about enabling/disabling something, I change it to `isInstrumentationPropertyEnabled`

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/GetObjectAdvice.java
mateuszrzeszutek: see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1843#discussion_r542354113

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/GetObjectAdvice.java
```diff
@@ -0,0 +1,33 @@
...
+package io.opentelemetry.javaagent.instrumentation.spring.batch.job;
+
+import io.opentelemetry.javaagent.instrumentation.api.ContextStore;
+import io.opentelemetry.javaagent.instrumentation.api.InstrumentationContext;
+import io.opentelemetry.javaagent.instrumentation.spring.batch.ContextAndScope;
+import net.bytebuddy.asm.Advice;
+import org.springframework.batch.core.JobExecution;
+import org.springframework.batch.core.JobExecutionListener;
+
+public class GetObjectAdvice {
```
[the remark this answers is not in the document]
mateuszrzeszutek: I think at some point I had 2 type instrumentations using the same advice, but that is no longer the point - I'll move it into the type instrumentation class. Thanks!

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobBuilderHelperInstrumentation.java
```diff
@@ -0,0 +1,52 @@
...
+            .and(isProtected())
+            .and(takesArguments(1))
+            .and(takesArgument(0, named("org.springframework.batch.core.Job"))),
+        this.getClass().getName() + "$EnhanceAdvice");
+  }
+
+  public static class EnhanceAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.FieldValue("properties") JobBuilderHelper.CommonJobProperties properties) {
```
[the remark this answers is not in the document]
mateuszrzeszutek: Very good point, done.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobBuilderHelperInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  public ElementMatcher<? super TypeDescription> typeMatcher() {
+    // Java DSL Job config
+    return named("org.springframework.batch.core.job.builder.JobBuilderHelper");
+  }
+
+  @Override
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    return singletonMap(
+        named("enhance")
```
anuraaga, line 36: Do you want to instrument enhance instead of the exit of the constructor?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobFactoryBeanInstrumentation.java
```diff
@@ -0,0 +1,60 @@
...
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    return singletonMap(
+        named("getObject").and(isPublic()).and(takesArguments(0)),
+        this.getClass().getName() + "$GetObjectAdvice");
+  }
+
+  public static class GetObjectAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.FieldValue(value = "jobExecutionListeners", readOnly = false)
```
anuraaga, line 44: Do you think we can avoid the private field access by

- Instrument constructor exit to call `setJobExecutionListeners(empty_array)`
- Instrument `setJobExecutionListeners` to add our listener to the provided array

?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/step/TracingStepExecutionListener.java
```diff
@@ -0,0 +1,62 @@
...
+    return null;
+  }
+
+  @Override
+  public int getOrder() {
+    return HIGHEST_PRECEDENCE;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
anuraaga, line 51: Do we need these weird equals/hashCode methods?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationModule.java
```diff
@@ -0,0 +1,62 @@
...
+      instrumentations.add(new JobFactoryBeanInstrumentation());
+    }
+    if (isTracingEnabled("step")) {
+      instrumentations.add(new StepBuilderHelperInstrumentation());
+    }
+    return instrumentations;
+  }
+
+  protected boolean defaultEnabled() {
+    // TODO: replace this with an experimental flag
```
iNikem, line 59: Why? Enabling specific instrumentation is totally OK, no?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/TracingJobExecutionListener.java
```diff
@@ -0,0 +1,60 @@
...
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    return o != null && getClass() == o.getClass();
+  }
+
+  @Override
+  public int hashCode() {
+    return 0;
```
iNikem, line 58: If `equals` compare classes, shouldn't we do `getClass().hashCode()` here?

iNikem (approved):

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationModule.java
mateuszrzeszutek: Hmm, I vaguely remembered that we're going to have a global experimental flag that enables all experimental features - I might be mistaken though, maybe I remember it wrong.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/step/TracingStepExecutionListener.java
mateuszrzeszutek: They guarantee that only one instance of job/step listener will be present in the ordered set of listeners that spring-batch uses. I'll add a comment explaining this because this is a bit unobvious.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobBuilderHelperInstrumentation.java
mateuszrzeszutek: Hmm, I don't really have a hard preference, but `enhance()` still seems like a proper instrumentation point for this. Is there any advantage of instrumenting the constructor that I might not be aware of?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobFactoryBeanInstrumentation.java
mateuszrzeszutek: Hmm, that's a good idea, I won't have to access the field this way.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobBuilderHelperInstrumentation.java
anuraaga: I think the main point was it seems like enhance can be called multiple times, so constructor makes it easier to prevent dedupe. Also not sure if we want our listener to be in the beginning or end - if end is good, then enhance is better. I think we are able to pick which one here.

mateuszrzeszutek: Ordering is defined by the `Ordered` interface which all listeners in this PR implement - in general, we want them as soon as possible. It doesn't really matter when we add the listener to the job/step, its order property is much more significant.
No duplication of tracing listeners is guaranteed by `equals()` & `hashCode()` methods.

anuraaga (approved):

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationModule.java
iNikem: Indeed, we have this idea...

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobFactoryBeanInstrumentation.java
```diff
@@ -0,0 +1,73 @@
...
+            .and(takesArguments(1))
+            .and(takesArgument(0, isArray())),
+        this.getClass().getName() + "$SetListenersAdvice");
+    return transformers;
+  }
+
+  public static class InitAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(@Advice.This JobFactoryBean jobFactory) {
+      jobFactory.setJobExecutionListeners(new Object[] {});
```
trask, line 52: took me a few min to understand this, worth a comment
```suggestion
      // this will trigger the advice below, which will make sure that the tracing listener is registered
      // even if the application never calls setJobExecutionListeners() directly
      jobFactory.setJobExecutionListeners(new Object[] {});
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/JobParserJobFactoryBeanInstrumentation.java
```diff
@@ -0,0 +1,74 @@
...
+            .and(takesArguments(1))
+            .and(takesArgument(0, isArray())),
+        this.getClass().getName() + "$SetListenersAdvice");
+    return transformers;
+  }
+
+  public static class InitAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(@Advice.This JobParserJobFactoryBean jobFactory) {
+      jobFactory.setJobExecutionListeners(new JobExecutionListener[] {});
```
trask, line 52:
```suggestion
      // this will trigger the advice below, which will make sure that the tracing listener is registered
      // even if the application never calls setJobExecutionListeners() directly
      jobFactory.setJobExecutionListeners(new JobExecutionListener[] {});
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/TracingJobExecutionListener.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public TracingJobExecutionListener(
+      ContextStore<JobExecution, ContextAndScope> executionContextStore) {
+    this.executionContextStore = executionContextStore;
+  }
+
+  @Override
+  public void beforeJob(JobExecution jobExecution) {
+    Context context = tracer().startSpan(jobExecution);
+    Scope scope = context.makeCurrent();
```
trask, line 29: is before/after always synchronous? if so, a comment about that will help future readers not worry about context leaking here

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/step/TracingStepExecutionListener.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public TracingStepExecutionListener(
+      ContextStore<StepExecution, ContextAndScope> executionContextStore) {
+    this.executionContextStore = executionContextStore;
+  }
+
+  @Override
+  public void beforeStep(StepExecution stepExecution) {
+    Context context = tracer().startSpan(stepExecution);
+    Scope scope = context.makeCurrent();
```
trask, line 30: same question about synchronous guarantee? for cautionary tale see https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/49c2a36811d10692f8b5b4669b150fe0fb0c4b3e/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java#L87-L93

trask (approved):

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/job/TracingJobExecutionListener.java
mateuszrzeszutek: I'm 99.99% sure that it's always synchronous -- I'll add the comment.

And FYI: right now I'm testing all major spring batch features (wait for PRs 😄 ) so I'm definitely going to cover all multithreading/async scenarios.

PR #1848, opened 2020-12-08
Add ComponentInstaller SPI and use it for OpenTelemetry SDK

pavolloffay:
Supersedes #1629

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
@@ -149,18 +149,7 @@ public class AgentInitializer {
     return false;
   }
 
-  private static boolean isJavaBefore9() {
+  public static boolean isJavaBefore9() {
     return System.getProperty("java.version").startsWith("1.");
-  }
-
-  public static boolean isJavaBefore9WithJfr() {
-    if (!isJavaBefore9()) {
-      return false;
-    }
-    // FIXME: this is quite a hack because there maybe jfr classes on classpath somehow that have
-    // nothing to do with JDK but this should be safe because only thing this does is to delay
-    // tracer install
-    String jfrClassResourceName = "jdk.jfr.Recording".replace('.', '/') + ".class";
-    return Thread.currentThread().getContextClassLoader().getResource(jfrClassResourceName) != null;
   }
 }
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
@@ -48,90 +48,6 @@ public class AgentInitializer {
 
   public static void initialize(Instrumentation inst, URL bootstrapUrl) {
     startAgent(inst, bootstrapUrl);
-
-    boolean appUsingCustomLogManager = isAppUsingCustomLogManager();
-
-    /*
-     * java.util.logging.LogManager maintains a final static LogManager, which is created during class initialization.
-     *
-     * JMXFetch uses jre bootstrap classes which touch this class. This means applications which require a custom log
-     * manager may not have a chance to set the global log manager if jmxfetch runs first. JMXFetch will incorrectly
-     * set the global log manager in cases where the app sets the log manager system property or when the log manager
-     * class is not on the system classpath.
-     *
-     * Our solution is to delay the initialization of jmxfetch when we detect a custom log manager being used.
-     *
-     * Once we see the LogManager class loading, it's safe to start jmxfetch because the application is already setting
-     * the global log manager and jmxfetch won't be able to touch it due to classloader locking.
-     */
-
-    /*
-     * Similar thing happens with AgentTracer on (at least) zulu-8 because it uses OkHttp which indirectly loads JFR
-     * events which in turn loads LogManager. This is not a problem on newer JDKs because there JFR uses different
-     * logging facility.
-     */
-    if (isJavaBefore9WithJfr() && appUsingCustomLogManager) {
-      log.debug("Custom logger detected. Delaying Agent Tracer initialization.");
-      registerLogManagerCallback(new InstallAgentTracerCallback());
-    } else {
-      installAgentTracer();
-    }
-  }
-
-  private static void registerLogManagerCallback(ClassLoadCallBack callback) {
-    try {
-      Class<?> agentInstallerClass =
-          AGENT_CLASSLOADER.loadClass("io.opentelemetry.javaagent.tooling.AgentInstaller");
-      Method registerCallbackMethod =
-          agentInstallerClass.getMethod("registerClassLoadCallback", String.class, Runnable.class);
-      registerCallbackMethod.invoke(null, "java.util.logging.LogManager", callback);
-    } catch (Exception ex) {
-      log.error("Error registering callback for " + callback.getName(), ex);
-    }
-  }
-
-  protected abstract static class ClassLoadCallBack implements Runnable {
-
-    @Override
-    public void run() {
-      /*
-       * This callback is called from within bytecode transformer. This can be a problem if callback tries
-       * to load classes being transformed. To avoid this we start a thread here that calls the callback.
-       * This seems to resolve this problem.
-       */
-      Thread thread =
-          new Thread(
-              new Runnable() {
-                @Override
-                public void run() {
-                  try {
-                    execute();
-                  } catch (Exception e) {
-                    log.error("Failed to run class loader callback {}", getName(), e);
-                  }
-                }
-              });
-      thread.setName("agent-startup-" + getName());
-      thread.setDaemon(true);
-      thread.start();
-    }
-
-    public abstract String getName();
-
-    public abstract void execute();
-  }
-
-  protected static class InstallAgentTracerCallback extends ClassLoadCallBack {
-
-    @Override
-    public String getName() {
-      return "agent-tracer";
-    }
-
-    @Override
-    public void execute() {
-      installAgentTracer();
-    }
   }
 
   private static synchronized void startAgent(Instrumentation inst, URL bootstrapUrl) {
@@ -142,30 +58,17 @@ public class AgentInitializer {
             agentClassLoader.loadClass("io.opentelemetry.javaagent.tooling.AgentInstaller");
         Method agentInstallerMethod =
             agentInstallerClass.getMethod("installBytebuddyAgent", Instrumentation.class);
-        agentInstallerMethod.invoke(null, inst);
+        ClassLoader savedContextClassLoader = Thread.currentThread().getContextClassLoader();
+        try {
+          Thread.currentThread().setContextClassLoader(AGENT_CLASSLOADER);
+          agentInstallerMethod.invoke(null, inst);
+        } finally {
+          Thread.currentThread().setContextClassLoader(savedContextClassLoader);
+        }
         AGENT_CLASSLOADER = agentClassLoader;
       } catch (Throwable ex) {
         log.error("Throwable thrown while installing the agent", ex);
       }
-    }
-  }
-
-  private static synchronized void installAgentTracer() {
-    if (AGENT_CLASSLOADER == null) {
-      throw new IllegalStateException("Agent should have been started already");
-    }
-    // TracerInstaller.installAgentTracer can be called multiple times without any problem
-    // so there is no need to have a 'agentTracerInstalled' flag here.
-    try {
-      // install global tracer
-      Class<?> tracerInstallerClass =
-          AGENT_CLASSLOADER.loadClass("io.opentelemetry.javaagent.tooling.TracerInstaller");
-      Method tracerInstallerMethod = tracerInstallerClass.getMethod("installAgentTracer");
-      tracerInstallerMethod.invoke(null);
-      Method logVersionInfoMethod = tracerInstallerClass.getMethod("logVersionInfo");
-      logVersionInfoMethod.invoke(null);
-    } catch (Throwable ex) {
-      log.error("Throwable thrown while installing the agent tracer", ex);
     }
   }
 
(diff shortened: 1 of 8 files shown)


iNikem:
Assuming that we need that kind of SPI (and @pavolloffay made a compelling argument that we do :) ), I like this implementation.

But for some reason it demonstrated to me how broken our agent installation is :) And what hacks we do in tests! Horrible!

pavolloffay:
> But for some reason it demonstrated to me how broken our agent installation is :) And what hacks we do in tests! Horrible!

maybe this SPI clarifies how things are bootstrapped :)

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -130,7 +130,7 @@ public String getName() {
 
     @Override
     public void execute() {
-      installAgentTracer();
+      //      installAgentTracer();
```
mateuszrzeszutek, line 133: Why did you comment this out?

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/ComponentInstaller.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.spi;
+
+/**
+ * {@link ComponentInstaller} can be used to install to install any implementation providers that
```
trask, line 9:
```suggestion
 * {@link ComponentInstaller} can be used to install any implementation providers that
```

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/ComponentInstaller.java
```diff
@@ -0,0 +1,33 @@
...
+ * are used by instrumentations. For instance Java agent uses this to install OpenTelemetry SDK. The
+ * instrumentation uses shaded OpenTelemetry API that lives in the bootstrap classlaoder and the
+ * implementation (SDK) is installed via service loader from agent's classloader. This way the
+ * application does not have a direct access to the OpenTelemetry SDK classes. The same approach can
+ * be done for other APIs used by custom instrumentations.
+ *
+ * <p>Before using service loader set the context classloader to agent's classloader e.g. {@code
+ * Thread.currentThread().setContextClassLoader(ComponentInstaller.class.getClassLoader())} if the
+ * component does not accept classloader when loading implementation via service loader.
```
trask, line 18: can we do this automatically before calling ComponentInstallers?

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/ComponentInstaller.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * <p>This is a service provider interface that requires implementations to be registered in {@code
+ * META-INF/services} folder.
+ */
+public interface ComponentInstaller {
+
+  /**
+   * Runs before instrumentations are installed to ByteBuddy. Execute only a minimal code because it
+   * agent boostrap sequence and instrumentations should be installed as early as possible.
```
trask, line 27:
```suggestion
   * Runs before instrumentations are installed to ByteBuddy. Execute only a minimal code because any classes loaded before the instrumentations are installed will have to be retransformed, which takes extra time, and more importantly means that fields can't be added to those classes and InstrumentationContext falls back to the less performant Map implementation for those classes.
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -30,15 +34,47 @@
...
+
+    OpenTelemetrySdkAccess.internalSetForceFlush(
+        (timeout, unit) ->
+            OpenTelemetrySdk.getGlobalTracerManagement().forceFlush().join(timeout, unit));
+  }
+
+  @Override
+  public void afterByteBuddyAgent() {
+    installAgentTracer();
+  }
```
trask, line 76: I'm not sure this is correct to call `installAgentTracer()` here, check out the existing callers of `installAgentTracer()`

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -159,7 +159,7 @@ private static synchronized void installAgentTracer() {
     try {
       // install global tracer
       Class<?> tracerInstallerClass =
-          AGENT_CLASSLOADER.loadClass("io.opentelemetry.javaagent.tooling.TracerInstaller");
+          AGENT_CLASSLOADER.loadClass("io.opentelemetry.javaagent.tooling.OpenTelemetryInstaller");
```
trask, line 162: it feels a little odd to use the OpenTelemetryInstaller class as both an SPI interface and as a static entry point here

pavolloffay: the PR is not ready yet :)

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/TestOpenTelemetryInstaller.java
```diff
@@ -0,0 +1,70 @@
...
+      setGlobalPropagators(
+          DefaultContextPropagators.builder()
+              .addTextMapPropagator(HttpTraceContext.getInstance())
+              .build());
+    }
+    OpenTelemetrySdk.getGlobalTracerManagement().addSpanProcessor(spanProcessor);
+  }
+
+  // Workaround https://github.com/open-telemetry/opentelemetry-java/pull/2096
+  public static void setGlobalPropagators(ContextPropagators propagators) {
```
anuraaga, line 47: Oh we should be able to remove this now since we're on 0.12.0

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/ComponentInstaller.java
pavolloffay: +1

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
pavolloffay: I have updated the PR, this has been removed from the agentInitializer

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/TestOpenTelemetryInstaller.java
pavolloffay: I can clean it up in a separate PR

pavolloffay:
I have spent some time on this PR, it should be ready for review.

pavolloffay:
All green please review :)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -160,7 +140,69 @@ public static ResettableClassFileTransformer installBytebuddyAgent(
...
+     *
+     * Once we see the LogManager class loading, it's safe to start jmxfetch because the application is already setting
+     * the global log manager and jmxfetch won't be able to touch it due to classloader locking.
+     */
+
+    /*
+     * Similar thing happens with AgentTracer on (at least) zulu-8 because it uses OkHttp which indirectly loads JFR
+     * events which in turn loads LogManager. This is not a problem on newer JDKs because there JFR uses different
+     * logging facility.
+     */
```
mateuszrzeszutek, line 192: JMXFetch is a DataDog thing that's no longer present in OTel javaagent, we should consider removing this code in separate issue/PR.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -81,30 +81,10 @@ public static void installBytebuddyAgent(Instrumentation inst) {
...
-      // the
-      // agent class loader, so that SPI finds the agent's (isolated) SDK, and (shaded)
-      // OpenTelemetry registers it, and then when instrumentation calls (shaded)
-      // OpenTelemetry.getGlobalTracerProvider() later, they get back the agent's (isolated) SDK
-      //
-      // but if we don't trigger this early registration, then if instrumentation is the first to
-      // call (shaded) OpenTelemetry.getGlobalTracerProvider(), then SPI can't see the agent class
-      // loader,
-      // and so (shaded) OpenTelemetry registers the no-op TracerFactory, and it cannot be replaced
-      // later
```
mateuszrzeszutek, line 98: Can you preserve this comment in `OpenTelemetryInstaller`?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
pavolloffay: I can clean it up once merged

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
pavolloffay: moved

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
pavolloffay: this should be correct

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -246,57 +143,11 @@ private static boolean isDebugMode() {
...
-
-    return false;
-  }
-
   private static boolean isJavaBefore9() {
     return System.getProperty("java.version").startsWith("1.");
   }
 
-  private static boolean isJavaBefore9WithJfr() {
+  public static boolean isJavaBefore9WithJfr() {
```
trask, line 150: can move this to AgentInstaller also

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
trask: oh i see this is delayed via the callback when needed 👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -160,7 +140,58 @@ public static ResettableClassFileTransformer installBytebuddyAgent(
...
+     * logging facility.
+     */
+    boolean appUsingCustomLogManager = isAppUsingCustomLogManager();
+    if (AgentInitializer.isJavaBefore9WithJfr() && appUsingCustomLogManager) {
+      log.debug("Custom logger detected. Delaying Agent Tracer initialization.");
+      for (ComponentInstaller componentInstaller : componentInstallers) {
+        registerClassLoadCallback(
+            "java.util.logging.LogManager",
+            new InstallComponentAfterByteBuddyCallback(componentInstaller));
+      }
```
trask, line 189: I think nice to pass the list into `InstallComponentsAfterByteBuddyCallback`, consolidating into a single callback

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -160,7 +140,58 @@ public static ResettableClassFileTransformer installBytebuddyAgent(
...
+      Iterable<ComponentInstaller> componentInstallers) {
+    ClassLoader savedContextClassLoader = Thread.currentThread().getContextClassLoader();
+    try {
+      Thread.currentThread().setContextClassLoader(AgentInstaller.class.getClassLoader());
+      for (ComponentInstaller componentInstaller : componentInstallers) {
+        componentInstaller.beforeByteBuddyAgent();
+      }
+    } finally {
+      Thread.currentThread().setContextClassLoader(savedContextClassLoader);
+    }
```
trask, line 158: I think setting the context class loader would make more sense (and be more self-explanatory) up in AgentInitializer.startAgent(), e.g.

```
        ClassLoader agentClassLoader = createAgentClassLoader("inst", bootstrapUrl);
        Class<?> agentInstallerClass =
            agentClassLoader.loadClass("io.opentelemetry.javaagent.tooling.AgentInstaller");
        Method agentInstallerMethod =
            agentInstallerClass.getMethod("installBytebuddyAgent", Instrumentation.class);
        ClassLoader savedContextClassLoader = Thread.currentThread().getContextClassLoader();
        try {
          Thread.currentThread().setContextClassLoader(agentClassLoader);
          agentInstallerMethod.invoke(null, inst);
        } finally {
          Thread.currentThread().setContextClassLoader(savedContextClassLoader);
        }
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -30,15 +34,40 @@
...
+    // the
+    // agent class loader, so that SPI finds the agent's (isolated) SDK, and (shaded)
+    // OpenTelemetry registers it, and then when instrumentation calls (shaded)
+    // OpenTelemetry.getGlobalTracerProvider() later, they get back the agent's (isolated) SDK
+    //
+    // but if we don't trigger this early registration, then if instrumentation is the first to
+    // call (shaded) OpenTelemetry.getGlobalTracerProvider(), then SPI can't see the agent class
+    // loader,
+    // and so (shaded) OpenTelemetry registers the no-op TracerFactory, and it cannot be replaced
+    // later
```
trask, line 59: I'm not sure this comment makes sense here

trask (commented):
❤️ the moving of code from AgentInitializer (bootstrap) to AgentInstaller (agent class loader)

a couple last suggestions

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/TestOpenTelemetryInstaller.java
trask: @pavolloffay I'll do this in #1643

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
pavolloffay: the method uses 
```
Thread.currentThread().getContextClassLoader().getResource(jfrClassResourceName) != null
```

which I guess might be different in the AgentInstaller (agent classloader). Should I change it to the bootstrap claassloader?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
pavolloffay: where to put it, @mateuszrzeszutek suggested to put it here.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
iNikem: Wait, is this delay needed at all now? We don't have `JMXFetch`, `AgentTracer` certainly does not use `OkHttp`. Do we need this?

And if yes, then do we need this delay here, for all components, or just inside `OpenTelemetryInstaller`?

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
trask: `Thread.currentThread().getContextClassLoader()` won't be different since it's based on the current thread and not the current class you are in

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
trask: I believe we still need the delay if we are using an exporter that uses SSL.

This bit me also when 8u231 was released, at the time I was convinced this enhancement was the cause: https://bugs.openjdk.java.net/browse/JDK-8148188

I think the chain goes like:

Agent -> SSL -> JFR -> j.u.l.

and then thankfully in Java 9, JFR doesn't use j.u.l. anymore, which prevents this chain from reaching j.u.l.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
trask: ya, this is a good place, sorry about that 👍

trask (approved):
Thanks @pavolloffay 😅

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
iNikem: @pavolloffay @trask  then do we need this delay here, for all components, or just inside `OpenTelemetryInstaller`?

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
pavolloffay: done

pavolloffay:
@trask thanks for the review. PR updated and rebased

mateuszrzeszutek (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/TestOpenTelemetryInstaller.java
trask: @anuraaga did this in #1924 👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
trask: seems ok to do it for all, so the components themselves don't need to worry about this weirdness

PR #1850, opened 2020-12-08
Instrument spring batch chunk

mateuszrzeszutek:
Continuation of #1843, #1852

### DIFF
--- a/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionTracer.java
+++ b/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionTracer.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.spring.batch.chunk;
+
+import static io.opentelemetry.api.trace.Span.Kind.INTERNAL;
+import static io.opentelemetry.javaagent.instrumentation.spring.batch.SpringBatchInstrumentationConfig.shouldCreateRootSpanForChunk;
+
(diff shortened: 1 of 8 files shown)


iNikem (approved):

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionTracer.java
```diff
@@ -0,0 +1,58 @@
...
+  public static ChunkExecutionTracer tracer() {
+    return TRACER;
+  }
+
+  public Context startSpan(ChunkContext chunkContext) {
+    String jobName = chunkContext.getStepContext().getJobName();
+    String stepName = chunkContext.getStepContext().getStepName();
+    SpanBuilder spanBuilder =
+        tracer.spanBuilder("BatchJob " + jobName + "." + stepName + ".Chunk").setSpanKind(INTERNAL);
+    if (shouldCreateRootSpanForChunk()) {
```
anuraaga, line 30: What's the use case for this flag by the way? It seems like it'd be nice if we can decide one or the other is better and stick with it

anuraaga (approved):
Just small comment but ok either way

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionTracer.java
mateuszrzeszutek: That one's controversial, I know - Spring Batch jobs can get veeery large, like millions of spans in a trace. We are still in the middle of experimenting with how to treat traces like these, one option was to split them into many smaller traces. Personally, I don't think that having a separate trace for each chunk makes sense from a semantic perspective (since it's still a single job execution), but I added this here just to be able to experiment with this. It's highly probable that it'll get removed in one of the future PRs.

PR #1852, opened 2020-12-08
Instrument spring batch tests

mateuszrzeszutek:
Continuation of #1843

This PR contains only additional tests for JSR-352 and XML config batch jobs, no instrumentation changes.

### DIFF
--- a/instrumentation/spring/spring-batch-3.0/javaagent/src/test/groovy/SpringBatchTest.groovy
+++ b/instrumentation/spring/spring-batch-3.0/javaagent/src/test/groovy/SpringBatchTest.groovy
@@ -13,6 +13,7 @@ import io.opentelemetry.instrumentation.test.AgentTestRunner
 import org.springframework.batch.core.JobParameter
 import org.springframework.context.ConfigurableApplicationContext
 import org.springframework.context.annotation.AnnotationConfigApplicationContext
+import org.springframework.context.support.ClassPathXmlApplicationContext
 
 abstract class SpringBatchTest extends AgentTestRunner {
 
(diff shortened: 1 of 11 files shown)


instrumentation/spring/spring-batch-3.0/javaagent/src/test/groovy/SpringBatchTest.groovy
```diff
@@ -78,3 +79,28 @@ class JavaConfigBatchJobTest extends SpringBatchTest implements ApplicationConfi
     new AnnotationConfigApplicationContext(SpringBatchApplication)
   }
 }
+
+class XmlConfigBatchJobTest extends SpringBatchTest implements ApplicationConfigTrait {
+  static final Config PREVIOUS_CONFIG = updateConfig {
```
iNikem, line 84: Do you think this can benefit from the similar change? https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1915

iNikem (commented):
Will there be tests for programmatic/Spring Bean creation of the batch job?

mateuszrzeszutek:
> Will there be tests for programmatic/Spring Bean creation of the batch job?

You mean using the same classes as in XML, but in Java? I haven't thought about it, but it's a good point - I'll introduce it in some future PR.

instrumentation/spring/spring-batch-3.0/javaagent/src/test/groovy/SpringBatchTest.groovy
mateuszrzeszutek: I've tried it but it won't work 😢 

Probably because `otel.instrumentation.spring-batch.enabled` is accessed when `AgentTestRunner` setup methods run, and they run first - before `SpringBatchTest`'s setup.

iNikem (approved):

trask (approved):

anuraaga (approved):

PR #1853, opened 2020-12-08
Fix liberty docker images

laurit:
- enable javaee-8.0 feature, without any features enabled nothing works as servlet support is also a feature
- liberty base image symlinks config directory, can't copy it because it copies the target directory instead of symlink


iNikem:
> * liberty base image symlinks config directory, can't copy it because it copies the target directory instead of symlink

Why is this a bad thing?

laurit:
> > * liberty base image symlinks config directory, can't copy it because it copies the target directory instead of symlink
> 
> Why is this a bad thing?

Because server.xml needs to end up in /opt/ol/wlp/usr/servers/defaultServer/server.xml if config is a plain directory then it will end up in /config/server.xml which is unrelated to server and won't do anything. The same thing applies to app.war.

iNikem (approved):

PR #1854, opened 2020-12-09
Create nexus staging repo in beginning of build instead of relying on…

anuraaga:
… auto routing.

While testing, I found the similar behavior to what the plugin mentions in its README that CI uploads tend to create multiple staging repos unless creating it as a singleton initially.


trask (approved):
nice

PR #1871, opened 2020-12-09
add Instrumenters support

as-polyakov:
PR for https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1429

Idea of this change is to prepare the ground for doing fully fledged instrumentations which goes beyond just traces. There are few steps there:
- we need to have an entry point for instrumentation combining metrics and traces
- at some stage we need to have right extension points so that instrumentation can extract meaningful information from application data (example is - instead of always naming spans in a certain way we might have an extension point where customers could provide their callback for synthesizing span name from request payload)

This PR does the following:
- most of what was called Tracers are now Instrumenters
- Instrumenters are meant to be an entry point for collecting metrics, traces and other observability data
- hence instead of calling startSpan() we do startOperation() which returns Context with the span started
- At some stage probably all subclasses of BaseInstrumenter needs to be renamed, for now I did it with just direct children and Http stack in particular


trask:
hey @parallelstream! uh-oh, I think we are going to have massive merge conflicts between this PR and #1840 😅

parallelstream:
hi @trask yeah...I guess this is slightly different approach as well. I was thinking about introducing an Operation as a structure but in the end felt like we won't really need yet another abstraction for something which is already presented as Span (which is "a logical unit of work")
my PR is basically a massive refactoring to extract Tracer in its simple and pure form and move all of the complexity in Instrumenter - these guys will deal with extracting data from requests, preparing contexts, etc. whereas tracers will be used just a a single tool to start spans together with meter - single tool to record measurements, both of which are used by Instrumenters. At some stage Instrumenters should also expose client-visible extension points for things like extracting contextual data from requests. This will allow clients to really extract meaningful data into spans and contexts and not just use defaults.
It's a lot of files changed due to this refactoring, but otherwise idea is pretty simple.

I am happy to join forces here and try to find a middle ground. Either way we both want to progress in the same direction :)

trask:
@parallelstream check out the latest updates to #1840, the separate `Operation` structure is gone now, so is aligning better with your PR.

on a separate topic, we discussed in the SIG meeting today (which btw everyone is welcome to attend, we have 2 every Thu to cover different timezones), that with tracing API 1.0 coming soon, and metrics API 1.0 not ready for a while, that we probably can't release GA artifacts from this repo that depend on non-GA metrics

that said, this repo is probably a few months away from 1.0, and so it's in our interest to try out metrics and see what our APIs should look like taking metrics in to consideration, and maybe metrics API will be 1.0 by the time this repo is 1.0, or we can always remove metrics sometime before this repo is 1.0, having designed the API in a way that they can be dropped back in as soon as the metrics API is 1.0

parallelstream:
@trask having no Operation makes sense to me and makes us closer. My changes are mostly mechanical - renaming to Instrumenters and using Context in startSpan. I can merge with yours but maybe the other way around is easier

trask:
closing, as some of these ideas have been implemented, and others are now being carried forward in #2596, thx @as-polyakov!

PR #1872, opened 2020-12-10
end to end latency benchmark

bhautikpip:
Using `wrk (HTTP benchmarking)` end to end latency benchmark for any sample app configured with agent and collector setup with receiver. I am also planning to add support for collecting cpu and memory data using `oshi` instrumentation probably in next PR. 

Inspired from this server integrated benchmark: https://github.com/openzipkin/zipkin/blob/master/benchmarks/src/test/java/zipkin2/server/ServerIntegratedBenchmark.java


benchmark-e2e/benchmark-e2e.gradle
```diff
@@ -0,0 +1,16 @@
+plugins {
+  id 'java'
+  id "io.freefair.lombok" version "5.1.0"
```
anuraaga, line 3: We don't use lombok, can you remove this? It looks like you aren't using it either

benchmark-e2e/benchmark-e2e.gradle
```diff
@@ -0,0 +1,16 @@
...
+}
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  compile 'org.apache.logging.log4j:log4j-api:2.7'
```
anuraaga, line 11: Stick with slf4j instead of log4j

anuraaga: Also everything can be `implementation`

benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Log4j2
+public class E2EAgentBenchmark {
+  private static final Logger LOG = LoggerFactory.getLogger(E2EAgentBenchmark.class);
+  private List<GenericContainer<?>> containers;
+
+  // configurable values
+  private String APP = System.getenv("APP_IMAGE");
```
anuraaga, line 27: How about defaults based on one of our smoke test images?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootSmokeTest.groovy#L20

In fact, if you can follow the pattern of the smoke tests to build the agent and add it to the container, then the benchmark will always be running against the current HEAD which is awesome.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/smoke-tests/smoke-tests.gradle#L34

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy#L90

benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private void runBenchmark() throws Exception {
+    String agentPath = System.getProperty("io.opentelemetry.smoketest.agent.shadowJar.path");
+
+    // docker images
+    final DockerImageName WRK_IMAGE = DockerImageName.parse("quay.io/dim/wrk:stable");
+    final DockerImageName APP_IMAGE =
+        DockerImageName.parse(
+            "ghcr.io/open-telemetry/java-test-containers:smoke-springboot-jdk8-20201204.400701583");
```
anuraaga, line 50: We could still have this reference an environment variable and default to this image name to allow benchmarking any image

anuraaga (approved):
Thanks!

trask (approved):
hi @bhautikpip, this is great, thanks!

can you add a job to the nightly build that runs it, either in this PR or in a subsequent PR?

iNikem:
@bhautikpip Build fails `:benchmark-e2e:checkstyleMain` task, can you please fix it?

bhautikpip:
> @bhautikpip Build fails `:benchmark-e2e:checkstyleMain` task, can you please fix it?

weirdly it passes in my local but will investigate what's the issue here. Thanks!

gradle/enforcement/checkstyle.xml
```diff
@@ -262,7 +262,7 @@
     <module name="AbbreviationAsWordInName">
       <property name="ignoreFinal" value="false"/>
       <!-- modified from default to support KHttp -->
-      <property name="allowedAbbreviationLength" value="1"/>
+      <property name="allowedAbbreviationLength" value="10"/>
```
trask, line 265: can this be changed back? it looks like the issue was the uppercase instance field `APP` that you fixed already
```suggestion
      <property name="allowedAbbreviationLength" value="1"/>
```

benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
```diff
@@ -0,0 +1,130 @@
...
+      app = "ghcr.io/open-telemetry/java-test-containers:smoke-springboot-jdk8-20201204.400701583";
+    }
+
+    String agentPath = System.getProperty("io.opentelemetry.smoketest.agent.shadowJar.path");
+
+    // docker images
+    final DockerImageName WRK_IMAGE = DockerImageName.parse("quay.io/dim/wrk:stable");
+    final DockerImageName APP_IMAGE = DockerImageName.parse(app);
+    final DockerImageName OTLP_COLLECTOR_IMAGE =
+        DockerImageName.parse("otel/opentelemetry-collector-dev:latest");
```
trask, line 60: it looks like checkstyle is now complaining about these, can u make them camelCase? that should resolve

anuraaga: I'm ok with loosening the checkstyle actually - I don't see why an abbreviation would need to be limitted to just 2 letters. (seems to not like WRK being 3 letters, APP, OTLP)

trask: it's complaining b/c those are using constant style naming `APP` instead of instance/local var naming `app`, which goes against https://google.github.io/styleguide/javaguide.html

trask: if you did `static final DockerImageName APP_IMAGE = DockerImageName.parse(app);` checkstyle wouldn't complain

trask: the "abbreviation" thing is red-herring, it's to allow something like "serverURL" if you like that instead of "serverUrl"

anuraaga: Oh duh - guess these should be static :)

bhautikpip:
I will submit fresh PR.

PR #1881, opened 2020-12-10
Document advice class name hard coding

breedx-splk:
This came up in the SIG call this morning.  This question has been asked before, so we're trying to help document it so that it is more easily and consistently answered.

### DIFF
--- a/docs/contributing/writing-instrumentation.md
+++ b/docs/contributing/writing-instrumentation.md
@@ -136,15 +136,15 @@ on a single `execute` method:
 
 ```
 transformers.put(
-  isMethod().and(named("execute")), 
+  isMethod().and(named("execute")),
   MyInstrumentationModule.class.getName() + "$WonderfulAdvice");
 ```
 
 Simply referring to the inner class and
 calling `getName()` would be easier to read and understand than
-this odd mix of string concatenation...but please NOTE:  **this is intentional** 
+this odd mix of string concatenation...but please NOTE:  **this is intentional**
 and should be maintained.
 
-Instrumentation modules are typically loaded by the user's classloader, and this 
+Instrumentation modules are typically loaded by the user's classloader, and this
 string concatenation is an optimization that prevents the actual advice class
 from being loaded.
--- a/docs/contributing/writing-instrumentation.md
+++ b/docs/contributing/writing-instrumentation.md
@@ -124,3 +124,27 @@ are not legal in Java 7 (and prior) bytecode.
 Because OpenTelemetry API has many common default methods (e.g. `Span.current()`),
 the `javaagent-api` artifact has a class `Java8BytecodeBridge` which provides static methods
 for accessing these default methods from advice.
+
+### Why hard code advice class names?
+
+Subclasses of `InstrumentationModule` will often implement advice classes static inner classes.
+These classes are referred to by name in the mappings from method descriptor to advice class,
+typically in the `transform()` method.
+
+For instance, this `MyInstrumentationModule` defines a single advice that mathches
+on a single `execute` method:
+
+```
+transformers.put(
+  isMethod().and(named("execute")), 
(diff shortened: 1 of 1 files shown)


docs/contributing/writing-instrumentation.md
```diff
@@ -124,3 +124,27 @@ are not legal in Java 7 (and prior) bytecode.
 Because OpenTelemetry API has many common default methods (e.g. `Span.current()`),
 the `javaagent-api` artifact has a class `Java8BytecodeBridge` which provides static methods
 for accessing these default methods from advice.
+
+### Why hard code advice class names?
+
+Subclasses of `InstrumentationModule` will often implement advice classes static inner classes.
```
iNikem, line 130:
```suggestion
Subclasses of `InstrumentationModule` will often implement advice classes as static inner classes.
```

docs/contributing/writing-instrumentation.md
```diff
@@ -124,3 +124,27 @@ are not legal in Java 7 (and prior) bytecode.
...
+on a single `execute` method:
+
+```
+transformers.put(
+  isMethod().and(named("execute")), 
+  MyInstrumentationModule.class.getName() + "$WonderfulAdvice");
+```
+
+Simply referring to the inner class and
+calling `getName()` would be easier to read and understand than
```
iNikem, line 144:
```suggestion
Simply writing `WonderfulAdvice.class.getName()` would be easier to read and understand than
```

docs/contributing/writing-instrumentation.md
```diff
@@ -124,3 +124,27 @@ are not legal in Java 7 (and prior) bytecode.
...
+this odd mix of string concatenation...but please NOTE:  **this is intentional** 
+and should be maintained.
+
+Instrumentation modules are typically loaded by the user's classloader, and this
```
iNikem, line 148:
```suggestion
Instrumentation modules are loaded by the agent's classloader, and this 
```

docs/contributing/writing-instrumentation.md
breedx-splk: Thanks.

docs/contributing/writing-instrumentation.md
breedx-splk: Thanks for clearing me up, I had it backwards!

trask (approved):
thanks @breedx-splk!

trask:
@breedx-splk looks like the branch needs a `./gradlew spotlessApply` (yes for .md files too 😄)

breedx-splk:
OMG for `.md`!  Never would have guessed....ugh!  On it.

docs/contributing/writing-instrumentation.md
iNikem: You decided against accepting this suggestion? It is fine, you just send mixed signals here :)

docs/contributing/writing-instrumentation.md
```diff
@@ -124,3 +124,27 @@ are not legal in Java 7 (and prior) bytecode.
 Because OpenTelemetry API has many common default methods (e.g. `Span.current()`),
 the `javaagent-api` artifact has a class `Java8BytecodeBridge` which provides static methods
 for accessing these default methods from advice.
+
+### Why hard code advice class names?
+
+Subclasses of `InstrumentationModule` will often implement advice classes as static inner classes.
```
mateuszrzeszutek, line 130: `transformers()` method belongs to the `TypeInstrumentation` interface and I think it's a bit better to focus on them in this particular gotcha. A module just brings everything together.

```suggestion
Implementations of `TypeInstrumentation` will often implement advice classes as static inner classes.
```

mateuszrzeszutek: It might be useful to mention that it's okay to keep the advice class in the module class or in the package.

anuraaga (approved):

trask:
Thanks @breedx-splk!

PR #1891, opened 2020-12-11
Update javadocs for generic types in HttpServerTracer

### LINKED ISSUE #791: Document the STORAGE parameter in HttpServerTracer
It is confusing when/if you need it, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/756#discussion_r459842332

Suggest to use `Void` if you do not need the `STORAGE` parameter for propagating the context to async handlers

breedx-splk:
This resolves #791

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
@@ -18,6 +18,19 @@ import java.util.concurrent.TimeUnit;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 // TODO In search for a better home package
+
+/**
+ * Base class for implementing Tracers for HTTP servers. It has 3 types that must be specified by
+ * subclasses:
+ *
+ * @param <REQUEST> - The specific type for HTTP requests
+ * @param <RESPONSE> - The specific type for HTTP responses
+ * @param <CONNECTION> - The specific type of HTTP connection, used to get peer address information
+ *     and HTTP flavor. Use Void if your subclass does not need the connection for peerHostIP,
+ *     peerPort, or flavor.
+ * @param <STORAGE> - Implementation specific storage type for attaching/getting the server context.
+ *     Use Void if your subclass does not have an implementation specific storage need.
+ */
 public abstract class HttpServerTracer<REQUEST, RESPONSE, CONNECTION, STORAGE> extends BaseTracer {
 
   // the class name is part of the attribute name, so that it will be shaded when used in javaagent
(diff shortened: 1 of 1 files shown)


jkwatson (approved):
seems good to me. thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -18,6 +18,19 @@
...
+
+/**
+ * Base class for implementing Tracers for HTTP servers. It has 3 types that must be specified by
+ * subclasses:
+ *
+ * @param <REQUEST> - The specific type for HTTP requests
+ * @param <RESPONSE> - The specific type for HTTP responses
+ * @param <CONNECTION> - The specific type of HTTP connection, used to get peer address information
+ *     and HTTP flavor. Use Void if your subclass does not need the connection for peerHostIP,
+ *     peerPort, or flavor.
```
trask, line 30: can you check if there any subclasses that don't use CONNECTION, I'm not sure if makes sense to use Void for this (unlike STORAGE)

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -18,6 +18,19 @@
...
+ * @param <STORAGE> - Implementation specific storage type for attaching/getting the server context.
```
anuraaga, line 31: Any doc is better than no doc so this LGTM, but maybe we can go further. `STORAGE` has always been very hard for me to grok. This comment still doesn't really tell me how to use this parameter, "what's server context", "what's implementation specific storage" are still remaining questions.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
breedx-splk: > can you check if there any subclasses that don't use CONNECTION, I'm not sure if makes sense to use Void for this (unlike STORAGE)

Good call. They all do.  I'll remove the `Void` bit.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
trask: Yeah, I'm planning to work on cleaning up the http server instrumentations after getting through the http clients, and will think on this.

trask (approved):

PR #1892, opened 2020-12-12
Shade trace propagators so they can be injected in instrumentation th…

anuraaga:
…at benefits from a hard-coded propagation format. And inject XRay propagator as helper in AWS SDK instrumentation.

Assuming my manual commands to do so worked OK, confirmed with my AWS SDK integration that the instrumentation works.

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
+++ b/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
@@ -30,7 +30,8 @@ public class AwsSdkInstrumentationModule extends InstrumentationModule {
       "io.opentelemetry.instrumentation.awssdk.v2_2.RequestType",
       "io.opentelemetry.instrumentation.awssdk.v2_2.SdkRequestDecorator",
       "io.opentelemetry.instrumentation.awssdk.v2_2.DbRequestDecorator",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.TracingExecutionInterceptor"
+      "io.opentelemetry.instrumentation.awssdk.v2_2.TracingExecutionInterceptor",
+      "io.opentelemetry.extension.trace.propagation.AwsXRayPropagator"
     };
   }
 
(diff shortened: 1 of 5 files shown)


javaagent-exporters/javaagent-exporters.gradle
```diff
@@ -67,4 +67,5 @@ shadowJar {
   relocate "io.opentelemetry.api", "io.opentelemetry.javaagent.shaded.io.opentelemetry.api"
   relocate "io.opentelemetry.spi", "io.opentelemetry.javaagent.shaded.io.opentelemetry.spi"
   relocate "io.opentelemetry.context", "io.opentelemetry.javaagent.shaded.io.opentelemetry.context"
+  relocate "io.opentelemetry.extension.trace.propagation", "io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.trace.propagation"
```
trask, line 70: if updating here, should probably update in `ExporterClassLoader` also for consistency

trask (approved):

iNikem:
> And inject XRay propagator as helper in AWS SDK instrumentation.

I don't understand, shouldn't you actually use XRay propagator somewhere in this PR?

instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
```diff
@@ -30,7 +30,8 @@ public AwsSdkInstrumentationModule() {
       "io.opentelemetry.instrumentation.awssdk.v2_2.RequestType",
       "io.opentelemetry.instrumentation.awssdk.v2_2.SdkRequestDecorator",
       "io.opentelemetry.instrumentation.awssdk.v2_2.DbRequestDecorator",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.TracingExecutionInterceptor"
+      "io.opentelemetry.instrumentation.awssdk.v2_2.TracingExecutionInterceptor",
+      "io.opentelemetry.extension.trace.propagation.AwsXRayPropagator"
```
anuraaga, line 34: @iNikem Here

iNikem: I still don't understand :( Here you _allow_ it to be used, by injecting it into user's classloader. Where or how do you actually use it?

anuraaga: Ah thought you remember from your PR comment before :P We use it here

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java#L112

But it currently fails with `ClassNotFoundException` because I forgot to add the helper. Would be caught in tests after #1643.

iNikem (approved):

PR #1893, opened 2020-12-13
HTTP client instrumentation cleanup, part 0

trask:
Splitting up #1840.

This PR makes a few small changes to `BaseTracer` and `HttpClientTacer` that other parts use.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -112,6 +116,15 @@ public void endExceptionally(Context context, Throwable throwable) {
     super.endExceptionally(span, throwable, -1);
   }
 
+  /** Convenience method primarily for bytecode instrumentation. */
+  public void endMaybeExceptionally(Context context, RESPONSE response, Throwable throwable) {
```
anuraaga, line 120: I'd put this method in a PR with at least one usage of it - inClientSpan / withClientSpan are quite clear here, but not this method which isn't used. One way to keep PRs small is to change just one usage in one PR, and remaining usages in a mechanical way in another.

trask: added example

anuraaga: I think a while ago we had a thought that this pattern is technically the most precise

```java
Throwable err = null;
try (context.makeCurrent()) {
  doWork();
} catch (Throwable t) {
  err = t;
} finally {
  span.end
}
```

If so I guess this class isn't only useful for bytecode manipulation - do you think we can remove the two variants and merge into a single `end` method eventually?

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
anuraaga, line 120:
```suggestion
  public void endMaybeExceptionally(Context context, RESPONSE response, @Nullable Throwable throwable) {
```

trask: Added TODO comment on this method to revisit

iNikem (approved):

mateuszrzeszutek (approved):

PR #1898, opened 2020-12-14
Don't allow to run SDK initialisation multiple times

pavolloffay:

This PR adds check to `AgentInitializer` to skip initialization if the agent has been already loaded. This is useful in dynamic attach as protection to install the agent only once.  The `startAgent` has already this check, but the SDK does not so it can be installed multiple times.

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
@@ -47,6 +47,11 @@ public class AgentInitializer {
   private static ClassLoader AGENT_CLASSLOADER = null;
 
   public static void initialize(Instrumentation inst, URL bootstrapUrl) {
+    if (AGENT_CLASSLOADER != null) {
+      // don't run initialization multiple times
+      return;
+    }
+
     startAgent(inst, bootstrapUrl);
 
     boolean appUsingCustomLogManager = isAppUsingCustomLogManager();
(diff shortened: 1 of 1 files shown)


javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -47,6 +47,11 @@
   private static ClassLoader AGENT_CLASSLOADER = null;
 
   public static void initialize(Instrumentation inst, URL bootstrapUrl) {
+    if (AGENT_CLASSLOADER != null) {
+      // don't run initialization multiple times
```
trask, line 51:
```suggestion
      // don't run initialization multiple times, initialize can be called multiple times via dynamic attach
```

trask (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -47,6 +47,11 @@
   private static ClassLoader AGENT_CLASSLOADER = null;
 
   public static void initialize(Instrumentation inst, URL bootstrapUrl) {
+    if (AGENT_CLASSLOADER != null) {
+      // don't run initialization multiple times, initialize can be called multiple times via dynamic attach
```
anuraaga, line 51: Just confirming, do we need the field to be volatile if accessed multiple times? With dynamic attach, is `initialize` guaranteed to read the memory from a previous update?

trask: I would guess we get no such guarantee, it's only accessed a few times, so little harm making it volatile, good call 👍

anuraaga: Alternatively if we cared about affecting the other reads of the variable on first init, we could add an `AtomicBoolean` only for checking reinitialization.

pavolloffay:
This PR might not be needed if https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1848/files#diff-348133242eecb9210e0c2679ef4893a1352811498dd11a93b977ef412f8dacfaL51 goes in

pavolloffay:
not needed, it will be done in #1848 by running only agentInstaller

Issue #894, opened 2020-08-05
http.status_code not matching what was returned


FrankSpitulski:
possible duplicate of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/688

I've managed to get a 401 response while the span's tag says http.status_code is 500. Please see the [README.md](https://github.com/FrankSpitulski/rest-api-example/blob/incorrect-status-code-tag/README.md) in my example.

https://github.com/FrankSpitulski/rest-api-example/tree/incorrect-status-code-tag

FrankSpitulski:
I did some debugging and found that this is a result of the response being uncommitted at the time that the end method is called causing the status code to be set to 500.

FrankSpitulski:
The Servlet3Advice.stopSpan assumes that if a response has a throwable then it should end the span right there, but shouldn't it wait until the response is committed before trying to end the span and record the status code?

trask:
Any ideas what we could instrument to capture when the response is committed?

trask:
oh, are you using `0.7.0`?

It looks like I added `isCommitted` check just before `0.7.0` was released:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v0.7.0/instrumentation/servlet/servlet-3.0/src/main/java/io/opentelemetry/auto/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java#L52

I think that check should prevent you from seeing `500` when the response hasn't been committed at that point

FrankSpitulski:
@trask sorry, should have mentioned. I did my most recent debugging session today with latest master, so hash 09c597e8.

FrankSpitulski:
I was just about to try setting the parent's status code from the servlet send response child span and not setting the default 500. That kind of breaks encapsulation though, but if the servlet method already exited and the response is not async I don't see another choice.

trask:
oh, ignore my comments above, reading comprehension 🙁

> this is a result of the response being uncommitted at the time that the end method is called causing the status code to be set to 500

yes, this is (mostly) intentional, because apps without error handlers will return `500` when an exception is thrown from the servlet

maybe instead we should not set _any_ `status_code` in this case, since we don't really know? we are capturing the exception, so there's some indication of failure on the span already

> I was just about to try setting the parent's status code from the servlet send response child span and not setting the default 500

if we don't call `Span.end()` at the end of a synchronous servlet, i'm not sure where we would call it?

[...]

Issue #1886, opened 2020-12-11
Server span handling is inconsistent


laurit:
Servlet integration can only capture requests that pass though a servlet filter or servlet. So on tomcat, which relies on servlet integration, request to paths that are outside of deployed application are not captured. On jetty where integration triggers before request reaches servlet such requests are captured. Similarly on liberty where static content is not served by a servlet, requests to static files are not captured when only servlet-4.0 feature is enabled, but are captured when javaee-8.0 feature is enabled because javaee-8.0 adds a servlet filter to the application. We should add integrations that handle server spans similarly to jetty for other popular app-servers. But in order to do that we should agree on how to name such server spans consistently across integrations.

Servlet integration sets span name to request.getContextPath() + request.getServletPath(). Jetty integration sets span name to HandlerCollection.handle for jetty9 and HandlerList.handle for jetty10. Could anybody elaborate what is the desired value for span name? Should it be changed so that server spans would have a consistent naming?

vovencij:
In general, the solution that is done for Jetty integration doesn't scale to everything else – jetty integration doesn't instrument some specific handler classes that process Servlets, so that servlet integration will fully handle requests to servlets and jetty integration will act as a fallback. Tomcat, for example doesn't have such architecture where it would be possible to not instrument some specific servlet-related classes to separate such cases. More scalable solution would be to allow for servlet handler to just update the name of the server span (which maybe later updated by more specific web-framework integration) in case it is already in the context.

iNikem:
Is my understanding below is correct?

There are HTTP servers, such as Jetty or Undertow, which provide their own framework of "Handlers" to response to incoming http requests. Some of built-in handlers actually implement Servlet API and delegate to some servlets. Current instrumentations either focus solely on Servlet API (e.g. Tomcat) or try to have two disjoint instrumentations: for Servlets and for everything else (e.g. Jetty).

This proposal is to have first class instrumentations for HTTP servers which would create SERVER spans for all intercepted requests, regardless if they later end up in Servlets or not. And our Servlet instrumentations will just update name and set additional attributes to already existing SERVER span. Similar to how e.g. Spring MVC instrumentation does.

If I understood you correctly, then I think this is great idea :)

iNikem:
[...]

PR #1902, opened 2020-12-14
Tomcat server handlers

vovencij:
This is a Tomcat version of a server handler, as described and discussed in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1886

Some notable findings:

- ~~Server instrumentation needs uncaught exceptions from outermost servlets/filters. These are passed via `AppServerBridge.THROWABLE_CONTEXT_KEY` context attribute.~~ Outermost servlets/filters set their unhandled exceptions with the current span.
- Exception handling in tomcat is done in a way, where a request passes the servlet stack two times – first pass is the initial request and the second pass is when tomcat tries to render an error page mapped in web.xml. Server span name should be updated only when doing a first pass. Thus the `AppServerBridge` context attribute and not the intuitive `callDepth` counter.


instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
```diff
@@ -0,0 +1,94 @@
...
+      return;
+    }
+    scope.close();
+
+    if (context == null) {
+      return;
+    }
+
+    Throwable throwableToReport =
+        throwable != null ? throwable : AppServerBridge.getUnhandledThrowable(context);
```
iNikem, line 60: Why do you need this `AppServerBridge.getUnhandledThrowable`? You can record the exception on SERVER span write when it happened inside servlet instrumentation. You don't need to pass it up.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/TomcatSmokeTest.groovy
```diff
@@ -0,0 +1,46 @@
...
+
+class TomcatSmokeTest extends AppServerTest {
+
+  protected String getTargetImage(int jdk, String serverVersion) {
+    "ghcr.io/open-telemetry/java-test-containers:tomcat-${serverVersion}-jdk$jdk-20201207.405832649"
+  }
+
+  def "Server Handler test"() {
+    setup:
+    startTarget(8, "7.0.107")
```
trask, line 19: is this meant to be parameterized by the test params below?

instrumentation/tomcat-7.0/javaagent/tomcat-7.0-javaagent.gradle
```diff
@@ -0,0 +1,17 @@
...
+    group = "org.apache.tomcat.embed"
+    module = "tomcat-embed-core"
+    // Tomcat 10 is about servlet 5.0
+    // 7.0.4 added Request.isAsync, which is needed
+    versions = "[7.0.4, 10)"
+  }
+}
+
+dependencies {
+  library group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '9.0.40'
```
trask, line 14: typically this is the lowest version, and then tests will be run against the latest version in nightly build using -PtestLatestDeps=true, in this case you don't quite want the latest version, so need to supply `latestDepTestLibrary` like below
```suggestion
  library group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '7.0.4'
  latestDepTestLibrary  group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '9.+'
```

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
```diff
@@ -0,0 +1,94 @@
...
+        try {
+          servletRequest
+              .getAsyncContext()
+              .addListener(new TagSettingAsyncListener(responseHandled, context));
+        } catch (IllegalStateException e) {
+          // thrown by tomcat if request was already handled while setting the listener.
+          tracer().end(context, response);
+        }
+      }
+    }
```
trask, line 92: Jetty and Servlet advice also does this, do you think applicable here also?
```suggestion

      // Check again in case the request finished before adding the listener.
      if (!request.isAsync() && responseHandled.compareAndSet(false, true)) {
        tracer().end(context, response);
      }
    }
```

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
```diff
@@ -0,0 +1,94 @@
...
+
+    Throwable throwableToReport =
+        throwable != null ? throwable : AppServerBridge.getUnhandledThrowable(context);
+
+    if (throwableToReport != null) {
+      if (response.isCommitted()) {
+        tracer().endExceptionally(context, throwableToReport, response);
+      } else {
+        tracer().endExceptionally(context, throwableToReport);
```
trask, line 66: can you add a comment here about why not passing response when it's not committed? it's not obvious to me at least

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
```diff
@@ -0,0 +1,94 @@
...
+      @Advice.Local("otelScope") Scope scope) {
+    if (scope == null) {
+      return;
+    }
+    scope.close();
+
+    if (context == null) {
+      return;
+    }
+
```
trask, line 58: no need for this null check I think
```suggestion
```

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
```diff
@@ -32,10 +36,18 @@ public static void onEnter(
...
         scope = attachedContext.makeCurrent();
       }
 
+      // We're interested only in the very first suggested name, as this is where the initial
+      // request arrived. There are potential forward and other scenarios, where servlet path
+      // may change, but we don't want this to be reflected in the span name.
+      if (!AppServerBridge.isBetterNameSuggested(attachedContext)) {
+        tracer().updateServerSpanName(httpServletRequest);
+        AppServerBridge.setBetterNameSuggested(attachedContext, true);
+      }
```
trask, line 49: the SERVER span name is also updated in several of the web framework instrumentations

trask (commented):
thanks @vovencij! do you think this will resolve #894?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Helper containers for the Context attributes for transferring certain information between servlet
+ * integration and app-server server handler integrations.
+ */
+public class AppServerBridge {
+
+  public static final ContextKey<AtomicReference<Throwable>> THROWABLE_CONTEXT_KEY =
```
anuraaga, line 19: It's better if we can have a single key / object with multiple fields for any particular concern, rather than multiple keys

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
vovencij: Web Framework instrumentations are expecting the SERVER span to be present and to update the server span name unconditionally. Servlet integration was either _creating_ a new SERVER span or exiting as soon as possible if a SERVER span was already present. This flag is actually needed for the Servlet instrumentation only to indicate that the servlet instrumentation itself has set the name. I probably should rename methods to better reflect what's happening.

vovencij:
> do you think this will resolve #894?

I think so. This is the trace I get from the demo case in that ticket in my branch:
image

And there is also an unhandled servlet exception reported in the trace:
image

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
vovencij: If the response is not yet committed, then response headers, including response code, are not yet written to the output stream. And the only reason response is passed to the `endExceptionally` is to read the response code.

.... And after looking at the https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/894 and into `endExceptionally` I see that in this case `http.status_code` will be set to 500 even if it wasn't actually sent. Maybe this has to be fixed also, but I'm reluctant to change that logic in the current PR.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
```diff
@@ -0,0 +1,83 @@
...
+
+  public static Context init(Context ctx) {
+    return ctx.with(AppServerBridge.THROWABLE_CONTEXT_KEY, new AtomicReference<>())
+        .with(AppServerBridge.SERVLET_SUGGESTED_BETTER_NAME_KEY, new AtomicBoolean(false));
+  }
+
+  /**
+   * This is for servlet instrumentation to record exceptions unhandled by servlets and filters,
+   * that app-server integrations are interested in (to fail the span exceptionally with attached
+   * throwable), but which may be swallowed by an app-server before execution arrives to an
```
iNikem, line 32: Why exactly do we want to record exception in SERVER span if that was swallowed inside it?

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
iNikem: This comment better be in the code :)

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
```diff
@@ -32,10 +36,18 @@ public static void onEnter(
 
     Context attachedContext = tracer().getServerContext(httpServletRequest);
     if (attachedContext != null) {
-      if (tracer().needsRescoping(attachedContext)) {
+      if (Servlet3HttpServerTracer.needsRescoping(attachedContext)) {
```
iNikem, line 39: Why this change?

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
vovencij: Good point! Looks that this works. Will change.

iNikem (commented):
I don't like `AppServerBridge` at all. I have no better suggestions right now, but I will think about it...

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
vovencij: `needsRescoping` is a static method. IntelliJ suggested.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
vovencij: I changed this part as you have suggested earlier - record throwable from the outermost servlet/filter instead of storing it in the context. Will push soon.

vovencij: > I don't like `AppServerBridge` at all. I have no better suggestions right now, but I will think about it...

After removing throwables from this bridge, the only thing it has left is the flag for the updated name. The flag is only used by the servlet instrumentation, but it must have the same lifetime as the context created by the app server, so this bridge must be added to the context by the server instrumentation. You are welcome to suggest a better name or  more elegant approach altogether.

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
vovencij: Coming soon...

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
vovencij: Fixed.

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatTracer.java
```diff
@@ -0,0 +1,124 @@
...
+
+    // context must be reattached, because it has new attributes compared to the one returned from
+    // startSpan().
+    attachServerContext(context, request);
+    return context;
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "tomcat";
```
laurit, line 42: Should be "io.opentelemetry.javaagent.tomcat" to be consistent with other integrations

instrumentation/tomcat-7.0/javaagent/tomcat-7.0-javaagent.gradle
vovencij: Actually I don't need `tomcat-embed-core` for the instrumentation. I only need it for the test. For the instrumentation I need two tomcat libraries - `org.apache.tomcat:tomcat-coyote` (handling of HTTP) and `org.apache.tomcat:tomcat-catalina` (Servlet implementation). However, these two dependencies must have the same version while muzzling (because some classes moved between those Jars at some point and there are `NoClassDefFound` errors from either one or other jar). Is there a way to configure muzzle in a way that two dependencies would have the same version at a time?

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/TomcatSmokeTest.groovy
vovencij: Quick answer - eventually yes, but not right now 😄 

Ideal solution eventually would be for the matrix to work as follows:

```
SmokeTestClass {
  forEach (configuration : matrix) {
    startTarget(configuration)
    testScenario1()
    testScenario2()
    ....
    tearDown(configuration)
 }
}
```

As a result there will be several separate test cases for each configuration in the matrix and the app server is started once per matrix entry.

Currently it is only possible to have either:

a) each `testScenarioX` will call `startTarget` which will effectively mean that each app server is started as many times as there are test cases. In case of this particular smoke test it would start 5 more tomcats.

or

b) There is one `testScenario` which contains all asserts in the world and it would be impossible to distinguish between failing tests.

The problem is I don't know yet how to achieve the ideal solution using neither Spock or JUnit.

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatTracer.java
vovencij: Yes, fixed.

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
vovencij: Removed.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
```diff
@@ -0,0 +1,65 @@
...
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Helper container for Context attributes for transferring certain information between servlet
+ * integration and app-server server handler integrations.
+ */
+public class AppServerBridge {
```
iNikem, line 16: I now seems to me that this whole logic can live in `ServletHttpServerTracer`. This will also remove some ifs from servlet advices.

instrumentation/tomcat-7.0/javaagent/tomcat-7.0-javaagent.gradle
trask: > Is there a way to configure muzzle in a way that two dependencies would have the same version at a time?

I don't think so. The way you have it currently using the aggregating artifact `tomcat-embed-core` seems reasonable to me.

trask (approved):
👍

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -126,4 +131,12 @@ private static String getSpanName(HttpServletRequest request) {
     }
     return spanName;
   }
+
+  public void updateServerSpanNameOnce(Context attachedContext, HttpServletRequest request) {
+    if (AppServerBridge.isPresent(attachedContext)
```
iNikem, line 136: Is this `isPresent` needed here? The result of `if` is the same without it

instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2Advice.java
```diff
@@ -24,14 +26,20 @@ public static void onEnter(
...
     if (!(request instanceof HttpServletRequest) || !(response instanceof HttpServletResponse)) {
       return;
     }
 
     HttpServletRequest httpServletRequest = (HttpServletRequest) request;
 
-    if (tracer().getServerContext(httpServletRequest) != null) {
+    Context serverContext = tracer().getServerContext(httpServletRequest);
+    if (serverContext != null) {
+      // We're interested only in the very first suggested name, as this is where the initial
```
iNikem, line 39: Shouldn't this comment go to `updateServerSpanNameOnce` method declaration site?

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+  public static void stopSpan(
+      @Advice.Argument(0) Request request,
+      @Advice.Argument(1) Response response,
+      @Advice.Thrown Throwable throwable,
+      @Advice.Local("otelContext") Context context,
+      @Advice.Local("otelScope") Scope scope) {
+    scope.close();
```
iNikem, line 49: `scope` can be null here

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
iNikem: @vovencij ?

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatTracer.java
```diff
@@ -0,0 +1,124 @@
...
+    implements TextMapPropagator.Getter<Request> {
+
+  private static final Logger log = LoggerFactory.getLogger(TomcatTracer.class);
+  private static final TomcatTracer TRACER = new TomcatTracer();
+
+  public static TomcatTracer tracer() {
+    return TRACER;
+  }
+
+  public Context startServerSpan(Request request, String adapterClassName) {
```
iNikem, line 30: Pass the whole `Method` here instead of `adapterClassName`. There is `getSpanName(Method)`

instrumentation/tomcat-7.0/javaagent/tomcat-7.0-javaagent.gradle
```diff
@@ -0,0 +1,22 @@
...
+    // Tomcat 10 is about servlet 5.0
+    // 7.0.4 added Request.isAsync, which is needed
+    versions = "[7.0.4, 10)"
+  }
+}
+
+dependencies {
+  library group: 'org.apache.tomcat.embed', name: 'tomcat-embed-core', version: '7.0.4'
+  implementation project(':instrumentation:servlet:servlet-3.0:javaagent')
+  implementation("org.slf4j:slf4j-api:1.7.30")
```
iNikem, line 16: I wonder, why no other instrumentation defines this dependency?

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
vovencij: `isPresent` is for the cases, where there is no app-server specific tracer (and thus no AppServerBridge in the context) which actually needs this update. Without this check camel tests, for example, will fail because the name changes unexpectedly.

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/ServerHandlerAdvice.java
vovencij: Checked Tomcat's source code, the check is relevant for older tomcats (7). Newer tomcats (e.g., 9) will throw IllegalStateException in this case, which is handled above. Added the check.

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatTracer.java
vovencij: But I don't have the method 🤔

iNikem: But you can have :) Like servlet advices do

instrumentation/tomcat-7.0/javaagent/tomcat-7.0-javaagent.gradle
vovencij: removed

instrumentation/tomcat-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatTracer.java
vovencij: Now you lost me. Servlet advices set server span name from the URL obtained from the Request. What method are you talking about?

iNikem: Take a look at `io.opentelemetry.javaagent.instrumentation.servlet.http.HttpServletResponseInstrumentationModule.SendAdvice`

vovencij: ok

iNikem (approved):

PR #1900, opened 2020-12-14
Fix import for @WithSpan in docs.

breedx-splk:
Users will be copy/pasting examples and will encounter problems.  Let's stick with the new package name in the docs too.

PR #1903, opened 2020-12-14
Use correct package

DaGeRe:
In the main readme, the package `import io.opentelemetry.extensions.auto.annotations.WithSpan;` is wrong, it needs to be `io.opentelemetry.extension.annotations.WithSpan` (extension without s).


trask:
thanks @DaGeRe! this just got fixed here #1900, definitely send in any other problems you find!

PR #1907, opened 2020-12-15
HTTP client instrumentation cleanup: apache-httpclient

trask:
Split out from #1840


instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/CommonsHttpClientTracer.java
```diff
@@ -21,6 +22,10 @@ public static CommonsHttpClientTracer tracer() {
     return TRACER;
   }
 
+  public Context startSpan(Context parentContext, HttpMethod request) {
+    return super.startSpan(parentContext, request, request);
+  }
```
trask, line 27: The goal is to make the super.startSpan() `protected`, and only expose simplified startSpan to users of the Tracer

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ContextScopePair.java
```diff
@@ -1,27 +0,0 @@
...
- * Copyright The OpenTelemetry Authors
- * SPDX-License-Identifier: Apache-2.0
- */
-
-package io.opentelemetry.javaagent.instrumentation.apachehttpclient.v4_0;
-
-import io.opentelemetry.context.Context;
-import io.opentelemetry.context.Scope;
-
-public class ContextScopePair {
```
trask, line 11: This class was unused already, just deleting it here

anuraaga (approved):

iNikem (approved):

PR #1908, opened 2020-12-15
HTTP client instrumentation cleanup: http-url-connection

trask:
Split out from #1840

### DIFF
--- a/instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
+++ b/instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
@@ -18,7 +18,6 @@ import static net.bytebuddy.matcher.ElementMatchers.named;
 import static net.bytebuddy.matcher.ElementMatchers.not;
 
 import com.google.auto.service.AutoService;
-import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.javaagent.instrumentation.api.CallDepth;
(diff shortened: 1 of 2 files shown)


anuraaga (approved):

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
```diff
@@ -79,75 +81,65 @@ public HttpUrlConnectionInstrumentationModule() {
...
-            // responseCode field is sometimes not populated.
-            // We can't call getResponseCode() due to some unwanted side-effects
-            // (e.g. breaks getOutputStream).
-            if (responseCode > 0) {
-              tracer().end(state.context, new HttpUrlResponse(connection, responseCode));
-              state.finished = true;
-            }
-          }
+      scope.close();
+      if (Span.fromContext(context).isRecording()) {
```
iNikem, line 133: We don't have similar condition in other instrumentations, right? Why do we need it here?

mateuszrzeszutek: `isRecording()` is the same as previous `!state.finished`. No idea why it was used here though

mateuszrzeszutek (approved):

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
trask: It's needed here because HttpURLConnection is a messy API and so instrumenting it is messy.

But I do think of checking `isRecording` as more of an optimization for avoiding costly things when Sampling is enabled, so think the usage here is a bit confusing, so brought back `HttpUrlState.finished` to track this (and added more comments).

iNikem (approved):

PR #1906, opened 2020-12-15
Hide spring non-spec'd attributes behind flag

trask:
Introduces
* `otel.instrumentation.spring-webflux.experimental-span-attributes`
* `otel.instrumentation.spring-webmvc.experimental-span-attributes`

PR #1915, opened 2020-12-15
Fix spring webflux tests

laurit:
Move setup from static initializer to setupSpec to ensure that property is set for both SpringWebfluxTest and SingleThreadedSpringWebfluxTest not only the one that runs first. Currently the one that runs first clears property in cleanupSpec and the other test fails.
This is a regression from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1906


iNikem (approved):

mateuszrzeszutek (approved):

trask (approved):
Thanks @laurit!

PR #1917, opened 2020-12-16
Add support for Undertow server runtime

iNikem:
Part of #1886 .

Changes in servlets should be replaced by those from #1902 when that is merged.


instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/undertow/UndertowInstrumentationModule.java
```diff
@@ -0,0 +1,98 @@
...
+      @Advice.OnMethodEnter(suppress = Throwable.class)
+      public static void onEnter(
+          @Advice.Origin Method method,
+          @Advice.Argument(value = 0, readOnly = false) HttpServerExchange exchange,
+          @Advice.Local("otelScope") Scope scope) {
+        Context attachedContext = tracer().getServerContext(exchange);
+        if (attachedContext != null) {
+          if (!Java8BytecodeBridge.currentContext().equals(attachedContext)) {
+            scope = attachedContext.makeCurrent();
+          }
```
trask, line 75: I don't think we generally perform this check (maybe we should?)
```suggestion
          scope = attachedContext.makeCurrent();
```

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/undertow/UndertowInstrumentationModule.java
```diff
@@ -0,0 +1,98 @@
...
+public class UndertowInstrumentationModule extends InstrumentationModule {
+
+  public UndertowInstrumentationModule() {
+    super("undertow", "undertow-2.0");
+  }
+
+  @Override
+  public ElementMatcher.Junction<ClassLoader> classLoaderMatcher() {
+    return hasClassesNamed("io.undertow.server.HttpHandler");
+  }
```
trask, line 42: is this needed? (see javadoc on `classLoaderMatcher`)

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/api/KeyHolder.java
```diff
@@ -0,0 +1,35 @@
...
+ * available to all tracers from every classloader.
+ *
+ * <p>But at the same time, being loaded by bootstrap classloader, this class itself cannot initiate
+ * the loading of {@link AttachmentKey} class. Class has to be loaded by <i>any</i> classloader that
+ * has it, e.g. by the classloader of a Tracer that uses this key holder. After that, <i>all</i>
+ * Tracers, loaded by all classloaders, will be able to use exactly the same sole instance of the
+ * key.
+ */
+public class KeyHolder {
+  public static final AtomicReference<AttachmentKey<Context>> contextKey = new AtomicReference<>();
```
trask, line 34: two thoughts:
* a bootstrap class can't really reference the AttachmentKey class, the reason this is working is because of erasure, but think better to not hide that, and use Object here in the signature
* I think better to use `Map<ClassLoader, Object>` (where the value is AttachmentKey), in case undertow _itself_ is loaded by two different class loaders, in which case you really do need two different AttachmentKey instances. The key of the map would be `AttachmentKey.class.getClassLoader()`.

anuraaga: How about a map from class to object to allow scaling this to any type? `Map<Class<?>, Object>`. And since classes from different classloaders aren't equal (I presume), wouldn't need `ClassLoader` in the key I believe.

trask: ya, I like that, it's approximately "contextStore" for a single static field

iNikem: Why do I want two different key instance if Undertow is loaded from two different classloaders? And how is the latter possible?

Important point is that any given request must use the same key all the time it passes through Undertow and WildFly.

trask: It's not going to happen in wildfly, since wildfly provides undertow in its root class loader, but it's _possible_, e.g. say you had two tomcat apps that each included undertow library.

And if you have two class loaders each with their own copy of undertow classes, and if you store an `AttachmentKey` from one class loader into the `AtomicReference`, and read it back out from the other class loader and cast it to `AttachmentKey` you will get `ClassCastException`.

iNikem: Now I understand :) Thank you for your patience :)

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/undertow/UndertowInstrumentationModule.java
iNikem: This is similar to `rescoping` check in Servlet3. If there is server context attached to the request and it differs from the current context, then we rescope into server context. If they are the same, no point in calling `makeCurrent`, is it?

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/undertow/UndertowInstrumentationModule.java
iNikem: Moved to `HandlerInstrumentation` as `classLoaderOptimization`

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/api/KeyHolder.java
iNikem: Done

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/undertow/UndertowInstrumentationModule.java
trask: I thought you were always on the side of less code in the Advice classes 😄

trask (approved):
👍

instrumentation/undertow/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/undertow/UndertowInstrumentationModule.java
iNikem: I am, but I still want to prevent unnecessary `makeCurrent` :)

PR #1918, opened 2020-12-16
Muzzle should add SPI classes defined in helperResourceNames as references

### LINKED ISSUE #1887: Muzzle should add SPI classes defined in helperResourceNames as references
Currently muzzle does not look at the injected resource files at all, so when we inject service provider files (like in log4j2 instrumentation) to make helper injection & muzzle validation work we need to add an empty advice class with a `muzzleCheck()` method. This is ugly and the javaagent behavior is non-obvious.

Muzzle compile time reference collection process could inspect files returned by the `helperResourceNames()` method and if the match one of the following patterns load classes defined in them as references.
* `META-INF/services/.*` - `ServiceLoader` SPI file pattern
* `/software/amazon/awssdk/services/[^/]+/execution.interceptors` - AWS SDK execution interceptors SPI file pattern

**Additional context**
This would prevent situations like https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1876 from happening

mateuszrzeszutek:
Resolves #1887

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
+++ b/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
@@ -22,17 +22,7 @@ public class AwsSdkInstrumentationModule extends InstrumentationModule {
 
   @Override
   public String[] additionalHelperClassNames() {
-    return new String[] {
-      "io.opentelemetry.javaagent.instrumentation.awssdk.v2_2.TracingExecutionInterceptor",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdk",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkHttpClientTracer",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkInjectAdapter",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.RequestType",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.SdkRequestDecorator",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.DbRequestDecorator",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.TracingExecutionInterceptor",
-      "io.opentelemetry.extension.trace.propagation.AwsXRayPropagator"
-    };
+    return new String[] {"io.opentelemetry.extension.trace.propagation.AwsXRayPropagator"};
   }
 
   /**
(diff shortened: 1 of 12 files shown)


instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
```diff
@@ -22,17 +22,7 @@ public AwsSdkInstrumentationModule() {
...
-      "io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdk",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkHttpClientTracer",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkInjectAdapter",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.RequestType",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.SdkRequestDecorator",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.DbRequestDecorator",
-      "io.opentelemetry.instrumentation.awssdk.v2_2.TracingExecutionInterceptor",
-      "io.opentelemetry.extension.trace.propagation.AwsXRayPropagator"
-    };
+    return new String[] {"io.opentelemetry.extension.trace.propagation.AwsXRayPropagator"};
```
anuraaga, line 25: Thanks a lot for the change!!! Just checking, will muzzle task fail if this class isn't included now?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollector.java
```diff
@@ -7,54 +7,103 @@
...
+   * @param resource path to the resource file, same as in {@link ClassLoader#getResource(String)}
+   * @see io.opentelemetry.javaagent.tooling.muzzle.InstrumentationClassPredicate
+   */
+  public void collectReferencesFromResource(String resource) {
+    if (!isSpiFile(resource)) {
+      return;
+    }
+
+    List<String> spiImplementations = new ArrayList<>();
+    try (InputStream stream = getResourceStream(resource)) {
```
anuraaga, line 65: Is it possible to use `Files.lines` or `Files.readAllLines`?

anuraaga (approved):

instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java
mateuszrzeszutek: Yes, it will:

```
FAILED MUZZLE VALIDATION: io.opentelemetry.javaagent.instrumentation.awssdk.v2_2.AwsSdkInstrumentationModule mismatches:
-- io.opentelemetry.instrumentation.awssdk.v2_2.AwsSdkHttpClientTracer:39 Missing class io.opentelemetry.extension.trace.propagation.AwsXRayPropagator
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollector.java
mateuszrzeszutek: I wanted to use one of those but they all accept `Path` - and we pretty much have to use `InputStream` here because of `connection.setUseCaches(false)`

trask (approved):
this is great! will really encourage us to instrument via official SPI hooks more 👍

PR #1924, opened 2020-12-17
Update SDK dependency to 0.13.0-SNAPSHOT.

anuraaga:


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/AgentTestRunner.java
```diff
@@ -344,29 +339,4 @@ protected static String getClassName(Class clazz) {
     }
     return className;
   }
-
-  // Workaround https://github.com/open-telemetry/opentelemetry-java/pull/2096
-  public static void setGlobalPropagators(ContextPropagators propagators) {
```
trask, line 349: 👍

trask (approved):
thanks for cleaning up the metrics code usage too, looks better 👍

gradle/dependencies.gradle
```diff
@@ -42,17 +42,17 @@ ext {
...
-    opentelemetryOtlp          : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporter-otlp', version: versions.opentelemetryOther),
-    opentelemetryZipkin        : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporter-zipkin', version: versions.opentelemetryOther),
-    opentelemetryPrometheus    : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporter-prometheus', version: versions.opentelemetryOther),
-    opentelemetryLogging       : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-exporter-logging', version: versions.opentelemetryOther),
-    opentelemetryProto         : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-proto', version: versions.opentelemetryAnother),
-    opentelemetryResources     : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-sdk-extension-resources', version: versions.opentelemetry),
+    opentelemetryKotlin        : dependencies.create(group: 'io.opentelemetry', name: 'opentelemetry-extension-kotlin', version: '0.13.0-20201217.041750-92'),
```
iNikem, line 45: Shouldn't you use variables from above here?

gradle/java.gradle
```diff
@@ -126,14 +126,16 @@ dependencies {
     // latest snapshots are used, because the opentelemetry-sdk snapshot pom references these
     // transitive dependencies using the generic SNAPSHOT instead of a specific snapshot
     implementation group: 'io.opentelemetry', name: 'opentelemetry-api', version: versions.opentelemetry
+    implementation group: 'io.opentelemetry', name: 'opentelemetry-api-trace', version: '0.13.0-20201217.041746-88'
```
iNikem, line 129: Do we want both `opentelemetry-api-trace` and `opentelemetry-api`?

instrumentation/kafka-streams-0.11/javaagent/src/test/groovy/KafkaStreamsTest.groovy
```diff
@@ -32,6 +30,9 @@ import org.springframework.kafka.test.utils.ContainerTestUtils
 import org.springframework.kafka.test.utils.KafkaTestUtils
 import spock.lang.Shared
 
+import java.util.concurrent.LinkedBlockingQueue
```
iNikem, line 33: Something wrong with your import settings?

gradle/dependencies.gradle
anuraaga: Don't know why (maybe too many artifacts now) but the snapshot versions are all over the place now. I don't think I can come up with a more maintaible set of variables, and I guess in the future if this continues (will check once how sonatype deals with it compared to jfrog) we probably won't attempt snapshot dependencies anymore since pretty crazy to manage these.

iNikem: Then we don't need those variables?

Can we use BOM with time-based snapshots?

anuraaga: I don't think there's a way to have a BOM with the time-based snapshots (time is computed by server, not publisher).

I'm thinking of removing the variables when updating to 0.13.0 final since this got to a state that works with some yak shaving and that will be much more direct / simple.

iNikem (approved):

PR #1929, opened 2020-12-17
Liberty server span handling

laurit:
Part of #1886


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
```diff
@@ -62,4 +62,18 @@ public static void setServletUpdatedServerSpanName(Context ctx, boolean value) {
...
+
+  /**
+   * Class used as key in CallDepthThreadLocalMap for counting servlet invocation depth in
+   * Servlet3Advice and Servlet2Advice. We can not use helper classes like Servlet3Advice and
+   * Servlet2Advice for determining call depth of server invocation because they can be injected
+   * into multiple class loaders.
+   *
+   * @return class used as a key in CallDepthThreadLocalMap for counting servlet invocation depth
+   */
+  public static Class<?> getServletKey() {
```
trask, line 74: `getCallDepthKey()` may be more descriptive

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyHttpDispatcherLinkAdvice.java
```diff
@@ -0,0 +1,66 @@
...
+  @Advice.OnMethodEnter(suppress = Throwable.class)
+  public static void onEnter(
+      @Advice.This HttpDispatcherLink httpDispatcherLink,
+      @Advice.FieldValue("isc") HttpInboundServiceContextImpl isc,
+      @Advice.Local("otelContext") Context context,
+      @Advice.Local("otelScope") Scope scope) {
+    LibertyRequestWrapper lrw = new LibertyRequestWrapper(httpDispatcherLink, isc.getRequest());
+    LibertyConnectionWrapper lcw =
+        new LibertyConnectionWrapper(httpDispatcherLink, isc.getRequest());
+    context = tracer().startSpan(lrw, lcw, null, lrw.getRequestUri());
```
trask, line 35: in general we use verbose Java naming style 😄, probably `lrw` -> `requestWrapper` and `lcw` -> `connectionWrapper` would be more consistent here (and similar in a couple other places)

instrumentation/liberty/liberty/javaagent/liberty-javaagent.gradle
```diff
@@ -0,0 +1,8 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+dependencies {
```
trask, line 3: can you add a comment in this file about this library not being available via maven dependency, which is why we need the compile-stub module, and we can't do muzzle library version range validation, and we can't do normal tests and rely on smoke tests for testing?

(and copy-paste or reference the comment from liberty-dispatcher-javaagent.gradle)

instrumentation/liberty/liberty-dispatcher/javaagent/liberty-dispatcher-javaagent.gradle
```diff
@@ -0,0 +1,5 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+dependencies {
```
trask, line 3: can you add comment in here about why this is separate module from liberty module? (or can they be combined?)

trask (commented):
thanks @laurit! these app server instrumentation 😅

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherInstrumentationModule.java
```diff
@@ -0,0 +1,53 @@
...
+import io.opentelemetry.javaagent.tooling.InstrumentationModule;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.List;
+import java.util.Map;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(InstrumentationModule.class)
+public class LibertyDispatcherInstrumentationModule extends InstrumentationModule {
```
iNikem, line 23: @trask should we have `classLoaderMatcher` here somewhere? Or `classLoaderOptimization`?

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherTracer.java
```diff
@@ -0,0 +1,109 @@
...
+    return libertyResponseWrapper.getStatus();
+  }
+
+  @Override
+  public @Nullable Context getServerContext(Object o) {
+    return null;
+  }
+
+  @Override
+  protected void attachServerContext(Context context, Object o) {}
```
iNikem, line 108: Why this is not implemented?

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyHttpDispatcherLinkAdvice.java
```diff
@@ -0,0 +1,66 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+import net.bytebuddy.asm.Advice;
+
+/**
+ * Instrumenting
+ * https://github.com/OpenLiberty/open-liberty/blob/master/dev/com.ibm.ws.transport.http/src/com/ibm/ws/http/dispatcher/internal/channel/HttpDispatcherLink.java
+ * We instrument sendResponse method that is called when - no application has been deployed under
+ * requested context root - something goes horribly wrong and server responds with Internal Server
+ * Error
```
iNikem, line 22: I think this javadoc is not formatted as you intended :)

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyResponseWrapper.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.liberty.dispatcher;
+
+import com.ibm.wsspi.http.channel.values.StatusCodes;
+
+public class LibertyResponseWrapper {
```
iNikem, line 10: Why do you need a wrapper for a single value?

instrumentation/liberty/liberty/javaagent/liberty-javaagent.gradle
iNikem: @laurit Can you also document this stub approach in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/writing-instrumentation.md ? As a separate PR, if you wish.

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyInstrumentationModule.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(InstrumentationModule.class)
+public class LibertyInstrumentationModule extends InstrumentationModule {
```
iNikem, line 24: Can you please add a little README to this module? Its instrumentations are not straightforward and depend on each other. It is quite hard to follow the instrumentation path from the code.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
```diff
@@ -0,0 +1,288 @@
...
+    traceIds.size() == 1
+
+    and: "Response contains Hello"
+    responseBody.contains("Hello")
+
+    and: "There is one server span"
+    traces.countSpansByKind(Span.SpanKind.SPAN_KIND_SERVER) == 1
+
+    and: "Expected span names"
+    traces.countSpansByName('/app/hello.txt') == 1
```
iNikem, line 61: This span name indeed violates the spec's requirement for low cardinality. We have to review other app server instrumentations as well for this problem. /cc @vovencij

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
```diff
@@ -0,0 +1,288 @@
...
+      .map { it.stringValue }
+      .findAny()
+      .isPresent()
+
+    where:
+    [appServer, jdk] << getTestParams()
+  }
+
+  @Unroll
+  def "#appServer test request for WEB-INF/web.xml on JDK #jdk"(String appServer, String jdk) {
```
iNikem, line 119: Just out of curiosity: why this test is important here?

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
```diff
@@ -0,0 +1,288 @@
...
+
+import io.opentelemetry.proto.trace.v1.Span
+import okhttp3.Request
+import spock.lang.Shared
+import spock.lang.Unroll
+
+import java.util.jar.Attributes
+import java.util.jar.JarFile
+
+class LibertySmokeTest extends AppServerTest {
```
iNikem, line 16: Should some (most?) of these tests live in `AppServerTest`? And tested for all app servers?

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -123,8 +123,16 @@ protected String requestHeader(HttpServletRequest httpServletRequest, String nam
...
   }
 
-  private static String getSpanName(HttpServletRequest request) {
+  public static String getSpanName(HttpServletRequest request) {
     String spanName = request.getServletPath();
+    String pathInfo = request.getPathInfo();
+    // getServletPath() returns "" when request doesn't target any servlet or
+    // when it targets servlet mapped to /*
+    // for example websphere liberty does not use a servlet for static content so
+    // requests for static files will have "" for servlet path
```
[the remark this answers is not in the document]
laurit: It is called from LibertyUpdateSpanAdvice for all requests including the ones that don't have a servelt

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherInstrumentationModule.java
laurit: I believe it is not needed because this one matches based on exact class name. `classLoaderOptimization` usually uses `hasClassesNamed` which uses `ClassLoader.getResource` to verify that given class loader can see some type that the instrumentation depends on. For example when you want to instrument all classes that implement some interface you first check whether current class loader can see that interface, if it doesn't the assumption is that this class loader can't contain classes that would be needed to instrumented by this module. The idea behind this is that checking whether class implements some interface is currently expensive so we try to avoid it by doing a cheaper check before it, but `ClassLoader.getResource` isn't free either so if it isn't really needed i would skip it.

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherTracer.java
laurit: Because it is not needed and I din't immediately see an easy way to implement it. This advice is only used when server didn't find matching application or got an internal error. Nothing that is called within this advice should require access to the span.

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyResponseWrapper.java
laurit: To be consistent with request and connection that have wrappers. Additionally I only used liberty classes from the wrappers

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
laurit: Yes they should. I plan to do it a bit later along with changing the tests for all the servers so that each configuration starts server once and runs all the tests.
Though for this to happen it would be really helpful if all servers behaved the same way or at least there was a clear understanding what span names they should use.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
laurit: As far as I know currently all servers behave that way. This is not the only scenario where span name has high cardianality /this-is-definitely-not-there-but-there-should-be-a-trace-nevertheless also is pretty much the same.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
laurit: Paths under WEB-INF are special, server should not show content for them. At least liberty cuts off request processing rather early for WEB-INF/web.xml. Added a test to verify that we do get a span for such a request.

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherTracer.java
iNikem: That deserves documentation in the code then

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
iNikem: Yep, and that also needs fixing...

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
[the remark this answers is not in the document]
iNikem: I don't understand. In what sense "is called"? `LibertyUpdateSpanAdvice` uses `LibertyHttpServerTracer`, not servlet tracer, doesn't it?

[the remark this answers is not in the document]
laurit: `LibertyHttpServerTracer` extends `Servlet3HttpServerTracer`

instrumentation/liberty/liberty-dispatcher/javaagent/liberty-dispatcher-javaagent.gradle
laurit: I initially had some issues keeping them in the same module, but now that I tried again it worked. I still would prefer to keep them separated as they really don't interact with each other and are in separate class loaders during runtime. Maybe I initially had something using servlet api inside dispatcher code and that caused it to fail to load because servlet api isn't available for dispatcher during runtime, idk.

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -141,8 +149,15 @@ private static String getSpanName(HttpServletRequest request) {
...
         && !AppServerBridge.isServerSpanNameUpdatedFromServlet(attachedContext)) {
-      getServerSpan(request).updateName(getSpanName(request));
+      updateSpanName(request);
       AppServerBridge.setServletUpdatedServerSpanName(attachedContext, true);
     }
   }
+
+  public void updateSpanName(HttpServletRequest request) {
+    Span span = getServerSpan(request);
+    if (span != null) {
```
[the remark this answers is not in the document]
laurit: It is possible when liberty request integration didn't start span because the method from where it is supposed to start it wasn't called, but the integration that is supposed to update span name is called. I'll change it so that it won't try to update span when it wasn't started.

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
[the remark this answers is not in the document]
laurit: It can. On liberty there is no default servlet for static content, but we still call getSpanName for requests that are for static files. Similarly default servlet can be removed/disable on at least tomcat, in that case it is possible to have a filter that sees the request but no servlet.

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyHttpServerTracer.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public static LibertyHttpServerTracer tracer() {
+    return TRACER;
+  }
+
+  @Override
+  public Context startSpan(HttpServletRequest request) {
+    // using request URI as span name as server isn't ready for calling request.getServletPath()
+    // span name will be updated a bit later when calling request.getServletPath() works
+    Context context = startSpan(request, request, request, request.getRequestURI());
```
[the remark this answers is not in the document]
laurit: Changed it to `HTTP {METHOD_NAME}` as suggested

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -123,8 +123,16 @@ protected String requestHeader(HttpServletRequest httpServletRequest, String nam
...
-  private static String getSpanName(HttpServletRequest request) {
+  public static String getSpanName(HttpServletRequest request) {
     String spanName = request.getServletPath();
+    String pathInfo = request.getPathInfo();
+    // getServletPath() returns "" when request doesn't target any servlet or
+    // when it targets servlet mapped to /*
+    // for example websphere liberty does not use a servlet for static content so
+    // requests for static files will have "" for servlet path
+    if (pathInfo != null && spanName.isEmpty()) {
+      spanName = pathInfo;
```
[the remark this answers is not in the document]
laurit: If `request.getServletPath()` is empty then set span name to `HTTP {METHOD_NAME}` as suggested.
Note that `request.getServletPath()` isn't always low cardinality.
If servlet is mapped to / then `getServletPath` for request to /foo/bar/baz is /foo/bar/baz. Default servlet that serves static content is typically mapped to /
If servlet is mapped to *.html then `getServletPath` for request to /foo/bar/baz.html is /foo/bar/baz.html.

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherTracer.java
laurit: comment added

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyInstrumentationModule.java
laurit: Added javadoc to LibertyInstrumentationModule

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertySmokeTest.groovy
laurit: Changed span name to `HTTP {METHOD_NAME}`

instrumentation/liberty/liberty/javaagent/liberty-javaagent.gradle
laurit: doc for using stubs added

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyInstrumentationModule.java
```diff
@@ -20,6 +20,20 @@
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 
+/**
+ * Instrumenting request handling in liberty.
```
iNikem, line 24:
```suggestion
 * Instrumenting request handling in Liberty.
```

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyInstrumentationModule.java
```diff
@@ -20,6 +20,20 @@
...
+ *   <li>On entry to WebApp.handleRequest remember request.
+ *   <li>On call to WebApp.isForbidden (called from WebApp.handleRequest) start span based on
+ *       remembered request. We don't start span immediately at the start or handleRequest because
+ *       HttpServletRequest isn't usable yet.
+ *   <li>On exit from WebAppDispatcherContext.setPathElements (called from WebApp.handleRequest)
+ *       update span name. We don't do it before because before this method is called we can't use
+ *       HttpServletRequest.getServletPath.
+ *   <li>On exit from WebApp.handleRequest close the span.
+ * </ul>
+ */
```
iNikem, line 36: Thanks! This is really helpful. Can you add links to corresponding advice classes as well?

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/ThreadLocalContext.java
```diff
@@ -57,9 +57,13 @@ public boolean startSpan() {
   /**
    * Test whether span should be updated.
    *
-   * @return true when span should be update, false when span was already updated
+   * @return true when span should be update, false when span was already updated or span wasn't
```
iNikem, line 60:
```suggestion
   * @return true when span should be updated, false when span was already updated or span wasn't
```

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyInstrumentationModule.java
laurit: Done.

instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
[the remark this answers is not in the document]
trask: this is a great point, I opened #1954 to track

instrumentation/liberty/liberty-dispatcher/javaagent/liberty-dispatcher-javaagent.gradle
trask: I think this is a very good reason if they are in different class loaders, can you add comment in the gradle file?

trask (approved):

instrumentation/liberty/liberty-dispatcher/javaagent/liberty-dispatcher-javaagent.gradle
laurit: comment added

iNikem (approved):

iNikem:
@laurit Please rebase

laurit:
@iNikem rebased

PR #1934, opened 2020-12-18
Don't compare requested version to current version since we always wa…

anuraaga:
…nt to override our default of Java 11

This branch which I copied in from the old way of doing things doesn't make sense with the new toolchains config.


trask (approved):
❤️ and thanks @breedx-splk and @laurit for uncovering this issue!

PR #1935, opened 2020-12-18
end to end latency benchmark

bhautikpip:
Using wrk (HTTP benchmarking) end to end latency benchmark for any sample app configured with agent and collector setup with receiver. I am also planning to add support for collecting cpu and memory data using oshi instrumentation probably in next PR.

Inspired from this server integrated benchmark: https://github.com/openzipkin/zipkin/blob/master/benchmarks/src/test/java/zipkin2/server/ServerIntegratedBenchmark.java

### DIFF
--- a/benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
+++ b/benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
@@ -25,7 +25,7 @@ import org.testcontainers.utility.DockerImageName;
 import org.testcontainers.utility.MountableFile;
 
 public class E2EAgentBenchmark {
-  private static String APP_NAME = System.getenv().getOrDefault("APP_IMAGE", "ghcr.io/open-telemetry/java-test-containers:smoke-springboot-jdk8-20201204.400701583");
+  private static final String APP_NAME = System.getenv().getOrDefault("APP_IMAGE", "ghcr.io/open-telemetry/java-test-containers:smoke-springboot-jdk8-20201204.400701583");
 
   private List<GenericContainer<?>> containers;
   private static final Logger LOG = LoggerFactory.getLogger(E2EAgentBenchmark.class);
--- a/benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
+++ b/benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.e2ebenchmark;
+
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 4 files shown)


benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
```diff
@@ -0,0 +1,130 @@
...
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.lifecycle.Startables;
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.MountableFile;
+
+public class E2EAgentBenchmark {
+  private static String app = System.getenv("APP_IMAGE");
```
anuraaga, line 28:
```suggestion
  private static final String APP_NAME = System.getenv("APP_IMAGE", "ghcr.io/open-telemetry/java-test-containers:smoke-springboot-jdk8-20201204.400701583");
```

benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
```diff
@@ -0,0 +1,130 @@
...
+    containers.forEach(GenericContainer::stop);
+  }
+
+  @Test
+  void run() throws Exception {
+    runBenchmark();
+  }
+
+  private void runBenchmark() throws Exception {
+    if (app == null || app.equals("")) {
```
anuraaga, line 55: Can remove these lines with above suggestion

anuraaga (commented):
Just some nits, thanks!

benchmark-e2e/src/main/java/io/opentelemetry/e2ebenchmark/E2EAgentBenchmark.java
bhautikpip: Sorry I did not understand this suggestion. `System.gentenv()` accepts 1 `String` param right?

anuraaga: Oops - sorry `System.env().getOrDefault`

anuraaga (approved):
Thanks!

benchmark-e2e/src/main/resources/collector-config.yml
```diff
@@ -0,0 +1,34 @@
...
+service:
+  extensions: health_check
+  pipelines:
+    traces:
+      processors:
+        - batch
+      receivers:
+        - otlp
+      exporters:
+        - logging
```
trask, line 27: can we leave out the exporter, or is there a no-op exporter? in hopes of reducing the impact of running a local collector on the benchmark

trask (approved):
thanks @bhautikpip!

benchmark-e2e/src/main/resources/collector-config.yml
bhautikpip: I couldn't find no-op exporter in [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/exporter). Wouldn't leaving out receiver throws some weird exceptions on the terminal? Since we are not exporting any data out from collector.

anuraaga: I haven't tried running collector without any exporters :) But if it throws errors when there's no exporter defined than using logging seems reasonable.

anuraaga (approved):

anuraaga:
Thanks @bhautikpip!

PR #1951, opened 2020-12-21
Expose context in async apache HTTP client instrumentation

pavolloffay:
I need to access context in my instrumentation that builds on top of this one.

### DIFF
--- a/instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
+++ b/instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
@@ -150,6 +150,12 @@ public class ApacheHttpAsyncClientInstrumentation implements TypeInstrumentation
       delegate.close();
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


iNikem (approved):

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
```diff
@@ -149,6 +149,10 @@ public void resetRequest() throws IOException {
     public void close() throws IOException {
       delegate.close();
     }
+
+    public Context getContext() {
```
trask, line 153: can you add a comment here about why this method exists? I'm worried someone (e.g. me) will remove it since it's unused

trask (approved):

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
pavolloffay: +1 added a comment. 

There will be probably more instrumentations that do not expose the context which makes it impossible for extending instrumentations to access the current span.

PR #1960, opened 2020-12-23
AWS lambda - configurable flush timeout

### LINKED ISSUE #1959: AWS lamba tracing handler - configure flush timeout
Currently the handler waits max 1 second before terminating lambda execution (which freezes all threads) and can prevent spans from being successfully ingested.

The timeout value should be configurable. Then the user can decide to allow longer wait if needed by their use case.  

- higher default value (ie 10 secs) - but some users can prefer lower execution latency than higher trace fidelity

**Additional context**

kubawach:
Closes #1959 

- handler's flush timeout is configurable now
- wrapper uses env prop to set the timeout (= 1 sec default)

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/README.md
+++ b/instrumentation/aws-lambda-1.0/library/README.md
@@ -3,10 +3,10 @@
 This package contains libraries to help instrument AWS lambda functions in your code.
 
 ## Using wrappers
-To use the instrumentation, configure `OTEL_LAMBDA_HANDLER` env property to your lambda handler method in following format `package.ClassName::methodName`
+To use the instrumentation, configure `OTEL_INSTRUMENTATION_AWS_LAMBDA_HANDLER` env property to your lambda handler method in following format `package.ClassName::methodName`
 and use one of wrappers as your lambda `Handler`.
 
-In order to configure a span flush timeout (default is set to 1 second), please configure `OTEL_LAMBDA_FLUSH_TIMEOUT` env property. The value is in seconds.
+In order to configure a span flush timeout (default is set to 1 second), please configure `OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT` env property. The value is in seconds.
 
 Available wrappers:
 - `io.opentelemetry.instrumentation.awslambda.v1_0.TracingRequestWrapper` - for wrapping regular handlers (implementing `RequestHandler`)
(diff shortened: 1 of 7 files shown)


kubawach:
@anuraaga please have a look

anuraaga (approved):
Thanks!

kubawach:
@anuraaga @trask @tylerbenson please merge as soon as convenient

anuraaga:
@trask Can you doublecheck this too? Thanks!

instrumentation/aws-lambda-1.0/library/README.md
```diff
@@ -6,6 +6,8 @@ This package contains libraries to help instrument AWS lambda functions in your
 To use the instrumentation, configure `OTEL_LAMBDA_HANDLER` env property to your lambda handler method in following format `package.ClassName::methodName`
 and use one of wrappers as your lambda `Handler`.
 
+In order to configure a span flush timeout (default is set to 1 second), please configure `OTEL_LAMBDA_FLUSH_TIMEOUT` env property. The value is in seconds.
```
trask, line 9: probably should follow standard config property naming? see #1414

kubawach: That's a very good idea, I'm all in favour of standardisation. This is of course a bit different than the "regular" usage of the config properties as here we're not configuring an instrumentation but a runtime wrapper (only via env property) that uses the aws lambda instrumentation

If such usage fits into the model that you've proposed (for me it sounds OK), I believe that the updated property name should be: `OTEL_INSTRUMENTATION_AWS-LAMBDA_FLUSH_TIMEOUT` There is also another config property, defining the handler, which would then become: `OTEL_INSTRUMENTATION_AWS-LAMBDA_HANDLER`.

@trask please have a look ^^

trask: that sounds good, with underscore instead of dash, e.g.

`OTEL_INSTRUMENTATION_AWS_LAMBDA_FLUSH_TIMEOUT`
and
`OTEL_INSTRUMENTATION_AWS_LAMBDA_HANDLER`

kubawach: Ah yes, kebab-case is for dot-property names only. Silly me :) Updated PR, please have a look.

mateuszrzeszutek (approved):

trask (approved):

PR #1961, opened 2020-12-23
Update vendor distro example to 0.13

iNikem:
@anuraaga @jkwatson some feedback on the current released state of configuration:

- It is confusing that`addSpanProcessor` is on `SdkTracerProvider`, not on builder.
- There is `setTraceConfig` on `SdkTracerProviderBuilder` and `updateActiveTraceConfig` on `SdkTracerProvider`. It is unclear if different names mean different functionality.
- Still awkward to set Sampler
- `sdkTracerProvider.addSpanProcessor` returns `void`, breaking fluent interface
- Interplay between `OpenTelemetrySdkBuilder.build()` and `GlobalOpenTelemetry` is confusing

### DIFF
--- a/examples/distro/build.gradle
+++ b/examples/distro/build.gradle
@@ -8,8 +8,8 @@ subprojects {
 
     ext {
         versions = [
-                opentelemetry            : "0.12.0",
-                opentelemetryJavaagent   : "0.12.0",
+                opentelemetry            : "0.13.1",
+                opentelemetryJavaagent   : "0.13.0",
                 bytebuddy                : "1.10.10",
                 guava                    : "20.0"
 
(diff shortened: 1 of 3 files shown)


anuraaga:
Thanks @iNikem!

Mutators are going away

https://github.com/open-telemetry/opentelemetry-java/pull/2385
https://github.com/open-telemetry/opentelemetry-java/pull/2402

Build / global interaction problem might be improved by https://github.com/open-telemetry/opentelemetry-java/pull/2387

Sampler being weird to configure - if it's about potentially separating Sampler from trace limits like I hinted at, I like the idea of splitting. But, even these proposals I make are more conservative than I'd like out of my fear of bogdan :)

trask (approved):

PR #1970, opened 2020-12-30
Close JarFile

laurit:


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExporterClassLoader.java
```diff
@@ -127,10 +128,9 @@ private static String getPackageName(String className) {
   }
 
   private static Manifest getManifest(URL url) {
-    try {
-      JarFile jarFile = new JarFile(url.getFile());
+    try (JarFile jarFile = new JarFile(url.toURI().getPath())) {
```
anuraaga, line 131: Do we need to change the `url.getFile()`?

anuraaga (commented):
Thanks, closing looks good but wondering if we need the other change

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExporterClassLoader.java
laurit: If path contains a space then `url.getFile()` will leave it as `%20`, `url.toURI().getPath()` decodes it as an actual space character

anuraaga: Does it cause a problem? I'm surprised the more direct way of getting a `File` would cause `JarFile` constructor to fail, if you're seeing an exception let's add a comment above this line on what exception we are trying to prevent with the indirect code.

laurit: `URL.getFile` returns a `String`. As the url is constructed with `new File(exporterJar).toURI().toURL();` which encodes space -> %20 when getting file from url we should use a way that translates %20 back to space.

anuraaga: So it sounds like there is a problem when the path has a space. If so yeah we need to fix it :) But it's confusing, so we need at least a comment documenting the failure case. Even better is a test that fails before the change and passes after, then no one will accidentally revert and regress.

Alternatively, we can keep the code similarly as before and file an issue - the PR title is about closing the jar file but is doing more than just that. It's ok but we need to do it completely, otherwise let's not make too many changes.

laurit: Added test

anuraaga (approved):
Thanks!

trask (approved):
nice 👍

PR #1979, opened 2021-01-04
Distinguish between the various servlet tracers for easier debugging.

jkwatson:
While trying to track down an issue with a customer, I ran into the fact that I couldn't figure out which servlet instrumentation was generating the spans. This will help with that.

### DIFF
--- a/instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
+++ b/instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
@@ -16,7 +16,7 @@ public class Servlet2HttpServerTracer extends ServletHttpServerTracer<ResponseWi
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
+    return "io.opentelemetry.javaagent.servlet22";
   }
 
   @Override
--- a/instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
+++ b/instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
@@ -16,7 +16,7 @@ public class Servlet2HttpServerTracer extends ServletHttpServerTracer<ResponseWi
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet22";
+    return "io.opentelemetry.javaagent.servlet-2.2";
   }
 
   @Override
--- a/instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
+++ b/instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
@@ -22,7 +22,7 @@ public class Servlet3HttpServerTracer extends ServletHttpServerTracer<HttpServle
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet30";
+    return "io.opentelemetry.javaagent.servlet-3.0";
   }
 
   @Override
--- a/instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
+++ b/instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
@@ -22,7 +22,7 @@ public class Servlet3HttpServerTracer extends ServletHttpServerTracer<HttpServle
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
+    return "io.opentelemetry.javaagent.servlet30";
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


trask:
hey @jkwatson, I was assuming that the instrumentation library name is supposed to be the same for all instrumentation produced by a given instrumentation library, which would rule out the `.request`, `.response` suffixes.

for adding the version number to the library name, we used to have that, but discussed and removed it in #1260.

definitely open to revisiting that discussion tho.

is this a case where you were debugging without access to the javaagent debug log?

jkwatson:
> hey @jkwatson, I was assuming that the instrumentation library name is supposed to be the same for all instrumentation produced by a given instrumentation library, which would rule out the .request, .response suffixes.

I'd be fine with that, although I don't think it matters too much, one way or the other.

> is this a case where you were debugging without access to the javaagent debug log?

Yes, all we had were a couple of sample spans in the final observability system. Being able to detect what servlet version was being instrumented would have been very useful.

trask:
> Yes, all we had were a couple of sample spans in the final observability system. Being able to detect what servlet version was being instrumented would have been very useful.

can you create an issue to revisit the decision made in #1260? I'd like to give others a chance to weigh in, and then apply the change consistently across all instrumentation

instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
```diff
@@ -16,7 +16,7 @@ public static Servlet2HttpServerTracer tracer() {
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
```
anuraaga, line 19: These actually seem like a no-brainer to me - our artifact names are also e.g., `opentelemetry-javaagent-servlet-2.2`, the name of the library. And these are supposed to reflect the name of the library.

I understand the potential for confusion due to the lack of something like a `+` but I think it's just the nature of how instrumentation libraries are named.

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/dispatcher/RequestDispatcherTracer.java
```diff
@@ -16,6 +16,6 @@ public static RequestDispatcherTracer tracer() {
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
+    return "io.opentelemetry.javaagent.servlet.request";
```
anuraaga, line 19: I don't know if I like it but I guess for consistency and to reflect "name of instrumentation library", these should use `.common`?

jkwatson: yeah, I wasn't really sure what these should say. `.common` seems to accurately and uniquely identify the instrumentation, though, and that's definitely the most important thing.

jkwatson:
> > Yes, all we had were a couple of sample spans in the final observability system. Being able to detect what servlet version was being instrumented would have been very useful.
> 
> can you create an issue to revisit the decision made in #1260? I'd like to give others a chance to weigh in, and then apply the change consistently across all instrumentation

#1992

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/dispatcher/RequestDispatcherTracer.java
jkwatson: updated

instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
```diff
@@ -16,7 +16,7 @@ public static Servlet2HttpServerTracer tracer() {
...
+    return "io.opentelemetry.javaagent.servlet22";
```
anuraaga, line 19:
```suggestion
    return "io.opentelemetry.javaagent.servlet-2.2";
```

Nit but we may as well use the same formatting as the artifact name for these

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
```diff
@@ -22,7 +22,7 @@ public static Servlet3HttpServerTracer tracer() {
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
+    return "io.opentelemetry.javaagent.servlet30";
```
anuraaga, line 25:
```suggestion
    return "io.opentelemetry.javaagent.servlet-3.0";
```

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/dispatcher/RequestDispatcherTracer.java
```diff
@@ -16,6 +16,6 @@ public static RequestDispatcherTracer tracer() {
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
+    return "io.opentelemetry.javaagent.servlet.common";
```
anuraaga, line 19:
```suggestion
    return "io.opentelemetry.javaagent.servlet-common";
```

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/http/HttpServletResponseTracer.java
```diff
@@ -16,6 +16,6 @@ public static HttpServletResponseTracer tracer() {
 
   @Override
   protected String getInstrumentationName() {
-    return "io.opentelemetry.javaagent.servlet";
+    return "io.opentelemetry.javaagent.servlet.common";
```
anuraaga, line 19:
```suggestion
    return "io.opentelemetry.javaagent.servlet-common";
```

instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2HttpServerTracer.java
jkwatson: seems reasonable to me!

anuraaga (approved):

mateuszrzeszutek (approved):

iNikem:
Anybody wants to chime in on this before we merge? I don't :)

PR #1985, opened 2021-01-05
Improve servlet22 response status handling

laurit:
If servlet doesn't set response status then currently there is a NPE when casting Integer -> int in Servlet2Advice#stopSpan. This happens when running "test exception" in JettyServlet2Test with other tests commented out. In JettyServlet2Test "test exception" currently passes only when it is run after "test error" as our integration ends up using response status set from previous request, reset response status when creating span to avoid this.


instrumentation/servlet/servlet-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v2_2/Servlet2Advice.java
```diff
@@ -43,6 +43,8 @@ public static void onEnter(
 
     context = tracer().startSpan(httpServletRequest);
     scope = context.makeCurrent();
+    // reset response status from previous request
```
trask, line 46:
```suggestion
    // reset response status from previous request
    // (some servlet containers reuse response objects to reduce memory allocations)
```

trask (approved):
nice find 👍

anuraaga (approved):
Thanks!

PR #1995, opened 2021-01-07
Make factory method for netty 4.0 attribute keys public

pavolloffay:
Our custom instrumentation needs to create additional netty AttributeKey's. It would be useful to expose this.

### DIFF
--- a/instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/AttributeKeys.java
+++ b/instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/AttributeKeys.java
@@ -42,6 +42,8 @@ public class AttributeKeys {
    * prevents an issue with Apache Atlas project were this class loaded by multiple class loaders,
    * while the Attribute class is loaded by a third class loader and used internally for the
    * cassandra driver.
+   *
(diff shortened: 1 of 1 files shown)


instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/AttributeKeys.java
```diff
@@ -43,7 +43,7 @@
    * while the Attribute class is loaded by a third class loader and used internally for the
    * cassandra driver.
    */
-  private static <T> AttributeKey<T> attributeKey(String key) {
+  public static <T> AttributeKey<T> attributeKey(String key) {
```
trask, line 46: can you add a comment about why it's public? I still have same concern as https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1951/files#r546868569

pavolloffay: +1 I have added a comment there

trask (approved):

pavolloffay:
Could this be merged? cc) @anuraaga @mateuszrzeszutek

mateuszrzeszutek (approved):

PR #1999, opened 2021-01-08
Update snapshot url in docs

breedx-splk:
The old jfrog url was still in `RELEASING.md` and I changed the links to go to the org (so that when users click it it's meaningful).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@ See [Running the tests](./docs/contributing/running-tests.md) for more details.
 
 For developers testing code changes before a release is complete, there are
 snapshot builds of the `master` branch. They are available from
-[Sonatype OSS snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/)
+[Sonatype OSS snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/)
 
 #### Building from source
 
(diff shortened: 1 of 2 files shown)


trask (approved):
👍

CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@ See [Running the tests](./docs/contributing/running-tests.md) for more details.
 
 For developers testing code changes before a release is complete, there are
 snapshot builds of the `master` branch. They are available from
-[Sonatype OSS snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/)
+[Sonatype OSS snapshots repository](https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/)
```
anuraaga, line 19: I don't have a strong preference but maybe we can include two URLs, one with repo one that goes to the group ID folder. The former is the only one that works when configuring repositories in e.g. Gradle, Maven.

anuraaga (approved):

CONTRIBUTING.md
trask: this is a good point, maybe something like

```suggestion
the Sonatype OSS snapshots repository at https://oss.sonatype.org/content/repositories/snapshots/ ([browse](https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/))
```

breedx-splk: This did occur to me when writing this originally, and I assumed (maybe poorly) that any developers who are savvy enough to use snapshot builds in their build are savvy enough to take off the group id, and that potentially having two urls could be confusing because then the user doesn't know which one to use.  This is fine tho, I like it. Let's do this.

anuraaga:
Thanks!

PR #2004, opened 2021-01-08
Don't override preconfigured propagators.

### LINKED ISSUE #1681: TracerInstaller overrides custom propagators
If some `TracerCustomizer` tries to provide its own propagators via calling `OpenTelemetry.setGlobalPropagators`, then `io.opentelemetry.javaagent.tooling.PropagatorsInitializer#initializePropagators` will override it anyway. At least with default `HttpTraceContext`. This makes it impossible to provide external/custom propagator.

Related to open-telemetry/opentelemetry-java#2022

Setting P1 because it limits the power of vendor distribution, which is P1.

breedx-splk:
This resolves #1681.

In that bug, it's described how the `PropagatorsInitializer` will overwrite any customized propagators configured earlier in the `GlobalOpenTelemetry.set(sdkWithCustomPropagators)`.

Now, in `initializePropagators()`, we pull the preconfigured propgatation from the global via `GlobalOpenTelemetry.get().getPropagators().getTextMapPropagator()` and pass this down.  If propagators were preconfigured (not noops), they will be composited along with the other kinds.

There weren't any tests around the existing behavior, so this backfills tests.  The bug was then fixed and the code cleaned up a little (but not too much, because sdk changes will likely result in additional changes in the near future).

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
@@ -5,11 +5,11 @@
 
 package io.opentelemetry.javaagent.tooling;
 
+import static java.util.stream.Collectors.toList;
+
 import com.google.common.collect.ImmutableMap;
-import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.baggage.propagation.W3CBaggagePropagator;
-import io.opentelemetry.api.trace.Span;
-import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.ContextPropagators;
@@ -18,13 +18,13 @@ import io.opentelemetry.extension.trace.propagation.AwsXRayPropagator;
 import io.opentelemetry.extension.trace.propagation.B3Propagator;
 import io.opentelemetry.extension.trace.propagation.JaegerPropagator;
 import io.opentelemetry.extension.trace.propagation.OtTracerPropagator;
-import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -58,19 +58,30 @@ public class PropagatorsInitializer {
    * </ul>
    */
   public static void initializePropagators(List<String> propagatorIds) {
-    /* Only override the default propagators *if* the user specified any. */
+    initializePropagators(propagatorIds, () -> GlobalOpenTelemetry.get().getPropagators().getTextMapPropagator(),
+        p -> GlobalOpenTelemetry.get().setPropagators(p));
+  }
+
+  //exists for testing
+  static void initializePropagators(List<String> propagatorIds, Supplier<TextMapPropagator> preconfiguredPropagator,
+      Consumer<ContextPropagators> globalSetter) {
+    ContextPropagators propagators = createPropagators(propagatorIds, preconfiguredPropagator.get());
+    // Register it in the global propagators:
+    globalSetter.accept(propagators);
+  }
+
+  private static ContextPropagators createPropagators(List<String> propagatorIds, TextMapPropagator preconfiguredPropagator) {
+    /* Only override the default propagators *if* the caller specified any. */
     if (propagatorIds.size() == 0) {
       // TODO this is probably temporary until default propagators are supplied by SDK
       //  https://github.com/open-telemetry/opentelemetry-java/issues/1742
-      OpenTelemetry.setGlobalPropagators(
-          ContextPropagators.create(
-              TextMapPropagator.composite(
-                  W3CTraceContextPropagator.getInstance(), W3CBaggagePropagator.getInstance())));
-      return;
+      return createPropagatorsRemovingNoops(Arrays.asList(preconfiguredPropagator,
+          W3CTraceContextPropagator.getInstance(), W3CBaggagePropagator.getInstance()));
     }
 
     List<Propagator> propagators = new ArrayList<>(propagatorIds.size());
     List<TextMapPropagator> textMapPropagators = new ArrayList<>();
+    textMapPropagators.add(preconfiguredPropagator);
 
     for (String propagatorId : propagatorIds) {
       Propagator propagator = TEXTMAP_PROPAGATORS.get(propagatorId);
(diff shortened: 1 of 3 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -82,73 +102,20 @@ public static void initializePropagators(List<String> propagatorIds) {
       }
     }
     if (propagators.size() > 1) {
-      textMapPropagators.add(new MultiPropagator(propagators));
+      textMapPropagators.addAll(propagators);
     } else if (propagators.size() == 1) {
       textMapPropagators.add(propagators.get(0));
     }
```
mateuszrzeszutek, line 108: Can't it be replaced by just `textMapPropagators.addAll(propagators);`? Or you can add directly to the `textMapPropagators` list, without using the `propagators` one

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
```diff
@@ -82,73 +102,20 @@ public static void initializePropagators(List<String> propagatorIds) {
...
-    private static boolean isSpanContextExtracted(Context context) {
-      return Span.fromContextOrNull(context) != null;
-    }
+  private static ContextPropagators createPropagatorsRemovingNoops(
+      List<TextMapPropagator> textMapPropagators) {
+    return ContextPropagators.create(
+        TextMapPropagator.composite(
+            textMapPropagators.stream()
+                .filter(propagator -> propagator != TextMapPropagator.noop())
+                .collect(toSet())));
```
mateuszrzeszutek, line 118: Just a thought: does the order of propagators matter here?

javaagent-tooling/src/test/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializerTest.java
```diff
@@ -0,0 +1,156 @@
...
+                        PropagatorsInitializer.Propagator.B3,
+                        PropagatorsInitializer.Propagator.JAEGER));
+  }
+
+  @Test
+  void initialize_idsAndPreconfigured() {
+    List<String> ids = Arrays.asList("jaeger", "xray");
+    AtomicReference<ContextPropagators> seen = new AtomicReference<>();
+    Consumer<ContextPropagators> setter = seen::set;
+    TextMapPropagator mockPreconfigured = mock(TextMapPropagator.class);
```
mateuszrzeszutek, line 139: Just a minor suggestion: you can pull `seen` and `mockPreconfigured` to the class level, since they're used in almost all test cases

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
breedx-splk: It shouldn't...I mean, I don't think so.

There's also [a recent spec change](https://github.com/open-telemetry/opentelemetry-specification/pull/1316/files#diff-adf3b6d20af39e3109b1907f4b1f356e90912b893693a308955dea13dd97612f) that does require this deduplication.  I would expect that if order mattered that it would also have to call out how to deduplicate (first wins vs. last wins, etc).

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
breedx-splk: Great call!  Originally this size check was there as an optimization to prevent unnecessary compositing.  That logic also exists in `TextMapPropagator.composite()`, so it doesn't need to be here.  Since the tests still pass, I'll make this change...thanks!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/PropagatorsInitializer.java
trask: It's not totally clear from the spec, but I think order may be relevant, e.g.

> The resulting composite Propagator will invoke the Propagators, Injectors, or Extractors, in the order they were specified.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#composite-propagator

breedx-splk: It's an interesting idea for sure.  There seems to be a lack of clarity on _how_ to deduplicate while maintaining order (first wins, last wins, etc).  The spec thinks it's important, but I'm not sure _why_.  Shouldn't matter, right?

This is further complicated by the SPI.  So if the agent config has `[b3, jaeger, b3]` and the spi has `[b3]`, what do we think the final deduplicated order should be?

trask: I think ordering is important on extraction, in case multiple propagators set the parent (propagated) span into the context, e.g. if a request has both jaeger and w3c headers.

It could potentially matter on injection) in case multiple propagators set the same http headers, though I think most propagators have their own set of http headers, so this seems less important of a concern.

> This is further complicated by the SPI. So if the agent config has [b3, jaeger, b3] and the spi has [b3], what do we think the final deduplicated order should be?

Which SPI are you thinking of?

breedx-splk: Seems edge casey, but I agree with the two possibilities you've outlined.

The SPI I was thinking about (and the entire motivation in submitting this PR) is along the TracerCustomizer, wherein it can set up its own propagators (like in the `DemoTracerCustomizer` example).  

If that happens, we will need to define if those come before or after the other agent propagators.  I would think that a custom one should come first....but I dunno.

trask: In `DemoTracerCustomizer`, can you programmatically order it as you wish? e.g. to put your custom propagator first, something like

```
GlobalOpenTelemetry.setPropagators(
    ContextPropagators.create(
        TextMapPropagator.composite(
            new DemoPropagator(), GlobalOpenTelemetry.getPropagators().getTextMapPropagator()))
```

trask: oh, I just saw https://github.com/open-telemetry/opentelemetry-java/pull/2475, @anuraaga is planning to bring over the new SDK auto-configuration to this repo after Java SDK 0.14.0, so probably we will inherit this SPI (and it's associated ordering)

trask: although, from vendor perspective, I'd think they would want to provide a "named" propagator, and have that configurable as part of `otel.propagators`, e.g. I would want to add an `applicationinsights` named propagator, and I'd want to override the default `otel.propagators` to be `tracecontext,baggage,applicationinsights`

breedx-splk: I haven't given it much thought, but I like the idea of symbolic names to help make things easier.  Seems outside the scope of this PR and #1681 though, doesn't it? 

Do we think there is a path forward with the approach in this PR (combine the preconfigured list ahead of the agent's configured list and dedupe), or do you think there's too many unknowns and other related changes for this to be viable?

anuraaga: For the propagator provider, while haven't done so yet I'm intending to make `ConfigProperties` public so that can be used to recognize the `otel.propagators` property without having to force interaction between the SPI and the property It makes me realize the provider has to support not returning a propagator :)

For ordering, I copied over this comment we've had since the beginning

`Invocation order of {@code TextMapPropagator#inject()} and {@code TextMapPropagator#extract()} for registered trace propagators is undefined.`
https://github.com/open-telemetry/opentelemetry-java/blob/master/api/context/src/main/java/io/opentelemetry/context/propagation/TextMapPropagator.java#L50

Spec doesn't talk about order

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/context/api-propagators.md#composite-propagator

So I don't think the issue of order is something we need to worry about on this repo as the SDK doesn't guarantee any order anyways :) If that seems like a bad idea, we would probably want to confirm with a spec issue.

trask: @anuraaga the spec section you linked to does talk about ordering:

> The resulting composite Propagator will invoke the Propagators, Injectors, or Extractors, in the order they were specified

anuraaga: Oops missed it :) Going to have to update the SDK javadoc. Do you think not enabling a provided propagator by default and requiring a `name` is a better approach to handle the ordering?

trask: > Do you think not enabling a provided propagator by default and requiring a `name` is a better approach to handle the ordering?

For the javaagent this seems the least confusing. Custom propagators then get treated the same as built-in propagators from configuration perspective.

Maybe the auto-configuration module needs a PropertySource SPI also, so that auto-configuration users can change the default `otel.propagators` value to include their custom propagator(s), same as javaagent distributions?

anuraaga: @trask I'm leaning towards having two, one that is automatically enabled (most common in apps I think) and one that registers an available propagator by name. Just wondering, is it common in these custom distros to want to allow a user to disable the custom propagator using `otel.propagators`?

trask: Application Insights moved to W3C a while ago, but we still provide legacy propagation for backwards compatibility. Our default is to propagate both, but it's nice for users who don't need the backwards compatibility to be able to disable the legacy format (reduce wire size a little).

I could see this being a similar predicament for other vendors/users as they migrate away from legacy propagation formats.

trask: @breedx-splk thanks for instigating this discussion, it was very timely and led to good last minute changes in https://github.com/open-telemetry/opentelemetry-java/pull/2475 which I think will address #1681 once @anuraaga brings in the new autoconfigure module

breedx-splk: Sure thing!  Glad there was something good that fell out of it.  Whatever we end up with, hopefully it will have test coverage.  :)

Going to close since it sounds like there's a lot of momentum on another superseding path.

breedx-splk:
Sounds like we are going to try this out for this release and then supersede it with config code changes shortly after.  Reopening for probable merging.

anuraaga (approved):

trask (approved):
I'm good punting on the ordering issue since we are planning to replace it with SDK autoconfigure module in the future 👍

Issue #344, opened 2020-04-23
Allow to disable auto-instrumentation for specified methods


iNikem:
There is a configuration property `ota.trace.classes.exclude`, which allows to exclude some classes from auto-instrumentation. I propose to add similar property `ota.trace.methods.exclude` to allow more fine-grained configuration: to exclude only specific methods from a class.

trask:
I'm curious if these properties are only needed for working around instrumentation bugs, or if there are other use cases?

iNikem:
Another use-case I can think of: some library out of user's control produces a huge amount of monitoring data either due to a bug or just high load. User may want to silence it.

trask:
Have you seen the instrumentation-level silencing configuration yet:

https://github.com/open-telemetry/opentelemetry-auto-instr-java/blob/v0.2.2/agent-bootstrap/src/main/java/io/opentelemetry/auto/config/Config.java#L288

We discussed this a bit in last week's SIG meeting ([sorry for the somewhat jumbled notes](https://docs.google.com/document/d/1WK9h4p55p8ZjPkxO75-ApI9m0wfea6ENZmMoFRvXSCw)).

Does this address what you are looking for?

johnbley:
We (Splunk) have some customers who have over-instrumented their app and find a configuration-based approach more flexible than re-changing code, particularly as an immediate response.  Given that the (otel) code to enable this is fairly localized I find the feature pretty innocuous.

iNikem:
@trask If you don't object this proposed change, can you please assign this task to me?

trask:
@johnbley oh, do you mean over-instrumented using `@Trace` annotations? just want to understand the use case for needing method-level excludes, then good to go

johnbley:
Yes, over-instrumenting with `@Trace` is what I meant.  Sorry for the confusing terminology.

trask:
@johnbley got it, thanks! @iNikem all yours!

PR #2007, opened 2021-01-09
Clarify README.md instructions for manual instrumentation

mpetazzoni:
See discussion #1945 for the background behind this change. The aim
is to make this repository's README easier to grok when trying to get
started instrumenting a Java application with OpenTelemetry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,10 @@ java -javaagent:path/to/opentelemetry-javaagent-all.jar \
      -jar myapp.jar
 ```
 
-### Configuration parameters (subject to change!)
+Learn how to add custom instrumentation in the [Manually Instrumenting](#manually-instrumenting)
+section.
+
+## Configuration parameters (subject to change!)
 
 Note: These parameter names are very likely to change over time, so please check
 back here when trying out a new version! Please report any bugs or unexpected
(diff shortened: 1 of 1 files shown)


mpetazzoni:
I've been added to the CLA list by @flands. Hoping this comment will tickle the EasyCLA bot to re-evaluate?

trask:
closing and re-opening to re-trigger CLA check(?)

anuraaga:
@mpetazzoni I think you still need to click through to have your GitHub account associated, though not sure.

mpetazzoni:
@anuraaga Thanks, it worked this time!

README.md
```diff
@@ -79,7 +79,97 @@ java -javaagent:path/to/opentelemetry-javaagent-all.jar \
...
+    </dependency>
+  </dependencies>
+```
+
+#### Gradle
+
+```groovy
+dependencies {
+    compile('io.opentelemetry:opentelemetry-api:0.11.0')
+    compile('io.opentelemetry:opentelemetry-extension-annotations:0.11.0')
```
anuraaga, line 114:
```suggestion
    implementation('io.opentelemetry:opentelemetry-api:0.11.0')
    implementation('io.opentelemetry:opentelemetry-extension-annotations:0.11.0')
```

README.md
```diff
@@ -79,7 +79,97 @@ java -javaagent:path/to/opentelemetry-javaagent-all.jar \
...
+}
+```
+
+#### Adding attributes to the current span
+
+A common need when instrumenting an application is to capture additional application-specific or
+business-specific information as additional attributes to an existing span from the automatic
+instrumentation. Grab the current span with `Span.current()` and use of the `setAttribute()`
```
anuraaga, line 122:
```suggestion
instrumentation. Grab the current span with `Span.current()` and use the `setAttribute()`
```

README.md
```diff
@@ -79,7 +79,97 @@ java -javaagent:path/to/opentelemetry-javaagent-all.jar \
...
+```
+
+Each time the application invokes the annotated method, it creates a span that denote its duration
+and provides any thrown exceptions. Unless specified as an argument to the annotation, the span name
+will be `<className>.<methodName>`.
+
+##### Suppressing `@WithSpan` instrumentation
+
+Suppressing `@WithSpan` is useful if you have code that is over-instrumented using `@WithSpan`
+and you want to suppress some of them without modifying the code.
```
anuraaga, line 157: I know it's copied, but while we're updating the docs, for this, I'm not quite sure how this happens. `@WithSpan` is quite an explicit addition to code, how could it lead to over-instrumentation? @trask do you know the background of this quote? Would be nice if there's a real-world example here

trask: I do know the background! see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/344#issuecomment-620818687

I don't think this section needs to be in top-level README, though, probably should be moved to https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/suppressing-instrumentation.md

README.md
```diff
@@ -79,7 +79,97 @@ java -javaagent:path/to/opentelemetry-javaagent-all.jar \
      -jar myapp.jar
 ```
 
-### Configuration parameters (subject to change!)
+### Manually instrumenting
```
trask, line 82: I'm not sure that the "Manually instrumenting" section should go above the "Configuration parameters" section?

Are you feeling like the README is getting too big, and stuff lower down is getting lost? Maybe we can split more things out into sub-pages and keep the main README lighter?

README.md
mpetazzoni: I would imagine this happens if you're pulling in a dependency that uses `@WithSpan` and you need to suppress _their_ usage.

README.md
mpetazzoni: Yes, that was the main issue I was trying to address here. If possible, I'd like to keep "restructuring" to another PR.

README.md
mpetazzoni: I'm not sure I understand this change. The use of `compile()` was from the original README.

anuraaga: `compile` is deprecated and I think Gradle throws deprecation warnings now if using it. Our README was old, and if we keep it old now it'll stay old for a long time :P

README.md
anuraaga: I think `OTEL_JAVAAGENT_EXCLUDE_CLASSES` is supposed to work for that case pretty well. So anyways, maybe I'll propose removing the flag but we don't have to do anything here :)

README.md
trask: > If possible, I'd like to keep "restructuring" to another PR.

No problem. I'm still not sure about "Manual instrumentation" section going to the top though. My thinking is that "Manual instrumentation" section is generally something users need after they first get up and running with auto-instrumentation.

mpetazzoni: I've moved it back below and added a small sentence in the Getting Started section. Let me know if that works for you.

anuraaga (approved):

mpetazzoni:
Thanks @anuraaga! Hopefully someone can merge before new conflicts appear.

anuraaga:
@mpetazzoni We generally merge after two approvals, should be merged by tomorrow at the latest.

jkwatson (approved):
👍

trask (approved):

Issue #1751, opened 2020-11-24
Lazy load servers in tests


anuraaga:
This may just be a bad bandaid, but just an idea. We often use test class callbacks for creating testing servers, e.g. `@Shared`, `@TestContainer`. Starting up servers in our huge build can fail for various reasons - unfortunately `gradle-retry-plugin` will not support retrying tests for this case. An extreme cludge, but we could workaround this by instead of using class callbacks, use lazy singletons that are actually executed in the test method itself - that should allow `gradle-retry-plugin` to work. If `gradle-retry-plugin` could work for all our tests, we could avoid `invision/retry` which restarts the entire Gradle build, with much more overhead than the individual retry of `gradle-retry-plugin`.

iNikem:
As additional benefit this will make the time to load server more visible, I think

trask:
I haven't seen failures due to this in a while, do you know if anything changed?

PR #2018, opened 2021-01-11
Instrument jsf action calls

laurit:


instrumentation/jsf/jsf-testing-common/src/main/resources/test-app/WEB-INF/web.xml
```diff
@@ -0,0 +1,20 @@
...
+  </filter>
+
+  <filter-mapping>
+    <filter-name>ExceptionFilter</filter-name>
+    <url-pattern>/*</url-pattern>
+  </filter-mapping>
+
+  <!--listener>
+    <listener-class>org.apache.myfaces.webapp.StartupServletContextListener</listener-class>
+  </listener-->
```
vovencij, line 19: remove?

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
```diff
@@ -0,0 +1,79 @@
...
+import javax.faces.component.ActionSource2;
+import javax.faces.component.UIViewRoot;
+import javax.faces.context.FacesContext;
+import javax.faces.el.EvaluationException;
+import javax.faces.el.MethodNotFoundException;
+import javax.faces.event.ActionEvent;
+
+public abstract class JsfTracer extends BaseTracer {
+
+  private static String getSpanName(String expressionString) {
```
vovencij, line 22: Could you add either a comment explaining the format of the expressionString, or add a unit-test with concrete examples. Or both.

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
```diff
@@ -0,0 +1,79 @@
...
+
+    if (result.contains("(") && result.contains(")")) {
+      result = result.replaceAll("\\(.*?\\)", "()");
+    }
+
+    return result;
+  }
+
+  public Span startSpan(ActionEvent event) {
+    if (event.getComponent() instanceof ActionSource2) {
```
vovencij, line 40: Would be nice to have some comments about this particular constraint and how large portion of other components are we missing and why?

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
```diff
@@ -0,0 +1,79 @@
...
+    if (serverSpan == null) {
+      return;
+    }
+
+    UIViewRoot uiViewRoot = facesContext.getViewRoot();
+    if (uiViewRoot == null) {
+      return;
+    }
+
+    String viewId = uiViewRoot.getViewId();
```
vovencij, line 63: Add a comment with the description of viewId and its potential cardinality.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/AbstractHttpServerTest.groovy
```diff
@@ -0,0 +1,68 @@
...
+import io.opentelemetry.instrumentation.test.utils.PortUtils
+import okhttp3.OkHttpClient
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import spock.lang.Shared
+
+abstract class AbstractHttpServerTest<SERVER> extends AgentTestRunner {
+  public static final Logger SERVER_LOGGER = LoggerFactory.getLogger("http-server")
+  static {
+    ((Logger) SERVER_LOGGER).setLevel(Level.DEBUG)
```
mateuszrzeszutek, line 20: It used to be `ch.qos.logback.classic.Logger` previously

Also, why not set it in the logback XML config instead?

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/AbstractHttpServerTest.groovy
```diff
@@ -0,0 +1,68 @@
...
+import ch.qos.logback.classic.Level
+import io.opentelemetry.instrumentation.test.AgentTestRunner
+import io.opentelemetry.instrumentation.test.utils.OkHttpUtils
+import io.opentelemetry.instrumentation.test.utils.PortUtils
+import okhttp3.OkHttpClient
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import spock.lang.Shared
+
+abstract class AbstractHttpServerTest<SERVER> extends AgentTestRunner {
```
mateuszrzeszutek, line 17: Instead of an abstract class, consider making it a trait (like `ApplicationConfigTrait` or `InstrumentationTestTrait`. `withRetryOnAddressAlreadyInUse` could possibly be moved to `PortUtils` to remove dependency on `AgentTestRunner` (`AgentTestRunner` does not seem to be a proper place for this method anyway)

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  @Override
+  String getContextPath() {
+    return "/jetty-context"
+  }
+
+  void addServlet(ServletContextHandler servletContext, String path, Class<Servlet> servlet) {
+    servletContext.addServlet(servlet, path)
+  }
```
mateuszrzeszutek, line 69: Is this method used?

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
```diff
@@ -0,0 +1,227 @@
...
+    def viewState = doc.selectFirst("[name=javax.faces.ViewState]")?.val()
+    def formAction = doc.selectFirst("#app-form").attr("action")
+    def jsessionid = formAction.substring(formAction.indexOf("jsessionid=") + "jsessionid=".length())
+
+    then:
+    viewState != null
+    jsessionid != null
+
+    when:
+    TEST_WRITER.clear()
```
mateuszrzeszutek, line 119: Nit: I think that you can move this line just after the `assertTraces` call, this way the whole trace logic for this test step would be wrapped in one block.

instrumentation/jsf/jsf-testing-common/src/main/resources/test-app/WEB-INF/web.xml
laurit: removed, thanks

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
laurit: I removed this method and used expression as span name as it is

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
laurit: Comment added. My understanding is that there aren't that many components that produce actions, those that do extend https://docs.oracle.com/javaee/7/api/javax/faces/component/UICommand.html which implements ActionSource2.

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
laurit: comment added. View id is a context relative path to the file that produces the view, which should be ok.

instrumentation/jsf/jsf-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsf/JsfTracer.java
```diff
@@ -0,0 +1,79 @@
...
+    if (result.startsWith("#{")) {
+      result = result.substring(2);
+    }
+
+    if (result.endsWith("}")) {
+      result = result.substring(0, result.length() - 1);
+    }
+
+    if (result.contains("(") && result.contains(")")) {
+      result = result.replaceAll("\\(.*?\\)", "()");
```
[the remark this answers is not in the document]
laurit: I removed this method altogether.

instrumentation/jsf/myfaces/javaagent/myfaces-javaagent.gradle
```diff
@@ -0,0 +1,26 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    group = "org.apache.myfaces.core"
+    module = "myfaces-impl"
+    versions = "[2.2,3)"
+    extraDependency "jakarta.el:jakarta.el-api:3.0.3"
```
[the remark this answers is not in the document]
laurit: changed muzzle range to [1.2,3) and added `assertInverse `

instrumentation/jsf/mojarra/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mojarra/ProcessActionAdvice.java
```diff
@@ -0,0 +1,45 @@
...
+package io.opentelemetry.javaagent.instrumentation.mojarra;
+
+import static io.opentelemetry.javaagent.instrumentation.mojarra.MojarraTracer.tracer;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Scope;
+import javax.faces.event.ActionEvent;
+import net.bytebuddy.asm.Advice;
+
+public class ProcessActionAdvice {
```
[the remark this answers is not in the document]
laurit: I'll change these to inner classes. The impression I got from looking at existing code was that the decision to put advice to inner class or standalone class seemed to be mostly dependent on the size of the advice class.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/AbstractHttpServerTest.groovy
laurit: Thanks, I have restored the cast to `ch.qos.logback.classic.Logger`. I'll see if I can move the configuration to xml.

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
laurit: Thanks, removed.

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
laurit: moved

testing-common/testing-common.gradle
```diff
@@ -3,6 +3,8 @@ apply from: "$rootDir/gradle/java.gradle"
 dependencies {
   api deps.guava
   api deps.spock
+  api deps.slf4j
```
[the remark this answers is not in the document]
laurit: Removed it. The reason I added these was because I have a common module that depends on testing-commons for tests that is used by 2 other modules which just extend test class from the base module. Groovy compile fails because in the class hierarchy of the test class some super class uses logback & slf4j. To me it seemed weird that just because I have an extra module I have to add additional dependencies. Perhaps adding CompileStatic to test class would also work, but I resorted to adding `api deps.testLogging` to the common module.

testing-common/testing-common.gradle
```diff
@@ -32,7 +32,7 @@ dependencies {
 
   api deps.groovy
 
-  testImplementation project(':instrumentation-api')
+  api project(':instrumentation-api')
```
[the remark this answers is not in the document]
laurit: Thanks. This is indeed the wrong place for this. Changed it back. Perhaps because I also had an issue with logging I automatically did a similar change here. The issue is again that I have a common module that is used by 2 other modules where muzzle check failed during runtime because it couldn't find `io.opentelemetry.instrumentation.api.tracer.BaseTracer` presumably because it wasn't registered as helper class during compile (idk might be something completely different). I changed common module from javaagent -> library which has `api project(':instrumentation-api')`. I hope this was the correct choice.

instrumentation/jsf/mojarra/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mojarra/MojarraInstrumentationModule.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.tooling.InstrumentationModule;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class MojarraInstrumentationModule extends InstrumentationModule {
+  public MojarraInstrumentationModule() {
+    super("mojarra");
```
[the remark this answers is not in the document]
laurit: Changed module name to `mojarra-1.2` and added it as additional instrumentation name

instrumentation/jsf/myfaces/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/myfaces/MyFacesInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
...
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.tooling.InstrumentationModule;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class MyFacesInstrumentationModule extends InstrumentationModule {
+  public MyFacesInstrumentationModule() {
+    super("myfaces");
```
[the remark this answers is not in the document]
laurit: Changed module name to `myfaces-1.2` and added it as additional instrumentation name

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/AbstractHttpServerTest.groovy
laurit: Moved code from abstract class to HttpServerTestTrait, also moved retry logic to RetryOnAddressAlreadyInUseTrait

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/AbstractHttpServerTest.groovy
laurit: The same pattern is also used in AgentTestRunner and OkHttpUtils. @mateuszrzeszutek If it is ok with you I would leave this as is to avoid putting too much stuff into this pr. Additionally I'm not sure how well it would work out, there are some tests that also include logback.xml. I suspect that having a logback.xml in testing-common could override logback.xml that is inside test.

mateuszrzeszutek: Sure, we can tackle this issue some time later

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
```diff
@@ -0,0 +1,238 @@
...
+    assertTraces(1) {
+      trace(0, 1) {
+        basicSpan(it, 0, getContextPath() + "/hello.xhtml", null)
+      }
+    }
+  }
+
+  def "test hello with faces url"() {
+    setup:
+    def url = HttpUrl.get(address.resolve("faces/hello.xhtml")).newBuilder().build()
```
mateuszrzeszutek, line 86: Nit: could be merged into a parametrized test with `test hello` -- this is basically the same scenario except for different URLs.

instrumentation/jsf/myfaces-1.2/javaagent/myfaces-1.2-javaagent.gradle
```diff
@@ -0,0 +1,27 @@
...
+    group = "org.apache.myfaces.core"
+    module = "myfaces-impl"
+    versions = "[1.2,3)"
+    extraDependency "jakarta.el:jakarta.el-api:3.0.3"
+    assertInverse = true
+  }
+}
+
+dependencies {
+  library group: 'org.apache.myfaces.core', name: 'myfaces-impl', version: '2.3.7'
```
mateuszrzeszutek, line 14: Right now we have a myfaces-1.2 module that tests version 2+ and a separate 1.2 testing module. Wouldn't it make more sense to keep 1.2 testing in the myfaces-1.2 module and have a separate one for 2+ tests?

Also, have you considered using a separate test task for 2+ tests instead of a separate module? Like `jms2Test` in the JMS instrumentation.

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
```diff
@@ -0,0 +1,238 @@
...
+  Server startServer(int port) {
+    List<String> configurationClasses = new ArrayList<>()
+    Collections.addAll(configurationClasses, WebAppContext.getDefaultConfigurationClasses())
+    configurationClasses.add(AnnotationConfiguration.getName())
+
+    WebAppContext webAppContext = new WebAppContext()
+    webAppContext.setContextPath(getContextPath())
+    webAppContext.setConfigurationClasses(configurationClasses)
+    // set up test application
+    webAppContext.setBaseResource(Resource.newSystemResource("test-app"))
```
mateuszrzeszutek, line 36: Instead of keeping separate app modules for 1.2 and 2.2, would putting them in different directories and changing the resource path work? That way the class extending `BaseJsfTest` would specify by an abstract method which version should be used

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
laurit: done

instrumentation/jsf/jsf-testing-common/src/main/groovy/BaseJsfTest.groovy
laurit: Done, merged both tests apps into common module.

instrumentation/jsf/myfaces-1.2/javaagent/myfaces-1.2-javaagent.gradle
laurit: Yeah, this turned out weird. I initially wanted to add integration with only 2.2 and 2.3 but as it turned out to also work for 1.2 I changed the version number to 1.2, but i still wanted to use latestDepTestLibrary to test latest version from 2.x series. As 1.2 & 2.2 have a bit different dependencies I suspect that putting them in one module will make getting them to run with correct dependencies tricky.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/RetryOnAddressAlreadyInUseTrait.groovy
```diff
@@ -0,0 +1,41 @@
...
+   * an available free port, because that kind of setup can fail sporadically if the available port
+   * gets re-used between when we find the available port and when we use it.
+   *
+   * @param closure the groovy closure to run with retry
+   */
+  static void withRetryOnAddressAlreadyInUse(Closure<?> closure) {
+    withRetryOnAddressAlreadyInUse(closure, 3)
+  }
+
+  static void withRetryOnAddressAlreadyInUse(Closure<?> closure, int numRetries) {
```
anuraaga, line 28: We already have three retry mechanisms available, gradle-retry-plugin, spock `@Retry`, and the GitHub actions workflow retry, would be great if we can avoid yet another retry mechanism.

Is it possible to use lazy-loaded singletons instead to take advantage of gradle-retry-plugin?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1751

laurit: This is not new code, it was extracted from AgentTestRunner as @mateuszrzeszutek suggested. Anyway ideally such retry should not be needed. To me it seems that for this retry to happen the port returned from `PortUtils.randomOpenPort` is either not usable (for example `PortUtils.randomOpenPort` uses `socket.setReuseAddress(true);` but if server does not then I guess server could fail to bind to that port) or someone else managed to pick the exact same port. It would be best to have each test worker to have its own separate port range that excludes ephemeral ports. Something like https://github.com/gradle/gradle/blob/master/subprojects/core/src/testFixtures/groovy/org/gradle/util/ports/FixedAvailablePortAllocator.groovy As this is not a new issue I would prefer for it to be worked on in a separate pr.

mateuszrzeszutek (approved):

instrumentation/jms-1.1/javaagent/src/jms2Test/groovy/SpringTemplateJms2Test.groovy
```diff
@@ -111,6 +114,8 @@ class SpringTemplateJms2Test extends AgentTestRunner {
         session -> template.getMessageConverter().toMessage("responded!", session)
       }
     }
+    // wait for thread to start, we expect the first span to be from receive
+    countDownLatch.await()
```
trask, line 118: 👍

instrumentation/jsf/jsf-common/library/src/main/java/io/opentelemetry/instrumentation/jsf/JsfTracer.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    UIViewRoot uiViewRoot = facesContext.getViewRoot();
+    if (uiViewRoot == null) {
+      return;
+    }
+
+    // JSF spec 7.6.2
+    // view id is a context relative path to the web application resource that produces the view,
+    // such as a JSP page or a Facelets page.
```
trask, line 52: 👍

instrumentation/jsf/myfaces-1.2/javaagent/src/myfaces12Test/resources/test-app-1.2-extra/META-INF/web-fragment.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<web-fragment version="3.0" xmlns="http://java.sun.com/xml/ns/javaee" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-fragment_3_0.xsd">
+
+  <!-- without this response status is 200 even when there is an exception -->
```
trask, line 5: 👍

trask (approved):

PR #2026, opened 2021-01-12
Make the agent log its version anymore when starting

### LINKED ISSUE #2023: Agent does not log its version anymore when starting
It's a minor thing but I find it really useful - it's easy to know which version is used when you're investigating an issue in a system that you do not have direct access to.

Anyway, the problem is that `OpenTelemetryInstaller#logVersionInfo()` is no longer called from anywhere after latest changes.

mateuszrzeszutek:
Fixes #2023

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -72,6 +72,7 @@ public class AgentInstaller {
   }
 
   public static void installBytebuddyAgent(Instrumentation inst) {
+    logVersionInfo();
     if (Config.get().getBooleanProperty(JAVAAGENT_ENABLED_CONFIG, true)) {
       Iterable<ComponentInstaller> componentInstallers = loadComponentProviders();
       installBytebuddyAgent(inst, componentInstallers);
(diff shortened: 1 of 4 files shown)


jkwatson:
Isn't the title of this PR backwards?

mateuszrzeszutek:
> Isn't the title of this PR backwards?

It is 🤦‍♂️ 
I forgot to remove the "not" word 😅

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -540,5 +541,11 @@ private static boolean isJavaBefore9WithJfr() {
     return Thread.currentThread().getContextClassLoader().getResource(jfrClassResourceName) != null;
   }
 
+  private static void logVersionInfo() {
+    VersionLogger.logAllVersions();
```
anuraaga, line 545: Can we add a test for this? I think we don't even need a smoke test anymore since we updated the normal ones to use the agent.

mateuszrzeszutek: Hmm, okay, but how do you test that? It's basically a log statement. Do you have any suggestion, anything in mind?

anuraaga: If it's easier actually, then smoke test is fine too. We have smoke tests checking ID injection I think so probably right there should work.

anuraaga (approved):
Thanks!

PR #2046, opened 2021-01-14
MongoDB 4 driver instrumentation

malafeev:
Based on https://github.com/DataDog/dd-trace-java/tree/master/dd-java-agent/instrumentation/mongo/driver-4


instrumentation/mongo/mongo-4/javaagent/mongo-4-javaagent.gradle
```diff
@@ -0,0 +1,26 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    group = "org.mongodb"
+    module = "mongodb-driver-core"
+    // this instrumentation is backwards compatible with early versions of the new API that shipped in 3.7
```
jyemin, line 7: I'm not sure if this complicates matters, but the upcoming 4.2.0 release will allow construction of a legacy `com.mongodb.MongoClient` with a `com.mongodb.MongoClientSettings`.  See https://github.com/mongodb/mongo-java-driver/commit/15b70cf7e5d41875bc46e57897f98d43ca60f4f6#diff-9edd83ced1b6f12502db95a96af13ed57e647efd84d2ddf5d040af4a653fb1edR211 (you have to expand the MongoClient diffs as they are large)

instrumentation/mongo/mongo-4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,82 @@
...
+          MongoClientInstrumentationModule.class.getName() + "$MongoClientAdvice");
+    }
+  }
+
+  public static class MongoClientAdvice {
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void injectTraceListener(
+        @Advice.This MongoClientSettings.Builder builder,
+        @Advice.FieldValue("commandListeners") List<CommandListener> commandListeners) {
```
anuraaga, line 62: We generally prefer to use public API instead of fields to reduce the chance that testLatestDeps will fail when implementation details change :)

instrumentation/mongo/mongo-4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,82 @@
...
+
+      if (!commandListeners.isEmpty()
+          && commandListeners
+              .get(commandListeners.size() - 1)
+              .getClass()
+              .getName()
+              .startsWith("io.opentelemetry.")) {
+        // we'll replace it, since it could be the 3.x async driver which is bundled with driver 4
```
anuraaga, line 70: So we'd want to filter out calls to the API to add listeners rather than filter here

instrumentation/mongo/mongo-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -154,6 +154,16 @@ private JsonWriterSettings createJsonWriterSettings(int maxNormalizedQueryLength
...
     } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException ignored) {
     }
+    if (settings == null) {
+      try {
+        settings = JsonWriterSettings.class.getConstructor(Boolean.TYPE).newInstance(false);
+      } catch (InstantiationException
+          | IllegalAccessException
+          | InvocationTargetException
+          | NoSuchMethodException ignored) {
+      }
```
anuraaga, line 164: What happens if this is null, is it ok?

settings.gradle
```diff
@@ -149,6 +149,7 @@ include ':instrumentation:logback:logback-1.0:testing'
 include ':instrumentation:methods:javaagent'
 include ':instrumentation:mongo:mongo-3.1:javaagent'
 include ':instrumentation:mongo:mongo-3.7:javaagent'
+include ':instrumentation:mongo:mongo-4:javaagent'
```
trask, line 152:
```suggestion
include ':instrumentation:mongo:mongo-4.0:javaagent'
```

instrumentation/mongo/mongo-4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.mongo.v4;
```
trask, line 6:
```suggestion
package io.opentelemetry.javaagent.instrumentation.mongo.v4_0;
```

instrumentation/mongo/mongo-4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,82 @@
...
+
+      if (!commandListeners.isEmpty()
+          && commandListeners
+              .get(commandListeners.size() - 1)
+              .getClass()
+              .getName()
+              .startsWith("io.opentelemetry.")) {
+        // we'll replace it, since it could be the 3.x async driver which is bundled with driver 4
+        commandListeners.remove(commandListeners.size() - 1);
+      }
```
trask, line 72: Is the code above needed? I think the code below will prevent double registration?

malafeev: removed

instrumentation/mongo/mongo-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
malafeev: no, null is not ok, we will get NPE later. But null shouldn't happen because constructor with boolean exists prior version 4.0. And since 4.0 there is a builder only.

instrumentation/mongo/mongo-4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4/MongoClientInstrumentationModule.java
malafeev: removed that code

instrumentation/mongo/mongo-4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4/MongoClientInstrumentationModule.java
malafeev: fixed

malafeev: it is consistent with all other mongo versions instrumentations, I updated matcher.

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,85 @@
...
+          MongoClientInstrumentationModule.class.getName() + "$MongoClientAdvice");
+    }
+  }
+
+  public static class MongoClientAdvice {
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void injectTraceListener(
+        @Advice.This MongoClientSettings.Builder builder,
+        @Advice.FieldValue("commandListeners") List<CommandListener> commandListeners) {
```
anuraaga, line 74: Thanks for using the public method - for suppression we would want to instrument `addCommandListener` instead to avoid accessing this private field. However, we've generally been skipping handling duplicate registration in an effort to come with a more general pattern so you don't have to worry about it either if it's tedious.

malafeev:
is anything else is expected from me? 
I ported instrumentation from DD and made it similar to mongo-3.7 instrumentation.

anuraaga:
@malafeev Do you think you can address https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2046#discussion_r558807628? We should avoid accessing the private field. I believe you changed to instrumenting `addCommandListener` but my intention was to keep instrumenting `build`, but call `addCommandListener` from inside instead of accessing the private variable. Currently we'll run our instrumentation every time a user executes `addCommandListener` which is weird.

malafeev:
@anuraaga , no, as I understand, we intercept only `build` method:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2046/files#diff-68a7a03ca69546307dd8eda75056b354fc429e9f87a5ec3602a3f9a4816d8294R64

```
    @Override
    public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
      return singletonMap(
          isMethod().and(isPublic()).and(named("build")).and(takesArguments(0)),
          MongoClientInstrumentationModule.class.getName() + "$MongoClientAdvice");
    }
```

anuraaga (approved):
@malafeev sorry, was reading the type matcher, not the transformer!

In other agents we aren't really doing any deduping, so we could possibly remove it here too, but I guess if it causes a problem with testLatestDeps we could do so later too.

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,85 @@
...
+    return singletonList(new MongoClientSettingsBuilderInstrumentation());
+  }
+
+  private static final class MongoClientSettingsBuilderInstrumentation
+      implements TypeInstrumentation {
+    @Override
+    public ElementMatcher<TypeDescription> typeMatcher() {
+      return named("com.mongodb.MongoClientSettings$Builder")
+          .and(
+              declaresMethod(
```
iNikem, line 49: Why the existence of this method is important in type matcher if transformers are applied to a completely different method?

iNikem (changes requested):

instrumentation/mongo/mongo-4.0/javaagent/mongo-4.0-javaagent.gradle
```diff
@@ -0,0 +1,26 @@
...
+}
+
+dependencies {
+  implementation(project(':instrumentation:mongo:mongo-common:javaagent')) {
+    exclude group: 'org.mongodb', module: 'mongo-java-driver'
+  }
+
+  library group: 'org.mongodb', name: 'mongodb-driver-core', version: '4.0.0'
+
+  testCompile group: 'org.mongodb', name: 'mongodb-driver-sync', version: '4.0.0'
```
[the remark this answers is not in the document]
malafeev: done

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
malafeev: DD uses:
```
@Override
    return named("com.mongodb.MongoClientSettings$Builder")
        .and(declaresField(named("commandListeners")));
}
```  
but here all mongo instrumentations uses `named("addCommandListener")` in `typeMatcher`

I guess there is a reason why during porting it was changed.

instrumentation/mongo/mongo-4.0/javaagent/mongo-4.0-javaagent.gradle
```diff
@@ -0,0 +1,26 @@
...
+
+muzzle {
+  pass {
+    group = "org.mongodb"
+    module = "mongodb-driver-core"
+    // this instrumentation is backwards compatible with early versions of the new API that shipped in 3.7
+    // the legacy API instrumented in mongo-3.1 continues to be shipped in 4.x, but doesn't conflict here
+    // because they are triggered by different types: MongoClientSettings(new) vs MongoClientOptions(legacy)
+    versions = "[3.7,)"
+    assertInverse = true
```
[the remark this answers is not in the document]
malafeev: but  before 4.0, non-uber jars have also been published, that's why 3.7 is here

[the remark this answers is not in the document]
iNikem: I don't follow, sorry. If this module is meant for mongo 4.0+, then why do we want to verify muzzle for 3.7?

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
iNikem: Hm. @trask @tylerbenson any ideas?

tylerbenson: It was changed here on our side: https://github.com/DataDog/dd-trace-java/pull/2255

iNikem: I think the DD way makes more sense.

instrumentation/mongo/mongo-4.0/javaagent/mongo-4.0-javaagent.gradle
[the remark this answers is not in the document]
malafeev: there are 2 types of mongo jars:
1. uber jar. But it doesn't exist since version 4.0. 
2. non-uber jar which exists since version 3.7 up to current version.


DD named module to support non-uber jar `mongo-4`. 
Should I rename it to `mongo-non-uber-3.7`?

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
malafeev: fixed

instrumentation/mongo/mongo-4.0/javaagent/mongo-4.0-javaagent.gradle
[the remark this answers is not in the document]
iNikem: Wait, so actual instrumentations are the same, right? The only difference is, essentially, the collection of packages used for testing? So this is not a new module for instrumentation, but a new module for testing? Like `glassfish-testing`?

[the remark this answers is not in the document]
malafeev: It's an actual instrumentation for non-uber jars.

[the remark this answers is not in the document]
iNikem: How does it differ from already existing `io.opentelemetry.javaagent.instrumentation.mongo.v3_7.MongoClientInstrumentationModule`? Instrumentation looks the same to me.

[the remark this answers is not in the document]
malafeev: yes, instrumentation looks the same.
but mongo-3.7 will not work for `mongodb-driver-core` 3.7 jar because of muzzle.  
`mongo-3.7` depends on `mongo-java-driver`
`mongo-4.0` depends on `mongodb-driver-core`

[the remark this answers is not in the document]
mateuszrzeszutek: Instead of copying the 3.7 instrumentation, would it be possible to add another muzzle task there and just move the tests to a separate 4.0/uberjar testing module?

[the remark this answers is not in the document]
mateuszrzeszutek: Adding

```groovy
  pass {
    group = "org.mongodb"
    module = "mongodb-driver-core"
    // this instrumentation is backwards compatible with early versions of the new API that shipped in 3.7
    // the legacy API instrumented in mongo-3.1 continues to be shipped in 4.x, but doesn't conflict here
    // because they are triggered by different types: MongoClientSettings(new) vs MongoClientOptions(legacy)
    versions = "[3.7,)"
    extraDependency "org.mongodb:bson:3.7.0"
    assertInverse = true
  }
```

to the 3.7 muzzle task definition should cover all the uberjar releases.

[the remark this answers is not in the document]
malafeev: @mateuszrzeszutek done in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2046/commits/5e5ea23168589a266496e8d5b2e3d1461651a558

iNikem (approved):

PR #2047, opened 2021-01-14
Instrument spring-batch: item-level spans

mateuszrzeszutek:
Another part of spring batch instrumentation, last level of span hierarchy - item-level spans.
This time I couldn't use any listeners for instrumentation, so I'm instrumenting spring batch internal classes. And the item level instrumentation is very chatty, so perhaps it should stay an optional instrumentation even if spring batch ends up being enabled by default.

### DIFF
--- a/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
+++ b/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
@@ -15,19 +15,23 @@ public final class SpringBatchInstrumentationConfig {
   private static final List<String> INSTRUMENTATION_NAMES =
       unmodifiableList(asList("spring-batch", "spring-batch-3.0"));
 
+  // the item level instrumentation is very chatty so it's disabled by default
+  private static final boolean ITEM_LEVEL_TRACING_ENABLED =
+      Config.get().isInstrumentationPropertyEnabled(instrumentationNames(), "item.enabled", false);
+  private static final boolean CREATE_ROOT_SPAN_FOR_CHUNK =
+      Config.get()
+          .isInstrumentationPropertyEnabled(instrumentationNames(), "chunk.root-span", false);
+
   public static List<String> instrumentationNames() {
     return INSTRUMENTATION_NAMES;
   }
 
-  // the item level instrumentation is very chatty so it's disabled by default
   public static boolean isItemLevelTracingEnabled() {
-    return Config.get()
-        .isInstrumentationPropertyEnabled(instrumentationNames(), "item.enabled", false);
+    return ITEM_LEVEL_TRACING_ENABLED;
   }
 
   public static boolean shouldCreateRootSpanForChunk() {
-    return Config.get()
-        .isInstrumentationPropertyEnabled(instrumentationNames(), "chunk.root-span", false);
+    return CREATE_ROOT_SPAN_FOR_CHUNK;
   }
 
   private SpringBatchInstrumentationConfig() {}
--- a/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
+++ b/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
@@ -19,9 +19,9 @@ public final class SpringBatchInstrumentationConfig {
     return INSTRUMENTATION_NAMES;
   }
 
-  public static boolean isTracingEnabled(String type) {
+  public static boolean isTracingEnabled(String type, boolean defaultEnabled) {
     return Config.get()
-        .isInstrumentationPropertyEnabled(instrumentationNames(), type + ".enabled", true);
+        .isInstrumentationPropertyEnabled(instrumentationNames(), type + ".enabled", defaultEnabled);
   }
 
   public static boolean shouldCreateRootSpanForChunk() {
--- a/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
+++ b/instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
@@ -16,7 +16,7 @@ public final class SpringBatchInstrumentationConfig {
       unmodifiableList(asList("spring-batch", "spring-batch-3.0"));
 
   // the item level instrumentation is very chatty so it's disabled by default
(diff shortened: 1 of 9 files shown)


instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemTracer.java
```diff
@@ -0,0 +1,88 @@
...
+    return TRACER;
+  }
+
+  // item-level listeners do not receive chunk/step context as parameters
+  // fortunately the whole chunk always executes on one thread - in Spring Batch chunk is almost
+  // synonymous with a DB transaction
+  // that makes it quite safe to store the current chunk context in a thread local and use it in
+  // each item span
+
+  private final ThreadLocal<WeakReference<ChunkContext>> currentChunk = new ThreadLocal<>();
```
anuraaga, line 32: The ThreadLocal cleanup seems easy to reason about, I'd probably skip the weak reference.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemTracer.java
```diff
@@ -0,0 +1,88 @@
...
+        Optional.ofNullable(currentChunk.get()).map(Reference::get).orElse(null);
+    if (chunkContext == null) {
+      return null;
+    }
+
+    String jobName = chunkContext.getStepContext().getJobName();
+    String stepName = chunkContext.getStepContext().getStepName();
+    Span span =
+        tracer
+            .spanBuilder("BatchJob " + jobName + "." + stepName + "." + itemOperationName)
```
anuraaga, line 69: It'd definitely be nice to get some spec issues going for this batch processing stuff

mateuszrzeszutek: Yes, definitely - I've filed a [spec issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1347) about that (now that my vacation is over I can start working on that 😄 ).

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemTracer.java
mateuszrzeszutek: Done!

trask:
Thanks for calling this out:

> the item level instrumentation is very chatty, so perhaps it should stay an optional instrumentation even if spring batch ends up being enabled by default

It sounds like `otel.instrumentation.spring-batch.item.enabled` should be false by default even now then?

anuraaga (approved):

mateuszrzeszutek:
> It sounds like otel.instrumentation.spring-batch.item.enabled should be false by default even now then?

Yes, definitely -- done.

mateuszrzeszutek:
By the way I ran into an interesting issue with muzzle here: when item level instrumentation was disabled and type instrumentations were not added to the list, by default the `byteBuddyjava` task didn't see advice classes used by those instrumentations - and thus did not register helper classes. I've changed the way those instrumentations are disabled, and now every type instrumentation is always active, but the configuration switch is moved to the advice class.

anuraaga (approved):

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/SpringBatchInstrumentationConfig.java
```diff
@@ -15,18 +15,23 @@
...
+          .isInstrumentationPropertyEnabled(instrumentationNames(), "chunk.root-span", false);
+
   public static List<String> instrumentationNames() {
     return INSTRUMENTATION_NAMES;
   }
 
-  public static boolean isTracingEnabled(String type) {
-    return Config.get()
-        .isInstrumentationPropertyEnabled(instrumentationNames(), type + ".enabled", true);
+  public static boolean isItemLevelTracingEnabled() {
```
iNikem, line 29: Should this be named similar to `shouldCreateRootSpanForChunk`?

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ChunkOrientedTaskletInstrumentation.java
```diff
@@ -0,0 +1,58 @@
...
+            .and(
+                takesArgument(
+                    1, named("org.springframework.batch.core.scope.context.ChunkContext"))),
+        this.getClass().getName() + "$ExecuteAdvice");
+  }
+
+  public static class ExecuteAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(@Advice.Argument(1) ChunkContext chunkContext) {
+      if (isItemLevelTracingEnabled()) {
```
iNikem, line 46: So we always instrument classes, even if we actually end up doing nothing? Not sure if this is a good idea...

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemTracer.java
```diff
@@ -0,0 +1,84 @@
...
+    return TRACER;
+  }
+
+  // item-level listeners do not receive chunk/step context as parameters
+  // fortunately the whole chunk always executes on one thread - in Spring Batch chunk is almost
+  // synonymous with a DB transaction
+  // that makes it quite safe to store the current chunk context in a thread local and use it in
+  // each item span
+
+  private final ThreadLocal<ChunkContext> currentChunk = new ThreadLocal<>();
```
iNikem, line 29: Not opposing this solution, but did you consider using our tracing `Context` for this?

mateuszrzeszutek: No, it did not even register in my mind... but now that I think of it, I like it more than the thread local variable - I'll refactor this. Thanks!

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ChunkOrientedTaskletInstrumentation.java
mateuszrzeszutek: I think that since this is a static final boolean constant JIT will remove this code and it won't matter at all in the end. But I agree that the previous version looked better - muzzle had problems accepting it though (couldn't see the instrumentations that are disabled by default) so I decided checking the flag in advice classes is still better than `additionalHelperClassNames()`

iNikem (approved):

PR #2052, opened 2021-01-14
fixing Apache Camel instrumentation propagation order

### LINKED ISSUE #2051: Apache Camel instrumentation - wrong parent propagated
**Describe the bug**
Apache Camel instrumentation contains a bug, causing wrong parent span to be propagated. Instead of initiating outbound (client, producer) span, a parent is used.

**Steps to reproduce**
Any use case where there is a propagation involved (eh a route: "direct -> http -> rest")

HTTP client span being a parent of the rest server span.

**What did you see instead?**
Direct span is parent of both client and server spans.

master 

**Environment**
NA

**Additional context**
NA

kubawach:
Fixes #2051

- current context is injected as parent for propagation AFTER the "current" span has been activated
- test cases now check parent/child relationship! :)


breedx-splk (approved):
This looks good to me.  A simple ordering fix with test coverage!

trask (approved):
👍

anuraaga (approved):

PR #2054, opened 2021-01-14
tweak the output of the logging exporter:

jkwatson:
* one span per line
* delineate the span name from the rest of the data


javaagent-exporters/logging/src/main/java/io/opentelemetry/javaagent/exporters/logging/LoggingExporter.java
```diff
@@ -25,13 +25,14 @@ public LoggingExporter(String prefix) {
   public CompletableResultCode export(Collection<SpanData> list) {
     StringBuilder stringBuilder = new StringBuilder();
     for (SpanData span : list) {
-
+      stringBuilder.setLength(0);
```
anuraaga, line 28: Ouch, heh.

Are you interested in deleting this to switch to the SDK exporter?

jkwatson: anything that gets us one span per line would be good for me. ;)

trask (approved):

anuraaga (approved):

PR #2059, opened 2021-01-15
Update to OTel 0.14.1

anuraaga:


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/exporters/otlp/OtlpGrpcSpanExporterProperties.java
```diff
@@ -5,8 +5,8 @@
 
 package io.opentelemetry.instrumentation.spring.autoconfigure.exporters.otlp;
 
-import static io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter.DEFAULT_DEADLINE_MS;
 import static io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter.DEFAULT_ENDPOINT;
+import static io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter.DEFAULT_TIMEOUT_MS;
```
anuraaga, line 9: @jkwatson Looks like I accidentally breaking-changed this field during some WIP on the timeout PRs

instrumentation/oshi/library/src/test/java/io/opentelemetry/instrumentation/oshi/AbstractMetricsTest.java
```diff
@@ -7,60 +7,77 @@
...
-      String metricName, String unit, MetricData.Type type, boolean checkNonZeroValue) {
+      String metricName, String unit, MetricDataType type, boolean checkNonZeroValue) {
     List<MetricData> metricDataList = testMetricExporter.metricDataList;
     for (MetricData metricData : metricDataList) {
       if (metricData.getName().equals(metricName)) {
         assertThat(metricData.getDescription()).isNotEmpty();
         assertThat(metricData.getUnit()).isEqualTo(unit);
-        assertThat(metricData.getPoints()).isNotEmpty();
+        metricData.getDoubleGaugeData().getPoints();
+        List<Point> points = new ArrayList<>();
```
anuraaga, line 62: @jkwatson Don't know if it matters, but the impact of losing `getPoints` was relatively large. Presumably I should not be instanceof below, but without doing a deeper dive into this test it's not easy to know what's a long gauge or a long sum so used this pattern.

mateuszrzeszutek (approved):

pavolloffay:
@anuraaga @trask @iNikem roughly when the 0.14.0 will be released?

instrumentation/oshi/library/src/test/java/io/opentelemetry/instrumentation/oshi/AbstractMetricsTest.java
jkwatson: paging @bogdandrutu

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/exporters/otlp/OtlpGrpcSpanExporterProperties.java
jkwatson: do you think we need a patch to address it?

instrumentation/oshi/library/src/test/java/io/opentelemetry/instrumentation/oshi/AbstractMetricsTest.java
jkwatson: This doesn't look like a huge impact to me. Am I missing something?

instrumentation/oshi/library/src/test/java/io/opentelemetry/instrumentation/oshi/AbstractMetricsTest.java
```diff
@@ -7,60 +7,77 @@
...
   public void verify(
-      String metricName, String unit, MetricData.Type type, boolean checkNonZeroValue) {
+      String metricName, String unit, MetricDataType type, boolean checkNonZeroValue) {
     List<MetricData> metricDataList = testMetricExporter.metricDataList;
     for (MetricData metricData : metricDataList) {
       if (metricData.getName().equals(metricName)) {
         assertThat(metricData.getDescription()).isNotEmpty();
         assertThat(metricData.getUnit()).isEqualTo(unit);
-        assertThat(metricData.getPoints()).isNotEmpty();
+        metricData.getDoubleGaugeData().getPoints();
```
bogdandrutu, line 61: This is a no-op line?

instrumentation/oshi/library/src/test/java/io/opentelemetry/instrumentation/oshi/AbstractMetricsTest.java
bogdandrutu: I would actually merge this logic with the next 20 lines:

switch metric.getType() {
  case GaugeDouble:
    List<DoublePoints> dpts = metricData.getDoubleGaugeData().getPoints();
    dpts.foreach(doublePoint -> assertThat(doublePoint.getValue()).isGreaterThan(0.0));
  ....
}

Or something like that (written from the phone so no access to IDE 👍 )

trask (approved):
thanks!! 😅

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/exporters/otlp/OtlpGrpcSpanExporterProperties.java
anuraaga: Nah don't think so we should just update the release notes

instrumentation/oshi/library/src/test/java/io/opentelemetry/instrumentation/oshi/AbstractMetricsTest.java
anuraaga: I would have to understand the test much better to e.g. understand whether to use gauge vs sum. Will go with my brain-dead change for now and hopefully someone can give some maintenance to the test, maybe after the metrics stuff stabilizes some more.

PR #2061, opened 2021-01-15
adding AWS SQS tests to Apache Camel instrumentation

### LINKED ISSUE #2060: SQS tests for Apache Camel instrumentation
Apache Camel instrumentation lacks SQS tests (especially important for ensuring propagation).l

Tests should be added :) 

NA

**Additional context**
SQS context propagation investigation in several use cases, also involving Apache Camel.

kubawach:
Closes #2060 

Adding SQS tests for Apache Camel. 
For this purpose, additional library (SQS local mock in fact) is added - https://github.com/softwaremill/elasticmq, licensed under Apache 2.0 (https://github.com/softwaremill/elasticmq/blob/master/LICENSE.txt).

### DIFF
--- a/instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
+++ b/instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
@@ -37,7 +37,7 @@ class SqsCamelTest extends AgentTestRunner {
     println getClass().name + " SQS server started at: localhost:$sqsPort/"
 
     def app = new SpringApplication(SqsConfig)
-    app.setDefaultProperties(ImmutableMap.of("sqs.port", sqsPort))
+    app.setDefaultProperties(["sqs.port": sqsPort]))
     server = app.run()
   }
 
@@ -178,4 +178,3 @@ class SqsCamelTest extends AgentTestRunner {
     }
   }
 }
-
--- a/instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
+++ b/instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
@@ -1,0 +1,183 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test
+
+import static io.opentelemetry.api.trace.Span.Kind.CLIENT
+import static io.opentelemetry.api.trace.Span.Kind.CONSUMER
+import static io.opentelemetry.api.trace.Span.Kind.INTERNAL
+import static io.opentelemetry.api.trace.Span.Kind.PRODUCER
+
+import com.google.common.collect.ImmutableMap
+import io.opentelemetry.instrumentation.test.AgentTestRunner
+import io.opentelemetry.instrumentation.test.utils.PortUtils
+import org.apache.camel.CamelContext
+import org.apache.camel.ProducerTemplate
+import org.elasticmq.rest.sqs.SQSRestServer
+import org.elasticmq.rest.sqs.SQSRestServerBuilder
+import org.springframework.boot.SpringApplication
+import org.springframework.context.ConfigurableApplicationContext
+import spock.lang.Shared
+
+class SqsCamelTest extends AgentTestRunner {
+
+  @Shared
+  ConfigurableApplicationContext server
+
+  @Shared
+  SQSRestServer sqs
+  @Shared
+  int sqsPort
+
+  def setupSpec() {
+    sqsPort = PortUtils.randomOpenPort()
+    sqs = SQSRestServerBuilder.withPort(sqsPort).withInterface("localhost").start()
+    println getClass().name + " SQS server started at: localhost:$sqsPort/"
+
+    def app = new SpringApplication(SqsConfig)
+    app.setDefaultProperties(ImmutableMap.of("sqs.port", sqsPort))
+    server = app.run()
+  }
+
+  def cleanupSpec() {
+    if (server != null) {
+      server.close()
+      server = null
+    }
+    if (sqs != null) {
+      sqs.stopAndWait()
+    }
+  }
+
+  def "simple sqs producer-consumer services"() {
+    setup:
+    def camelContext = server.getBean(CamelContext)
+    ProducerTemplate template = camelContext.createProducerTemplate()
+
+    when:
+    template.sendBody("direct:input", "{\"type\": \"hello\"}")
+    // wait for SQS to send back
+    Thread.sleep(2000)
+
+    then:
+    assertTraces(5) {
+      trace(0, 3) {
+        def parent = it
+        it.span(0) {
+          name "input"
+          kind INTERNAL
+          hasNoParent()
+          attributes {
+            "apache-camel.uri" "direct://input"
+          }
+        }
+        it.span(1) {
+          name "sqsCamelTest"
+          kind PRODUCER
+          parentSpanId parent.span(0).spanId
+          attributes {
+            "apache-camel.uri" "aws-sqs://sqsCamelTest?amazonSQSClient=%23sqsClient"
+            "messaging.destination" "sqsCamelTest"
+          }
+        }
+        it.span(2) {
+          name "sqsCamelTest"
+          kind CONSUMER
+          parentSpanId parent.span(1).spanId
+          attributes {
+            "apache-camel.uri" "aws-sqs://sqsCamelTest?amazonSQSClient=%23sqsClient&messageAttributeNames=traceparent"
+            "messaging.destination" "sqsCamelTest"
+            "messaging.message_id" String
+          }
+        }
+      }
+      trace(1, 1) {
+        it.span(0) {
+          name "SQS.ReceiveMessage"
+          kind CLIENT
+          hasNoParent()
+          attributes {
+            "aws.agent" "java-aws-sdk"
+            "aws.endpoint" "http://localhost:$sqsPort"
+            "aws.operation" "ReceiveMessageRequest"
+            "aws.queue.url" "http://localhost:$sqsPort/queue/sqsCamelTest"
+            "aws.service" "AmazonSQS"
+            "http.flavor" "1.1"
+            "http.method" "POST"
+            "http.status_code" 200
+            "http.url" "http://localhost:$sqsPort"
+            "net.peer.name" "localhost"
+            "net.peer.port" sqsPort
+            "net.transport" "IP.TCP"
+          }
+        }
+      }
+      trace(2, 1) {
+        it.span(0) {
+          name "SQS.DeleteMessage"
+          kind CLIENT
+          hasNoParent()
+          attributes {
+            "aws.agent" "java-aws-sdk"
+            "aws.endpoint" "http://localhost:$sqsPort"
+            "aws.operation" "DeleteMessageRequest"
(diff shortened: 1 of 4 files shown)


instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,183 @@
...
+import org.springframework.context.ConfigurableApplicationContext
+import spock.lang.Shared
+
+class SqsCamelTest extends AgentTestRunner {
+
+  @Shared
+  ConfigurableApplicationContext server
+
+  @Shared
+  SQSRestServer sqs
```
mateuszrzeszutek, line 30: Nit: have you considered running localstack testcontainers instead: https://www.testcontainers.org/modules/localstack/ ?

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsConfig.groovy
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package test
+
+
```
mateuszrzeszutek, line 8:
```suggestion
```

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,183 @@
...
+
+  def "simple sqs producer-consumer services"() {
+    setup:
+    def camelContext = server.getBean(CamelContext)
+    ProducerTemplate template = camelContext.createProducerTemplate()
+
+    when:
+    template.sendBody("direct:input", "{\"type\": \"hello\"}")
+    // wait for SQS to send back
+    Thread.sleep(2000)
```
mateuszrzeszutek, line 62: Instead of waiting consider adding a processor calling `CyclicBarrier#await()` in the SQS consumer route - a similar thing is done in `JaxRsHttpServerTest`.

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,183 @@
...
+          kind INTERNAL
+          hasNoParent()
+          attributes {
+            "apache-camel.uri" "direct://input"
+          }
+        }
+        it.span(1) {
+          name "sqsCamelTest"
+          kind PRODUCER
+          parentSpanId parent.span(0).spanId
```
mateuszrzeszutek, line 79: You can use a bit shorter assertion:

```suggestion
          childOf span(0)
```

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,183 @@
...
+          parentSpanId parent.span(1).spanId
+          attributes {
+            "apache-camel.uri" "aws-sqs://sqsCamelTest?amazonSQSClient=%23sqsClient&messageAttributeNames=traceparent"
+            "messaging.destination" "sqsCamelTest"
+            "messaging.message_id" String
+          }
+        }
+      }
+      trace(1, 1) {
+        it.span(0) {
```
mateuszrzeszutek, line 97: I think that there's no need for `it.` here, just `span(0)` should work:

```suggestion
        span(0) {
```

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,183 @@
...
+            "net.peer.name" "localhost"
+            "net.peer.port" sqsPort
+            "net.transport" "IP.TCP"
+          }
+        }
+      }
+      trace(3, 1) {
+        it.span(0) {
+          name "SQS.ReceiveMessage"
+          kind CLIENT
```
mateuszrzeszutek, line 141: Just a question: can you explain those separate AWS CLIENT traces? I'm interested in a reason why those spans appear separately

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: Localstack uses elasticMQ underneath :D

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: That's really nice!

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: I think that right now it's not even needed (local SQS).

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: Nice tip, done.

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: Delete is easy - it's just "cleanup" done after processing batch, so after produce / consume trace is done.

kubawach: "Receive" spans are quite a different story - it took me while to understand whole process :D 

Basically, when SQS consumer (bean) is up, it starts sending "Receive" messages to SQS. Each such message can be seen as a CONSUMER, but only after a message is in fact available and consumed. So, once one of the messages succeeds, Camel starts "consumer" route and that's where Camel's instrumentation kicks in, creating a CONSUMER span. As there is a propagation enabled for Camel, the CONSUMER is tied to Camel's PRODUCER span issued earlier. So far so good. AWS SDK instrumentation does not have a trace propagation (!), therefore "RECEIVE" span will hang detached from the internal-producer-consumer trace.

Now things are getting interesting. Once I enable trace propagation for AWS SDK (and change span KIND for SQS "ReceiveMessage" which IMO should be CONSUMER not CLIENT), there are two options:
- two CONSUMER spans in the trace: internal - producer (camel) - consumer (AWS SDK) - consumer (camel)
- camel consumer will be suppressed (!) and the trace will have one consumer span - the one from AWS SDK: internal - producer (camel) - consumer (AWS SDK)

@anuraaga @trask @jkwatson I'd very much like to hear your opinions on the above question (what should happen when there is Camel and AWS SDK with trace propagation) ?

By the way - thanks @mateuszrzeszutek for bringing up this matter :)

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,181 @@
...
+import org.springframework.context.ConfigurableApplicationContext
+import spock.lang.Shared
+
+class SqsCamelTest extends AgentTestRunner {
+
+  @Shared
+  ConfigurableApplicationContext server
+
+  @Shared
+  SQSRestServer sqs
```
anuraaga, line 30: I'd still prefer localstack over running the server inline - we've generally had better stability running these sort of servers through testcontainers than inline even if they're Java (e.g., cassandra), generally because docker does a better job of randomizing the port than `PortUtils`

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
```diff
@@ -0,0 +1,181 @@
...
+  @Shared
+  int sqsPort
+
+  def setupSpec() {
+    sqsPort = PortUtils.randomOpenPort()
+    sqs = SQSRestServerBuilder.withPort(sqsPort).withInterface("localhost").start()
+    println getClass().name + " SQS server started at: localhost:$sqsPort/"
+
+    def app = new SpringApplication(SqsConfig)
+    app.setDefaultProperties(ImmutableMap.of("sqs.port", sqsPort))
```
anuraaga, line 40:
```suggestion
    app.setDefaultProperties(["sqs.port": sqsPort]))
```

kubawach: Thanks, nice syntax sugar :)

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: I considered full blown localstack (which uses elasticmq underneath) but figured out it's too heavy (adding a dockerized container - making test slower and more complex) for such a small use case (eventually a few test cases for not-that-critical instrumentation). So basically a cost/effect thing. But since we have testcontainers already in the stack... :)

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
anuraaga: As long as that small use case doesn't cause flakiness anything's fine :) We have just had better success with test containers I think.

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
anuraaga: More because of the coming change to remove Guava :)

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
kubawach: Done :)

kubawach:
Switched to testcontainers / localstack. Pls have another look at this.

anuraaga (approved):
Thanks!

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/test/SqsCamelTest.groovy
mateuszrzeszutek: I think it'd make the most sense if camel spans were internal - in the context of this test:

```
INTERNAL input (camel)
--> INTERNAL sqsCamelTest (camel)
    --> PRODUCER SQS.SendMessage (AWS SDK)
        --> CONSUMER SQS.ReceiveMessage (AWS SDK)
            --> INTERNAL sqsCamelTest (camel)
```

Camel is just a wrapper library that does not send any message to any queue directly, it relies on other libraries - and it makes sense for the camel instrumentation to rely on other instrumentations (AWS SDK here).

We need to have trace propagation in AWS SDK and change SQS spans to PRODUCER/CONSUMER for it to happen though.

mateuszrzeszutek (approved):

Issue #2048, opened 2021-01-14
Remove testing-common dependency on guava


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Right now `testing-common` depends on guava 20 and it's impossible to update the guava version because several instrumentations use/depend on this version (cassandra, grizzly, ratpack, etc).
With #1643 being merged the javaagent code now uses a separate, shaded guava that will not interfere with the older version used by the instrumented lib (and if it does then it's a bug).
After we remove the test dependency on guava javaagent will use its own version, the instrumented library under test will use its own and we'll be free to upgrade the javaagent one to the newest version.

PR #2062, opened 2021-01-15
Remove testing-common dependency on guava

mateuszrzeszutek:
And bump guava version to the latest one

Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2048

### DIFF
--- a/testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/TraceAssert.groovy
+++ b/testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/TraceAssert.groovy
@@ -27,7 +27,7 @@ class TraceAssert {
                           @ClosureParams(value = SimpleType, options = ['io.opentelemetry.instrumentation.test.asserts.TraceAssert'])
                           @DelegatesTo(value = TraceAssert, strategy = Closure.DELEGATE_FIRST) Closure spec) {
     def spans = getTrace(spanSupplier, traceId)
-    def startTime = System.currentTimeMillis()
+    def startTime = System.nanoTime()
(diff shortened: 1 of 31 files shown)


trask (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/TraceAssert.groovy
```diff
@@ -28,8 +27,8 @@ class TraceAssert {
                           @ClosureParams(value = SimpleType, options = ['io.opentelemetry.instrumentation.test.asserts.TraceAssert'])
                           @DelegatesTo(value = TraceAssert, strategy = Closure.DELEGATE_FIRST) Closure spec) {
     def spans = getTrace(spanSupplier, traceId)
-    Stopwatch stopwatch = Stopwatch.createStarted()
-    while (spans.size() < expectedSize && stopwatch.elapsed(TimeUnit.SECONDS) < 10) {
+    def startTime = System.currentTimeMillis()
```
anuraaga, line 30: Use nanoTime for measurements, not wall time (this would be the previous Guava behavior)

Same for all the similar code

anuraaga (commented):
Thanks a lot! Just one comment

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/TraceAssert.groovy
mateuszrzeszutek: 👍 

Done!

anuraaga (approved):

Issue #2031, opened 2021-01-12
jaxrs-client instrumentation causes failure in JerseyInvocation due to CompletableFuture cast


ben-childs-docusign:
**Describe the bug**
After adding the opentelemetry java agent to our existing application we saw this error during startup:
```
java.lang.ClassCastException: class io.opentelemetry.javaagent.instrumentation.jaxrsclient.v2_0.WrappedFuture cannot be cast to class java.util.concurrent.CompletableFuture 
        at org.glassfish.jersey.client.JerseyInvocation$AsyncInvoker.method(JerseyInvocation.java:556)
	at org.glassfish.jersey.client.JerseyInvocation$AsyncInvoker.method(JerseyInvocation.java:535)
	at com.spotify.docker.client.DefaultDockerClient.request(DefaultDockerClient.java:2710)
	at com.spotify.docker.client.DefaultDockerClient.version(DefaultDockerClient.java:600)
	at com.spotify.docker.client.DefaultDockerClient.assertApiVersionIsAbove(DefaultDockerClient.java:2910)
	at com.spotify.docker.client.DefaultDockerClient.listNodes(DefaultDockerClient.java:2239)
```

**Steps to reproduce**
Install the open telemetry agent in a service that uses the docker client or any other library that uses JerseyInvocation$AsyncInvoker.method

**What did you expect to see?**
Tracing should not break JerseyInvocation$AsyncInvoker

**What version are you using?**
v0.12.0

**Environment**
Compiler: AdoptOpenJDK 11.0.9
OS: MacOS X 10.15 

**Additional context**
After reviewing the jersey code it looks like it expects to receive a `CompletableFuture<T>` from JerseyInvocation.submit:
https://github.com/eclipse-ee4j/jersey/blob/95c08d35331478ce47c4ae339c7f8542bf69fa22/core-client/src/main/java/org/glassfish/jersey/client/JerseyInvocation.java
```java
        @Override
        public <T> CompletableFuture<T> method(final String name, final GenericType<T> responseType) {
            if (responseType == null) {
                throw new IllegalArgumentException(LocalizationMessages.RESPONSE_TYPE_IS_NULL());
            }
            builder.requestContext.setMethod(name);
            return (CompletableFuture<T>) new JerseyInvocation(builder).submit(responseType);
        }
```

However the jersey instrumentation code wraps the future using `WrappedFuture<T>` which inherits from `Future<T>`
[...]

PR #2071, opened 2021-01-18
Fix integration with jersey-client 2.30

laurit:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2031
We can't use WrappedFuture on jersey-client 2.30 as  jersey-client attempts to cast it to CompletableFuture.

### DIFF
--- a/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/jaxrs-client-2.0-common-javaagent.gradle
+++ b/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/jaxrs-client-2.0-common-javaagent.gradle
@@ -44,8 +44,8 @@ dependencies {
 
   testInstrumentation project(':instrumentation:apache-httpclient:apache-httpclient-4.0:javaagent')
 
-  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '[2+, 3)'
-  latestDepTestLibrary group: 'org.glassfish.jersey.core', name: 'jersey-client', version: '[2+, 3)'
+  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '2.+'
+  latestDepTestLibrary group: 'org.glassfish.jersey.core', name: 'jersey-client', version: '2.+'
   latestDepTestLibrary group: 'org.apache.cxf', name: 'cxf-rt-rs-client', version: '3.2.6'
   latestDepTestLibrary group: 'org.jboss.resteasy', name: 'resteasy-client', version: '3.0.26.Final'
 }
--- a/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/jaxrs-client-2.0-common-javaagent.gradle
+++ b/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/jaxrs-client-2.0-common-javaagent.gradle
@@ -44,8 +44,8 @@ dependencies {
 
   testInstrumentation project(':instrumentation:apache-httpclient:apache-httpclient-4.0:javaagent')
 
-  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '2.27'
-  latestDepTestLibrary group: 'org.glassfish.jersey.core', name: 'jersey-client', version: '2.27'
+  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '[2+, 3)'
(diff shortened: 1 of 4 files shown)


instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/JerseyClientUtil.java
```diff
@@ -0,0 +1,45 @@
...
+package io.opentelemetry.javaagent.instrumentation.jaxrsclient.v2_0;
+
+import static io.opentelemetry.javaagent.instrumentation.jaxrsclient.v2_0.JaxRsClientTracer.tracer;
+
+import io.opentelemetry.context.Context;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Future;
+import org.glassfish.jersey.client.ClientRequest;
+
+public class JerseyClientUtil {
```
mateuszrzeszutek, line 15: Nit: make it final and add a private constructor.

mateuszrzeszutek (approved):

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/JerseyClientUtil.java
laurit: done

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/JerseyClientUtil.java
```diff
@@ -0,0 +1,47 @@
...
+    } else {
+      if (!(future instanceof WrappedFuture)) {
+        future = new WrappedFuture<>(future, context);
+      }
+    }
+
+    return future;
+  }
+
+  public static void handleException(ClientRequest context, Throwable exception) {
```
anuraaga, line 39: We generally add instrumentation-specific handling to the `*Tracer` classes themselves instead of separating utils, would that work here?

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/jaxrs-client-2.0-common-javaagent.gradle
```diff
@@ -44,8 +44,8 @@ dependencies {
 
   testInstrumentation project(':instrumentation:apache-httpclient:apache-httpclient-4.0:javaagent')
 
-  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '2.27'
-  latestDepTestLibrary group: 'org.glassfish.jersey.core', name: 'jersey-client', version: '2.27'
+  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '[2+, 3)'
+  latestDepTestLibrary group: 'org.glassfish.jersey.core', name: 'jersey-client', version: '[2+, 3)'
```
trask, line 48: this is more consistent with other gradle files:
```suggestion
  latestDepTestLibrary group: 'org.glassfish.jersey.inject', name: 'jersey-hk2', version: '2.+'
  latestDepTestLibrary group: 'org.glassfish.jersey.core', name: 'jersey-client', version: '2.+'
```

trask (approved):
Thanks! One suggestion below, in addition to @anuraaga's suggestion.

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/JerseyClientUtil.java
laurit: Tracer class is in a different module `jaxrs-client-2.0-common` adding these methods there would also require moving `WrappedFuture` there. Adding a tracer to `jaxrs-client-2.0-jersey-2.0` also doesn't seem like a good option because this module doesn't really have any specialised tracing code.

iNikem (approved):

PR #2074, opened 2021-01-19
Replae grpc-netty-shaded with grpc-netty

anuraaga:
The main reason to use shaded is to prevent classpath conflicts, but we're in an isolated classloader and should be able to manage this ok. The other reason / downside is shaded includes 3 versions of boringssl for linux, mac, and windows. Now that latest versions of Java 8 include all the TLS features of Java 11, I think this is not as helpful as before (I've done some benchmarking and found very little difference with some advantages for JDK TLS) so we can save 3MB on our agent jar by using grpc-netty and not adding tcnative.

### DIFF
--- a/javaagent-exporters/javaagent-exporters.gradle
+++ b/javaagent-exporters/javaagent-exporters.gradle
@@ -77,3 +77,16 @@ shadowJar {
   relocate "io.opentelemetry.extension.kotlin", "io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.kotlin"
   relocate "io.opentelemetry.extension.trace.propagation", "io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.trace.propagation"
 }
+
+// We can't apply boms globally to all projects because instrumentation needs fine-grained control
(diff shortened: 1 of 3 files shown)


javaagent-exporters/javaagent-exporters.gradle
```diff
@@ -77,3 +77,26 @@ shadowJar {
...
   relocate "io.opentelemetry.extension.trace.propagation", "io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.trace.propagation"
 }
+
+// We can't apply boms globally to all projects because instrumentation needs fine-grained control
+// of versions. For now we just add to exporters which get the most benefit.
+subprojects {
+  dependencies {
+    configurations.all {
+      // Kotlin compiler classpaths don't support BOM nor need it.
+      if (it.name.endsWith('Classpath') && !it.name.startsWith('kotlin')) {
```
trask, line 87: do we have any kotlin dependencies from javaagent-exporters?

trask (approved):

anuraaga:
Ended up deleting the whole folder in another PR, let me just close this one

PR #2077, opened 2021-01-19
Use SDK Autoconfigure module

anuraaga:
I realized that right now there's no way to configure ID generator with agent 0.14.0 - we relied on OpenTelemetry SPI for that, but don't use SPI anymore, instead using the `OpenTelemetrySdkBuilder`. So I think I'm going to need to release a patch version with this change. And presumably more patches as we work out any more kinks introduced which there must be :)


javaagent-tooling/javaagent-tooling.gradle
```diff
@@ -37,6 +44,8 @@ dependencies {
   implementation deps.autoservice
   implementation deps.slf4j
 
+  implementation group: 'io.grpc', name: 'grpc-netty', version: '1.34.1'
```
anuraaga, line 47: Not using `shaded`. Not bothering with bom for now since it's just one project

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentTracerProviderConfigurer.java
```diff
@@ -0,0 +1,123 @@
...
+      String exporterJar, Properties config, SdkTracerProviderBuilder builder) {
+    URL url;
+    try {
+      url = new File(exporterJar).toURI().toURL();
+    } catch (MalformedURLException e) {
+      log.warn("Filename could not be parsed: " + exporterJar + ". Exporter is not installed");
+      log.warn("No valid exporter found. Tracing will run but spans are dropped");
+      return;
+    }
+    ExporterClassLoader exporterLoader =
```
anuraaga, line 68: Going to add an exporter SPI to SDK since it seems to make sense. But I wonder if it would be possible to use from agent given we load external JARs into separate classloader like this. Any idea?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -53,181 +31,60 @@ public void beforeByteBuddyAgent() {
...
-      } else {
-        // Try to create embedded exporter
-        installExporters(Config.get().getListProperty(EXPORTERS_CONFIG, DEFAULT_EXPORTERS), config);
-      }
+      copySystemProperties();
+
+      OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
+      // TODO(anuraaga): Remove this workdaround after autoconfiguration is fixed to actually
+      // register the global.
+      GlobalOpenTelemetry.set(sdk);
```
anuraaga, line 40: Oops

iNikem:
@anuraaga Can we please not hurry with merging this PR? I want to get up to speed with new SDK configuration story in order to better understand this PR. But don't know how much time will it take. And I am VERY interested in this topic :)

anuraaga:
@iNikem Hopefully we can merge this week? :)

javaagent-tooling/javaagent-tooling.gradle
iNikem: Why not using `shaded`?

javaagent-tooling/javaagent-tooling.gradle
```diff
@@ -21,14 +21,21 @@ dependencies {
...
   implementation deps.opentelemetrySdk
+  implementation deps.opentelemetrySdkAutoconfigure
   implementation deps.opentelemetrySdkMetrics
   implementation deps.opentelemetryKotlin
   implementation deps.opentelemetryTraceProps
   implementation(deps.opentelemetryResources) {
     // exclude sdk-common to avoid bumping guava version
     exclude group: 'io.opentelemetry', module: 'opentelemetry-sdk-common'
   }
+  implementation deps.opentelemetryZipkin
```
iNikem, line 38: Move somewhere closer to all other exporters? Or sort alphabetically?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -53,181 +31,60 @@ public void beforeByteBuddyAgent() {
...
-
-  private static synchronized void installExportersFromJar(String exporterJar, Properties config) {
-    URL url;
-    try {
-      url = new File(exporterJar).toURI().toURL();
-    } catch (MalformedURLException e) {
-      log.warn("Filename could not be parsed: " + exporterJar + ". Exporter is not installed");
-      log.warn("No valid exporter found. Tracing will run but spans are dropped");
-      return;
+  // OpenTelemetrySdkAutoConfiguration currently only supports configuration from environment. We
```
iNikem, line 48: But `io/opentelemetry/sdk/autoconfigure/ConfigProperties.java:24` does read from system properties?

And I still don't quite understand what does this method do :)

anuraaga: By environment, I meant both env vars and system properties here.

This method is to forward properties to the autoconfiguration for agent's unique sources, file and `PropertySource`. We may or may not want to move the handling of these to autoconfigure, or do something different but this is to play with a first iteration and works ok I think.

javaagent-tooling/javaagent-tooling.gradle
anuraaga: Sorry ended up merging another PR into this one since I deleted everything here :)

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2074

Adds 3MB (10%) to the agent jar with questionable benefit on latest Java 8.

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentTracerProviderConfigurer.java
trask: ya, not sure how to make it play nice with the autoconfigure env vars

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootWithSamplingSmokeTest.groovy
```diff
@@ -20,7 +20,10 @@ class SpringBootWithSamplingSmokeTest extends SmokeTest {
 
   @Override
   protected Map<String, String> getExtraEnv() {
-    return ["OTEL_CONFIG_SAMPLER_PROBABILITY": String.valueOf(SAMPLER_PROBABILITY)]
+    return [
+      "OTEL_TRACE_SAMPLER": String.valueOf("parentbased_traceidratio"),
```
trask, line 24: 😄
```suggestion
      "OTEL_TRACE_SAMPLER": "parentbased_traceidratio",
```

trask (approved):
This is really nice 👍

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootWithSamplingSmokeTest.groovy
anuraaga: Oops - looked like a groovyism at first

jkwatson:
I'm confused about the idgenerator bit. Can't you configure one via the `SdkTracerProviderBuilder` via the `SdkTracerProviderConfigurer` SPI? It seems like a rare enough thing to have to provide a custom id generator, that a custom SPI implementation seems like it would be an ok approach to me.

anuraaga:
@jkwatson That works after this PR, not before where we aren't using the autoconfigure module :P

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -102,6 +102,7 @@ abstract class SmokeTest extends Specification {
       .withEnv("OTEL_BSP_MAX_EXPORT_BATCH_SIZE", "1")
       .withEnv("OTEL_BSP_SCHEDULE_DELAY_MILLIS", "10")
       .withEnv("OTEL_EXPORTER_OTLP_ENDPOINT", "collector:55680")
+      .withEnv("OTEL_EXPORTER_OTLP_INSECURE", "true")
```
jkwatson, line 105: curious why this is now required. Are we defaulting to requiring TLS somewhere?

anuraaga: Oh shoot, I had meant to flag you on this and forgot, nice eyes. 

Yeah because of my use of a primitive type instead of nullable, it ended up defaulting to TLS with autoconfigure

https://github.com/open-telemetry/opentelemetry-java/blob/v0.14.1/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java#L55

Luckily with 0.15 it'll only use the scheme from the endpoint

https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/SpanExporterConfiguration.java#L50

jkwatson: @breedx-splk and I ran into this today while setting doing some benchmarking work around some possible tweaks to the jdbc instrumentation. Definitely confused us for a while!

anuraaga: @jkwatson Oh - yeah I guess it affects HEAD agent right now, but won't affect the released 0.15.0 since we'll migrate to the scheme by then.

Issue #1079, opened 2020-08-21
SmokeTest should declare agent's shadowJar as input


iNikem:
It seems that currently gradle does not re-run smoke tests if agent's shadow jar has changed. This may lead to missed regressions during subsequent builds of the same PR

PR #2085, opened 2021-01-20
Smoke tests declare shadowJar as an input to be rebuilt

vovencij:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1079

Fix Gradle deprecation warnings.

### DIFF
--- a/smoke-tests/smoke-tests.gradle
+++ b/smoke-tests/smoke-tests.gradle
@@ -27,7 +27,7 @@ dependencies {
 }
 
 test {
-  inputs.files(tasks.findByPath(':javaagent:shadowJar').outputs.files.singleFile)
+  inputs.files(tasks.findByPath(':javaagent:shadowJar').outputs.files)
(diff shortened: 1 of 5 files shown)


smoke-tests/smoke-tests.gradle
```diff
@@ -27,7 +27,7 @@ dependencies {
 }
 
 test {
-  dependsOn ':javaagent:shadowJar'
+  inputs.files(tasks.findByPath(':javaagent:shadowJar').outputs.files.singleFile)
```
iNikem, line 30: Can/should it be `tasks.findByPath(':javaagent:shadowJar').outputs.files`?

iNikem (approved):

smoke-tests/smoke-tests.gradle
vovencij: Seems that it should.

PR #2087, opened 2021-01-20
Rename "normalize" to "sanitize"

breedx-splk:
As suggested in #2065, it makes more sense to call this "sanitization" instead of "normalization", so this is just a rename for clarity.

It doesn't tackle the `normalizeQuery()` method on the DatabaseClientTracer, because that appears to be wholly separate concern (more like "extractQuery()" or something).

### DIFF
--- a/javaagent-api/src/test/groovy/io/opentelemetry/javaagent/instrumentation/api/db/sanitizer/SqlSanitizerTest.groovy
+++ b/javaagent-api/src/test/groovy/io/opentelemetry/javaagent/instrumentation/api/db/sanitizer/SqlSanitizerTest.groovy
@@ -3,17 +3,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.instrumentation.api.db.normalizer
+package io.opentelemetry.javaagent.instrumentation.api.db.sanitizer
 
 import spock.lang.Specification
 import spock.lang.Timeout
 
 @Timeout(20)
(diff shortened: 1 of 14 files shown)


javaagent-api/src/test/groovy/io/opentelemetry/javaagent/instrumentation/api/db/sanitizer/SqlSanitizerTest.groovy
```diff
@@ -116,9 +116,9 @@ class SqlNormalizerTest extends Specification {
     for (int i = 0; i < 10000; i++) {
       s.append("SELECT * FROM TABLE WHERE FIELD = 1234 AND ")
     }
-    String normalized = SqlNormalizer.normalize(s.toString())
+    String normalized = SqlSanitizer.sanitize(s.toString())
```
anuraaga, line 119: sanitized

anuraaga (approved):
Thanks!

trask (approved):
👍

mateuszrzeszutek (approved):

javaagent-api/src/test/groovy/io/opentelemetry/javaagent/instrumentation/api/db/sanitizer/SqlSanitizerTest.groovy
breedx-splk: solid.

PR #2090, opened 2021-01-21
WIP: First draft of proposed async/binding aspect APIs

HaloFour:
This is an early draft of a proposal to extend the `@WithSpan` annotation aspects to support the following features ported from a Comcast flavor of this aspect used in a number of Spring WebFlux applications:

1. Supporting asynchronous return values
2. Map method parameters to span attributes
3. Treat specific thrown exceptions as not errors

I've tried to dump most of what we have into the repo as-is for the sake of discussion.  It does not build due to an assortment of missing dependencies and bits of this would certainly have to be broken up and refactored for a number of different reasons.  I also see a number of places I would like to refactor the API for the sake of simplifying it and unifying the behavior behind specific interfaces to keep much of the logic out of the different specific AOP implementations.


trask:
@HaloFour ok to close this?

HaloFour:
Closed, breaking up into smaller targeted PRs.

PR #2091, opened 2021-01-21
JMS instrumentation: do not create spans if no messages are received

### LINKED ISSUE #1989: JMS instrumentation creates spans when no messages are received
It is common for users and frameworks to poll JMS for messages.  Depending on the system, there could be long periods where no messages are received.  Spring `@JmsListener` implementations, for example, poll JMS every second...and if no messages have been sent, the instrumentation still creates jms receive CONSUMER spans every second.  Any invocation of  `MessageConsumer.receive(timeout)` that times out without a message is also in this boat.

This potentially creates a large number of undesirable spans that the user will have to sift through, at worst, and possibly even pay for.  It is expected that spans should not be created if no messages are actually read from JMS.

The simplest way to reproduce this is to wire up a spring jms application that uses `@JmsListener`.  Spring will create a `DefaultMessageListenerContainer.AsyncMessageListenerInvoker` that will poll every second, ultimately ending up in `JmsDestinationAccessor` which calls `MessageConsumer.receive(timeout)` which is [ultimately handled by the JmsMessageConsumerInstrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java#L43).

The solution might be as simple as returning early from `ConsumerAdvice.stopSpan()` before the span is created.

malafeev:
fix #1989


instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -57,18 +55,13 @@ public static long onEnter() {
...
       if (message == null) {
-        destination =
-            InstrumentationContext.get(MessageConsumer.class, MessageDestination.class)
-                .get(consumer);
-        if (destination == null) {
-          destination = MessageDestination.UNKNOWN;
-        }
+        // Do not create span when no message is received
+        return;
       } else {
```
breedx-splk, line 65: Not a huge deal, but my preference is to remove the redundant `else` when there is a return above it.

breedx-splk (approved):
This looks good to me.  Thanks for picking this up!  🎉

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java
malafeev: fixed

anuraaga (approved):

trask (approved):

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -57,21 +55,15 @@ public static long onEnter() {
 
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void stopSpan(
-        @Advice.This MessageConsumer consumer,
         @Advice.Enter long startTime,
         @Advice.Return Message message,
         @Advice.Thrown Throwable throwable) {
       MessageDestination destination;
       if (message == null) {
-        destination =
```
iNikem, line 66: `JmsSessionInstrumentation` can be removed now, right?

iNikem (changes requested):

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java
malafeev: removed

iNikem (approved):

malafeev:
looks like ready to merge.
can anybody with permissions merge it?

PR #2093, opened 2021-01-21
AWS SDK - adding SQS propagation

### LINKED ISSUE #2092: Implement SQS context propagation for AWS SDK 1.11 instrumentation
Current AWS SDK 1.11 instrumentation does not offer a context propagation for SQS, which requires the values to be transported using message attributes.

Proper context propagation between producer and consumer when using SQS via AWS SDK 1.11.

NA

**Additional context**
NA

kubawach:
Closes #2092 

Scope of this PR:
- implementation of SQS context propagation for AWS SDK 1.11
- refactoring of the AWS SDK 1.11 module - split into shared library, core and SQS (thanks @mateuszrzeszutek for he suggestion)
- integration test for AWS SDK 1.11 SQS propagation

Technical details:
- SQS - currently only the main use case - send / receiver message supported
- SQS module depends on the core module - therefore also Tracing Handler depends in runtime on the core to get/save metadata in the cache (!!) - avoids code duplication

Open questions:
1. AWS SDK 1.11 split - does proposed (library, core - javaagent, SQS - javaagent) makes sense?
2. Spans created for SQS communication. I left core module intact, so currently for each SQS message, there will be:
- core - http client span
- sqs - queue (producer / consumer)
Therefore on outbound path we see CLIENT span (http with http and aws attributes) followed by a PRODUCER span (with aws attributes only). This is:
- clearly against the spec (CLIENT span followed by PRODUCER)
- causing issues - as two instrumentations interleave (first core, than sqs on send and receive) closing of the SQS span reinstates parent (CLIENT - core span) as the current one (!)
The question here is what are the expectations? A couple of solutions are possible, for example:
- checking for CLIENT span before creating PRODUCER one - in this situation PRODUCER won't be created, so we will have a trace like: CLIENT(send) -> CLIENT (receive) - possibly in the same span due to context propagation done for SQS; this solution however will not clearly "show" to the customers that this is in fact a messaging communication (PRODUCER / CONSUMER)
- preventing SDK core spans for SQS messages - in this situation we would get PRODUCER (send) -> CONSUMER (receive) spans, but HTTP attributes would be missing (not part of SQS spans)
- preventing coupling (parent / child) between SDK core and SQS spans - in this situation we would get three traces - CLIENT, PRODUCER -> CONSUMER, CLIENT

Calling experts to shed a light on ^^ matters - @anuraaga @trask @iNikem @mateuszrzeszutek :)


jkwatson:
I think I would have expected the span parenting to be PRODUCER -> CLIENT, rather than the other way around, since the http communication is supposed to be the CLIENT, whereas the PRODUCER is the logical queue production span, which happens above the http layer.

anuraaga:
@jkwatson's proposed ordering sounds right in general, but I guess it breaks parenting because we would expect a consumer to be linked to the producer, not to the client span. So we need to propagate the producer span, not the client span, which is what our SDK instrumentation does.

I think this may be fundamentally a problem with the AWS API's design to propagate the header on the SDK request itself into SQS, when really the API is an implementation detail wrapping another concept - a better modeling could have been to have the producer span context in a separate header. client connects to server, producer connects to consumer. Due to backwards compatibility issues, I suspect this won't ever change though.

That being said, it does make me wonder of the intent of this PR - is anything needed for SQS propagation in general? I figured v1 would also have the v2 behavior, meaning that a client span should be getting propagated already. Is that true? If so, I guess this is more about instrumenting a producer span, not as much about context propagation. I would probably special case SQS to create producer spans with both the http and faas attributes instead of client spans, replacing them completely - the client span doesn't really provide much since there's no corresponding server span. If SQS was to export a server span corresponding to the processing of a message before dispatching it to a receiver, then things get tricky, but perhaps such a situation could be handled if it ever came up?

If just using a single span, then I don't know if we'd need a completely separate instrumentation. I'm imagining something small like `if (service == SQS) span.kind(PRODUCER); fillFaasAttributes(span);`

kubawach:
@jkwatson @anuraaga thanks for your comments, I appreciate them a lot

With regards to the fundamental question "why" there were two reasons:
1. Get "proper" messaging spans - PRODUCER / CONSUMER (that would be easily done with and if as @anuraaga already pointed out)
2. Get full context propagation between producing and consuming sides of the SQS interaction. As far as I can tell, with HTTP headers only (as in the original code), context is not propagated from "Send" message to "Receive" one. In other words consumer (if not running within same JVM) does not know about the producer and can set parent properly. In order to fix this, parent context needs to be propagated using the message attributes mechanism. This is also important in scenarios involving eg Apache Camel consuming messages from SQS (as the Camel instrumentation can also understand context from the attributes - appropriate PR is in progress)

LMK what you think, very happy to discuss this with more knowledgable folks :)

anuraaga:
Ah the missing piece might be that http header is actually propagated. It just has to be the aws x-ray format.

https://docs.aws.amazon.com/xray/latest/devguide/xray-services-sqs.html

Technically this doc title and OTel designation for the format are both wrong - it's aws trace header and nothing specific to x-ray. I didn't bother correcting the terminology in OTel because it's effectively outdated now as aws should migrate to w3c - but that will take years.

So yeah, I realize the main difference between v2 and v1 is probably that in the v2 SDK I forced the propagator to be the AwsXrayPropagator to get as much propagation as possible. It would be good to do this for v1 too. If the SDK always sends the aws header, it should make it into that system attribute. Remember, that's what we are using in the lambda instrumentation as well.

kubawach:
@anuraaga ahhh yes, that's exactly what I missed and went with "regular" propagation headers propagation. Thanks! 
So the plan now would be to:
- enforce XRayPropagator for SDK v1 (just as it's done for v2)
- add if-s for setting PRODUCER / CONSUMER spans for SQS (leaving both HTTP and AWS attributes)
- remove extra instrumentation split and get back to a single one (without dependency on AWS SQS library)
- correct tests, especially for Camel - as discussed with @mateuszrzeszutek in case of Camel SQS trace should be: INTERNAL (Camel) -> PRODUCER (AWS SDK) -> CONSUMER (AWS SDK) -> INTERNAL (Camel); this way all of the magic propagation stays within AWS SDK (whichever is used)

LMK if ^^ sounds correct :)

anuraaga:
@kubawach SGTM thanks!

iNikem:
@kubawach should this be converted to Draft for now?

PR #2094, opened 2021-01-22
Cache sql sanitized extraction

breedx-splk:
This builds on top of #2087 and will be rebased when that is in...so maybe wait to review?

The discussion in #2065 talks about sql parsing being expensive.  If we cache the sql statement string to `SqlStatementInfo` we can prevent having to do this heavy lifting more than once for each statement.  A guava LRU cache was selected to prevent caching excessive numbers of statements on the heap in the even the user does dynamic/inconsistent sql string creation.

This seems like a pretty reasonable compromise -- it keeps the utility of sanitization and db/operation extraction while preventing duplicate efforts in common situations.

A benchmark based on spring-petclinic-rest (which uses hibernate) shows the following with 5VUs, 500 iterations:
*  No agent : avg = 98.83 p95 = 119.73
*  With agent no cache: avg = 319.32 p95 = 525.50
*  With agent with cache: avg = 129.67 p95 = 174.81
(all times in ms)


jkwatson:
This seems like a pretty nice performance win. 5000 might even be too many to cache. If we're willing to put metrics into our code, it would be rad to register an observer that would report how many items were in the cache.

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
```diff
@@ -69,25 +74,35 @@ public static Connection connectionFromStatement(Statement statement) {
...
+    try {
+      return sqlToStatementInfoCache.get(
+          sql,
+          () -> {
+            log.debug("SQL statement cache miss");
+            try {
+              return SqlStatementInfoExtractor.extract(sanitizeSql(sql));
+            } catch (ParseException e) {
+              log.debug("Could not extract sql info", e);
+              return new SqlStatementInfo(null, null, null);
```
iNikem, line 87: Can we use a single pre-made instance for this?

mateuszrzeszutek: I don't think that's necessary - this is a "should never happen" scenario, as the lexer actually matches all possible input.

iNikem (approved):

mateuszrzeszutek:
Muzzle failed locally:

```
FAILED MUZZLE VALIDATION: io.opentelemetry.javaagent.instrumentation.jdbc.JdbcInstrumentationModule mismatches:
-- io.opentelemetry.javaagent.instrumentation.jdbc.JdbcUtils:31 Missing class com.google.common.cache.CacheBuilder
-- io.opentelemetry.javaagent.instrumentation.jdbc.JdbcUtils:79 Missing class com.google.common.cache.Cache
```

This is the second time I've noticed that it passes in GHA but fails (correctly) when run locally. Do we use some sort of cache for muzzle tasks?

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
```diff
@@ -24,6 +27,8 @@
   private static final boolean NORMALIZATION_ENABLED = isQueryNormalizationEnabled("jdbc");
 
   private static Field c3poField = null;
+  private static final Cache<String, SqlStatementInfo> sqlToStatementInfoCache =
+      CacheBuilder.newBuilder().maximumSize(5000).build();
```
mateuszrzeszutek, line 31: I wonder if we can use our own `WeakCache` instead of guava-cache - right now it's located in the `javaagent-bootstrap` module, but is only used in `javaagent-tooling`. If we moved it to `javaagent-api` we could use it in the instrumentations too.

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
```diff
@@ -69,8 +74,27 @@ public static Connection connectionFromStatement(Statement statement) {
...
+            log.debug("SQL statement cache miss");
+            try {
+              return SqlStatementInfoExtractor.extract(sanitizeSql(sql));
+            } catch (ParseException e) {
+              log.debug("Could not extract sql info", e);
+              return new SqlStatementInfo(null, null, null);
+            }
+          });
+    } catch (ExecutionException e) {
+      log.info("Sql statement cache error", e);
```
mateuszrzeszutek, line 91: Hmm, should we really log it as info? Users will see this exception (if it ever happens...) with the default javaagent logging.

iNikem:
> Muzzle failed locally:
> 
> ```
> FAILED MUZZLE VALIDATION: io.opentelemetry.javaagent.instrumentation.jdbc.JdbcInstrumentationModule mismatches:
> -- io.opentelemetry.javaagent.instrumentation.jdbc.JdbcUtils:31 Missing class com.google.common.cache.CacheBuilder
> -- io.opentelemetry.javaagent.instrumentation.jdbc.JdbcUtils:79 Missing class com.google.common.cache.Cache
> ```
> 
> This is the second time I've noticed that it passes in GHA but fails (correctly) when run locally. Do we use some sort of cache for muzzle tasks?

All GHA runs use gradle cache. Both remote and "local". Seems like muzzle tasks don't play nicely with caches?

mateuszrzeszutek:
> All GHA runs use gradle cache. Both remote and "local". Seems like muzzle tasks don't play nicely with caches?

I've created an issue for that: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2094#issuecomment-765671043

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
breedx-splk: I like this idea. Would it be ok for me to create a follow-up issue to do this work?

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
breedx-splk: I chose `info` because the docs indicate that it is unusual for this to happen, and is usually related to an underlying `InterruptedException`.  I wouldn't think the user would see this much, but if it's happening a lot it would be good to know wouldn't it? 

I don't feel strongly, and I could back it down to debug.

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
mateuszrzeszutek: Sure! 👍 

I think it could work a bit similar to `WeakMap` (interface in `javaagent-api`, implementation in `javaagent-tooling`).

breedx-splk:
Closing to prevent premature merging (until I can get a muzzle fix in).

breedx-splk:
Going to try this as a draft instead of closing.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/GuavaBoundedCache.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.opentelemetry.javaagent.tooling;
+
+import com.google.common.cache.Cache;
+import io.opentelemetry.javaagent.instrumentation.api.BoundedCache;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
+
+public class GuavaBoundedCache<K, V> implements BoundedCache<K, V> {
```
anuraaga, line 13: Can be package private I think

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
```diff
@@ -69,8 +72,27 @@ public static Connection connectionFromStatement(Statement statement) {
     return connection;
   }
 
+  public static SqlStatementInfo sanitizeAndExtractInfo(String sql) {
+    try {
+      return sqlToStatementInfoCache.get(
```
anuraaga, line 77: Is it possible / makes sense to only use the cache with prepared statements? The bounding makes it safe, but I guess for raw queries the cache will just thrash all the time.

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
```diff
@@ -69,8 +72,27 @@ public static Connection connectionFromStatement(Statement statement) {
...
+          sql,
+          k -> {
+            log.debug("SQL statement cache miss");
```
anuraaga, line 80:
```suggestion
            log.trace("SQL statement cache miss");
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentTooling.java
```diff
@@ -33,6 +36,21 @@ static void registerWeakMapProvider() {
...
+  /**
+   * Instances of BoundCache are backed by a guava instance that lives in the agent classloader and
+   * is bridged to user/instrumentation classloader through the BoundedCache.Provider interface.
+   */
+  static void registerBoundedCacheProvider() {
+    BoundedCache.Provider.registerIfAbsent(
+        new BoundedCache.Builder() {
+          @Override
+          public <K, V> BoundedCache<K, V> build(long maxSize) {
+            Cache<K, V> cache = CacheBuilder.newBuilder().maximumSize(5000).build();
```
mateuszrzeszutek, line 48:
```suggestion
            Cache<K, V> cache = CacheBuilder.newBuilder().maximumSize(maxSize).build();
```

javaagent-tooling/src/test/java/io/opentelemetry/javaagent/tooling/GuavaBoundedCacheTest.java
```diff
@@ -0,0 +1,27 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+
+class GuavaBoundedCacheTest {
+
+  void testCache() {
+    fail("OH MY");
```
mateuszrzeszutek, line 17: Hmm, to be removed?

iNikem: Build is green! :D

mateuszrzeszutek: 😱

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
breedx-splk: I went ahead and added it here, since it was going to be difficult to get past the muzzle issue without tackling it here.

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
breedx-splk: There's another code path that uses `JdbcMaps.preparedStatements` for caching prepared statements.   It's a contract between the `JdbcTracer` and the `ConnectionInstrumentation`.  Would prefer to circle back on this later if it becomes a problem...although it might be difficult to trace without internal metrics.  I'll open an issue to track the idea of adding metrics for cache misses.

javaagent-tooling/src/test/java/io/opentelemetry/javaagent/tooling/GuavaBoundedCacheTest.java
breedx-splk: Yeah, those java unit tests in that dir don't get run.  🤷 
I'll convert it to groovy for now.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/GuavaBoundedCache.java
```diff
@@ -0,0 +1,29 @@
...
+    this.delegate = delegate;
+  }
+
+  @Override
+  public V get(K key, Function<? super K, ? extends V> mappingFunction) {
+    try {
+      return delegate.get(key, () -> mappingFunction.apply(key));
+    } catch (ExecutionException e) {
+      throw new BoundedCache.Exception(e);
+    }
```
trask, line 27: and then you can delete BoundedCache.Exception and the associated catch blocks
```suggestion
    try {
      return delegate.get(key, () -> mappingFunction.apply(key));
    } catch (ExecutionException e) {
      // this shouldn't happen because the java.util.function.Function doesn't declare any checked exceptions
      throw new IllegalStateException(e);
    }
```

breedx-splk: This is a good point.  I had only really noticed the contract through its signature and not the javadoc which does clearly say that it only throws if the value lookup throws.  🤘  I'll change it.

anuraaga (approved):

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
```diff
@@ -22,6 +23,8 @@
   private static final boolean NORMALIZATION_ENABLED = isQueryNormalizationEnabled("jdbc");
 
   private static Field c3poField = null;
+  private static final BoundedCache<String, SqlStatementInfo> sqlToStatementInfoCache =
+      BoundedCache.build(5000);
```
trask, line 27: a little concerned about this being pretty large, imagining 5000 hibernate queries could easily top 5MB, thoughts on dropping it to 1000?

trask (approved):

mateuszrzeszutek (approved):

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcUtils.java
breedx-splk: Oh sure.  I think @jkwatson was also thinking this number was too large, so I'll lower it to 1000.  Before I do that, I'll also sarcastically and jokingly say OMG 5 WHOLE MEGABYTES! 😝

PR #2099, opened 2021-01-22
Match classloader when applying netty instrumentation.

anuraaga:
Currently, both netty 4.0 and 4.1 are attempted to be applied on every classload because 4.0 does not fail type matching, only muzzle reference check. I'm not sure if this is a bug or not in our muzzle reference check handling, presumably it should only do so once. Either way, it seems like good practice to explicitly check where we can when targeting multiple versions.

This makes debug logs much more readable since previously every successful 4.1 transformation was also accompanied by a mysterious failure of the 4.0 instrumentation. Presumably some speed boost too.


iNikem (approved):

pavolloffay (approved):

trask:
> This makes debug logs much more readable

💯

PR #2105, opened 2021-01-22
Hide the context keys in the BaseTracer.

jkwatson:
And, provide the appropriate methods to get access.


iNikem (approved):
Niiiice :)

jkwatson:
> Niiiice :)

perhaps, but why so many broken tests???

jkwatson:
> > Niiiice :)
> 
> perhaps, but why so many broken tests???

figured out. somehow the client key got non-static. fixed.

PR #2106, opened 2021-01-22
Consolidate span suppression logic into the BaseTracer

jkwatson:
Depends on #2105 (will rebase after that is merged)

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -80,7 +80,7 @@ public abstract class BaseTracer {
     return Span.current();
   }
 
-  protected final boolean shouldNotSuppressSpan(Kind proposedKind, Context context) {
+  protected final boolean shouldStartSpan(Kind proposedKind, Context context) {
     switch (proposedKind) {
       case CLIENT:
         return !inClientSpan(context);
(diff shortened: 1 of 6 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -83,6 +80,25 @@ public Span getCurrentSpan() {
     return Span.current();
   }
 
+  protected final boolean shouldNotSuppressSpan(Kind proposedKind, Context context) {
```
iNikem, line 83: Can we name this method `shouldStartSpan`? Current name is too close to double negation IMO

iNikem (approved):

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
jkwatson: I don't like the double negative either. But,I also want something that references suppression.  Unless we want everything to always call this method before starting a span. We could reverse the logic if we're ok with all the current callers negating it.

iNikem: Well, in theory as of now all instrumentations should/could make a check before starting a span. Either to prevent nested SERVER/CLIENT spans or to check `CallDepth` for INTERNAL spans. This may mean that "we want everything to always call this method before starting a span". But nested client spans may become allowed...

jkwatson: right. the main reason for this change was to make it so that we could have one place to handle the supression logic, especially if the logic was going to change. Happy to rename to `shouldStartSpan` if folks think that's not too generic.

jkwatson: I also just realized that this method is now unit-testable. I'll do the rename and add some unit tests.

trask (approved):
thx!

iNikem (approved):

anuraaga (approved):
Cool

Issue #1888, opened 2020-12-11
Add support for apache dubbo.


chenmudu:
Maybe can do some enhancements to apache dubbo.  Is this feasible?   thanks very much!

tydhot:
@iNikem assign to me plz

anuraaga:
Thanks @tydhot :) If it's possible to start with library instrumentation so we support both agent and non-agent users, that would be great!

tydhot:
@anuraaga Thank you for your advice :)

mateuszrzeszutek:
Implemented in #2108 - closing.

PR #2108, opened 2021-01-25
add support 4 Apache Dubbo

tydhot:
support apache dubbo version: 2.7.0+
feature #1888


tydhot:
@anuraaga Is this what you said about start with library instrumentation?:)

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/DubboClientTracer.java
```diff
@@ -0,0 +1,41 @@
...
+
+  protected DubboClientTracer() {}
+
+  protected DubboClientTracer(Tracer tracer) {
+    super(tracer);
+  }
+
+  public Span startSpan(String name) {
+    SpanBuilder spanBuilder = tracer.spanBuilder(name).setSpanKind(CLIENT);
+    spanBuilder.setAttribute(SemanticAttributes.RPC_SYSTEM, "dubbo");
```
anuraaga, line 28: Separate from this PR it will be nice to have `abstract String getRpcSystem()` on `RpcClientTracer` to abstract this away.

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/DubboInjectAdapter.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.dubbo.apache.v2_7.client;
+
+import io.opentelemetry.context.propagation.TextMapPropagator;
+import org.apache.dubbo.rpc.RpcInvocation;
+
+public class DubboInjectAdapter implements TextMapPropagator.Setter<RpcInvocation> {
```
anuraaga, line 11:
```suggestion
class DubboInjectAdapter implements TextMapPropagator.Setter<RpcInvocation> {
```

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.dubbo.rpc.Result;
+import org.apache.dubbo.rpc.RpcContext;
+import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.RpcInvocation;
+
+public class TracingClientFilter implements Filter {
+
+  private final DubboClientTracer tracer;
+
+  public TracingClientFilter() {
```
anuraaga, line 28:
```suggestion
  private TracingClientFilter() {
```

Since we have factory method

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    this.tracer = new DubboClientTracer();
+  }
+
+  public static TracingClientFilter newFilter() {
```
anuraaga, line 32: How about a factory that accepts `DubboClientTracer` as a parameter too

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    return new TracingClientFilter();
+  }
+
+  @Override
+  public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
+    if (invocation instanceof RpcInvocation) {
```
anuraaga, line 38: Let's reduce indent

```java
if (!(invocation instanceoc RpcInvocation)) {
  return invoker.invoke(invocation);
}
```

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
```diff
@@ -0,0 +1,59 @@
...
+      String methodName = invocation.getMethodName();
+      Span span = tracer.startSpan(methodName);
+      DubboHelper.prepareSpan(span, methodName, invoker);
+      Context context = Context.current().with(span);
```
anuraaga, line 42: `tracer.withClientSpan(Context.current(), span)`

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
```diff
@@ -0,0 +1,59 @@
...
+          .getTextMapPropagator()
+          .inject(context, (RpcInvocation) invocation, SETTER);
+      Result result;
+      try (Scope ignored = span.makeCurrent()) {
+        result = invoker.invoke(invocation);
+      } catch (Throwable e) {
+        tracer.endExceptionally(span, e);
+        throw e;
+      }
+      tracer.endSpan(span, result);
```
anuraaga, line 54: Does dubbo only support synchronous call?

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
```diff
@@ -0,0 +1,59 @@
...
+
+  public static TracingClientFilter newFilter() {
+    return new TracingClientFilter();
+  }
+
+  @Override
+  public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
+    if (invocation instanceof RpcInvocation) {
+      String methodName = invocation.getMethodName();
+      Span span = tracer.startSpan(methodName);
```
anuraaga, line 40: Span name must be the full name

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#span-name

So I guess it is `invoker.getInterface().getName() + '/' + invocation.getMethodName()`

anuraaga: I guess we can add a utility in the helper for it for use in server side too

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/common/DubboHelper.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.dubbo.rpc.Invoker;
+import org.apache.dubbo.rpc.Result;
+
+public final class DubboHelper {
+
+  private DubboHelper() {}
+
+  public static void prepareSpan(Span span, String methodName, Invoker<?> invoker) {
+    span.setAttribute(
+        SemanticAttributes.RPC_SERVICE, invoker.getInterface().getSimpleName() + ":" + methodName);
```
anuraaga, line 20: `RPC_SERVICE` is only the class name, so I guess `invoker.getInterface().getName()`

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#attributes

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/common/DubboHelper.java
```diff
@@ -0,0 +1,29 @@
...
+    if (methodName != null) {
```
anuraaga, line 21: Can methodname be null?

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/server/TracingServerFilter.java
```diff
@@ -0,0 +1,53 @@
...
+import io.opentelemetry.instrumentation.dubbo.apache.v2_7.common.DubboHelper;
+import org.apache.dubbo.rpc.Filter;
+import org.apache.dubbo.rpc.Invocation;
+import org.apache.dubbo.rpc.Invoker;
+import org.apache.dubbo.rpc.Result;
+import org.apache.dubbo.rpc.RpcContext;
+import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.RpcInvocation;
+
+public class TracingServerFilter implements Filter {
```
anuraaga, line 21: Similar comments as client/side

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/dubbo/apache/v2_7/DubboFilterInstrumentation.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Override
+  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+    return singletonMap(
+        named("getActivateExtension")
+            .and(takesArgument(0, named("org.apache.dubbo.common.URL")))
+            .and(takesArgument(1, String.class))
+            .and(takesArgument(2, String.class))
+            .and(isPublic()),
+        DubboFilterInstrumentation.class.getName() + "$AddFilterAdvice");
```
anuraaga, line 45: Does it work to use resource injection instead?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java#L33

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInitializationInstrumentation.java#L23

anuraaga (commented):
Thanks a lot for the instrumentation!

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
tydhot: Dubbo automatically loads filters based on their name by SPI. If this is set to private, the library will not work.Maybe I should remove the factory method?

anuraaga: Ah ok - yeah let's remove the factory then.

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/dubbo/apache/v2_7/DubboFilterInstrumentation.java
tydhot: In dubbo, I found that importing resources directly does not work.It should be declared or configured in advance.Maybe code here it's easier to understand now。

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/TracingClientFilter.java
tydhot: I have read Apache SkyWalking and pinpoint's dubbo instrumentation, it seemed that none of them has extra processing for asynchronous calls.And dubbo haven't had streaming call yet.

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/DubboClientTracer.java
tydhot: Maybe it would be more appropriate to do it in another pr.

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/dubbo/apache/v2_7/DubboFilterInstrumentation.java
anuraaga: Can you describe more what the issue is? This injection happens on classload, so it's as early as it needs to be I think. For the type matcher in `DubboInitializationInstrumentation`, I think you can just use `ExtensionLoader`.

Looking at this version of the instrumentation, I see we instrument one of these similarly named methods

https://github.com/apache/dubbo/blob/master/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java#L246

It's unclear whether this is complete, or if it will need maintenance in the future. We can be more confident if we use the official mechanism though, which I guess is the resource file?

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/client/DubboClientTracer.java
```diff
@@ -0,0 +1,49 @@
...
+  public Span startSpan(String interfaceName, String methodName) {
+    SpanBuilder spanBuilder =
+        tracer.spanBuilder(DubboHelper.getSpanName(interfaceName, methodName)).setSpanKind(CLIENT);
+    spanBuilder.setAttribute(SemanticAttributes.RPC_SYSTEM, "dubbo");
+    Span span = spanBuilder.startSpan();
+    DubboHelper.prepareSpan(span, methodName, interfaceName);
+    return span;
+  }
+
+  public Context withClient(Span span) {
```
anuraaga, line 36: I don't think we need this trivial helper, and the name is confusing I guess

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/common/DubboHelper.java
```diff
@@ -0,0 +1,29 @@
...
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.apache.dubbo.rpc.Result;
+
+public final class DubboHelper {
+
+  private DubboHelper() {}
+
+  public static void prepareSpan(Span span, String methodName, String interfaceName) {
```
anuraaga, line 17: Order it `interfaceName`, then `methodName` which is more common for class+method

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/test/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTest.groovy
```diff
@@ -0,0 +1,35 @@
...
+import io.opentelemetry.instrumentation.dubbo.apache.v2_7.api.HelloService
+import io.opentelemetry.instrumentation.dubbo.apache.v2_7.impl.HelloServiceImpl
+import io.opentelemetry.instrumentation.test.AgentTestTrait
+import org.apache.dubbo.config.ReferenceConfig
+import org.apache.dubbo.config.RegistryConfig
+import org.apache.dubbo.config.ServiceConfig
+
+class DubboTest extends AbstractDubboTest implements AgentTestTrait {
+  @Override
+  ServiceConfig configureServer() {
```
anuraaga, line 17: We generally have the tests differ only in the way they are instrumenting. Is it possible to have the registry / service definition happen in the base class, only difference is non-agent version calls addFilter if it's required?

instrumentation/dubbo/dubbo-apache-2.7/library/src/test/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTest.groovy
```diff
@@ -0,0 +1,37 @@
...
+class DubboTest extends AbstractDubboTest implements InstrumentationTestTrait {
+  @Override
+  ServiceConfig configureServer() {
+    def registerConfig = new RegistryConfig()
+    registerConfig.setAddress("N/A")
+    ServiceConfig<HelloServiceImpl> service = new ServiceConfig<>()
+    service.setInterface(HelloService)
+    service.setRef(new HelloServiceImpl())
+    service.setRegistry(registerConfig)
+    service.setFilter("OtelServerFilter")
```
anuraaga, line 24: Is setFilter required even though we have the resource files?

instrumentation/dubbo/dubbo-apache-2.7/testing/src/main/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/AbstractDubboTest.groovy
```diff
@@ -0,0 +1,139 @@
...
+
+    then:
+    response == "hello"
+    responseAsync.get() == "hello"
+
+    assertTraces(2) {
+      trace(0, 3) {
+        basicSpan(it, 0, "parent")
+        span(1) {
+          name "org.apache.dubbo.rpc.service.GenericService/\$invoke"
```
anuraaga, line 76: I would generally expect the client service / method to match the server one. In dubbo, is it not possible for the client to have the name `HelloService/hello` instead of this generic name? I see `setInterface` and the method name `hello` in the invocation, so it seems like the client has all the information.

tydhot: Dubbo will use javaassist to set up a proxy for the interface to call directly instead of generic invoke, but I find that the proxy cannot work in groovy's unit test.So I use generic invoke here instead.

anuraaga: I see - we'll need to add a smoke-test to verify the behavior in a future PR then. Can you add a comment here?

tydhot: I'd love to do it

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/test/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTest.groovy
tydhot: yes,I will fix it.

instrumentation/dubbo/dubbo-apache-2.7/library/src/test/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTest.groovy
tydhot: I found that if I add the Activate annotation at the beginning of a class it will be loaded automatically, and it is worked well in library.But maybe Dubbo's spi is more complicated than I thought, if I inject the resource file directly, it will be something wrong in javaagent's unit test.It will be the exception like failOnContextLeak I've spent a lot of time checking, but I don't have a clue.It seems that the process of Dubbo loading the server filter is different from that of the client filter.I have changed the code in javaagent in order to make it work by Activate annotation, and it can avoid the exception failOnContextLeak now in javaagent's unit test.

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/test/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTest.groovy
tydhot: I've made the unit test code cleaner

tydhot:
Hi @anuraaga !The library module I had refer to the implementation of Zipkin and rewrite it.Now the code is cleaner and additional support for asynchronous future is added.In library, the filter will be Autoload.
 And I tried to inject the resource file in javaagent,the test case will be an exception like context leak.So I decided to remove the filter's autoload and add it when method back instead in javaagent module.It seemed work and the code is cleaner than before. Plz review it:)!

tydhot:
@anuraaga I have changed the way and now the javaagent module is liked aws's javaagent module.I think It is more confident to do like it so.

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTracer.java
```diff
@@ -0,0 +1,73 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapPropagator.Getter;
+import io.opentelemetry.instrumentation.api.tracer.RpcServerTracer;
+import io.opentelemetry.instrumentation.api.tracer.utils.NetPeerUtils;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.apache.dubbo.rpc.Result;
+import org.apache.dubbo.rpc.RpcContext;
+import org.apache.dubbo.rpc.RpcInvocation;
+
+class DubboTracer extends RpcServerTracer<RpcInvocation> {
```
anuraaga, line 24: A bit weird but it's ok - I filed #2132

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/TracingFilter.java
```diff
@@ -0,0 +1,75 @@
...
+  @Override
+  public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
+    if (!(invocation instanceof RpcInvocation)) {
+      return invoker.invoke(invocation);
+    }
+    String methodName = invocation.getMethodName();
+    String interfaceName = invoker.getInterface().getName();
+    RpcContext rpcContext = RpcContext.getContext();
+    Kind kind = rpcContext.isProviderSide() ? SERVER : CLIENT;
+    Context context;
```
anuraaga, line 44:
```suggestion
    final Context context;
```

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/TracingFilter.java
```diff
@@ -0,0 +1,75 @@
...
+    if (kind.equals(CLIENT)) {
+      context = tracer.startClientSpan(interfaceName, methodName);
+      GlobalOpenTelemetry.getPropagators()
+          .getTextMapPropagator()
+          .inject(context, (RpcInvocation) invocation, SETTER);
+    } else {
+      context = tracer.startServerSpan(interfaceName, methodName, (RpcInvocation) invocation);
+    }
+    Span span = Span.fromContext(context);
+    Result result;
```
anuraaga, line 54:
```suggestion
    final Result result;
```

instrumentation/dubbo/dubbo-apache-2.7/testing/src/main/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/AbstractDubboTest.groovy
anuraaga: Ah sorry I meant a comment in the code :P `// Dubbo's codegen does not work properly from Groovy tests, so we need to verify in a smoke test instead.`

instrumentation/dubbo/dubbo-apache-2.7/testing/src/main/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/AbstractDubboTest.groovy
```diff
@@ -0,0 +1,155 @@
...
+
+    when:
+    GenericService genericService = ReferenceConfigCache.getCache().get(reference) as GenericService
+    def o = new Object[1]
+    o[0] = "hello"
+    def response = runUnderTrace("parent") {
+      genericService.$invoke("hello", [String.getName()] as String[], o)
+    }
+
+    and:
```
anuraaga, line 79: I'm still not a spock expert but I think the general pattern would be

```
when:
then:
and:
when:
then:
```

Or it may make sense to separate sync and async. Currently the assertions being in one big block make it hard to understand.

anuraaga (commented):
Thanks a lot, just some small comments

instrumentation/dubbo/dubbo-apache-2.7/testing/src/main/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/AbstractDubboTest.groovy
tydhot: Yes, I will separate them.

anuraaga:
@tydhot Sorry for the delay in checking this out - I found a problem with our tooling with a fix in #2179. 

Now there is a different issue where the agent doesn't inject the references from the dubbo file since it's not a standard formatted SPI file. I think we can tweak our mechanism for resolving from SPI so instrumentation can customize the parsing technique. We basically need a way for 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2108/files#diff-92953523ffff3580ebb3ff4d0a1b1276d3aacfa375416f2a36e070117022bacfR1 to be parsed by 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollector.java#L60

@mateuszrzeszutek what do you think of this?

instrumentation/dubbo/dubbo-apache-2.7/javaagent/dubbo-apache-2.7-javaagent.gradle
```diff
@@ -0,0 +1,17 @@
...
+    group = "org.apache.dubbo"
+    module = "dubbo"
+    versions = "[2.7.0,)"
+  }
+}
+
+dependencies {
+  implementation project(':instrumentation:dubbo:dubbo-apache-2.7:library')
+
+  library group: 'org.apache.dubbo', name: 'dubbo', version: '2.7.0'
```
anuraaga, line 14: We target tests to oldest version supported

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/dubbo/apache/v2_7/DubboInstrumentationModule.java
```diff
@@ -0,0 +1,39 @@
...
+@AutoService(InstrumentationModule.class)
+public class DubboInstrumentationModule extends InstrumentationModule {
+  public DubboInstrumentationModule() {
+    super("dubbo", "dubbo-apache-2.7");
+  }
+
+  @Override
+  public String[] helperResourceNames() {
+    return new String[] {
+      "META-INF/services/org.apache.dubbo.rpc.Filter",
```
anuraaga, line 26: I found `/services` is handled by dubbo too, so this works with our muzzle

instrumentation/dubbo/dubbo-apache-2.7/javaagent/src/main/resources/META-INF/services/org.apache.dubbo.rpc.Filter
```diff
@@ -0,0 +1 @@
+io.opentelemetry.javaagent.shaded.instrumentation.dubbo.apache.v2_7.OpenTelemetryFilter
```
anuraaga, line 1: I found it automatically transfers the classname to a name if not provided, which works well for us and ensures working with Muzzle

https://github.com/apache/dubbo/blob/master/dubbo-common/src/main/java/org/apache/dubbo/common/extension/ExtensionLoader.java#L1011

testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/OtlpInMemoryMetricExporter.java
```diff
@@ -69,9 +69,6 @@ void reset() {
 
   @Override
   public CompletableResultCode export(Collection<MetricData> metrics) {
-    for (MetricData metric : metrics) {
```
anuraaga, line 72: Really spammy - let's restore in some form if metrics gets more stable

anuraaga:
Did some debugging and got something that seems to work, almost there

instrumentation/dubbo/dubbo-apache-2.7/javaagent/dubbo-apache-2.7-javaagent.gradle
```diff
@@ -0,0 +1,17 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
```
anuraaga, line 1: Can we name the folders either `apache-dubbo/apache-dubbo-2.7` for consistency with other apache projects, or just `dubbo/dubbo-2.7`? I guess dubbo is a unique enough name that apache may not be needed but whatever seems more familiar to users I guess

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboHelper.java
```diff
@@ -0,0 +1,29 @@
...
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.apache.dubbo.rpc.Result;
+
+class DubboHelper {
+
+  private DubboHelper() {}
+
+  public static void prepareSpan(Span span, String interfaceName, String methodName) {
```
anuraaga, line 17: Methods can be package private

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/DubboTracer.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public Context startServerSpan(
+      String interfaceName, String methodName, RpcInvocation rpcInvocation) {
+    SpanBuilder spanBuilder =
+        tracer
+            .spanBuilder(DubboHelper.getSpanName(interfaceName, methodName))
+            .setSpanKind(SERVER)
+            .setParent(
+                GlobalOpenTelemetry.getPropagators()
```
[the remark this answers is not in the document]
anuraaga: How about trying this again, it should work now. The issue previously was we applied the filter twice, once found in library instrumentation (hence needing #2179) and once in agent instrumentation.

tydhot:
@anuraaga Yes, I founded that the classname in resource's name had to be  renamed by shaded just before. It took me a lot of time to debug it...

instrumentation/dubbo/dubbo-apache-2.7/library/src/main/java/io/opentelemetry/instrumentation/dubbo/apache/v2_7/OpenTelemetryFilter.java
```diff
@@ -0,0 +1,75 @@
...
+    String methodName = invocation.getMethodName();
+    String interfaceName = invoker.getInterface().getName();
+    RpcContext rpcContext = RpcContext.getContext();
+    Kind kind = rpcContext.isProviderSide() ? SERVER : CLIENT;
+    final Context context;
+    if (kind.equals(CLIENT)) {
+      context = tracer.startClientSpan(interfaceName, methodName);
+      GlobalOpenTelemetry.getPropagators()
+          .getTextMapPropagator()
+          .inject(context, (RpcInvocation) invocation, SETTER);
```
anuraaga, line 49: Not for this PR, nor necessarily you, but `inject` needs to be moved from `HttpClientTracer` to `BaseTracer`

instrumentation/dubbo/dubbo-apache-2.7/testing/src/main/groovy/io/opentelemetry/instrumentation/dubbo/apache/v2_7/AbstractDubboTest.groovy
anuraaga: Can you go ahead and add this code comment suggestion?

anuraaga (commented):
Few hopefully easy comments left, thanks for the patience

instrumentation/dubbo/dubbo-apache-2.7/javaagent/dubbo-apache-2.7-javaagent.gradle
tydhot: In China, Alibaba's Dubbo still has a certain number of users

anuraaga: Ah got it - then yeah let's use `apache-dubbo/apache-dubbo-2.7`

tydhot: Maybe it would be more cautious to separate Alibaba Dubbo and Apache Dubbo

instrumentation/dubbo/dubbo-apache-2.7/javaagent/dubbo-apache-2.7-javaagent.gradle
tydhot: Before Dubbo 2.7.0, it is called alibaba Dubbo.And it is called Apache Dubbo since 2.7.0.

instrumentation/dubbo/dubbo-apache-2.7/javaagent/dubbo-apache-2.7-javaagent.gradle
tydhot: Ok I got it

anuraaga (approved):
Thanks?

tydhot:
@anuraaga hah, it was a challenging and interesting development experience.

anuraaga:
@tydhot So sorry my comment said "Thanks?" that was a typo for "Thanks!". On phone's Japanese keyboard, `?` is too close to `!` :(

tydhot:
@anuraaga ah, I got it immediately. BTW, I have translated doc in this project 70 percent , it help me a lot :)!  Espaclly the doc about class loader .       https://github.com/open-telemetry/docs-cn/pull/49

instrumentation/apache-dubbo/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apache/dubbo/v2_7/DubboTracer.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public void endSpan(Span span, Result result) {
+    span.setStatus(DubboHelper.statusFromResult(result));
+    end(span);
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.javaagent.dubbo";
```
trask, line 62:
```suggestion
    return "io.opentelemetry.javaagent.apache-dubbo";
```

trask (approved):
great work @tydhot!

instrumentation/apache-dubbo/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apache/dubbo/v2_7/DubboTracer.java
tydhot: ok, I got it :)

instrumentation/apache-dubbo/apache-dubbo-2.7/javaagent/src/main/resources/META-INF/services/org.apache.dubbo.rpc.Filter
```diff
@@ -0,0 +1 @@
+io.opentelemetry.javaagent.shaded.instrumentation.apache.dubbo.v2_7.OpenTelemetryFilter
```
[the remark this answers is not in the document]
tydhot: In should be lived in library in order to make it can be autoload when user only use the library.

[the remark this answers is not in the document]
trask: oh, yes, I missed that it's in both places. you should be able to delete the one in the javaagent instrumentation, and it will use (and shade appropriately) the one from the library instrumentation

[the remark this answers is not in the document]
tydhot: oh, it works! This is really beyond my expectation. I think I should  read the source code of this set carefully during the Chinese Spring Festival holiday.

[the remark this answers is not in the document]
tydhot: It seemed that Task :instrumentation:aws-lambda-1.0:library:compileJava failed.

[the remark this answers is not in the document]
trask: Oh no looks like main branch is busted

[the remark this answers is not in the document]
trask: sorry about that! main is fixed now by #2213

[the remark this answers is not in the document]
tydhot: ah, it is ok now :)

trask:
thanks again @tydhot!

tydhot:
@trask Well, I hope this is a good start for the Chinese community :)

PR #2111, opened 2021-01-25
Add instrumentation for spring-ws

laurit:
Create internal span for methods annotated with spring-ws annotations PayloadRoot, SoapAction or Action.


instrumentation/spring/spring-ws-2.0/javaagent/spring-ws-2.0-javaagent.gradle
```diff
@@ -0,0 +1,43 @@
...
+  testImplementation group: 'wsdl4j', name: 'wsdl4j', version: '1.6.3'
+  testImplementation group: 'com.sun.xml.messaging.saaj', name: 'saaj-impl', version: '1.5.2'
+  testImplementation group: 'javax.xml.bind', name: 'jaxb-api', version: '2.2.11'
+  testImplementation group: 'com.sun.xml.bind', name: 'jaxb-core', version: '2.2.11'
+  testImplementation group: 'com.sun.xml.bind', name: 'jaxb-impl', version: '2.2.11'
+
+  testInstrumentation project(':instrumentation:servlet:servlet-3.0:javaagent')
+
+  latestDepTestLibrary group: 'org.springframework.boot', name: 'spring-boot-starter-web-services', version: '+'
+  latestDepTestLibrary group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '+'
```
iNikem, line 42: @anuraaga do we need to specify this if we already have `testLibrary`?

instrumentation/spring/spring-ws-2.0/javaagent/src/test/groovy/test/boot/SpringWsTest.groovy
```diff
@@ -0,0 +1,159 @@
...
+import org.springframework.oxm.jaxb.Jaxb2Marshaller
+import org.springframework.util.ClassUtils
+import org.springframework.ws.client.core.WebServiceMessageCallback
+import org.springframework.ws.client.core.WebServiceTemplate
+import org.springframework.ws.soap.addressing.client.ActionCallback
+import org.springframework.ws.soap.client.SoapFaultClientException
+import org.springframework.ws.soap.client.core.SoapActionCallback
+import spock.lang.Shared
+
+class SpringWsTest extends AgentTestRunner implements HttpServerTestTrait<ConfigurableApplicationContext> {
```
iNikem, line 29: Have you thought about extending `HttpServerTest`? Some of our frameworks instrumentations use it for tests

iNikem (approved):
Please update the list of supported libraries in README as well

instrumentation/spring/spring-ws-2.0/javaagent/src/test/groovy/test/boot/SpringWsTest.groovy
laurit: Imo `HttpServerTest` is not suitable for testing a soap webservice. To use `HttpServerTest` I would probably need to disable most of the test and hard code soap request content instead of using `WebServiceTemplate`.

instrumentation/spring/spring-ws-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/ws/SpringWsTracer.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public static SpringWsTracer tracer() {
+    return TRACER;
+  }
+
+  public Span startSpan(Method method) {
+    Span springWsSpan = super.startSpan(method);
+    springWsSpan.setAttribute(
+        SemanticAttributes.CODE_NAMESPACE, method.getDeclaringClass().getName());
+    springWsSpan.setAttribute(SemanticAttributes.CODE_FUNCTION, method.getName());
```
trask, line 25: this feels like a "route", should we also update the SERVER span name?

instrumentation/spring/spring-ws-2.0/javaagent/spring-ws-2.0-javaagent.gradle
laurit: removed these

instrumentation/spring/spring-ws-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/ws/AnnotatedMethodInstrumentation.java
```diff
@@ -0,0 +1,79 @@
...
+      new String[] {
+        "org.springframework.ws.server.endpoint.annotation.PayloadRoot",
+        "org.springframework.ws.soap.server.endpoint.annotation.SoapAction",
+        "org.springframework.ws.soap.addressing.server.annotation.Action"
+      };
+
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return declaresMethod(isAnnotatedWith(namedOneOf(ANNOTATION_CLASSES)));
+  }
```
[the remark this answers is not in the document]
laurit: `classLoaderOptimization ` added, latestDepTestLibrary removed

instrumentation/spring/spring-ws-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/ws/SpringWsTracer.java
laurit: It seemed a bit weird to have server span with the exact same name as a nested internal span so I wasn't sure whether it would be a good idea. If you believe server span should also be renamed then I'll do it.

anuraaga (approved):

instrumentation/spring/spring-ws-2.0/javaagent/src/test/groovy/test/boot/SpringWsTest.groovy
```diff
@@ -0,0 +1,159 @@
...
+
+  HelloResponse makeRequest(methodName, name) {
+    WebServiceTemplate webServiceTemplate = new WebServiceTemplate(marshaller)
+
+    Object request = null
+    WebServiceMessageCallback callback = null
+    if ("hello" == methodName) {
+      HelloRequest tmp = new HelloRequest()
+      tmp.setName(name)
+      request = tmp
```
mateuszrzeszutek, line 69: I think you can use groovy constructor syntax and avoid tmp variable:

```suggestion
      request = new HelloRequest(name: name)
```

instrumentation/spring/spring-ws-2.0/javaagent/src/test/groovy/test/boot/SpringWsTest.groovy
```diff
@@ -0,0 +1,159 @@
...
+        serverSpan(it, 0, getContextPath() + "/ws", null, expectedException)
+        handlerSpan(it, 1, methodName, span(0), expectedException)
+      }
+    }
+
+    where:
+    methodName << ["hello", "helloSoapAction", "helloWsAction"]
+  }
+
+  static serverSpan(TraceAssert trace, int index, String operation, Object parentSpan = null, Throwable exception = null) {
```
mateuszrzeszutek, line 127: `parentSpan` is always null in this test, you can remove it

mateuszrzeszutek (approved):

instrumentation/spring/spring-ws-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/ws/SpringWsTracer.java
trask: good point 👍

trask (approved):

iNikem:
@laurit Can you rebase it now to fix failing tests?

PR #2112, opened 2021-01-25
Add documentation for available instrumentation customisation options

### LINKED ISSUE #1623: Analysis of instrumentation customisation
Related to #566, #1622 and #778

I am going to summarise my current understanding of various options for overriding any given instrumentation in custom vendor distributions. The goal is not to arrive at an ideal solution, but at a solution good enough for GA. I focus on the following motivation for such customisation:

> Instrumentation X from Otel distribution creates span that I don't like and I want to change it in my vendor distro.

To simplify the following discussion, let me take as an example some database client instrumentation that creates a span for database call and extracts data from db connection to provide attributes for that span.

## I don't want this span at all

The easiest case. You can just pre-configure your distribution and disable given instrumentation.

## I want to add/modify some attributes and their values does NOT depend on a specific db connection instance.

E.g. you want to add some data from call stack as span attribute. In this case just provide your custom SpanProcessor. No need for touching instrumentation itself.

## I want to add/modify some attributes and their values depend on a specific db connection instance.

Write a _new_ instrumentation, which injects its own advice into the same method as the original instrumentation. Use `getOrder` method to ensure it is run after the original instrumentation. Now you can augment current span with new information.

## I want to remove some attributes

Write custom exporter or use attribute filtering functionality in Collector

## I don't like Otel span at all. I want to significantly modify it and its lifecycle

Disable existing instrumentation. Write a new one, which injects Advice into the same (or better) method as the original instrumentation. Write your own Advice for this. Use existing Tracer directly or extend it. As you have your own Advice, you can control which Tracer you use.

======

The analysis above has some assumptions built-in. First, it assumes that custom instrumentation writer has access to the source code of the original instrumentation. They will have to copy-paste code from `transformers()` method of the original instrumentation. This means, among other things, that changes in the original instrumentation's `transformers` method may be breaking changes for custom instrumentation.

Second, it assumes that Otel `TypeInstrumentations` are sufficiently fine-grained. If an instrumentation has a large number of advices packed into it, it may be difficult to replace just one of them.

Also this analysis does not address the problem of changing a lot of similar instrumentations at once. E.g. "in all database client instrumentations add this extra attribute". This may be achieved by extending #778.

iNikem:
Closes #1623

### DIFF
--- a/examples/distro/README.md
+++ b/examples/distro/README.md
@@ -24,3 +24,36 @@ customize that
 * [DemoSpanProcessor](custom/src/main/java/com/example/javaagent/DemoSpanProcessor.java) - custom `SpanProcessor`
 * [DemoSpanExporter](custom/src/main/java/com/example/javaagent/DemoSpanExporter.java) - custom `SpanExporter`
 * [DemoServlet3Instrumentation](instrumentation/servlet-3/src/main/java/com/example/javaagent/instrumentation/DemoServlet3Instrumentation.java) - additional instrumentation
+
+## Instrumentation customisation
+
+There are several options to override or customise instrumentation provided by the upstream agent.
+The following description follows one specific use-case:
+
+> Instrumentation X from Otel distribution creates span that I don't like and I want to change it in my vendor distro.
+
+As an example, let us take some database client instrumentation that creates a span for database call
+and extracts data from db connection to provide attributes for that span.
+
+### I don't want this span at all
+The easiest case. You can just pre-configure your distribution and disable given instrumentation.
+
+### I want to add/modify some attributes and their values does NOT depend on a specific db connection instance.
+E.g. you want to add some data from call stack as span attribute. 
+In this case just provide your custom `SpanProcessor`.
+No need for touching instrumentation itself.
+
+### I want to add/modify some attributes and their values depend on a specific db connection instance.
+Write a _new_ instrumentation which injects its own advice into the same method as the original one.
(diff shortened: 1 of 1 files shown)


examples/distro/README.md
```diff
@@ -24,3 +24,36 @@ customize that
 * [DemoSpanProcessor](custom/src/main/java/com/example/javaagent/DemoSpanProcessor.java) - custom `SpanProcessor`
 * [DemoSpanExporter](custom/src/main/java/com/example/javaagent/DemoSpanExporter.java) - custom `SpanExporter`
 * [DemoServlet3Instrumentation](instrumentation/servlet-3/src/main/java/com/example/javaagent/instrumentation/DemoServlet3Instrumentation.java) - additional instrumentation
+
+## Instrumentation customisation
```
jkwatson, line 28: British spelling, eh? Should we standardise on the British, or American spellings i wonder?

jkwatson:
This is good. It would be even better if we could provide examples of each of these options in the example distro, and link to them. But, this is a good start!  👍

trask (approved):

examples/distro/README.md
iNikem: I am simple Russian guy :). Whatever spelling suggests me in the specific tool...

iNikem:
> This is good. It would be even better if we could provide examples of each of these options in the example distro, and link to them. But, this is a good start! 👍

I have added the link to already existing example.

PR #2113, opened 2021-01-25
Replace two JavaCC lexers with a single JFlex one

mateuszrzeszutek:
I'mm introduce a single property to control the sanitization in next PR - this one just replaces the lexer implementation.

@breedx-splk if you find some time, could you run your petclinic benchmark against the agent built from this branch? I'd be very grateful for that 😊


javaagent-api/javaagent-api.gradle
```diff
@@ -10,20 +10,15 @@ apply from: "$rootDir/gradle/publish.gradle"
...
-      outputDir = file("${project.buildDir}/generated/javacc/sql")
-    }
-    statmentInfoExtractor {
-      inputFile = file('src/main/javacc/SqlStatementInfoExtractor.jj')
-      packageName = 'io.opentelemetry.javaagent.instrumentation.api.db.sanitizer'
-      outputDir = file("${project.buildDir}/generated/javacc/sql")
-    }
-  }
+def jflexTargetDir = file"${project.buildDir}/generated/jflex/sql"
+def jflexTask = tasks.create("sqlSanitizerJflex", org.xbib.gradle.plugin.JFlexTask) {
```
mateuszrzeszutek, line 14: The jflex plugin was supposed to automatically create a task for this, but it's currently broken... I'll create an issue in their project. Meanwhile, a more manual approach still works

javaagent-api/src/main/jflex/SqlSanitizer.flex
```diff
@@ -0,0 +1,378 @@
...
+    boolean expectingTableName = false;
+    boolean mainTableSetAlready = false;
+    int identifiersAfterMainFromClause = 0;
+
+    boolean handleFrom() {
+      if (parenLevel == 0) {
+        // main query FROM clause
+        expectingTableName = true;
+        return false;
+      } else {
```
breedx-splk, line 145: would prefer to remove redundant `else` when `if{}` returns.

breedx-splk (approved):
I really like this approach.  I definitely appreciate that there are multiple duplicate blocks in type-specific normalizers that have collapsed into one simple statement now!  🎉 

For those who wanted to see some additional context, @mateuszrzeszutek posted some benchmark results here:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/2065#discussioncomment-302561

I definitely glossed over the middle of the lexer impl, but didn't see anything too shocking.  I'm going to also try running the other benchmarks (spring pet clinic rest + a6) and will report back.

javaagent-api/src/main/jflex/SqlSanitizer.flex
trask: I think this may be subjective, and I don't think there's an established convention in this codebase (I've seen lots of both styles). Should we add to https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/docs/contributing/style-guideline.md? IIRC @jkwatson has expressed this preference previously as well.

breedx-splk:
I ran the benchmarks this afternoon and got these results:
Nice looking results. 

| whch jar | avg (ms) | p95 (ms) |
|---------|----------|-----------|
| No agent  | 77.91 | 103.46 |
| Nightly snap (no cache, no jflex) | 314.61 | 499.00 |
| Nightly snap normalizer disabled | 200.10 | 306.28 |
| Better-lexer | 128.27 | 265.94 |
| Better-lexer normalizer disabled | 101.03 | 148.36 |
| Better-lexer jdbc inst disabled | 80.17 | 99.85 |

breedx-splk:
I also looked at GC.  The snapshot build had 85 GCs and this branch only had 29 GCs.  The total time in stop the world decreased from 0.578s in the snapshot to 0.403s in this branch.

More data from the benchmarks:

## Latest SNAPSHOT GC
image

## Better-lexer branch
image

breedx-splk:
And with the `profile` config, we can look at this significant reduction in TLABs.  I thought this was surprising, but if you 

## SNAPSHOT TLABs
image
image

## Better Lexer TLABs
image
image

iNikem (approved):

javaagent-api/src/main/jflex/SqlSanitizer.flex
```diff
@@ -0,0 +1,378 @@
...
+IDENTIFIER        = ([:letter:] | "_") ([:letter:] | [0-9] | [_.])*
+BASIC_NUM         = [.+-]* [0-9] ([0-9] | [eE.+-])*
+HEX_NUM           = "0x" ([a-f] | [A-F] | [0-9])+
+QUOTED_STR        = "'" ("''" | [^'])* "'"
+DOUBLE_QUOTED_STR = "\"" ("\"\"" | [^\"])* "\""
+DOLLAR_QUOTED_STR = "$$" [^$]* "$$"
+WHITESPACE        = [ \t\r\n]+
+
+%{
+  public static SqlStatementInfo sanitize(String statement) {
```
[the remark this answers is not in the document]
mateuszrzeszutek: I was already planning to do something similar in the next PR, because I don't want to use `Config` inside generated code - but I guess I can just do it here.

javaagent-api/src/main/jflex/SqlSanitizer.flex
```diff
@@ -0,0 +1,378 @@
+/*
```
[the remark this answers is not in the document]
mateuszrzeszutek: I'll move the entire `db` package in the another PR - there are ~4 classes that need to be moved together.

javaagent-api/src/main/jflex/SqlSanitizer.flex
```diff
@@ -0,0 +1,378 @@
...
+DOUBLE_QUOTED_STR = "\"" ("\"\"" | [^\"])* "\""
+DOLLAR_QUOTED_STR = "$$" [^$]* "$$"
+WHITESPACE        = [ \t\r\n]+
+
+%{
+  public static SqlStatementInfo sanitize(String statement) {
+    if (statement == null) {
+        return new SqlStatementInfo(null, null, null);
+    }
+    SqlSanitizer sanitizer = new SqlSanitizer(new java.io.StringReader(statement));
```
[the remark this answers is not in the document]
mateuszrzeszutek: Yeah, that would be awesome... and probably would improve the performance even more, since char[] copying seems to take a lot of time in Jason's perf test. Unfortunately jflex does not allow this degree of customisation.

The generated code is pretty much unreadable.
I think that maybe we could use a semi-manual approach: keep a flex grammar file + e.g. patch file with customizations applied to the generated code; whenever somebody has to change something they would have to generate+apply changes by themselves. And fix the patch when it breaks 😅

anuraaga (approved):
Thanks!

mateuszrzeszutek:
Thanks for performance tests @breedx-splk 🙏  It's great to see how much the response times & GC have improved - and this should get even better once we introduce the cache 🎉 

>  I definitely appreciate that there are multiple duplicate blocks in type-specific normalizers that have collapsed into one simple statement now!

Those duplicates will disappear in next PR, when I introduce one property to rule them all 😄

javaagent-api/src/main/jflex/SqlSanitizer.flex
```diff
@@ -0,0 +1,372 @@
...
+
+%unicode
+%ignorecase
+
+COMMA             = ","
+OPEN_PAREN        = "("
+CLOSE_PAREN       = ")"
+OPEN_COMMENT      = "/*"
+CLOSE_COMMENT     = "*/"
+IDENTIFIER        = ([:letter:] | "_") ([:letter:] | [0-9] | [_.])*
```
johnbley, line 24: I'm so happy that the abomination I created to represent the unicode letter character class is gone!

johnbley (approved):

mateuszrzeszutek:
(Rebased to get the netty 4.1 test fix)

trask (approved):
👍

PR #2115, opened 2021-01-26
Clean up some spring-boot auto-configuration to get rid of deprecations

jkwatson:
Also, update the docs to point at a more recent release.

### DIFF
--- a/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfiguration.java
+++ b/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfiguration.java
@@ -5,21 +5,23 @@
 
 package io.opentelemetry.instrumentation.spring.autoconfigure;
 
-import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.api.trace.TracerProvider;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
-import io.opentelemetry.sdk.trace.SpanProcessor;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.sdk.trace.SdkTracerProviderBuilder;
 import io.opentelemetry.sdk.trace.config.TraceConfig;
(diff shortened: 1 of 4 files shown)


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfiguration.java
```diff
@@ -34,38 +36,30 @@
...
+    return tracerProvider.get(tracerProperties.getName());
   }
 
-  private void addSpanProcessors(List<SpanExporter> spanExporters) {
-    List<SpanProcessor> spanProcessors =
-        spanExporters.stream()
-            .map(spanExporter -> SimpleSpanProcessor.builder(spanExporter).build())
-            .collect(Collectors.toList());
+  @Bean
+  @Scope("singleton")
```
trask, line 44: I think this is the default
```suggestion
```

trask (approved):

anuraaga (approved):

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfiguration.java
jkwatson: fair enough. :)

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/TracerAutoConfiguration.java
```diff
@@ -5,21 +5,23 @@
...
 import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
```
jkwatson, line 24:
```suggestion
```

PR #2100, opened 2021-01-22
Add Netty instrumentation handlers after matched handler rather than at end of pipeline.

anuraaga:
While I think it's generally a bug, in practice we can't be sure that the pipeline will propagate events we need through to the end because who knows what the user's handlers will be doing. We can be sure that Netty's default handlers do though, so `addAfter` is safer than `addLast`.

Fixes #2098 

This is close, tests pass but I get this mysterious error. @trask do you know the magic?

```
Transformed classes match global libraries ignore matcher: [org.springframework.core.ReactiveAdapterRegistry$EmptyCompletableFuture]
java.lang.AssertionError: Transformed classes match global libraries ignore matcher: [org.springframework.core.ReactiveAdapterRegistry$EmptyCompletableFuture]
	at io.opentelemetry.instrumentation.test.AgentTestRunner.agentCleanup(AgentTestRunner.java:111)
	at org.spockframework.util.ReflectionUtil.invokeMethod(ReflectionUtil.java:200)
	at org.spockframework.runtime.model.MethodInfo.invoke(MethodInfo.java:113)
	at org.spockframework.runtime.extension.MethodInvocation.proceed(MethodInvocation.java:98)
	at org.spockframework.spring.SpringInterceptor.interceptCleanupSpecMethod(SpringInterceptor.java:82)
	at org.spockframework.runtime.extension.AbstractMethodInterceptor.intercept(AbstractMethodInterceptor.java:39)
	at org.spockframework.runtime.extension.MethodInvocation.proceed(MethodInvocation.java:97)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at org.junit.vintage.engine.execution.RunnerExecutor.execute(RunnerExecutor.java:43)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
[...]

PR #2122, opened 2021-01-26
Fix netty 4.1 tests

laurit:


mateuszrzeszutek (approved):

trask (approved):
@laurit can you add a bit to the description of this PR? e.g. what the failure was, and (if known) when the failure was introduced?

laurit:
@trask currently most of the pull requests are failing to build because of netty 4.1 tests. It was last changed in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2100 though as it didn't fail idk, maybe it only fails together with some other pr that was merged around the same time.

Issue #1644, opened 2020-11-16
Publish testing-common


iNikem:
In order to allow vendor distribution to unit-test their custom instrumentation we should publish our test infrastructure, at least `testing-common` module.

anuraaga:
Will we force all vendors to use Spock? Ideally not :-)

iNikem:
Don't see anything bad in this :) What do you suggest?

anuraaga:
Ideally we can move as much code into Java classes that don't depend on Spock but just operate on the agent, or our instrumentation APIs, etc, probably by presenting some lifecycle methods (`beforeAgentTest`, etc). Then, it probably (?) wouldn't be hard to call these from any testing framework, spock or junit. Hopefully #1146 will help move even more logic out of our spock lifecycle classes and this may become even simpler. I think it's good to give downstream a choice in test frameworks.

PR #2133, opened 2021-01-28
Publish agent-for-testing and testing-common

mateuszrzeszutek:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1644. There's Groovy/Spock -> Java refactoring left, but it's better to leave that for another PR (or a series of them).

Published `agent-for-testing` (which includes `agent-exporter`) and `testing-common`.
I've removed `test-utils` module because it only contained one small class -- less artifacts to publish.


iNikem (approved):

trask (approved):
> I've removed `test-utils` module because it only contained one small class

👍

Issue #907, opened 2020-08-06
Remove CallDepth tracking from all spans except INTERNAL


trask:
Once #465 is implemented (or as it is implemented across instrumentations), we can remove CallDepth tracking since that will then be redundant.

trask:
Probably worth implementing #1009 at the same time.

iNikem:
Btw, I realized that we have nice symmetry here. Server spans should verify already existing server span, client spans should verify already existing client span, internal spans should verify call depth. But essentially this is the same pattern everywhere.

PR #2135, opened 2021-01-28
Add a metric for counting the number of suppressed spans.

jkwatson:
This will help in diagnosing mysterious "missing span" issues.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -85,28 +85,33 @@ public abstract class BaseTracer {
     return parentContext.with(span).with(CONTEXT_SERVER_SPAN_KEY, span);
   }
 
+<<<<<<< HEAD
   protected final boolean shouldStartSpan(SpanKind proposedKind, Context context) {
     boolean result;
+=======
+  public Scope startScope(Span span) {
+    return Context.current().with(span).makeCurrent();
+  }
+
+  protected final boolean shouldStartSpan(Kind proposedKind, Context context) {
+    boolean suppressed = false;
+>>>>>>> 4174bc5c4 (cleanup from PR comments)
     switch (proposedKind) {
       case CLIENT:
-        result = !inClientSpan(context);
+        suppressed = inClientSpan(context);
         break;
       case SERVER:
-        result = !inServerSpan(context);
+        suppressed = inServerSpan(context);
         break;
-      default:
-        result = true;
     }
-    if (!result) {
-      if (log.isDebugEnabled()) {
-        log.debug(
-            "Suppressing "
-                + proposedKind
-                + " span. Suppressed instrumentation: "
-                + getInstrumentationName());
-      }
+    if (suppressed && log.isDebugEnabled()) {
+      log.debug(
+          "Suppressing "
+              + proposedKind
+              + " span. Suppressed instrumentation: "
+              + getInstrumentationName());
     }
-    return result;
+    return !suppressed;
   }
 
   private boolean inClientSpan(Context parentContext) {
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -7,6 +7,10 @@ package io.opentelemetry.instrumentation.api.tracer;
 
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.common.Labels;
+import io.opentelemetry.api.metrics.GlobalMetricsProvider;
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.api.metrics.Meter;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
@@ -22,6 +26,9 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 public abstract class BaseTracer {
+  private static Meter meterProvider =
+      GlobalMetricsProvider.getMeter("io.opentelemetry.instrumentation.api.tracer");
+
   private static final Logger log = LoggerFactory.getLogger(BaseTracer.class);
 
   // Keeps track of the server span for the current trace.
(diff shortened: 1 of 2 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -25,7 +25,7 @@
 import org.slf4j.LoggerFactory;
 
 public abstract class BaseTracer {
-  private static final Logger log = LoggerFactory.getLogger(HttpServerTracer.class);
+  private static final Logger log = LoggerFactory.getLogger(BaseTracer.class);
```
jkwatson, line 28: heh. this was also wrong and very confusing when I saw it in the logs while verifying. :)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -81,14 +81,27 @@ public Span getCurrentSpan() {
...
-        return !inClientSpan(context);
+        result = !inClientSpan(context);
+        break;
       case SERVER:
-        return !inServerSpan(context);
+        result = !inServerSpan(context);
+        break;
       default:
-        return true;
+        result = true;
```
breedx-splk, line 93: you could move this assignment to the declaration and remove the default to save a couple lines of code.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -81,14 +81,27 @@ public Span getCurrentSpan() {
...
     }
+    if (!result) {
+      if (log.isDebugEnabled()) {
```
breedx-splk, line 96: prefer to roll these nested `if`s to an `&&`.

breedx-splk (approved):
Looks good to me, a few small things for polish if you wanna.

trask (commented):
Oh, I just remembered, we have been removing CallDepth tracking (which prevents an instrumentation from creating multiple nested spans) in places where it is redundant with the nested CLIENT span check (https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/907), but that means those instrumentations will trigger this "suppressed instrumentation" log even though that's not really the case.

I'm not really sure how to detect a "real" suppressed instrumentation 🤔

jkwatson:
> Oh, I just remembered, we have been removing CallDepth tracking (which prevents an instrumentation from creating multiple nested spans) in places where it is redundant with the nested CLIENT span check (#907), but that means those instrumentations will trigger this "suppressed instrumentation" log even though that's not really the case.
> 
> I'm not really sure how to detect a "real" suppressed instrumentation 🤔

Can you point me at an example of this? Maybe we need to update the code to only use this code path when we're actually doing "real" span suppression (whatever that means).

trask:
Check out #1811 for a bunch of places where I removed the CallDepth tracking since we can now rely on nested client span suppression.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -81,14 +81,27 @@ public Span getCurrentSpan() {
   }
 
   protected final boolean shouldStartSpan(Kind proposedKind, Context context) {
+    boolean result;
```
iNikem, line 84: Rename `result` to `suppressed` and remove some of negations?

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
jkwatson: done

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
jkwatson: done

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
jkwatson: done

jkwatson:
> Oh, I just remembered, we have been removing CallDepth tracking (which prevents an instrumentation from creating multiple nested spans) in places where it is redundant with the nested CLIENT span check (#907), but that means those instrumentations will trigger this "suppressed instrumentation" log even though that's not really the case.
> 
> I'm not really sure how to detect a "real" suppressed instrumentation 🤔

So... is this an objection to this PR? Or just a musing that the word "suppression" might not be 100% accurate? If we're suppressing spans, we're suppressing them, whether or not it's due to too many nested spans, or other reasons, no?

trask:
> So... is this an objection to this PR? Or just a musing that the word "suppression" might not be 100% accurate? If we're suppressing spans, we're suppressing them, whether or not it's due to too many nested spans, or other reasons, no?

I'm not sure how it's going to be useful since it's going to log lots of false positives due to the removal of CallDepth tracking.

Interestingly, in order to implement https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/465#issuecomment-769429325, we will probably have to bring back CallDepth tracking (or something similar), so we know whether we should suppress due to "CallDepth", or whether we should add an event to current span.

jkwatson:
> > So... is this an objection to this PR? Or just a musing that the word "suppression" might not be 100% accurate? If we're suppressing spans, we're suppressing them, whether or not it's due to too many nested spans, or other reasons, no?
> 
> I'm not sure how it's going to be useful since it's going to log lots of false positives due to the removal of CallDepth tracking.
> 
> Interestingly, in order to implement [#465 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/465#issuecomment-769429325), we will probably have to bring back CallDepth tracking (or something similar), so we know whether we should suppress due to "CallDepth", or whether we should add an event to current span.

Are those actually false positives though? We are suppressing the spans in those cases, aren't we?

jkwatson:
I changed this to use a LongCounter to keep track of the suppressed spans. There is one optimization to be made if we want to go this route.

jkwatson:
@trask @iNikem @anuraaga thoughts on the current state, using metrics for this?

trask (approved):
> thoughts on the current state, using metrics for this?

I think it's a cool idea, let's try it out 👍

anuraaga (approved):

iNikem:
Metrics are cool, but... :) They are alpha and will remain such for a long time. Do we want for `instrumentation-api` do depend on alpha artifact? We cannot then release this API as stable., can we?

anuraaga:
@iNikem Hmm this is a good point. Even if it's an implementation detail for us, it could still cause dependency hell for users regardless.

But my guess is metrics will stabilize before instrumentation - we still have the question of naming things tracer vs instrumenter, only because of metrics.

So I'm ok with it and if we actually release instrumentation first at the time we can flag guard or something similar.

iNikem:
> But my guess is metrics will stabilize before instrumentation

You are an optimist about metrics and pessimist about instrumentations :D

jkwatson:
Just thinking out loud, but another option would be to create some sort of "Supportability" interface that could be implemented by by whatever the user would prefer (logging, or metrics, or micrometer metrics, or a no-op). So, we could decouple the implementation from the need for providing some sort of observability for the internals of the instrumentation APIs/agent.

iNikem (approved):

PR #2137, opened 2021-01-29
Update to SDK 0.15.0

anuraaga:


iNikem (approved):
But this got me thinking... Metrics are still alpha and there seems to be a motion for their major overhaul. Can we NOT support metrics and save us from the pain of keeping up to date with ever changing metrics API?

anuraaga:
@iNikem Yeah already ended up updating them so left them in, but deleting would be ok too I think.

Also FYI, made a largish change to lightstep propagator test https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2137/commits/739b6a66b755727e3cfa6b327b5ec8918e408d37

mateuszrzeszutek (approved):

PR #2139, opened 2021-01-29
Wicket instrumentation

laurit:
- Update server span name based on displayed page
- Record unhandled exceptions on server span


instrumentation/wicket-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/wicket/WicketInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
...
+package io.opentelemetry.javaagent.instrumentation.wicket;
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.tooling.InstrumentationModule;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.Arrays;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class WicketInstrumentationModule extends InstrumentationModule {
```
iNikem, line 15: Please add classloader matcher to avoid inspecting every class by type matchers.

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
```diff
@@ -0,0 +1,21 @@
...
+  pass {
+    group = 'org.apache.wicket'
+    module = 'wicket'
+    versions = "[6.0.0,]"
+    assertInverse = true
+  }
+}
+
+dependencies {
+  library group: 'org.apache.wicket', name: 'wicket', version: '8.0.0'
```
iNikem, line 13: Why not 6.0.0?

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
```diff
@@ -0,0 +1,21 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    group = 'org.apache.wicket'
+    module = 'wicket'
+    versions = "[6.0.0,]"
+    assertInverse = true
```
iNikem, line 8: I don't think this line is needed. @trask @mateuszrzeszutek Do we have any guidelines, when to use inverse assert and when not?

iNikem (changes requested):

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
mateuszrzeszutek: I don't think that we have any "official" recommendation, but I remember that in some discussion somebody has said that "it should have been turned on by default" - and I can agree with that.
If there's some old library version that we know it can't be instrumented properly, but muzzle reference check does not block it from being applied, then `assertInverse` forces the author of the instrumentation module to use `classLoaderMatcher()` to make sure that this old version won't be instrumented (and thus won't break the instrumented application).

instrumentation/wicket-6.0/javaagent/src/test/groovy/WicketTest.groovy
```diff
@@ -0,0 +1,97 @@
...
+import okhttp3.RequestBody
+import okhttp3.Response
+import org.apache.wicket.protocol.http.WicketFilter
+import org.eclipse.jetty.server.Server
+import org.eclipse.jetty.servlet.DefaultServlet
+import org.eclipse.jetty.servlet.ServletContextHandler
+import org.eclipse.jetty.util.resource.FileResource
+import org.jsoup.Jsoup
+
+class WicketTest extends AgentTestRunner implements HttpServerTestTrait<Server> {
```
mateuszrzeszutek, line 23: Can you extend `AgentInstrumentationSpecification` instead? It'll replace direct usages of `AgentTestRunner`

```suggestion
class WicketTest extends AgentInstrumentationSpecification implements HttpServerTestTrait<Server> {
```

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
laurit: Sure I can change it to 6.0.0 but it was released in 24-Aug-2012, 8.0.0 is from 16-May-2018. Our tests don't always run agains the earliest possible version, for example springmvc-3.1 runs agains 4.3.20, tomcat-7.0 runs agains 9.0.41, jaxrs-2.0-jersey-2.0 runs against 2.16. Should we test with the earliest possible version or something that has not yet started to fossilize?

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
laurit: I thought that `assertInverse` is used to ensure that version range starts with earliest possible version where muzzle checks pass. For example adding `assertInverse = true` to struts-2.3 would fail `./gradlew :instrumentation:struts-2.3:javaagent:muzzle`

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
iNikem: We should test earliest support version. If you have added support for Wicket 6+, then we should test with 6.0 :)

If some modules don't do that, then it is their fault and we should fix it.

iNikem: Or have clearly documented reason, why not.

instrumentation/wicket-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/wicket/WicketInstrumentationModule.java
laurit: Is this really needed? Existing instrumentations that match only by class name don't use it unless there are separate instrumentations for different version of the same thing e.g. servlet.

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
mateuszrzeszutek: > For example adding assertInverse = true to struts-2.3 would fail ./gradlew :instrumentation:struts-2.3:javaagent:muzzle

Then it would be a good idea to introduce a `classLoaderMatcher()` to struts-2.3; or take a look at the instrumentation and try yo determine why it does not work for earlier versions.

Anyway, since usage patterns of `assertInverse` seem to be a bit unclear, we should improve the muzzle docs - I'll take care of it.

instrumentation/wicket-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/wicket/WicketInstrumentationModule.java
iNikem: All `InstrumentationModules` should have classloader matches. Matching every class is much more expensive that eliminating the whole classloader.

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
laurit: Instrumentation probably does work, test might not. Muzzle fails with MUZZLE PASSED Struts2InstrumentationModule BUT FAILURE WAS EXPECTED. Considering Struts 2.3.1 is dated 08-Dec-2011 not bothering with earlier versions is probably a deliberate choice.

instrumentation/wicket-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/wicket/WicketInstrumentationModule.java
laurit: Bytebuddy applies all matchers to input class. Class loader and type matcher are run together with
```
      public boolean matches(TypeDescription typeDescription, ClassLoader classLoader, JavaModule module, Class<?> classBeingRedefined, ProtectionDomain protectionDomain) {
        return this.moduleMatcher.matches(module) && this.classLoaderMatcher.matches(classLoader) && this.typeMatcher.matches(typeDescription);
      }
```
For class loader matcher to have a positive perf effect matching class loader needs to be less expensive than matching type. If type matcher is just a `String.equals` then it is likely to be less expensive than class loader matcher which is map lookup and when not found a `ClassLoader.getResouce`. Obviously for name based matchers it would best if they were kept in map name->list of matchers which would eliminate a bunch of matching.

iNikem: Hmmm, @trask @tylerbenson @mateuszrzeszutek help me here :) I was sure that classloader matches are there for performance optimisation. Was I wrong?

instrumentation/wicket-6.0/javaagent/wicket-6.0-javaagent.gradle
laurit: Changed supported version to 8.0, turns out I forgot to test with 6.0 and test didn't pass there.

instrumentation/wicket-6.0/javaagent/src/test/groovy/WicketTest.groovy
laurit: Changed.

instrumentation/wicket-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/wicket/WicketInstrumentationModule.java
tylerbenson: The classloader matcher was added mainly in places where the type matcher was relatively expensive (eg exploring the type hierarchy), but for a simple `named` matcher it's not really needed (unless I missed some other motivation).

mateuszrzeszutek: Classloader matchers defined in `TypeInstrumentation#classLoaderOptimization()` are used purely for optimization. These are usually defined when we use `extends`/`implements` type matcher.

`InstrumentationModule#classLoaderMatcher()` is used whenever there's a need to manually disable instrumenting some library versions. E.g. library had class `A` in version 1.0, but we want to instrument `[2.0,)` and the class is not directly used in instrumentation code (so muzzle does not have reference to `A`) - `classLoaderMatcher()` can be used in this case to make sure that 1.0 will never be instrumented.

Since all matchers are run together for every class I don't think it is necessary for the instrumentation module to use classloader matchers for optimization - the ones in `TypeInstrumentation` should be enough.

mateuszrzeszutek (approved):

instrumentation/wicket-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/wicket/WicketInstrumentationModule.java
iNikem: Thank you for explanation! I indeed was confused :)

iNikem (approved):

Issue #2153, opened 2021-02-01
Flaky test: :instrumentation:okhttp:okhttp-3.0:javaagent:test


anuraaga:
I haven't seen a flaky test failure without any containers involved in a loong time I think. Recording for posterity, but I stupidly forgot to copy in the build scan and am just guessing it is an extremely rare case of going over our retry count.

```
2021-02-01T08:29:19.5321908Z > Task :instrumentation:okhttp:okhttp-3.0:javaagent:test
2021-02-01T08:29:19.5322899Z 
2021-02-01T08:29:19.5324158Z OkHttp3AsyncTest > connection error dropped request[0] FAILED
2021-02-01T08:29:19.5325510Z     Condition not satisfied:
2021-02-01T08:29:19.5326180Z 
2021-02-01T08:29:19.5327867Z     jdk.internal.reflect.GeneratedMethodAccessor18.invoke(Unknown Source)
2021-02-01T08:29:19.5329296Z 
2021-02-01T08:29:19.5330264Z     assert val == value
2021-02-01T08:29:19.5331287Z            |   |  |
2021-02-01T08:29:19.5332835Z            |   |  'java.lang.NullPointerException'
2021-02-01T08:29:19.5334221Z            |   false
2021-02-01T08:29:19.5335574Z            'java.net.ConnectException'
2021-02-01T08:29:19.5338710Z         at io.opentelemetry.instrumentation.test.asserts.InMemoryExporterAssert.assertTraces(InMemoryExporterAssert.groovy:59)
2021-02-01T08:29:19.5344593Z         at io.opentelemetry.instrumentation.test.AgentTestRunner.assertTraces(AgentTestRunner.java:121)
2021-02-01T08:29:19.5348210Z         at io.opentelemetry.instrumentation.test.base.HttpClientTest.connection error dropped request(HttpClientTest.groovy:341)
2021-02-01T08:29:19.5350545Z 
2021-02-01T08:29:19.5351398Z         Caused by:
2021-02-01T08:29:19.5352297Z         Assertion failed: 
2021-02-01T08:29:19.5352842Z 
2021-02-01T08:29:19.5353688Z         assert val == value
2021-02-01T08:29:19.5354661Z                |   |  |
2021-02-01T08:29:19.5355992Z                |   |  'java.lang.NullPointerException'
2021-02-01T08:29:19.5357157Z                |   false
2021-02-01T08:29:19.5358314Z                'java.net.ConnectException'
[...]

PR #2154, opened 2021-02-01
Disable retry on connection failure for okhttp3 tests

laurit:
Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2153
`OkHttp3AsyncTest` waits for request to complete with `latch.await(30, SECONDS)` if response or failure is not received in this time then the request will fail with `NullPointerException` by reading response that it has not received yet. 
`HttpClientTest "connection error dropped request"` expects the request to fail, connection timeout is set to `5s`. Usually it completes in ~10s without retry it completes in `~5s`. Same test with okhttp2 does not retry the request and completes in `5s`.
This PR removes retry from okhttp3 in the hopes that this makes test less flaky. I have seen this failure multiple times, but I have not seen a similar failure for okhttp2.


iNikem (approved):

trask (approved):

anuraaga:
@laurit Can you update the PR description with the motivation of the change? I don't quite understand how connection retry is related to the issue linked.

anuraaga:
Thanks!

Issue #2155, opened 2021-02-01
Review ignored HttpClientTest


iNikem:
A lot of subclasses of `HttpClientTest` ignore one or more tests. This leaves us blind to the actual problems with our instrumentations, where we don't support quite common cases, e.g. redirects. We should review and fix them all.

jaydeluca:
To add some context to this issue and to make sure I understand it - this is related to test classes that extend `AbstractHttpClientTest` and configure things like `optionsBuilder.disableTestRedirects()` which then causes a subset of tests [to be ignored](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/5daa37e782069aa767c3fc7ba3baf916b93ac2c1/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java#L254).

The ask is to update all the instrumentations that opt out of various tests so that they all properly instrument things like redirects, so that we can then remove these configurations.

Is that correct?

trask:
this is my understanding as well 👍

PR #2156, opened 2021-02-01
Add causality test with multiple concurrent requests

iNikem:
Several things happen in this PR, sorry for such a soup. I can split it into several PR if needed...

First, a new "causality test" was introduced to the `HttpClientTest`. It stresses our instrumentations of http clients in highly concurrent settings. Please read the test comments.

Second, this new test has to be disabled for a lot of http clients. This is very concerning, see #2155

Third, Reactor Netty 0.9 test was converted to the subclass of `HttpClientTest`.

Fourth, integration with Reactor Netty 1.0 was introduced.


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -467,6 +519,10 @@ abstract class HttpClientTest extends AgentInstrumentationSpecification {
     true
   }
 
+  boolean testCausality() {
+    true
```
jkwatson, line 523: Should we default to false here until we've got most of the instrumentation happy with these tests? Or is better to default to true and explicitly call out where we aren't working yet?

iNikem: I think the latter, yes. This way it will be easier to remedy the situation: just try to remove overrides one by one

README.md
```diff
@@ -352,7 +352,7 @@ These are the supported libraries and frameworks:
 | [RabbitMQ Client](https://github.com/rabbitmq/rabbitmq-java-client)                                                                   | 2.7+                           |
 | [Ratpack](https://github.com/ratpack/ratpack)                                                                                         | 1.4+                           |
 | [Reactor](https://github.com/reactor/reactor-core)                                                                                    | 3.1+                           |
-| [Reactor Netty](https://github.com/reactor/reactor-netty)                                                                             | 0.9+ (not including 1.0)       |
+| [Reactor Netty](https://github.com/reactor/reactor-netty)                                                                             | 0.9+                           |
```
trask, line 355: 👍

instrumentation/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/ReactorNettyInstrumentationModule.java
```diff
@@ -0,0 +1,101 @@
...
+import reactor.core.publisher.Mono;
+import reactor.netty.Connection;
+import reactor.netty.http.client.HttpClient;
+import reactor.netty.http.client.HttpClientRequest;
+
+/**
+ * This instrumentation solves the problem of the correct context propagation through the roller
+ * coaster of Project Reactor and Netty thread hopping. It uses two public hooks of {@link
+ * HttpClient}: {@link HttpClient#mapConnect(BiFunction)} and {@link
+ * HttpClient#doOnRequest(BiConsumer)} two short-cut context propagation from the caller to Reactor
```
trask, line 37: maybe i misunderstood though
```suggestion
 * HttpClient#doOnRequest(BiConsumer)} to pass context propagation from the caller to Reactor
```

instrumentation/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/ReactorNettyInstrumentationModule.java
```diff
@@ -0,0 +1,101 @@
...
+ * to Netty.
+ */
+@AutoService(InstrumentationModule.class)
+public class ReactorNettyInstrumentationModule extends InstrumentationModule {
+
+  public ReactorNettyInstrumentationModule() {
+    super("reactor-netty", "reactor-netty-0.9");
```
trask, line 44:
```suggestion
    super("reactor-netty", "reactor-netty-1.0");
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -395,6 +404,49 @@ abstract class HttpClientTest extends AgentInstrumentationSpecification {
     method = "HEAD"
   }
 
+  /**
+   * This test fires a large number of concurrent requests.
+   * Each request first hits a HTTP server and then makes another client request.
+   * The goal of this test is to verify that in highly concurrent environment our instrumentations
+   * for http clients (especially inherently concurrent ones, such as Netty or Reactor) correctly
+   * propagate trace context.
+   */
```
trask, line 413: is the second (downstream) http request needed? is testing concurrency of the initial request enough (under `runUnderTrace` to verify propagation incoming->outgoing)?

also (and maybe this is what I'm missing) why is the test named "causality test" instead of "concurrency test"?

instrumentation/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/ReactorNettyInstrumentationModule.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public List<TypeInstrumentation> typeInstrumentations() {
+    return singletonList(new HttpClientInstrumentation());
+  }
+
+  public static class HttpClientInstrumentation implements TypeInstrumentation {
+    @Override
+    public ElementMatcher<TypeDescription> typeMatcher() {
+      return named("reactor.netty.http.client.HttpClient");
```
trask, line 55: since both 0.9 and 1.0 apply to the same class, would be good to make sure that they don't conflict, e.g.
* add muzzle `assertInverse` if possible
* add a `testInstrumentation` dependency from 1.0 on 0.9 and vice versa (we do this in a couple of places, but should probably do it uniformly wherever we have instrumentation split across library versions)

trask (commented):
> First, a new "causality test" was introduced to the `HttpClientTest`. It stresses our instrumentations of http clients in highly concurrent settings. Please read the test comments.

❤️

> Second, this new test has to be disabled for a lot of http clients. This is very concerning, see #2155

😢

> Third, Reactor Netty 0.9 test was converted to the subclass of `HttpClientTest`.

👍

> Fourth, integration with Reactor Netty 1.0 was introduced.

🎉

instrumentation/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/ReactorNettyInstrumentationModule.java
anuraaga: Also debug logging will be confusing, it'd be nice to have a `classLoaderMatcher` to differentiate rather than only use the muzzle check

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
iNikem: "Causality" is due to internal jargon used by @johnbley : we want to check that outgoing request is correctly attributed to the incoming one, hence "causality".

> is the second (downstream) http request needed? is testing concurrency of the initial request enough (under runUnderTrace to verify propagation incoming->outgoing)?

In most cases indeed it would be sufficient to use `runUnderTrace` and to verify that the span, which was active before `doRequest`, is correctly propagated across the wire. But e.g. take an example of async servlets or WebFlux as `/front` request handlers (as separate tests, not part of `HttpClientTest`, but using the same test scenario). In these cases it's obvious where our instrumentation will attribute even the first span. May be it is sufficient to have them as the target of the second request, I don't know...

instrumentation/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/ReactorNettyInstrumentationModule.java
iNikem: Good points, will do.

iNikem: Done

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -395,6 +404,49 @@ abstract class HttpClientTest extends AgentInstrumentationSpecification {
...
+      pool.submit(job)
+    }
+    latch.countDown()
+
+    then:
+    assertTraces(count) {
+      count.times { traceIndex ->
+        trace(traceIndex, 4) {
+          clientSpan(it, 0, null, method, frontUrl)
+          serverSpan(it, 1, span(0))
```
jkwatson, line 441: I wonder if we should also have a common span attribute that varies with each requests (maybe via a tiny bit of added manual instrumentation pulling a value from a header?) so we can very that we're not just "getting lucky" and happen to end up with 4 lined up spans on each request?

johnbley: You mean like https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/4d59fc0c8198343fbb2bd29a02f197ae48324815/instrumentation/kotlinx-coroutines/javaagent/src/test/groovy/KotlinCoroutineInstrumentationTest.groovy#L200 ?

iNikem: Yeah, I skipped that in hurry. I will see if I can do that easily enough.

iNikem: Done

iNikem:
@mateuszrzeszutek @trask Can you help me? I don't understand why muzzle fails on Reactor Netty with
```
MUZZLE PASSED NettyInstrumentationModule BUT FAILURE WAS EXPECTED

> Task :instrumentation:reactor-netty-0.9:javaagent:muzzle-AssertFail-io.projectreactor.netty-reactor-netty-1.0.3 FAILED
```

Why does it fail due to `NettyInstrumentationModule`?

mateuszrzeszutek:
Hmm, it looks like for `fail` directives the instrumentation classloader is missing the reactor-netty instrumentation module:

(example for 1.0)
```
> Task :instrumentation:reactor-netty-1.0:javaagent:muzzle-AssertFail-io.projectreactor.netty-reactor-netty-0.9.12.RELEASE FAILED
CHECKING NettyInstrumentationModule
MUZZLE PASSED NettyInstrumentationModule BUT FAILURE WAS EXPECTED

> Task :instrumentation:reactor-netty-1.0:javaagent:muzzle-AssertPass-io.projectreactor.netty-reactor-netty-http-1.0.3
CHECKING NettyInstrumentationModule
CHECKING ReactorNettyInstrumentationModule
```

There must be something wrong with how `MuzzlePlugin` prepares classloaders

iNikem:
@mateuszrzeszutek I look at `io.opentelemetry.javaagent.tooling.muzzle.matcher.MuzzleGradlePluginUtil#assertInstrumentationMuzzled` and I see that 
```
        if (passed && !assertPass) {
          System.err.println(
              "MUZZLE PASSED "
                  + instrumentationModule.getClass().getSimpleName()
                  + " BUT FAILURE WAS EXPECTED");
          throw new RuntimeException("Instrumentation unexpectedly passed Muzzle validation");
```

Meaning that even if there are 2 modules present, the plugin throws exception and fails the build as soon as any of them pass muzzle. Which does not make sense in my case. I depend on classes from netty instrumentation, thus `NettyInstrumentationModule` is present on my classpath, but it certainly passes all muzzle checks, regardless of Reactor Netty version.

mateuszrzeszutek:
> Meaning that even if there are 2 modules present, the plugin throws exception and fails the build as soon as any of them pass muzzle.

Oh, that's true! I was so concentrated on the CL & gradle plugin thing that I missed the loop 😅 
Yes, I suppose that's where the problem is  - this `passed && !assertPass` if should be outside of the loop.

iNikem:
I went another way... Let's see...

johnbley (commented):
LGTM, thanks so much for pushing on this much-needed test type.

mateuszrzeszutek (approved):

anuraaga (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
trask: > take an example of async servlets or WebFlux as `/front` request handlers

ya, this makes sense to me, but since these tests don't do this, it's not clear why the extra complexity is needed (though I'm ok with it if you think good to keep it)

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -395,6 +401,58 @@ abstract class HttpClientTest extends AgentInstrumentationSpecification {
...
+    def latch = new CountDownLatch(1)
+
+    def pool = Executors.newFixedThreadPool(4)
+
+    when:
+    count.times { index ->
+      def job = {
+        latch.await()
+        runUnderTrace("Parent span " + index) {
+          Span.current().setAttribute("traceAge", index)
```
trask, line 427: `traceAge` made me think date, maybe `traceId`? ok, no, that's probably more confusing, maybe `testId`? or `testBaggage` 😁? or leave as is totally good too

instrumentation/netty/netty-4.1/library/netty-4.1-library.gradle
```diff
@@ -0,0 +1,5 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
```
trask, line 1:
```suggestion
apply from: "$rootDir/gradle/instrumentation-library.gradle"
```

trask: also, AttributeKeys below should be moved out of `javaagent` package

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
iNikem: I have removed the second http call :)

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
iNikem: I will leave it as it is for now, I don't think any other proposed name is much better :(

trask: oh, i had a better idea for name:
* attribute name: `test.request.id`
* header name: `test-request-id` (if we make it all lowercase then don't need special case for google-http-client below)

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -395,6 +401,58 @@ abstract class HttpClientTest extends AgentInstrumentationSpecification {
...
   }
 
+  /**
+   * This test fires a large number of concurrent requests.
+   * Each request first hits a HTTP server and then makes another client request.
+   * The goal of this test is to verify that in highly concurrent environment our instrumentations
+   * for http clients (especially inherently concurrent ones, such as Netty or Reactor) correctly
+   * propagate trace context.
+   */
+  def "causality test"() {
```
trask, line 411: causality -> concurrency? I think that will make it clearer for most people what these tests are doing

trask (approved):
Really great test addition 👍

Two last suggestions, no worries if you disagree please ignore.

PR #2171, opened 2021-02-02
Correct hibernate dtd address

laurit:
Dtd address should be what the framework expects so that it could use the copy of dtd that is bundled inside the jar instead of downloading it.


mateuszrzeszutek (approved):

anuraaga (approved):

iNikem (approved):

trask:
> Dtd address should be what the framework expects so that it could use the copy of dtd that is bundled inside the jar instead of downloading it.

nice

PR #2179, opened 2021-02-03
Filter out library jar from tests since it's shadowed into the agent …

anuraaga:
…/ agent-testing jar.

Found this when debugging the dubbo test failure, which is having unintended influence from the library jar on the test classpath. 

As with the previous classpath filtering, I can't come up with a clean check, but the heuristic is relatively safe (I guess?).


tydhot:
Thanks a lot!

trask (approved):
thanks for the comments on this trickiness!

PR #2181, opened 2021-02-03
Elasticsearch: fix unexpected spans for ConnectException: Connection refused

laurit:
- Configure elasticsearch to run in a single node cluster. This skips discovery of other cluster nodes. Currently this discovery process doesn't always complete before tests start to run so tests may fail because there is an unexpected span for `java.net.ConnectException: Connection refused`
- Shut down `PreBuiltTransportClient` to avoid it attempting to reconnect with a shut down node which can produce unexpected span for `java.net.ConnectException: Connection refused` in subsequent test


iNikem (approved):

mateuszrzeszutek (approved):

anuraaga (approved):

trask:
nice, thx!

PR #2184, opened 2021-02-03
Allow getting current call depth

mateuszrzeszutek:
That's a really minor thing, but it can be useful in vendor distributions when trying add some additional code to existing instrumentation.


trask (approved):

trask:
can you add a comment on that method that it's useful from vendor distros, so we don't accidentally remove it thinking it's unused?

iNikem (approved):

pavolloffay:
+1 In some causes I had to use additional `CallDepth` since I wanted to use the same class as a key.

pavolloffay (approved):

anuraaga (approved):

PR #2185, opened 2021-02-03
Fix JSP 2.3 muzzle

mateuszrzeszutek:
Tomcat 10.0.2 was released just today; version 10.0.x is supposed to use JSP 3.0 (and servlet 5.0) so this broke the muzzle task.

### DIFF
--- a/instrumentation/jsp-2.3/javaagent/jsp-2.3-javaagent.gradle
+++ b/instrumentation/jsp-2.3/javaagent/jsp-2.3-javaagent.gradle
@@ -6,7 +6,8 @@ muzzle {
     module = "tomcat-jasper"
     // range [7.0.0,7.0.19) and version 8.0.9 missing from maven
     // tomcat 10 uses JSP 3.0
-    versions = "[7.0.19,8.0.9),(8.0.9,10)"
+    versions = "[7.0.19,10)"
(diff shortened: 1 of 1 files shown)


instrumentation/jsp-2.3/javaagent/jsp-2.3-javaagent.gradle
```diff
@@ -5,7 +5,8 @@ muzzle {
     group = "org.apache.tomcat"
     module = "tomcat-jasper"
     // range [7.0.0,7.0.19) and version 8.0.9 missing from maven
-    versions = "[7.0.19,8.0.9),(8.0.9,)"
+    // tomcat 10 uses JSP 3.0
+    versions = "[7.0.19,8.0.9),(8.0.9,10)"
```
iNikem, line 9: Can we use `[7.0.19, 10)` and `skipVersions`?

iNikem (approved):

instrumentation/jsp-2.3/javaagent/jsp-2.3-javaagent.gradle
mateuszrzeszutek: Yep, definitely!

anuraaga (approved):

PR #2196, opened 2021-02-04
Remove all remaining direct AgentTestRunner usages

mateuszrzeszutek:
Now everybody's forced to use `AgentInstrumentationSpecification` when testing agent instrumentations

### DIFF
--- a/testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/AgentTestRunner.java
+++ b/testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/AgentTestRunner.java
@@ -15,26 +15,9 @@ import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.instrumentation.test.asserts.InMemoryExporterAssert;
 import io.opentelemetry.javaagent.testing.common.AgentTestingExporterAccess;
 import io.opentelemetry.javaagent.testing.common.TestAgentListenerAccess;
-import java.util.Collections;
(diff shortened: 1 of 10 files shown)


trask (approved):
niiice

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/AgentTestRunner.java
```diff
@@ -15,26 +15,9 @@
...
-/**
- * A spock test runner which automatically applies instrumentation and exposes a global trace
- * writer.
- *
- * <p>To use, write a regular spock test, but extend this class instead of {@link
- * spock.lang.Specification}.
- */
-@SpecMetadata(filename = "AgentTestRunner.java", line = 0)
-public abstract class AgentTestRunner extends Specification {
+public final class AgentTestRunner {
```
iNikem, line 20: Should it be public? If yes, then some javadoc is needed.

mateuszrzeszutek: Hmm, do you mind if I do it in the next PR? I planned to make `AgentTestRunner` and `InstrumentationTestRunner` implement a common interface, put them in the java source set (and in the next next PR introduce a JUnit base class/extension) -- anyway, I'll place some Javadoc in the next step, when I completely refactor this class.

iNikem (approved):

PR #2197, opened 2021-02-04
Add constructors/methods to the base tracers to enable injecting an OpenTelemetry instance

jkwatson:
I haven't gone and updated any of the many other sub-tracers to inject the global, but having the capability means we can start migrating toward that model. This will enable the spring sleuth integration to not have to mess around with the global instance during startup.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -205,16 +205,14 @@ public abstract class BaseTracer {
     return extract(GlobalOpenTelemetry.getPropagators(), carrier, getter);
   }
 
-  public static <C> Context extract(ContextPropagators propagators, C carrier,
-      TextMapPropagator.Getter<C> getter) {
+  public static <C> Context extract(
+      ContextPropagators propagators, C carrier, TextMapPropagator.Getter<C> getter) {
     ContextPropagationDebug.debugContextLeakIfEnabled();
 
     // Using Context.ROOT here may be quite unexpected, but the reason is simple.
     // We want either span context extracted from the carrier or invalid one.
     // We DO NOT want any span context potentially lingering in the current context.
-    return propagators
-        .getTextMapPropagator()
-        .extract(Context.root(), carrier, getter);
+    return propagators.getTextMapPropagator().extract(Context.root(), carrier, getter);
   }
 
   /** Returns span of type SERVER from the current context or <code>null</code> if not found. */
@@ -226,5 +224,4 @@ public abstract class BaseTracer {
   public static Span getCurrentServerSpan(Context context) {
     return context.get(CONTEXT_SERVER_SPAN_KEY);
   }
-
 }
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -45,6 +45,8 @@ public abstract class BaseTracer {
     propagators = GlobalOpenTelemetry.getPropagators();
   }
 
+  /** @deprecated prefer to pass in an OpenTelemetry instance, instead. */
+  @Deprecated
   public BaseTracer(Tracer tracer) {
     this.tracer = tracer;
     this.propagators = GlobalOpenTelemetry.getPropagators();
@@ -201,11 +203,18 @@ public abstract class BaseTracer {
     span.recordException(throwable);
   }
 
-  public static <C> Context extract(C carrier, TextMapPropagator.Getter<C> getter) {
+  /** @deprecated We should eliminate all static usages so we can use the non-global propagators. */
+  @Deprecated
+  public static <C> Context extractWithGlobalPropagators(
+      C carrier, TextMapPropagator.Getter<C> getter) {
     return extract(GlobalOpenTelemetry.getPropagators(), carrier, getter);
   }
 
-  public static <C> Context extract(
+  public <C> Context extract(C carrier, TextMapPropagator.Getter<C> getter) {
+    return extract(propagators, carrier, getter);
+  }
+
+  private static <C> Context extract(
       ContextPropagators propagators, C carrier, TextMapPropagator.Getter<C> getter) {
     ContextPropagationDebug.debugContextLeakIfEnabled();
 
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -45,7 +45,12 @@ public abstract class BaseTracer {
     propagators = GlobalOpenTelemetry.getPropagators();
   }
 
-  /** @deprecated prefer to pass in an OpenTelemetry instance, instead. */
+  /**
+   * Prefer to pass in an OpenTelemetry instance, rather than just a Tracer, so you don't have to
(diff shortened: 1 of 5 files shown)


instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
```diff
@@ -111,7 +110,7 @@ public HttpHost getTarget() {
     @Override
     public HttpRequest generateRequest() throws IOException, HttpException {
       HttpRequest request = delegate.generateRequest();
-      GlobalOpenTelemetry.getPropagators()
+      tracer().getPropagators()
```
jkwatson, line 113: I updated this particular instrumentation because I wanted to make sure that the basic idea was workable.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -32,6 +32,18 @@
 
   protected static final String USER_AGENT = "User-Agent";
 
+  protected HttpClientTracer() {
+    super();
+  }
+
+  protected HttpClientTracer(Tracer tracer) {
```
trask, line 39:
```suggestion
  @Deprecated
  protected HttpClientTracer(Tracer tracer) {
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -36,13 +38,25 @@
...
 
   protected final Tracer tracer;
+  protected final ContextPropagators propagators;
 
   public BaseTracer() {
     tracer = GlobalOpenTelemetry.getTracer(getInstrumentationName(), getVersion());
+    propagators = GlobalOpenTelemetry.getPropagators();
   }
 
   public BaseTracer(Tracer tracer) {
```
trask, line 48:
```suggestion
  @Deprecated
  public BaseTracer(Tracer tracer) {
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -188,14 +202,17 @@ public void addThrowable(Span span, Throwable throwable) {
   }
 
   public static <C> Context extract(C carrier, TextMapPropagator.Getter<C> getter) {
+    return extract(GlobalOpenTelemetry.getPropagators(), carrier, getter);
+  }
+
+  public static <C> Context extract(
+      ContextPropagators propagators, C carrier, TextMapPropagator.Getter<C> getter) {
```
trask, line 209: I'd suggest converting this to non-static so that it can use the new `propagators` instance field (I don't think it was particularly important for it to be static)
```suggestion
  public <C> Context extract(C carrier, TextMapPropagator.Getter<C> getter) {
```

trask (approved):
Thanks for this!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
jkwatson: yeah, I was trying to decide on whether to deprecate or not. If you think so, then I'm 💯  on that.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
jkwatson: the caller didn't have access to a BaseTracer, so I didn't want to go too crazy unwinding that sweater.

trask: ya no worries, can you add a TODO here

```
// TODO convert to non-static, so it can use the propagators instance field
```

jkwatson: Here's the troublesome usage:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/5f263644da9383616016d2cc007a3301f7885c4b/instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/server/ContextDispatcher.java#L53

trask: I think this could be converted to:

```
Context context = tracer().extract(payload, GETTER);
```

jkwatson: I can deprecate the static method and convert all the other usages to use a non-static one. That's not a huge change, and is a good move in the right direction.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
jkwatson: done

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
jkwatson: done

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -187,15 +203,25 @@ public void addThrowable(Span span, Throwable throwable) {
     span.recordException(throwable);
   }
 
-  public static <C> Context extract(C carrier, TextMapPropagator.Getter<C> getter) {
+  /** @deprecated We should eliminate all static usages so we can use the non-global propagators. */
+  @Deprecated
+  public static <C> Context extractWithGlobalPropagators(
```
jkwatson, line 208: how's this approach, @trask ? There's only one usage of it in the project at the moment.

trask: awesome

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -36,13 +38,32 @@
...
   public BaseTracer() {
     tracer = GlobalOpenTelemetry.getTracer(getInstrumentationName(), getVersion());
+    propagators = GlobalOpenTelemetry.getPropagators();
   }
 
+  /**
+   * Prefer to pass in an OpenTelemetry instance, rather than just a Tracer, so you don't have to
+   * use the GlobalOpenTelemetry Propagator instance.
+   *
+   * @deprecated prefer to pass in an OpenTelemetry instance, instead.
```
anuraaga, line 52: These artifacts aren't used much so not a big deal IMO, but as general practice, we'd link to the new method, `@deprecated Use {@link #BaseTracer(OpenTelemetry)} instead`

jkwatson: that's a good idea. Happy to amend this PR, or follow-up later.

anuraaga (approved):
Nah wouldn't worry too much about the doc here

iNikem (approved):

mateuszrzeszutek (approved):

PR #2206, opened 2021-02-05
Remove kotlin core libs from javaagent distro

trask:
these are adding ~3mb to the javaagent distro

### DIFF
--- a/javaagent-tooling/javaagent-tooling.gradle
+++ b/javaagent-tooling/javaagent-tooling.gradle
@@ -28,7 +28,12 @@ dependencies {
   implementation deps.opentelemetrySdk
   implementation deps.opentelemetrySdkAutoconfigure
   implementation deps.opentelemetrySdkMetrics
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

javaagent-tooling/javaagent-tooling.gradle
```diff
@@ -28,7 +28,12 @@ dependencies {
   implementation deps.opentelemetrySdk
   implementation deps.opentelemetrySdkAutoconfigure
   implementation deps.opentelemetrySdkMetrics
-  implementation deps.opentelemetryKotlin
+  implementation (deps.opentelemetryKotlin) {
```
iNikem, line 31: @anuraaga do you think that Kotlin extension should bring Kotlin with it? It should have `compileOnly` dependency on Kotlin stdlib?

iNikem (approved):

PR #2209, opened 2021-02-05
Add upload jar step to releasing doc

trask:


anuraaga:
Well if we're finally making a PR anyways think we can do it to the workflow instead? :)

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/.github/workflows/release-build.yml#L42

iNikem:
Or https://github.com/signalfx/splunk-otel-java/blob/main/.github/workflows/release.yaml#L31

trask:
Closing in favor of #2237

Issue #1716, opened 2020-11-20
Document possible values for otel.propagators


iNikem:
Currently README lists possible values for that property, but does not explain their meaning. E.g. what is `ottracer` propagator?

Also please document which ones of the supported propagators are for baggage and which ones are for traces.

FrankSpitulski:
ottracer uses OT_TRACER_* headers that lightstep uses in opentracing

PR #2222, opened 2021-02-08
aws lambda instrumentation - http headers propagation fix

### LINKED ISSUE #2131: AWS lambda instrumentation fails to propagate HTTP headers context
**Describe the bug**
- OTEL lambda tracer (io.opentelemetry.instrumentation.awslambda.v1_0.AwsLambdaTracer) sets parent span using either XRAY header or HTTP headers
- XRAY is preferred if header (env value - _X_AMZN_TRACE_ID) is set
- HTTP is used if XRAY parent is invalid (or XRAY header was not set)
- Currently AWS sets XRAY header not matter lambda settings - ie even with XRAY disabled in the console
- XRAY disabled - value is eg `Root=1-60115e2e-3bc996f80f14f6e33d18f37b;Parent=4761858a3c680442;Sampled=0` (notice sampled settings)
- XRAY enabled - value is eg `Root=1-60115e2e-3bc996f80f14f6e33d18f37b;Parent=4761858a3c680442;Sampled=1` (notice sampled settings)

**Steps to reproduce**
- deploy lambda with OTEL wrapper, configure any exporter

- HTTP headers should be used (if set) once XRAY is disabled

**What did you see instead?**
- XRAY propagation in all cases (enabled, disabled)

- latest (0.14.0, master)

**Environment**
NA

**Additional context**

1. First question is whether AWS should set XRAY header when tracing is disabled? If it didn't (which I think was the case some time ago), everything would work as expected, that is:
- XRAY enabled - propagation of XRAY context
- XRAY disabled, HTTP headers - propagation of HTTP context
- XRAY disabled, no HTTP headers - no propagation (no parent span)

2. If above behaviour can't be changed, I believe that the right way is to set HTTP propagation as preferred, this way we will get slightly different but still acceptable (in my opinion of course! :) ) behaviour: 
- XRAY enabled, HTTP headers - propagation of HTTP context
- XRAY enabled,  no HTTP headers - propagation of XRAY context
- XRAY disabled, HTTP headers - propagation of HTTP context
- XRAY disabled, no HTTP headers - propagation of XRAY context with parent that should not be sampled

@anuraaga please have an "expert look" here

kubawach:
Fixes #2131

- if AWS parent ("_X_AMZN_TRACE_ID" env var) is set but yields parent that should not be samples (sample = false), HTTP parent is tried

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/README.md
+++ b/instrumentation/aws-lambda-1.0/library/README.md
@@ -92,7 +92,7 @@ Context propagation for this instrumentation can be done either with X-Ray propa
 
 ### X-Ray propagation
 This instrumentation supports propagating traces using the `X-Amzn-Trace-Id` format for both normal
-requests and SQS requests. X-Ray propagation is always enabled, there is no need to configure it explicitely. 
+requests and SQS requests. X-Ray propagation is always enabled, there is no need to configure it explicitely.
 
 ### HTTP headers based propagation
 For API Gateway (HTTP) requests instrumented by using one of following methods:
--- a/instrumentation/aws-lambda-1.0/library/README.md
+++ b/instrumentation/aws-lambda-1.0/library/README.md
@@ -87,36 +87,12 @@ public class MyBatchHandler extends TracingSQSEventHandler {
 
 ## Trace propagation
 
+Context propagation for this instrumentation can be done either with X-Ray propagation or regular HTTP propagation. If X-Ray is enabled for instrumented lambda, it will be preferred. If X-Ray is disabled, HTTP propagation will be tried (that is HTTP headers will be read to check for a valid trace context).
+
+
 ### X-Ray propagation
 This instrumentation supports propagating traces using the `X-Amzn-Trace-Id` format for both normal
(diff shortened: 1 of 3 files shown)


mateuszrzeszutek (approved):

iNikem:
I think this should be documented somewhere

kubawach:
@iNikem do you have also an idea where?

iNikem:
> @iNikem do you have also an idea where?

In to-be-created https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1716 :D

kubawach:
@iNikem currently if needed I could add a sentence about that in `instrumentation/aws-lambda-1.0/library/README.md`. What do you think?

anuraaga (approved):

kubawach:
@iNikem added a paragraph about x-ray vs http propagation
@anuraaga removed paragraph about configuring X-ray propagation cause as far as I understand currently it's always enabled for the instrumentation (parent span resolving we discussed over :) ). Please have a look.

anuraaga:
Thanks, yeah we had never updated that.

PR #2223, opened 2021-02-08
Refactor AgentTestRunner: extract a common interface, convert to Java class

mateuszrzeszutek:
Traits got a lot smaller, all differences between agent and library instrumentation tests were moved to `InstrumentationTestRunner`. Next part will introduce JUnit 5 extensions (or base classes, I'll see which one is less painful to use 😄 )


testing-common/src/main/java/io/opentelemetry/instrumentation/testing/AgentTestRunner.java
```diff
@@ -0,0 +1,72 @@
...
+ * An implementation of {@link InstrumentationTestRunner} that delegates most of its calls to the
+ * OpenTelemetry Javaagent that this process runs with. It uses the {@link
+ * AgentTestingExporterAccess} bridge class to retrieve exported traces and metrics data from the
+ * agent classloader.
+ */
+public final class AgentTestRunner implements InstrumentationTestRunner {
+  static {
+    ((Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME)).setLevel(Level.WARN);
+    ((Logger) LoggerFactory.getLogger("io.opentelemetry")).setLevel(Level.DEBUG);
+  }
```
trask, line 27: I'm not sure if we still need this after #1643, but we can revisit later, no harm

trask (approved):
👍

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/AgentTestRunner.java
mateuszrzeszutek: I'll verify this and fix it in one of the next PRs.

anuraaga:
@mateuszrzeszutek By the way, if we can use, or have ideas for improving, https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk/testing that'd be pretty sweet

mateuszrzeszutek:
> @mateuszrzeszutek By the way, if we can use, or have ideas for improving, https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk/testing that'd be pretty sweet

Those assertions are pretty nice, it would be great if we could reuse them in the instrumentation repo... ideally we should only have a thin groovy wrapper over them.

PR #2226, opened 2021-02-08
README and docs reorg

breedx-splk:
This is follow-up from a community call last week, where we discussed the README.  The current state of the README is very long and it requires a lot of scanning/scrolling/searching to find what you might be looking for. With the goal of making things easier to adopt for newcomers, I set out to reduce some of the first-view GIANT WALL OF HIGHLY TECHNICAL TEXT impression.

This change pulls out a few larger chunks of the README into their own docs (within the docs subdir).  I feel that this separation is natural and should be easier for newcomers to navigate.  It also adds table of contents to the main readme and to the new files. In a few places I've tried to add friendly/encouraging language and/or make wording simpler.

It's probably obvious, but if you want to see this in action you can [click through to the PR branch](https://github.com/breedx-splk/opentelemetry-java-instrumentation/tree/docs_reorg).


docs/manual-instrumentation.md
```diff
@@ -0,0 +1,115 @@
...
+  * [Gradle](#gradle)
+- [Adding attributes to the current span](#adding-attributes-to-the-current-span)
+- [Creating spans around methods with `@WithSpan`](#creating-spans-around-methods-with---withspan-)
+  * [Suppressing `@WithSpan` instrumentation](#suppressing---withspan--instrumentation)
+  * [Creating spans around methods with `otel.instrumentation.methods.include`](#creating-spans-around-methods-with-otelinstrumentationmethodsinclude)
+- [Creating spans manually with a Tracer](#creating-spans-manually-with-a-tracer)
+
+# Dependencies
+
+> :warning: starting with 0.6.0, and prior to version 1.0.0, `opentelemetry-javaagent-all.jar`
```
trask, line 21: I think it's long enough ago now, safe to remove the "starting with"
```suggestion
> :warning: prior to version 1.0.0, `opentelemetry-javaagent-all.jar`
```

docs/manual-instrumentation.md
```diff
@@ -0,0 +1,115 @@
...
+
+// ...
+
+Span span = Span.current();
+span.setAttribute(..., ...);
+```
+
+# Creating spans around methods with `@WithSpan`
+
+Another common situation is to capture a span around an existing first-party code method. The
```
trask, line 73:
```suggestion
Another common situation is to capture a span corresponding to one of your methods. The
```

docs/manual-instrumentation.md
```diff
@@ -0,0 +1,115 @@
...
+| otel.instrumentation.methods.include |                                 | Add instrumentation for specific methods in lieu of `@WithSpan`.
+Format is "my.package.MyClass1[method1,method2];my.package.MyClass2[method3]" |
+
+# Creating spans manually with a Tracer
+
+OpenTelemetry offers a tracer to easily enable custom instrumentation throughout your application.
+See the [OpenTelemetry Java
+QuickStart](https://github.com/open-telemetry/opentelemetry-java/blob/master/QUICKSTART.md#tracing)
+for an example of how to configure the tracer and use the Tracer, Scope and Span interfaces to
+instrument your application.
```
trask, line 115: I think we want to recommend getting the tracer from GlobalOpenTelemetry (and not configuring the tracer)

docs/supported-libraries.md
```diff
@@ -0,0 +1,138 @@
+
+# Supported libraries, frameworks, and application servers
+
+We automatically instrument and support a huge number of libraries, frameworks,
+and application servers... right out of the box!
+
+Don't see your favorite tool listed here?  Consider [filing an issue](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues),
+or [contributing](../CONTRIBUTING.md).
```
trask, line 8: 😬 😂 👍

README.md
```diff
@@ -82,357 +90,34 @@ java -javaagent:path/to/opentelemetry-javaagent-all.jar \
 Learn how to add custom instrumentation in the [Manually Instrumenting](#manually-instrumenting)
 section.
```
trask, line 91: may not need this anymore, since that section is near the top now
```suggestion
```

docs/supported-libraries.md
```diff
@@ -0,0 +1,138 @@
...
+you can enable it by adding the following system property:
+`-Dotel.instrumentation.grizzly.enabled=true`
+
+## Suppressing specific auto-instrumentation
+
+See [suppressing specific auto-instrumentation](suppressing-instrumentation.md)
+
+## Logger MDC (Mapped Diagnostic Context) auto-instrumentation
+
+See [Logger MDC auto-instrumentation](logger-mdc-instrumentation.md)
```
trask, line 138: I think "suppressing instrumentation" may be better in agent-config.md

And Logger MDC might deserve it's own section on README, since it's different from everything else, as it's about configuring your _application_ instead of configuring the _agent_.

trask (approved):
Thanks for tackling this, it is much needed ❤️

docs/manual-instrumentation.md
breedx-splk: If it's phrased "prior to version 1.0.0" and we hit 1.1 doesn't that mean that section isn't applicable any longer?   Maybe we should say "starting with 1.0.0", or did I completely misread it?

breedx-splk: Oh, nm, I misread it.  I'll take this suggestion.

docs/supported-libraries.md
breedx-splk: Spirit of supporting encouragement.  😁

docs/manual-instrumentation.md
breedx-splk: Ok, good call out.  I changed the wording in here to try and make this better and avoid suggesting the old way.

docs/supported-libraries.md
breedx-splk: Yup, love both of these.  Will change now.

docs/manual-instrumentation.md
```diff
@@ -0,0 +1,121 @@
...
+will be `<className>.<methodName>`.
+
+## Suppressing `@WithSpan` instrumentation
+
+Suppressing `@WithSpan` is useful if you have code that is over-instrumented using `@WithSpan`
+and you want to suppress some of them without modifying the code.
+
+| System property                 | Environment variable            | Purpose                                                                                                                                  |
+|---------------------------------|---------------------------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| trace.annotated.methods.exclude | TRACE_ANNOTATED_METHODS_EXCLUDE | Suppress `@WithSpan` instrumentation for specific methods.
```
mateuszrzeszutek, line 98: I think that this property has been changed a while ago to `otel.instrumentation.opentelemetry-annotations.exclude-methods`:

```suggestion
| otel.instrumentation.opentelemetry-annotations.exclude-methods | OTEL_INSTRUMENTATION_OPENTELEMETRY_ANNOTATIONS_EXCLUDE_METHODS | Suppress `@WithSpan` instrumentation for specific methods.
```

mateuszrzeszutek (approved):

docs/manual-instrumentation.md
trask: I can confirm this very very long property name 😂.

I will accept the suggestion and merge the PR.

Thanks again @breedx-splk!

Issue #2129, opened 2021-01-27
Clean up BaseTracer a bit


mateuszrzeszutek:
I think that we can make a few improvements to the `BaseTracer` class:

* Make `startSpan()` return `Context` and `end()` accept `Context`
* Encapsulate `debugContextLeak()` logic in the `ContextPropagationDebug` class
* Remove `getCurrentSpan()` (it's hardly used at all)
* Remove `startScope()` and just use `context.makeCurrent()` instead

PR #2232, opened 2021-02-09
Make BaseTracer#startSpan() return Context

mateuszrzeszutek:
This resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2129

There are still lots of tracers that use `Span`s instead of `Context`s, but at least the classes in `instrumentation-api` all use `Context` now - all the other ones can be refactored a bit later. Also, while working on this I thought that maybe we should consider adding a recommendation somewhere in the Javadoc to only use `Context` in tracer public API, and avoid using `Span`s. And add as many attributes as possible to the `SpanBuilder`, not the `Span` - to make them visible in `SpanProcessor#onStart()`. WDYT?


instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelEventNotifier.java
```diff
@@ -59,7 +59,7 @@ private void onExchangeSending(ExchangeSendingEvent ese) {
 
     String name =
         sd.getOperationName(ese.getExchange(), ese.getEndpoint(), CamelDirection.OUTBOUND);
-    Span span = CamelTracer.TRACER.startSpan(name, sd.getInitiatorSpanKind());
+    Span span = Span.fromContext(CamelTracer.TRACER.startSpan(name, sd.getInitiatorSpanKind()));
     sd.pre(span, ese.getExchange(), ese.getEndpoint(), CamelDirection.OUTBOUND);
     ActiveSpanManager.activate(ese.getExchange(), span, sd.getInitiatorSpanKind());
     CamelPropagationUtil.injectParent(Context.current(), ese.getExchange().getIn().getHeaders());
```
trask, line 65: would be good to use the context returned from `startSpan` in activate(), e.g.
```suggestion
    Context context = CamelTracer.TRACER.startSpan(name, sd.getInitiatorSpanKind());
    Span span = Span.fromContext(context);
    sd.pre(span, ese.getExchange(), ese.getEndpoint(), CamelDirection.OUTBOUND);
    ActiveSpanManager.activate(ese.getExchange(), context, sd.getInitiatorSpanKind());
    CamelPropagationUtil.injectParent(context, ese.getExchange().getIn().getHeaders());
```

trask (approved):
> There are still lots of tracers that use `Span`s instead of `Context`s, but at least the classes in `instrumentation-api` all use `Context` now - all the other ones can be refactored a bit later. 

👍

> Also, while working on this I thought that maybe we should consider adding a recommendation somewhere in the Javadoc to only use `Context` in tracer public API, and avoid using `Span`s. And add as many attributes as possible to the `SpanBuilder`, not the `Span` - to make them visible in `SpanProcessor#onStart()`. WDYT?

👍 👍 👍 (`BaseTracer` class-level javadoc seems like a good spot?)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -66,18 +67,29 @@ public ContextPropagators getPropagators() {
...
+    return startSpan(spanNameForClass(clazz));
+  }
+
+  public Context startSpan(Method method) {
+    return startSpan(spanNameForMethod(method));
   }
 
-  public Span startSpan(Method method) {
-    String spanName = spanNameForMethod(method);
+  public Context startSpan(String spanName) {
```
anuraaga, line 78: Hmm one high level comment - I think we have a dream of setting more attributes on `Span.Builder` where we can, currently we never do. Will we lose that with this sort of API? I guess if there's a `protected fillAttributes(Span.Builder)` on tracers, it should be ok just want to confirm.

instrumentation/ratpack-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/ratpack/ErrorHandlerAdvice.java
```diff
@@ -17,11 +18,13 @@
   @Advice.OnMethodEnter(suppress = Throwable.class)
   public static void captureThrowable(
       @Advice.Argument(0) Context ctx, @Advice.Argument(1) Throwable throwable) {
-    Optional<Span> span = ctx.maybeGet(Span.class);
-    if (span.isPresent()) {
+    Optional<io.opentelemetry.context.Context> otelContext =
+        ctx.maybeGet(io.opentelemetry.context.Context.class);
```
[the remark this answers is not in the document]
mateuszrzeszutek: I know 😂 

I'm willing to bet a small and insignificant sum of money that if somebody created a ranking of most commonly used words in all Java code in the world "context" would be somewhere in the top 5

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
mateuszrzeszutek: I think that all those `startSpan()` methods in the base tracer are not really meant for extension -- they're more like the simplest way of creating a span. Usually you need to pass something more than a string span name, or a `Class` to extract some useful attributes after all.
I think that we should explicitly state that in the javadoc: if you want something more than a bare span, write your own `startSpan()` method.

mateuszrzeszutek:
(Rebased to get the muzzle fix)

anuraaga (approved):

PR #2234, opened 2021-02-09
Fix build deadlock

trask:
Identified via the build deadlock logs from: https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/551220496#artifacts

Similar to gradle output deadlock as we've seen before:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/ca21955bd97c4320977fd23988eb430cf26d30fa/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/context/FieldBackedProvider.java#L78-L84

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -345,7 +345,6 @@ public class AgentInstaller {
         JavaModule module,
         boolean loaded,
         DynamicType dynamicType) {
-      log.debug("Transformed {} -- {}", typeDescription.getName(), classLoader);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -344,30 +344,22 @@ public void onTransformation(
         ClassLoader classLoader,
         JavaModule module,
         boolean loaded,
-        DynamicType dynamicType) {
-      log.debug("Transformed {} -- {}", typeDescription.getName(), classLoader);
-    }
+        DynamicType dynamicType) {}
```
anuraaga, line 347: Do you think we'll later add something that adds log messages to a list or something to print them? Guess `Transformed` is a pretty useful message for debugging instrumentation load.

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
trask: hm, you're right, this is a pretty useful debug log

that's not a bad idea, under a system property, we could write to a queue, and have another thread log it (so the logging won't happen inside of the class file transform)

will give this PR another try 👍

anuraaga: Talked offline, let's follow up so our current builds (especially release build) are more stable

PR #2236, opened 2021-02-10
Bring back debug logging under class file transform

trask:


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/context/FieldBackedProvider.java
```diff
@@ -205,7 +197,7 @@ public void visitMethodInsn(
                         .equals(owner)
                     && CONTEXT_GET_METHOD.getName().equals(name)
                     && Type.getMethodDescriptor(CONTEXT_GET_METHOD).equals(descriptor)) {
-                  log.trace("Found context-store access in {}", instrumenterClass.getName());
+                  log.debug("Found context-store access in {}", instrumenterClass.getName());
```
trask, line 200: we may want to keep this logging at `trace`, but since this was primary culprit of deadlocking earlier, moving it back to debug to verify deadlocking is gone now

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeLogger.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Debug logging that is performed under class file transform needs to use this class, because
+ * gradle deadlocks sporadically under the following sequence:
+ * <li>Gradle triggers a class to load while it is holding a lock
+ * <li>Class file transform occurs (under this lock) and the agent writes to System.out
+ * <li>(Because gradle hijacks System.out), gradle is called from inside of the class file transform
+ * <li>Gradle tries to grab a different lock during it's implementation of System.out
+ */
+public class SafeLogger {
```
anuraaga, line 21:
```suggestion
public class GradleSafeLogger {
```

maybe

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeLogger.java
```diff
@@ -0,0 +1,103 @@
...
+
+  private static final boolean ENABLE_GRADLE_SAFE_LOGGING =
+      Boolean.getBoolean("otel.internal.enableGradleSafeLogging");
+
+  private static final BlockingQueue<LogMessage> logMessageQueue;
+
+  static {
+    if (ENABLE_GRADLE_SAFE_LOGGING) {
+      logMessageQueue = new ArrayBlockingQueue<>(100);
```
anuraaga, line 30: `100` seems a bit low, but maybe it's enough if the build's passing :)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeLogger.java
```diff
@@ -0,0 +1,103 @@
...
+    return new SafeLogger(LoggerFactory.getLogger(clazz));
+  }
+
+  private SafeLogger(Logger logger) {
+    this.logger = logger;
+  }
+
+  public void debug(String format, Object arg) {
+    if (logMessageQueue != null) {
+      logMessageQueue.add(new LogMessage(logger, format, arg));
```
anuraaga, line 52: `offer` instead of `add` so it doesn't crash when full?

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeLogger.java
iNikem, line 21: I feel uneasy with having build system specific class in distribution, where it is not actually needed.

trask: I like this name, it does look a little weird in the places where it's used tho, related to @iNikem's point about any mention of gradle being weird, will sleep on it 😄

trask: Ya, but I don't have any better ideas currently.

@anuraaga and I have both noticed an uptick in the build deadlocking lately, so would like to do something.

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeLogger.java
trask: @iNikem I renamed it to `TransformSafeLogger`, and renamed the property to `otel.internal.enableTransformSafeLogging`, so it now feels a more generalized, not just a gradle hack, which seems better, and in theory this problem could occur in production systems that also hijack System.out.

Please re-review at your leisure.

iNikem (approved):

anuraaga (approved):

PR #2241, opened 2021-02-10
Add JUnit 5 extensions for instrumentation testing

mateuszrzeszutek:
I think this PR resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1644 - testing-common is far from ideal, but at least it's now usable bu the vendor distributions.

I'll create a separate after-ga issue listing all the stuff that needs to be done to make this lib much nicer: reusing otel sdk testing assertions, changing package structure, moving more utils from groovy to java etc.

### DIFF
--- a/testing-common/src/test/java/io/opentelemetry/instrumentation/testing/junit/LibraryInstrumentationExtensionTest.java
+++ b/testing-common/src/test/java/io/opentelemetry/instrumentation/testing/junit/LibraryInstrumentationExtensionTest.java
@@ -11,7 +11,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import io.opentelemetry.instrumentation.test.utils.TraceUtils;
 import io.opentelemetry.sdk.trace.data.SpanData;
 import java.util.List;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.RepeatedTest;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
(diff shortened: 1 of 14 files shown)


iNikem (approved):

testing-common/src/test/java/io/opentelemetry/instrumentation/testing/junit/LibraryInstrumentationExtensionTest.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+class LibraryInstrumentationExtensionTest {
+  @RegisterExtension
+  static final LibraryInstrumentationExtension instrumentation =
+      LibraryInstrumentationExtension.create();
+
+  @Test
```
anuraaga, line 22: Does `@RepeatedTest` help to verify the exports get cleared in between?

anuraaga (approved):

testing-common/src/test/java/io/opentelemetry/instrumentation/testing/junit/LibraryInstrumentationExtensionTest.java
mateuszrzeszutek: That's a great idea, done. Thanks!

PR #2242, opened 2021-02-10
Test jax-rs on wildfly

laurit:
The purpose of this test is to demonstrate that our jax-rs intrumentation instrumentes bytecode proxies generated for jax-rs resources that are ejbs or cdi beans which is undesirable.
I modified `GlobalIgnoresMatcher` to exclude classes with `$$` in their name (except `$$anon$`). This is based on the assumption that while `$$` is perfectly legal in a java class name it is really only used by bytcode generators.


iNikem (approved):

mateuszrzeszutek (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/src/test/groovy/WildflyRestTest.groovy
```diff
@@ -0,0 +1,70 @@
...
+      .readEntity(String)
+
+    then:
+    result == "hello"
+
+    and:
+    assertTraces(1) {
+      trace(0, 2) {
+        span(0) {
+          name ~/.*\${path}/
```
trask, line 55: only curious, why don't we know the context path?

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/jaxrs-2.0-wildfly-testing.gradle
```diff
@@ -0,0 +1,69 @@
...
+  testServer "org.wildfly:wildfly-dist:18.0.0.Final@zip"
+}
+
+// extract wildfly dist, path is used from arquillian.xml
+task setupServer(type: Copy) {
+  from zipTree(configurations.testServer.singleFile)
+  into file('build/server/')
+}
+
+// logback-classic contains /META-INF/services/javax.servlet.ServletContainerInitializer
```
anuraaga, line 32: This looks familiar - @trask I see to recall we had a better way around this?

anuraaga (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/jaxrs-2.0-wildfly-testing.gradle
```diff
@@ -0,0 +1,69 @@
...
+}
+
+test.dependsOn modifyLogbackJar, setupServer
+
+test {
+  doFirst {
+    // --add-modules is unrecognized on jdk8, ignore it instead of failing
+    jvmArgs "-XX:+IgnoreUnrecognizedVMOptions"
+    // needed for java 11
+    jvmArgs "--add-modules=java.se"
```
trask, line 58: this seems odd, can you add a bit more explanation to the comment?

trask (approved):
given the complex configuration required, would this make more sense as a smoke test?

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/jaxrs-2.0-wildfly-testing.gradle
trask: yes, it does look familiar 😄

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1635#discussion_r523703382

if this becomes a common problem, we could have a test-logging module that does the exclusion, and instrumentation.gradle can depend on the test-logging module, instead of pulling in logback directly

anuraaga: Or switch to log4j :)

trask: ya, i'm not even sure what we are logging directly from test code (separate from the agent's internal logging)

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/src/test/groovy/WildflyRestTest.groovy
laurit: Because the deployed archive name is an uuid. I guess I could add a jboss-web.xml and try to set context path to something predictable, but doesn't seem like it is worth the effort so I just changed it to extract context path from the uri that arquillian gives us and used that.
As for using smoke tests, the current setup where you need to first push test app and only after it gets merged can push the test because you need the image id from ghcr.io is annoying. The mess with logback-classic and jvmArgs wouldn't be needed if instead of embedded container we would use managed container where arquillian starts the server itself. Then we would need to somehow add proper arguments for agent to arquillian.xml but as test and server are not in the same process then would also need to somehow get spans from server to test which would complicate things.

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/jaxrs-2.0-wildfly-testing.gradle
laurit: https://stackoverflow.com/questions/52852192/how-to-run-wildfly-14-with-java-11
wildfly needs java.se module for idk what. I guess in later versions they just add this in their startup scripts, but as we are using embedded wildfly we need to make sure that the environment is suitable for it ourself

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-wildfly-testing/src/test/groovy/WildflyRestTest.groovy
trask: I created #2274, hopefully we can make adding new smoke tests less annoying in the future

PR #2249, opened 2021-02-10
Start smoke test backend and collector only once

laurit:
Start backend and collector once for smoke tests and reuse it for all tests that run on the same worker. This should make smoke test runs a bit faster.

### DIFF
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
@@ -38,7 +38,8 @@ abstract class SmokeTest extends Specification {
   protected static final OkHttpClient CLIENT = OkHttpUtils.client()
 
   @Shared
-  private Network network = Network.newNetwork()
+  private Backend backend = Backend.getInstance()
+
   @Shared
   protected String agentPath = System.getProperty("io.opentelemetry.smoketest.agent.shadowJar.path")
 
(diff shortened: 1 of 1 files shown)


iNikem (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -243,4 +219,59 @@ abstract class SmokeTest extends Specification {
...
+
+    int getMappedPort(int originalPort) {
+      return backend.getMappedPort(originalPort)
+    }
+
+    def cleanup() {
+    }
+
+    // currently unused
+    def stop() {
```
mateuszrzeszutek, line 271: Nit: consider adding a shutdown hook for stoping those docker containers

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -243,4 +219,59 @@ abstract class SmokeTest extends Specification {
...
+    static Backend getInstance() {
+      return INSTANCE
+    }
+
+    def setup() {
+      // we start backend & collector once for all tests
+      if (started) {
+        return
+      }
+      started = true
```
mateuszrzeszutek, line 241: Do smoke tests run concurrently? If so, consider using an `AtomicBoolean` instead (and making the method synchronized)

mateuszrzeszutek (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
laurit: My understanding is that tests run serially within a process, there can be multiple process, but they don't share anything. It is possible to make junit run tests in parallel with multiple threads but we don't use it.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
laurit: shutdown hook added

trask (approved):

PR #2250, opened 2021-02-10
fixing possible NPE in Apache Camel instrumentation

### LINKED ISSUE #2243: Camel instrumentation can suffer from NPE in AWS parent propagation scenario
**Describe the bug**
If SQS message is produced without propagation header (AWSTraceHeader) set, consuming by Apache Camel instrumentation can cause NPE.

**Steps to reproduce**
- send message without propagation (eg by not instrumented code)
- consume with Camel
- parent extraction (io.opentelemetry.javaagent.instrumentation.apachecamel.CamelPropagationUtil#extractAwsPropagationParent) can suffer from NPE due to NULL value

- propagation should work

**What did you see instead?**

- master

**Environment**
NA

**Additional context**
NA

kubawach:
Fixes #2243

Chance of NPE in case of SQS parent propagation from a message without AWSTraceHeader (eg not instrumented producer).

### DIFF
--- a/instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
+++ b/instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
@@ -59,11 +59,8 @@ final class CamelPropagationUtil {
 
     @Override
     public String get(Map<String, Object> map, String key) {
-      return (hasValidValue(map, key) ? map.get(key).toString() : null);
-    }
-
-    private boolean hasValidValue(Map<String, Object> map, String key) {
-      return (map.containsKey(key) && (map.get(key) != null));
+      Object value = map.get(key);
+      return (value == null ? null : value.toString());
     }
   }
 
--- a/instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
+++ b/instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
@@ -58,8 +58,12 @@ final class CamelPropagationUtil {
     }
 
     @Override
-    public String get(Map<String, Object> map, String s) {
-      return (map.containsKey(s) ? map.get(s).toString() : null);
+    public String get(Map<String, Object> map, String key) {
+      return (hasValidValue(map, key) ? map.get(key).toString() : null);
+    }
+
+    private boolean hasValidValue(Map<String, Object> map, String key) {
+      return (map.containsKey(key) && (map.get(key) != null));
     }
   }
 
--- a/instrumentation/apache-camel-2.20/javaagent-unittests/apache-camel-2.20-javaagent-unittests.gradle
+++ b/instrumentation/apache-camel-2.20/javaagent-unittests/apache-camel-2.20-javaagent-unittests.gradle
@@ -4,7 +4,9 @@ dependencies {
   testImplementation project(':instrumentation:apache-camel-2.20:javaagent')
   testImplementation group: 'org.apache.camel', name: 'camel-core', version: '2.20.1'
   testImplementation group: 'org.apache.camel', name: 'camel-aws', version: '2.20.1'
+  testImplementation group: 'org.apache.camel', name: 'camel-http', version: '2.20.1'
 
+  testImplementation deps.opentelemetryTraceProps
   testImplementation deps.opentelemetryExtAws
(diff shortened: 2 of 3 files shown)


instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
```diff
@@ -58,8 +58,12 @@ static void injectParent(Context context, final Map<String, Object> exchangeHead
...
 
     @Override
-    public String get(Map<String, Object> map, String s) {
-      return (map.containsKey(s) ? map.get(s).toString() : null);
+    public String get(Map<String, Object> map, String key) {
+      return (hasValidValue(map, key) ? map.get(key).toString() : null);
+    }
+
+    private boolean hasValidValue(Map<String, Object> map, String key) {
+      return (map.containsKey(key) && (map.get(key) != null));
```
breedx-splk, line 66: I think that calling `containsKey()` prior to `get()` is clunky.  This `get()` is also repeated in the caller just a few lines above.  I suggest combining these methods into a simple 2-liner:
```
Object value = map.get(key);
return value == null : null : value.toString();
```

breedx-splk (changes requested):
This looks good to me. Had one small suggestion around readability.  👍

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
kubawach: Makes sense :)

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
```diff
@@ -58,8 +58,9 @@ static void injectParent(Context context, final Map<String, Object> exchangeHead
     }
 
     @Override
-    public String get(Map<String, Object> map, String s) {
-      return (map.containsKey(s) ? map.get(s).toString() : null);
+    public String get(Map<String, Object> map, String key) {
+      Object value = map.get(key);
+      return (value == null ? null : map.get(key).toString());
```
breedx-splk, line 63: You still have the double get!  :). You can just say `value.toString()` here.

kubawach: Too fast, too furious 🙈 But in reality I probably should stop working after 6 PM 😬

breedx-splk (approved):
Looks good to me, thanks!

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
```diff
@@ -58,8 +58,9 @@ static void injectParent(Context context, final Map<String, Object> exchangeHead
     }
 
     @Override
-    public String get(Map<String, Object> map, String s) {
-      return (map.containsKey(s) ? map.get(s).toString() : null);
+    public String get(Map<String, Object> map, String key) {
+      Object value = map.get(key);
+      return (value == null ? null : value.toString());
```
anuraaga, line 63: I agree this new style is cleaner but my eyes aren't being able to see the NPE fix - is the map bring mutated between containsKey and get previously?

anuraaga (approved):

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
iNikem: I understand that map has an explicit value `null` assigned to the given key.

iNikem (approved):

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/CamelPropagationUtil.java
kubawach: Nikita is right, previously we only checked if there is a key/value mapping and dereferenced value if so.

anuraaga: Got it, thanks!

PR #2251, opened 2021-02-10
Remove deps from javaagent

trask:
e.g. jackson is currently (unnecessarily) bundled under `inst/`

we already have netty under `inst/` due to the otlp/jaeger exporters, but is unnecessarily included if someone is not using netty in their javaagent exporter

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
+++ b/instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
@@ -25,11 +25,11 @@ dependencies {
   // in public API.
   library group: 'com.amazonaws', name: 'aws-lambda-java-events', version: '2.2.1'
 
-  implementation(
+  compileOnly(
     'com.fasterxml.jackson.core:jackson-databind:2.10.5.1',
     'com.fasterxml.jackson.module:jackson-module-afterburner:2.9.10',
     'commons-io:commons-io:2.2')
-  implementation deps.slf4j
+  compileOnly deps.slf4j
 
   implementation deps.opentelemetryExtAws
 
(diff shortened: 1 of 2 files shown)


instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
```diff
@@ -25,17 +25,22 @@ dependencies {
...
     'commons-io:commons-io:2.2')
-  implementation deps.slf4j
+  compileOnly deps.slf4j
 
   implementation deps.opentelemetryExtAws
 
   // 1.2.0 allows to get the function ARN
   testLibrary group: 'com.amazonaws', name: 'aws-lambda-java-core', version: '1.2.0'
 
+  testImplementation(
```
anuraaga, line 39: This means we will rely on the jackson brought in by the lambda library right? It's a bit unsafe because I think jackson is just an implementation detail and could change to awsjson or something in the future. But the probability is low and muzzle / test latest deps would catch it I think so it seems ok.

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
```diff
@@ -25,17 +25,22 @@ dependencies {
...
+    'com.fasterxml.jackson.core:jackson-databind:2.10.5.1',
+    'com.fasterxml.jackson.module:jackson-module-afterburner:2.9.10',
```
anuraaga, line 41: Let's go ahead and remove afterburner not useful for tests.

anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

iNikem:
@trask please rebase to fix muzzle failures

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
trask: 👍 done in #2271

instrumentation/aws-lambda-1.0/library/aws-lambda-1.0-library.gradle
trask: we weren't injecting (shaded) jackson as `additionalHelperClassNames`, so we were already relying on jackson being present in the user's class loader

Issue #2157, opened 2021-02-01
Instrumentation for Apache HttpClient5


tafli:
It seems that only Apache HttpClient 2/4 are supported right now. Are there any plans for supporting v5?

My project is using Apache HttpClient5 and I wanted to use the auto instrumentation for it.

anuraaga:
Fixed in #2254

anuragagarwal561994:
@anuraaga the current instrumentation provides the metrics for classic httpclient5 instrumentation, do we have a plan to also provide the instrumentation for the http async client.

I can help contribute, but will need a little guidance for the same.

trask:
hi @anuragagarwal561994! I would recommend opening a new issue for this. I don't believe anyone is working on httpclient5 async instrumentation, that would be a great addition if you want to contribute!

PR #2254, opened 2021-02-11
Add support for apache httpclient5

laurit:
Adds support for apache httpclient5 classic. Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2157 though the issue does not specify whether they are interested in class or async httpclient.


anuraaga:
Hi @laurit - can we start with library instrumentation for this? We should generally use the library's native support for instrumentation instead of intercepting execute where it's possible - so if we start with library instrumentation registering interceptors, I think we'll know how to then wire that into the javaagent code.

https://hc.apache.org/httpcomponents-client-5.0.x/httpclient5/xref-test/org/apache/hc/client5/http/examples/ClientInterceptors.html

laurit:
@anuraaga This is an adaption of httpclient-4.0 instrumentation and works exactly the same way. The approach you suggested would work when http client is built by `HttpClientBuilder`. `HttpClients.createMinimal` https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClients.java does not use `HttpClientBuilder` and might need a different approach. I won't have time to work on this in the near future. If you believe this approach is unacceptable then let me know and I'll close this pr.

iNikem (approved):
If this is exact adaption of the instrumentation for 4.0, then I am Ok with this. But I somewhat agree with @anuraaga sentiment, so will let him decide if we accept this without library instrumentation.

anuraaga:
I do want to discourage agent-only PRs in general to not box out non-agent users, but at the same time I feel very bad to delete working code. If @laurit doesn't have time for it though that's totally fine and I would recommend merging if this is just copy paste from our other instrumentation with package rename, let's not waste it.

FYI the biggest thing on my mind is it's 2021 - if a new HTTP library is released without a proper instrumentation hook (which may be the case given the comment about `createMinimal`) we probably need to get in touch with the authors of the code 🤷‍♂️

anuraaga (approved):
Did a skim and I'm ok with this. @laurit just to be clear thanks for doing this, regardless of where this ends up long term it's for our users. I really appreciate it!

instrumentation/apache-httpclient/apache-httpclient-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v5_0/ApacheHttpClientInstrumentationModule.java
```diff
@@ -0,0 +1,310 @@
...
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void methodEnter(
+        @Advice.Argument(0) ClassicHttpRequest request,
+        @Advice.Argument(value = 1, typing = Assigner.Typing.DYNAMIC, readOnly = false) Object arg1,
+        @Advice.Argument(
+                value = 2,
+                optional = true,
+                typing = Assigner.Typing.DYNAMIC,
+                readOnly = false)
+            Object arg2,
```
trask, line 194: instead of dynamic typing, can you split this into two advice classes?

instrumentation/apache-httpclient/apache-httpclient-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v5_0/ApacheHttpClientInstrumentationModule.java
```diff
@@ -0,0 +1,310 @@
...
+    public static void methodEnter(
+        @Advice.Argument(0) HttpHost host,
+        @Advice.Argument(1) ClassicHttpRequest request,
+        @Advice.Argument(value = 2, typing = Assigner.Typing.DYNAMIC, readOnly = false) Object arg2,
+        @Advice.Argument(
+                value = 3,
+                optional = true,
+                typing = Assigner.Typing.DYNAMIC,
+                readOnly = false)
+            Object arg3,
```
trask, line 274: same here

trask (commented):
can you also port #2276 to this new instrumentation?

instrumentation/apache-httpclient/apache-httpclient-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v5_0/WrappingStatusSettingResponseHandler.java
```diff
@@ -0,0 +1,36 @@
...
+      Context context, Context parentContext, HttpClientResponseHandler<T> handler) {
+    this.context = context;
+    this.parentContext = parentContext;
+    this.handler = handler;
+  }
+
+  @Override
+  public T handleResponse(ClassicHttpResponse response) throws IOException, HttpException {
+    tracer().end(context, response);
+    try (Scope ignored = parentContext.makeCurrent()) {
```
trask, line 32: comment added to #2276 just before merging (if this is still the same in httpclient5?):
```suggestion
    // ending the span before executing the callback handler (and scoping the callback handler to
    // the parent context), even though we are inside of a synchronous http client callback
    // underneath HttpClient.execute(..), in order to not attribute other CLIENT span timings that
    // may be performed in the callback handler to the http client span (and so we don't end up with
    // nested CLIENT spans, which we currently suppress)
    try (Scope ignored = parentContext.makeCurrent()) {
```

trask (approved):
👍

PR #2262, opened 2021-02-12
AWS SDK instrumentation - DynamoDB attributes

### LINKED ISSUE #1198: AWS instrumentation - DynamoDB statements

kubawach:
Closes #1198

Currently as PoC, after discussions concerning the previous attempt (custom "query" creation per DynamoDb operation).

Goals:
- automate creation of attributes
- map https://github.com/anuraaga/aws-opentelemetry-java-instrumentation/blob/master/docs/aws-sdk-semantic-conventions.md 

Remarks:
- PoC, so may (and should) po polished, optimised and tested before releasing
- if the approach is valid, will add all operations as per AWS SDK sem conventions

@anuraaga @trask please have a look - I'm primarily looking for validation of the presented approach :)


anuraaga:
Is the timing coincidentally the same as reviving my spec pr? Guessing not but interestingly close :)

https://github.com/open-telemetry/opentelemetry-specification/pull/1422

Sorry I didn't get to look at the code yet at all but if you can compare with that and see any possible improvements on both sides, or otherwise confirm they aren't really related, that would be great and will check this out more by Monday

kubawach:
@anuraaga pure coincidence, but out of exactly the same reason - I was busy end of last year but got some extra time now ;) Proposed PR aims to provide implementation for setting attributes, without defining them - I planned to use your drafted sem convs :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/DbRequestDecorator.java
```diff
@@ -26,5 +41,58 @@ public void decorate(Span span, SdkRequest sdkRequest, ExecutionAttributes attri
...
+    }
+    if (target instanceof List) {
+      return ((List<Object>) target).stream().map(this::serialize).collect(Collectors.joining());
+    }
+    // simple type
+    return target.toString();
+  }
+
+  @Nullable
+  private Object next(Object current, String fieldName) {
```
anuraaga, line 89: I believe it's possible to extract request parameters without reflection, can you check it?

https://github.com/aws/aws-xray-sdk-java/blob/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/java/com/amazonaws/xray/interceptors/TracingInterceptor.java#L148

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/DynamoDbRequest.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package io.opentelemetry.instrumentation.awssdk.v2_2;
+
+import static io.opentelemetry.instrumentation.awssdk.v2_2.FieldMapping.of;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.awssdk.core.SdkRequest;
+
+public enum DynamoDbRequest {
```
anuraaga, line 13: While for an initial implementation it wouldn't be required, but ideally we generate this code from the YAML file itself, similar to how we generate the attribute constants

https://github.com/open-telemetry/opentelemetry-java/tree/main/buildscripts/semantic-convention

Do you think it could work?

For reference, aws xray sdk java currently uses a JSON file shared among the SDKs to define the mapping, but the OTel yaml should be able to serve the same purpose I believe.

https://github.com/aws/aws-xray-sdk-java/tree/master/aws-xray-recorder-sdk-aws-sdk-v2/src/main/resources/com/amazonaws/xray/interceptors

anuraaga: Ah - it does mean reverse parsing "natural language" like "The value of the ... request parameter". I meant to be consistent for parsability, but it might not be, and it's definitely fragile. Hopefully we can find a structured way to model those rules though in that yaml.

kubawach: Makes perfect sense, manual converstion from spec into Java code just does not sound right. I'll open a follow-up issue for that.

kubawach: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2291

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/DbRequestDecorator.java
kubawach: Yes, there is a statically typed way but only for SdkRequest. Other model classes also have appropriate method, but even though it's the same method it is not in the interface (SdkPojo). So I could propose an optimisation - direct call for first level properties and dynamic for nested (see changed class).

kubawach:
@anuraaga please have a look - updated the code for the initial implementation. If the direction is ok, I'll add rest of DynamoDb requests (hardcoded for now), some tests and a bit of clean up.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdkRequest.java
```diff
@@ -0,0 +1,78 @@
...
+  // generic requests
+  DynamoDbRequest(DynamoDB, "DynamoDbRequest"),
+  S3Request(S3, "S3Request"),
+  SqsRequest(SQS, "SqsRequest"),
+  KinesisRequest(Kinesis, "KinesisRequest"),
+  // specific requests
+  CreateTable(
+      DynamoDB,
+      "CreateTableRequest",
+      of("awssdk.global_secondary_indexes", "globalSecondaryIndexes"),
```
anuraaga, line 32: Thanks for adapting my initial proposal :) Can you update these with what I've set in this latest PR?

https://github.com/open-telemetry/opentelemetry-specification/pull/1422

Notably, `aws.dyanamodb.*` instead of `awssdk.*` (we can dream that the server will populate these some day right? :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
```diff
@@ -0,0 +1,83 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.core.SdkRequest;
+
+public class FieldMapper {
+
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+  static {
+    OBJECT_MAPPER.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
```
anuraaga, line 25: I don't know if it requires using internal classes, if so maybe not a good idea. But if you can take a peek on whether we can use the AWS SDK to serialize, that will be safer I guess

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
```diff
@@ -0,0 +1,83 @@
...
+      return list.stream().map(this::serialize).collect(Collectors.joining());
+    }
+    // simple type
+    return target.toString();
+  }
+
+  @Nullable
+  private Object next(Object current, String fieldName) {
+    try {
+      Method method = current.getClass().getMethod(fieldName);
```
anuraaga, line 76: We should cache this, something like `ClassValue<ConcurrentHashMap<String, Method>>`

anuraaga (commented):
Thanks for helping with this! Generally looks good

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
anuraaga: Timely issue, we're not sure why this cropped up now, but this is also presumably because of sidestepping the SDK when serializing a pojo

https://github.com/aws/aws-xray-sdk-java/issues/259

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
kubawach: Done.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
kubawach: Switched to AWS SDK marshalling. A bit hacky - pls have a look :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdkRequest.java
kubawach: Done :)

kubawach:
@anuraaga another revision (with AWS SDK marshalling). Please have a look (will add more tests! :) ).

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Nullable
+  private String serialize(SdkPojo sdkPojo) {
+    Optional<ContentStreamProvider> optional =
+        protocolMarshaller.marshall(sdkPojo).contentStreamProvider();
+    return optional
+        .map(
+            csp -> {
+              try {
+                return IoUtils.toUtf8String(csp.newStream());
```
mateuszrzeszutek, line 53: `csp.newStream()` is never closed

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
```diff
@@ -0,0 +1,87 @@
...
+    for (FieldMapping fieldMapping : request.type().fields(type)) {
+      mapFields(fieldValueProvider, fieldMapping, span);
+    }
+  }
+
+  private void mapFields(
+      Function<String, Object> fieldValueProvider, FieldMapping fieldMapping, Span span) {
+    // traverse path
+    String[] path = fieldMapping.getFields();
+    Object target = fieldValueProvider.apply(camelCase(path[0]));
```
mateuszrzeszutek, line 62: This `camelCase()` call wouldn't be needed if fields in `AwsSdkRequestType` were already camelCased, would it?

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
```diff
@@ -120,12 +132,92 @@ abstract class AbstractAws2ClientTest extends InstrumentationSpecification {
...
 
     where:
     [service, operation, method, path, requestId, builder, call] << dynamoDbRequestDataTable(DynamoDbAsyncClient.builder())
   }
 
+  def assertCreateTableRequest(path, method, requestId) {
+    assertTraces(1) {
+      trace(0, 1) {
+        span(0) {
+          name "DynamoDb.CreateTable"
```
mateuszrzeszutek, line 154: Does it make sense to use the `<db.operation> <db.name>.<table>` naming convention for DynamoDB?

instrumentation/aws-sdk/aws-sdk-2.2/library/src/test/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapperTest.java
```diff
@@ -0,0 +1,104 @@
...
+    verifyNoMoreInteractions(span);
+  }
+
+  @Test
+  public void shouldMapResponseFieldsOnly() {
+
+    // given
+    AwsSdkRequest awsSdkRequest = BatchWriteItem;
+    MethodHandleFactory methodHandleFactory = new MethodHandleFactory();
+    Serializer serializer = mock(Serializer.class);
```
mateuszrzeszutek, line 85: Consider converting `serializer` and `span` mocks to class fields, they're used in all test cases

instrumentation/aws-sdk/aws-sdk-2.2/library/aws-sdk-2.2-library.gradle
```diff
@@ -4,6 +4,11 @@ dependencies {
   implementation deps.opentelemetryExtAws
 
   library group: 'software.amazon.awssdk', name: 'aws-core', version: '2.2.0'
+  library group: 'software.amazon.awssdk', name: 'aws-json-protocol', version: '2.2.0'
 
+  implementation 'com.fasterxml.jackson.core:jackson-databind:2.11.2'
```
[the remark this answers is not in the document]
kubawach: It's already added to lambda ;) But here it's legacy - followed @anuraaga recommendation and switched serialization over to AWS SDK.

instrumentation/aws-sdk/aws-sdk-2.2/library/aws-sdk-2.2-library.gradle
```diff
@@ -4,6 +4,11 @@ dependencies {
...
   testImplementation project(':instrumentation:aws-sdk:aws-sdk-2.2:testing')
+
+  testImplementation group: 'org.assertj', name: 'assertj-core', version: '3.18.1'
```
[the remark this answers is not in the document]
kubawach: Noice. Will add both (cause, why not? :) )

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
```diff
@@ -0,0 +1,87 @@
...
+    serializer = new Serializer();
+    methodHandleFactory = new MethodHandleFactory();
+  }
+
+  FieldMapper(Serializer serializer, MethodHandleFactory methodHandleFactory) {
+    this.methodHandleFactory = methodHandleFactory;
+    this.serializer = serializer;
+  }
+
+  void mapFields(SdkRequest sdkRequest, AwsSdkRequest request, Span span) {
```
anuraaga, line 29: Since this is a `FieldMapper`, `mapToAttributes` for the method name? I've read the word a few times now and every time I have to remember what it's mapping to

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
```diff
@@ -0,0 +1,80 @@
...
+package io.opentelemetry.instrumentation.awssdk.v2_2;
+
+import java.io.IOException;
+import java.net.URI;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import software.amazon.awssdk.core.SdkPojo;
```
anuraaga, line 15: I don't see the word `internal` in this list - nice :)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
```diff
@@ -0,0 +1,80 @@
...
+class Serializer {
+
+  private final ProtocolMarshaller<SdkHttpFullRequest> protocolMarshaller;
+
+  Serializer() {
+    protocolMarshaller =
+        AwsJsonProtocolFactory.builder()
+            .clientConfiguration(
+                SdkClientConfiguration.builder()
+                    .option(SdkClientOption.ENDPOINT, URI.create("http://empty"))
```
anuraaga, line 35: Let's add a line comment describing this hack

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
```diff
@@ -0,0 +1,80 @@
...
+            })
+        .orElse(null);
+  }
+
+  private String serialize(Collection<Object> collection) {
+    return collection.stream().map(this::serialize).collect(Collectors.joining(","));
+  }
+
+  @Nullable
+  String serialize(Object target) {
```
anuraaga, line 66: Confirming target can't be null?

instrumentation/aws-sdk/aws-sdk-2.2/library/aws-sdk-2.2-library.gradle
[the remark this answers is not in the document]
kubawach: Ahhh didn't have assertj as it was added later than my branch was created. It's now here after rebase. That solves the mystery.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/test/java/io/opentelemetry/instrumentation/awssdk/v2_2/SerializerTest.java
```diff
@@ -0,0 +1,63 @@
...
+    assertThat(serialized).isEqualTo("one,two,three");
+  }
+
+  @Test
+  public void shouldSerializeMapAsKeyCollection() {
+    // given
+    Map<String, Object> map = new HashMap<>();
+    map.put("uno", Long.valueOf(1));
+    map.put("dos", new LinkedHashMap<>());
+    map.put("tres", "cuatro");
```
anuraaga, line 57: cinco cinco seis

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdkRequestType.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.List;
+import java.util.Map;
+
+enum AwsSdkRequestType {
+  S3(request("aws.bucket.name", "Bucket")),
+  SQS(request("aws.queue.url", "QueueUrl"), request("aws.queue.name", "QueueName")),
+  Kinesis(request("aws.stream.name", "StreamName")),
+  DynamoDB(
+      request("aws.table.name", "TableName"),
+      request(SemanticAttributes.DB_NAME.getKey(), "TableName"));
```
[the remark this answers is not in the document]
kubawach: Legacy. Definitely should be consistent.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdkRequestType.java
```diff
@@ -0,0 +1,31 @@
...
+
+package io.opentelemetry.instrumentation.awssdk.v2_2;
+
+import static io.opentelemetry.instrumentation.awssdk.v2_2.FieldMapping.request;
+
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.List;
+import java.util.Map;
+
+enum AwsSdkRequestType {
```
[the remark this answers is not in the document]
kubawach: For me it was more of a AWS SDK request type as contains field mappings - which belong more to a request than to a service. So in other words this model a "generic request type" rather than "a service".

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
```diff
@@ -120,12 +132,92 @@ abstract class AbstractAws2ClientTest extends InstrumentationSpecification {
...
+            "${SemanticAttributes.HTTP_FLAVOR.key}" "1.1"
+            "aws.service" "DynamoDb"
+            "aws.operation" "CreateTable"
+            "aws.agent" "java-aws-sdk"
+            "aws.requestId" "$requestId"
+            "aws.table.name" "sometable"
+            "${SemanticAttributes.DB_SYSTEM.key}" "dynamodb"
+            "${SemanticAttributes.DB_NAME.key}" "sometable"
+            "${SemanticAttributes.DB_OPERATION.key}" "CreateTable"
+            "aws.dynamodb.global_secondary_indexes" "{\"IndexName\":\"globalIndex\",\"KeySchema\":[{\"AttributeName\":\"attribute\"}],\"ProvisionedThroughput\":{\"ReadCapacityUnits\":10,\"WriteCapacityUnits\":12}}"
```
anuraaga, line 175: `global_secondary_indexes` is a list field (I think), it looks like this is doing a single value assertion?

anuraaga (commented):
Thanks a lot!

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
kubawach: So, here's the deal - in AWS model, fields are CamelCased, but getters start with lowercase (FieldName vs fieldName). First field in a path is retrieved by getFieldValue call which takes field name, rest is retrieved with getter call. So either way we're going to transform the name, assuming we want to stay consistent with naming (avoid putting first camel cased and rest lowercased).

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapping.java
```diff
@@ -0,0 +1,63 @@
...
+
+  static FieldMapping response(String attribute, String fieldPath) {
+    return new FieldMapping(Type.RESPONSE, attribute, fieldPath);
+  }
+
+  FieldMapping(Type type, String attribute, String fieldPath) {
+    this.type = type;
+    this.attribute = attribute;
+    this.fieldPath = fieldPath;
+    this.fields = fieldPath.split("\\.");
```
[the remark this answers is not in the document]
kubawach: I wanted to avoid interpretation issues - with path it's obvious that "nested" fields are required, with array it would look like a bunch of fields on the same level. In other words - decided on readability vs performance gain here.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapping.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  String getAttribute() {
+    return attribute;
+  }
+
+  String[] getFields() {
+    return fields;
+  }
```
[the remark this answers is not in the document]
kubawach: Done.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapping.java
```diff
@@ -0,0 +1,63 @@
...
+
+  Type getType() {
+    return type;
+  }
+
+  private final Type type;
+  private final String attribute;
+  private final String fieldPath;
```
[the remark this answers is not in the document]
kubawach: Yep.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapping.java
```diff
@@ -0,0 +1,63 @@
...
+  private final String[] fields;
+
+  public static final Map<Type, List<FieldMapping>> map(FieldMapping[] fieldMappings) {
```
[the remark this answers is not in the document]
kubawach: Sounds good.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
kubawach: Internally it's BAIS, which does not need to be closed (per javadoc: Closing a {@code ByteArrayInputStream} has no effect). Yet, it can be easily changed to try-with-resources to be super-super safe (for a future changes for example).

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -36,39 +32,12 @@
...
-      new ExecutionAttribute<>(TracingExecutionInterceptor.class.getName() + ".RequestType");
+  static final ExecutionAttribute<AwsSdkRequest> AWS_SDK_REQUEST_ATTRIBUTE =
+      new ExecutionAttribute<>(TracingExecutionInterceptor.class.getName() + ".AwsSdkRequest");
 
   static final String COMPONENT_NAME = "java-aws-sdk";
 
-  private static final Map<RequestType, SdkRequestDecorator> TYPE_TO_DECORATOR = mapDecorators();
-  private static final Map<String, String> FIELD_TO_ATTRIBUTE = mapFieldToAttribute();
-
-  private static Map<RequestType, SdkRequestDecorator> mapDecorators() {
```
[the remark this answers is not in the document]
kubawach: YES! :D

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: Dunno, out of scope of this PR I guess.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/FieldMapper.java
kubawach: Good idea!

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
kubawach: Would require a member of a collection (field type) to be NULL. IMO not possible, but will add a guard, just to stay safe.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/test/java/io/opentelemetry/instrumentation/awssdk/v2_2/SerializerTest.java
kubawach: Exaaactly! 😆

instrumentation/aws-sdk/aws-sdk-2.2/testing/src/main/groovy/io/opentelemetry/instrumentation/awssdk/v2_2/AbstractAws2ClientTest.groovy
kubawach: Serializer was implemented to just join elements with a delimiter, without parentheses denoting an array. Changed that and incidentally discovered that AWS SDK marshaller can't be re-used (throws NPE in such case!). So thanks for pointing that out :)

kubawach:
@mateuszrzeszutek @anuraaga please have another look, resolved all comments. Notable changed:
- additional NPE guards
- improved serialization (fixed a bug in AWS SDK usage)
- FieldMapping field paths are now fields (Camel Cased) with optional lowercasing for getter calls (so the other way around than it used to - better performance and improved readability)

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdkRequestType.java
```diff
@@ -0,0 +1,31 @@
...
+      request("aws.table.name", "TableName"),
+      request(SemanticAttributes.DB_NAME.getKey(), "TableName"));
+
+  private final Map<FieldMapping.Type, List<FieldMapping>> fields;
+
+  AwsSdkRequestType(FieldMapping... fieldMappings) {
+    this.fields = FieldMapping.groupByType(fieldMappings);
+  }
+
+  public List<FieldMapping> fields(FieldMapping.Type type) {
```
anuraaga, line 28: Think there are a few package private classes with public methods we can make package private

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
anuraaga: Let's add `@Nullable` - since we're interfacing with another SDK we need to expect the worse and consider it correct.

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
```diff
@@ -0,0 +1,92 @@
...
+        AwsJsonProtocolFactory.builder()
+            .clientConfiguration(
+                SdkClientConfiguration.builder()
+                    // AwsJsonProtocolFactory requires any URI to be present
+                    .option(SdkClientOption.ENDPOINT, URI.create("http://empty"))
+                    .build())
+            .build();
+  }
+
+  private ProtocolMarshaller<SdkHttpFullRequest> createMarshaller() {
```
anuraaga, line 43: Small nit let's move the private methods to the bottom

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/Serializer.java
```diff
@@ -0,0 +1,92 @@
...
+                return IoUtils.toUtf8String(cspIs);
+              } catch (IOException e) {
+                return null;
+              }
+            })
+        .orElse(null);
+  }
+
+  private String serialize(Collection<Object> collection) {
+    return asJsonArray(collection.stream().map(this::serialize).collect(Collectors.joining(",")));
```
anuraaga, line 66: Let's remove the `asJsonArray` method. I do like your general style of particularly readable methods even when they're short :) But `asJsonArray` only works for certain input, it isn't actually converting to an array it's just adding brackets, and `asJsonArray("foo bar")` would be malformed.

This method itself though could be renamed to `asJsonArray` if you'd like :)

anuraaga (approved):
Thanks, just nits

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/AwsSdkRequestType.java
kubawach: Definitely ;)

kubawach:
Nits done :) Awaiting merge (or an additional review ;) ).

mateuszrzeszutek (approved):

PR #2263, opened 2021-02-12
Add instrumentation for rocketmq

### LINKED ISSUE #1916: rocketMq  support
Maybe can do some enhancements to rocketMq. Is this feasible? thanks very much!

addname:
add rocketmq support  

Resolves #1916


addname:
@iNikem  Hello! I integrated "group: 'org.apache.rocketmq', name: 'rocketmq-test', version: '4.8.0'" code, and wrote the test according to the specification of Writing instrumentation, and then it was run and succeeded in local PC:

gradlew :instrumentation:rocketmq-client-4.8:javaagent:test --tests RocketMqClientTest

Configure project :benchmark
Inferred project: opentelemetry-java-instrumentation, version: 0.17.0-SNAPSHOT

BUILD SUCCESSFUL in 11s
41 actionable tasks: 41 up-to-date

But, rocketmq-test is not normal working in this PR's test. I don't know why, whether does it allow to create rocketmq store files?

PR test logs:
15:26:16.103 [AllocateMappedFileService] WARN RocketmqStore - AllocateMappedFileService interrupted, possibly by shutdown.
15:26:16.121 [TransactionalMessageCheckService] ERROR RocketmqTransaction - Check error
java.util.concurrent.RejectedExecutionException: Task org.apache.rocketmq.broker.out.BrokerOuterAPI$1@7a8308fa rejected from org.apache.rocketmq.broker.latency.BrokerFixedThreadPoolExecutor@41264e9f[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 7]

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,90 @@
...
+    return TRACER;
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.javaagent.rocketmq-client";
+  }
+
+  public Span startSpan(List<MessageExt> msgs) {
+    MessageExt msg = msgs.get(0);
```
mateuszrzeszutek, line 34: What if there are >1 messages? The rest won't be traced at all?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,90 @@
...
+            .setSpanKind(CONSUMER)
+            .setParent(extractParent(msg))
+            .setAttribute(SemanticAttributes.MESSAGING_SYSTEM, "rocketmq")
+            .setAttribute(SemanticAttributes.MESSAGING_DESTINATION, msg.getTopic())
+            .setAttribute(SemanticAttributes.MESSAGING_DESTINATION_KIND, "topic")
+            .setAttribute(SemanticAttributes.MESSAGING_OPERATION, "process")
+            .setAttribute(
+                SemanticAttributes.MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES, getStoreSize(msgs))
+            .startSpan();
+    onConsume(span, msg);
```
mateuszrzeszutek, line 47: Can you make `onConsume` accept a `SpanBuilder` instead? It is generally advisable to put as much attributes as it is possible before starting the span

instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
```diff
@@ -0,0 +1,16 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    group = "org.apache.rocketmq"
+    module = 'rocketmq-client'
+    versions = "[4.8.0,)"
```
mateuszrzeszutek, line 7: Can you add `assertInverse = true` to make sure that all other versions are not instrumented?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,90 @@
...
+
+  String getBrokerHost(MessageExt msg) {
+    if (msg.getStoreHost() != null) {
+      return msg.getStoreHost().toString().replace("/", "");
+    } else {
+      return null;
+    }
+  }
+
+  public void endConcurrentlySpan(Span span, ConsumeConcurrentlyStatus status) {
```
mateuszrzeszutek, line 83: This method doesn't actually end the span, how about renaming it to sth like `afterConsume`?

instrumentation/rocketmq-client-4.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rocketmq/RocketMqClientApiImplInstrumentation.java
```diff
@@ -0,0 +1,104 @@
...
+        @Advice.Local("otelSpan") Span span,
+        @Advice.Local("otelScope") Scope scope) {
+
+      Context parent = Java8BytecodeBridge.currentContext();
+      span = tracer().startProducerSpan(addr, msg);
+      Context newContext = parent.with(span);
+      try {
+        Java8BytecodeBridge.getGlobalPropagators()
+            .getTextMapPropagator()
+            .inject(newContext, requestHeader, SETTER);
```
mateuszrzeszutek, line 60: Hmm, wouldn't it be worth it to check `RocketMqClientConfig.isPropagationEnabled()` here as well?

settings.gradle
```diff
@@ -250,3 +253,5 @@ def setBuildFile(project) {
 rootProject.children.each {
   setBuildFile(it)
 }
+
+
```
mateuszrzeszutek, line 257: Nit: unnecessary blank lines
```suggestion
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingMessageInterceptor.java
```diff
@@ -0,0 +1,34 @@
...
+package io.opentelemetry.instrumentation.rocketmq;
+
+import io.opentelemetry.api.trace.Span;
+import java.util.List;
+import org.apache.rocketmq.client.consumer.listener.ConsumeConcurrentlyStatus;
+import org.apache.rocketmq.client.consumer.listener.ConsumeOrderlyStatus;
+import org.apache.rocketmq.common.message.Message;
+import org.apache.rocketmq.common.message.MessageExt;
+
+public class TracingMessageInterceptor {
```
mateuszrzeszutek, line 15: Hmm, this class does not seem to be that useful to anybody using the library. Do we even need a library instrumentation if the particular lib (rocketmq) cannot be easily instrumented? @anuraaga @iNikem @trask

instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetery/instrumentation/rocketmq/AbstractRocketMqClientLibraryTest.groovy
```diff
@@ -0,0 +1,161 @@
...
+import spock.lang.Shared
+import spock.lang.Unroll
+
+import static io.opentelemetry.api.trace.SpanKind.CONSUMER
+import static io.opentelemetry.api.trace.SpanKind.PRODUCER
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.basicSpan
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.runUnderTrace
+
+@Unroll
+abstract class AbstractRocketMqClientLibraryTest extends InstrumentationSpecification{
```
mateuszrzeszutek, line 26: There's no need to keep that in the common `testing` module if it only tests library instrumentation code.

instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetery/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
```diff
@@ -0,0 +1,196 @@
...
+  private String brokerAddr
+
+  @Shared
+  Message msg
+
+  @Shared
+  int consumeTime = 1000
+
+  @Shared
+  BaseConf baseConf =new BaseConf()
```
mateuszrzeszutek, line 53: Nit formatting: can you add a space after `=` everywhere where it's missing? Thanks

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/dledger/DLedgerProduceAndConsumeIT.java
```diff
@@ -0,0 +1,39 @@
...
+package base.dledger;
+
+import static base.IntegrationTestBase.nextPort;
+
+import base.BaseConf;
+import base.IntegrationTestBase;
+import org.apache.rocketmq.common.BrokerConfig;
+import org.apache.rocketmq.store.config.MessageStoreConfig;
+
+public class DLedgerProduceAndConsumeIT {
```
mateuszrzeszutek, line 15: This class is unused, can you remove it?

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    return createResult;
+  }
+
+  public static boolean initTopic(String topic, String nsAddr, String clusterName) {
+    return initTopic(topic, nsAddr, clusterName, 8);
+  }
+
+  public static void deleteFile(File file) {
```
mateuszrzeszutek, line 152: This method is unused, can you remove it?

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/BaseConf.java
```diff
@@ -0,0 +1,175 @@
...
+      String nsAddr, String topic, String producerGoup, String instanceName) {
+    RMQNormalProducer producer = new RMQNormalProducer(nsAddr, topic, producerGoup, instanceName);
+    if (debug) {
+      producer.setDebug();
+    }
+    mqClients.add(producer);
+    return producer;
+  }
+
+  public static RMQTransactionalProducer getTransactionalProducer(
```
mateuszrzeszutek, line 97: There are a couple of unused methods in this class, can you remove them?

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/BaseConf.java
```diff
@@ -0,0 +1,175 @@
...
+public class BaseConf {
+  public static final String nsAddr;
+  public static final String broker1Addr;
+  protected static String broker1Name;
+  protected static final String clusterName;
+  protected static int brokerNum;
+  protected static final NamesrvController namesrvController;
+  protected static final BrokerController brokerController1;
+  protected static final List<Object> mqClients = new ArrayList<Object>();
+  protected static boolean debug = false;
```
mateuszrzeszutek, line 36: This property does not seem to be ever set to true, can you remove it?

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/BaseConf.java
```diff
@@ -0,0 +1,175 @@
...
+    namesrvController = IntegrationTestBase.createAndStartNamesrv();
+    nsAddr = "127.0.0.1:" + namesrvController.getNettyServerConfig().getListenPort();
+    brokerController1 = IntegrationTestBase.createAndStartBroker(nsAddr);
+    clusterName = brokerController1.getBrokerConfig().getBrokerClusterName();
+    broker1Name = brokerController1.getBrokerConfig().getBrokerName();
+    broker1Addr = "127.0.0.1:" + brokerController1.getNettyServerConfig().getListenPort();
+    brokerNum = 2;
+  }
+
+  public BaseConf() {}
```
mateuszrzeszutek, line 51: All methods & fields in this class are static, can you make the constructor private (and class final) and call static methods everywhere without creating a `BaseConf` instance?

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -0,0 +1,158 @@
...
+  private static final AtomicInteger port = new AtomicInteger(40000);
+
+  public static synchronized int nextPort() {
+    return port.addAndGet(random.nextInt(10) + 10);
+  }
+
+  protected static final Random random = new Random();
+
+  public static String createBaseDir() {
+    String baseDir = System.getProperty("user.home") + SEP + "unitteststore-" + UUID.randomUUID();
```
mateuszrzeszutek, line 50: How about using `Files.createTempDirectory()` instead? Or using the JUnit `TemporaryFolder`/`TempDir` rule

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+    TMPE_FILES.add(file);
+    return baseDir;
+  }
+
+  public static NamesrvController createAndStartNamesrv() {
+    String baseDir = createBaseDir();
+    NamesrvConfig namesrvConfig = new NamesrvConfig();
+    NettyServerConfig nameServerNettyServerConfig = new NettyServerConfig();
+    namesrvConfig.setKvConfigPath(baseDir + SEP + "namesrv" + SEP + "kvConfig.json");
```
mateuszrzeszutek, line 67: Consider using `Path` and `Path.resolve()` instead

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -0,0 +1,158 @@
...
+    nameServerNettyServerConfig.setListenPort(nextPort());
+    NamesrvController namesrvController =
+        new NamesrvController(namesrvConfig, nameServerNettyServerConfig);
+    try {
+      Assert.assertTrue(namesrvController.initialize());
+      logger.info("Name Server Start:{}", nameServerNettyServerConfig.getListenPort());
+      namesrvController.start();
+    } catch (Exception e) {
+      logger.info("Name Server start failed", e);
+      System.exit(1);
```
mateuszrzeszutek, line 79: Since it's test code you can just let the exception propagate, there's no need to `exit(1)` manually

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -0,0 +1,158 @@
...
+                "topic[%s] is created failed after:%d ms",
+                topic, System.currentTimeMillis() - startTime));
+        break;
+      } else {
+        TestUtils.waitForMoment(500);
+        continue;
+      }
+    }
+
+    return createResult;
```
mateuszrzeszutek, line 145: `createResult` will always be true, in any other case an `AssertionError` is thrown

mateuszrzeszutek (commented):
Hey, thanks for contributing! I've left a few comments, please take a look at them 🙏 

Sorry for taking so long to review this PR, it somehow went under our radar 😞

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingMessageInterceptor.java
anuraaga: Rocketmq seems to have some functionality to support tracing.

https://www.github.com/apache/rocketmq/tree/master/client%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Frocketmq%2Fclient%2Ftrace

https://github.com/apache/rocketmq/wiki/RIP-6-Message-Trace

If we can use it to provide library instrumentation and have the agent code just register it that's awesome. Agree that if a library prevents no way to hook in for instrumentation agent only is fine (unavoidable).

mateuszrzeszutek: Nice, that looks promising - using `ConsumeMessageHook` & `SendMessageHook` for otel tracing might be a good idea.

addname:
@mateuszrzeszutek Thank you for your review:). I'll read your comments carefully and improve the code.

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
addname: When msgs. size( )>1, do we create one span or  several span? 
1.We create several span,but their "SemanticAttributes" vaules are the same. 
2. Create one span, whose span name is marked that this span processes multiple data, for example : test-topic process size(or batch) 10 
I think the second one is better. What's your opinion?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingMessageInterceptor.java
anuraaga: @addname Would you be able to look into using hooks to instrument in library instrumentation, and then just registering that in the javaagent instrumentation? I would probably start with a PR that only includes library instrumentation if this looks feasible.

addname: @anuraaga OK, but registering hooks in the javaagent instrumentation seems a little difficult.

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
mateuszrzeszutek: Hmm, I think that messages may have different attributes, like different message id or queue offset.
Maybe we can do it the same way AWS SQS lambda handler instrumentation does in this scenario:

* create one root span for the whole message listener (without a parent);
* for each message create a child span and add a link to the producer span (extracted from the message headers).

This way each message will get its own span containing a link to propagated producer span, and the whole consumer processing will be wrapped in a single span. WDYT? @anuraaga

mateuszrzeszutek: Oh, and you can check out `TracingSqsEventHandler`, `TracingSqsMessageHandler` and `AwsLambdaMessageTracer` for an example of this.

mateuszrzeszutek: This situation is also described in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#batch-receiving

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingMessageInterceptor.java
addname: @anuraaga @mateuszrzeszutek That's cool! I've tried and it is available use hooks to register in the iavaagent instrumentation. Later, I will reconstruct javaagent, library, and testing, totally 3 modules, and then commit them.

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqProducerTracer.java
```diff
@@ -0,0 +1,65 @@
...
+    return "io.opentelemetry.javaagent.rocketmq-client";
+  }
+
+  public Context startProducerSpan(String addr, Message msg, Context parentContext) {
+    SpanBuilder spanBuilder = spanBuilder(spanNameOnProduce(msg), PRODUCER);
+    onProduce(spanBuilder, msg, addr);
+    return withClientSpan(parentContext, spanBuilder.startSpan());
+  }
+
+  public void onCallback(Context context, SendResult sendResult) {
```
mateuszrzeszutek, line 37: This method seems to be unused, can it be removed?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingSendMessageHookImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+    context.setMqTraceContext(traceContext);
+  }
+
+  @Override
+  public void sendMessageAfter(SendMessageContext context) {
+    if (context == null || context.getMqTraceContext() == null || context.getSendResult() == null) {
+      return;
+    }
+    tracer().afterProduce((Context) context.getMqTraceContext(), context.getSendResult());
```
mateuszrzeszutek, line 44: Can you check if `context.getMqTraceContext() instanceof Context`? It should always be a `Context`, but it doesn't hurt to check it just in case.

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqProducerTracer.java
```diff
@@ -0,0 +1,65 @@
...
+    return TRACER;
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.javaagent.rocketmq-client";
+  }
+
+  public Context startProducerSpan(String addr, Message msg, Context parentContext) {
+    SpanBuilder spanBuilder = spanBuilder(spanNameOnProduce(msg), PRODUCER);
```
mateuszrzeszutek, line 32: We usually pass `Context` as the first argument to any tracer method - can you change it?

Also, please add the context as the parent of the new span:

```suggestion
  public Context startProducerSpan(String addr, Message msg, Context parentContext) {
    SpanBuilder spanBuilder =
        spanBuilder(spanNameOnProduce(msg), PRODUCER).setParent(parentContext);
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingSendMessageHookImpl.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+    Context traceContext =
+        tracer()
+            .startProducerSpan(context.getBrokerAddr(), context.getMessage(), Context.current());
+    if (RocketMqClientConfig.isPropagationEnabled()) {
+      GlobalOpenTelemetry.getPropagators()
+          .getTextMapPropagator()
+          .inject(traceContext, context.getMessage().getProperties(), SETTER);
+    }
+    context.setMqTraceContext(traceContext);
```
mateuszrzeszutek, line 36: Hmm, instead of storing just the context how about adding a simple `ContextAndScope` tuple class to this module and storing `(context, context.makeCurrent())` in the trace context object?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
```diff
@@ -0,0 +1,32 @@
...
+  @Override
+  public void consumeMessageBefore(ConsumeMessageContext context) {}
+
+  @Override
+  public void consumeMessageAfter(ConsumeMessageContext context) {
+    if (context == null || context.getMsgList() == null || context.getMsgList().isEmpty()) {
+      return;
+    }
+    Context traceContext = tracer().startSpan(Context.current(), context.getMsgList());
+    tracer().end(traceContext);
```
mateuszrzeszutek, line 30: The actual processing logic will not be included in the trace if you start and immediately close the span. How about using `ConsumeMessageContext#mqTraceContext` field to store the context and scope?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,105 @@
...
+              .setAttribute(SemanticAttributes.MESSAGING_OPERATION, "receive");
+      Context rootContext = withClientSpan(parentContext, spanBuilder.startSpan());
+      for (MessageExt message : msgs) {
+        createChildSpan(rootContext, message);
+      }
+      return rootContext;
+    }
+  }
+
+  public void createChildSpan(Context parentContext, MessageExt msg) {
```
mateuszrzeszutek, line 52: This method and all methods below can be private

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqProducerTracer.java
```diff
@@ -0,0 +1,65 @@
...
+    onProduce(spanBuilder, msg, addr);
+    return withClientSpan(parentContext, spanBuilder.startSpan());
```
mateuszrzeszutek, line 34: It's not really a client span, it's a producer span:

```suggestion
    return parentContext.with(spanBuilder.startSpan());
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,105 @@
...
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.javaagent.rocketmq-client";
+  }
+
+  public Context startSpan(Context parentContext, List<MessageExt> msgs) {
+    MessageExt msg = msgs.get(0);
+    if (msgs.size() == 1) {
+      SpanBuilder spanBuilder = startSpanBuilder(msg).setParent(extractParent(msg));
+      return withClientSpan(parentContext, spanBuilder.startSpan());
```
mateuszrzeszutek, line 36: Same as in the producer tracer: it's not a client span, it's a consumer span

```suggestion
      return parentContext.with(spanBuilder.startSpan());
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      SpanBuilder spanBuilder =
+          tracer
+              .spanBuilder(msg.getTopic() + " receive")
```
mateuszrzeszutek, line 40: Are all messages received by the same consumer guaranteed to be in the same topic?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,105 @@
...
+      }
+      return rootContext;
+    }
+  }
+
+  public void createChildSpan(Context parentContext, MessageExt msg) {
+    SpanBuilder childSpanBuilder =
+        startSpanBuilder(msg)
+            .setParent(parentContext)
+            .addLink(Span.fromContext(extractParent(msg)).getSpanContext());
```
mateuszrzeszutek, line 56: In this particular case if the propagation is turned off you shouldn't add a link - it'll end up pointing to the `parentContext` anyway.

instrumentation/rocketmq-client-4.8/library/src/test/groovy/io/opentelemetry/instrumentation/rocketmq/RocketMqClientTest.groovy
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.rocketmq.remoting.common.RemotingHelper
+import org.apache.rocketmq.test.listener.rmq.order.RMQOrderListener
+import spock.lang.Shared
+
+import static io.opentelemetry.api.trace.SpanKind.CONSUMER
+import static io.opentelemetry.api.trace.SpanKind.PRODUCER
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.basicSpan
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.runUnderTrace
+
+class RocketMqClientTest extends InstrumentationSpecification implements LibraryTestTrait {
```
mateuszrzeszutek, line 24: Can this class extend `AbstractRocketMqClientTest` and reuse the test cases? It'd be really great if we had the same set of test cases for library & javaagent instrumentations.

instrumentation/rocketmq-client-4.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rocketmq/RocketMqProducerInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl;
+import org.apache.rocketmq.client.producer.DefaultMQProducer;
+
+public class RocketMqProducerInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
```
mateuszrzeszutek, line 27: There's no need for a `classLoaderOptimization()` in this case - it is only needed when we use complex `typeMatcher`s that need to examine the byte code (e.g. extends, implementsInterface)

instrumentation/rocketmq-client-4.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rocketmq/RocketMqConsumerInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+import org.apache.rocketmq.client.consumer.DefaultMQPushConsumer;
+import org.apache.rocketmq.client.impl.consumer.DefaultMQPushConsumerImpl;
+
+public class RocketMqConsumerInstrumentation implements TypeInstrumentation {
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
```
mateuszrzeszutek, line 27: Same here, a `classLoaderOptimization()` in not really necessary for a simple `typeMatcher()`

instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetery/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
```diff
@@ -0,0 +1,190 @@
...
+          }
+        }
+      }
+      cleanup:
+      producer.shutdown()
+      consumer.shutdown()
+      BaseConf.deleteTempDir()
+    }
+  }
+}
```
mateuszrzeszutek, line 189: Can you add some test cases that verify the batch message consumer scenario?

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.rocketmq.logging.InternalLogger;
+import org.apache.rocketmq.logging.InternalLoggerFactory;
+import org.apache.rocketmq.namesrv.NamesrvController;
+import org.apache.rocketmq.remoting.netty.NettyClientConfig;
+import org.apache.rocketmq.remoting.netty.NettyServerConfig;
+import org.apache.rocketmq.store.config.MessageStoreConfig;
+import org.apache.rocketmq.test.util.MQAdmin;
+import org.junit.Assert;
+
+public class IntegrationTestBase {
```
mateuszrzeszutek, line 29: The whole rocketmq setup & start procedure seems to be pretty complicated. Do you think you can use testcontainers instead to just run a docker image? That's what we usually try to do in our tests when we need to use some external dependency/service

instrumentation/rocketmq-client-4.8/library/src/test/groovy/io/opentelemetry/instrumentation/rocketmq/RocketMqClientTest.groovy
addname: @mateuszrzeszutek If reuse javaagent instrumentations test cases，
when use "gradlew :instrumentation:rocketmq-client-4.8:library:test --tests RocketMqClientTest”to test for library instrumentation，the configuration of library.gradle doesn't work.
tasks.withType(Test) {
  jvmArgs "-Dotel.instrumentation.rocketmq-client.experimental-span-attributes=true"
}
Would you pls how to deal with this situation? 

In the process of javaagent instrumentation test,the same test configuration of instrumentation.gradle works.

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
anuraaga: In aws-lambda instrumentation we work around this by checking the topics and use "multiple sources" if they aren't all the same, maybe good to do it here too

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
anuraaga: I filed this spec issue we may be able to handle this in the SDK too https://github.com/open-telemetry/opentelemetry-specification/pull/1492

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,105 @@
...
+      spanBuilder.setAttribute("messaging.rocketmq.queue_offset", msg.getQueueOffset());
+      spanBuilder.setAttribute("messaging.rocketmq.broker_address", getBrokerHost(msg));
+    }
+  }
+
+  String spanNameOnConsume(MessageExt msg) {
+    return msg.getTopic() + " process";
+  }
+
+  String getBrokerHost(MessageExt msg) {
```
anuraaga, line 98:
```suggestion
  @Nullable
  String getBrokerHost(MessageExt msg) {
```

instrumentation/rocketmq-client-4.8/library/src/test/groovy/io/opentelemetry/instrumentation/rocketmq/RocketMqClientTest.groovy
addname: Does library instrumentation test support the configuration?
tasks.withType(Test) {
jvmArgs ...
}
I didn't find examples in this repo.

mateuszrzeszutek: Hi @addname ,
Fix for this is already on the main branch - if you merge/rebase your library tests should no longer fail on missing config.

instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetery/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
addname: Done

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
addname: Hi @mateuszrzeszutek ,
The official of testcontainers didn't provide the module of RocketMQ. It is not easy to customize testcontainers anywhere to run RocketMQ  docker image for dependency. For example, in this condition, brokerIP1 usually is suggested to be configurated to physical IP. It is easier to use current rocketmq-test library ways.

instrumentation/rocketmq-client-4.8/library/src/test/groovy/io/opentelemetry/instrumentation/rocketmq/RocketMqClientTest.groovy
addname: @mateuszrzeszutek  nice, ran successfully!

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
anuraaga: Let's leave this code as is, next SDK will handle invalid links for us

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,101 @@
...
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.List;
+import org.apache.rocketmq.common.message.MessageExt;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class RocketMqConsumerTracer extends BaseTracer {
```
anuraaga, line 20:
```suggestion
final class RocketMqConsumerTracer extends BaseTracer {
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,101 @@
...
+
+  private static final RocketMqConsumerTracer TRACER = new RocketMqConsumerTracer();
+
+  public static RocketMqConsumerTracer tracer() {
```
anuraaga, line 24: Move the static singleton to the javaagent, library instrumentation shouldn't have statics anymore

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,101 @@
...
+    return TRACER;
+  }
+
```
anuraaga, line 27:
```suggestion
  RocketMqConsumerTracer(OpenTelemetry openTelemetry) {
    super(openTelemetry);
  }
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqConsumerTracer.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.javaagent.rocketmq-client";
+  }
+
+  public Context startSpan(Context parentContext, List<MessageExt> msgs) {
```
anuraaga, line 33:
```suggestion
  Context startSpan(Context parentContext, List<MessageExt> msgs) {
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqProducerTracer.java
```diff
@@ -0,0 +1,58 @@
...
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.apache.rocketmq.client.producer.SendResult;
+import org.apache.rocketmq.common.message.Message;
+
+public class RocketMqProducerTracer extends BaseTracer {
```
anuraaga, line 18: Same comments as above

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TextMapExtractAdapter.java
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.rocketmq;
+
+import io.opentelemetry.context.propagation.TextMapGetter;
+import java.util.Map;
+
+public class TextMapExtractAdapter implements TextMapGetter<Map<String, String>> {
```
anuraaga, line 11:
```suggestion
final class TextMapExtractAdapter implements TextMapGetter<Map<String, String>> {
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TextMapInjectAdapter.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.rocketmq;
+
+import io.opentelemetry.context.propagation.TextMapSetter;
+import java.util.Map;
+
+public class TextMapInjectAdapter implements TextMapSetter<Map<String, String>> {
```
anuraaga, line 11:
```suggestion
final class TextMapInjectAdapter implements TextMapSetter<Map<String, String>> {
```

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
```diff
@@ -0,0 +1,42 @@
...
+
+package io.opentelemetry.instrumentation.rocketmq;
+
+import static io.opentelemetry.instrumentation.rocketmq.RocketMqConsumerTracer.tracer;
+
+import io.opentelemetry.context.Context;
+import org.apache.rocketmq.client.hook.ConsumeMessageContext;
+import org.apache.rocketmq.client.hook.ConsumeMessageHook;
+
+public class TracingConsumeMessageHookImpl implements ConsumeMessageHook {
```
anuraaga, line 14:
```suggestion
final class TracingConsumeMessageHookImpl implements ConsumeMessageHook {
```

anuraaga: Please add a `RocketMqTracing` entry point which can return the hooks

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingSendMessageHookImpl.java
```diff
@@ -0,0 +1,52 @@
...
+
+import static io.opentelemetry.instrumentation.rocketmq.RocketMqProducerTracer.tracer;
+import static io.opentelemetry.instrumentation.rocketmq.TextMapInjectAdapter.SETTER;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.context.Context;
+import org.apache.rocketmq.client.hook.SendMessageContext;
+import org.apache.rocketmq.client.hook.SendMessageHook;
+
+public class TracingSendMessageHookImpl implements SendMessageHook {
```
anuraaga, line 16:
```suggestion
final class TracingSendMessageHookImpl implements SendMessageHook {
```

instrumentation/rocketmq-client-4.8/library/src/test/groovy/io/opentelemetry/instrumentation/rocketmq/RocketMqClientTest.groovy
```diff
@@ -0,0 +1,27 @@
...
+import io.opentelemetery.instrumentation.rocketmq.AbstractRocketMqClientTest
+import io.opentelemetry.instrumentation.test.LibraryTestTrait
+import org.apache.rocketmq.test.listener.rmq.order.RMQOrderListener
+
+class RocketMqClientTest extends AbstractRocketMqClientTest implements LibraryTestTrait {
+
+  @Override
+  void configureMQProducer() {
+    producer=BaseConf.getProducer(BaseConf.nsAddr)
+    producer.getDefaultMQProducerImpl().registerSendMessageHook(new TracingSendMessageHookImpl())
```
anuraaga, line 18: We want the pattern to be that as much logic as possible is in the parent class, library instrumentation only registers tracing, while javaagent test usually has an empty body. I think it should be possible if you provide producer and consumer as method arguments to these abstract methods

anuraaga (commented):
Thanks a lot @addname! Recently we've settled on a pattern for library instrumentation, do you mind updating to follow it?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/grpc-1.5/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_5/GrpcTracing.java

- `RocketMqTracing` entrypoint that accepts `OpenTelemetry` for creating a default, or for creating a builder
- Use builder for configuring "propagation enabled" and "capture experimental attributes"
- No singleton in library instrumentation, move to javaagent and have javaagent code configure the builder based on the `Config`
- `*Tracer` classes only have one package-private constructor accepting `OpenTelemetry` and other configuration options
- Everything except for the entrypoint class is package-private

addname:
@anuraaga Thank you for your review，I'd love to update it.

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqTracing.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package io.opentelemetry.instrumentation.rocketmq;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.instrumentation.api.config.Config;
+import org.apache.rocketmq.client.hook.ConsumeMessageHook;
+import org.apache.rocketmq.client.hook.SendMessageHook;
+
+public final class RocketMqTracing {
```
anuraaga, line 13: Let's add javadoc to these two public classes. You can for the most part copy from `grpc`, with find-replace `Grpc` with `RocketMq`

anuraaga (commented):
Thanks, almost there

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqTracingBuilder.java
```diff
@@ -7,6 +7,7 @@
```
anuraaga, line 7: For `setPropagationEnabled`

`Sets whether the trace context should be written from producers / read from consumers for propagating through messaging.`

addname: This is currently the case, written it in TracingSendMessageHookImpl, and read it in RocketMqConsumerTracer.

anuraaga: Sorry I meant you can copy-paste that as the javadoc for the `setPropagationEnabled` method which currently doesn't have javadoc (and GitHub didn't let me comment on it previously since I was looking at the commit, not PR, oops.

addname: That's OK. :smile:

anuraaga (approved):
Thanks!

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/BaseConf.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.rocketmq.client.exception.MQClientException;
+import org.apache.rocketmq.client.producer.DefaultMQProducer;
+import org.apache.rocketmq.common.MQVersion;
+import org.apache.rocketmq.namesrv.NamesrvController;
+import org.apache.rocketmq.remoting.protocol.RemotingCommand;
+import org.apache.rocketmq.test.listener.AbstractListener;
+import org.apache.rocketmq.test.util.MQRandomUtils;
+import org.apache.rocketmq.test.util.RandomUtil;
+
+final class BaseConf {
```
mateuszrzeszutek, line 20: Can you make this class public? Right now it IntelliJ treats it as unused, which is really misleading.

mateuszrzeszutek (commented):
Thanks! I've got just a few more minor comments, but other than that it looks really good 👍

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RocketMqTracing.java
```diff
@@ -0,0 +1,70 @@
...
+  public static RocketMqTracing create(OpenTelemetry openTelemetry) {
+    return newBuilder(openTelemetry)
+        .setPropagationEnabled(
+            Config.get()
+                .getBooleanProperty("otel.instrumentation.rocketmq-client.propagation", true))
+        .setCaptureExperimentalSpanAttributes(
+            Config.get()
+                .getBooleanProperty(
+                    "otel.instrumentation.rocketmq-client.experimental-span-attributes", true))
+        .build();
```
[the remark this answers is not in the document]
addname: Hi, @mateuszrzeszutek 
This configuration seems to be invalid again, for library instrumentation testing?
tasks.withType(Test) {
jvmArgs ...
}

The current configuration is used for library instrumentation testing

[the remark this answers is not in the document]
anuraaga: Ah - you can just initialize using the builder in the library test don't need to pass in a flag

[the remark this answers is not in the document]
addname: @anuraaga  nice, but in this case the create method is unused, can it be removed? The GrpcTracing create method of grpc is only used in the library instrumentation testing.

addname:
muzzle (:instrumentation:kotlinx-coroutines:javaagent) failed, how to deal with it?

mateuszrzeszutek (approved):
Thanks!

mateuszrzeszutek:
> muzzle (:instrumentation:kotlinx-coroutines:javaagent) failed, how to deal with it?

That looks like a random failure, I'll re-run these checks for you.

trask:
Thanks @addname!

Issue #765, opened 2020-07-23
AgentTestTrait can't be used when instrumenting something initialized by ClassRule


anuraaga:
For example, Armeria provides junit integration that makes it easy to start / stop a server during a test. But if we use it, we can't test instrumentation of that server because it will be loaded before our trait, which runs bytecode instrumentation. This isn't the end of the world, since we can forward to the classrule lifecycle methods from Spock lifecycle methods, but if there's a better solution it'd be nice.

PR #2265, opened 2021-02-12
Use ClassRule in armeria tests

mateuszrzeszutek:
Closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/765

There's no longer any problem with using `@ClassRule`s after https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1643 was implemented - we no longer do any code transformations in test code (not directly at least)


trask (approved):
Nice

anuraaga (approved):
Yay!

PR #2272, opened 2021-02-13
Include status code for HttpUrlConnection exceptions

### LINKED ISSUE #2244: http-url-connection instrumentation creates on error spans without http.status_code attribute
**Describe the bug**

Spans for HTTP calls made with _HttpURLConnection_ or a client like _RestTemplate_ from Spring Boot, which uses it, have no `http.status_code` attribute on errors with response status 4xx and 5xx. If a response has status 1xx, 2xx or 3xx the attribute is present.

| **Status 1xx, 2xx and 3xx**          | **Status 4xx and 5xx**               |
| -------------------------------------------- | ----------------------------------------- |
| http-status-200 | http-status-500 |

**Steps to reproduce**

I've created a sample project to reproduce this issue at [https://github.com/rkondrashov/opentelemetry-span-without-status](https://github.com/rkondrashov/opentelemetry-span-without-status).

1. Build the project with `mvn clean package`.
2. Start the service, httpbin and jaeger using `docker-compose up -d` and wait until all components are started.
3. Call `./calls.sh` script to perform some HTTP calls to the service.
4. Take a look at traces in jaeger UI - [http://localhost:16686](http://localhost:16686).
5. Stop all components using `docker-compose down`.

Alternatively the service could be started without docker:

1. Build the project with `mvn -Dexec.skip clean package`.
2. Start the service:
```bash
java -javaagent:path/to/opentelemetry-javaagent-all.jar \
       -Dotel.traces.exporter=logging \
       -Dotel.metrics.exporter=none \
       -jar target/opentelemetry-span-without-status-1.0.0.jar \
       --httpbin.uri=https://httpbin.org/status
```
3. Call `./calls.sh` script to perform some HTTP calls to the service. 

Spans on error with response status 4xx or 5xx have the `http.status_code` attribute.

**What did you see instead?**

The attribute `http.status_code` in spans on error with response status 4xx or 5xx missing.

v0.16.0

**Environment**
Compiler: AdoptOpenJDK 11.0.9
OS: Mint 20.1

**Additional context**

The same issue with java agent v0.13.x, v0.14.x and v0.15.x.

breedx-splk:
Resolves #2244 

The `HttpUrlConnection` implementation is highly complex.  There are code paths that generate and throw an exception, which causes inconsistency around `responseCode` being populated or not (since it is both a method an a field).

There's also a neat note in the instrumentation module that says:
```
        // responseCode field is sometimes not populated.
        // We can't call getResponseCode() due to some unwanted side-effects
        // (e.g. breaks getOutputStream).
```

So this adds instrumentation around `HttpUrlConnection.getResponseCode()`.  Method enter is effectively a no-op, but on exit we get the span and add in the status code and mark it as an error if it's 400/500 level.

### DIFF
--- a/instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
+++ b/instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
@@ -19,6 +19,7 @@ import static net.bytebuddy.matcher.ElementMatchers.not;
 
 import com.google.auto.service.AutoService;
 import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.javaagent.instrumentation.api.CallDepth;
@@ -92,9 +93,8 @@ public class HttpUrlConnectionInstrumentationModule extends InstrumentationModul
         @Advice.Origin("#m") String methodName
         ) {
 
-      System.out.println("///////////// yes enter " + methodName);
-      System.out.println("/////////////     enter httpUrlState = " + httpUrlState);
       if(methodName.equals("getResponseCode")){
+        // Nothing to do here, prevent changing/impacting
         return;
       }
       callDepth = CallDepthThreadLocalMap.getCallDepth(HttpURLConnection.class);
@@ -122,7 +122,6 @@ public class HttpUrlConnectionInstrumentationModule extends InstrumentationModul
         return;
       }
 
-      System.out.println("///////////// start span in " + methodName);
       Context context = tracer().startSpan(parentContext, connection);
       httpUrlState = new HttpUrlState(context);
       storage.put(connection, httpUrlState);
@@ -140,25 +139,17 @@ public class HttpUrlConnectionInstrumentationModule extends InstrumentationModul
         @Advice.Local("otelCallDepth") CallDepth callDepth,
         @Advice.Return(typing = Assigner.Typing.DYNAMIC) Object returnValue) {
 
-      System.out.println("///////////// yes exit " + methodName);
-      System.out.println("/////////////     exit httpUrlState = " + httpUrlState);
       if(methodName.equals("getResponseCode")){
-        System.out.println("///////////// EXIT getResponseCode returnValue = "  + returnValue);
-        System.out.println("/////////////     urlState = " + httpUrlState);
-        System.out.println("/////////////     throwable = " + throwable);
-
         if(httpUrlState == null){
-          System.out.println("/////////////     )) param is null");
           ContextStore<HttpURLConnection, HttpUrlState> storage =
               InstrumentationContext.get(HttpURLConnection.class, HttpUrlState.class);
           httpUrlState = storage.get(connection);
-          if(httpUrlState != null){
-            System.out.println("/////////////     )) got one from the storage! " + httpUrlState);
-            Span span = Span.fromContext(httpUrlState.context);
-            span.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, (int)returnValue);
-          }
-          else{
-            System.out.println("/////////////     )) FAILED to get one from the storage!");
+        }
+        if(httpUrlState != null){
+          Span span = Span.fromContext(httpUrlState.context);
+          span.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, (int)returnValue);
+          if((int)returnValue >= 400){
+            span.setStatus(StatusCode.ERROR);
           }
         }
         return;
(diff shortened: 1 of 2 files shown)


instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
```diff
@@ -85,8 +89,13 @@ public static void methodEnter(
         @Advice.FieldValue("connected") boolean connected,
         @Advice.Local("otelHttpUrlState") HttpUrlState httpUrlState,
         @Advice.Local("otelScope") Scope scope,
-        @Advice.Local("otelCallDepth") CallDepth callDepth) {
+        @Advice.Local("otelCallDepth") CallDepth callDepth,
+        @Advice.Origin("#m") String methodName) {
 
+      if (methodName.equals("getResponseCode")) {
```
mateuszrzeszutek, line 95: It looks like for `getResponseCode()` instrumentation has nothing in common with the other instrumented methods -- consider extracting it to a separate advice class and just put another map entry in the `transformers()` map.

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
```diff
@@ -126,7 +135,23 @@ public static void methodExit(
...
+        ContextStore<HttpURLConnection, HttpUrlState> storage =
+            InstrumentationContext.get(HttpURLConnection.class, HttpUrlState.class);
+        httpUrlState = storage.get(connection);
+        if (httpUrlState != null) {
+          Span span = Span.fromContext(httpUrlState.context);
+          span.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, (int) returnValue);
+          if ((int) returnValue >= 400) {
+            span.setStatus(StatusCode.ERROR);
+          }
```
mateuszrzeszutek, line 150: Please use `HttpStatusConverter` - same as `HttpClientTracer` does.

mateuszrzeszutek (commented):
I also have an alternative idea: consider putting status code in `HttpUrlState` and using this new field to create a `HttpUrlResponse` - this way you won't have to set attributes yourself, you'll be able to rely on the tracer to do so.

iNikem (commented):
What @mateuszrzeszutek said.

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
breedx-splk: Yeah, if you look at the earlier commits I did have a new field in the `HttpUrlState` class, but it is only a local and isn't shared with the other advice calls for the other methods.  I will see what it looks like with a new advice.

breedx-splk: I think it's an improvement.  Thanks for the nudge.

iNikem (approved):

instrumentation/http-url-connection/javaagent/src/test/groovy/HttpUrlConnectionTest.groovy
```diff
@@ -293,6 +294,28 @@ class HttpUrlConnectionTest extends HttpClientTest {
     }
   }
 
+  def "error span"(){
+    def uri = server.address.resolve("/error")
+    when:
+    def url = uri.toURL()
+    runUnderTrace("parent") {
+      doRequest(method, uri, ["is-test-server": "false"])
```
trask, line 302: and then see below for corresponding validation
```suggestion
      doRequest(method, uri)
```

instrumentation/http-url-connection/javaagent/src/test/groovy/HttpUrlConnectionTest.groovy
```diff
@@ -293,6 +294,28 @@ class HttpUrlConnectionTest extends HttpClientTest {
...
+    }
+
+    then:
+    def expectedException = new IOException("Server returned HTTP response code: 500 for URL: $url")
+    thrown(IOException)
+    assertTraces(1) {
+      trace(0, 2 + extraClientSpans()) {
+        basicSpan(it, 0, "parent", null, expectedException)
+        clientSpan(it, 1, span(0), method, uri, 500, expectedException)
+      }
```
trask, line 312: there's no need for extraClientSpans() below, since that's not overridden in this test, but let's keep it, b/c I think we should move this test up to HttpClientTest - I'll open separate issue for that
```suggestion
      trace(0, 3 + extraClientSpans()) {
        basicSpan(it, 0, "parent", null, expectedException)
        clientSpan(it, 1, span(0), method, uri, 500, expectedException)
        serverSpan(it, 2 + extraClientSpans(), span(1 + extraClientSpans()))
      }
```

trask (approved):
👍

Btw, I hadn't noticed previously that we don't test the `/error` url from `HttpClientTest`. I'll open a separate issue to move your new `/error` test up to `HttpClientTest`.

mateuszrzeszutek (approved):

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentationModule.java
```diff
@@ -148,6 +158,24 @@ public static void methodExit(
     }
   }
 
+  public static class GetResponseCodeAdvice {
+
+    @Advice.OnMethodExit
+    public static void methodExit(
+        @Advice.This HttpURLConnection connection,
+        @Advice.Return(typing = Assigner.Typing.DYNAMIC) int returnValue) {
```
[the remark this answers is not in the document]
breedx-splk: Good catch, that was a carry-over from when they were combined.  I should have caught it when I changed the type.  :). Will fix.

PR #2276, opened 2021-02-13
Propagate context to apache httpclient async callback

trask:
Noticed this while reviewing #2254


anuraaga (approved):

anuraaga (commented):
Can we add a test?

trask:
> Can we add a test?

Good call, there was already test infra for callbacks I forgot about 👍

mateuszrzeszutek (approved):

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/WrappingStatusSettingResponseHandler.java
```diff
@@ -8,22 +8,28 @@
...
+    this.parentContext = parentContext;
     this.handler = handler;
   }
 
   @Override
   public T handleResponse(HttpResponse response) throws IOException {
     tracer().end(context, response);
-    return handler.handleResponse(response);
+    try (Scope ignored = parentContext.makeCurrent()) {
+      return handler.handleResponse(response);
```
iNikem, line 32: I don't understand it here. Why exactly do we want to scope the process of reading http response into parent context? We create span around `HttpClient.execute` method, which takes `ResponseHandler`. That `ResponseHandler` handler is called inside `execute` method. It seems much more sensible to me to call `handler.handlerResponse` inside `context` and before `tracer().end`. What do I miss?

iNikem (changes requested):

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/WrappingStatusSettingResponseHandler.java
mateuszrzeszutek: As far as I understand whatever `ResponseHandler` does is no longer the HTTP call, but user code processing the response (which may be anything).
We probably need a comment with an explanation here.

iNikem: But that user code is still executed during the call to `HttpClient.execute`. If we create span around that method, then handler's execution time should be attributed to that span.

trask: @iNikem this is the pattern we apply to async callback handlers across all CLIENT instrumentation (async http clients, async database clients).

Do you feel there's something different about this particular one, or are you thinking we should change this behavior across all instrumentation?

And if we do parent callback handlers to the CLIENT span, would spans created by those callback handlers parent the CLIENT span? (and then we get into unresolved nested CLIENT span issues)

anuraaga: Ah I think @iNikem is saying that this isn't an async callback since response handler is called inline, e.g.,

https://github.com/apache/httpcomponents-client/blob/rel/v4.0.3/httpclient/src/main/java/org/apache/http/impl/client/AbstractHttpClient.java#L735

For an async callback, we would expect `execute` to return, and handler called at some later point. But this is just a synchronous response transformation. Since we instrument `execute` I wouldn't expect to need to instrument the response handler at all IIUC.

Presumably `httpasyncclient` does have a proper async pattern though :-)

trask: I think users are going to be very confused if we include callback timing (async or sync) inside of http client timing.

They are going to see an HTTP call to Service A taking a long time, and think there's a problem with Service A, when the slowness was really coming from a callback handler that called Service B which responded slowly.

anuraaga: @trask A `ResponseHandler` which also makes a RPC call, despite it all being synchronous? That's pretty intense. I get the modeling better now though - yeah it's synchronous but is similar to async instrumentation where we time start of request to start of use callback. I've never seen a pattern where we shadow a client span context with its parent like this though, we always remount the parent context after the client context is unmounted in async calls. I guess there shouldn't be any harm but it does make context much harder to reason about 🤯

iNikem: I understand your point, @trask. Indeed, it seems correct to separate callback timing from http call timing. But at the same time it makes sense to me that a single line of source code, `client.execute`, should not be split into two spans. We do similarly confusing thing in `HttpClientResponseTracingHandler` for Netty.

But then it outgrows the scope of this PR, so let's merge it.

iNikem (approved):

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/WrappingStatusSettingResponseHandler.java
trask: > But then it outgrows the scope of this PR, so let's merge it.

👍

Added comment to the code:

```
    // ending the span before executing the callback handler (and scoping the callback handler to
    // the parent context), even though we are inside of a synchronous http client callback
    // underneath HttpClient.execute(..), in order to not attribute other CLIENT span timings that
    // may be performed in the callback handler to the http client span (and so we don't end up with
    // nested CLIENT spans, which we currently suppress)
```

PR #2277, opened 2021-02-13
Remove unnecessary dynamic typing

trask:
Noticed this while reviewing #2254

### DIFF
--- a/instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
+++ b/instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
@@ -255,12 +255,7 @@ public class ApacheHttpClientInstrumentationModule extends InstrumentationModule
     public static void methodEnter(
         @Advice.Argument(0) HttpHost host,
         @Advice.Argument(1) HttpRequest request,
-        @Advice.Argument(
-                value = 2,
-                optional = true,
-                typing = Assigner.Typing.DYNAMIC,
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
```diff
@@ -198,9 +192,7 @@ public static void methodEnter(
       scope = context.makeCurrent();
 
       // Wrap the handler so we capture the status code
-      if (handler instanceof ResponseHandler) {
-        handler = new WrappingStatusSettingResponseHandler(context, (ResponseHandler<?>) handler);
-      }
+      handler = new WrappingStatusSettingResponseHandler<>(context, handler);
```
laurit, line 195: It might be better to check `handler` for `null` before wrapping to ensure that in case of `null` handler execute would fail when it checks args passed to the method instead of failing with `NullPointerException` in `WrappingStatusSettingResponseHandler.handleResponse`

iNikem (changes requested):
Lauri has a valid point about NPE.

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientInstrumentationModule.java
trask: thanks! fixed

iNikem (approved):

PR #2283, opened 2021-02-14
Remove unnecessary shutdown hook

trask:
`CommonTaskExecutor` is only used for cleaning `WeakMapSuppliers.WeakConcurrent` maps, so no need for shutdown hook

(this PR is motivated by getting rid of an unnecessary warning message that this shutdown hook logs sometimes)


anuraaga (approved):
I didn't look at this at all to know the implications, but random thought is we could look into "direct" weak concurrent map, where it's cleaned up on each call. More predictable, and possibly better, performance than a separate thread.

mateuszrzeszutek (approved):

iNikem (approved):

PR #2288, opened 2021-02-15
Remove Hibernate from javaagent jar

mateuszrzeszutek:
I've noticed that we bundle entire Hibernate 4.3 + its dependencies in the javaagent jar - which is not likely to be needed.


anuraaga (approved):
Ow, thanks!

trask (approved):

PR #2295, opened 2021-02-16
Fix main branch build

trask:
Now with added logging from #2236 (which I should still to revert a bit of, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2236#discussion_r573459946), the retry bug hit on main branch, I had already removed retry from pr builds: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1643#issuecomment-749979883

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,16 +22,6 @@ jobs:
         with:
           job-id: jdk11
 
-      - name: Build
-        uses: nick-invision/retry@v2.4.0
-        env:
-          S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -22,16 +22,6 @@ jobs:
         with:
           job-id: jdk11
 
-      - name: Build
-        uses: nick-invision/retry@v2.4.0
-        env:
-          S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
-          S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
-        with:
-          command: ./gradlew build --stacktrace
```
iNikem, line 31: Wait, you are removing the whole build step here :)

trask: well, that would fix the build, right? 🤦‍♂️

anuraaga (approved):

iNikem (approved):

PR #2300, opened 2021-02-16
Improve library tests' stability

mateuszrzeszutek:


iNikem (approved):

instrumentation/aws-lambda-1.0/library/src/test/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequestTest.java
```diff
@@ -22,10 +22,12 @@
 import java.io.IOException;
 import java.io.InputStream;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 class ApiGatewayProxyRequestTest {
 
+  @BeforeEach
```
anuraaga, line 30: Is BeforeEach enough?

anuraaga (approved):
This feels a bit extreme, adding it almost everywhere. Though ideally these all go away when the instrumentation supports injecting the SDK so it's probably fine for now, I think we'll want to clean these up later to reason about the tests better though.

instrumentation/aws-lambda-1.0/library/src/test/java/io/opentelemetry/instrumentation/awslambda/v1_0/ApiGatewayProxyRequestTest.java
mateuszrzeszutek: That's true -- fixed. Thanks!

PR #2305, opened 2021-02-16
Move slf4j-simple initialization from bootstrap into tooling

trask:
In general, we want bootstrap init to be slim, just enough to get us over into the AgentClassLoader.

Also, this allows future vendor logging customization hooks to happen inside of tooling, which is where other vendor customization happens.

cc: @pavolloffay

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/LoggingConfigurer.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/LoggingConfigurer.java
@@ -1,0 +1,62 @@
+package io.opentelemetry.javaagent.tooling;
+
+class LoggingConfigurer {
+
+  private static final String SIMPLE_LOGGER_SHOW_DATE_TIME_PROPERTY =
+      "io.opentelemetry.javaagent.slf4j.simpleLogger.showDateTime";
+  private static final String SIMPLE_LOGGER_DATE_TIME_FORMAT_PROPERTY =
+      "io.opentelemetry.javaagent.slf4j.simpleLogger.dateTimeFormat";
+  private static final String SIMPLE_LOGGER_DATE_TIME_FORMAT_DEFAULT =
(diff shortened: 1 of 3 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/LoggingConfigurer.java
```diff
@@ -0,0 +1,67 @@
...
+   */
+  private static boolean isDebugMode() {
+    String tracerDebugLevelSysprop = "otel.javaagent.debug";
+    String tracerDebugLevelProp = System.getProperty(tracerDebugLevelSysprop);
+
+    if (tracerDebugLevelProp != null) {
+      return Boolean.parseBoolean(tracerDebugLevelProp);
+    }
+
+    String tracerDebugLevelEnv =
```
anuraaga, line 59: Since this is in the agent, can we use `Config.get()`?

trask: logging gets set up before config (so that we can log config errors 😄)

trask: or wait, maybe we should load config first, so that we can configure logging, now I'm confused...

anuraaga: I see - that seems like enough reason to not use it

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/LoggingConfigurer.java
trask: ya, I'll think more about this in subsequent PRs, planning to do more work in this area (working towards the idea you have mentioned before of "agent builder" API instead of sprawling SPIs).

iNikem (approved):

mateuszrzeszutek (approved):

PR #2307, opened 2021-02-16
Implement spanName for MongoClientTracer

### LINKED ISSUE #2306: MongoDB Span Names don't match specification
**Describe the issue**
The MongoDB implementation is currently specifying the query statement as a span name, rather than a low-cardinality value as recommended by the specification. It should be possible for the span name to follow the recommended convention of: `<db.operation> <db.name>.<db.mongodb.collection>`

jamal:
Adds a basic implementation for spanName in MongoClientTracer that uses the recommended  name from the specification.

Please let me know if this is not the right way to solve this and I'm happy to iterate on it. :)

Fixes #2306

### DIFF
--- a/instrumentation/mongo/mongo-3.1/javaagent/src/test/groovy/MongoClientTest.groovy
+++ b/instrumentation/mongo/mongo-3.1/javaagent/src/test/groovy/MongoClientTest.groovy
@@ -37,6 +37,7 @@ class MongoClientTest extends MongoBaseTest {
   def "test create collection"() {
     setup:
     MongoDatabase db = client.getDatabase(dbName)
+    client.listDatabaseNames()
 
     when:
     db.createCollection(collectionName)
--- a/instrumentation/mongo/mongo-3.1/javaagent/src/test/groovy/MongoClientTest.groovy
+++ b/instrumentation/mongo/mongo-3.1/javaagent/src/test/groovy/MongoClientTest.groovy
@@ -37,7 +37,6 @@ class MongoClientTest extends MongoBaseTest {
   def "test create collection"() {
     setup:
     MongoDatabase db = client.getDatabase(dbName)
-    client.listDatabaseNames()
 
     when:
     db.createCollection(collectionName)
--- a/instrumentation/mongo/mongo-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
+++ b/instrumentation/mongo/mongo-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
@@ -85,8 +85,7 @@ public class MongoClientTracer extends DatabaseClientTracer<CommandStartedEvent,
   }
 
   @Override
-  public String spanName(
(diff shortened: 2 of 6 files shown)


instrumentation/mongo/mongo-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
```diff
@@ -84,6 +84,29 @@ protected InetSocketAddress peerAddress(CommandStartedEvent event) {
...
+  protected String spanName(
+      CommandStartedEvent event, BsonDocument document, String normalizedQuery) {
+    String dbName = dbName(event);
+    if (event.getCommandName() == null) {
+      return dbName == null ? DB_QUERY : dbName;
+    }
+
+    String collectionName = collectionName(event);
+    StringBuilder name = new StringBuilder();
+    name.append(event.getCommandName()).append(' ');
```
anuraaga, line 97: Does this have a chance of a span name with a trailing space? Perhaps we can add a unit test for that case.

anuraaga (commented):
Thanks @jamal!

instrumentation/mongo/mongo-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/MongoClientTracer.java
jamal: Ah, good call. There are operations that can probably be done through the Java driver that wouldn't have a DB Name. I'll update and add a test for that. Thanks!

jamal: I've updated this and added a test that returns null for both dbName and the collection name. I don't know if its possible to actually have a situation with no dbName (even management commands use `admin` as the dbName) but its good to be safe :)

instrumentation/mongo/mongo-3.1/javaagent/src/test/groovy/MongoClientTest.groovy
```diff
@@ -37,6 +37,7 @@ class MongoClientTest extends MongoBaseTest {
   def "test create collection"() {
     setup:
     MongoDatabase db = client.getDatabase(dbName)
+    client.listDatabaseNames()
```
mateuszrzeszutek, line 40: Is this line needed?

mateuszrzeszutek (approved):
LGTM 👍

instrumentation/mongo/mongo-3.1/javaagent/src/test/groovy/MongoClientTest.groovy
jamal: Ah, no. I accidentally left that in there. Cleaning it up.

mateuszrzeszutek (approved):

iNikem (approved):

anuraaga (approved):
Thanks!

PR #2309, opened 2021-02-17
Don't String.format when determining if instrumentation is enabled.

anuraaga:
Randomly found this - we should never use the slow `String.format` in non-error code. This usage I guess has some impact on startup time.


trask (approved):
👍

anuraaga:
Unrelated muzzle failure, merging

iNikem:
Just a comment: I heard a lot of times when Brian Goetz speaking about future java plans etc he mentioned that `String.format` should be the default way to format a string and Java/JVM team does want to make it fast enough to actually become a default.

anuraaga:
@iNikem Once Java 30 is out (meaning everyone's on Java 11)? :P It's too bad that when targeting pre Java9, we have no way of taking advantage of the improvements being made to string :(

PR #2311, opened 2021-02-17
Enable runtime metrics on agent startup.

anuraaga:
Sometimes we get asked about them. It doesn't hurt to turn them on.


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -36,6 +39,11 @@ public static synchronized void installAgentTracer() {
       OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
       OpenTelemetrySdkAccess.internalSetForceFlush(
           (timeout, unit) -> sdk.getSdkTracerProvider().forceFlush().join(timeout, unit));
+
+      if (Config.get().isInstrumentationEnabled(Collections.singleton("runtime-metrics"), true)) {
+        GarbageCollector.registerObservers();
+        MemoryPools.registerObservers();
+      }
```
mateuszrzeszutek, line 46: What do you think about creating a separate `ComponentInstaller` in the `:instrumentation:runtime-metrics:javaagent` module? This way we could avoid adding instrumentation code (well, sort of) to javaagent-tooling module.

instrumentation/runtime-metrics/javaagent/src/test/groovy/io/instrumentation/javaagent/instrumentation/runtimemetrics/RuntimeMetricsTest.groovy
```diff
@@ -0,0 +1,24 @@
...
+
+class RuntimeMetricsTest extends AgentInstrumentationSpecification {
+
+  def "test runtime metrics is enabled"() {
+    when:
+    def conditions = new PollingConditions(timeout: 10, initialDelay: 1.5, factor: 1.25)
+
+    then:
+    conditions.eventually {
+      assert metrics.any { it.name == "runtime.jvm.gc.collection" }
```
iNikem, line 19: Where does this `metrics` come from?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
iNikem: ^^ This is a VERY good idea

instrumentation/runtime-metrics/javaagent/src/test/groovy/io/instrumentation/javaagent/instrumentation/runtimemetrics/RuntimeMetricsTest.groovy
mateuszrzeszutek: `InstrumentationSpecification`

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
trask: oh, that is a good idea 👍

anuraaga: Hmm - I tried and it's not trivial since it leads to a dependency cycle in Gradle, `javaagent-tooling` depends on `runtime-metrics:javaagent` depends on `javaagent-tooling`. I'm also not sure it's that much better - it's nowhere close to the pattern we use in any other `javaagent` module so working around these issues seems to be counter-intuitive?

anuraaga: BTW, I think the way I'd solve it is to have a folder, ideally not called `javaagent` since it seems so inconsistent, that does not `apply from: instrumentation.gradle` since it doesn't need any of its patterns. And have a `javaagent-testing` folder which does which only includes unit tests. It's highly inconsistent though.

trask: `ComponentInstaller` is in `javaagent-spi` not `javaagent-tooling`, does using that directly break the cycle?

anuraaga: We add javaagent-tooling to all the instrumentations. But we've refactored a lot, perhaps that's not needed anymore let me see if there's a simple fix

trask: oh right, all the bytebuddy matchers are in javaagent-tooling, I'm good with this as-is, and we can revisit later how to decouple it

mateuszrzeszutek: Hmm, why does `javaagent-tooling` need to depend on `runtime-metrics:javaagent`? It does not depend directly on all other instrumentations, they all get added to `javaagent` directly.

anuraaga: @mateuszrzeszutek Oops - forgot it'd be automatically added, yeah let me try it

instrumentation/runtime-metrics/javaagent/src/test/groovy/io/instrumentation/javaagent/instrumentation/runtimemetrics/RuntimeMetricsTest.groovy
anuraaga: I had the exact same confusion when reading SystemMetricsTest. I renamed these to `getMetrics()` since I think it's less mysterious.

anuraaga (commented):
Thanks for the tip!

iNikem (approved):

trask (approved):
👍

Issue #2304, opened 2021-02-16
Add traceFlags to logging MDC instead of sampled boolean


trask:
The [logging spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#trace-context-in-legacy-formats) mentions:

> - "trace_id" for [TraceId](data-model.md#field-traceid), hex-encoded.
> - "span_id" for [SpanId](data-model.md#field-spanid), hex-encoded.
> - "trace_flags" for [trace flags](data-model.md#field-traceflags), formatted
  according to W3C traceflags format.

To follow this better, it seems would be good to add traceFlags to MDC instead of sampled boolean.

PR #2312, opened 2021-02-17
Add 'trace_flags' to logging MDC

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2304

* Remove 'sampled' flag
* Rename MDC keys to the current logging spec convention: trace_id instead of traceId

I'll update the logging doc after release.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/log/LoggingContextConstants.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/log/LoggingContextConstants.java
@@ -5,6 +5,8 @@
 
 package io.opentelemetry.instrumentation.api.log;
 
+import io.opentelemetry.api.trace.SpanContext;
+
 /**
  * This class contains several constants used in logging libraries' Mapped Diagnostic Context
  * instrumentations.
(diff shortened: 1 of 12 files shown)


iNikem (approved):

trask (approved):
I'm good with this change, or with the existing java-style MDC property names (`traceId`, `spanId`, `traceFlags`).

mateuszrzeszutek:
> I'm good with this change, or with the existing java-style MDC property names (traceId, spanId, traceFlags).

Some more explanation for my changes: I think that `trace_id` is a bit more consistent, we use the same name everywhere - and if somebody chooses to log the whole MDC (just use `%X` pattern) they'll get something like `trace_id=12313445 ...` by default.

anuraaga (approved):

kazuki43zoo:
Oh... this change broken compatibility with json layout provided by spring-cloud-gcp-logging... Unfortunately...

https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/552f227d6b508a4af54166934f176c6fc38ea3f8/spring-cloud-gcp-logging/src/main/java/com/google/cloud/spring/logging/StackdriverTraceConstants.java#L57

Issue #1494, opened 2020-10-27
Some missing tests for capturing servlet context path


trask:
PR #1418 started capturing servlet context path as part of the span name "route", but we are still missing tests for this behavior in the `jaxrs-1.0` and `spring-webflux` modules.

trask:
Closing, looks like this was addressed.

PR #2317, opened 2021-02-17
Fix Operation name logging for Netty based applications

neinoi:
Fix Operation name logging for Netty based applications

(@trask Report of PR #1494 on ApplicationInsights-Java repo)

### DIFF
--- a/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
+++ b/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerRequestTracingHandler.java
@@ -45,7 +45,7 @@ public class HttpServerRequestTracingHandler extends SimpleChannelUpstreamHandle
 
     HttpRequest request = (HttpRequest) message;
 
-    Context context = tracer().startSpan(request, ctx.getChannel(), channelTraceContext, request.getUri().split("\\?")[0]);
+    Context context = tracer().startSpan(request, ctx.getChannel(), channelTraceContext, "HTTP " + request.getMethod());
     try (Scope ignored = context.makeCurrent()) {
       ctx.sendUpstream(event);
       // the span is ended normally in HttpServerResponseTracingHandler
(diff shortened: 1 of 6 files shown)


laurit:
See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md
> Instrumentation MUST NOT default to using URI path as span name

neinoi:
@laurit if I follow your recommandation, all my operations will be « HTTP GET » … so I can let « netty.request » the result is the same.

These are more usefull : https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span

iNikem (changes requested):
As @laurit correctly noticed, using full URL (even without query string) violates Otel semantic convention and general requirement for low cardinality of the span names.

neinoi:
Here without my modifications : 

image
I agree, for low cardinality it’s perfect, all calls have the same name.

Here after my modifications (first version, before my last modifications) : 
image
It’s not perfect, but I think a bit better.

trask:
Hi @neinoi! As appealing as it is to use the path as the span name, it doesn't work well for applications that have an unbounded number of (parameterized) paths.

Since the one-size-fits-all span name is not that great (and we all agree that `HTTP GET` is not that great 😄), there are ways for applications to customize the span name.

Since you are using Application Insights, if you open an issue in that repo, I can explain the different options for customizing the span name (which are a bit specific to Application Insights, since we don't use the OTel Collector, and our OTel API interoperability is under a preview flag).

If you'd like to continue with this PR, changing the span name to `HTTP {method}` would be great, we missed `netty.request` previously when updating to the OpenTelemetry semantic conventions, and it would be a good update this.

neinoi:
Hi @trask ,
I’ve modified the code to have `HTTP {method}` for the span name.
For the options, is this the good place to go : [java-standalone-telemetry-processors](https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-telemetry-processors) ?
(I will transmit these informations to my successor in my current project)

trask (commented):
Thanks @neinoi!

> For the options, is this the good place to go : [java-standalone-telemetry-processors](https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-standalone-telemetry-processors) ?

Yes, that's a good option. Another option is https://docs.microsoft.com/en-us/azure/azure-monitor/app/java-in-process-agent#override-the-request-telemetry-name-using-the-2x-sdk.

trask (approved):

anuraaga (approved):

iNikem (approved):

PR #2320, opened 2021-02-17
Instrument all jetty handlers

laurit:
Due to the excluded handlers embedded jetty used in our tests usually runs without using jetty instrumentation even when test has a dependency to it.


trask (approved):
nice 👍

anuraaga (approved):

iNikem (approved):

PR #2322, opened 2021-02-17
Allow Gradle build without JDK/bin in PATH

agoallikmaa:
Currently when running a Gradle build (also when importing in IDE), the RMI module invokes the `rmic` binary, assuming it is in `PATH`. This does not work in cases where no JDK `bin` directory is in `PATH`, but `JAVA_HOME` is used instead to tell Gradle where the JDK is.

Added looking for the `rmic` binary in `${java.home}/bin` (or `${java.home}/../bin`) before falling back to invoking it without specifying its full path.

### DIFF
--- a/instrumentation/rmi/javaagent/rmi-javaagent.gradle
+++ b/instrumentation/rmi/javaagent/rmi-javaagent.gradle
@@ -8,7 +8,14 @@ muzzle {
 
 task "rmic", dependsOn: testClasses {
   def clazz = 'rmi.app.ServerLegacy'
-  String command = """rmic -g -keep -classpath ${sourceSets.test.output.classesDirs.asPath} -d ${buildDir}/classes/java/test ${clazz}"""
+
+  // Try one level up too in case java.home refers to jre directory inside jdk directory
+  def rmicBinaryPath = ['/bin/rmic', '/../bin/rmic'].findResult {
+    def path = new File(System.getProperty("java.home"), it).getAbsoluteFile()
+    path.isFile() ? path.toString() : null
+  } ?: "rmic"
+
+  String command = """$rmicBinaryPath -g -keep -classpath ${sourceSets.test.output.classesDirs.asPath} -d ${buildDir}/classes/java/test ${clazz}"""
   command.execute().text
 }
 


instrumentation/rmi/javaagent/rmi-javaagent.gradle
```diff
@@ -8,7 +8,14 @@ muzzle {
...
   def clazz = 'rmi.app.ServerLegacy'
-  String command = """rmic -g -keep -classpath ${sourceSets.test.output.classesDirs.asPath} -d ${buildDir}/classes/java/test ${clazz}"""
+
+  // Try one level up too in case java.home refers to jre directory inside jdk directory
+  def rmicBinaryPath = ['/bin/rmic', '/../bin/rmic'].findResult {
+    def path = new File(System.getProperty("java.home"), it).getAbsoluteFile()
+    path.isFile() ? path.toString() : null
+  } ?: "rmic"
+
+  String command = """$rmicBinaryPath -g -keep -classpath ${sourceSets.test.output.classesDirs.asPath} -d ${buildDir}/classes/java/test ${clazz}"""
```
iNikem, line 18: Why do we do this during task creation? Shouldn't this be in `doLast`? How does it work at all??

agoallikmaa: Not actually sure why it is done this way, just changed it because I do not keep a JDK in `PATH`, and because this may cause tooling from different JDKs to be mixed.

But it does indeed look like `doLast` was intended here, because if you run the tests from a clean state, the `rmic` runs before everything else, and thus sees no input (which would be produced by test compile task, which runs later). However, the test actually passes without the files generated by `rmic` being present at all, which raises the question of why even have them.

So it might actually make sense to throw this task away, since all clean builds up to now have always passed without the `rmic` outputs anyway, so functionally nothing would change. Alternatively, would have to change the test so it actually makes use of the stubs (and thus fails without their presence) if the intent was to make sure instrumentation doesn't break anything with stubs, but that seems like a full task and not a simple fix as was intended with this PR.

iNikem (approved):

trask (approved):

PR #2328, opened 2021-02-18
Inject OpenTelemetrySdk into lambda library instrumentation instead o…

anuraaga:
…f using global.

This change is needed to update to 0.17.0

I have changed the wrappers to use `autoconfigure` instead of "the global", I guess in general the global would be a noop without that anyways. But because our test infrastructure is not set up for handling library instrumentation which has its own autoconfigure, and we probably wouldn't update it for the one-and-only case where library instrumentation depends on the SDK, we'll need to add some smoke-tests eventually.

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ParentContextExtractor.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ParentContextExtractor.java
@@ -20,7 +20,7 @@ public class ParentContextExtractor {
 
   private static final String AWS_TRACE_HEADER_ENV_KEY = "_X_AMZN_TRACE_ID";
 
-  static Context extract(Map<String, String> headers) {
+  static Context extract(Map<String, String> headers, BaseTracer tracer) {
     Context parentContext = null;
     String parentTraceHeader = System.getenv(AWS_TRACE_HEADER_ENV_KEY);
     if (parentTraceHeader != null) {
@@ -28,7 +28,7 @@ public class ParentContextExtractor {
     }
     if (!isValidAndSampled(parentContext)) {
       // try http
-      parentContext = fromHttpHeaders(headers);
+      parentContext = fromHttpHeaders(headers, tracer);
     }
     return parentContext;
   }
@@ -42,8 +42,8 @@ public class ParentContextExtractor {
     return (parentSpanContext.isValid() && parentSpanContext.isSampled());
   }
 
-  private static Context fromHttpHeaders(Map<String, String> headers) {
-    return BaseTracer.extractWithGlobalPropagators(lowercaseMap(headers), MapGetter.INSTANCE);
+  private static Context fromHttpHeaders(Map<String, String> headers, BaseTracer tracer) {
+    return tracer.extract(lowercaseMap(headers), MapGetter.INSTANCE);
   }
 
   // lower-case map getter used for extraction
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
@@ -7,10 +7,21 @@ package io.opentelemetry.instrumentation.awslambda.v1_0;
 
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
 
 /**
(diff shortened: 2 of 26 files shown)


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/ParentContextExtractor.java
```diff
@@ -20,15 +20,15 @@
 
   private static final String AWS_TRACE_HEADER_ENV_KEY = "_X_AMZN_TRACE_ID";
 
-  static Context extract(Map<String, String> headers) {
+  static Context extract(Map<String, String> headers, BaseTracer tracer) {
```
anuraaga, line 23: This is a bad API, but untangling would add a lot to this PR I think. Would like to address it in the future as tracer APIs are quite unstable anyways.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
```diff
@@ -7,10 +7,21 @@
...
 
 /**
  * Wrapper for {@link TracingRequestHandler}. Allows for wrapping a lambda proxied through API
  * Gateway, enabling single span tracing and HTTP context propagation.
  */
 public class TracingRequestApiGatewayWrapper
-    extends TracingRequestWrapperBase<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {}
+    extends TracingRequestWrapperBase<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+
+  public TracingRequestApiGatewayWrapper() {
```
anuraaga, line 19: Requires smoke test to verify

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
```diff
@@ -21,38 +24,28 @@
  */
 public abstract class TracingRequestHandler<I, O> implements RequestHandler<I, O> {
 
-  private static final long DEFAULT_FLUSH_TIMEOUT_SECONDS = 1;
+  protected static final Duration DEFAULT_FLUSH_TIMEOUT = Duration.ofSeconds(1);
 
   private final AwsLambdaTracer tracer;
-  private final long flushTimeout;
```
anuraaga, line 27: Was confused on the unit, so took the chance to clean it up to explicit units while here

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsEventHandler.java
```diff
@@ -50,7 +56,6 @@ public Void doHandleRequest(SQSEvent event, Context context) {
       } else {
         tracer.end(otelContext);
       }
-      LambdaUtils.forceFlush();
```
anuraaga, line 53: I realized this flush wasn't actually needed since the base class handles it

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
```diff
@@ -86,11 +87,15 @@ protected abstract void doHandleRequest(InputStream input, OutputStream output,
 
     private final OutputStream delegate;
     private final io.opentelemetry.context.Context otelContext;
+    private final OpenTelemetrySdk openTelemetrySdk;
 
     private OutputStreamWrapper(
-        OutputStream delegate, io.opentelemetry.context.Context otelContext) {
+        OutputStream delegate,
+        io.opentelemetry.context.Context otelContext,
+        OpenTelemetrySdk openTelemetrySdk) {
```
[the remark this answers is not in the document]
jkwatson: It's definitely *NOT* a good thing, but if the lambda instrumentation needs to call `forceFlush` it's the only option we have right now...unless we introduce a shim interface that exposes that function, and make people construct an implementation that hides the SDK.

trask (approved):

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestStreamHandler.java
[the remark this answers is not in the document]
anuraaga: Yeah - it's not a good thing, but when discussing in spec repo and sig we decided that for now, lambda requiring the dependency is fine since it's the only way.

https://github.com/open-telemetry/opentelemetry-specification/issues/1287#issuecomment-745697250

anuraaga:
@iNikem Going to go ahead and merge this but happy to follow up with anything

PR #2339, opened 2021-02-19
Publish a BOM

anuraaga:


trask (approved):

bom-alpha/bom-alpha.gradle
```diff
@@ -0,0 +1,39 @@
+plugins {
+  id("java-platform")
+}
+
+apply from: "$rootDir/gradle/publish.gradle"
+
+description = "OpenTelemetry Instrumentation Bill of Materials (Alpha)"
+group = "io.opentelemetry.instrumentation"
+archivesBaseName = "opentelemetry-instrumentation-bom-alpha"
```
anuraaga, line 9: I think we'll probably have two boms, `javaagent-bom` for custom distros, and `instrumentation-bom` for library instrumentation users. For now I include everything into this and we can clean up as things stabilize some.

PR #2345, opened 2021-02-19
Migrate Armeria to HttpServer/ClientTest

anuraaga:


instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/server/OpenTelemetryService.java
```diff
@@ -52,20 +50,25 @@ private OpenTelemetryService(HttpService delegate, ArmeriaServerTracer serverTra
...
     Context context = serverTracer.startSpan(req, ctx, null, spanName, requestStartTimeNanos);
 
-    if (Span.fromContext(context).isRecording()) {
+    Span span = Span.fromContext(context);
+    if (span.isRecording()) {
       ctx.log()
           .whenComplete()
           .thenAccept(
               log -> {
+                if (log.responseHeaders().status() == HttpStatus.NOT_FOUND) {
```
anuraaga, line 67: /cc @trustin @minwoox @ikhoon to confirm this is the most natural way of setting our route name to the path itself when not found.

anuraaga: Though it may not matter - our unit test is checking that the span name is the path name, but for not found, shouldn't our instrumentation generally fallback to `HTTP {method}` @trask @iNikem?

trustin: @anuraaga As you noticed, you can't distinguish who returned 404 (fallback or your service) with `log.responseHeaders().status() == HttpStatus.NOT_FOUND`.

If you did not bind any services on `Route.ofCatchAll()`, you could do this:

    ctx.config().route() == Route.ofCatchAll()

.. but this isn't a great experience. How about this:

    ctx.config().route().isFallback()

..? I guess we can add this in the next minor update.

anuraaga: @trustin Yeah I think that could be helpful :)

iNikem (approved):

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/server/OpenTelemetryService.java
trustin: https://github.com/line/armeria/issues/3365

PR #2351, opened 2021-02-19
Fix debug logging

trask:
This was broken by #2305. I will create an issue for adding smoke test coverage for this.

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
@@ -9,8 +9,6 @@ import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


jkwatson (approved):
I can confirm that this restores the missing debug logging.

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -61,19 +62,19 @@ public AgentClassLoader(
...
+      throw new IllegalStateException("URL malformed.  Unsupported JDK?", e);
     }
 
     if (!AGENT_INITIALIZER_JAR.isEmpty()) {
       URL url;
       try {
         url = new File(AGENT_INITIALIZER_JAR).toURI().toURL();
       } catch (MalformedURLException e) {
-        log.warn(
+        throw new IllegalStateException(
```
anuraaga, line 73: This only happens due to a bug in packaging and we assume that a any published distribution at least has enough testing to know it won't cause this crash I guess. It seems fine to me.

anuraaga (approved):

iNikem (approved):

PR #2360, opened 2021-02-20
Remove trivial uses of Guava from agent.

anuraaga:
The only remaining usage in the agent is caching, and there is usage at compile time by muzzle of the graph library and some test usage. So we'll be able to replace the caching with caffeine for a bit smaller distribution size and probably better cache performance.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -33,7 +33,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
@@ -491,18 +490,15 @@ public class AgentInstaller {
     @Override
     public Iterable<Iterable<Class<?>>> resolve(Instrumentation instrumentation) {
       // filter out our agent classes and injected helper classes
-      return iterableOf(
-          () ->
-              streamOf(delegate.resolve(instrumentation))
-                  .map(classes -> iterableOf(() -> streamOf(classes).filter(c -> !isIgnored(c)))));
+      return () -> streamOf(delegate.resolve(instrumentation)).map(this::filterClasses).iterator();
+    }
+
+    private Iterable<Class<?>> filterClasses(Iterable<Class<?>> classes) {
+      return () -> streamOf(classes).filter(c -> !isIgnored(c)).iterator();
     }
 
     private static <T> Stream<T> streamOf(Iterable<T> iterable) {
       return StreamSupport.stream(iterable.spliterator(), false);
-    }
-
-    private static <T> Iterable<T> iterableOf(Supplier<Stream<T>> stream) {
-      return () -> stream.get().iterator();
     }
 
     private static boolean isIgnored(Class<?> c) {
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -11,7 +11,6 @@ import static net.bytebuddy.matcher.ElementMatchers.any;
 import static net.bytebuddy.matcher.ElementMatchers.nameStartsWith;
 import static net.bytebuddy.matcher.ElementMatchers.none;
 
-import com.google.common.collect.Iterables;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.instrumentation.api.internal.BootstrapPackagePrefixesHolder;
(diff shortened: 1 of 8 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -489,8 +489,16 @@ public void onComplete(
     @Override
     public Iterable<Iterable<Class<?>>> resolve(Instrumentation instrumentation) {
       // filter out our agent classes and injected helper classes
-      return Iterables.transform(
-          delegate.resolve(instrumentation), i -> Iterables.filter(i, c -> !isIgnored(c)));
+      return () ->
```
iNikem, line 492: Hope you do understand this and tested it :) Looks like black magic to me

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
```diff
@@ -160,9 +159,8 @@ private boolean doesMatch(ClassLoader loader) {
     Set<HelperReferenceWrapper.Method> plainMethods = new HashSet<>();
     collectMethodsFromTypeHierarchy(helperWrapper, abstractMethods, plainMethods);
 
-    Set<HelperReferenceWrapper.Method> unimplementedMethods =
-        Sets.difference(abstractMethods, plainMethods);
-    for (HelperReferenceWrapper.Method unimplementedMethod : unimplementedMethods) {
+    abstractMethods.removeAll(plainMethods);
```
iNikem, line 162: I think there is a change of behaviour here: in the new version you mutate `abstractMethods`, the old version did not that, did it?

anuraaga: Yeah but it's a local variable that wasn't used after the call to difference. So I'm ok with it I could make a copy instead though

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
anuraaga: The Stream (single use) vs Iterable (multiuse) dilemma is always annoying. The safe way to convert a Stream into an Iterable is creating a Lambda that initializes the Stream and returns the iterator, which is what an Iterable is.

The alternative is to collect(toList) each time which I'm ok with but I don't know this part well enough to know if the extra allocations would be problematic.

trask: I think naming the Iterable <-> Stream transformations makes it (a bit) easier to read:

```
    public Iterable<Iterable<Class<?>>> resolve(Instrumentation instrumentation) {
      // filter out our agent classes and injected helper classes
      return iterableOf(
          streamOf(delegate.resolve(instrumentation))
              .map(classes -> iterableOf(streamOf(classes).filter(c -> !isIgnored(c)))));
    }

    private static <E> Stream<E> streamOf(Iterable<E> iterable) {
      return StreamSupport.stream(iterable.spliterator(), false);
    }

    private static <E> Iterable<E> iterableOf(Stream<E> stream) {
      return stream::iterator;
    }
```

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
anuraaga: Unfortunately `stream:: iterator` isn't a real Iterable since it can't be replayed, so `toIterable` would need to accept `Supplier<Stream>`. Not sure if that ends up better.

anuraaga: Tried, how is it?

trask: > Unfortunately `stream:: iterator` isn't a real Iterable since it can't be replayed

😭

ok, how about

```
    public Iterable<Iterable<Class<?>>> resolve(Instrumentation instrumentation) {
      // filter out our agent classes and injected helper classes
      return () -> streamOf(delegate.resolve(instrumentation)).map(this::filterClasses).iterator();
    }

    private Iterable<Class<?>> filterClasses(Iterable<Class<?>> classes) {
      return () -> streamOf(classes).filter(c -> !isIgnored(c)).iterator();
    }

    private static <T> Stream<T> streamOf(Iterable<T> iterable) {
      return StreamSupport.stream(iterable.spliterator(), false);
    }
```

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
anuraaga: Sure seems good

PR #2361, opened 2021-02-20
Better context passing for Netty 4.1

### LINKED ISSUE #1312: Netty instrumentation bug.
I found a new way to break our Netty instrumentation: https://github.com/open-telemetry/opentelemetry-java-instrumentation/compare/master...iNikem:netty-bug?expand=1

The core of the issue is `io.opentelemetry.instrumentation.auto.netty.v4_1.NettyChannelPipelineInstrumentation.ChannelPipelineConnectAdvice`. It stores the context, which was active during Netty channel connect, to the channel. This context is then used by `HttpClientRequestTracingHandler` to scope the actual http call.

My test executes Netty connect manually, outside any active context. As a result `ChannelPipelineConnectAdvice` stores an invalid span to the Netty channel and it becomes the parent of http call span. Totally ignoring actual span that was active when call was made by test.

If I comment out `ChannelPipelineConnectAdvice`, test passes. Unfortunately, some other tests start failing. E.g. for Ratpack. Thus this can of worms require very careful handling...

iNikem:
Follows #2323 
Fixes #1312

This PR changes one important aspect of Netty 4.1 instrumentation. Before, we remembered the context that was active when Netty establishes a network connection. Now we remember the context that was active during write operation on the channel. This is the lowest level operation in Netty to trigger the actual request execution.

Now, integrations with specific http client libraries which are based on Netty have to ensure the context propagation from request initiation method in their high level API to the point where library's implementation calls Netty.

This PR also rewrites Netty41ClientTest to actually use Netty, and not higher level AsyncHttpClient.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
@@ -50,25 +50,30 @@ public final class ContextPropagationDebug {
       if (currentSpan != null) {
         log.error("It contains this span: {}", currentSpan);
       }
-      List<StackTraceElement[]> locations = ContextPropagationDebug.getLocations(current);
-      if (locations != null) {
-        StringBuilder sb = new StringBuilder();
-        Iterator<StackTraceElement[]> i = locations.iterator();
-        while (i.hasNext()) {
-          for (StackTraceElement ste : i.next()) {
-            sb.append("\n");
-            sb.append(ste);
-          }
-          if (i.hasNext()) {
-            sb.append("\nwhich was propagated from:");
-          }
-        }
-        log.error("a context leak was detected. it was propagated from:{}", sb);
-      }
+
+      debugContextPropagation(current);
 
       if (FAIL_ON_CONTEXT_LEAK) {
         throw new IllegalStateException("Context leak detected");
       }
+    }
+  }
+
+  public static void debugContextPropagation(Context context) {
+    List<StackTraceElement[]> locations = ContextPropagationDebug.getLocations(context);
+    if (locations != null) {
+      StringBuilder sb = new StringBuilder();
+      Iterator<StackTraceElement[]> i = locations.iterator();
+      while (i.hasNext()) {
+        for (StackTraceElement ste : i.next()) {
+          sb.append("\n");
+          sb.append(ste);
+        }
+        if (i.hasNext()) {
+          sb.append("\nwhich was propagated from:");
+        }
+      }
+      log.error("a context leak was detected. it was propagated from:{}", sb);
     }
   }
 
--- a/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureInstrumentation.java
+++ b/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureInstrumentation.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.netty.v4_1;
+
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.implementsInterface;
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.ClassLoaderMatcher.hasClassesNamed;
+import static java.util.Collections.singletonMap;
+import static net.bytebuddy.matcher.ElementMatchers.isMethod;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
+
+import io.netty.util.concurrent.GenericFutureListener;
+import io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class ChannelFutureInstrumentation implements TypeInstrumentation {
+
+  @Override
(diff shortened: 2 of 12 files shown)


instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ClientTest.groovy
```diff
@@ -287,21 +285,6 @@ class Netty41ClientTest extends HttpClientTest implements AgentTestTrait {
     null != channel.pipeline().remove('Netty41ClientTest$OtherSimpleHandler#0')
   }
 
-  def "when a traced handler is added from an initializer we still detect it and add our channel handlers"() {
```
iNikem, line 290: This test genuinely fails on  Netty 4.1.0.Final, and only works starting from 4.1.5.Final,  see https://github.com/netty/netty/pull/5588

It worked before this PR only because test used AHC, which used netty version newer than 4.1.0.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
```diff
@@ -50,27 +50,32 @@ public static void debugContextLeakIfEnabled() {
...
+
+      debugContextPropagation(current);
 
       if (FAIL_ON_CONTEXT_LEAK) {
         throw new IllegalStateException("Context leak detected");
       }
     }
   }
 
+  public static void debugContextPropagation(Context context) {
```
anuraaga, line 62:
```suggestion
  private static void debugContextPropagation(Context context) {
```

Don't see it used

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientRequestTracingHandler.java
```diff
@@ -24,7 +24,7 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise prm) {
       return;
     }
 
-    Context parentContext = ctx.channel().attr(AttributeKeys.CONNECT_CONTEXT).getAndRemove();
+    Context parentContext = ctx.channel().attr(AttributeKeys.WRITE_CONTEXT).getAndRemove();
```
anuraaga, line 27: Can you add some documentation maybe in `AttributeKeys` on when we'd parent on the connect context vs the write context? Or do we really need two keys?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
iNikem: I used it during the debugging of this PR. That was the reason for extraction as a separate method.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientRequestTracingHandler.java
iNikem: connect context is used in one specific use-case: when we create a span for connect failure. I plan to follow up with a separate PR that makes that use-case more obvious in the code. Is it Ok if I add documentation then, if it will still be needed?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
anuraaga: Well if we keep it public we need `// Visible for testing` maybe to make that clear :)

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
iNikem: I mean, I can make it private. Then whenever a developer wants to use it during debugging, they just make it public again :)

anuraaga: Ah - yeah that's fine, it's why my suggestion does that :P

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureInstrumentation.java
```diff
@@ -0,0 +1,52 @@
...
+import io.netty.util.concurrent.GenericFutureListener;
+import io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge;
+import io.opentelemetry.javaagent.tooling.TypeInstrumentation;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class ChannelFutureInstrumentation implements TypeInstrumentation {
```
trask, line 24: should we add this to netty 3.8/4.0 instrumentation also?

trask (approved):

instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ClientTest.groovy
iNikem: Indeed, will do that

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureInstrumentation.java
iNikem: None of the tests fail, so seems like none of our supported higher level libraries use similar functionality. Thus - YAGNI :) Less instrumentation is better, right?

trask: consistency is good too, but I have no objection to leaving 3.8/4.0 behind in this regard until (if ever) needed 👍

PR #2371, opened 2021-02-22
Update distro example to 0.17.0

mateuszrzeszutek:


examples/distro/build.gradle
```diff
@@ -2,53 +2,53 @@ group 'io.opentelemetry.example'
...
+        dependencies.create(group: 'com.google.auto.service', name: 'auto-service', version: '1.0-rc7'),
+        dependencies.create(group: 'com.google.auto', name: 'auto-common', version: '0.8'),
+        dependencies.create(group: 'com.google.guava', name: 'guava', version: "${versions.guava}"),
+      ],
+      autoValueAnnotations: "com.google.auto.value:auto-value-annotations:${versions.autoValue}",
+    ]
+  }
+
+  repositories {
+    jcenter()
```
iNikem, line 32: JCenter is sunsetting, we should remove it.

examples/distro/custom/src/main/java/com/example/javaagent/DemoResourceProvider.java
```diff
@@ -1,11 +1,29 @@
...
 
-public class DemoResourceProvider extends ResourceProvider {
   @Override
-  protected Attributes getAttributes() {
-    return Attributes.builder().put("custom.resource", "demo").build();
+  public Resource createResource(ConfigProperties config) {
+    return DemoResource.get();
+  }
+
+  private static final class DemoResource {
```
iNikem, line 15: Why do we need separate class? Can we just inline? Is `createResource` called several times?

mateuszrzeszutek: We can, I just copied the way it's done in the SDK.

iNikem (approved):

trask (approved):

PR #2373, opened 2021-02-22
Add custom distro instrumentation unit test example

mateuszrzeszutek:


trask (approved):
can you also change the github action `example-distro` jobs from `build` to `test` so they will run the test too? Looks like I should have done that in #2292 to begin with, since we had smoke-tests already

anuraaga:
`build` runs tests too 😀 assemble is the one that doesn't

examples/distro/gradle/instrumentation.gradle
```diff
@@ -0,0 +1,63 @@
...
+tasks.withType(Test).configureEach {
+  jvmArgs "-Dotel.javaagent.debug=true"
+  jvmArgs "-javaagent:${configurations.testAgent.files.first().absolutePath}"
+  jvmArgs "-Dotel.initializer.jar=${shadowJar.archiveFile.get().asFile.absolutePath}"
+  jvmArgs "-Dinternal.testing.disable.global.library.ignores=true"
+  // always run with the thread propagation debugger to help track down sporadic test failures
+  jvmArgs "-Dotel.threadPropagationDebugger=true"
+  jvmArgs "-Dotel.internal.failOnContextLeak=true"
+  // always print muzzle warnings
+  jvmArgs "-Dio.opentelemetry.javaagent.slf4j.simpleLogger.log.muzzleMatcher=warn"
```
iNikem, line 49: @mateuszrzeszutek Doesn't this mean that only WARN and above will be logged from muzzle? Meaning the opposite of what you wanted? Also `-Dotel.javaagent.debug=true` should already tweak muzzle logging as needed.

mateuszrzeszutek: It only logs WARNs anyway. But you're right, this is completely not needed, `otel.java.debug` enables it anyway - I'll fix that.

PR #2374, opened 2021-02-22
precompile pattern and use char replacement.

breedx-splk:
I saw some allocations being done in a stack trace that pointed to the compiling of a regex inside of `Utils.getInternalName()`.  Upon closer inspection, it became clear that:

1. the pattern should probably be precompiled for efficiency
2. there was duplicate code to do this replacement
3. 🔥 I think there was a bug in the regex.  It had expected a literal `$` at the end of the resource name after the `.class`.  That didn't match the comments, nor does that make immediate sense to me, so I changed it.  I believe it was trying to match the end of string, in which case the `$` should not have been escaped.  Marking with a 🔥 because this is a change of behavior.
 
So this PR consolidates that duplication, fixes the bug, does a precompilation of the pattern, and backfills some missing unit tests.

It also changes a single character `replaceAll()` over to just a basic character replace in `convertToInnerClassName()`.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Utils.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Utils.java
@@ -30,6 +30,8 @@ public class Utils {
     }
   }
 
+  private static final String CLASS_SUFFIX = ".class";
+
   /** Return the classloader the core agent is running on. */
   public static ClassLoader getAgentClassLoader() {
     return AgentInstaller.class.getClassLoader();
@@ -46,10 +48,10 @@ public class Utils {
 
   /** com.foo.Bar to com/foo/Bar.class */
   public static String getResourceName(String className) {
-    if (className.endsWith(".class")) {
+    if (className.endsWith(CLASS_SUFFIX)) {
       return className;
     }
-    return className.replace('.', '/') + ".class";
+    return className.replace('.', '/') + CLASS_SUFFIX;
   }
 
   /** com/foo/Bar.class to com.foo.Bar */
@@ -63,9 +65,8 @@ public class Utils {
   }
 
   private static String stripDotClassSuffix(String resourceName) {
-    String suffix = ".class";
-    if (resourceName.endsWith(suffix)) {
-      return resourceName.substring(0, resourceName.length() - suffix.length());
+    if (resourceName.endsWith(CLASS_SUFFIX)) {
+      return resourceName.substring(0, resourceName.length() - CLASS_SUFFIX.length());
     }
     return resourceName;
   }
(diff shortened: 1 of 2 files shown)


breedx-splk:
Not ready yet, one sec...

breedx-splk:
ready for a look.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Utils.java
```diff
@@ -39,29 +42,31 @@ public static ClassLoader getAgentClassLoader() {
...
   }
 
   /** com.foo.Bar to com/foo/Bar */
   public static String getInternalName(String resourceName) {
-    return resourceName.replaceAll("\\.class\\$", "").replace('.', '/');
+    return stripDotClassSuffix(resourceName).replace('.', '/');
+  }
+
+  private static String stripDotClassSuffix(String resourceName) {
+    return CLASS_SUFFIX_PATTERN.matcher(resourceName).replaceAll("");
```
anuraaga, line 69: Maybe we can just do `endsWith` and `substring`?

iNikem: You mean `indexOf`+`substring`?

anuraaga: Thought endsWith is ok since it's a fixed length so just subtract for the index. indexOf would end up searching the entire string.

breedx-splk: Yup, I like it!

mateuszrzeszutek (approved):

trask (approved):
thanks for the tests 👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Utils.java
```diff
@@ -39,29 +39,35 @@ public static ClassLoader getAgentClassLoader() {
...
   }
 
   /** com.foo.Bar to com/foo/Bar */
   public static String getInternalName(String resourceName) {
-    return resourceName.replaceAll("\\.class\\$", "").replace('.', '/');
+    return stripDotClassSuffix(resourceName).replace('.', '/');
+  }
+
+  private static String stripDotClassSuffix(String resourceName) {
+    String suffix = ".class";
```
iNikem, line 66: Should this be a constant then?

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/Utils.java
breedx-splk: Oh sure, it's in a few places.  Will do.

anuraaga (approved):
Thanks!

Issue #1874, opened 2020-12-10
Hide all span attributes not defined by spec behind an experimental flag


trask:
Otherwise, if/when they are defined by the spec, we will have to keep the old ones around forever too if we don't want to introduce a breaking change.

trask:
All of these are done, except for aws-sdk-2.2, because `Config.get()` will return the default (empty) config for library instrumentation users

A couple possible solutions would be to init the default config with system properties for library instrumentation users, or to have an instrumentation-specific configuration API that library users (and the javaagent) can use to configure it.

PR #2376, opened 2021-02-23
Hide non-specd attributes behind experimental flag

trask:
Even though instrumentation attributes are not being declared stable with 1.0, I still think #1874 is a good thing.

Still todo for #1874 after this PR (will doc on the issue)
* elasticsearch
* apache-camel
* (will leave up to @anuraaga if we apply to aws-sdk, I'm assuming not)

### DIFF
--- a/instrumentation/couchbase/couchbase-2.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_6/CouchbaseConfig.java
+++ b/instrumentation/couchbase/couchbase-2.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_6/CouchbaseConfig.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.javaagent.instrumentation.couchbase.v2_6;
 
 import io.opentelemetry.instrumentation.api.config.Config;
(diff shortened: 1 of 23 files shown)


anuraaga (approved):
We can flag guard aws SDK too - the current ones are quite different from my proposal in the spec. Though I'll enable the flag by default in the aws distro of the agent :)

iNikem (approved):

instrumentation/couchbase/couchbase-2.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_6/CouchbaseConfig.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.couchbase.v2_6;
+
+import io.opentelemetry.instrumentation.api.config.Config;
+
+public class CouchbaseConfig {
```
mateuszrzeszutek, line 10: Nit: consider adding a private constructor and making it final

mateuszrzeszutek (approved):

mateuszrzeszutek:
By the way, should we introduce a single experimental flag that enables all of them?

trask:
> By the way, should we introduce a single experimental flag that enables all of them?

I sort of don't want to make it too easy to use them 😄, I want to motivate people who have some need for these attributes to spec them.

But can be persuaded otherwise.

Issue #2273, opened 2021-02-13
Wrangle @Ignore tests in http-url-connection instrumentation.


breedx-splk:
The instrumentation has 72 tests.  60 are passing, but 12 are ignored. 

I think that we should probably look at those ignored ones and see if/whey they exist and why they might be `@Ignore`d.  If they are not important/relevant, we should delete them.

PR #2382, opened 2021-02-23
Enable disabled tests for HttpUrlConnection

laurit:
Closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2273


iNikem (approved):
Thanks! 🙏

mateuszrzeszutek (approved):

instrumentation/http-url-connection/javaagent/src/test/groovy/HttpUrlConnectionTest.groovy
```diff
@@ -142,7 +163,7 @@ class HttpUrlConnectionTest extends HttpClientTest implements AgentTestTrait {
       connection = url.openConnection()
       connection.useCaches = useCaches
       connection.addRequestProperty("is-test-server", "false")
-      assert activeSpan() != null
+      assert Span.current() != null
```
trask, line 166: Span.current() never returns null

(same is other places below)

```suggestion
      assert Span.current().getSpanContext().isValid()
```

trask (approved):
nice 👍

PR #2383, opened 2021-02-23
Add missing Unroll annotations

laurit:


iNikem:
@trask Did not we have some auto unroll or something?

mateuszrzeszutek (approved):

trask (approved):
> @trask Did not we have some auto unroll or something?

that would be nice, but doesn't ring a bell

mateuszrzeszutek:
If we add `@Unroll` to the `InstrumentationSpecification` it'll probably work everywhere

laurit:
Without unroll method is run for both parameters given in where section, but in test report method is present only once and #methodName is not replaced in its name. With unroll test report has 2 method and #methodName is replaced with parameter.
@iNikem maybe you meant https://github.com/szpak/spock-global-unroll

iNikem:
> @iNikem maybe you meant https://github.com/szpak/spock-global-unroll

Yes, I was sure we had this 🤔

iNikem (approved):

PR #2385, opened 2021-02-23
Only run smoke tests if requested

### LINKED ISSUE #2058: Guard some smoke tests behind CI flag?
When developing, especially for infrastructure work like updating dependencies, it's nice to run `./gradlew check` to make sure everything is fine. But currently smoke tests take 15+min on Mac (presumably Docker is much slower on Mac than on CI), and actually fail with a timeout for me. I guess many of the tests are sort of "for good measure", I'm thinking the many app servers. Does it make sense to slice them out to only run on CI?

iNikem:
Fixes #2058


mateuszrzeszutek (approved):

trask (approved):
👍

PR #2395, opened 2021-02-24
Link to autoconfigure docs.

### LINKED ISSUE #1716: Document possible values for otel.propagators

anuraaga:
Fixes #1716

### DIFF
--- a/docs/agent-config.md
+++ b/docs/agent-config.md
@@ -8,99 +8,16 @@ behavior you find.
 
 ## Contents
 
-* [Exporters](#exporters)
-  + [OTLP exporter (both span and metric exporters)](#otlp-exporter--both-span-and-metric-exporters-)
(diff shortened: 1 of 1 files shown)


trask:
Let's discuss briefly in SIG call in 40 min, there was some discussion in last week morning PST SIG call that "Multiple locations for docs are ok given stability instead of making it hard to find", but worth revisiting.

iNikem (approved):

trask (approved):
discussed in SIG call, the goal is to have primary user documentation at opentelemetry.io, and worth the duplication over there as needed for nice user experience, and removing duplication here makes sense

Issue #2128, opened 2021-01-27
Use low cardinality span names in all DB API instrumentations


mateuszrzeszutek:
All DB API instrumentations that use SQL-like languages (cassandra, couchbase, geode) should use the same span name format as the JDBC instrumentation. (It might be worth it to refactor `DatabaseClientTracer` a little and make `normalizeQuery()` return a type variable; and rename it)

PR #2398, opened 2021-02-24
Refactor DatabaseClientTracer (in preparation for low-cardinality span names)

mateuszrzeszutek:
First step of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2128

This is purely a refactoring, traces emitted by database instrumentation haven't changed - that will be the subject of the next PRs, where I'll correct span names (and set additional attributes, like `db.operation`, where applicable) in all DB instrumentations.


trask (approved):
thanks for splitting out the refactoring from the behavior change 👍

PR #2402, opened 2021-02-24
Move rest of non-specd attributes to experimental

### LINKED ISSUE #1874: Hide all span attributes not defined by spec behind an experimental flag

trask:
Almost completes #1874, except for aws-sdk-2.2, because we don't have solution for `Config.get()` in library instrumentation yet.


iNikem (approved):

anuraaga (approved):

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/KafkaSpanDecorator.java
```diff
@@ -38,6 +39,11 @@
   private static final String TOPIC = "kafka.TOPIC";
   private static final String OFFSET = "kafka.OFFSET";
 
+  private static final boolean CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES =
+      Config.get()
+          .getBooleanProperty(
+              "otel.instrumentation.apache-camel.experimental-span-attributes", false);
```
mateuszrzeszutek, line 45: Hm, can you use the constant from `BaseSpanDecorator` if you make it protected?

mateuszrzeszutek (approved):

mateuszrzeszutek:
@trask what's the problem with calling `Config.get()` in library instrumentations?

trask:
> @trask what's the problem with calling `Config.get()` in library instrumentations?

`Config.get()` will return the default (empty) config for library instrumentation users

A couple possible solutions would be to init the default config with system properties for library instrumentation users, or to have an instrumentation-specific configuration API that library users (and the javaagent) can use to configure it.

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/KafkaSpanDecorator.java
trask: 👍 will make this change (in this PR) tomorrow

PR #2406, opened 2021-02-25
grpc smoke test: Add logging of trace_id to pattern and target log4j2 2.13.2

anuraaga:


anuraaga:
After looking at targeting 2.14.0 with this / instrumentation, realized it's probably easier to just monkeypatch a fix :)

trask (approved):

PR #2407, opened 2021-02-25
Instrument log4j to include bugfix of contextdataprovider mechanism.

### LINKED ISSUE #2403: Log4j instrumentation hijacks the user's logger
**Describe the bug**

The users logger does not function correctly when the log4j instrumentation is in place.  Adding `-Dotel.instrumentation.log4j.enabled=false` JVM arg to disable the instrumentation allows the users log4j logger to operate again...but breaks any chance of using MDC.

**Steps to reproduce**
Add an application dependency on log4j-core 2.13.2.  Build a method like this:

```
    @WithSpan("so-many-socks")
    private void doSomething() {
        Span span = Span.current();
        span.setAttribute("foo", "bar");
        logger.info("This is a test");
    }
```

The `This is a test` string should be logged.  In my case, I was logging to both a file and the console.

**What did you see instead?**
The message is not logged, neither to the file nor to the console.

Tried with latest (0.17.0) but also tried with 0.14.0 (same results).

**Environment**
Compiler: n/a (downloaded the agent from github)
OS: OSX
Runtime (if different from JDK above): adopt openjdk 11

anuraaga:
Depends on #2406 

Fixes #2403

### DIFF
--- a/instrumentation/log4j/log4j-2-testing/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/ListAppender.java
+++ b/instrumentation/log4j/log4j-2-testing/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/ListAppender.java
@@ -7,7 +7,9 @@ package io.opentelemetry.instrumentation.log4j.v2_13_2;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.appender.AbstractAppender;
@@ -28,13 +30,14 @@ public class ListAppender extends AbstractAppender {
 
   private static final ListAppender INSTANCE = new ListAppender();
 
-  private final List<LogEvent> events = Collections.synchronizedList(new ArrayList<LogEvent>());
+  private final List<LoggedEvent> events =
+      Collections.synchronizedList(new ArrayList<LoggedEvent>());
 
   public ListAppender() {
     super("ListAppender", null, null, true);
   }
 
-  public List<LogEvent> getEvents() {
+  public List<LoggedEvent> getEvents() {
     return events;
   }
 
@@ -44,7 +47,12 @@ public class ListAppender extends AbstractAppender {
 
   @Override
   public void append(LogEvent logEvent) {
-    events.add(logEvent);
+    // Event object may be reused by the framework so copy the data we need.
+    LoggedEvent copied =
+        new LoggedEvent(
+            logEvent.getMessage().getFormattedMessage(),
+            new HashMap<>(logEvent.getContextData().toMap()));
+    events.add(copied);
   }
 
   @PluginFactory
@@ -54,5 +62,23 @@ public class ListAppender extends AbstractAppender {
           "Use name=\"ListAppender\" in log4j2-test.xml instead of " + name);
     }
     return INSTANCE;
+  }
+
+  public static class LoggedEvent {
+    private final String message;
+    private final Map<String, String> contextData;
+
+    LoggedEvent(String message, Map<String, String> contextData) {
+      this.message = message;
+      this.contextData = contextData;
+    }
+
+    public String getMessage() {
+      return message;
+    }
+
+    public Map<String, String> getContextData() {
+      return contextData;
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


instrumentation/log4j/log4j-2-testing/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/ListAppender.java
```diff
@@ -44,7 +47,12 @@ public void clearEvents() {
 
   @Override
   public void append(LogEvent logEvent) {
-    events.add(logEvent);
+    // Event object may be reused by the framework so copy the data we need.
```
anuraaga, line 50: Never knew log4j has such an optimization, too bad most spring users never get to see it :D

trask (approved):
nice 👍

instrumentation/log4j/log4j-2.13.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_13_2/BugFixingInstrumentation.java
```diff
@@ -0,0 +1,53 @@
...
+    return Collections.singletonMap(
+        isMethod()
+            .and(isPublic())
+            .and(named("injectContextData"))
+            .and(takesArgument(0, List.class)),
+        BugFixingInstrumentation.class.getName() + "$BugFixingAdvice");
+  }
+
+  public static class BugFixingAdvice {
+    @Advice.OnMethodEnter
```
mateuszrzeszutek, line 46: I know that nothing wrong should happen here, but just in case...

```suggestion
    @Advice.OnMethodEnter(suppress = Throwable.class)
```

instrumentation/log4j/log4j-2.13.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_13_2/Log4j2InstrumentationModule.java
```diff
@@ -39,8 +39,7 @@ public Log4j2InstrumentationModule() {
 
   @Override
   public List<TypeInstrumentation> typeInstrumentations() {
-    // have to return at least 1 type instrumentation so that helpers get injected
-    return singletonList(new EmptyTypeInstrumentation());
+    return Arrays.asList(new BugFixingInstrumentation(), new EmptyTypeInstrumentation());
```
mateuszrzeszutek, line 42: Can you remove `EmptyTypeInstrumentation`? It's not needed anymore, since there's an actual type instrumentation in this module.

mateuszrzeszutek (approved):

instrumentation/log4j/log4j-2.13.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_13_2/BugFixingInstrumentation.java
anuraaga: I've always wondered what this means - the javadoc says all exceptions are suppressed automatically, what does this syntax do?

instrumentation/log4j/log4j-2.13.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_13_2/Log4j2InstrumentationModule.java
anuraaga: The bugfix only applies to one of the codepaths in log4j, if it's using the other one it won't ever be applied so still need it.

instrumentation/log4j/log4j-2.13.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_13_2/BugFixingInstrumentation.java
mateuszrzeszutek: That's a good question, I've never checked the javadocs and just thoughtlessly assumed that this is required.

I read a bit through the ByteBuddy code and I think I know why we add it everywhere: ByteBuddy uses a different `SuppressionHandler` when you specify `suppress`, one that calls our handler: `ExceptionHandlers.EXCEPTION_STACK_HANDLER` - which just logs the error.

breedx-splk:
It's kinda bonkers to me to include a bugfix by way of instrumentation, but it seems only helpful and also allows the instrumentation to, you know, actually work.  🤘

Issue #2356, opened 2021-02-20
KubernetesClient instrumentation NPE


lshangq:
**Describe the bug**
I get this exception in  KubernetesClientTracer because  getSetter() return null. Should it be implement as a empty function, because okhttp client has added TracingInterceptor

```
java.lang.NullPointerException: setter
	at java.util.Objects.requireNonNull(Objects.java:228) ~[na:1.8.0_232]
	at io.opentelemetry.javaagent.shaded.io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator.inject(W3CTraceContextPropagator.java:102) ~[na:0.16.0-mi0.0.0-SNAPSHOT]
	at io.opentelemetry.javaagent.shaded.io.opentelemetry.context.propagation.MultiTextMapPropagator.inject(MultiTextMapPropagator.java:49) ~[na:0.16.0-mi0.0.0-SNAPSHOT]
	at io.opentelemetry.javaagent.instrumentation.kubernetesclient.KubernetesClientTracer.startSpan(KubernetesClientTracer.java:43) ~[na:na]
	at io.opentelemetry.javaagent.instrumentation.kubernetesclient.TracingInterceptor.intercept(TracingInterceptor.java:22) ~[na:na]
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142) ~[okhttp-3.14.9.jar!/:na]
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:117) ~[okhttp-3.14.9.jar!/:na]
	at io.opentelemetry.javaagent.instrumentation.okhttp.v3_0.TracingInterceptor.intercept(TracingInterceptor.java:31) ~[na:na]
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:142) ~[okhttp-3.14.9.jar!/:na]
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:117) ~[okhttp-3.14.9.jar!/:na]
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:229) ~[okhttp-3.14.9.jar!/:na]
	at okhttp3.RealCall.execute(RealCall.java:81) ~[okhttp-3.14.9.jar!/:na]
[...]

Issue #1012, opened 2020-08-16
Write integration tests for kubernetes-client


trask:
Currently it only has unit tests.

Wondering if we can write tests for it similar to the `aws-sdk` tests by standing up a fake endpoint and having `aws-sdk` talk to it.

Also, it looks like `kubernetes-client` uses `okhttp3`, so this would allow us to test with `:instrumentation:okhttp:okhttp-3.0` and verify that we aren't getting duplicate `CLIENT` spans.

PR #2416, opened 2021-02-25
Rewrite kubernetes-client instrumentation so that it works

mateuszrzeszutek:
(and add tests)

Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2356
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1012


instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientInstrumentationModule.java
```diff
@@ -37,40 +43,97 @@ public KubernetesClientInstrumentationModule() {
...
+        @Advice.Argument(0) Call httpCall,
+        @Advice.Local("otelContext") Context context,
+        @Advice.Local("otelScope") Scope scope) {
+      Context parentContext = Java8BytecodeBridge.currentContext();
+      if (!tracer().shouldStartSpan(parentContext)) {
+        return;
+      }
+
+      Request request = httpCall.request();
+      context = tracer().startSpan(parentContext, request, request.newBuilder());
```
trask, line 83: I don't follow what we are doing with the instance from `newBuilder`?

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientTracer.java
```diff
@@ -73,24 +59,42 @@ protected String requestHeader(Request request, String name) {
...
 
   @Override
-  protected String responseHeader(Response response, String name) {
-    return response.header(name);
+  protected String responseHeader(ApiResponse<?> response, String name) {
+    Map<String, List<String>> responseHeaders =
+        response.getHeaders() == null ? Collections.emptyMap() : response.getHeaders();
+    return responseHeaders.getOrDefault(name, Collections.emptyList()).stream()
+        .findFirst()
+        .orElse(null);
```
trask, line 67: this is not a very convenient api 🤷‍♂️

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/TracingApiCallback.java
```diff
@@ -0,0 +1,55 @@
...
+import io.kubernetes.client.openapi.ApiResponse;
+import io.opentelemetry.context.Context;
+import java.util.List;
+import java.util.Map;
+
+public class TracingApiCallback<T> implements ApiCallback<T> {
+  private final ApiCallback<T> delegate;
+  private final Context context;
+
+  public TracingApiCallback(ApiCallback<T> delegate, Context context) {
```
trask, line 21: should we pass in `parentContext` also for running the callbacks?

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientInstrumentationModule.java
mateuszrzeszutek: Inject the context into HTTP headers - `Request` is immutable.

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientTracer.java
anuraaga: I'd do `return !values.isEmpty() ? values.get(0) : null` instead of streaming

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientInstrumentationModule.java
mateuszrzeszutek: Good catch - I should have been more cautious when copying code from the OkHttp instrumentation 😂

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/TracingApiCallback.java
mateuszrzeszutek: Good point, done.

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientInstrumentationModule.java
mateuszrzeszutek: Fixed - I've also modified tests to verify that propagation works as expected.

instrumentation/kubernetes-client-7.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kubernetesclient/KubernetesClientTracer.java
mateuszrzeszutek: Null safety in Java is fun in general 😂

trask (approved):
👍

PR #2418, opened 2021-02-25
Remove some deprecated BaseTracer#end(Span) usages (part 2)

mateuszrzeszutek:


instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java
```diff
@@ -76,13 +76,15 @@ public static void onEnter(
 
       MessageDestination messageDestination =
           tracer().extractDestination(message, defaultDestination);
-      span = tracer().startProducerSpan(messageDestination, message);
-      scope = tracer().startProducerScope(span, message);
+      context = tracer().startProducerSpan(messageDestination, message);
+      // TODO: why are we propagating context only in this advice class? the other one does not
+      // inject current span context into JMS message
```
trask, line 81: oh, good find, can you open an issue for this?

trask (approved):

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java
mateuszrzeszutek: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2420

iNikem (approved):

PR #2421, opened 2021-02-26
Use dynamic resolution of SQS to allow working without it.

### LINKED ISSUE #2408: AWS SDK V1 instrumentation doesn't apply if SQS isn't on the classpath
AWS SDK v1 is highly modular so it's common for users to only have a subset of the services on the classpath. Unfortunately, muzzle currently blocks the whole instrumentation

```
[opentelemetry.auto.trace 2021-02-25 08:03:29:958 +0000] [main] WARN muzzleMatcher - Instrumentation skipped, mismatched references were found: aws-sdk -- io.opentelemetry.javaagent.instrumentation.awssdk.v1_11.AwsSdkInstrumentationModule on org.springframework.boot.loader.LaunchedURLClassLoader@70d63e05
[opentelemetry.auto.trace 2021-02-25 08:03:29:959 +0000] [main] WARN muzzleMatcher - -- io.opentelemetry.javaagent.instrumentation.awssdk.v1_11.TracingRequestHandler:86 Missing class com.amazonaws.services.sqs.model.ReceiveMessageResult
[opentelemetry.auto.trace 2021-02-25 08:03:29:959 +0000] [main] WARN muzzleMatcher - -- io.opentelemetry.javaagent.instrumentation.awssdk.v1_11.TracingRequestHandler:51 Missing class com.amazonaws.services.sqs.model.SendMessageRequest
[opentelemetry.auto.trace 2021-02-25 08:03:29:959 +0000] [main] WARN muzzleMatcher - -- io.opentelemetry.javaagent.instrumentation.awssdk.v1_11.TracingRequestHandler:57 Missing class com.amazonaws.services.sqs.model.ReceiveMessageRequest
[opentelemetry.auto.trace 2021-02-25 08:03:29:959 +0000] [main] WARN muzzleMatcher - -- io.opentelemetry.javaagent.instrumentation.awssdk.v1_11.TracingRequestHandler:88 Missing class com.amazonaws.services.sqs.model.Message
```

I think someone might have suggested we have separate instrumentation module for SQS and I incorrectly thought it'd be ok to have it together 😅

anuraaga:
Using reflection is probably the worst possible solution but the simplest for now. Some ideas that come to mind for improvement.

- Add an annotation that can suppress classes from muzzle, similar to animal sniffer `IgnoreJreRequirement`. We'd be expected to thoroughly code review suppressed classes
- Have a system for defining classes that operate on plugins of the instrumented library. 

```java
interface PluginInstrumentation {
  boolean isEnabled();

  ... various methods
}
```

Muzzle would implement `isEnabled` to return `true` only if the references check out. And instrumentation would guard calls into "various methods" with `isEnabled`. Muzzle could probably also verify this guard is happening with some work.

Fixes #2408


anuraaga:
Implemented dynamic dispatch as a stopgap - it'd be nice if 1.0 doesn't have broken aws sdk v1 instrumentation :)

mateuszrzeszutek (approved):

mateuszrzeszutek:
I think that we could add a method annotation that both stops muzzle from analyzing it and wraps the method with `try {...} catch (LinkageError)` to make sure that the runtime won't blow up no matter what - we'd still need to make sure that there are appropriate `if`s guarding the annotated code, catching an error should be the last resort.

The plugin approach would probably be really complicated to implement (and understand)

PR #2177, opened 2021-02-02
cassandra table name should split out keyspace (#1298)

FrankSpitulski:
This addresses an issue in PR #1314 https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1314#issuecomment-771552383 to split the keyspace out from the table name if needed.

@mateuszrzeszutek

PR #2423, opened 2021-02-26
Use low cardinality span names in Cassandra instrumentation

mateuszrzeszutek:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2128

### DIFF
--- a/instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/CassandraDatabaseClientTracer.java
+++ b/instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/CassandraDatabaseClientTracer.java
@@ -18,6 +18,7 @@ import io.opentelemetry.javaagent.instrumentation.api.db.SqlStatementSanitizer;
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes.DbSystemValues;
 import java.net.InetSocketAddress;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 public class CassandraDatabaseClientTracer
     extends DatabaseClientTracer<Session, String, SqlStatementInfo> {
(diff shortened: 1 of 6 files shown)


iNikem (approved):

instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/CassandraDatabaseClientTracer.java
```diff
@@ -34,22 +35,27 @@ protected String getInstrumentationName() {
 
   @Override
   protected SqlStatementInfo sanitizeStatement(String statement) {
-    return SqlStatementSanitizer.sanitize(statement);
+    return SqlStatementSanitizer.sanitize(statement).mapTable(this::stripKeyspace);
```
trask, line 38: why strip keyspace? I think that's equivalent of schema, which should be included in `db.sql.table`:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes-for-specific-technologies

mateuszrzeszutek: It was added here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2177

@FrankSpitulski can you explain why we're stripping the keyspace?

FrankSpitulski: We already have the keyspace as a separate tag and stripping it makes it more consistent for searching. Without stripping you’d have to search for key.table and table if you wanted queries on a table.

trask: oh I missed that this is capturing `db.cassandra.table` and not `db.sql.table`

though the spec says the same thing for `db.cassandra.table`:

> The name of the primary table that the operation is acting upon, including the schema name (if applicable)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#cassandra

I created issue #2476 to discuss/resolve the difference between our impl and the spec, so we can ignore this issue for this PR 😄

trask (approved):

PR #2433, opened 2021-02-27
Update instrumentation names

### LINKED ISSUE #1992: The Tracer name (instrumentation name) should always uniquely identify the instrumentation module
This is related to PR #1979 and would be a change from the decision made in #1260 . 

It is important when debugging spans to know precisely what instrumentation module created the spans, especially when there are crazy classpaths in play in the end-user's application. For this to work, every instrumentation module should use a unique name for the Tracers that are created. So, when there are multiple versions of the servlet instrumentation, for example, the tracer name should reflect which one is being used.

trask:
Resolves #1992

~Note: needs test fixes from #2431~


jkwatson (approved):

iNikem (commented):
Weak approval :)

mateuszrzeszutek (approved):

trask:
> Weak approval :)

will wait for @anuraaga to make sure we have enough agreement on this change

PR #2434, opened 2021-02-27
Hide non-specd attributes for aws-sdk-2.2

### LINKED ISSUE #1874: Hide all span attributes not defined by spec behind an experimental flag

trask:
Closes #1874


instrumentation/external-annotations/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/extannotations/TraceAnnotationsInstrumentationModule.java
```diff
@@ -125,20 +126,19 @@ public AnnotatedMethodsInstrumentation() {
           TraceAdvice.class.getName());
     }
 
-    private static Set<String> configureAdditionalTraceAnnotations(Config config) {
-      String configString = config.getProperty(TRACE_ANNOTATIONS_CONFIG);
-      if (configString == null) {
+    private static Set<String> configureAdditionalTraceAnnotations(String include) {
```
iNikem, line 129:
```suggestion
    private static Set<String> configureAdditionalTraceAnnotations(String config) {
```

iNikem (approved):

trask:
Moving this back to draft status until we decide on #2444

anuraaga:
/cc @mateuszrzeszutek 

For this and #2467 I'm hoping they're temporary, and it's ok. I'd expect us to make these programmatic eventually though. After #2463 I'm planning on tackling aws-sdk-2.2 next, and have an `AwsSdkTracingBuilder.setEnableExperimentalAttributes` type of pattern in mind.

mateuszrzeszutek:
> I'm planning on tackling aws-sdk-2.2 next, and have an `AwsSdkTracingBuilder.setEnableExperimentalAttributes` type of pattern in mind.

That sounds great - and if we refactor all library instrumentations to use this pattern we might be able to remove `Config` from instrumentation-api

anuraaga (approved):

mateuszrzeszutek (approved):

Issue #1616, opened 2020-11-11
spotlessKotlin shouldn't prevent wildcard import?


anuraaga:
Not a kotlin expert, but FWIU, wildcard imports are relatively standard in kotlin for being the basis of kotlin DSLs, such as coroutines

iNikem:
[Koltin coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html) do not say anything about this

trask:
I have no objection to allowing wildcard imports in kotlin code

trask:
@anuraaga is it ok to close this since we only have one kotlin file and the [editorconfig issue is now fixed](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2440#issuecomment-787743323) (and we can revisit in the future when we have more kotlin)?

PR #2440, opened 2021-03-01
Update old name

trask:
There were still some references around to the old `io.opentelemetry.auto` name, including one in `GlobalIgnoresMatcher` that was probably impacting startup overhead by not ignoring all of the classes under `io.opentelemetry.javaagent`.


anuraaga (approved):

trask:
Btw, Intellij was adding kotlin wildcard imports, which spotless was then complaining about, so I snuck a change to editorconfig so Intellij would stop doing this. I'm also good with #1616 in the future too.

instrumentation/jaxws/jws-1.1/javaagent/src/test/java/io/opentelemetry/test/WebServiceFromInterface.java
```diff
@@ -3,8 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.instrumentation.jaxws.jws.v1_1;
+package io.opentelemetry.test;
 
+import io.opentelemetry.javaagent.instrumentation.jaxws.jws.v1_1.WebServiceDefinitionInterface;
```
mateuszrzeszutek, line 8: Should we move `WebServiceDefinitionInterface` out of the javaagent package as well?

mateuszrzeszutek (approved):

iNikem (approved):

PR #2443, opened 2021-03-01
Rename peer mapping config

trask:
Since it's not spec'd, I propose to flag it as experimental.

If this PR is accepted, I will submit second draft PR to update docs.


anuraaga (commented):
Hmm - I think this is an attribute designed to work around javaagent-ness (can't programatically set a peer name). Similar to our instrumentation.enabled properties, I guess we'd leave it non-experimental? (or make everything experimental including those).

trask:
👍 makes sense to me, closing this in favor of issue #2447

PR #2449, opened 2021-03-01
Use low cardinality span names in Couchbase instrumentation

mateuszrzeszutek:
Also move unit tests to a separate module, previously they passed purely by accident (necessary classes were injected by the javaagent)

Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2128


iNikem:
WDYM "by accident"? When I see "unit tests" and "injected classes" I got confused? :)

mateuszrzeszutek:
Before this PR if you run the test by itself it'd always fail with a `NoClassDefFoundError`. That's because the class is not present on the classpath, because `src/main` (that contains `CouchbaseQuerySanitizer`) is put into `otel.initializer.jar` and is not reachable in the test code.
However, when you run the full `test` task, the actual javaagent tests run before the unit test - and the javaagent injects all helper classes to the test classloader. In other words, this test can only pass if the javaagent tests run before it.

trask (approved):
> Also move unit tests to a separate module, previously they passed purely by accident

😱 👍

iNikem:
Ok, what about instead of introducing this strange-looking `javaagent-unittests` module we move that class and its test to `library` module?

mateuszrzeszutek:
> Ok, what about instead of introducing this strange-looking `javaagent-unittests` module we move that class and its test to `library` module?

Unfortunately right now the `SqlStatementSanitizer` is in the javaagent-api module (because of the bounded cache thing) so it can't be moved to library module.
And this `javaagent-unittests` module is not that strange, we have a couple of them already in this project.

trask:
> And this `javaagent-unittests` module is not that strange, we have a couple of them already in this project.

and documentation 👍 https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation.md#writing-java-agent-unit-tests

iNikem:
> And this `javaagent-unittests` module is not that strange, we have a couple of them already in this project.

Still strange :)

PR #2452, opened 2021-03-01
Remove deprecated BaseTracer#extractWithGlobalPropagators() methods

mateuszrzeszutek:


instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/ContextPayload.java
```diff
@@ -42,9 +34,8 @@ public ContextPayload(Map<String, String> context) {
     this.context = context;
   }
 
-  public static ContextPayload from(Span span) {
+  public static ContextPayload from(Context context) {
     ContextPayload payload = new ContextPayload();
-    Context context = Context.current().with(span);
     GlobalOpenTelemetry.getPropagators().getTextMapPropagator().inject(context, payload, SETTER);
```
jkwatson, line 39: there's still a global call in here...that seems like a problem.

mateuszrzeszutek: Yep - currently all `inject()` calls are using global propagators. That's going to be the subject for my next PR 😄

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/ContextPayload.java
```diff
@@ -24,12 +22,6 @@
 
   private static final Logger log = LoggerFactory.getLogger(ContextPayload.class);
 
-  private static final Tracer TRACER = GlobalOpenTelemetry.getTracer("io.opentelemetry.auto.rmi");
```
jkwatson, line 27: this is a good change, but I don't see how it's related to the topic of the PR.

mateuszrzeszutek: It was just a bit of unused code - I thought that removing it in a separate PR would be a bit of excess of form, but you're absolutely right that it't not related to this PR.

jkwatson (approved):

trask (approved):

anuraaga (approved):

iNikem (approved):

PR #2453, opened 2021-03-01
Add BaseTracer#inject() method

mateuszrzeszutek:
And avoid using GlobalOpenTelemetry#getPropagators()

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -16,6 +16,7 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.context.ContextKey;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.context.propagation.TextMapGetter;
+import io.opentelemetry.context.propagation.TextMapSetter;
 import io.opentelemetry.instrumentation.api.InstrumentationVersion;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.instrumentation.api.context.ContextPropagationDebug;
@@ -74,10 +75,6 @@ public abstract class BaseTracer {
   public BaseTracer(OpenTelemetry openTelemetry) {
     this.tracer = openTelemetry.getTracer(getInstrumentationName(), getVersion());
     this.propagators = openTelemetry.getPropagators();
-  }
-
-  public ContextPropagators getPropagators() {
-    return propagators;
   }
 
   public Context startSpan(Class<?> clazz) {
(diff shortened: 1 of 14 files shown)


trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -253,6 +250,10 @@ public void addThrowable(Span span, Throwable throwable) {
     return propagators.getTextMapPropagator().extract(Context.root(), carrier, getter);
   }
 
+  public <C> void inject(Context context, C carrier, TextMapSetter<C> setter) {
```
jkwatson, line 253: public method in a public API should have some javadoc, I think.

jkwatson (approved):
would like some javadoc on the new method, but otherwise, 👍

anuraaga (approved):

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
```diff
@@ -110,10 +110,7 @@ public HttpHost getTarget() {
     @Override
     public HttpRequest generateRequest() throws IOException, HttpException {
       HttpRequest request = delegate.generateRequest();
-      tracer()
-          .getPropagators()
-          .getTextMapPropagator()
-          .inject(context, request, tracer().getSetter());
+      tracer().inject(context, request, tracer().getSetter());
```
iNikem, line 113: This looks like an overload is missing.

iNikem (approved):

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientInstrumentation.java
mateuszrzeszutek: Nice catch! Fixed.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
mateuszrzeszutek: Definitely. I'm planning to make a separate PR just to add Javadocs to `BaseTracer` sometime in the near future.

Issue #2456, opened 2021-03-01
Add test for otel.instrumentation.common.db-statement-sanitizer.enabled


trask:
This is a public configuration property, so we should have at least one test for it.

MartinDai:
Can I try it?

trask:
> Can I try it?

sure!

PR #2457, opened 2021-03-01
Rename two props to otel.instrumentation.common.*

### LINKED ISSUE #2447: Rename otel.endpoint.peer.service.mapping to match other instrumentation properties?
It should probably start with either `otel.instrumentation.*` or `otel.javaagent.*` to match other properties.

Based on discussion so far in #2444, probably should start with `otel.instrumentation.`

Maybe under a "common" instrumentation node, e.g.
`otel.instrumentation.common.peer-service-mapping` to note that it's shared across multiple instrumentations?

We could use similar for `otel.instrumentation.common.db-statement-sanitizer.enabled`.

trask:
Closes #2447 

Missing test logged to #2456


anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

trask:
I will send a follow-up draft doc PR

PR #2462, opened 2021-03-02
Use atomic updaters instead of atomic fields to reduce allocations fo…

anuraaga:
…r commonly created objects.

Randomly found this when looking though datadog changes and think it's easy to add here. It's a bit of extra code but otherwise is free memory (very commonly used pattern in Netty).

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
@@ -19,7 +19,7 @@ public class AppServerBridge {
       ContextKey.named("opentelemetry-servlet-app-server-bridge");
 
   private static final AtomicIntegerFieldUpdater<AppServerBridge>
-      servlerUpdatedServerSpanNameUpdater =
+      servletUpdatedServerSpanNameUpdater =
           AtomicIntegerFieldUpdater.newUpdater(
               AppServerBridge.class, "servletUpdatedServerSpanName");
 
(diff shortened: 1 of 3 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
```diff
@@ -42,42 +50,30 @@ public static Context init(Context ctx, boolean shouldRecordException) {
...
-   *     <code>false</code> otherwise.
+   * @return <code>true</code>, if the server span name had not already been updated by servlet
+   *     integration, or <code>false</code> otherwise.
    */
-  public static boolean shouldUpdateServerSpanName(Context ctx) {
+  public static boolean setUpdatedServerSpanName(Context ctx) {
     AppServerBridge appServerBridge = ctx.get(AppServerBridge.CONTEXT_KEY);
     if (appServerBridge != null) {
-      return !appServerBridge.servletUpdatedServerSpanName.get();
+      return servlerUpdatedServerSpanNameUpdater.compareAndSet(appServerBridge, FALSE, TRUE);
```
anuraaga, line 72: This previously wasn't using CAS but I presume it wanted to, or otherwise I could just change to a simple `volatile`.

mateuszrzeszutek (approved):

mateuszrzeszutek:
That's pretty nice, I didn't know about those field updater classes before 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/AppServerBridge.java
```diff
@@ -18,6 +18,14 @@
   private static final ContextKey<AppServerBridge> CONTEXT_KEY =
       ContextKey.named("opentelemetry-servlet-app-server-bridge");
 
+  private static final AtomicIntegerFieldUpdater<AppServerBridge>
+      servlerUpdatedServerSpanNameUpdater =
```
jkwatson, line 22: type: `servler`

trask (approved):

iNikem:
Just a note for the future: I think that if we propose some performance optimisations, we have to accompany them with benchmarking results demonstrating the actual effect. In general it's too easy to be wrong about performance :)

trask:
> Just a note for the future: I think that if we propose some performance optimisations, we have to accompany them with benchmarking results demonstrating the actual effect. In general it's too easy to be wrong about performance :)

I think it's ok to apply some judgement. Some optimizations are not so hard to reason about (like this one), and I wouldn't want to make contributing them harder than necessary with a blanket rule.

iNikem:
Agree on blanket rule

PR #2467, opened 2021-03-02
Initialize Config for library instrumentations

mateuszrzeszutek:
Extracted from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2434

It turns out that this `Config` fix is needed in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2263


trask (approved):
approving in case this is needed as temporary measure, otherwise it sounds like there's agreement with @anuraaga proposal that library instrumentation will only support programmatic configuration

anuraaga:
Yeah - either way moving back to programmatic will be easier once there's a flag in the first place :)

anuraaga (approved):

anuraaga:
Should we merge this or just #2434?

trask:
> Should we merge this or just #2434?

Sounds good, I marked it ready

iNikem (approved):

PR #2474, opened 2021-03-03
Refactor gRPC library instrumentation to expose minimal API.

anuraaga:


instrumentation/grpc-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_5/GrpcClientBuilderBuildInstrumentation.java
```diff
@@ -46,16 +45,7 @@
...
-      for (ClientInterceptor interceptor : interceptors) {
-        if (interceptor instanceof TracingClientInterceptor) {
-          shouldRegister = false;
-          break;
-        }
-      }
-      if (shouldRegister) {
-        interceptors.add(0, TracingClientInterceptor.newInterceptor());
-      }
+      interceptors.add(0, GrpcInterceptors.CLIENT_INTERCEPTOR);
```
anuraaga, line 48: We aren't focusing on duplicate instrumentation yet so I just removed what probably wasn't working anyways due to shading.

instrumentation/grpc-1.5/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_5/GrpcTracingBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.grpc.v1_5;
+
+import io.opentelemetry.api.OpenTelemetry;
+
+/** A builder of {@link GrpcTracing}. */
+public final class GrpcTracingBuilder {
```
anuraaga, line 11: Note, I proactively went with a builder for this single property here because unlike SDK, where we generally proactively define functionality, with instrumentation it's more common to be reactive and require different settings.

mateuszrzeszutek (approved):

pavolloffay (approved):

PR #2477, opened 2021-03-03
Add a caching API based on caffeine for use from instrumentation, not just javaagent

anuraaga:
For now just adds the most minimal API, `computeIfAbsent`.


anuraaga:
@ben-manes After I experimented with gradle-shadow-plugin's minimization feature, I found that it's only about a 350K dependency to add caffeine for our purposes, so we're going with it instead of the concurrentlinkedhashmap. If you have a chance to skim this to see if it's a sane approach, that would be much appreciated.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/caching/CacheTest.java
```diff
@@ -0,0 +1,121 @@
...
+      // Wait for GC to be reflected.
+      await()
+          .untilAsserted(
+              () -> {
+                caffeineCache.cleanup();
+                assertThat(caffeineCache.keySet()).hasSize(1);
+              });
+      assertThat(cache.computeIfAbsent(dog, unused -> "bark")).isEqualTo("bark");
+      dog = null;
+      System.gc();
```
ben-manes, line 82: GcFinalizable from guava testlib is great for this as more predictable

ben-manes:
That’s great!

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/caching/CacheTest.java
anuraaga: Thanks for the hint! I've always wondered about this but in practice have yet to see a failure due to only calling `gc` once and continue with this pattern. Looking at it

https://github.com/google/guava/blob/master/guava-testlib/src/com/google/common/testing/GcFinalization.java#L186

I feel as if it would be mostly the same to move `System.gc` into the await() predicate, possibly tweaking the polling interval, which is nice to reduce the number of ways of doing something. Might do one or the other if this flakes in the future.

javaagent-bootstrap/javaagent-bootstrap.gradle
```diff
@@ -7,6 +7,18 @@ apply from: "$rootDir/gradle/publish.gradle"
...
 
+// patch inner class from Caffeine to avoid ForkJoinTask from being loaded too early
+sourceSets {
+  patch {
+    java {}
+  }
+}
+jar {
+  from(sourceSets.patch.output) {
+    include 'com/github/benmanes/caffeine/cache/BoundedLocalCache$PerformCleanupTask.class'
```
mateuszrzeszutek, line 18: Will this file be shaded too? Also, why `javaagent-bootstrap` and not the new module?

instrumentation-api-caching/src/main/java/io/opentelemetry/instrumentation/api/caching/CacheBuilder.java
```diff
@@ -0,0 +1,39 @@
...
+  public CacheBuilder setWeakKeys() {
+    caffeine.weakKeys();
+    return this;
+  }
+
+  /** Returns a new {@link Cache} with the settings of this {@link CacheBuilder}. */
+  public <K, V> Cache<K, V> build() {
+    @SuppressWarnings("unchecked")
+    com.github.benmanes.caffeine.cache.Cache<K, V> delegate =
+        (com.github.benmanes.caffeine.cache.Cache<K, V>) caffeine.build();
```
mateuszrzeszutek, line 34: Why not add `<K, V>` as type parameters of `CacheBuilder`?

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/caching/CacheTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+      CaffeineCache<?, ?> caffeineCache = ((CaffeineCache<?, ?>) cache);
+      assertThat(cache.computeIfAbsent("cat", unused -> "meow")).isEqualTo("meow");
+      assertThat(caffeineCache.keySet()).hasSize(1);
+
+      assertThat(cache.computeIfAbsent("cat", unused -> "bark")).isEqualTo("meow");
+      assertThat(caffeineCache.keySet()).hasSize(1);
+
+      assertThat(cache.computeIfAbsent("dog", unused -> "bark")).isEqualTo("bark");
+      caffeineCache.cleanup();
```
mateuszrzeszutek, line 46: Just wondering: what does `cleanup()` exactly do here, does it remove values over max size?
Does this test pass without it? Would adding it to `unbounded()` make that test fail?

instrumentation-api-caching/src/main/java/io/opentelemetry/instrumentation/api/caching/CacheBuilder.java
anuraaga: Yeah Caffeine has them - since we don't have loading cache at least for now I couldn't find a reason to have them on the builder. This is the same trick Caffeine uses to convert `Caffeine<Object, Object>` to `Cache<K, V>`.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/caching/CacheTest.java
anuraaga: Yup it ensures values over max are removed since Caffeine does that asynchronously. Without it out fails, but for non evictions it's not needed

javaagent-bootstrap/javaagent-bootstrap.gradle
anuraaga: Oh good catch, I don't think this is doing anything because of the non shading (I'll need to come up with a bespoke test for this I guess). It's here since only the agent cares, not library instrumentation.

anuraaga: For additional reference there is a gap here vs datadog - they keep Caffeine in the agent classloader but we are shading and putting in the bootstrap classloader instead to allow the same instrumentation to work for both library and agent.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/caching/CacheTest.java
ben-manes: If you prefer, set the executor to Runnable::run for a direct call. The async isn’t needed for the cache’s own maintenance as that is fast, but we have optional callbacks to user code so it’s protected if that is slow.

javaagent-bootstrap/javaagent-bootstrap.gradle
anuraaga: I fixed the package name but couldn't come up with a way to add a test for this. Anyone have any ideas?

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/caching/CacheTest.java
anuraaga: @ben-manes Thanks for the tip! Going to leave it as is for now to use the defaults as a baseline. Will set up a simple benchmark for our own sanity checking later, compare the two, and then pick one.

javaagent-bootstrap/javaagent-bootstrap.gradle
anuraaga: Was able to come up with a test, a bit annoying since I need a different Gradle project to be able to check the patched jar itself.

mateuszrzeszutek (approved):
Nice! 🎉

trask (approved):
nice 👍

Issue #2469, opened 2021-03-02
Does tracer name in spring-boot-autoconfigure need to be configurable?


mateuszrzeszutek:
`SpringWebMvcServerTracer` is one of the few classes left that still use the deprecated `BaseTracer(Tracer)` constructor. It turns out that it's used by the spring-boot-autoconfigure library instrumentation that actually constructs a `Tracer` with a custom instrumentation name and uses the constructor to pass it.
Is this custom instrumentation name functionality necessary? Can we remove it and pass `OpenTelemetry` instead of a `Tracer`?

trask:
👍 on removing it and passing `OpenTelemetry`

PR #2478, opened 2021-03-03
Remove tracer name configuration from spring-boot-autoconfigure

mateuszrzeszutek:
* Use OpenTelemetry wherever it makes sense
* Refactor @WithSpan advice since it made no sense before
* Rename @EnableOpenTelemetryTracing to @EnableOpenTelemetry (it won't be just tracing in the future)
* And add a getOpenTelemetry() method to InstrumentationExtension & InstrumentationSpecification

Closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2469

This PR turned out a bit bigger than I expected it to, but our spring-boot-autoconfigure module was *really* out-of-date. I'll create 3 more issues to fix it completely (I guess?).


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTracer.java
```diff
@@ -0,0 +1,52 @@
...
+        return withClientSpan(parentContext, span);
+      default:
+        return parentContext.with(span);
+    }
+  }
+
+  private String spanName(WithSpan annotation, Method method) {
+    String spanName = annotation.value();
+    if (spanName.isEmpty()) {
+      return method.getDeclaringClass().getSimpleName() + "." + method.getName();
```
trask, line 48:
```suggestion
      return spanNameForMethod(method);
```

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfigurationTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+import org.springframework.context.annotation.Bean;
+
+/** Spring Boot auto configuration test for {@link OpenTelemetryAutoConfiguration}. */
+class OpenTelemetryAutoConfigurationTest {
+  @TestConfiguration
+  static class CustomTracerConfiguration {
+    @Bean
+    public OpenTelemetry customOpenTelemetry() {
+      return OpenTelemetry.noop();
```
trask, line 26: noop in the wild 😁 cc: @jkwatson

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
```diff
@@ -39,99 +43,191 @@ public String testWithSpanWithValue() {
...
+                trace
+                    .hasSize(2)
+                    .hasSpansSatisfyingExactly(
+                        parentSpan -> parentSpan.hasName("parent").hasKind(INTERNAL),
+                        span ->
+                            span.hasName("WithSpanTester.testWithSpan")
+                                .hasKind(INTERNAL)
+                                // otel SDK assertions need some work before we can comfortably use
+                                // them in this project...
+                                .hasParentSpanId(traces.get(0).get(0).getSpanId())));
```
trask, line 110: cc: @anuraaga 👍

instrumentation/spring/spring-web-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/httpclients/RestTemplateInterceptor.java
```diff
@@ -19,25 +17,26 @@
...
 
-    Context context = tracer().startSpan(parentContext, request, request.getHeaders());
+    Context context = tracer.startSpan(parentContext, request, request.getHeaders());
     try (Scope ignored = context.makeCurrent()) {
       ClientHttpResponse response = execution.execute(request, body);
-      tracer().end(context, response);
+      tracer.end(context, response);
       return response;
     }
+    // TODO: endExceptionally?
```
trask, line 40: 👍

trask (commented):
Thanks! 😅

trask (approved):

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
```diff
@@ -39,99 +43,191 @@ public String testWithSpanWithValue() {
...
+        "parent",
+        () -> {
+          withSpanTester.testWithSpan();
+          return null;
+        });
+
+    // then
+    List<List<SpanData>> traces = instrumentation.waitForTraces(1);
+    assertThat(traces)
+        .hasSize(1)
```
anuraaga, line 98: Don't need `hasSize` for either since the next check is `Exactly`

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
```diff
@@ -39,99 +43,191 @@ public String testWithSpanWithValue() {
...
+        .hasTracesSatisfyingExactly(
+            trace ->
+                trace
+                    .hasSize(2)
+                    .hasSpansSatisfyingExactly(
+                        parentSpan -> parentSpan.hasName("parent").hasKind(INTERNAL),
+                        span ->
+                            span.hasName("WithSpanTester.testWithSpan")
+                                .hasKind(INTERNAL)
+                                // otel SDK assertions need some work before we can comfortably use
```
anuraaga, line 108: Will `hasParentSpanIndex`, the index within the trace, solve it?

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/LibraryTestRunner.java
```diff
@@ -70,6 +70,11 @@ public void clearAllExportedData() {
     forceFlushCalled = false;
   }
 
+  @Override
+  public OpenTelemetrySdk getOpenTelemetry() {
```
anuraaga, line 74: We should almost never need the SDK in tests (instrumentation is supposed to avoid the SDK) so I think the old name was better when the return type is `OpenTelemetrySdk`. So I figured for the rare cases (aws lambda, spring) where we need the SDK, calling into `testRunner.openTelemetrySdk` in library tests was OK, while in the more common case we had `LibraryTestTrait.getOpenTelemetry()`. Do we need to make a change here or is it ok to revert it?

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
mateuszrzeszutek: Yes, I think it will.

I initially thought that we could use the same pattern as in the javaagent asserts: `hasParentSpan(trace.span(0))` but maybe just the index is better.
On a slightly related topic: what about asserting on span links? We would have to pass the trace index there too, ex. `hasLinkIndex(traceIndex, spanIndex)`

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/LibraryTestRunner.java
mateuszrzeszutek: I've added back the `getOpenTelemetrySdk()` method to the library runner and reverted back the AWS tests.

trask:
(merged main into this PR to get the `apache-camel-2.20` test fixes)

anuraaga (approved):

Issue #2442, opened 2021-03-01
Consolidate otel.exporter.jar and otel.initializer.jar into single feature otel.javaagent.extensions?


trask:
`otel.javaagent.extensions` could be a single jar file, a list of jar files, or a folder with jar files

It would behave same as `otel.initializer.jar` where those jar files just get added to `AgentClassLoader` (or even better follow @jsuereth's proposal of loading them in a mostly-isolated child of `AgentClassLoader`).

And all of the normal autoconfigure SPIs can be used to add exporters, span processors, propagators.

I like the runtime shading of the existing `ExporterClassLoader`, but would probably vote to do away with it due to startup overhead, and make sure we document how to shade properly using both gradle and maven.

trask:
Adding note here about benefits of runtime shading: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/2570#discussioncomment-492019

trask:
Also note that as part of this we will drop `io.opentelemetry.javaagent.tooling.exporter.SpanExporterFactory`, and only support `io.opentelemetry.sdk.autoconfigure.spi.ConfigurableSpanExporterProvider` from `io.opentelemetry:opentelemetry-sdk-extension-autoconfigure`

tobias-:
Nit: They're `otel.javaagent.experimental.exporter.jar` and `otel.javaagent.experimental.initializer.jar`

sumitm-iiit:
@trask : I was trying out "otel.javaagent.experimental.initializer.jar" system property. It seems to work when i provide the path of the exact jar file. But when i provide the path of a folder that contains multiple jars, it does not seem to load those jars. Is that something expected as based on your description above it seems its expected to work with folders containing multiple jars as well.

trask:
> when i provide the path of a folder that contains multiple jars, it does not seem to load those jars

ya, this has not been implemented yet, but is planned

sumitm-iiit:
@trask : Any idea by when it would be implemented.

iNikem:
> @trask : Any idea by when it would be implemented.

During the June

trask:
this is done! please see https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/examples/extension#introduction

sumitm-iiit:
@trask : Thanks for notifying.

PR #2486, opened 2021-03-04
Rename more properties

trask:
Make `otel.exporter.jar` and `otel.initializer.jar` experimental because we are planning to replace with `otel.javaagent.extensions` soon (#2442):
* `otel.exporter.jar` -> `otel.javaagent.experimental.exporter.jar`
* `otel.initializer.jar` -> `otel.javaagent.experimental.initializer.jar`

Put all instrumentation settings that are not specific to one `[id]` under `otel.instrumentation.common.*`:
* `otel.instrumentation.default-enabled` -> `otel.instrumentation.common.default-enabled`

Follow existing convention of using `.enabled`:
* `otel.instrumentation.kafka.client-propagation` -> `otel.instrumentation.kafka.client-propagation.enabled`

Rename to better name(?):
* `otel.instrumentation.spring-batch.chunk.root-span` -> `otel.instrumentation.spring-batch.chunk.new-trace`

I think these are experimental(?):
* `otel.javaagent.runtime.context.field.injection` -> `otel.javaagent.experimental.field-injection.enabled`
* `otel.threadPropagationDebugger` -> `otel.javaagent.experimental.thread-propagation-debugger.enabled`

Use `otel.javaagent.testing.*` for settings used during testing only:
* `internal.testing.disable.global.library.ignores` -> `otel.javaagent.testing.additional-library-ignores.enabled` (and flipped)
* `otel.internal.failOnContextLeak` -> `otel.javaagent.testing.fail-on-context-leak`
* `otel.internal.enableTransformSafeLogging` -> `otel.javaagent.testing.transform-safe-logging.enabled`

---------------------

Note to self: I owe two follow-up PRs for this:
* Documentation updates
* Example distro updates (can't be merged until after release so the distro can be updated to latest release)


instrumentation/spring/spring-batch-3.0/javaagent/spring-batch-3.0-javaagent.gradle
```diff
@@ -30,7 +30,7 @@ test.finalizedBy(tasks.register("testChunkRootSpan", Test) {
   filter {
     includeTestsMatching '*ChunkRootSpanTest'
   }
-  jvmArgs '-Dotel.instrumentation.spring-batch.chunk.root-span=true'
+  jvmArgs '-Dotel.instrumentation.spring-batch.chunk.new-trace=true'
```
anuraaga, line 33: Strong +1 to new-trace always much more grokkable than talking about root spans.

anuraaga (approved):

instrumentation/spring/spring-batch-3.0/javaagent/spring-batch-3.0-javaagent.gradle
mateuszrzeszutek: Should we add `experimental` to this property?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
```diff
@@ -22,9 +22,9 @@
       ContextKey.named("thread-propagation-locations");
 
   private static final boolean THREAD_PROPAGATION_DEBUGGER =
-      Boolean.getBoolean("otel.threadPropagationDebugger");
+      Boolean.getBoolean("otel.javaagent.experimental.thread-propagation-debugger.enabled");
```
mateuszrzeszutek, line 25: What do you think about `otel.javaagent.debug.thread-propagation.enabled` instead? We could also set its default value to `Config.isAgentDebugEnabled()`

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

PR #2491, opened 2021-03-04
Use ClassValue instead of WeakMap to cache span name

laurit:
As values stored in `WeakMap` may reference the key used (key is class, value may contain method of the same class which has a reference to class) it does not serve intended purpose.


anuraaga (approved):

mateuszrzeszutek (approved):

trask (approved):
> As values stored in `WeakMap` may reference the key used (key is class, value may contain method of the same class which has a reference to class) it does not serve intended purpose.

😱 💯

PR #2500, opened 2021-03-04
Updates to thread propagation debug property

trask:
Based on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2486#discussion_r587364869


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
```diff
@@ -22,9 +23,13 @@
       ContextKey.named("thread-propagation-locations");
 
   private static final boolean THREAD_PROPAGATION_DEBUGGER =
-      Boolean.getBoolean("otel.javaagent.experimental.thread-propagation-debugger.enabled");
+      Config.get()
+          .getBooleanProperty(
+              "otel.javaagent.debug.thread-propagation.enabled",
```
anuraaga, line 28: Don't we need the experimental still? If I'm not mistaken, actually this flag is supposed to go away at some point since the SDK includes a similar flag now.

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/context/ContextPropagationDebug.java
trask: makes sense, reverted name, but kept the new default

mateuszrzeszutek (approved):

Issue #2502, opened 2021-03-05
Request for OTel support in AWS Java SDK v1.x 


zaheershaik:
We are trying to manually instrument our API calls to DyanamoDB and SQS, but noticed that there is no  OTel support for AWS Java SDK 1.x version, unlike [2.x versions like this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/aws-sdk/aws-sdk-2.2/library). It is not an easy task to migrate all the older versions of DDB calls from 1.x to 2.x version. It'd really help to have support for this older version of SDK.

anuraaga:
Hi @zaheershaik - this is definitely on our radar, I've started to prepare for it with #2508

PR #2508, opened 2021-03-05
Use reflection to get request metadata instead of instrumentation.

anuraaga:
Similar to what we do in aws sdk v2. For #2502, while we may need to keep some of the special javaagent instrumentation to work around bugs that will not be workaroundable in library instrumentation, the `RequestInstrumentation` is the one that has the largest impact on library instrumentation right now.


instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/Aws1ClientTest.groovy
```diff
@@ -258,6 +258,7 @@ class Aws1ClientTest extends AgentInstrumentationSpecification {
             "aws.endpoint" "https://s3.amazonaws.com"
             "aws.operation" "HeadBucket"
             "aws.agent" "java-aws-sdk"
+            "aws.bucket.name" "someBucket"
```
anuraaga, line 261: And somehow improved coverage!

mateuszrzeszutek (approved):

trask (approved):

PR #2512, opened 2021-03-05
Add support for Grails

laurit:
- set server span name to /context/controller/action
- create span for grails controller invocations


instrumentation/grails-3.0/javaagent/src/test/groovy/test/GrailsTest.groovy
```diff
@@ -0,0 +1,155 @@
...
+  boolean hasExceptionOnServerSpan() {
+    true
+  }
+
+  @Override
+  boolean hasResponseSpan(ServerEndpoint endpoint) {
+    endpoint == REDIRECT || endpoint == ERROR || endpoint == NOT_FOUND
+  }
+
+  boolean hasErrorPageSpans(ServerEndpoint endpoint) {
```
mateuszrzeszutek, line 97: Nit: add `@Override`

mateuszrzeszutek (approved):

instrumentation/grails-3.0/javaagent/grails-3.0-javaagent.gradle
```diff
@@ -0,0 +1,44 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+muzzle {
+  pass {
+    group = "org.grails"
+    module = "grails-web-url-mappings"
+    versions = "[3.0,)"
+    skipVersions += ['3.1.15']
+    assertInverse true
```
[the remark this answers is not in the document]
laurit: Thanks, it definitely wasn't intentional, but seems to work as far as I can tell. Will add the equals sign to be consistent with other usages of `assertInverse`

mateuszrzeszutek (approved):

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringWebMvcTracer.java
```diff
@@ -110,6 +119,18 @@ private void onRender(SpanBuilder span, ModelAndView mv) {
...
+  protected Throwable unwrapThrowable(Throwable throwable) {
+    if (throwable instanceof InvocationTargetException) {
+      return throwable.getCause();
+    }
+    return super.unwrapThrowable(throwable);
+  }
+
+  public void addUnwrappedThrowable(Span span, Throwable throwable) {
+    addThrowable(span, unwrapThrowable(throwable));
+  }
```
trask, line 132: I think it would be reasonable for addThrowable to always call unwrapThrowable, similar to endExceptionally (current impl is good for this PR) cc: @mateuszrzeszutek

trask (commented):
> set server span name to /context/controller/action

Is it possible to get and use the url-based route instead?

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springwebmvc/SpringWebMvcTracer.java
trask: nm, just reviewing #2513 now and see this is already happening there 👍

iNikem:
> > set server span name to /context/controller/action
> 
> Is it possible to get and use the url-based route instead?

I think that span name based on `controller/action` is a good one. But `/context/controller/action` mixes url-based name (context) with source code based (controller/action). Also it would be more idiomatic to have dot as separator: `conroller.action`

I don't have a good suggestion atm.

iNikem (approved):

trask:
> `/context/controller/action` mixes url-based name (context) with source code based (controller/action).
>
> I don't have a good suggestion atm.

Ya, me either, that's why I'm hoping there's a reasonable way to get and use the url-based route.

> Also it would be more idiomatic to have dot as separator: controller.action

Interestingly the [rpc semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#span-name) use `classname/methodname`

laurit:
It might be better to just accept that span name is something that identifies a trace. How exactly will depend on the circumstances, it might not be possible to always get a result that immediately makes sense.
For rest urls it is easy, there is a path that nicely describes the current operation, we only need to get rid of parameters in path. Using `className.methodName` is not that great, what if application has multiple classes with same name in different packages or there are multiple applications that have a class with same name? In case of paths context path helps to distinguish different applications, maybe we should use `/contextPath/className.methodName` instead? The downside would be that it is a bit less obvious how span name is constructed and if you only have a single application then context path is redundant anyway. Also when using class name it might happen that the actual controller class has been proxied and we end up with a name like `$Proxy.foo`. Sometimes neither path not controller are a good fix, for example in jax-ws framework integration I appended operation name to span name, this really is not a part of path, just something that is inside the xml that was submitted to the web service.
Grails tutorials use the following mapping
```
package demo

class UrlMappings {

    static mappings = {
        "/$controller/$action?/$id?(.$format)?"{
            constraints {
                // apply constraints here
            }
        }

        "/"(view:"/index")
        "500"(view:'/error')
        "404"(view:'/notFound')
    }
}
```
So I'd assume that `/context/controller/action` is actually a path that a lot of grails applications use because that seems to be the convention. For test I used a different mapping scheme just to have endpoints where `HttpServerTest` expects them to be. I agree that all of this is not ideal, but this seemed to be the easiest way to get span names that identify the operation. I guess could also use `controllerClass.methodName` but that doesn't have context path and then server span would have the same name as nested handler span. It should be possible to also get the matched pattern which for our tests would produce good names like `/context/success/*` but fail miserably for the mapping from tutorial and produce something like `/context/*/*/*`. With some effort it might be possible to replace controller and action wildcards in pattern with actual names and translate parameter wildcards back to variable names.

iNikem:
@laurit convinced me on the status quo

trask (approved):
> "/$controller/$action?/$id?(.$format)?"

oh my, I see why this is problematic

> With some effort it might be possible to replace controller and action wildcards in pattern with actual names and translate parameter wildcards back to variable names.

agree status quo is good, thanks for the explanations 👍

anuraaga (approved):

PR #2514, opened 2021-03-05
Add support for elasticsearch7

laurit:
Tweak elasticsearch test and instrumentation a bit so it would work for versions starting from 6.5

### DIFF
--- a/instrumentation/elasticsearch/elasticsearch-transport-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/transport/v6_0/Elasticsearch6TransportClientInstrumentationModule.java
+++ b/instrumentation/elasticsearch/elasticsearch-transport-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/transport/v6_0/Elasticsearch6TransportClientInstrumentationModule.java
@@ -11,6 +11,7 @@ import static java.util.Collections.singletonList;
 import static java.util.Collections.singletonMap;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
 import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.namedOneOf;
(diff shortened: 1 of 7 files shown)


mateuszrzeszutek (approved):
LGTM 👍 
Just wondering, what about the elasticsearch-rest-6.4 instrumentation?
It has this suspicious TODO, maybe similar test there changes will make it work?
```
  // TODO: The tests are incompatible with 7.x.  The instrumentation may be as well.
  latestDepTestImplementation group: 'org.elasticsearch.client', name: 'elasticsearch-rest-client', version: '6.+'
```

laurit:
I did look at it and while tests can be made to work the instrumentation also needs a bit of work because of https://github.com/elastic/elasticsearch/commit/a7046e001ce6b5397deee35c0a0e348291ffe034 Previously request methods ended up in `performRequestAsyncNoCatch` now sync and async methods need to be instrumented separately.

instrumentation/elasticsearch/elasticsearch-transport-6.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/transport/v6_0/Elasticsearch6TransportClientInstrumentationModule.java
```diff
@@ -57,7 +57,12 @@ public Elasticsearch6TransportClientInstrumentationModule() {
       return singletonMap(
           isMethod()
               .and(named("execute"))
-              .and(takesArgument(0, named("org.elasticsearch.action.Action")))
+              .and(
+                  takesArgument(
+                      0,
+                      namedOneOf(
+                          "org.elasticsearch.action.Action",
```
anuraaga, line 64: This sort of change happened on minor version 6.5? Geez

anuraaga (approved):
Luckily this is gone from 8.x+ so won't need more maintenance (probably)

iNikem (approved):

Issue #2479, opened 2021-03-03
RestTemplateInterceptor does not call endExceptionally() when throwable is thrown


mateuszrzeszutek:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/spring/spring-web-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/httpclients/RestTemplateInterceptor.java

PR #2516, opened 2021-03-05
Fix RestTemplateInterceptor so that it calls endExceptionally() on exception

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2479

I've added a few testing changes in this PR because I wouldn't be able to unit-test my change otherwise:
* implemented a new `TraceUtils` class in testing-common; contrary to the old one it uses `BaseTracer` to create spans, so we can now test span suppression;
* added mockito to dependencies.gradle

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -31,6 +31,7 @@ ext {
     prometheus        : "0.9.0",
     assertj           : '3.19.0',
     awaitility        : '4.0.3',
+    mockito           : '3.6.0',
     // Caffeine 2.x to support Java 8+. 3.x is 11+.
     caffeine          : '2.9.0',
     testcontainers    : '1.15.2'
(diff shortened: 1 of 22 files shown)


laurit:
Maybe we could also test it the same way as we do for other http clients.
```
import io.opentelemetry.api.GlobalOpenTelemetry
import io.opentelemetry.instrumentation.spring.httpclients.RestTemplateInterceptor
import io.opentelemetry.instrumentation.test.AgentTestTrait
import io.opentelemetry.instrumentation.test.base.HttpClientTest
import org.springframework.http.HttpEntity
import org.springframework.http.HttpHeaders
import org.springframework.http.HttpMethod
import org.springframework.http.client.SimpleClientHttpRequestFactory
import org.springframework.web.client.RestTemplate

class RestTemplateTest extends HttpClientTest implements AgentTestTrait {

  @Override
  int doRequest(String method, URI uri, Map<String, String> headers, Closure callback) {
    SimpleClientHttpRequestFactory clientHttpRequestFactory = new SimpleClientHttpRequestFactory()
    clientHttpRequestFactory.setConnectTimeout(CONNECT_TIMEOUT_MS)
    RestTemplate restTemplate = new RestTemplate(clientHttpRequestFactory)
    RestTemplateInterceptor restTemplateInterceptor = new RestTemplateInterceptor(GlobalOpenTelemetry.get())
    restTemplate.getInterceptors().add(restTemplateInterceptor)

    HttpHeaders httpHeaders = new HttpHeaders()
    headers.each {
      httpHeaders.add(it.key, it.value)
    }
    HttpEntity<?> requestEntity = new HttpEntity<>("body", httpHeaders)
    def status = restTemplate.exchange(uri, HttpMethod.resolve(method), requestEntity, String).getStatusCodeValue()
    callback?.call()
    return status
  }
}
```

trask (approved):
> Maybe we could also test it the same way as we do for other http clients.
👍

iNikem (approved):

mateuszrzeszutek:
> Maybe we could also test it the same way as we do for other http clients.

Yeah, definitely 🤦‍♂️  I don't know how I did not think of this earlier 😅

mateuszrzeszutek:
Done!
I've left only one unit test with mocks: the one that verifies that we do not create a seconds CLIENT span. This test should also be added to the `HttpClientTest`, but let's do it in another PR.

PR #2518, opened 2021-03-05
AWS SDK - S3 to SQS propagation test

kubawach:
- adds S3 to SQS notification propagation test (currently disabled - needs https://github.com/localstack/localstack/issues/3686 but works with live AWS)
- fixes small bug with AWS tracing header based propagation - context needs to be root() otherwise for messages without the attribute CONSUMER spans will get CLIENT as their parent (!)

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/SqsParentContext.java
+++ b/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/SqsParentContext.java
@@ -34,7 +34,7 @@ class SqsParentContext {
     String traceHeader = systemAttributes.get(AWS_TRACE_SYSTEM_ATTRIBUTE);
     return AwsXrayPropagator.getInstance()
         .extract(
-            Context.current(),
+            Context.root(),
             Collections.singletonMap("X-Amzn-Trace-Id", traceHeader),
             MapGetter.INSTANCE);
   }
--- a/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/S3TracingTest.groovy
+++ b/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/S3TracingTest.groovy
@@ -1,0 +1,449 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import static io.opentelemetry.api.trace.SpanKind.CLIENT
+import static io.opentelemetry.api.trace.SpanKind.CONSUMER
+
+import com.amazonaws.services.s3.AmazonS3Client
+import com.amazonaws.services.s3.model.BucketNotificationConfiguration
+import com.amazonaws.services.s3.model.ObjectListing
+import com.amazonaws.services.s3.model.QueueConfiguration
+import com.amazonaws.services.s3.model.S3Event
+import com.amazonaws.services.s3.model.S3ObjectSummary
+import com.amazonaws.services.s3.model.SetBucketNotificationConfigurationRequest
+import com.amazonaws.services.sqs.AmazonSQSAsyncClient
+import com.amazonaws.services.sqs.model.GetQueueAttributesRequest
+import com.amazonaws.services.sqs.model.PurgeQueueRequest
+import com.amazonaws.services.sqs.model.ReceiveMessageRequest
+import io.opentelemetry.instrumentation.test.AgentInstrumentationSpecification
+import org.slf4j.LoggerFactory
+import org.testcontainers.containers.localstack.LocalStackContainer
+import org.testcontainers.containers.output.Slf4jLogConsumer
+import org.testcontainers.utility.DockerImageName
+import spock.lang.Ignore
+import spock.lang.Shared
+
+class S3TracingTest extends AgentInstrumentationSpecification {
+
+  @Shared
+  LocalStackContainer localstack
+  @Shared
+  AmazonSQSAsyncClient sqsClient
+  @Shared
+  AmazonS3Client s3Client
+
+  def setupSpec() {
+
+    localstack = new LocalStackContainer(DockerImageName.parse("localstack/localstack:latest"))
+      .withServices(LocalStackContainer.Service.SQS, LocalStackContainer.Service.SNS)
+      .withEnv("DEBUG", "1")
+      .withEnv("SQS_PROVIDER", "elasticmq")
+    localstack.start()
+
+    sqsClient = AmazonSQSAsyncClient.asyncBuilder()
+      .withEndpointConfiguration(localstack.getEndpointConfiguration(LocalStackContainer.Service.SQS))
+      .withCredentials(localstack.getDefaultCredentialsProvider())
+      //.withRegion(Regions.US_EAST_1)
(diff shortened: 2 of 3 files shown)


instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/S3TracingTest.groovy
```diff
@@ -0,0 +1,449 @@
...
+    setQueuePolicy(queueUrl, queueArn)
+    createBucket(bucketName)
+    enableS3Notifications(bucketName, queueArn)
+
+    when:
+    Thread.sleep(2000)
+    // test message, auto created by AWS
+    sqsClient.receiveMessage(new ReceiveMessageRequest(queueUrl))
+    s3Client.putObject(bucketName, "testKey", "testData")
+    Thread.sleep(3000)
```
iNikem, line 119: Is it possible to replace `Thread.sleep` with intermediate `assertTraces` ?

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/SqsParentContext.java
```diff
@@ -34,7 +34,7 @@ static Context ofSystemAttributes(Map<String, String> systemAttributes) {
     String traceHeader = systemAttributes.get(AWS_TRACE_SYSTEM_ATTRIBUTE);
     return AwsXrayPropagator.getInstance()
         .extract(
-            Context.current(),
+            Context.root(),
```
iNikem, line 37: I am trying to understand this change. Is there a test that failed before but now passes?

kubawach: It's the second point of the description :)

Basically with context.current() the model for consuming a message without parent tracing attribute would be:
- client (http receive) span parent of consumer (messaging) span
while with parent tracing attribute:
- producer (sending) span parent of all consumer (messaging) span
- client (http receive) span decoupled (modelling http receive operation)

With this change modelling is the same no matter if the attribute is present, that is client (http receive) span is always decoupled from consume (messaging) span.

@iNikem does this ^^ make sense for you now?

iNikem: Partially :D I still don't understand it fully, but now at least I have a sense, that you did the right thing :)

anuraaga: @kubawach Is it possible to create the consumer span after the client span is closed?

Basically we can imagine this 

------- Server ---------------------
---Recieve--- ---Consume---

I think we need the consume span to link to server right?

Though I recall a very similar conversation with @trask and @inikem a few months ago and can't remember what we decided.

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/S3TracingTest.groovy
mateuszrzeszutek: Or at least `waitForTraces`

kubawach: I don't think that's possible. In both cases sleep() is added to give the infrastructure (AWS now, localstack in the future) some time to propagate requested sync.
First sleep - enabling S3 notifications to SQS queue and sending "test" message (consumed in the next step). If we don't do that, receive message will still execute but won't consume the message (and will break the asserts).
Second sleep - gives S3 time to create notification message and deliver it to SQS. 

I believe that once appropriate localstack feature is implemented, the test will run *without* any sleeps (execution of notification will be sync with s3 upload).

@mateuszrzeszutek & @iNikem ^^

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/SqsParentContext.java
kubawach: @anuraaga yep, we could definitely change the order in TracingRequestHandler#afterResponse so that consume spans are simply created after client is closed. For this particular scenario would yield the same results (no bogus parent-child between client and consume in case of lacking AWS tracing parent attribute).

With regards to linking - currently consume is linked to server (produce) - provided the info is available as a system attribute of a message. Linking "client" with "server" (produce) however would be much harder to do, as "client" is created upon http request when the parent tracing info is not yet available. AFAIK our API does not allow to set the parent later on.

To sum up - if you prefer to stay with Context.current() for  SqsParentContext#ofSystemAttributes I can switch the code to create consumer spans after client is closed :)

trask: I may not be fully following, but if it helps, we use `Context.root()` in other messaging instrumentation (e.g. JMS, Kafka, RabbitMQ) for message listeners which have messages pushed to them, treating them as entry points (similar to http server entry points where we also use `Context.root()`)

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/S3TracingTest.groovy
iNikem: > First sleep - enabling S3 notifications to SQS queue and sending "test" message

Does this result in a trace? If yes, then we can wait for this trace to arrive instead of sleep, no?

kubawach: The problem with SQS consumer is that it will create a similar trace for "ReceiveMessage" with and without consumed message. So if we send "Receive" and wait for trace we will get one, but "test" message may not be consumed and break the trace later on (we check for parent-child link and for "test" message we won't get one). 

To avoid explicit "Thread.sleep()" we could utilise "getWaitTimeSeconds" property of the ReceiveMessageRequest - albeit it will still mean waiting ;) @iNikem LMK if this would be better.

iNikem: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html#sqs-long-polling

This reads to me that you can tell `receiveMessage` to wait until message is available. Thus we will wait only as long as needed and not defensively for 2-3 seconds in any case.

iNikem (approved):

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/S3TracingTest.groovy
kubawach: Yeah, that's what I meant (same setting). Will test with AWS and commit change if it works as expected :)

PR #2519, opened 2021-03-07
Use new version of publish / staging plugin which has merged the two …

anuraaga:
…projects.

It's the year of projects merging together!
https://github.com/gradle-nexus/publish-plugin


gradle/publish.gradle
```diff
@@ -102,7 +89,6 @@ private String artifactPrefix(Project p, String archivesBaseName) {
 }
 
 rootProject.tasks.release.finalizedBy tasks.publishToSonatype
-rootProject.tasks.closeAndReleaseRepository.mustRunAfter tasks.publishToSonatype
```
anuraaga, line 105: This is handled by the plugin now

anuraaga:
/cc @asarkar I randomly noticed you're looking into maven publishing plugins. Here's yet another one :P Our build doesn't function with any other plugin since we need the feature this one has of initializing the staging repository before uploading, instead of using the implicitly created repository and working on it. It's because with the count of our artifacts, Maven always flakes and creates a second staging repository during a single build if we don't initialize it explicitly first, causing other plugins to fail.

asarkar:
@anuraaga 
> I randomly noticed you're looking into maven publishing plugins.

I have my fingers in many pies, can you refresh my memory?

iNikem (approved):

anuraaga:
@asarkar https://github.com/DanySK/maven-central-gradle-plugin/issues/88

trask (approved):

mateuszrzeszutek (approved):

PR #2524, opened 2021-03-08
Load couchbase otel instrumentation for 3.1

### LINKED ISSUE #2520: Instrumentation for Couchbase Client 3.x
Opentelemetry agent works with Couchbase client(2.x) well. Do you have a plan to support 3.x client?

Thanks

anuraaga:
`tracing-opentelemetry-0.3.3` only supports couchbase 3.1, not 3.0. It seems quite tricky to deal with external library instrumentation if they have version alignment requirements, which is fairly reasonable. Let's see how it goes I guess.

Fixes #2520


instrumentation/couchbase/couchbase-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v3_0/CouchbaseEnvironmentInstrumentation.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Collections;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class CouchbaseEnvironmentInstrumentation implements TypeInstrumentation {
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
```
iNikem, line 25: Is it really optimizing anything?

iNikem (approved):

instrumentation/couchbase/couchbase-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v3_0/CouchbaseEnvironmentInstrumentation.java
anuraaga: Oops - at first was going to do a subclass check but realized I shouldn't need it

instrumentation/couchbase/couchbase-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v3_1/CouchbaseInstrumentationModule.java
```diff
@@ -0,0 +1,43 @@
...
+    super("couchbase", "couchbase-3.1");
+  }
+
+  @Override
+  protected String[] additionalHelperClassNames() {
+    return new String[] {
+      "com.couchbase.client.tracing.opentelemetry.OpenTelemetryRequestSpan",
+      "com.couchbase.client.tracing.opentelemetry.OpenTelemetryRequestTracer"
+    };
+  }
```
mateuszrzeszutek, line 29: This is interesting: if we're going to include external library instrumentations in the javaagent (and we're probably going to do that, since it seems to be a good way to make the agent a bit slimmer & simpler) then we need to improve muzzle a bit.
Right now muzzle not only does not collect helper classes from 3rd party packages, it also collects no references whatsoever. Which means that we actually don't have the muzzle check both in build time and runtime.
It should be possible to define the set of packages for each module, by default it'd be same as `InstrumentationClassPredicate` - I'm going to open an issue for that.

mateuszrzeszutek (approved):

instrumentation/couchbase/couchbase-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v3_1/CouchbaseInstrumentationModule.java
mateuszrzeszutek: Turns out there already was one: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1395
I've just added my 3 cents there

anuraaga:
As we added to #1395 will go ahead and merge this

PR #2525, opened 2021-03-08
Extract library instrumentation for AWS SDK v1

### LINKED ISSUE #2502: Request for OTel support in AWS Java SDK v1.x

anuraaga:
The tricky part of this was that our instrumentation also applied bytecode fixes to work around buggy behavior in the AWS SDK v1 itself related to exceptions, in particular making sure that we are able to close an active scope even with exceptions. This is not possible in library instrumentation, but we also don't really need to activate the scope in it because apache http client won't be automatically instrumented / need suppression. So I've basically split the request handler into two, agent one that delegates to library instrumentation but also creates a scope for use in the agent.

Fixes #2502


instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v1_11/AwsHttpClientInstrumentation.java
```diff
@@ -38,22 +40,26 @@
...
+        @Advice.Return Response<?> response,
         @Advice.Thrown Throwable throwable) {
-      if (throwable != null) {
-        ContextScopePair scope = request.getHandlerContext(CONTEXT_SCOPE_PAIR_CONTEXT_KEY);
-        if (scope != null) {
-          request.addHandlerContext(CONTEXT_SCOPE_PAIR_CONTEXT_KEY, null);
-          tracer().endExceptionally(scope.getContext(), throwable);
-          scope.closeScope();
-        }
+      if (throwable instanceof Exception) {
```
anuraaga, line 57: There is a small regression in my desire to delegate the actual tracing to the library instrumentation - `Error` will result in unclosed traces. I feel this should be rare enough to live with until we get a report of problems.

mateuszrzeszutek (approved):

PR #2528, opened 2021-03-08
Add support for elasticsearch 7 rest client

laurit:


iNikem:
Looking at muzzle configurations for `elasticsearch-rest-6.4` and `elasticsearch-rest-7.0` it seems to me that the former instrumentation will be applied to version 7.0 as well. I don't think this is what we want. Can you try to add `elasticsearch-rest-6.4` as testInstrumentation to `elasticsearch-rest-7.0` and see what will happen?

anuraaga (approved):

iNikem (approved):

PR #2530, opened 2021-03-08
Add asynchronous tracing for Java 8 CompletableFuture in WithSpanAdvice

HaloFour:
Adds basic support for asynchronous tracing with the OTel annotation WithSpanAdvice.  Only supports Java 8 `CompletionStage<T>` and `CompletableFuture<T>` but the `MethodSpanStrategy` interface should be able to support most/any promise-like return types.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/async/MethodSpanStrategyContextKey.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.tracer.async;
+
+import io.opentelemetry.context.ContextKey;
+
+final class MethodSpanStrategyContextKey {
```
iNikem, line 10: Why a separate class for this? Key can live in `MethodSpanStrategy`, right?

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
```diff
@@ -187,4 +202,132 @@ void suppressServerSpan() throws Throwable {
...
+    @Test
+    @DisplayName("should end Span on complete")
+    void onComplete() throws Throwable {
+      CompletableFuture<String> future = new CompletableFuture<>();
+
+      // when
+      withSpan("parent", () -> withSpanTester.testAsyncCompletionStage(future));
+
+      try {
+        instrumentation.waitForTraces(1, 20, TimeUnit.SECONDS);
```
iNikem, line 219: So this test will always take at least 20 seconds? Thus this test class will take more than a minute to complete. We have to come up with some other way :(

iNikem (commented):
This is definitely an interesting proposal :) But it needs some documentation. E.g. `package-info.java` in `io.opentelemetry.instrumentation.api.tracer.async`.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/async/MethodSpanStrategyContextKey.java
HaloFour: It could, I was following the convention set by `Span`+`SpanContextKey` where the `ContextKey<T>` isn't publicly accessible, but if that doesn't matter I can make it a static final field on the `MethodSpanStrategy` interface instead.

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTest.java
HaloFour: Agreed, I need to find a better way to assert the negative here and would love some pointers.  Some of the Groovy tests seem to sleep for 500ms and I'll be happy to follow the pattern.

Actually this test is currently "failing" as the parent span is recorded, which I wasn't expecting, so I'll either need to stop running it under the parent span or find a better way to assert that only the parent span has completed.

HaloFour:
I will park this for feedback again.

I've moved some more of the machinery to `instrumentation-api` to a new abstract class derived from `BaseTracer` called `BaseMethodTracer`.  This will likely be reused by `WithSpanAdvice` and `MethodAdvice`, but I didn't think it necessarily belonged in `BaseTracer` as it wouldn't apply to most instrumentation, only those designed specifically to intercept method invocation.

I changed the `MethodSpanStrategies` to be a singleton instance instead of static.  This is kind of a placeholder class and I'm not sure what the best route would be to handle all of the reactive/future-like types that can be supported by a given framework.  Right now it would require manually registering additional `MethodSpanStrategy` implementations in code for each reactive type in shared mutable state.  I imagine a goal would be that when using the Java agent that it would know which types exist and to automatically register supporting strategies.

Names and other conventions are certainly up for discussion.

Thank you!

anuraaga:
Thanks @HaloFour! To be honest, I was only expecting a change in https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations because I forgot about the spring aspect :) It does seem like we'd want to make something generic to apply to both, but does it make sense to first create something non-generic in just `otelannotations` (our primary supported use case right now) and then make things generic from there after we understand the implementation fully?

HaloFour:
@anuraaga 

Done and done, shifted all of the potentially sharable bits over to the otelannotations instrumentation module.  I targeted the Spring version as the projects I've been working on are Spring WebFlux.

I also see a `MethodAdvice` instrumentation which uses ByteBuddy but without using the `WithSpan` annotation?  Surprised there isn't an AspectJ flavor also.  I can certainly understand any hesitation about having this much specialized code in instrumentation-api.  Perhaps it could live in a separate instrumentation-aop module or the like if this is accepted and we want to share the machinery across the different instrumentation modules?

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/CompletableFutureMethodSpanStrategy.java
```diff
@@ -0,0 +1,51 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.otelannotations.async;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import java.util.concurrent.CompletableFuture;
+
+enum CompletableFutureMethodSpanStrategy implements MethodSpanStrategy {
```
mateuszrzeszutek, line 12: Isn't `CompletableFuture` just an implementation of `CompletionStage`? Wouldn't `CompletionStageMethodSpanStrategy` work for both of them?

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanTracer.java
```diff
@@ -23,19 +25,33 @@ public static WithSpanTracer tracer() {
 
   private static final Logger log = LoggerFactory.getLogger(WithSpanTracer.class);
 
+  private final MethodSpanStrategies methodSpanStrategies = MethodSpanStrategies.getInstance();
+
   public Context startSpan(
       Context parentContext, WithSpan applicationAnnotation, Method method, SpanKind kind) {
+
+    Context spanStrategyContext = withMethodSpanStrategy(parentContext, method);
```
mateuszrzeszutek, line 33: Hmm, do you really need to store the strategy in the context? I mean, you could just call it directly in the `WithSpanAdvice#stopSpan()` method using the returned value.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/MethodSpanStrategies.java
```diff
@@ -0,0 +1,44 @@
...
+  private MethodSpanStrategies(Map<Class<?>, MethodSpanStrategy> strategies) {
+    this.strategies = strategies;
+  }
+
+  public void registerStrategy(Class<?> returnType, MethodSpanStrategy strategy) {
+    strategies.put(returnType, strategy);
+  }
+
+  public MethodSpanStrategy resolveStrategy(Class<?> returnType) {
+    return strategies.getOrDefault(returnType, MethodSpanStrategy.synchronous());
```
mateuszrzeszutek, line 42: Hmm, this won't work if you pass some `CustomFuture` that implements `CompletionStage`: you'd have to iterate over the map and check if `entry.getKey().isAssignableFrom(returnType)`

HaloFour: Although even with `CompletionStage`/`CompletableFuture` if you don't return the composed future that will cause any composition done by the consumer of the annotated method to count towards the duration of the span since the registered notifications are fired in reverse order and on the same thread by default.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanTracer.java
HaloFour: Agreed, I've removed the use of context here.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanTracer.java
```diff
@@ -69,6 +72,21 @@ public static SpanKind toAgentOrNull(
     }
   }
 
+  /**
+   * Denotes the end of the invocation of the traced method with a successful result which will end
+   * the span stored in the passed {@code context}. If the method returned a value representing an
+   * asynchronous operation then the span will remain open until the asynchronous operation has
```
mateuszrzeszutek, line 78: Nit:
```suggestion
   * asynchronous operation then the span will not be finished until the asynchronous operation has
```

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+    return returnType == CompletionStage.class || returnType == CompletableFuture.class;
+  }
+
+  @Override
+  public Object end(BaseTracer tracer, Context context, Class<?> returnType, Object result) {
+    if (result instanceof CompletableFuture) {
+      CompletableFuture<?> future = (CompletableFuture<?>) result;
+      if (future.isDone()) {
+        return endSynchronously(future, tracer, context);
+      }
```
mateuszrzeszutek, line 27: What's the reason for this block? I've tried removing it, added the following tests:

```groovy
  def "should capture span for completed CompletableFuture"() {
    setup:
    def future = CompletableFuture.completedFuture("Done")
    new TracedWithSpan().completableFuture(future)

    expect:
    assertTraces(1) {
      trace(0, 1) {
        span(0) {
          name "TracedWithSpan.completableFuture"
          kind SpanKind.INTERNAL
          hasNoParent()
          errored false
          attributes {
          }
        }
      }
    }
  }

  def "should capture span for completed exceptionally CompletableFuture"() {
    setup:
    def future = new CompletableFuture()
    future.completeExceptionally(new IllegalArgumentException("Boom"))
    new TracedWithSpan().completableFuture(future)

    expect:
    assertTraces(1) {
      trace(0, 1) {
        span(0) {
          name "TracedWithSpan.completableFuture"
          kind SpanKind.INTERNAL
          hasNoParent()
          errored true
          errorEvent(IllegalArgumentException, "Boom")
          attributes {
          }
        }
      }
    }
  }
```

and everything still passed.

HaloFour: It's an optimization, one that can't really be detected in the tests.  It avoids the allocation of the callback lambda if the future is already completed.  It's not entirely necessary and I can remove it.

mateuszrzeszutek: Can you add a comment explaining why it was added? I think it's fine to leave it here, but it should be obvious why we're doing this.

Also, WDYT about adding tests that cover this path?

HaloFour: Can do.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Checks to see if the {@link CompletableFuture} has already been completed and if so
+   * synchronously ends the span to avoid additional allocations and overhead registering for
+   * notification of completion.
+   */
+  private boolean endSynchronously(
+      CompletableFuture<?> future, BaseTracer tracer, Context context) {
+
+    if (future.isDone()) {
```
mateuszrzeszutek, line 45: Nit: you can return early instead:

```java
if (!future.isDone()) {
  return false;
}

**Review by mateuszrzeszutek — APPROVED** [review]

**Comment by mateuszrzeszutek** [comment]
LGTM 👍

File: instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java [hunk 9]
```diff
@@ -0,0 +1,78 @@
...
+  INSTANCE;
+
+  @Override
+  public boolean supports(Class<?> returnType) {
+    return returnType == CompletionStage.class || returnType == CompletableFuture.class;
+  }
+
+  @Override
+  public Object end(BaseTracer tracer, Context context, Class<?> returnType, Object result) {
+    if (result instanceof CompletableFuture) {
```
trask, line 23: to avoid problem if method is declared to return a subclass of `CompletableFuture`... although that's not possible due to `supports` implementation above... still, feels a bit safer in case this pattern is extended/copied in the future
```suggestion
    if (returnType == CompletableFuture.class) {
```

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+      CompletableFuture<?> future = (CompletableFuture<?>) result;
+      if (endSynchronously(future, tracer, context)) {
+        return future;
+      }
+      return endWhenComplete(future, tracer, context);
+    } else if (result instanceof CompletionStage) {
```
trask, line 29:
```suggestion
    } else if (returnType == CompletionStage.class) {
```

trask (approved):
😎

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
HaloFour: I've added a double check by having `end` call `supports` on the `returnType`.  I left the use of `instanceof` as once the method signature is checked it is safe to deal with subclasses, e.g. if the method is declared to return `CompletionStage<T>` but actually returns `CompletableFuture<T>`.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+    return true;
+  }
+
+  /**
+   * Registers for notification of the completion of the {@link CompletionStage} at which time the
+   * span will be ended.
+   */
+  private CompletionStage<?> endWhenComplete(
+      CompletionStage<?> stage, BaseTracer tracer, Context context) {
+    return stage.whenComplete(
```
anuraaga, line 71: If the stage is already complete I think this is guaranteed to be synchronous - I guess we can remove endSynchronously?

anuraaga (commented):
Sorry for delay, just a small comment but this mostly looks ready

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+  INSTANCE;
+
+  @Override
+  public boolean supports(Class<?> returnType) {
+    return returnType == CompletionStage.class || returnType == CompletableFuture.class;
+  }
+
+  @Override
+  public Object end(BaseTracer tracer, Context context, Class<?> returnType, Object result) {
+    if (supports(returnType)) {
```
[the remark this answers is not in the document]
HaloFour: > does this always return `true`?

It should, yes, unless someone grabs an instance of this strategy and uses it pathologically.  I agree that a check is probably not really necessary here.

> or maybe these strategy classes should be limited to a single returnType, which would make this situation easier to reason about / more self-explanatory?

That's how I originally organized the code but I found that there's a lot of overlap between the two, especially given that you'd expect that the most common implementation of the `CompletionStage<T>` interface would be the `CompletableFuture<T>` class, and you're guaranteed that `CompletableFuture#whenComplete` will return a `CompletableFuture<T>` due to covariant overloads.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/async/Jdk8MethodStrategy.java
HaloFour: I double checked to be sure and yes, `whenComplete` should always be synchronous, at least given how it's implemented in `CompletableFuture<T>`.  Checking and completing synchronously was more about optimizing away the need for the callback and the extra allocations that requires.  It's not observable, but it is cheaper/faster.  But if that's not worth the complexity I can remove it.

trask:
I sent a PR to your PR with a couple of suggestions https://github.com/HaloFour/opentelemetry-java-instrumentation/pull/1

anuraaga (approved):

trask:
thanks for the new feature @HaloFour 🎉

HaloFour:
Excellent news!  Glad I could contribute.

From here what might the plan be to extend this to other aspect-based instrumentation, especially `WithSpanAspect`/`WithSpanAspectTracer`?  The actual implementation within the instrumentation classes seems simple enough but the strategy approach could be shared, if that were to live in `instrumentation-api` or a new module.

Then there is expanding the list of strategies to other asynchronous types, like Reactor, RxJava, Guava, etc.

I'd love to be involved with both but I can understand wanting to let this "soak" first.

Issue #2529, opened 2021-03-08
Lettuce 6 instrumentation Request


mnadeem:
**Is your feature request related to a problem? Please describe.**
Currently Redis is instrumented with Lettuce 5 and not with 6

**Describe the solution you'd like**
Lettuce 6 instrumentation as it is happening for Lettuce 5

**Describe alternatives you've considered**
N/A

**Additional context**
https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2436 refer for more details

PR #2533, opened 2021-03-09
Extract library instrumentation for lettuce 5.1

anuraaga:
Preparation for #2529


anuraaga:
Github's not rendering this diff well :( The test files are mostly just copy-paste, except I had to keep one async test in the javaagent directory since it seems to rely on executor instrumentation for now. Otherwise no real logic change, just making `Tracer` a parameter to `OpenTelemetryTracing` per usual pattern.

mateuszrzeszutek (approved):

PR #2536, opened 2021-03-09
Minor InstrumentationClassPredicate refactor (don't treat javaagent-t…

mateuszrzeszutek:
…ooling as helper classes)


trask (approved):
👍

anuraaga (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v2_0/JaxRsAnnotationsTracer.java
```diff
@@ -12,7 +12,7 @@
 import io.opentelemetry.instrumentation.api.servlet.ServletContextPath;
 import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
 import io.opentelemetry.instrumentation.api.tracer.ServerSpan;
-import io.opentelemetry.javaagent.tooling.ClassHierarchyIterable;
+import io.opentelemetry.javaagent.instrumentation.api.ClassHierarchyIterable;
```
mateuszrzeszutek, line 15: Fun thing: it turns out we were using a class from `javaagent-tooling` in jaxrs instrumentations 😱

Issue #2538, opened 2021-03-10
Mismatch in column names for benchmark integration test perf_results.csv file


shantanu-vsbhosale:
**Steps to reproduce**
1. Run the benchmark integration test and see the perf_results.csv file.
2.  There is a mismatch in column names and 2 columns are empty. (Server Start RSS,Server Load Increase RSS) 

**Issue**
In the shell file opentelemetry-java-instrumentation/benchmark-integration/run-perf-test.sh line no 129
`header="$header,Agent CPU Burn,Server CPU Burn,Agent RSS Delta,Server Max RSS,Server Start RSS,Server Load Increase RSS"`
There are **6** columns names related to server statistics but when we are writing the data to csv at line number 175
`echo "$result_row,$server_cpu,$server_max_rss,$server_start_rss,$server_load_increase_rss" >> $test_csv_file`
There are only **4** columns related to server statistics

We are not calculating **Agent CPU Burn** & **Agent RSS Delta** but are filled with data whereas
last 2 columns **Server Start RSS & Server Load Increase RSS** are empty.

**Resolution**
Remove the Agent CPU Burn & Agent RSS Delta column names from line no 129. The data will match appropriately to the column name.

<img width="1084" alt="perf_results_empty" src="https://user-images.githubusercontent.com/80242052/110578233-59620c80-818a-11eb-98f3-384f688897a8.png">
[perf_results.csv.zip](https://github.com/open-telemetry/opentelemetry-java-instrumentation/files/6113263/perf_results.csv.zip)

shantanu-vsbhosale:
@trask, @tylerbenson Can you look at this issue ? I have raised PR for the same at #2542

PR #2542, opened 2021-03-10
Column name mismatch in perf_results.csv file for benchmark-integration test

shantanu-vsbhosale:
This is a PR for issue #2538


trask (approved):
Thanks @shantanu-vsbhosale!

If you are benchmarking the javaagent, have a look at the `benchmark-e2e` directory for the new benchmark harness we are using/building out, probably we will remove the `benchmark-integration` directory at some point.

cc: @breedx-splk

PR #2548, opened 2021-03-10
Move and rename NetPeerUtils

mateuszrzeszutek:
* Rename NetPeerUtils to NetPeerAttributes; inject instance instead of using a global var
* Rename SpanAttributeSetter to AttributeSetter (might be used to set on AttributesBuilder in the future, who knows)
* Deprecate default BaseTracer constructor: library instrumentations are supposed to inject all dependencies, the javaagent should explicitly pass globals


instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaTracingBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Used to specify a mapping from hostnames or IP addresses (key of the passed {@code
+   * peerServiceMapping} map) to peer services (value of the map). The peer service is added as an
+   * attribute to a span whose host or IP match the mapping.
+   *
+   * @see <a
+   *     href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-remote-service-attributes">peer.service
+   *     attribute specification</a>
+   */
+  public ArmeriaTracingBuilder setPeerServiceMapping(Map<String, String> peerServiceMapping) {
```
trask, line 30: plural seems slightly better to me, but maybe too late to change b/c of env var `OTEL_INSTRUMENTATION_COMMON_PEER_SERVICE_MAPPING`
```suggestion
  public ArmeriaTracingBuilder setPeerServiceMappings(Map<String, String> peerServiceMappings) {
```

trask (approved):

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaTracingBuilder.java
anuraaga, line 30: Library doesn't need a peer service mapping, that's an agent-specific concept since you can't make code changes. When initializing library instrumentation for a client code could just explicitly pass in the peer service. So this could be just peer service, or just `Attributes` to build.

I've been thinking about this a bit while prototyping a new API based on instrumenters and extractors here

https://github.com/open-telemetry/opentelemetry-java-instrumentation/compare/main...anuraaga:instrumentation-extractor

The intention TBH is to remove the Tracers if that works out - should we take a break from refactoring Tracer?

mateuszrzeszutek: > Library doesn't need a peer service mapping, that's an agent-specific concept

> Library doesn't need a peer service mapping, that's an agent-specific concept 

👍 

> 
> I've been thinking about this a bit while prototyping a new API based on instrumenters and extractors here
> 
> [main...anuraaga:instrumentation-extractor](https://github.com/open-telemetry/opentelemetry-java-instrumentation/compare/main...anuraaga:instrumentation-extractor)
> 
> The intention TBH is to remove the Tracers if that works out - should we take a break from refactoring Tracer?

Nice, I actually had a somewhat similar idea about `HttpExtractor`, the (almost) only difference being I wanted to name it `HttpAttributes` 😄  
I'll take a closer look at your changes today, but there's one question that I wanted to ask you right away: what about instrumentations that call `tracer().onRequest()` or `tracer().onException()` in different places than start/end? I had a notion that in tracers (or instrumenters should we move on from tracers) we should expose more "lifecycle" methods than just start/end. WDYT?

anuraaga: I need to try to adapt these sort of libraries to see what happens indeed - two lifecycle methods keeps things very simple for many cases, so depending on how common the separate lifecycles is, keeping the data between the exception point to end point, for example, could be managed by the library instrumentation too at the expense of some object storage. I'm hoping for the minimum of lifecycle methods though.

trask: @mateuszrzeszutek does it make sense to remove the `setPeerServiceMapping` methods and merge this? I ❤️ refactoring forward, all the cleanup and consistency helps us visualize and implement the next layer of refactorings.

mateuszrzeszutek: Done 👍

anuraaga (approved):

PR #2549, opened 2021-03-10
AWS SDK - S3 - SNS - SQS propagation tests

kubawach:
- adds S3 -> SNS -> SQS propagation test (disabled currently as requiring localstack features implementation)
- some refactoring to make those test more readable

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/AwsConnector.groovy
+++ b/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/AwsConnector.groovy
@@ -1,0 +1,188 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import com.amazonaws.regions.Regions
+import com.amazonaws.services.s3.AmazonS3Client
+import com.amazonaws.services.s3.model.BucketNotificationConfiguration
+import com.amazonaws.services.s3.model.ObjectListing
+import com.amazonaws.services.s3.model.QueueConfiguration
+import com.amazonaws.services.s3.model.S3Event
+import com.amazonaws.services.s3.model.S3ObjectSummary
+import com.amazonaws.services.s3.model.SetBucketNotificationConfigurationRequest
+import com.amazonaws.services.s3.model.TopicConfiguration
+import com.amazonaws.services.sns.AmazonSNSAsyncClient
+import com.amazonaws.services.sns.model.CreateTopicResult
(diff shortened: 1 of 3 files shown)


instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/AwsConnector.groovy
```diff
@@ -0,0 +1,188 @@
...
+    awsConnector.localstack.start()
+    awsConnector.localstack.followOutput(new Slf4jLogConsumer(LoggerFactory.getLogger("test")))
+
+    awsConnector.sqsClient = AmazonSQSAsyncClient.asyncBuilder()
+      .withEndpointConfiguration(awsConnector.localstack.getEndpointConfiguration(LocalStackContainer.Service.SQS))
+      .withCredentials(awsConnector.localstack.getDefaultCredentialsProvider())
+      .build()
+
+    awsConnector.s3Client = AmazonS3Client.builder()
+      .withEndpointConfiguration(awsConnector.localstack.getEndpointConfiguration(LocalStackContainer.Service.SNS))
```
jkwatson, line 51: is this supposed to be `LocalStackContainer.Service.S3` ?

iNikem (approved):

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/groovy/AwsConnector.groovy
kubawach: Yes, definitely! Thanks for spotting this. Funny thing the test works with live AWS as localstack config is not used. But would fail in the future once we enable it.

anuraaga: Just wondering, how do you test against live AWS? Is it worth having a system property to allow this same code to run both localstack and live?

kubawach: Using my own AWS account (and env vars with credentials). 

I was thinking about proposing that we could set up and OTEL AWS account and move tests interacting with external system to a separate suite. One that would run during build, but would not fail the process (increased test flakiness due to external system communication). What would you say about that?

anuraaga: If our CNCF provided account has budget for it it seems like a nice idea. If not, we can do similar in https://github.com/aws-observability/aws-otel-java-instrumentation for free - though I don't think I'd let spock into that repo ;)

anuraaga (approved):

PR #2551, opened 2021-03-10
Update the agent-config docs.

jkwatson:
This provides more direct linking to the auto-instrumentation docs that people might need.

Mainly, it reduces by one the number of links that a user might have to click to get to what they need.

### DIFF
--- a/docs/agent-config.md
+++ b/docs/agent-config.md
@@ -1,4 +1,4 @@
-# Configuration options
+# Agent Configuration
 
 ## NOTE: subject to change!
 
--- a/docs/agent-config.md
+++ b/docs/agent-config.md
@@ -2,8 +2,8 @@
 
 ## NOTE: subject to change!
 
-Note: The environment variables/system properties in this document are very likely to change over time. 
-Please check back here when trying out a new version! 
+Note: The environment variables/system properties in this document are very likely to change over time.
+Please check back here when trying out a new version!
 
 Please report any bugs or unexpected behavior you find.
 
--- a/docs/agent-config.md
+++ b/docs/agent-config.md
@@ -17,23 +17,23 @@ Please report any bugs or unexpected behavior you find.
 ## SDK Autoconfiguration
 
 The SDK's autoconfiguration module is used for basic configuration of the agent. Read the
-[docs](https://github.com/open-telemetry/opentelemetry-java/tree/v1.0.0/sdk-extensions/autoconfigure)
+[docs](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure)
 to find settings such as configuring export or sampling.
 
 Here are some quick links into those docs for the configuration options for specific portions of the SDK & agent:
 
 * [Exporters](#exporters)
-  + [OTLP exporter (both span and metric exporters)](https://github.com/open-telemetry/opentelemetry-java/tree/v1.0.0/sdk-extensions/autoconfigure/README.md#otlp-exporter-both-span-and-metric-exporters)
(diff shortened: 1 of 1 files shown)


docs/agent-config.md
```diff
@@ -1,25 +1,40 @@
-# Configuration parameters
+# Configuration options
```
breedx-splk, line 1: Or maybe just "Agent configuration"

docs/agent-config.md
```diff
@@ -1,25 +1,40 @@
...
 
 The SDK's autoconfiguration module is used for basic configuration of the agent. Read the
-[docs](https://github.com/open-telemetry/opentelemetry-java/tree/v1.0.0/sdk-extensions/autoconfigure#customizing-the-opentelemetry-sdk)
+[docs](https://github.com/open-telemetry/opentelemetry-java/tree/v1.0.0/sdk-extensions/autoconfigure)
 to find settings such as configuring export or sampling.
 
+Here are some quick links into those docs for the configuration options for specific portions of the SDK & agent:
+
+* [Exporters](#exporters)
+  + [OTLP exporter (both span and metric exporters)](https://github.com/open-telemetry/opentelemetry-java/tree/v1.0.0/sdk-extensions/autoconfigure/README.md#otlp-exporter-both-span-and-metric-exporters)
```
breedx-splk, line 26: Hmmm...I'm not convinced that linking to the specific version (instead of `main` branch) is a good idea.  Can you convince me?

breedx-splk (approved):
A definite improvement.  🏆

docs/agent-config.md
jkwatson: I have no idea...it's what was linked to in this document, so I stuck with it.

docs/agent-config.md
```diff
@@ -44,14 +59,6 @@ The following property may be used to disable it:
...
-with caution*
-
-The OpenTelemetry SDK exposes SPI [hooks](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/spi)
-for customizing its behavior, such as the `Resource` attached to spans or the `Sampler`.
-
 Because the automatic instrumentation runs in a different classpath than the instrumented application, it is not possible for customization in the application to take advantage of this customization. In order to provide such customization, you can provide the path to a JAR file, including an SPI implementation using the system property `otel.initializer.jar`. Note that this JAR needs to shade the OpenTelemetry API in the same way as the agent does. The simplest way to do this is to use the same shading configuration as the agent from [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/cfade733b899a2f02cfec7033c6a1efd7c54fd8b/java-agent/java-agent.gradle#L39). In addition, you must specify the `io.opentelemetry.javaagent.shaded.io.opentelemetry.api.trace.spi.TraceProvider` to the name of the class that implements the SPI.
```
trask, line 62: I think this section is still relevant (though maybe too advanced for main agent config page, and scheduled to change a bit in #2442)

docs/agent-config.md
trask: `main` sounds reasonable to me

docs/agent-config.md
jkwatson: yeah, I missed this line. I think we should remove it from this page, since it's not relevant for normal end-users and will probably confuse most people who are the target audience.

mateuszrzeszutek (approved):

docs/agent-config.md
jkwatson: updated to point at `main`

trask (approved):
❤️ :shipit:

PR #2559, opened 2021-03-11
Use ClassValue instead of ClassLoader -> Class weakmap since it shoul…

anuraaga:
…d be same thing.

Randomly noticed this - two `Class` in different `ClassLoader` would have different `ClassValue` right?


mateuszrzeszutek (approved):

iNikem:
Will want @trask opinion here. It was his suggestion to use `ClassLoader` here.

anuraaga:
@richardstartin Is that a laugh of agreement or a laugh at my naivete? :P

richardstartin:
@anuraaga I just love to see the love for `ClassValue` spreading, it's arguably the most useful class in the JDK for people writing instrumentation and it goes all the way back to JDK7.

trask (approved):
> Will want @trask opinion here. It was his suggestion to use `ClassLoader` here.

this addresses the same issue, but much better implementation ❤️

Issue #2499, opened 2021-03-04
Undertow span is finished before last servlet filter


pavolloffay:
**Describe the bug**

Undertow instrumentation finishes server request span before the first servlet filter exits.

This has two implications:
* potentially wrong timing information
* additional instrumentation cannot add information to span in the servlet filter because the span has already been finished. 
  

**Steps to reproduce**
https://github.com/wildfly/quickstart/blob/11.0.0.Final/jaxrs-client/src/main/java/org/jboss/as/quickstarts/jaxrsclient/rest/ContactResourceRESTService.java
Deploy on http://download.jboss.org/wildfly/10.1.0.Final/wildfly-10.1.0.Final.zip

**What did you expect to see?**

Span should be recording events after `chain.doFilter(request, response);`

**What did you see instead?**

Span is not recording events after `chain.doFilter(request, response);`

**What version are you using?**
0.17.0

**Environment**
JDK 11
WF 10.1.0

pavolloffay:
cc) @laurit

pavolloffay:
The issue seems to be present only for post request 

```
curl -ivX POST -H "Content-Type: application/json" localhost:8080/jaxrs-client/rest/contacts -d '{"id": 15, "name": "jdoe"}'
```

PR #2560, opened 2021-03-11
Fix prematurely ending server span on undertow

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2499

Currently we end server span on undertow from `ExchangeCompletionListener` which is called when response is sent to the client. This may happen before request processing code has completed. Change span ending logic so that span is ended when all handlers have exited and exchange has been completed.
@pavolloffay could you verify that this fixes the issue you had


pavolloffay:
thanks for the fix @laurit. I will have a look at this shortly.

pavolloffay:
Maybe there could be a test that adds a filter and verifies that span is recording events after the `.doFilter`

instrumentation/undertow/javaagent/undertow-javaagent.gradle
```diff
@@ -4,7 +4,8 @@ muzzle {
   pass {
     group = "io.undertow"
     module = 'undertow-core'
-    versions = "[2.0.0.Final,)"
+    versions = "[1.4.0.Final,)"
```
iNikem, line 7: Change README as well, please

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/undertow/UndertowRequestContext.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package io.opentelemetry.instrumentation.api.undertow;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/** Helper container for keeping track of request processing state in undertow. */
+public class UndertowRequestContext {
```
mateuszrzeszutek, line 13: Can you move it to `javaagent-api`? This class does not need to be available in the library instrumentations.

Also, nit: please make it final and add a private constructor.

instrumentation/undertow/javaagent/undertow-javaagent.gradle
laurit: added undertow to supported-libraries.md

instrumentation/undertow/javaagent/src/test/groovy/UndertowServerTest.groovy
```diff
@@ -66,4 +93,65 @@ class UndertowServerTest extends HttpServerTest<Undertow> implements AgentTestTr
...
+          hasNoParent()
+          name "HTTP GET"
+
+          event(0) {
+            eventName "before-event"
+          }
+          event(1) {
+            eventName "after-event"
+          }
+          errorEvent(Exception, "exception after sending response", 2)
```
trask, line 151: can you add `attributes` validation also? e.g. would be good to see that we're capturing `200` even though also capturing the exception

trask (approved):
thx for tracking this down!

pavolloffay (approved):

PR #2566, opened 2021-03-12
Remove unnecessary libs from the javaagent shadowJar

mateuszrzeszutek:
And make it a bit lighter: 36 MB -> 21 MB

### DIFF
--- a/instrumentation/cdi-testing/cdi-testing.gradle
+++ b/instrumentation/cdi-testing/cdi-testing.gradle
@@ -1,4 +1,7 @@
-ext.skipPublish = true
+ext {
+  packageInAgentBundle = false
+  skipPublish = true
+}
 apply from: "$rootDir/gradle/instrumentation.gradle"
 
 dependencies {
--- a/instrumentation/cdi-testing/cdi-testing.gradle
+++ b/instrumentation/cdi-testing/cdi-testing.gradle
@@ -1,5 +1,4 @@
 ext {
-  packageInAgentBundle = false
   skipPublish = true
 }
 apply from: "$rootDir/gradle/instrumentation.gradle"
--- a/javaagent-tooling/javaagent-tooling.gradle
+++ b/javaagent-tooling/javaagent-tooling.gradle
@@ -28,13 +28,7 @@ dependencies {
   implementation deps.opentelemetrySdk
   implementation deps.opentelemetrySdkAutoconfigure
   implementation deps.opentelemetrySdkMetrics
-  implementation(deps.opentelemetryKotlin) {
-    // opentelemetry-extension-kotlin classes are injected into user classpath
-    // where kotlin core libraries are already present
-    exclude group: 'org.jetbrains.kotlin'
-    exclude group: 'org.jetbrains.kotlinx'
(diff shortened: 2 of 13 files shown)


instrumentation/cdi-testing/cdi-testing.gradle
```diff
@@ -1,4 +1,7 @@
-ext.skipPublish = true
+ext {
+  packageInAgentBundle = false
+  skipPublish = true
```
iNikem, line 3: This is great idea, but it bothers me, that one of these options is "positive" and another is "negative". Can we make them similar?

mateuszrzeszutek: We definitely should.

anuraaga: Sorry didn't look closely - do `testing` libraries need to use instrumentation.gradle?

mateuszrzeszutek: It turns the flag is not needed at all, we can just add only projects that are named `javaagent` (thanks to our strict module naming convention)

mateuszrzeszutek: > Sorry didn't look closely - do `testing` libraries need to use instrumentation.gradle?

Some of the `testing` artifacts are not libraries, but modules that contain only tests for a particular environment - for example `glassfish-testing`, or those three classloader testing modules.

trask (approved):
> And make it a bit lighter: 36 MB -> 21 MB

😱😱😱

Can I approve this multiple times?!??

javaagent-tooling/javaagent-tooling.gradle
```diff
@@ -31,8 +31,9 @@ dependencies {
   implementation(deps.opentelemetryKotlin) {
     // opentelemetry-extension-kotlin classes are injected into user classpath
     // where kotlin core libraries are already present
-    exclude group: 'org.jetbrains.kotlin', module: 'kotlin-stdlib-common'
-    exclude group: 'org.jetbrains.kotlinx', module: 'kotlinx-coroutines-core'
+    exclude group: 'org.jetbrains.kotlin'
+    exclude group: 'org.jetbrains.kotlinx'
+    exclude group: 'org.jetbrains', module: 'annotations'
```
trask, line 36: btw, I just remembered, I don't think these exclusions are needed anymore since https://github.com/open-telemetry/opentelemetry-java/pull/2793

anuraaga (approved):

iNikem (approved):

javaagent-tooling/javaagent-tooling.gradle
iNikem: @mateuszrzeszutek Are you going to look into this or just go ahead and merge?

mateuszrzeszutek: I'll take a look at it 👀

mateuszrzeszutek: Javaagent jar without these exclusions is exactly the same, they can be removed 👍

Issue #2201, opened 2021-02-05
Smoke tests on Windows


iNikem:
We have smoke tests for different application services, such as WildFly or Tomcat. Currently they run only on Ubuntu Linux. Is there an interest in having those tests run on Windows as well? If yes, and if https://github.com/open-telemetry/community/issues/629 means that we actually can do that, then I can contribute them.

trask:
Yes, this would be great 👍

iNikem:
Closed by #2617

PR #2568, opened 2021-03-12
Build Windows images for smoke tests

iNikem:
Relates to #2201


trask (approved):
cool, never used windows containers before!

PR #2569, opened 2021-03-12
Let the ComponentInstaller be passed a Config instance

breedx-splk:
This also prefers using the immutable Config instances over the Properties that are created from the config.

This is a breaking change for SPI implementations.  I initially tried to keep both interface methods around (one that takes the config, one that doesn't) but I think that it just served to muddy up the interface.

### DIFF
--- a/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
+++ b/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
@@ -5,8 +5,8 @@
 
 package io.opentelemetry.javaagent.spi.exporter;
 
-import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.sdk.metrics.export.MetricProducer;
+import java.util.Properties;
 import java.util.Set;
 
 /**
@@ -26,7 +26,7 @@ public interface MetricServer {
    * @param producer The metric producer
    * @param config The configuration
    */
-  void start(MetricProducer producer, Config config);
+  void start(MetricProducer producer, Properties config);
 
   /**
    * Returns names of metric servers supported by this factory.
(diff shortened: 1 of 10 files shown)


javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/SpanExporterFactory.java
```diff
@@ -21,7 +21,7 @@
    * @param config The configuration
    * @return An implementation of a {@link SpanExporter}
    */
-  SpanExporter fromConfig(Properties config);
+  SpanExporter fromConfig(Config config);
```
trask, line 24: let's not change this here, we are going to remove `SpanExporterFactory` and `MetricExporterFactory` in #2442, and changing this will create unnecessary churn for @jsuereth and others who are using it in the meanwhile

breedx-splk: 🤷 ok.

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
```diff
@@ -26,7 +26,7 @@
    * @param producer The metric producer
    * @param config The configuration
    */
-  void start(MetricProducer producer, Properties config);
+  void start(MetricProducer producer, Config config);
```
trask, line 29: I don't know if anyone is using this (I hope not) but let's wait to break this SPI until #2442 also

trask (approved):

javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricServer.java
iNikem: @trask did you mean to ask @breedx-splk to revert this change as well?

mateuszrzeszutek (approved):

iNikem:
@breedx-splk Please rebase again. Main branch failure was fixed

PR #2573, opened 2021-03-15
Fix docker login on Windows

iNikem:


anuraaga (approved):

.github/workflows/build-test-matrix.yaml
```diff
@@ -59,11 +59,7 @@ jobs:
           job-id: matrix-smoke
 
       - name: Login to GitHub Package Registry
-        uses: docker/login-action@v1
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GHCR_TOKEN }}
+        run: echo ${{ secrets.GHCR_TOKEN }} | docker login ghcr.io -u ${{ github.repository_owner }} --password-stdin
```
anuraaga, line 62: Nice if you can file and or link to an issue :)

iNikem: It's just a follow-up to broken #2568, because it turned out that action runs only on Linux :(

anuraaga: Yeah :) I just meant we should link to https://github.com/docker/login-action/issues/47 since prefer to use official actions where we can (was out and about before but found it now)

iNikem: A, that is a good point!

anuraaga (approved):

PR #2577, opened 2021-03-15
Run smoke tests on Windows as well

### LINKED ISSUE #2201: Smoke tests on Windows

iNikem:
Closes #2201

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,26 @@ jobs:
         run: ./gradlew build --stacktrace
         working-directory: examples/distro
 
+  smoke-test-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up JDK 11 for running Gradle
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Restore cache
+        uses: burrunan/gradle-cache-action@v1.10
+        with:
+          job-id: smokeTests
+
+      - name: Test
+        run: ./gradlew :smoke-tests:test
+
   test:
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,6 +54,9 @@ jobs:
   smoke-test-windows:
     runs-on: windows-latest
     steps:
+      - name: Support longpaths
+        run: git config --system core.longpaths true
+
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/AppServerTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/AppServerTest.groovy
@@ -36,7 +36,7 @@ abstract class AppServerTest extends SmokeTest {
 
   @Override
   protected String getTargetImage(String jdk) {
-    throw new UnsupportedOperationException("App servers tests should use getAppServerImage")
+    throw new UnsupportedOperationException("App servers tests should use getTargetImagePrefix")
   }
 
   @Override
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/AppServerTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/AppServerTest.groovy
@@ -26,8 +26,25 @@ abstract class AppServerTest extends SmokeTest {
     def appServer = AppServerTestRunner.currentAppServer(this.getClass())
     serverVersion = appServer.version()
     jdk = appServer.jdk()
-    startTarget(jdk, serverVersion)
+
+    boolean isWindows = System.getProperty("os.name").toLowerCase().contains("windows");
+    startTarget(jdk, serverVersion, isWindows)
   }
+
+  @Override
+  protected String getTargetImage(String jdk) {
+    throw new UnsupportedOperationException("App servers tests should use getAppServerImage")
+  }
+
+  @Override
+  protected String getTargetImage(String jdk, String serverVersion, boolean windows) {
+    String platformSuffix = windows ? "-windows" : ""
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yaml
```diff
@@ -51,6 +51,26 @@ jobs:
...
+        with:
+          java-version: 11
+
+      - name: Restore cache
+        uses: burrunan/gradle-cache-action@v1.10
+        with:
+          job-id: smokeTests
+
+      - name: Test
+        run: ./gradlew :smoke-tests:test
```
anuraaga, line 72: I had one random thought that it's very hard to verify this actually tests windows - I don't know if GitHub actions supports it now but don't see any reason it couldn't run linux containers on Windows runner similar to how they run on mac runner. Is there a way to make sure our containers were set up correctly? If we log the container base os during the tests it's enough, just confirming.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/JettySmokeTest.groovy
```diff
@@ -15,7 +15,7 @@ package io.opentelemetry.smoketest
 @AppServer(version = "10.0.0", jdk = "15-openj9")
 class JettySmokeTest extends AppServerTest {
 
-  protected String getTargetImage(String jdk, String serverVersion) {
-    "ghcr.io/open-telemetry/java-test-containers:jetty-${serverVersion}-jdk$jdk-20210223.592806654"
+  protected String getTargetImagePrefix() {
+    "ghcr.io/open-telemetry/java-test-containers:jetty"
```
anuraaga, line 19: I think we decided here that versioning is not important for Java app container tests - it's probably because if versioning caused a problem, other tests would fail. I think it's ok but don't know if we've documented such a decision anywhere

iNikem: Who decided? When? `AppServerTest` still requests a specific container version.

anuraaga: Oh I didn't put them together - I guess XXX is waiting for the other fix? Then no worries.

anuraaga (approved):

.github/workflows/ci.yaml
iNikem: I added an assertion into tests to verify OS

anuraaga: Cool!

trask (approved):
nice 👍

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/AppServerTest.groovy
```diff
@@ -21,14 +22,33 @@ abstract class AppServerTest extends SmokeTest {
...
     jdk = appServer.jdk()
-    startTarget(jdk, serverVersion)
+
+    isWindows = System.getProperty("os.name").toLowerCase().contains("windows")
+    startTarget(jdk, serverVersion, isWindows)
+  }
+
+  @Override
+  protected String getTargetImage(String jdk) {
+    throw new UnsupportedOperationException("App servers tests should use getAppServerImage")
```
mateuszrzeszutek, line 39: There's no method named `getAppServerImage` here

mateuszrzeszutek (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/AppServerTest.groovy
iNikem: Thanks, fixed

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootSmokeTest.groovy
```diff
@@ -11,11 +11,13 @@ import io.opentelemetry.proto.collector.trace.v1.ExportTraceServiceRequest
 import java.util.jar.Attributes
 import java.util.jar.JarFile
 import okhttp3.Request
+import spock.lang.IgnoreIf
 import spock.lang.Unroll
 
+@IgnoreIf({ os.windows })
```
anuraaga, line 17: Should we instead have `WindowsSmokeTest` or something and have `SmokeTest` base class disable itself on windows?

iNikem: `IgnoreIf` does not work on superclass, you have to use it on every concrete class. But yeah, our hierarchy is not convenient for this: we want to run only `AppServerTest` and not other subclasses of `SmokeTest`. That's the only way I found.

anuraaga: I didn't look in detail but was hoping something like `assumeThat` could be used in a base class to control the execution. Mostly want to avoid forgetting this on new tests given windows are the odd-one-out. Can you confirm once more that it's not possible?

anuraaga (approved):
Thanks!

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SpringBootSmokeTest.groovy
iNikem: I didn't find a place to put `assumeThat` on class level :(

iNikem:
Damn! I am an idiot! I totally missed that testcontainers don't run on Windows Server right now and we at Splunk run these tests in quite different way. I will scrap this PR and will have to come up with another one

anuraaga:
@iNikem No clue at all if it works but the first thing that came to mind is trying to disable ryuk on Windows

https://github.com/testcontainers/testcontainers-java/issues/2960#issuecomment-652835627

PR #2581, opened 2021-03-16
Add logback classloader optimization

anuraaga:
While muzzle may prevent this from being a performance issue anyways, I guess we usually want our classloader optimization for this sort of checker.


mateuszrzeszutek (approved):

mateuszrzeszutek:
Muzzle check is executed after `typeMatcher()`, so it won't help with performance here.

iNikem (approved):

PR #2587, opened 2021-03-17
Move attributes to span builder for use by samplers

### LINKED ISSUE #388: Make attributes such as `http.url` available before a span is started
Hi, we run into similar issue as #230 and wish to know the URL when doing sampling decision.

But it seems like with the change in https://github.com/open-telemetry/opentelemetry-auto-instr-java/pull/348/files , the http.url is still not available for example in the Servlet3Advice, the DECORATOR which adds the http.url attribute is not run until the span is started.

Wondering if it's feasible to change the DECORATOR to allow a beforeSpanStart or even change the onRequest and onConnection to take SpanBuilder and have them executed before the span is started?

Many thanks in advance!

trask:
Moves the majority of attributes in `HttpClientTracer` and `HttpServerTracer` from being set on `Span` to being set on `SpanBuilder`, which allows these attributes to be used by samplers.

Fixes #388

### DIFF
--- a/instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
+++ b/instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.instrumentation.servlet;
 
 import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanContext;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.TextMapGetter;
 import io.opentelemetry.instrumentation.api.servlet.AppServerBridge;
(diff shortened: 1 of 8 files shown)


instrumentation-core/servlet-2.2/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -27,7 +28,14 @@
   private static final Logger log = LoggerFactory.getLogger(ServletHttpServerTracer.class);
 
   public Context startSpan(HttpServletRequest request, String spanName) {
-    return startSpan(request, request, request, spanName);
+    Context context = startSpan(request, request, request, spanName);
+
+    SpanContext spanContext = Span.fromContext(context).getSpanContext();
+    // we do this e.g. so that servlet containers can use these values in their access logs
+    request.setAttribute("traceId", spanContext.getTraceId());
+    request.setAttribute("spanId", spanContext.getSpanId());
```
trask, line 36: I'll send a follow-up PR to change these attribute names to `trace_id` and `span_id` to match logging MDC attribute names.

mateuszrzeszutek (approved):
👍 👍 👍

iNikem (approved):

Issue #2591, opened 2021-03-17
ComponentInstallers should run after LogManager regardless of Java version/JFR presence


mateuszrzeszutek:
**Describe the bug**
I've spent a few hours debugging an issue happening on openj9 (IBM) JVM & WildFly where the following exception would happen:

```
java.lang.IllegalStateException: WFLYLOG0078: The logging subsystem requires the log manager to be org.jboss.logmanager.LogManager. The subsystem has not be initialized and cannot be used. To use JBoss Log Manager you must add the system property "java.util.logging.manager" and set it to "org.jboss.logmanager.LogManager"
```

It turns out that in `AgentInstaller#installComponentsAfterByteBuddy()` the agent decided to run the component installers directly, not asynchronously, because JFR is not present on the classpath (JDK <= 8 & custom LogManager conditions passed).
Then in our splunk distro we had a custom `ComponentInstaller` that tried to initialize some metrics and used a `javax.management.NotificationEmitter` -- which indirectly uses JUL, so it initialized `LogManager`, which of course couldn't find the JBoss `LogManager` in the `AgentClassLoader` and defaulted to the JVM implementation. And then the error happened.

I believe that this might become a problem in the future for the vanilla OTel javaagent too, once we implement more standard JVM metrics (that call `javax.management` classes).

To solve this I think that we should remove the `jdk <= 8 && containsJfr` check and only leave the custom `LogManager` one. The JFR stuff (JMXFetch?) is clearly copied from the DD agent and doesn't make sense in OTel javaagent.

**Environment**
Java 1.8.0_282. JVM Eclipse OpenJ9 VM - Eclipse OpenJ9 - openj9-0.24.0

**Additional context**
This is pretty similar to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2546, who knows, maybe the simplified condition will fix it too

trask:
just a heads up on another problem (besides j.u.l.) with using PlatformMBeanServer too early :sob:

https://github.com/glowroot/glowroot/blob/29114b2cabe5115cd39a50774039f8bdf99a5f31/agent/core/src/main/java/org/glowroot/agent/util/LazyPlatformMBeanServer.java#L118-L119

(this was supporting Java 6+, so possibly the situation is better now?)

mateuszrzeszutek:
> (this was supporting Java 6+, so possibly the situation is better now?)

I have no idea, we probably won't know unless something fails on this 🙈

PR #2592, opened 2021-03-17
ComponentInstallers should run after LogManager if a custom one is detected

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2591

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.javaagent.tooling;
 
+import static io.opentelemetry.javaagent.bootstrap.AgentInitializer.isJavaBefore9;
 import static io.opentelemetry.javaagent.tooling.Utils.getResourceName;
 import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.NameMatchers.namedOneOf;
 import static io.opentelemetry.javaagent.tooling.matcher.GlobalIgnoresMatcher.globalIgnoresMatcher;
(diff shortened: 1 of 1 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -170,28 +177,24 @@ private static void installComponentsBeforeByteBuddy(
...
+    // class initialization. Some ComponentInstaller implementations may use jre bootstrap classes
+    // which touch this class (e.g. JFR classes or some MBeans).
+    // This means applications which require a custom LogManager may not have a chance to set the
+    // global LogManager if one of those ComponentInstallers runs first: it will incorrectly
+    // set the global LogManager to the default JVM one in cases where the instrumented application
+    // sets the LogManager system property or when the custom LogManager class is not on the system
+    // classpath.
+    // Our solution is to delay the initialization of ComponentInstallers when we detect a custom
+    // log manager being used.
+    // Once we see the LogManager class loading, it's safe to run ComponentInstaller#after() because
```
trask, line 190:
```suggestion
    // Once we see the LogManager class loading, it's safe to run ComponentInstaller#afterByteBuddyAgent() because
```

trask (approved):
I wonder if we can keep the Java 8 condition. IIRC (and I likely do not), starting in Java 9, Java classes themselves no longer use j.u.l., instead using a new internal logging API.

iNikem:
@trask @mateuszrzeszutek what is the drawback of late installation of components?

mateuszrzeszutek:
> I wonder if we can keep the Java 8 condition. IIRC (and I likely do not), starting in Java 9, Java classes themselves no longer use j.u.l., instead using a new internal logging API.

That's true, Java 9 introduced the new `System.Logger` interface that's supposed to be used internally (https://openjdk.java.net/jeps/264) - let's keep the Java 8 condition then.

> @trask @mateuszrzeszutek what is the drawback of late installation of components?

Hmm, runtime/JVM metrics starting a bit later? The only risk that I can think of is that LogManager never loads and `ComponentInstaller#after...()` never gets called, but with all the conditions and checks that we have this is a very minor risk.

iNikem:
`OpenTelemetryInstaller` is also `ComponentInstaller`. Anything bad happens if its installation is delayed?

mateuszrzeszutek:
No, because it doesn't override the `afterByteBuddyAgent()` method - this "delay after LogManager" thing only affects the `after...` method.

iNikem (approved):

PR #2596, opened 2021-03-18
New Instrumenter API 

anuraaga:
Still draft quality but I think this is ready for some initial feedback

The biggest difference from the current APIs is avoiding inheritance in favor of stitching together parts. The notion is that there are only two big logic behavioral differences between instrumenters - server instrumenters extract context and client ones inject it. Other than that, it's just customization - HTTP semantic conventions drives the customization of span attributes, etc, but there is no big difference in logic it's sort of configuration. So instead of using inheritance, I encapsulate attribute extraction, span name extraction, and status extraction into composable parts. This allows two things that I think an inheritance version has trouble with

- Resolve different conflicting semantic conventions. For example, AWS spans are notorious for having different functionality that needs to be reconciled per-method for determining span name from options of RPC, Messaging, Database, or Armeria also can be used both for RPC or just plain HTTP. So these different types of conventions don't fit neatly into an inheritance tree, we need to be able to compose them. I haven't added an example of this yet in this PR though

- Allow user customization of points they need with builder pattern. I think this style allows keeping instrumentation classes private without losing customization, because customization can be defined in terms of the instrumentation-api. With inheritance, we'd probably need to expose instrumentation classes to allow methods to be overridden.

More than happy to hear thoughts, there are probably some terrible aspects of this too. Though most issues won't be ironed out until trying it on more instrumentation in future PRs.

/cc @as-polyakov


mateuszrzeszutek (commented):
Overall I really like this proposition - the composition is so much nicer (and flexible!) than the inheritance 👍 👍 👍 

Various observations & ideas:
* For some instrumentations (like DB instrumentations) we will probably have to create some DTOs that collect several objects into a `REQUEST` - I don't think that's bad though, and some instrumentations already do this because of various reasons.
* Those instrumentations that have more lifecycle phases than just start & end will be a bit problematic: we'll have to work around all those loose `onRequest()`, `onException()`, `onSomething()` calls.

mateuszrzeszutek:
@anuraaga I've implemented several of my comments in this PR: https://github.com/anuraaga/opentelemetry-java-instrumentation/pull/20
I'd be happy to hear some feedback & comments.
In short, I've completely removed inheritance and added a instrumenter builder that allows the API user to choose which behavior (server or client propagation, internal span) they want in their instrumenter.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public Context start(Context parentContext, REQUEST request) {
+    SpanKind kind = spanKind(request);
+    SpanBuilder spanBuilder =
+        tracer.spanBuilder(spanName(request)).setSpanKind(kind).setParent(parentContext);
+
+    AttributesBuilder attributesBuilder = Attributes.builder();
+    for (AttributesExtractor<? super REQUEST, ? super RESPONSE> extractor : extractors) {
+      extractor.onStart(attributesBuilder, request);
```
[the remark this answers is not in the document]
trask: @anuraaga @jkwatson what do you think about moving `AttributeSetter` interface to the API to unify setting attributes on `Span` and `SpanBuilder`?

Another option we have is to just use `SpanBuilder` here, and come up with a separate solution for the 2 instrumentations that don't fit this usage.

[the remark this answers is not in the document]
anuraaga: Yeah the thought has come to mind but also with the caveat that indeed, I think it's very rare to share the same code for both SpanBuilder and Span which sounds like it's the case.

trask:
thx @anuraaga and @mateuszrzeszutek for giving so much thought to this!

I agree with all the big ideas 👍

once you're ready, I suggest we go for it: merge and start migrating instrumentation one-by-one, uncovering edge cases and working out the smaller stuff as we go

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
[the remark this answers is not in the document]
mateuszrzeszutek: > I think it's very rare to share the same code for both SpanBuilder and Span which sounds like it's the case.

That interface was added to reuse `NetPeerUtils` methods, which was called in various situations 😄 
`NetAttributesExtractor` fixes that issue, so the only use case left that I can imagine is `onRequest()` situation - which we have to solve or find a workaround for anyway.

anuraaga:
Sorry for the delay, this should be ready for review

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+ *   <li>Call {@link Instrumenter#end(Context, Object, Object, Throwable)} at the end of a request.
+ * </ul>
+ */
+public class Instrumenter<REQUEST, RESPONSE> {
+
+  /** Returns a new {@link InstrumenterBuilder}. */
+  public static <REQUEST, RESPONSE> InstrumenterBuilder<REQUEST, RESPONSE> newBuilder(
+      OpenTelemetry openTelemetry,
+      String instrumentationName,
+      SpanNameExtractor<? super REQUEST> spanNameExtractor) {
```
bogdandrutu, line 44: I would prefer to limit the arguments in this method, maybe just otel and instrumentationName are enough and ask users to set the other properties in the builder?

The main reason is that metrics may also have some special "extractor" etc, so I would say let's just have things that are common (or at least we believe are common at this point between traces and metrics).

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+      StatusExtractor<? super REQUEST, ? super RESPONSE> statusExtractor,
+      List<? extends AttributesExtractor<? super REQUEST, ? super RESPONSE>> extractors,
+      ErrorCauseExtractor errorCauseExtractor) {
+    this.tracer = tracer;
+    this.spanNameExtractor = spanNameExtractor;
+    this.spanKindExtractor = spanKindExtractor;
+    this.statusExtractor = statusExtractor;
+    this.extractors = extractors;
+    this.errorCauseExtractor = errorCauseExtractor;
+  }
```
bogdandrutu, line 68: As long as this is package protected passing all the arguments is fine, but if we plan to make this public would suggest to have an `Option` pojo to allow us to add more arguments.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+        tracer
+            .spanBuilder(spanNameExtractor.extract(request))
+            .setSpanKind(spanKind)
+            .setParent(parentContext);
+
+    AttributesBuilder attributes = Attributes.builder();
+    for (AttributesExtractor<? super REQUEST, ? super RESPONSE> extractor : extractors) {
+      extractor.onStart(attributes, request);
+    }
+    attributes.build().forEach((key, value) -> spanBuilder.setAttribute((AttributeKey) key, value));
```
bogdandrutu, line 89: why not directly calling add to the builder?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+   * thrown by the operation, or {@code null} if none was thrown.
+   */
+  public void end(Context context, REQUEST request, RESPONSE response, @Nullable Throwable error) {
+    Span span = Span.fromContext(context);
+
+    AttributesBuilder attributes = Attributes.builder();
+    for (AttributesExtractor<? super REQUEST, ? super RESPONSE> extractor : extractors) {
+      extractor.onEnd(attributes, request, response);
+    }
+    attributes.build().forEach((key, value) -> span.setAttribute((AttributeKey) key, value));
```
bogdandrutu, line 116: why not directly calling add to the span?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+ * instrumenting a library, there will generally be three steps.
+ *
+ * <ul>
+ *   <li>Create an {@link Instrumenter} using {@link InstrumenterBuilder}. Use the builder to
+ *       configure any library-specific customizations, and also expose useful knobs to your user.
+ *   <li>Call {@link Instrumenter#start(Context, Object)} at the beginning of a request.
+ *   <li>Call {@link Instrumenter#end(Context, Object, Object, Throwable)} at the end of a request.
+ * </ul>
+ */
+public class Instrumenter<REQUEST, RESPONSE> {
```
bogdandrutu, line 38: Are there cases where only one generic argument is needed? Would this API fit that as well?

anuraaga: Messaging cases will only have one - in theory it might work but it's not clear. We're going to keep on iterating as we apply to more instrumentations instead of trying to predict too much in advance.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
anuraaga: These are all the required parameters for tracing so they're here - the only default span name we could provide is something like `unknown`, but we'd rather have instrumentation be required to pass it in. Or we could have a "required builder method" but that's not as compile-time-safe.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
anuraaga: This wouldn't become public, that's why we have the builder

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
anuraaga: Since it seems extremely likely that we'll use `Attributes` for metrics too (first version I'd just extract the string key/values when adding our first metric here). If performance is a concern, I'm planning on creating a special implementation of `AttributesBuilder` which is faster but unsafe for general consumption.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
mateuszrzeszutek: We could make tracing/metrics optional "components" of the instrumenter:

```java
Instrumenter.builder(openTelemetry, instrumentationName)
    .enableTracing(spanNameExtractor)         // whatever's required for tracing
    .enableMetrics(metricsSomethingExtractor) // whatever's required for metrics
    // ...
    .newInstrumeter();
```

This API would be a bit more explicit, but also more elastic and less prone to backwards incompatible changes.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
mateuszrzeszutek: > Messaging cases will only have one 

Not necessarily - RocketMQ producer receives `SendResult` as a response, and RabbitMQ sets some attributes using a `Command` that's not available at start span (we might want to either refactor the instrumentation or use `Command` as a (part of) response)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+// TODO(anuraaga): Need to define what are actually useful knobs, perhaps even providing a
+// base-class
+// for instrumentation library builders.
+/**
+ * An instrumenter of the start and end of a request/response lifecycle. Almost all instrumentation
+ * of libraries falls into modeling start and end, generating observability signals from these such
+ * as a tracing {@link Span}, or metrics such as the duration taken, active requests, etc. When
+ * instrumenting a library, there will generally be three steps.
```
mateuszrzeszutek, line 29: Four steps if you include `shouldStart()`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
mateuszrzeszutek, line 38: Should we make `Instrumenter` and `InstrumenterBuilder` interfaces and keep the implementation details in package protected classes? I think that all the implementation details might make the API a bit less readable.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/HttpStatusExtractorTest.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Test
+  void noStatus() {
+    when(extractor.statusCode(anyMap(), anyMap())).thenReturn(null);
+    assertThat(
+            StatusExtractor.http(extractor)
+                .extract(Collections.emptyMap(), Collections.emptyMap(), null))
+        .isEqualTo(StatusCode.UNSET);
+    // Presence of exception has no effect.
```
mateuszrzeszutek, line 50:
```suggestion
    // Presence of exception has effect.
```

WDYT about splitting those test cases?

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/InetSocketAddressNetAttributesExtractorTest.java
```diff
@@ -0,0 +1,90 @@
...
+                SemanticAttributes.NetTransportValues.IP_TCP.getValue()));
+  }
+
+  @Test
+  void fullAddress() {
+    InetSocketAddress address = new InetSocketAddress("github.com", 123);
+    assertThat(address.getAddress().getHostAddress()).isNotNull();
+    when(extractor.getAddress(any(), any())).thenReturn(address);
+    AttributesBuilder attributes = Attributes.builder();
+    extractor.onStart(attributes, null);
```
mateuszrzeszutek, line 52: Combining a mock with final method makes those test kind of non-obvious. How about passing socket address as RQ:

```java
  private InetSocketAddressNetAttributesExtractor<InetSocketAddress, Void> extractor =
      new InetSocketAddressNetAttributesExtractor<InetSocketAddress, Void>() {
        @Override
        protected InetSocketAddress getAddress(InetSocketAddress inetSocketAddress, Void unused) {
          return inetSocketAddress;
        }

        @Override
        protected String transport(InetSocketAddress inetSocketAddress) {
          return null;
        }
      };
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -526,9 +532,17 @@ abstract class HttpClientTest extends InstrumentationSpecification {
       }
       attributes {
         "${SemanticAttributes.NET_TRANSPORT.key}" "IP.TCP"
-        "${SemanticAttributes.NET_PEER_NAME.key}" uri.host
+        if (uri.port == UNUSABLE_PORT) {
+          // TODO(anuraaga): For the unusbale port, there isn't actually a peer so we shouldn't be
```
mateuszrzeszutek, line 536:
```suggestion
          // TODO(anuraaga): For the unusable port, there isn't actually a peer so we shouldn't be
```

mateuszrzeszutek (commented):
@anuraaga please let me know if you need any help with implementing suggestions from comments, I'd like to help with this PR.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * Starts a new operation to be instrumented. The {@code parentContext} is the parent of the
+   * resulting instrumented operation and should usually be {@code Context.current()}. The {@code
+   * request} is the request object of this operation. The returned {@link Context} should be
+   * propagated along with the operation and passed to {@link #end(Context, Object, Object,
+   * Throwable)} when it is finished.
+   */
+  public Context start(Context parentContext, REQUEST request) {
```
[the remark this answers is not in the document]
anuraaga: For better or worse the Armeria instrumentation wasn't using it I guess. I'd like to add it in a follow-up after making sure it's needed.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ClientInstrumenter.java
```diff
@@ -0,0 +1,45 @@
...
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.context.propagation.TextMapSetter;
+import java.util.List;
+
+final class ClientInstrumenter<REQUEST, RESPONSE> extends Instrumenter<REQUEST, RESPONSE> {
```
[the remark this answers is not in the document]
anuraaga: Ah yeah - I understood the comment, I am still on the fence on it. Considering how much more common server/client is than the other forms I wonder if it's not so bad to have server / client words even when it's actually consumer / producer. Was thinking of addressing this in a follow-up when using from messaging.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/NetAttributesExtractor.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  protected final void onEnd(AttributesBuilder attributes, REQUEST request, RESPONSE response) {
+    set(attributes, SemanticAttributes.NET_PEER_IP, peerIp(request, response));
+
+    // TODO(anuraaga): Clients don't have peer information available during the request usually.
+    // By only resolving them after the response, we can simplify the code a lot but sacrifice
+    // having them available during sampling on the server side. Revisit if that seems important.
+    set(attributes, SemanticAttributes.NET_PEER_NAME, peerName(request, response));
+    set(attributes, SemanticAttributes.NET_PEER_PORT, peerPort(request, response));
```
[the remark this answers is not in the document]
anuraaga: Yeah I am considering adding peer.service as a parameter actually to methods that return a client instrumenter. Maybe not but since we don't have any good instrumentation tests exercising this was going to think in a follow-up.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
anuraaga: I wouldn't split impl just for that I think, if we expect other implementations than yeah. We can have detailed docs at the class level eventually which would be the best for either case I guess.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ClientInstrumenter.java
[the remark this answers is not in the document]
mateuszrzeszutek: Just to clarify: we can still have `newServerInstrumenter(TextMapGetter)`, `new(Upstream|Downstream)Instrumenter()` would be just for the situation when you have to pass a custom `SpanKindExtractor`

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/InetSocketAddressNetAttributesExtractorTest.java
anuraaga: Yeah I was a bit concerned so with confirmation I guess it is a problem. Sounds good - by the way what does RQ stand for? :D

anuraaga:
Thanks a lot @mateuszrzeszutek - I've left some thumbs on ones that seem trivial for this PR and comments on ones I lean towards following up. If you'd like to send a pr for them sounds great or else will address tomorrow

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/InetSocketAddressNetAttributesExtractorTest.java
mateuszrzeszutek: Request -- that's an abbreviation that I'm used to, somehow it slipped my mind that it may be not that obvious 😄

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ClientInstrumenter.java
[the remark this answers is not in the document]
anuraaga: Yeah it's a tradeoff between introducing new concepts vs having precise terminology. If server and producer are close enough to be two ducks (i.e, if it quacks it's a duck) then consolidating might be ok. I'm not too sure on it yet.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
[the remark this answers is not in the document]
mateuszrzeszutek: It is, it was added very recently here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2646/files#diff-6cbd0e0fc20c6b027748332652df89740a096fd45730f2aaba66d0d97eea1540

mateuszrzeszutek:
https://github.com/anuraaga/opentelemetry-java-instrumentation/pull/21

anuraaga:
This should be ready for another review now, and I think it's probaby a reasonable first try to iterate on in future PRs

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaTracingBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.armeria-1.3";
+
+  private final OpenTelemetry openTelemetry;
+
+  private final List<AttributesExtractor<? super RequestContext, ? super RequestLog>>
+      additionalExtractors = new ArrayList<>();
+
+  private Function<
+          SpanStatusExtractor<RequestContext, RequestLog>,
+          ? extends SpanStatusExtractor<? super RequestContext, ? super RequestLog>>
```
trask, line 34: 😅

instrumentation/armeria-1.3/library/armeria-1.3-library.gradle
```diff
@@ -1,5 +1,5 @@
 apply from: "$rootDir/gradle/instrumentation-library.gradle"
-apply plugin: "net.ltgt.errorprone"
+// apply plugin: "net.ltgt.errorprone"
```
trask, line 2: add a TODO here?

trask (approved):
> I think it's probaby a reasonable first try to iterate on in future PRs

👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpAttributesExtractor.java
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public abstract class HttpAttributesExtractor<REQUEST, RESPONSE>
```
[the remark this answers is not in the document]
anuraaga: Hmm - I think it means we wouldn't have `SpanStatusExtractor.http()` (not in same package anymore) and just use `HttpStatusExtractor.create()` or something. I'm wondering about this vs extracting an `attributes` package with `HttpAttributesExtractor`, `NetAttributesExtractor`, etc, that does keep the public API together with some separation.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpAttributesExtractor.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Override
+  protected final void onStart(AttributesBuilder attributes, REQUEST request) {
+    set(attributes, SemanticAttributes.HTTP_METHOD, method(request));
+    set(attributes, SemanticAttributes.HTTP_URL, url(request));
+    set(attributes, SemanticAttributes.HTTP_TARGET, target(request));
+    set(attributes, SemanticAttributes.HTTP_HOST, host(request));
+    set(attributes, SemanticAttributes.HTTP_ROUTE, route(request));
+    set(attributes, SemanticAttributes.HTTP_SCHEME, scheme(request));
+    set(attributes, SemanticAttributes.HTTP_USER_AGENT, userAgent(request));
```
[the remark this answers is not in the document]
anuraaga: For this API, we just set what was provided indeed, I don't think we can model something like that at this layer since we don't know what can be inferred, or if we do then it needs to be more explicit in the spec I guess, that spec language is way too vague to be actually implementable I think.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -0,0 +1,127 @@
...
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.tracer.ClientSpan;
+import io.opentelemetry.instrumentation.api.tracer.ServerSpan;
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+// TODO(anuraaga): Need to define what are actually useful knobs, perhaps even providing a
+// base-class
+// for instrumentation library builders.
```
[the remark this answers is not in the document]
anuraaga: Once the API settles down we definitely need an example / docs. Is working on instrumentation modules, in this PR, Armeria, enough for these early iterations?

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpAttributesExtractor.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Override
+  protected Long responseContentLength(RequestContext ctx, RequestLog requestLog) {
+    return requestLog.responseLength();
+  }
+
+  @Override
+  protected @Nullable Long responseContentLengthUncompressed(
+      RequestContext ctx, RequestLog requestLog) {
+    return null;
```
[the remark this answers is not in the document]
anuraaga: `return null` by default does make instrumentation easier to write, but I worry it's too easy. Explicitly opting-out of an attribute because you can't implement it seems to ensure higher coverage of the attributes than requiring opt-in. `return null` isn't so bad and what IntelliJ generates by default anyways I guess

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/OpenTelemetryClient.java
```diff
@@ -32,23 +32,16 @@ public HttpResponse execute(ClientRequestContext ctx, HttpRequest req) throws Ex
...
 
     Span span = Span.fromContext(context);
     if (span.isRecording()) {
       ctx.log()
           .whenComplete()
           .thenAccept(
               log -> {
-                clientTracer.getNetPeerAttributes().setNetPeer(span, ctx.remoteAddress());
-
                 long requestEndTimeNanos = requestStartTimeNanos + log.responseDurationNanos();
```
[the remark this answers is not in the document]
anuraaga: Decided at least for now to remove it and might rethink later

instrumentation/armeria-1.3/library/armeria-1.3-library.gradle
anuraaga: Oops

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
anuraaga: @mateuszrzeszutek - Hmm - this is an interesting proposal, would like to think it through more as a followup. One open question is whether we want to make it easy to only enable one signal and not the other.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpAttributesExtractor.java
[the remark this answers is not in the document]
iNikem: WDYM, vague? Spec gives several options for required attribute combinations and then warns agains duplication.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpAttributesExtractor.java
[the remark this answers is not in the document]
mateuszrzeszutek: Or `HttpSemanticConventions.spanName()`/`HttpSemanticConventions.status()` in the `http` package

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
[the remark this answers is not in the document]
mateuszrzeszutek: That was just an idea for future, once the API is somewhat stable - I think we should concentrate on covering most of use cases with Instrumenter right now.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpAttributesExtractor.java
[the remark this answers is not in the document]
anuraaga: Well these attributes are defined by the spec so it could say exactly which attributes should be filled. If it's the three sets of attributes and says attributes SHOULD NOT be filled from one set if another is filled, it would be clearer. I don't know either if this is a ordered list by priority. So vague ;)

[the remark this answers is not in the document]
anuraaga: E.g, in that text server name is mentioned descriptively but it could just be said it should always be filled when available.

mateuszrzeszutek (approved):

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpAttributesExtractor.java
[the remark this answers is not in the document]
trask: I like the idea of forcing explicit implementation (and hopefully documentation) of the unsupported conventions

Issue #2554, opened 2021-03-11
Muzzle should fail at compile time if it fails at runtime


anuraaga:
Currently, muzzle may fail at runtime even if the compile time check succeeds, for example if instrumentation refers to plugin classes in the AWS SDK. This seems to be a bug in muzzle - the `muzzle` extension declaration only specifies `aws-core` as an artifact, but it seems to also take in instrumentation's `compileClasspath` and can pass some checks that might fail in production. I think we should respect the `muzzle` declaration only for checks, not using the compile classpath - if extra dependencies are needed, they can be added to the muzzle block too, it's how to have a more robust check.

mateuszrzeszutek:
Adding compile deps to the classloader in MuzzlePlugin is definitely weird - why would you want to do that? Especially when there's even an `extraDependency` flag in the muzzle directive configuration, which seems *perfect* for including additional deps in the muzzle task classpath.

jaydeluca:
is this still a problem or did #2599 address it?

PR #2599, opened 2021-03-18
Make sure muzzle build-time check actually validates anything

### LINKED ISSUE #2110: Investigate muzzle builds not failing when they should on GHA
**Describe the bug**
Sometimes muzzle tasks on GHA succeed even though they should not - and when you run the same muzzle task locally it fails (as expected).

See: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2094#issuecomment-765569295, https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2094#issuecomment-765671043

Muzzle tasks on GHA should behave same as locally (i.e. fail when they are expected to fail).

iNikem:
Fixes #2110


trask:
🤯 really looking forward to hearing the next chapter of this story (I do hope it's not too scary).

buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -95,10 +95,7 @@ class MuzzlePlugin implements Plugin<Project> {
     project.tasks.compileMuzzle.dependsOn(bootstrapProject.tasks.compileJava)
     project.tasks.compileMuzzle.dependsOn(toolingProject.tasks.compileJava)
     project.afterEvaluate {
-      project.tasks.compileMuzzle.dependsOn(project.tasks.compileJava)
-      if (project.tasks.getNames().contains('compileScala')) {
-        project.tasks.compileMuzzle.dependsOn(project.tasks.compileScala)
-      }
+      project.tasks.compileMuzzle.dependsOn(project.tasks.classes)
```
mateuszrzeszutek, line 98: This seems to be the actual fix of our muzzle problems - what does it change? We still use `compileJava` two lines above, is that OK?

instrumentation/play-ws/play-ws-common/library/play-ws-common-library.gradle
```diff
@@ -1,7 +1,9 @@
-apply from: "$rootDir/gradle/instrumentation.gradle"
+apply from: "$rootDir/gradle/instrumentation-library.gradle"
 
 def scalaVersion = '2.12'
 
 dependencies {
+  compileOnly project(':javaagent-api')
+  compileOnly project(':javaagent-tooling')
```
mateuszrzeszutek, line 7: Hmm, it's not really a "piece of library instrumentation" if it depends on javaagent artifacts... I wonder if we need another "module type" like `javaagent-shared` for those modules. (In another issue/PR though)

gradle/instrumentation.gradle
```diff
@@ -20,6 +19,11 @@ apply from: "$rootDir/gradle/instrumentation-common.gradle"
 
 if (projectDir.name == 'javaagent') {
   archivesBaseName = projectDir.parentFile.name
+
+  //This is a special project that has only ComponentInstaller but not InstrumentationModule
+  if (archivesBaseName != 'runtime-metrics') {
```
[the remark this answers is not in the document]
iNikem: I have thought about that. Currently if there is no muzzle directive we verify against JDK classes. I don't know if we actually need that. @trask ?

instrumentation/rxjava-2.0/javaagent/rxjava-2.0-javaagent.gradle
```diff
@@ -4,7 +4,9 @@ muzzle {
   pass {
     group = "io.reactivex.rxjava2"
     module = "rxjava"
-    versions = "[2.0.0,)"
+    versions = "[2.0.6,)"
+    extraDependency "io.reactivex.rxjava2:rxjava:2.0.6"
```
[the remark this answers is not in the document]
iNikem: this module does not depend on this jar. `library` module does. This is the same case as with aws-sdk. We should change MuzzlePlugin to be able to handle such situation automatically, but currently it does not.

buildSrc/src/main/groovy/MuzzlePlugin.groovy
iNikem: Yes, toolingProject and bootstrapProject are Ok to use `compileJava`, although we may want to change that to `classes` as well for consistency. The problem is that in instrumentation modules we use a separate bytebuddy compilation plugin. And that plugin messes with `compileJava` task. Using `classes` instead guarantees that byte buddy ran as well.

instrumentation/play-ws/play-ws-common/library/play-ws-common-library.gradle
iNikem: Yeah, I am not happy about this as well, but don't have any better solution that does not explode the scope.

gradle/instrumentation.gradle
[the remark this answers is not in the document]
mateuszrzeszutek: Hmm, we do have a separate directive syntax for JDK:

```groovy
muzzle {
  pass {
    coreJdk()
  }
}
```

instrumentation/rxjava-2.0/javaagent/rxjava-2.0-javaagent.gradle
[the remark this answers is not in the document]
mateuszrzeszutek: Maybe this will be fixed by https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2554 once it's implemented

buildSrc/src/main/groovy/MuzzlePlugin.groovy
mateuszrzeszutek: Can you add a comment about that? That was kind of non-obvious fix

gradle/instrumentation.gradle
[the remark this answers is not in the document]
anuraaga: Agree that we shouldn't run muzzle if nothing is declared if it's simple to implement.

instrumentation/play-ws/play-ws-common/library/play-ws-common-library.gradle
anuraaga: It look like the common -> library pattern worked out ok for most artifacts. For this one how about same hack as runtime metrics?

buildSrc/src/main/groovy/MuzzlePlugin.groovy
iNikem: I simplified and unified this.

instrumentation/play-ws/play-ws-common/library/play-ws-common-library.gradle
iNikem: Got back to javaagent

mateuszrzeszutek (approved):

gradle/java.gradle
```diff
@@ -94,6 +94,7 @@ jar {
 repositories {
   mavenCentral()
   mavenLocal()
+  jcenter()
```
trask, line 97: is this change intentional?

instrumentation/elasticsearch/elasticsearch-rest-common/library/elasticsearch-rest-common-library.gradle
```diff
@@ -0,0 +1,6 @@
+apply from: "$rootDir/gradle/instrumentation-library.gradle"
+
+dependencies {
+  compileOnly project(':javaagent-api')
```
trask, line 4: this doesn't look like it's needed
```suggestion
```

instrumentation/elasticsearch/elasticsearch-transport-common/library/elasticsearch-transport-common-library.gradle
```diff
@@ -0,0 +1,6 @@
+apply from: "$rootDir/gradle/instrumentation-library.gradle"
+
+dependencies {
+  compileOnly project(':javaagent-api')
```
trask, line 4: same
```suggestion
```

instrumentation/hibernate/hibernate-common/library/hibernate-common-library.gradle
```diff
@@ -0,0 +1,9 @@
+/*
+ * Classes that are common to all versions of the Hibernate instrumentation.
+ */
+
+apply from: "$rootDir/gradle/instrumentation-library.gradle"
+
+dependencies {
+  compileOnly project(':javaagent-api')
```
trask, line 8: since this needs `javaagent-api`, does it make sense to keep this module as `hibernate-common-javaagent`?

instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
```diff
@@ -4,7 +4,7 @@ muzzle {
   pass {
     group = "org.apache.rocketmq"
     module = 'rocketmq-client'
-    versions = "[4.8.0,)"
+    versions = "[4.0.0,)"
```
trask, line 7: @addname if tests pass against 4.0, let's run tests against that version and rename the module to `rocketmq-client-4.0`

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
```diff
@@ -18,6 +19,7 @@
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 
+@AutoService(InstrumentationModule.class)
```
trask, line 22: um, how were tests passing? 😟

instrumentation/jaxws/jaxws-common/library/jaxws-common-library.gradle
```diff
@@ -0,0 +1,5 @@
+apply from: "$rootDir/gradle/instrumentation-library.gradle"
+
+dependencies {
+  compileOnly project(':javaagent-api')
+}
```
trask, line 5: doesn't look needed
```suggestion
```

instrumentation/play-ws/play-ws-common/javaagent/play-ws-common-javaagent.gradle
```diff
@@ -3,5 +3,7 @@ apply from: "$rootDir/gradle/instrumentation.gradle"
 def scalaVersion = '2.12'
 
 dependencies {
+  compileOnly project(':javaagent-api')
+  compileOnly project(':javaagent-tooling')
```
trask, line 7: looks unused after switching back to `-javaagent`
```suggestion
```

trask (commented):
since you updated muzzle to ignore modules that don't have a muzzle directive, should we move the `-common-library` modules back to `-common-javaagent`? seems a little confusing to have `-library` modules (which will show up under the maven groupId with other library modules) for instrumentation that does not really have library instrumentation

instrumentation/grails-3.0/javaagent/grails-3.0-javaagent.gradle
```diff
@@ -5,7 +5,7 @@ muzzle {
     group = "org.grails"
     module = "grails-web-url-mappings"
     versions = "[3.0,)"
-    skipVersions += ['3.1.15']
+    skipVersions += ['3.1.15', '3.3.6']
```
trask, line 8: I saw a recent build fail on `3.3.6` too, but I don't see it in the maven central metadata: https://repo1.maven.org/maven2/org/grails/grails-web-url-mappings/maven-metadata.xml, and I can't reproduce it locally..

trask: I pushed this `skipVersions` line to #2611 and will merge there to unblock a couple of PRs that look stuck on this too

trask:
(fyi I merged main into your branch to resolve the merge conflict I created)

instrumentation/grails-3.0/javaagent/grails-3.0-javaagent.gradle
iNikem: Grails module has a separate repository, `https://repo.grails.org/grails/core`. It does have `3.3.6` but that has a SNAPSHOT transitive dependency which we fail to retrieve.

trask: oh I totally missed that, thx (I just needed to read down a few lines 😭)

gradle/java.gradle
iNikem: No, removed.

instrumentation/hibernate/hibernate-common/library/hibernate-common-library.gradle
iNikem: Moved back

mateuszrzeszutek (approved):

anuraaga (approved):

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
trask: ah, looks like rxjava2 javaagent tests just need the executor instrumentation to pass, I guess that's not too surprising

iNikem: It should be. This means that test is not sensitive and thus useless :(

trask: cc: @piotr-sumo when you have time can you see if there's a test we can add for rxjava2 that doesn't pass purely due to executor instrumentation?

anuraaga: This is being addressed in #2672

instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
trask: nevermind, i took a quick look and tests didn't compile against earlier versions, not a problem to keep as-is

PR #2602, opened 2021-03-19
Add protocol-core to muzzle dependencies for aws sdk v2.

anuraaga:


buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -358,6 +358,9 @@ class MuzzlePlugin implements Plugin<Project> {
       config.dependencies.add(dep)
     }
     for (String additionalDependency : muzzleDirective.additionalDependencies) {
+      if (additionalDependency.count(":") < 2) {
```
iNikem, line 361: What does this condition mean?

anuraaga: Added comment, thanks

mateuszrzeszutek: Use the same version as the main artifact? Maybe we should use some placeholder like `$MAIN_LIB_VERSION` and stirng.replace() it here.

instrumentation/aws-sdk/aws-sdk-2.2/javaagent/aws-sdk-2.2-javaagent.gradle
```diff
@@ -5,6 +5,9 @@ muzzle {
     group = "software.amazon.awssdk"
     module = "aws-core"
     versions = "[2.2.0,)"
+    // Used by all SDK services, the only case it isn't is an SDK extension such as a custom HTTP
+    // client, which is not target of instrumentation anyways.
+    additionalDependencies = ["software.amazon.awssdk:protocol-core"]
```
mateuszrzeszutek, line 10: I think that we should use the `extraDependency` syntax that's being used everywhere else

```suggestion
        extraDependency "software.amazon.awssdk:protocol-core"
```

mateuszrzeszutek (approved):

instrumentation/aws-sdk/aws-sdk-2.2/javaagent/aws-sdk-2.2-javaagent.gradle
anuraaga: Oops

buildSrc/src/main/groovy/MuzzlePlugin.groovy
anuraaga: I thought of a placeholder approach - I couldn't use `$` though since then users have to be careful on single quotes vs double quotes or the variable will be replaced by Groovy, not us. Just checking for a versionless artifact seems reasonably simple to me (or I could use a different template syntax like `|VERSION|`

mateuszrzeszutek (approved):

iNikem (approved):

buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -358,6 +358,10 @@ class MuzzlePlugin implements Plugin<Project> {
...
+        // Dependency definition without version, use the artifact's version.
```
trask, line 362: nice

Issue #2586, opened 2021-03-16
:instrumentation:rocketmq-client-4.8:library:test is flaky


iNikem:
It sporadically fails during PRs and fails 2 times out of 4 when trying to run locally

trask:
cc: @addname

iNikem:
Some more problems with those tests:

- They never shutdown resources started in `BaseConf` and in `IntegrationTestBase`, such as broker, name server, producer and consumer
- They never clean up the files they created, `BaseConf.deleteTempDir()` is never called
- `RocketMqClientTest` uses deprecated methods. Is it expected? Should end-users do the same?

addname:
My local run has been successful, let me solve this issue

trask:
thx @addname!

if you are not able to repro locally, try reverting #2595 and open a PR against main branch in your own repo in order to trigger GitHub Actions CI (doing this in your own repo will allow you to re-run CI a couple of times if needed)

if you have trouble reproducing this way let us know!

mateuszrzeszutek:
I've noticed that only the `test rocketmq produce and batch consume` test case fails randomly (most of the time): the consumer spans never appear. Maybe this'll be a hint for you.

addname:
thanks @trask @mateuszrzeszutek 

Rocketmq client (producer/consumer) can only use hooks through `getDefaultMQProducerImpl()` and `getDefaultMQPushConsumerImpl()`. Rocketmq client marks `getDefaultMQProducerImpl()` and `getDefaultMQPushConsumerImpl()` as @Deprecated, but the current version does not provide alternative methods. If a later version provides a recommended method, I will update it.

The test class of rocketmq clinet uses hooks in the same way.
https://github.com/apache/rocketmq/blob/master/client/src/test/java/org/apache/rocketmq/client/producer/DefaultMQProducerTest.java

PR #2610, opened 2021-03-20
Fix support rocketmq client test

addname:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2586

### DIFF
--- a/instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
+++ b/instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
@@ -17,5 +17,6 @@ dependencies {
 }
 
 tasks.withType(Test) {
+  jvmArgs "--add-exports=java.base/jdk.internal.ref=ALL-UNNAMED"
   jvmArgs "-Dotel.instrumentation.rocketmq-client.experimental-span-attributes=true"
 }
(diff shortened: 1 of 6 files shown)


instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
```diff
@@ -17,5 +17,6 @@ dependencies {
 }
 
 tasks.withType(Test) {
+  jvmArgs += ["--add-exports", "java.base/jdk.internal.ref=ALL-UNNAMED"]
```
anuraaga, line 20: Thanks @addname - is this needed for rocketmq server, not client? If so, is it possible to use test containers with docker instead of having to use this custom Java config? We use it in a few tests.

addname: `brokerController.shutdown()` needs to be set `--add-exports`

iNikem (approved):

trask (approved):
thx @addname!

instrumentation/rocketmq-client-4.8/javaagent/rocketmq-client-4.8-javaagent.gradle
anuraaga: Sounds like we'd be able to use testcontainers then - we can do it in another PR though

anuraaga (approved):

PR #2611, opened 2021-03-21
Increase smoke test timeout

trask:
Saw a failure due to the 30 min timeout. Checked and saw a couple of successful runs in the 20-25 min range, so makes sense to bump for now until we break it up into multiple jobs.


iNikem (approved):
Some day we are going to need a more long-term solution :(

anuraaga (approved):

PR #2601, opened 2021-03-19
Use Caffeine for weak maps

anuraaga:
I'd like to use weak maps in library instrumentation too in some cases and we already have it as part of the caching API. I checked some very basic benchmarks and it looks like weaklockfree could be a bit better at low contention with caffeine better at high contention, though they're pretty similar. I did notice with Datadog that they seem to use Caffeine for Java 8+ and presumably benchmarked much better than me so think this is ok. https://github.com/DataDog/dd-trace-java/pull/2044 It wouldn't be hard to switch to weaklockfree in the future though if needed.

Edit: squashed just has a silly bug

```
Benchmark                                                                              Mode  Cnt          Score           Error   Units
WeakMapBenchmark.threads1_caffeine                                                    thrpt   15          5.298 ±         0.787  ops/us
WeakMapBenchmark.threads1_caffeine:heap.total.after                                   thrpt   15  157146589.867 ±   2315509.766   bytes
WeakMapBenchmark.threads1_caffeine:heap.total.before                                  thrpt   15   45857723.733 ±   4631019.532   bytes
WeakMapBenchmark.threads1_caffeine:heap.used.after                                    thrpt   15   57109048.533 ±  26323528.636   bytes
WeakMapBenchmark.threads1_caffeine:heap.used.before                                   thrpt   15    8902554.133 ±     22938.052   bytes
WeakMapBenchmark.threads1_caffeine:·gc.alloc.rate                                     thrpt   15        810.426 ±       150.078  MB/sec
WeakMapBenchmark.threads1_caffeine:·gc.alloc.rate.norm                                thrpt   15        240.566 ±        21.685    B/op
WeakMapBenchmark.threads1_caffeine:·gc.churn.G1_Eden_Space                            thrpt   15        784.116 ±       156.570  MB/sec
WeakMapBenchmark.threads1_caffeine:·gc.churn.G1_Eden_Space.norm                       thrpt   15        231.973 ±        22.565    B/op
WeakMapBenchmark.threads1_caffeine:·gc.churn.G1_Survivor_Space                        thrpt   15          0.001 ±         0.001  MB/sec
WeakMapBenchmark.threads1_caffeine:·gc.churn.G1_Survivor_Space.norm                   thrpt   15         ≈ 10⁻³                    B/op
WeakMapBenchmark.threads1_caffeine:·gc.count                                          thrpt   15        337.000                  counts
WeakMapBenchmark.threads1_caffeine:·gc.time                                           thrpt   15         91.000                      ms
[...]

PR #2613, opened 2021-03-22
Run instrumentation that injects helper resources last.

anuraaga:
While working on #2601 I noticed that ever since adding HelperResources, there's a weak map in the bootstrap classloader. It brought me to think about #2612 - since weak map uses runnable, does it cause runnable to become uninstrumentable? This is just a strawman's PR, without knowing how to verify #2612 it doesn't make too muh sense.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -246,7 +246,10 @@ public class AgentInstaller {
     return SafeServiceLoader.load(
             InstrumentationModule.class, AgentInstaller.class.getClassLoader())
         .stream()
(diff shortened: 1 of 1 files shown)


trask:
I'm not sure that the order of applying instrumentation matters for this problem, the important thing is that we minimize the classes we load/use before installing the instrumentation (i.e. before installing the bytebuddy agent).

anuraaga:
Yeah - I thought if helperresources is known to classload e.g., Runnable (not confirmed yet) running last would allow instrumentation that modifies Runnable to have a better chance of applying before classload.

trask:
I'm not totally following, but in case it helps, any class that is loaded under `ClassFileTransformer.transform()` sadly won't be transformed itself

trask:
or maybe that is your point... this one may be easier to talk through on ThuFri

anuraaga:
Discussed and realized it's not the order instrumentation is applied that matters - if a class happens to be loaded anywhere during transform it will cause problems regardless of order.

PR #2614, opened 2021-03-22
Migrate rocketmq test to testcontainers.

anuraaga:
@addname I've tried migrating the rocketmq server to testcontainers. It seems to start up correctly and I've confirmed with `MQAdmin.getCluster` that the nameserver has the correct IP address of the broker. But still the messages don't seem to send correctly - I have tried both with manual `MQAdmin.createTopic` and `autoCreateTopicEnabled` (latter is better if it works for simpler test) and it seems to log a failure to create the topic. I couldn't find any problems in logs when using `docker exec -i containerid bash` to inspect. Do you know how to debug this sort of rocketmq connection issue? Thanks!


addname (commented):
Whether it is this exception: `MQClientException: No route info of this topic`?

You can try to configure the broker server as follows:
Set in broker.conf:
brokerIP1 = rocketmq server external IP (physical IP)
autoCreateTopicEnable = true

https://github.com/apache/rocketmq/issues/504
https://github.com/apache/rocketmq-docker/blob/master/product/README.md

In addition, we are now using rocketmq client version 4.8.0

anuraaga:
@addname I realized the problem is that we can't override the broker IP address for the client. The nameserver and broker are assigned docker IP addresses separate from localhost, and docker maps localhost ports to those containers. It means we can pass the mapped nameserver address to the client, but it will still retrieve docker IP addresses from the nameserver (which it received from the broker) and the client attempts to connect directly to it which can't work. So I guess it's not possible to use testcontainers unless it's possible to override at least the port connected to at client connect time (`brokerIP1` can only override the IP which would help, but the port isn't known since it's created by testcontainers).

addname:
The broker running in the docker container uses the property `brokerIP1 `to specify the address of the host, rocketmq docker runs successfully, and there is no problem with the port. Can testcontainersnen use the property `brokerIP1 `to specify the address of the host?

Setting rocketmq broker IP1 in docker
 >cd /apache-rocketmq 
 >echo "brokerIP1=10.6.99.105" > conf/broker.conf
 >nohup sh bin/mqbroker -n 10.6.99.105:9876 -c conf/broker.conf &

anuraaga:
@addname The issue seems to be that with testcontainers, we don't know the real port until after the container starts, which seems to be too late to reconfigure the DNS or broker. Oh well, at least the test seems to have gotten less flaky :)

PR #2615, opened 2021-03-22
Make muzzle reference creation package(s) configurable

### LINKED ISSUE #1395: Make muzzle reference creation package(s) configurable

mateuszrzeszutek:
... and use this new feature in Couchbase 3.1 instrumentation - now it has actual muzzle checks.

Resolves #1395

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
@@ -253,15 +253,28 @@ public abstract class InstrumentationModule {
   }
 
   /**
+   * This is an internal helper method for muzzle code generation: generating {@code invokedynamic}
+   * instructions in ASM is so painful that it's much simpler and readable to just have a plain old
+   * Java helper function here.
+   */
+  @SuppressWarnings("unused")
+  protected final Predicate<String> additionalLibraryInstrumentationPackage() {
+    return this::isHelperClass;
+  }
+
+  /**
    * Instrumentation modules can override this method to specify additional packages (or classes)
    * that should be treated as "library instrumentation" packages. Classes from those packages will
    * be treated by muzzle as instrumentation helper classes: they will be scanned for references and
    * automatically injected into the application classloader if they're used in any type
-   * instrumentation. The class name predicate returned by this method is added to the default ones
-   * defined in {@link InstrumentationClassPredicate}.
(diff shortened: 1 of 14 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
```diff
@@ -251,6 +252,18 @@ private String mainInstrumentationName() {
...
 
+  /**
+   * Instrumentation modules can override this method to specify additional packages (or classes)
+   * that should be treated as "library instrumentation" packages. Classes from those packages will
+   * be treated by muzzle as instrumentation helper classes: they will be scanned for references and
+   * automatically injected into the application classloader if they're used in any type
+   * instrumentation. The class name predicate returned by this method is added to the default ones
+   * defined in {@link InstrumentationClassPredicate}.
+   */
+  public Predicate<String> additionalLibraryInstrumentationPackage() {
```
iNikem, line 263: Please describe in more details what String is being passed to this predicate.

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
trask: a thought, not sure, on trying to make it clearer
```suggestion
  public boolean isHelperClass(String className) {
    return false;
  }
```

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
mateuszrzeszutek: Good point 👍 

It's a good chance to learn how method references are represented in the bytecode 😂

PR #2616, opened 2021-03-22
Add runtime.jvm.gc.collection.count metric

felixbarny:
The OTel agent already reports the collection time per GC. 

The collection count can be useful to compare the count and the duration to see if there are a lot of short GCs or few long GCs, for example.
Please let me know whether you think adding the collection count would also be interesting for OTel.

Background: I'm working for Elastic and we're trying to convert the OTel semantic convention attributes to the Elastic Common Schema so that users can use the OTel agent with the Elastic Stack and the Elastic APM UI. As a part of that effort, we also want to convert the metrics (see https://github.com/elastic/apm-server/issues/4919). We have a screen in our UI displays [JVM runtime metrics](https://www.elastic.co/guide/en/kibana/current/metrics.html). One of the charts uses not only the collection time but also the collection count per GC.

### DIFF
--- a/instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/GarbageCollector.java
+++ b/instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/GarbageCollector.java
@@ -25,8 +25,8 @@ import java.util.List;
  * <p>Example metrics being exported:
  *
  * <pre>
- *   runtime.jvm.gc.collection.time{gc="PS1"} 6.7
- *   runtime.jvm.gc.collection.count{gc="PS1"} 1
+ *   runtime.jvm.gc.time{gc="PS1"} 6.7
+ *   runtime.jvm.gc.count{gc="PS1"} 1
  * </pre>
  */
 public final class GarbageCollector {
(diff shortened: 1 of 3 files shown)


iNikem (approved):

instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/GarbageCollector.java
```diff
@@ -51,6 +52,18 @@ public static void registerObservers() {
               }
             })
         .build();
+    meter
+        .longSumObserverBuilder("runtime.jvm.gc.collection.count")
+        .setDescription("The number of collections that have occurred for a given JVM garbage collector.")
```
felixbarny, line 57:
```suggestion
        .setDescription(
            "The number of collections that have occurred for a given JVM garbage collector.")
```

mateuszrzeszutek (approved):

trask (approved):
hey @felixbarny, good to see you here!

anyone have any objections to renaming these:

* runtime.jvm.gc.collection -> runtime.jvm.gc.time
* runtime.jvm.gc.collection.count -> runtime.jvm.gc.count

we haven't declared stability of metric names, so it's ok for us to rename existing metrics, and I think these are both shorter and clearer

I can do this in a follow-up PR also.

anuraaga (approved):
Agree with @trask's names :) There are many GC'd languages, getting `runtime.*.gc.*` into the spec would probably be nice

PR #2618, opened 2021-03-22
Add asynchronous tracing to Spring Boot Autoconfigure WithSpanAspect

HaloFour:
Updates the WithSpanAspect in Spring Boot Autoconfigure module to support asynchronous return types.  Refactors the MethodSpanStrategies implementations from OpenTelemetry Annotations module to enable sharing between instrumentation projects.

Open question:  Should the async method strategy machinery belong in the instrumentation-api module or would it be better suited in a separate AOP module?


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectTracer.java
```diff
@@ -42,4 +46,24 @@ private String spanName(WithSpan annotation, Method method) {
...
+   * Denotes the end of the invocation of the traced method with a successful result which will end
+   * the span stored in the passed {@code context}. If the method returned a value representing an
+   * asynchronous operation then the span will not be finished until the asynchronous operation has
+   * completed.
+   *
+   * @param returnType Return type of the traced method.
+   * @param returnValue Return value from the traced method.
+   * @return Either {@code returnValue} or a value composing over {@code returnValue} for
+   *     notification of completion.
+   * @throws ClassCastException if returnValue is not an instance of returnType
```
trask, line 60: iirc i added this line 😬
```suggestion
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/async/package-info.java
```diff
@@ -2,4 +2,4 @@
  * Provides implementations of strategies for tracing methods that return asynchronous and reactive
  * values so that the span can be ended when the asynchronous operation completes.
  */
-package io.opentelemetry.javaagent.instrumentation.otelannotations.async;
+package io.opentelemetry.instrumentation.api.tracer.async;
```
trask, line 5: maybe `.async` -> `.future` or `.promise`

`.async` feels a bit general

trask (approved):
👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/async/package-info.java
HaloFour: I'd probably gravitate to `reactive` but that's probably as general as `async` is.  Of your suggestions I think `future` probably fits better in the Java ecosystem.

mateuszrzeszutek (approved):
WDYT about renaming `MethodSpanStrategy/-ies` to `AsyncSpanEndStrategy/-ies`? (or something similar that emphasises the asynchronous nature of this code)

trask:
> WDYT about renaming `MethodSpanStrategy/-ies` to `AsyncSpanEndStrategy/-ies`? (or something similar that emphasises the asynchronous nature of this code)

that sounds nice, and if we don't find an AsyncSpanEndStrategy, we can handle the default synchronous case without a strategy, e.g.

```
  public Object end(Context context, Class<?> returnType, Object returnValue) {
    ...
    AsyncSpanEndStrategy asyncSpanEndStrategy = asyncSpanEndStrategies.resolveStrategy(returnType);
    if (asyncSpanEndStrategy != null) {
      return asyncSpanEndStrategy.end(this, context, returnValue);
    }
    end(context);
    return returnValue;
  }
```

trask:
> If the interface is `AsyncSpanEndStrategy` does it make sense for the package name to remain `async` then?

Ya, let's keep it `async` unless anyone else has thoughts.

trask:
@HaloFour looks like it needs a `./gradlew spotlessApply`

trask:
🚢

trask:
> I could open a follow up PR to change it. At least the class is package-private so it doesn't really matter.

👍 (not a problem even if it was public, as none of the instrumentation APIs have been declared stable yet)

PR #2609, opened 2021-03-20
Servlet 5 API, reorganize servlet modules

agoallikmaa:
The following changes are made in this PR:
* `servlet-common` is split up into `servlet-common` and `servlet-javax-common` where the former does not directly use servlet classes, but defines an interface `ServletAccessor<REQUEST, RESPONSE>` which is used to access request and response objects, and the latter applies instrumentations previously found in `servlet-common`.
* `:instrumentation-core:servlet-2.2` is migrated to `:instrumentation:servlet:servlet-common:library`
* Both `servlet-2.2:javaagent` and `servlet-3.0:javaagent` are split into `javaagent` and `library` modules where the tracer moves to library.
* Shared `TypeInstrumentation` classes moved to `servlet-common:javaagent` and they take base package (either `javax.servlet` or `jakarta.servlet`) and advice class name as constructor parameters
* New module `servlet-jakarta` which applies instrumentations to `jakarta.servlet` classes (the ones which are done by `servlet-javax-common` and `servlet-3.0` for `javax.servlet` packages)
* Matrix project gets a subproject called `servlet-jakarta` which is Servlet API 5.0 equivalent of the test war project. Added Jetty 11 test images which use the new war file. Manually tested that Jetty tests that do not require the `jetty-8.0` module pass with those images.

To get a general overview of the structure of `instrumentation/servlet` directory, read the updated readme in there.

Most of the changes to existing servlet modules were made to reduce code duplication, but some code is still duplicated for the following cases:
* Advice classes of `servlet-3.0` and `servlet-jakarta`. As they modify local variables injected into instrumented methods, extracting common methods would require some additional allocations and complexity.
* Span name detection in `Servlet3HttpServerTracer` and `JakartaServletHttpServerTracer`, as it uses too many different Servlet API classes to reasonably cover with generics.

PR #2619, opened 2021-03-23
Add support for vaadin web framework

laurit:
Vaadin relies heavily on javascript which results in completely undescriptive traces as almost every user action results in a javascript rpc call. This pr attempts to improve the situation in following ways
- set server span name to currently displayed route
- if route is not available set server span name to name of request handler that accepted the request
- add spans describing rpc calls


instrumentation/vaadin-14.2/vaadin-testing/src/main/groovy/test/vaadin/AbstractVaadinTest.groovy
```diff
@@ -0,0 +1,141 @@
...
+import org.testcontainers.Testcontainers
+import org.testcontainers.containers.BrowserWebDriverContainer
+import org.testcontainers.containers.output.Slf4jLogConsumer
+import spock.lang.Shared
+
+abstract class AbstractVaadinTest extends AgentInstrumentationSpecification implements HttpServerTestTrait<ConfigurableApplicationContext> {
+  private static final Logger logger = LoggerFactory.getLogger(AbstractVaadinTest)
+
+  @Shared
+  BrowserWebDriverContainer<?> chrome
```
trask, line 31: 👍

settings.gradle
```diff
@@ -250,6 +250,9 @@ include ':instrumentation:struts-2.3:javaagent'
 include ':instrumentation:tomcat-7.0:javaagent'
 include ':instrumentation:twilio-6.6:javaagent'
 include ':instrumentation:undertow-1.4:javaagent'
+include ':instrumentation:vaadin-14.2:javaagent'
+include ':instrumentation:vaadin-14.2:vaadin-14.2-testing'
+include ':instrumentation:vaadin-14.2:vaadin-testing'
```
trask, line 255: is it possible to use `testSets` to be more consistent with how other instrumentations do multi-version testing?

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
```diff
@@ -0,0 +1,163 @@
...
+        // append event type to make span name more descriptive
+        spanName += "/" + eventType;
+      }
+    }
+    return super.startSpan(spanName);
+  }
+
+  @Override
+  protected String getInstrumentationName() {
+    return "io.opentelemetry.javaagent.vaadin";
```
trask, line 144: we changed convention recently (#2433)
```suggestion
    return "io.opentelemetry.javaagent.vaadin-14.2";
```

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinInstrumentationModule.java
```diff
@@ -0,0 +1,330 @@
...
+        context = tracer().startVaadinServiceSpan(vaadinService, method);
+        scope = context.makeCurrent();
+      }
+
+      @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+      public static void onExit(
+          @Advice.Thrown Throwable throwable,
+          @Advice.Local("otelContext") Context context,
+          @Advice.Local("otelScope") Scope scope) {
+        if (scope == null) {
```
iNikem, line 95: `scope` cannot really be null here, can it?

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private static final VaadinTracer TRACER = new VaadinTracer();
+
+  public static VaadinTracer tracer() {
+    return TRACER;
+  }
+
+  private VaadinTracer() {
+    super(GlobalOpenTelemetry.get());
+  }
```
iNikem, line 37: @anuraaga What is our current preferred form of tracer constructors?

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
```diff
@@ -0,0 +1,163 @@
...
+
+    String spanName = spanNameForMethod(requestHandler.getClass(), method);
+    VaadinServiceContext vaadinServiceContext = current.get(SERVICE_CONTEXT_KEY);
+    if (vaadinServiceContext != null && !vaadinServiceContext.isRequestHandled()) {
+      Span span = ServerSpan.fromContextOrNull(current);
+      if (span != null) {
+        // set server span name to request handler name
+        // we don't really know whether this request handler is going to be the one
+        // that process the request, if it isn't then next handler will also update
+        // server span name
```
iNikem, line 84: Do handlers call each other (and thus they are nested) or are they called one after another?

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private static class VaadinServiceContext {
+    final String vaadinServiceSpanName;
+    boolean requestHandled;
+
+    VaadinServiceContext(String vaadinServiceSpanName) {
+      this.vaadinServiceSpanName = vaadinServiceSpanName;
+    }
+
+    void setRequestHandled() {
```
iNikem, line 155: Do we need some thread-safety here?

instrumentation/vaadin-14.2/javaagent/vaadin-14.2-javaagent.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+  testLibrary 'com.vaadin:vaadin-spring-boot-starter:16.0.0'
+
+  testImplementation project(':instrumentation:vaadin-14.2:vaadin-testing')
+  testImplementation(project(':testing-common')) {
+    exclude(module: 'jetty-server')
+  }
+
+  testInstrumentation project(':instrumentation:servlet:servlet-3.0:javaagent')
+  testInstrumentation project(':instrumentation:servlet:servlet-common:javaagent')
```
iNikem, line 37: Please take a look that changes from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2609 don't mess with you

instrumentation/vaadin-14.2/javaagent/vaadin-14.2-javaagent.gradle
```diff
@@ -0,0 +1,39 @@
...
+  testInstrumentation project(':instrumentation:tomcat-7.0:javaagent')
```
iNikem, line 38: Why jetty from `testing-common` is not good enough?

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinInstrumentationModule.java
laurit: Fixed.

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
laurit: As far as I understand both can happen. Main request processing calls handlers sequentially, but there are some handlers that call other handlers. Our instrumentation should only care about handlers in main processing and ignore the rest.

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
laurit: I'd say no, it should all happen on the same thread, though making it a bit more thread safe wouldn't hurt either.

instrumentation/vaadin-14.2/javaagent/vaadin-14.2-javaagent.gradle
laurit: Rebased.

instrumentation/vaadin-14.2/javaagent/vaadin-14.2-javaagent.gradle
laurit: `testing-common` has an ancient jetty version and `org.springframework.boot:spring-boot-starter-web` already depends on tomcat. Excluding jetty from `testing-common` is done in multiple projects, it might be better if `testing-common` didn't depend on jetty.

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinInstrumentationModule.java
```diff
@@ -0,0 +1,329 @@
...
+        context = tracer().startRpcInvocationHandlerSpan(rpcInvocationHandler, method, jsonObject);
+        scope = context.makeCurrent();
+      }
+
+      @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+      public static void onExit(
+          @Advice.Thrown Throwable throwable,
+          @Advice.Local("otelContext") Context context,
+          @Advice.Local("otelScope") Scope scope) {
+        if (scope == null) {
```
iNikem, line 272: scope cannot be null here

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinInstrumentationModule.java
```diff
@@ -0,0 +1,329 @@
...
+        context = tracer().startClientCallableSpan(componentClass, methodName);
+        scope = context.makeCurrent();
+      }
+
+      @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+      public static void onExit(
+          @Advice.Thrown Throwable throwable,
+          @Advice.Local("otelContext") Context context,
+          @Advice.Local("otelScope") Scope scope) {
+        if (scope == null) {
```
iNikem, line 320: same here

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
anuraaga: Ah sorry for late reply. I don't think it matters too much for Javaagent instrumentation so this seems ok.

iNikem (approved):

instrumentation/vaadin-14.2/testing/src/main/groovy/test/vaadin/AbstractVaadin14Test.groovy
```diff
@@ -0,0 +1,82 @@
...
+
+    return handlers
+  }
+
+  @Override
+  void assertFirstRequest() {
+    assertTraces(VAADIN_14_4 ? 5 : 4) {
+      def handlers = getRequestHandlers("BootstrapHandler")
+      trace(0, 3 + handlers.size()) {
+        basicSpan(it, 0, getContextPath() + "/main", null)
```
mateuszrzeszutek, line 35: I've just realized that `basicSpan()` does not check span kind at all...
`span(0)` should always be a SERVER span, right? Can you add a helper method in the base test class that does that? (I assume changing `basicSpan()` should be left for another PR)

instrumentation/vaadin-14.2/testing/src/main/groovy/test/vaadin/AbstractVaadin14Test.groovy
```diff
@@ -0,0 +1,82 @@
...
+      }
+      trace(1, 2) {
+        basicSpan(it, 0, getContextPath() + "/*", null)
+        basicSpan(it, 1, "ApplicationDispatcher.forward", span(0))
+      }
+      trace(2, 2) {
+        basicSpan(it, 0, getContextPath() + "/*", null)
+        basicSpan(it, 1, "ApplicationDispatcher.forward", span(0))
+      }
+      trace(3, 2) {
```
mateuszrzeszutek, line 52: Are those other traces significant at all? If it's possible, can you leave a comment explaining what they're representing? (ex. `/favicon.ico`)

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinTracer.java
```diff
@@ -0,0 +1,164 @@
...
+      RpcInvocationHandler rpcInvocationHandler, Method method, JsonObject jsonObject) {
+    String spanName = spanNameForMethod(rpcInvocationHandler.getClass(), method);
+    if ("event".equals(rpcInvocationHandler.getRpcType())) {
+      String eventType = jsonObject.getString("event");
+      if (eventType != null) {
+        // append event type to make span name more descriptive
+        spanName += "/" + eventType;
+      }
+    }
+    return super.startSpan(spanName);
```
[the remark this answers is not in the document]
laurit: It is on the receiving end of the rpc.

instrumentation/vaadin-14.2/testing/src/main/groovy/test/vaadin/AbstractVaadinTest.groovy
```diff
@@ -0,0 +1,141 @@
...
+  abstract void assertFirstRequest()
+
+  abstract void assertButtonClick()
+
+  def "test vaadin"() {
+    setup:
+    def driver = getWebDriver()
+    waitForStart(driver)
+
+    driver.manage().timeouts().implicitlyWait(30, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
laurit: Moved. If by line below you mean `driver.get(address.resolve("main").toString())` then it is needed, `waitForStart` fetches the same page and after it is loaded clears traces, so need to fetch it again. `waitForStart` clears traces because there is some background activity during the first page load that creates an unpredictable amount of traces.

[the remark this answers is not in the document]
mateuszrzeszutek: Can you add a comment explaining why it's there?

instrumentation/vaadin-14.2/testing/src/main/groovy/test/vaadin/AbstractVaadin14Test.groovy
laurit: These aren't significant, they are for javascript files. Comment added.

instrumentation/vaadin-14.2/testing/src/main/groovy/test/vaadin/AbstractVaadin14Test.groovy
laurit: Added server span assertion.

instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/VaadinInstrumentationModule.java
laurit: Null check removed.

settings.gradle
laurit: Done.

mateuszrzeszutek (approved):

PR #2622, opened 2021-03-23
Run smoketests in parallel

laurit:


smoke-tests/smoke-tests.gradle
```diff
@@ -38,6 +38,25 @@ test {
   //In addition to that we disable them by default on local machines
   enabled = enabled && (System.getenv("CI") != null || findProperty('runSmokeTests'))
 
+  def suites = [
+    "glassfishAndJetty": ["**/GlassFishSmokeTest.*", "**/JettySmokeTest.*"],
```
iNikem, line 42: Sometimes the simplest solution is the best :D Why these buckets specifically?

iNikem (approved):

smoke-tests/smoke-tests.gradle
laurit: No particular reason. It shouldn't really matter which way these are grouped as long as suite run time is lower than longest build step.

trask (approved):

PR #2629, opened 2021-03-24
Reimplement finding open ports

laurit:
This pr implements a new strategy for finding open ports that hopefully reduces the chance of getting `java.net.BindException: Address already in use`. Ports are allocated from 11000 to 32768. 11000 is chosen because most services should use a lower port and 32768 is chosen because it should be below ephemeral port range on most operating systems. Allocator divides port range to chunks of 100 ports and attempts to bind the first port of the chunk. If binding succeeds allocator claims this chunk and returns ports from it sequentially. If binding fails allocator assumes that another allocator that is running in some other process has claimed this chunk and attempts to claim next chunk. It is also possible to find consecutive open ports, this can come handy when the server that we test binds multiple ports where the second port can't be directly set but is computed from the first port e.g. in `RocketMqClientTest` server also bind to port with offset of -2 from the port that we set.

### DIFF
--- a/instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
+++ b/instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
@@ -5,6 +5,7 @@
 
 package base;
 
+import io.opentelemetry.instrumentation.test.utils.PortUtils;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -12,7 +13,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Random;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.rocketmq.broker.BrokerController;
 import org.apache.rocketmq.common.BrokerConfig;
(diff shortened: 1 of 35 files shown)


iNikem (approved):

instrumentation/rocketmq-client-4.8/testing/src/main/java/base/IntegrationTestBase.java
```diff
@@ -77,7 +70,9 @@ public static NamesrvController createAndStartNamesrv() {
     namesrvConfig.setKvConfigPath(kvConfigPath.toString());
     namesrvConfig.setConfigStorePath(namesrvPath.toString());
 
-    nameServerNettyServerConfig.setListenPort(nextPort());
+    // find 3 consecutive open ports and use the last one of them
+    // rocketmq will also bind to give port - 2
```
trask, line 74:
```suggestion
    // rocketmq will also bind to given port - 2
```

testing-common/src/main/java/io/opentelemetry/instrumentation/test/utils/PortUtils.java
```diff
@@ -6,26 +6,23 @@
...
-      socket.setReuseAddress(true);
-      socket.close();
-      return socket.getLocalPort();
-    } catch (IOException ioe) {
-      ioe.printStackTrace();
-      return -1;
-    }
+  private static final PortAllocator portAllocator = new PortAllocator();
+
+  /** Find consecutive open ports. */
```
trask, line 18:
```suggestion
  /** Find consecutive open ports, returning the first one in the range. */
```

testing-common/src/main/java/io/opentelemetry/instrumentation/test/utils/PortAllocator.java
```diff
@@ -0,0 +1,129 @@
...
+                  }
+                }));
+  }
+
+  /** Find open port. */
+  int getPort() {
+    return getPorts(1);
+  }
+
+  /** Find consecutive range of open ports. */
```
trask, line 58:
```suggestion
  /** Find consecutive range of open ports, returning the first one in the range. */
```

trask (approved):
🤞

PR #2632, opened 2021-03-25
Use direct caffeine executor.

anuraaga:
I don't know what the actual problem is with the ForkJoinPool default executor but while I easily had crashes with it, after changing to direct execution, the crashes seem to have gone away. So just 70% sure this PR helps :P

We only have bounds, no async work, so direct execution of cleanup shouldn't really decrease performance much. (/cc @ben-manes JFYI since you had suggested this in the first place). In the future we could consider our `CommonTaskExecutor`, which presumably doesn't have the problem (not sure the essential difference vs ForkJoinPool would be though) if it seems like it would help.

### DIFF
--- a/instrumentation-api-caching/src/main/java/io/opentelemetry/instrumentation/api/caching/CacheBuilder.java
+++ b/instrumentation-api-caching/src/main/java/io/opentelemetry/instrumentation/api/caching/CacheBuilder.java
@@ -52,6 +52,8 @@ public final class CacheBuilder {
     }
     if (executor != null) {
       caffeine.executor(executor);
+    } else {
+      caffeine.executor(Runnable::run);
     }
     @SuppressWarnings("unchecked")
     com.github.benmanes.caffeine.cache.Cache<K, V> delegate =


anuraaga:
Somewhat coincidentally (I thought I'd look into why RunnableWrapper seems to be applied twice to webflux tasks), I noticed this sort of stack trace, I guess the advantage of `CommonTaskExecutor` (or any other executor) is we don't apply instrumentation to it

```
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:1379)
	at io.opentelemetry.javaagent.instrumentation.api.concurrent.RunnableWrapper.wrapIfNeeded(RunnableWrapper.java:32)
	at java.base/java.util.concurrent.ForkJoinPool.execute(ForkJoinPool.java:2455)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.internal.shaded.caffeine.cache.BoundedLocalCache.scheduleDrainBuffers(BoundedLocalCache.java:1425)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.internal.shaded.caffeine.cache.BoundedLocalCache.afterRead(BoundedLocalCache.java:1161)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.internal.shaded.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2348)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.internal.shaded.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:108)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.internal.shaded.caffeine.cache.LocalManualCache.get(LocalManualCache.java:62)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.caching.CaffeineCache.computeIfAbsent(CaffeineCache.java:21)
	at io.opentelemetry.javaagent.tooling.bytebuddy.matcher.ClassLoaderMatcher$ClassLoaderHasClassesNamedMatcher.matches(ClassLoaderMatcher.java:68)
	at io.opentelemetry.javaagent.tooling.bytebuddy.matcher.ClassLoaderMatcher$ClassLoaderHasClassesNamedMatcher.matches(ClassLoaderMatcher.java:33)
```

iNikem (approved):

laurit:
Imo the problem is that classes loaded during transformation are not instrumented. Apparently caffeine code called from transform manages to pull in `RunnableWrapper` and it isn't transformed :(
```
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -16,6 +16,8 @@ import static net.bytebuddy.matcher.ElementMatchers.none;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.javaagent.bootstrap.AgentClassLoader;
 import io.opentelemetry.javaagent.instrumentation.api.SafeServiceLoader;
+import io.opentelemetry.javaagent.instrumentation.api.concurrent.CallableWrapper;
+import io.opentelemetry.javaagent.instrumentation.api.concurrent.RunnableWrapper;
 import io.opentelemetry.javaagent.instrumentation.api.internal.BootstrapPackagePrefixesHolder;
 import io.opentelemetry.javaagent.spi.BootstrapPackagesProvider;
 import io.opentelemetry.javaagent.spi.ByteBuddyAgentCustomizer;
@@ -82,6 +84,8 @@ public class AgentInstaller {
     BootstrapPackagePrefixesHolder.setBoostrapPackagePrefixes(loadBootstrapPackagePrefixes());
     // this needs to be done as early as possible - before the first Config.get() call
     ConfigInitializer.initialize();
+    RunnableWrapper.class.getName();
+    CallableWrapper.class.getName();
   }
```
seems to also make the failing vertx test pass on jdk15.

trask (approved):

anuraaga:
@laurit Discussed with @trask and we think that the inline cleanup should be fine for this anyways - a thread's going to have to do this work anyways and batching them doesn't make a real difference.

For the callback wrappers, there does seem to be some fragility in that if they get loaded in wrong places it can break things, but we think a better solution is for any wrappers we provide to be implemented in code instead of relying on instrumentation, since they're our code in the first place :)

PR #2634, opened 2021-03-25
Add gRPC context bridge

anuraaga:
I was originally thinking of mapping otel context to grpc context for library instrumentation and opposite for agent. I decided against it because

- When using multiple libraries, e.g. Armeria and gRPC, it forces user to think about the canonical backing store. This is not often an issue but could cause some potential issues.
- It fits with the philosophy of the agent backing off if library instrumentation is found. Not a huge deal also since that doesn't have to apply to context propagation.

It means that we don't get to actually use the nice customizability of the OTel context storage - oh well at least we got a taste of what it can do in the nice and efficient agent/app context bridge.

It means that the implementation is extremely obtuse and fairly high overhead though - gRPC storage override only provides void methods and unextendable Context class so we have to side-channel state information. The least they could have done is allow `attach` to return something :(

The implementation method is basically the same as our previous experimentation with `otelInGrpc` - https://github.com/open-telemetry/opentelemetry-java/blob/main/context/src/otelInGrpcTest/java/io/opentelemetry/context/GrpcContextStorageProvider.java

The same bits are attached to each other, just using context values instead of wrappers and threadlocal-weakmap-of-queues-of-scopes instead of a return value from the attach method.


instrumentation/grpc-1.5/javaagent/grpc-1.5-javaagent.gradle
```diff
@@ -36,4 +36,10 @@ dependencies {
...
 
   testImplementation project(':instrumentation:grpc-1.5:testing')
-}
\ No newline at end of file
+}
+
+test {
+  // The agent context debug mechanism isn't compatible with the bridge approach which may add a
+  // gRPC context to the root.
+  jvmArgs "-Dotel.javaagent.experimental.thread-propagation-debugger.enabled=false"
```
anuraaga, line 44: Took me a day to realize this :)

instrumentation/grpc-1.5/library/src/main/java/io/grpc/override/ContextStorageOverride.java
```diff
@@ -0,0 +1,122 @@
...
+
+  // Because the extension point is void, there is no way to return information about the backing
+  // OpenTelemetry context when attaching gRPC context. So the only option is to have this
+  // side-channel to keep track of scopes. Because the same context can be attached to multiple
+  // threads, we must use a ThreadLocal here - on the bright side it means the map doesn't have to
+  // be concurrent. This will add an additional threadlocal lookup when attaching / detaching gRPC
+  // context, but not when accessing the current. In many applications, this means a small
+  // difference
+  // since those operations are rare, but in highly reactive applications where the overhead of
+  // ThreadLocal was already a problem, this makes it worse.
```
anuraaga, line 39: On the bright side, users do have a way forward - because the bridge is bi-directional, they can swap out gRPC propagation bit by bit without affecting their application. The more OTel propagation they use, the less the overhead of this bridge will be.

mateuszrzeszutek (approved):

trask (commented):
sorry for the delay, i'll finish reviewing tomorrow

instrumentation/grpc-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_5/GrpcInstrumentationModule.java
```diff
@@ -18,9 +18,16 @@ public GrpcInstrumentationModule() {
     super("grpc", "grpc-1.5");
   }
 
+  @Override
+  protected String[] additionalHelperClassNames() {
+    return new String[] {"io.grpc.override.ContextStorageOverride"};
```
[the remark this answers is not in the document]
mateuszrzeszutek: Hmm, that would disable the entire gRPC instrumentation... still, it might be necessary 🤷‍♂️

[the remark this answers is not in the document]
anuraaga: Hmm - I guess it means I shouldn't inject this and just wrap the library's context storage by instrumenting `LazyStorage.get()`. Should I do that in this PR or a followup?

instrumentation/grpc-1.5/library/src/main/java/io/grpc/override/ContextStorageOverride.java
```diff
@@ -0,0 +1,122 @@
...
+      // gRPC context which has an OTel context associated with it via a call to
+      // ContextStorageOverride.current(). Using it as the base allows it to be propagated together
+      // with the gRPC context.
+      newOtelContext = base.with(GRPC_CONTEXT, toAttach);
+    } else {
+      // gRPC context without an OTel context associated with it. This is only possible when
+      // attaching a context directly created by Context.ROOT, e.g., Context.ROOT.with(...) which
+      // is not common. We go ahead and assume the OTel context should also be reset to root in this
+      // case.
+      newOtelContext = io.opentelemetry.context.Context.root().with(GRPC_CONTEXT, toAttach);
```
trask, line 68: what do you think of erroring on the side of not breaking otel context in this case and using otel `Context.current()` here? e.g. I'm guessing one use case for grpc Context.ROOT.attach() is at request entry, as a safety for not leaking context, similar to what we do with otel context

trask (approved):
> I guess it means I shouldn't inject this and just wrap the library's context storage by instrumenting LazyStorage.get(). Should I do that in this PR or a followup?

👍 author's choice 😄

instrumentation/grpc-1.5/library/src/main/java/io/grpc/override/ContextStorageOverride.java
anuraaga: Yeah - that's why I thought we should reset the otel context too, it seemed like you wouldn't ever reset one and not the other, similar to how we always propagate them together.

trask: but in that case, it will break our otel propagation, b/c the agent attaches the SERVER span to the otel context even earlier in the request entry

anuraaga: I guess that makes sense - and ideally we just don't leak context anyways so it doesn't matter too much :D

instrumentation/grpc-1.5/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_5/internal/ContextStorageBridge.java
```diff
@@ -0,0 +1,122 @@
...
+      // gRPC context which has an OTel context associated with it via a call to
+      // ContextStorageOverride.current(). Using it as the base allows it to be propagated together
+      // with the gRPC context.
+      newOtelContext = base.with(GRPC_CONTEXT, toAttach);
+    } else {
+      // gRPC context without an OTel context associated with it. This is only possible when
+      // attaching a context directly created by Context.ROOT, e.g., Context.ROOT.with(...) which
+      // is not common. We go ahead and assume the OTel context should also be reset to root in this
+      // case.
+      newOtelContext = io.opentelemetry.context.Context.root().with(GRPC_CONTEXT, toAttach);
```
trask, line 68: I think `io.opentelemetry.context.Context.current()` is probably a better choice for auto-instrumentation at least, in case the application is attaching gRPC root context at the beginning of request (but after we attach otel context)

anuraaga: Oops, forgot to apply this

PR #2637, opened 2021-03-26
Enable strict context check and fix some context issues.

anuraaga:


instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
```diff
@@ -168,6 +164,10 @@ public void onExecutionFailure(
   }
 
   private void clearAttributes(ExecutionAttributes executionAttributes) {
+    Scope scope = executionAttributes.getAttribute(SCOPE_ATTRIBUTE);
```
anuraaga, line 167: 😱

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
```diff
@@ -27,20 +27,18 @@ public void consumeMessageBefore(ConsumeMessageContext context) {
     if (context == null || context.getMsgList() == null || context.getMsgList().isEmpty()) {
       return;
     }
-    Context traceContext = tracer.startSpan(Context.current(), context.getMsgList());
-    ContextAndScope contextAndScope = new ContextAndScope(traceContext, traceContext.makeCurrent());
-    context.setMqTraceContext(contextAndScope);
+    Context otelContext = tracer.startSpan(Context.current(), context.getMsgList());
```
anuraaga, line 30: Like similar netty libraries, there's no thread guarantee for before and end. Since there's no downstream instrumentation it's not a big deal anyways

instrumentation/akka-http-10.0/javaagent/akka-http-10.0-javaagent.gradle
```diff
@@ -62,3 +62,7 @@ compileVersion101TestGroovy {
   classpath = classpath.plus(files(compileVersion101TestScala.destinationDir))
   dependsOn compileVersion101TestScala
 }
+
+tasks.withType(Test) {
+  jvmArgs "-Dio.opentelemetry.javaagent.shaded.io.opentelemetry.context.enableStrictContext=false"
```
trask: nice to see where this fails

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
trask: 💯

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
```diff
@@ -28,22 +65,40 @@ public static boolean shouldAttachStateToTask(Object task) {
...
+    }
+
+    // Long running task which decorates other tasks. We should probably be using ExecutionContext
+    // for propagation in Scala but in the meantime this is not needed along with other concurrent
+    // instrumentation and causes context leaks.
+    if (taskClass.getName().equals("akka.util.SerializedSuspendableExecutionContext")) {
+      return false;
+    }
+
+    // Wrapper of tasks for dispatch - the wrapped task should have context already and this doesn't
```
anuraaga, line 87: No idea what I'm talking about! But seems like a reasonable hypothesis for now, how akka flakes.

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
```diff
@@ -14,6 +14,43 @@
 /** Utils for concurrent instrumentations. */
 public class ExecutorInstrumentationUtils {
 
+  private static final ClassValue<Boolean> NOT_INSTRUMENTED_RUNNABLE_ENCLOSING_CLASS =
```
trask, line 17: nice, i'm trying to remember to look out for opportunities to use also 😄

trask (approved):

anuraaga:
Sorry for the churn, couldn't help but fix some of the leaks. The remaining will take some deeper dives so I think this is it for now

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
iNikem: I don't get it. Both why this change is significant and your reaction to it :D

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
iNikem: What about e.g. user-provided message hooks that want to augment existing span?

iNikem (approved):

instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/TracingExecutionInterceptor.java
anuraaga: It's a real context leak with big damage - if sync AWS SDK request failed and a subsequent client request of some form was made, it would parent to the AWS SDK request (well it'd be suppressed as a result usually)! Hooray for strict context checking

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
anuraaga: We will need to provide something like `RocketMqTracing.getOpenTelemetryContext(ConsumeMessageContext)` to let them extract it I guess. In the meantime, if someone had such code using `Context.current()` in `consumeMessageAfter` it would generally be buggy with the chance of `Context.current()` not actually corresponding to the hook's message due to context leakage so this fix is important first.

I hope such hooks will be able to use our instrumentation-api instead of that though.

anuraaga:
Unfortunately not able to repro CI failures well but it's nice that it finds them. As far as I can tell there might be something fundamentally wrong with our netty instrumentation, will examine separately first.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/InstrumentationSpecification.groovy
```diff
@@ -36,6 +37,13 @@ abstract class InstrumentationSpecification extends Specification {
     testRunner().clearAllExportedData()
   }
 
+  def cleanup() {
```
mateuszrzeszutek, line 40: Hmm, perhaps we should consider adding an `afterTest()` method to `InstrumentationTestRunner`and calling it here and in `InstrumentationExtension`?

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
```diff
@@ -27,20 +27,18 @@ public void consumeMessageBefore(ConsumeMessageContext context) {
     if (context == null || context.getMsgList() == null || context.getMsgList().isEmpty()) {
       return;
     }
-    Context traceContext = tracer.startSpan(Context.current(), context.getMsgList());
-    ContextAndScope contextAndScope = new ContextAndScope(traceContext, traceContext.makeCurrent());
```
mateuszrzeszutek, line 31: Can `ContextAndScope` be removed now?

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
```diff
@@ -14,6 +14,43 @@
...
+      new ClassValue<Boolean>() {
+        @Override
+        protected Boolean computeValue(Class<?> enclosingClass) {
+          // Avoid context leak on jetty. Runnable submitted from SelectChannelEndPoint is used to
+          // process a new request which should not have context from them current request.
+          if (enclosingClass.getName().equals("org.eclipse.jetty.io.nio.SelectChannelEndPoint")) {
```
mateuszrzeszutek, line 23: Just an idea, but maybe those class names should be specified by instrumentation modules? Each instrumented library may bring its own set of excluded classes (also see `AdditionalLibraryIgnoresMatcher`) and maybe it'd better to define them in the lib instrumentation.

mateuszrzeszutek (approved):

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
anuraaga: Yeah! That's the approach Datadog has implemented, it would be nice for us to follow it at some point.

anuraaga: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/ExcludeFilterProvider.java

instrumentation/ratpack-1.4/javaagent/ratpack-1.4-javaagent.gradle
```diff
@@ -20,3 +20,7 @@ dependencies {
     testImplementation group: 'com.sun.activation', name: 'jakarta.activation', version: '1.2.2'
   }
 }
+
+tasks.withType(Test) {
+  jvmArgs "-Dio.opentelemetry.javaagent.shaded.io.opentelemetry.context.enableStrictContext=false"
```
jkwatson, line 25: does this mean that our ratpack instrumentation actually leaks context, or is it an artifact of something else? If we are really leaking contexts, should we create issues to fix all of these, so we make sure to circle back and fix them?

anuraaga: Good point, filed an issue

instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
anuraaga: Yup, thanks

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/InstrumentationSpecification.groovy
anuraaga: Thanks for pointing out InstrumentationExtension - adding afterTest seemed a bit weird since it's the same code for agent and library. For now I just copied the code since it isn't so much and I think we can revisit later.

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
mateuszrzeszutek: Nice - I originally thought about adding some `void excludeClasses(ExcludeBuilder)` method to `InstrumentationModule`

PR #2641, opened 2021-03-26
Finish Netty 4.1 spans after response has completed not when it started.

anuraaga:
Currently, we end an outbound span as soon as headers are sent, even if the data is chunked and it will take some time to send it all. We should end the span after the last (potentially only but still should wait for the I/O) write is finished instead.

Guessing this applies to other Netty instrumentations too but just started with 4.1 to confirm the approach.

### DIFF
--- a/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
+++ b/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
@@ -9,13 +9,19 @@ import static io.opentelemetry.javaagent.instrumentation.netty.v4_1.client.Netty
 
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
+import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpResponse;
+import io.netty.handler.codec.http.LastHttpContent;
 import io.netty.util.Attribute;
+import io.netty.util.AttributeKey;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.instrumentation.netty.v4_1.AttributeKeys;
 
 public class HttpClientResponseTracingHandler extends ChannelInboundHandlerAdapter {
+
+  private static final AttributeKey<HttpResponse> HTTP_RESPONSE =
+      AttributeKey.valueOf(HttpClientResponseTracingHandler.class, "HTTP_RESPONSE");
 
   @Override
   public void channelRead(ChannelHandlerContext ctx, Object msg) {
@@ -25,8 +31,15 @@ public class HttpClientResponseTracingHandler extends ChannelInboundHandlerAdapt
       return;
     }
 
-    if (msg instanceof HttpResponse) {
+    if (msg instanceof FullHttpResponse) {
       tracer().end(context, (HttpResponse) msg);
+    } else if (msg instanceof HttpResponse) {
+      // Headers before body has been sent, store them to use when finishing the span.
+      ctx.channel().attr(HTTP_RESPONSE).set((HttpResponse) msg);
+    } else if (msg instanceof LastHttpContent) {
+      // Not a FullHttpResponse so this is content that has been sent after headers. Finish the
+      // span using what we stored in attrs.
+      tracer().end(context, ctx.channel().attr(HTTP_RESPONSE).get());
     }
 
     // We want the callback in the scope of the parent, not the client span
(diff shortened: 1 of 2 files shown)


anuraaga:
@iNikem your favorite instrumentation!

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
```diff
@@ -25,8 +31,15 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) {
...
-    if (msg instanceof HttpResponse) {
+    if (msg instanceof FullHttpResponse) {
       tracer().end(context, (HttpResponse) msg);
+    } else if (msg instanceof HttpResponse) {
+      // Headers before body has been sent, store them to use when finishing the span.
+      ctx.channel().attr(HTTP_RESPONSE).set((HttpResponse) msg);
+    } else if (msg instanceof LastHttpContent) {
+      // Not a FullHttpResponse so this is content that has been sent after headers. Finish the
+      // span using what we stored in attrs.
+      tracer().end(context, ctx.channel().attr(HTTP_RESPONSE).get());
```
iNikem, line 42: I find it confusing that you talk about "sending" here. This is _client_ response handler, it receives data, not sends.

iNikem (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
anuraaga: Yup thanks

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
```diff
@@ -9,14 +9,20 @@
...
 import io.netty.util.Attribute;
+import io.netty.util.AttributeKey;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.instrumentation.netty.v4_1.AttributeKeys;
 
 public class HttpClientResponseTracingHandler extends ChannelInboundHandlerAdapter {
 
+  private static final AttributeKey<HttpResponse> HTTP_RESPONSE =
+      AttributeKey.valueOf(HttpClientResponseTracingHandler.class, "HTTP_RESPONSE");
```
trask, line 24: just to match existing convention
```suggestion
      AttributeKey.valueOf(HttpClientResponseTracingHandler.class, "http-response");
```

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/server/HttpServerResponseTracingHandler.java
```diff
@@ -7,29 +7,73 @@
...
 import io.netty.handler.codec.http.HttpResponse;
+import io.netty.handler.codec.http.LastHttpContent;
+import io.netty.util.AttributeKey;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 
 public class HttpServerResponseTracingHandler extends ChannelOutboundHandlerAdapter {
 
+  private static final AttributeKey<HttpResponse> HTTP_RESPONSE =
+      AttributeKey.valueOf(HttpServerResponseTracingHandler.class, "HTTP_RESPONSE");
```
trask, line 24:
```suggestion
      AttributeKey.valueOf(HttpServerResponseTracingHandler.class, "http-response");
```

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/server/HttpServerResponseTracingHandler.java
```diff
@@ -7,29 +7,73 @@
...
-    if (context == null || !(msg instanceof HttpResponse)) {
+    if (context == null) {
       ctx.write(msg, prm);
       return;
     }
 
+    final ChannelPromise writePromise;
+
+    if (msg instanceof LastHttpContent) {
+      if (prm == ctx.voidPromise()) {
```
trask, line 37:
```suggestion
      if (prm.isVoid()) {
```

trask (approved):
👍

PR #2642, opened 2021-03-26
Close Response in HttpServerTest

anuraaga:


mateuszrzeszutek (approved):

iNikem (approved):
Have you verified that failed asserts are correctly displayed and they fail the build?

trask (approved):

instrumentation/grails-3.0/javaagent/src/test/groovy/test/ErrorController.groovy
```diff
@@ -5,14 +5,16 @@
...
+
 import grails.artefact.Controller
 import grails.web.Action
 
 class ErrorController implements Controller {
 
   @Action
   def index() {
-    render "Error"
+    render ERROR.body
```
anuraaga, line 17: I have no idea why closing the okhttp response seems to have correctly caused GrailsTest to fail - it is set up to map 500 through this controller, which is also why `errorPageSpans` is set in the test. I can't think of any reason for the number of spans to be correct but the body assertion to be wrong. /cc @laurit

anuraaga:
> Have you verified that failed asserts are correctly displayed and they fail the build?

Yup me and CI :)

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -254,9 +262,12 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
     def response = client.newCall(request).execute()
 
     expect:
-    response.code() == REDIRECT.status
-    response.header("location") == REDIRECT.body ||
-      response.header("location") == "${address.resolve(REDIRECT.body)}"
+    response.withCloseable {
+      assert response.code() == REDIRECT.status
```
anuraaga, line 266: Digging more to try to understand what's going on, I found that most of our tests were actually closing, just not these header-only ones

https://github.com/square/okhttp/blob/parent-3.14.9/okhttp/src/main/java/okhttp3/ResponseBody.java#L48

That being said it's cleaner code to always use closeable without relying on this magic close-by-`string()` invocation.

But given that, I can only hypothesis that not closing in redirect test had an adverse affect on the error test. Even so, I think it's only supposed to be about whether a pooled buffer is returned to the buffer pool, and even if there happened to be buffer corruption, I'd expect it to have the word `/redirected` and not `controller error`. So still not exactly sure why tests would have started to correctly fail after this change but won't let Murphy affect my Saturday much more :)

instrumentation/tomcat-7.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatHandlerTest.groovy
```diff
@@ -56,6 +56,11 @@ class TomcatHandlerTest extends HttpServerTest<Tomcat> implements AgentTestTrait
     endpoint == REDIRECT || endpoint == ERROR
   }
 
+  @Override
+  boolean testErrorBody() {
```
anuraaga, line 60: @laurit I tried both

```
resp.sendError(serverEndpoint.getStatus(), serverEndpoint.getBody())
resp.setStatus(serverEndpoint.getStatus()
```

but they didn't work. If you think this loosening of the test case can be fixed, please address in a followup :)

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
laurit: Smoke tests had the same issue, there I added
```
    cleanup:
    response?.close()
```

I believe that in this case we were just bitten by magic of spock. Considering that in
```
  def "test a"() {
    expect:
    1 == 1
    1 == 0
  }

  def "test b"() {
    expect:
    1 == 1
    if (1 == 1) {
      1 == 0
    }
  }
```
a fails but b passes I suppose you just enabled an assertion that was previously ignored :)

instrumentation/tomcat-7.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/tomcat7/TomcatHandlerTest.groovy
laurit: Instead of `assert response.body().string() == ERROR.body` we could use `assert response.body().string().contains(ERROR.body)` then it would pass with `resp.sendError(serverEndpoint.getStatus(), serverEndpoint.getBody())`. Alternatives would be to figure out how to influence tomcat error page to print just the expected content or remove this assertion completely.

trask: > or remove this assertion completely
👍

trask: on second thought, maybe the existing check does provide some value...

laurit: `TomcatServlet3Test` replaces tomcat `ErrorReportValve` copy-pasting that should help

PR #2646, opened 2021-03-28
Fix webflux client filter subscribe

trask:
I should not be allowed to write reactive code 😞


anuraaga (commented):
Is there a test case we can add?

iNikem (approved):

anuraaga (approved):

mateuszrzeszutek (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -137,6 +138,31 @@ abstract class HttpClientTest extends InstrumentationSpecification {
     method << BODY_METHODS
   }
 
+  def "basic #method request with CLIENT parent"() {
```
iNikem, line 141: Should test name mention that we test nested client spans suppression here?

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -137,6 +138,31 @@ abstract class HttpClientTest extends InstrumentationSpecification {
...
+    then:
+    status == 200
+    assertTraces(2) {
+      trace(0, 1) {
+        basicSpan(it, 0, "parent-client-span")
+      }
+      trace(1, 1) {
+        serverSpan(it, 0)
+      }
+    }
```
iNikem, line 159: I think this deserves more comments. This claims that trace propagation should NOT happen in this test, which is very different from all other tests and thus will be confusing for readers.

PR #2652, opened 2021-03-29
Add support for GWT

laurit:
Add span around java method called from javascript.


instrumentation/gwt-2.0/javaagent/gwt-2.0-javaagent.gradle
```diff
@@ -0,0 +1,93 @@
...
+  testImplementation "org.testcontainers:selenium:${versions.testcontainers}"
+  testImplementation 'org.seleniumhq.selenium:selenium-java:3.141.59'
+
+  testImplementation(project(':testing-common')) {
+    exclude group: 'org.eclipse.jetty', module: 'jetty-server'
+  }
+  testImplementation group: 'org.eclipse.jetty', name: 'jetty-webapp', version: '9.4.35.v20201120'
+}
+
+task copyTestWebapp(type: Copy) {
```
iNikem, line 48: Wouldn't it be easier to have a separate module for test app?

iNikem (approved):
Anything we can do for SERVER span name?

instrumentation/gwt-2.0/javaagent/gwt-2.0-javaagent.gradle
laurit: Assuming that we want to compile the test app with the same version of gwt that we use to run the test even if it is doable having the test app in another module would probably make this even more confusing.

laurit: As for the server span name suggestions are welcome. Currently it is the servlet mapping where the service is mapped as far as I can tell each service should be mapped to a different path so it should be a rather good name. Alternative would be to set server span to invoked method name, but then server span and nested span would have the same name.

instrumentation/gwt-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/gwt/GwtTracer.java
```diff
@@ -0,0 +1,56 @@
...
+package io.opentelemetry.javaagent.instrumentation.gwt;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import java.lang.reflect.Method;
+
+public class GwtTracer extends BaseTracer {
+  private static final ContextKey<Object> PRC_CONTEXT_KEY =
```
mateuszrzeszutek, line 15: Typo:
```suggestion
  private static final ContextKey<Object> RPC_CONTEXT_KEY =
```

instrumentation/gwt-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/gwt/GwtInstrumentationModule.java
```diff
@@ -0,0 +1,113 @@
...
+    public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
+      Map<ElementMatcher<MethodDescription>, String> transformers = new HashMap<>();
+
+      transformers.put(
+          named("invokeAndEncodeResponse")
+              .and(takesArguments(5))
+              .and(takesArgument(0, named(Object.class.getName())))
+              .and(takesArgument(1, named(Method.class.getName())))
+              .and(takesArgument(3, named("com.google.gwt.user.server.rpc.SerializationPolicy")))
+              .and(takesArgument(4, named(int.class.getName()))),
```
mateuszrzeszutek, line 64: There's no need for `named()` matcher when referencing standard Java types:
```suggestion
              .and(takesArgument(0, Object.class))
              .and(takesArgument(1, Method.class))
              .and(takesArgument(3, named("com.google.gwt.user.server.rpc.SerializationPolicy")))
              .and(takesArgument(4, int.class)),
```

mateuszrzeszutek (approved):

instrumentation/gwt-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/gwt/GwtInstrumentationModule.java
laurit: Thanks, didn't know that. Should be fixed now.

mateuszrzeszutek (approved):

PR #2653, opened 2021-03-29
Start smoke test backend and collector once

laurit:
and reuse them to run all tests.

### DIFF
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
@@ -26,7 +26,7 @@ abstract class SmokeTest extends Specification {
   protected static final OkHttpClient CLIENT = OkHttpUtils.client()
 
   protected static final TestContainerManager containerManager = createContainerManager()
-  private static boolean containerManagerStarted = false;
+  private static boolean containerManagerStarted = false
 
   @Shared
   private TelemetryRetriever telemetryRetriever
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
@@ -26,7 +26,6 @@ abstract class SmokeTest extends Specification {
   protected static final OkHttpClient CLIENT = OkHttpUtils.client()
 
   protected static final TestContainerManager containerManager = createContainerManager()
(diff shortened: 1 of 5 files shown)


mateuszrzeszutek (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -47,7 +48,14 @@ abstract class SmokeTest extends Specification {
   }
 
   def setupSpec() {
-    containerManager.startEnvironment()
+    // TestContainerManager starts backend and collector, we want to start them only once
+    // and reuse for all tests
+    if (!containerManagerStarted) {
+      containerManagerStarted = true
+      containerManager.startEnvironment()
+      Runtime.addShutdownHook { containerManager.stopEnvironment() }
```
iNikem, line 56: Should this be encapsulated inside `ContainerManager`?

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/SmokeTest.groovy
```diff
@@ -79,7 +87,6 @@ abstract class SmokeTest extends Specification {
   }
 
   def cleanupSpec() {
-    containerManager.stopEnvironment()
   }
```
trask, line 90: remove method?

trask (approved):

smoke-tests/src/test/java/io/opentelemetry/smoketest/TestContainerManager.java
```diff
@@ -12,6 +12,8 @@
 public interface TestContainerManager {
   void startEnvironment();
```
mateuszrzeszutek, line 13: This method is no longer used in any test class, right? WDYT about adding the `if (!started)` check to it instead of creating a new `startEnvironmentOnce()` method?

laurit: There are 2 implementations of `TestContainerManager` i didn't want to change both of them. I could remove `startEnvironment` and `stopEnvironment` from `TestContainerManager` interface and add them as abstract to `AbstractTestContainerManager`.

mateuszrzeszutek: If they're not used externally then moving them into the base class would be a great idea.

iNikem (approved):

PR #2654, opened 2021-03-29
Remove unnecessary libs from testSets test classpath

mateuszrzeszutek:
I have no idea how testSets test tasks previously worked - runtime classpath certainly wasn't filtered, so how were jvm args passed? My gradle-fu is too low to understand that 😂 

Anyway, this PR should unblock the "prevent duplicate telemetry" one that I'm planning to open.

### DIFF
--- a/gradle/instrumentation.gradle
+++ b/gradle/instrumentation.gradle
@@ -108,38 +108,41 @@ shadowJar {
 
 evaluationDependsOn(":testing:agent-for-testing")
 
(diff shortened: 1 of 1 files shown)


iNikem (approved):

trask (approved):
heads up for merge conflict with #2658 (cc: @anuraaga)

anuraaga (approved):

PR #2658, opened 2021-03-30
Package instrumentation into agent jar for testing.

anuraaga:
I originally used initializer jar basically to reduce affect on build time. But with gRPC bridge, I realized that it can have a meaningful impact on the correctness of the build - currently instrumentation classpath is packaged without obfuscation so is available on the agent classpath. For the most part, this doesn't cause problems, but when a class like `io.grpc.override.ContextStorageOverride` does, then it affects the entire agent in a way that wouldn't happen in real usage.

Unfortunately, the task does seem to take some time, but not a significant amount I guess in the large scheme of the build.


gradle/instrumentation.gradle
```diff
@@ -58,6 +58,17 @@ afterEvaluate {
 
     testImplementation deps.testcontainers
 
+    // Include instrumentations instrumenting core JDK classes tp ensure interoperability with other instrumentation
+    testInstrumentation project(':instrumentation:executors:javaagent')
```
anuraaga, line 62: Moved here for `mergeServiceFiles()`

trask (approved):
nice 👍

testing/agent-exporter/agent-exporter.gradle
```diff
@@ -14,22 +15,10 @@ dependencies {
   annotationProcessor deps.autoservice
   compileOnly deps.autoservice
 
-  implementation project(':javaagent-tooling')
```
anuraaga, line 17: I had to move this to allow its SPIs to be merged with instrumentations. It made maintaining the agent-for-testing test difficult - I just deleted it since while it was a nice sanity check while getting that testing infrastructure in place, now it doesn't really provide much value.

iNikem:
> currently instrumentation classpath is packaged without obfuscation

You mean packaged into initializer jar? 

I don't like this change in the long run. If we are thinking about alternative ways of distributing 3rd party instrumentations, then one such way is obviously an initializer jar/folder. Thus I think that we have to support this mode of operation: minimal agent with required instrumentations such as classloaders and executors plus additional instrumentations in a separate jar.

anuraaga:
> Thus I think that we have to support this mode of operation: minimal agent with required instrumentations such as classloaders and executors plus additional instrumentations in a separate jar.

I think this requires the agent classloader isolation we've talked about - it sounds nice but I'm not ready to work on that right now. This PR makes the test agent closer to the agent we publish so I guess it's a reasonable reflection of the current state even if we hope to change it later.

iNikem:
I will wait for a green build before reviewing for approval :)

gradle/instrumentation.gradle
anuraaga: Looks like this approach won't be feasible - it really does explode things too much to do this all in each project. With this build, the build cache (which the shadowJar is a part of) gets too large to upload. I think I'll punt on this PR in favor of @iNikem's volunteered work of the properly isolated extensions jar :) So I'll also punt on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2634#discussion_r603104376 until then I think as I can't test behavior with / without an override present since it's always in the bootstrap for our tests right now.

PR #2659, opened 2021-03-30
Cache GWT compilation and only generate one permutation.

anuraaga:


instrumentation/gwt-2.0/javaagent/gwt-2.0-javaagent.gradle
```diff
@@ -45,11 +45,6 @@ dependencies {
   testImplementation group: 'org.eclipse.jetty', name: 'jetty-webapp', version: '9.4.35.v20201120'
 }
 
-task copyTestWebapp(type: Copy) {
```
anuraaga, line 48: @laurit This wasn't needed right?

instrumentation/gwt-2.0/javaagent/src/testapp/java/test/gwt/Greeting.gwt.xml
```diff
@@ -3,6 +3,8 @@
 
   <entry-point class="test.gwt.client.GreetingEntryPoint" />
 
+  <set-property name="user.agent" value="safari" />
```
trask, line 6:
```suggestion
  <!-- only generate one permutation to make the build faster -->
  <set-property name="user.agent" value="safari" />
```

trask (approved):

instrumentation/gwt-2.0/javaagent/src/testapp/java/test/gwt/Greeting.gwt.xml
```diff
@@ -3,6 +3,9 @@
 
   <entry-point class="test.gwt.client.GreetingEntryPoint" />
 
+  <!-- only generate one permutation to make the build faster -->
+  <set-property name="user.agent" value="safari" />
```
iNikem, line 7: Why safari if test uses chrome?

instrumentation/gwt-2.0/javaagent/gwt-2.0-javaagent.gradle
```diff
@@ -59,11 +54,10 @@ task compileGwt(dependsOn: classes, type: JavaExec) {
   def extraDir = "$buildDir/testapp/extra"
   def warDir = "$buildDir/testapp/war"
 
-  outputs.dir warDir
+  outputs.cacheIf { true }
```
iNikem, line 57: Can you explain this?

instrumentation/gwt-2.0/javaagent/src/testapp/java/test/gwt/Greeting.gwt.xml
anuraaga: GWT is from way before Chrome :) Back then it was all Webkit and I suspect it serves safari pages to Chrome.

instrumentation/gwt-2.0/javaagent/gwt-2.0-javaagent.gradle
anuraaga: Yeah it's from this https://docs.gradle.org/current/userguide/build_cache.html#using_the_runtime_api

instrumentation/gwt-2.0/javaagent/src/testapp/java/test/gwt/Greeting.gwt.xml
anuraaga: Possible values are here https://gwt.googlesource.com/gwt/+/master/user/src/com/google/gwt/useragent/UserAgent.gwt.xml#21

iNikem (approved):

instrumentation/gwt-2.0/javaagent/gwt-2.0-javaagent.gradle
laurit: It should be needed and test fails without this for me.

anuraaga: Hmm - myself nor CI could reproduce any failure. Since compileGwt builds the war, what does this task do?

laurit: This task copies greeting.html and WEB-INF/web.xml to war that is deployed by jetty. Without these files the application shouldn't really work because greeting.html is the page that we use for testing and if it is missing there is nothing to test. It is strange that it passed in ci, maybe it cached the test result and didn't really run the test? Anyway seems to have started to fail now https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/2225667052

PR #2660, opened 2021-03-30
Merge war resources

anuraaga:
Something about outputing to the same directory from two tasks seems to have caused CI to pass before without copying the WEB-INF in. This should be more robust.


anuraaga:
@laurit

laurit:
looks good to me

iNikem (approved):

Issue #1677, opened 2020-11-18
Extending InstrumentationModules


pavolloffay:
**Is your feature request related to a problem? Please describe.**

I would like to extend instrumentation modules in my custom build of the Java agent. The main motivation is to avoid bugs like https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1274 which the instrumentation modules solve.

Also the custom instrumentations depend on the core so if the core module is not applied due to mismatch the custom instrumentation should not apply either as it for instance uses current span started in the core module.

**Describe the solution you'd like**

A way to add `TypeInstrumentation` to the instrumentation modules in this project.

**Describe alternatives you've considered**

Extending instrumentation modules from this project has a couple of problems currently:
1. the helper classes via `super.helperClasses()` does not work properly if the super class uses `packageName` - it will be the package name of the extended class.
2. the parent class will be discovered and applied by `AgentInstaller` - this is more of a discussion on how it should be filtered out from the final distribution but also unit tests (in the custom instrumentation maven module).

**Additional context**

Related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1623

mateuszrzeszutek:
Number 1 will be solved by https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1374; after the muzzle improvement is implemented there will be no need for the `packageName` field at all.

I don't have any idea about what to do with number 2 yet.

iNikem:
@pavolloffay @mateuszrzeszutek Can the second problem be solved by custom distro disabling upstream module and then providing new name for its own override? And then extending `typeInstrumentations()`.

To simplify that we may automatically add one more `additionalInstrumentationName` to all our modules, equal to the class name of the module. Then there are much less chance to interfere with client configuration.

mateuszrzeszutek:
> @pavolloffay @mateuszrzeszutek Can the second problem be solved by custom distro disabling upstream module and then providing new name for its own override? And then extending typeInstrumentations().

Yes, it could be done this way. Instead of extending you could use composition, e.g.:
```java
class MyServletModule extends InstrumentationModule {
  private final InstrumentationModule delegate = new OtelServletModule();

  MyServletModule() { super("my-name"); }

  String[] getMuzzleHelperClassNames() {
    return delegate.getMuzzleHelperClassNames();
  }

  List<TypeInstrumentation> typeInstrumentations() {
    // ...
  }
}
```
[...]

PR #2661, opened 2021-03-30
Prevent duplicate telemetry when using both library and auto instrumentation

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/903

I've removed the library instrumentation shading (`io.opentelemetry.instrumentation.*`); instrumentation-api is still shaded though (`io.opentelemetry.instrumentation.api.*`).
Since now the agent will disable the javaagent instrumentation when it encounters a library class I think that this offers a similar level of safety as shading everything.

### DIFF
--- a/instrumentation/apache-dubbo/apache-dubbo-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachedubbo/v2_7/DubboInstrumentationModule.java
+++ b/instrumentation/apache-dubbo/apache-dubbo-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachedubbo/v2_7/DubboInstrumentationModule.java
@@ -33,6 +33,11 @@ public class DubboInstrumentationModule extends InstrumentationModule {
   }
 
   @Override
+  public boolean isLibraryInstrumentationClass(String className) {
+    return className.startsWith("io.opentelemetry.instrumentation.apachedubbo.v2_7");
+  }
+
+  @Override
   public List<TypeInstrumentation> typeInstrumentations() {
     return Collections.singletonList(new DubboInstrumentation());
   }
--- a/instrumentation/couchbase/couchbase-3.1/javaagent/couchbase-3.1-javaagent.gradle
+++ b/instrumentation/couchbase/couchbase-3.1/javaagent/couchbase-3.1-javaagent.gradle
@@ -22,3 +22,8 @@ dependencies {
 
   testImplementation group: "org.testcontainers", name: "couchbase", version: versions.testcontainers
 }
+
+configurations {
+  // exclude lib from test classpath, otherwise the javaagent would be skipped
+  testImplementation.exclude group: 'com.couchbase.client', module: 'tracing-opentelemetry'
+}
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
@@ -5,7 +5,9 @@
 
 package io.opentelemetry.javaagent.tooling;
 
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.callWhenTrue;
 import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.AgentElementMatchers.failSafe;
+import static io.opentelemetry.javaagent.tooling.bytebuddy.matcher.ClassLoaderMatcher.hasAnyClassesNamed;
 import static java.util.Arrays.asList;
 import static net.bytebuddy.matcher.ElementMatchers.any;
 import static net.bytebuddy.matcher.ElementMatchers.isAnnotatedWith;
@@ -29,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import net.bytebuddy.agent.builder.AgentBuilder;
 import net.bytebuddy.description.annotation.AnnotationSource;
 import net.bytebuddy.description.method.MethodDescription;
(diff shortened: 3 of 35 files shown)


instrumentation/apache-dubbo/apache-dubbo-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachedubbo/v2_7/DubboInstrumentationModule.java
```diff
@@ -32,6 +32,11 @@ public DubboInstrumentationModule() {
     return hasClassesNamed("org.apache.dubbo.rpc.Filter");
   }
 
+  @Override
+  public boolean isLibraryInstrumentationClass(String className) {
+    return className.startsWith("io.opentelemetry.instrumentation.apachedubbo.v2_7");
```
iNikem, line 37: Can this be replaced with `this.getClass().getPackage().getName()` and moved to superclass?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
```diff
@@ -175,6 +179,40 @@ public final AgentBuilder instrument(AgentBuilder parentAgentBuilder) {
...
+      return classLoaderMatcher();
+    }
+
+    ElementMatcher.Junction<ClassLoader> libraryMatcher =
+        callWhenTrue(
+            hasAnyClassesNamed(libraryHelperClasses), this::logLibraryInstrumentationDetected);
+    return classLoaderMatcher().and(not(libraryMatcher));
+  }
+
+  private void logLibraryInstrumentationDetected() {
```
iNikem, line 210: Mmm, if we don't shade library instrumentations, we will detect those classes from within our javaagent jar, no? Or am I asking nonsense?

mateuszrzeszutek: We won't, because they're in the agent classloader.

instrumentation/apache-dubbo/apache-dubbo-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachedubbo/v2_7/DubboInstrumentationModule.java
mateuszrzeszutek: And remove `.javaagent`, yes.

I thought about this, but unfortunately this would break vendor specific instrumentation modules that extend ones from this repo. I think that @pavolloffay actually had encountered this problem before.

On the other hand, if we officially recommended composition instead of extending (like I suggested [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1677#issuecomment-734949495)) then we could make all modules final and reap benefits of our package naming convention.

mateuszrzeszutek:
(Rebasing to make those context-leak-failing tests pass)

mateuszrzeszutek:
There's also the matter of testing this: I thought of adding a testSet that contains the library instrumentation and basically running the same test as the library instrumentation module - any duplicate telemetry would fail the tests.

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
```diff
@@ -175,6 +179,40 @@ public final AgentBuilder instrument(AgentBuilder parentAgentBuilder) {
...
+    log.debug(
+        "Skipping instrumentation {} because library instrumentation was detected on classpath",
+        mainInstrumentationName());
```
trask, line 213: 👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
```diff
@@ -342,6 +365,26 @@ public int getOrder() {
...
+   * instrumentation helper classes: they will be scanned for references and automatically injected
+   * into the application classloader if they're used in any type instrumentation.
+   *
+   * <p>In addition to that, the javaagent will prevent the instrumentations from this module from
+   * being applied when it detects that the application classloader already contains one of the
+   * library classes. This behavior prevents interference between library and javaagent
+   * instrumentation (for example, duplicate telemetry).
+   *
+   * @param className The name of the class that may or may not be a library instrumentation class.
+   */
```
trask, line 383: 👍

trask (approved):
nice 👍

instrumentation/couchbase/couchbase-3.1/javaagent/couchbase-3.1-javaagent.gradle
```diff
@@ -21,4 +21,9 @@ dependencies {
...
 
   testImplementation group: "org.testcontainers", name: "couchbase", version: versions.testcontainers
-}
\ No newline at end of file
+}
+
+configurations {
+  // exclude lib from test classpath, otherwise the javaagent would be skipped
+  testImplementation.exclude group: 'com.couchbase.client', module: 'tracing-opentelemetry'
+}
```
trask, line 29: is this a general problem for all javaagent instrumentation that is based on library instrumentation? or only a problem if the javaagent instrumentation depends on 3rd party library instrumentation?

mateuszrzeszutek: No, it's only a problem when you use a 3rd party library instrumentation. Our library instrumentation jars are filtered out in `gradle/instrumentation.gradle`.

anuraaga:
Sorry for very late review - are we confident classpath is a good enough signal to back off? For example, I might depend on a library which uses gRPC, and actually initializes gRPC library instrumentation itself, but that doesn't mean my own handlers will have it applied. So I thought we were going to need to go with an approach where the agent detects whether the library instrumentation is actually being used, not just the class presence.

trask:
ah, that's a good point

mateuszrzeszutek:
Oh, that is an excellent point...

And the worst thing is that there's no reasonable workaround for that scenario: you can't exclude the grpc library instrumentation because the library that uses it will break.

Hmm, I'll probably revert this and go back to the drawing board.

PR #2665, opened 2021-03-30
Add AsyncSpanEndStrategy implementation for RxJava 2 instrumentation

HaloFour:
Adds an implementation of `AsyncSpanEndStrategy` which supports RxJava 2 publishers returned from methods annotated with the `@WithSpan` annotation.

### DIFF
--- a/instrumentation/rxjava-2.0/javaagent/src/test/groovy/RxJava2WithSpanInstrumentationTest.groovy
+++ b/instrumentation/rxjava-2.0/javaagent/src/test/groovy/RxJava2WithSpanInstrumentationTest.groovy
@@ -8,8 +8,20 @@ import io.opentelemetry.instrumentation.rxjava2.TracedWithSpan
 import io.opentelemetry.instrumentation.test.AgentInstrumentationSpecification
 import io.opentelemetry.instrumentation.test.AgentTestTrait
 import io.reactivex.Completable
+import io.reactivex.Flowable
+import io.reactivex.Maybe
+import io.reactivex.Observable
+import io.reactivex.Single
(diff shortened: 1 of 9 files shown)


trask (commented):
@HaloFour can you add javaagent tests also? (verifying the spans produced from `@WithSpan` with these new RxJava return types)

Unfortunately lots extra can go wrong with bytecode instrumentation (shading, class loaders, version conflicts, ...), so having real javaagent tests helps our confidence level a lot.

instrumentation/rxjava-2.0/javaagent/src/test/groovy/RxJava2WithSpanInstrumentationTest.groovy
```diff
@@ -0,0 +1,856 @@
...
+import io.reactivex.subjects.CompletableSubject
+import io.reactivex.subjects.MaybeSubject
+import io.reactivex.subjects.SingleSubject
+import io.reactivex.subjects.UnicastSubject
+import io.reactivex.subscribers.TestSubscriber
+import org.reactivestreams.Publisher
+import org.reactivestreams.Subscriber
+import org.reactivestreams.Subscription
+
+class RxJava2WithSpanInstrumentationTest extends AgentInstrumentationSpecification implements AgentTestTrait {
```
mateuszrzeszutek, line 26: There's no need to implement `AgentTestTrait` when you already extend `AgentInstrumentationSpecification` - it's already implemented:

```suggestion
class RxJava2WithSpanInstrumentationTest extends AgentInstrumentationSpecification {
```

trask (approved):
thx!

PR #2666, opened 2021-03-30
Fix muzzle failure

trask:
May just be a blip mid-publishing new `2.2.7.Final` release, but it's holding up a couple of PRs due to CI failure.

Probably will fix itself at some point, `maven-metadata.xml` at
https://repo1.maven.org/maven2/io/undertow/undertow-core/
contains `2.2.7.Final`, but there's no `2.2.7.Final` published yet.

Still think worth merging this to unblock PRs. Or alternatively, we could override the check and merge PRs blocked on this anyways.


tylerbenson (approved):

trask:
Problem still not fixed, merging.

iNikem:
@trask you can open an issue in sonatype issue track and ask them to refresh they metadata.

trask:
it got fixed, reverting this change in #2671

PR #2671, opened 2021-03-31
Remove unnecessary skipVersions

trask:
Well, looks like undertow-core `2.2.7.Final` finally reached maven central, so can at least revert #2666, but curious how many of these `skipVersion` are no longer needed, especially now that we are not pulling versions from jcenter anymore...


iNikem (approved):

instrumentation/elasticsearch/elasticsearch-transport-5.3/javaagent/elasticsearch-transport-5.3-javaagent.gradle
```diff
@@ -5,19 +5,17 @@ muzzle {
     group = "org.elasticsearch.client"
     module = "transport"
     versions = "[5.3.0,6.0.0)"
-    // TODO (trask) try to remove the skipVersions in the future
-    //  not sure what's wrong with this recent version, hopefully just publishing blip that
-    //  will work itself out
+    // version 7.11.0 depends on org.elasticsearch:elasticsearch:7.11.0 which depends on
+    // org.elasticsearch:elasticsearch-plugin-classloader:7.11.0 which does not exist
     skipVersions += ['7.11.0']
```
iNikem, line 10: If this is needed here, why it's not needed in `elasticsearch-transport-5.0-javaagent.gradle` above?

iNikem (approved):

instrumentation/elasticsearch/elasticsearch-transport-5.3/javaagent/elasticsearch-transport-5.3-javaagent.gradle
trask: this is a good question, may be getting lucky, I just sent a temporary commit to remove the muzzle range limit so it will test everything

trask: Good call 👍, I will fix the remaining failures (4!) tomorrow.

buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -31,14 +31,15 @@ import org.gradle.api.Plugin
...
+
 /**
  * muzzle task plugin which runs muzzle validation against a range of dependencies.
  */
 class MuzzlePlugin implements Plugin<Project> {
   /**
    * Select a random set of versions to test
    */
-  private static final int RANGE_COUNT_LIMIT = 10
+  private static final int RANGE_COUNT_LIMIT = 10000
```
mateuszrzeszutek, line 42: Revert? 😄

instrumentation/couchbase/couchbase-2.6/javaagent/couchbase-2.6-javaagent.gradle
```diff
@@ -5,7 +5,7 @@ muzzle {
     group = 'com.couchbase.client'
     module = 'java-client'
     versions = "[2.6.0,3)"
-    // Version 2.7.5 and 2.7.8 were not released properly and muzzle cannot test against it causing failure.
+    // these version were released as ".bundle" instead of ".jar"
```
mateuszrzeszutek, line 8:
```suggestion
    // these versions were released as ".bundle" instead of ".jar"
```

instrumentation/couchbase/couchbase-3.1/javaagent/couchbase-3.1-javaagent.gradle
```diff
@@ -5,7 +5,7 @@ muzzle {
     group = "com.couchbase.client"
     module = "java-client"
     versions = "[3.1,)"
-    // Version 2.7.5 and 2.7.8 were not released properly and muzzle cannot test against it causing failure.
+    // these version were released as ".bundle" instead of ".jar"
```
mateuszrzeszutek, line 8:
```suggestion
    // these versions were released as ".bundle" instead of ".jar"
```

mateuszrzeszutek (approved):

PR #2672, opened 2021-03-31
Don't instrument rxjava2 internal task class.

anuraaga:


instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
```diff
@@ -41,15 +43,23 @@ public RxJava2InstrumentationModule() {
...
       return Collections.singletonMap(
-          isTypeInitializer(),
-          RxJava2InstrumentationModule.class.getName() + "$RxJavaPluginsAdvice");
+          isMethod(), RxJava2InstrumentationModule.class.getName() + "$RxJavaPluginsAdvice");
     }
   }
 
   public static class RxJavaPluginsAdvice {
+    @Advice.OnMethodEnter
+    public static int trackCallDepth() {
```
anuraaga, line 52: This is a terrible solution! For now just sending to demonstrate that previously the assembly wasn't getting called. The change to disable executor instrumentation for rxjava tasks caused this test to fail until I fixed this part.

anuraaga: If decorating every method is really the only way, I'd at least use a `ClassValue<AtomicBoolean>` instead of calldepth.

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
```diff
@@ -41,15 +43,23 @@ public RxJava2InstrumentationModule() {
...
+    }
+
     @Advice.OnMethodExit(suppress = Throwable.class)
-    public static void postStaticInitializer() {
+    public static void postStaticInitializer(@Advice.Enter int callDepth) {
+      if (callDepth > 0) {
+        return;
+      }
       TracingAssembly.enable();
+      CallDepthThreadLocalMap.reset(RxJavaPlugins.class);
```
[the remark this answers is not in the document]
anuraaga: I just pushed a change to remove the call depth tracking in favor of a boolean so `enable()` is only called once ever - mainly since `enable` uses `synchronized` as we didn't expect the contention to matter.

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
```diff
@@ -41,15 +42,16 @@ public RxJava2InstrumentationModule() {
...
+          isMethod(), RxJava2InstrumentationModule.class.getName() + "$RxJavaPluginsAdvice");
     }
   }
 
   public static class RxJavaPluginsAdvice {
-    @Advice.OnMethodExit(suppress = Throwable.class)
-    public static void postStaticInitializer() {
-      TracingAssembly.enable();
+
+    // TODO(anuraaga): Replace with adding a type initializer to RxJavaPlugins
```
anuraaga, line 51: @trask Perhaps we can merge and fix forward if the type initializer stuff is too complicated right now.

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
```diff
@@ -14,40 +14,68 @@
 /** Utils for concurrent instrumentations. */
 public class ExecutorInstrumentationUtils {
 
-  private static final ClassValue<Boolean> NOT_INSTRUMENTED_RUNNABLE_ENCLOSING_CLASS =
+  private static final ClassValue<Boolean> INSTRUMENTED_RUNNABLE_CLASS =
```
iNikem, line 17: Are changes in this class related to RxJava changes?

anuraaga: Changes in this class caused the rxjava test to fail unexpectedly :( There's no way to verify the behavior of the rxjava instrumentation without the changes in this class (as we found in your muzzle PR) so they're somewhat tied.

anuraaga: Though the change is larger than it could have been here since I figured I had too many `if` statements by now to not cache it

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
iNikem: Do we have a registered issue for this? If yes, can you reference it here?

iNikem (approved):

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
trask, line 51:
```suggestion
    // TODO(anuraaga): Replace with adding a type initializer to RxJavaPlugins
    // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2685
```

trask (approved):
👍

instrumentation/rxjava-2.0/javaagent/src/main/java/io/opentelemetry/instrumentation/rxjava2/RxJava2InstrumentationModule.java
trask: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2685, and pushed to add to the comment

PR #2675, opened 2021-03-31
Split HttpClientTest execution methods for sync and callback

anuraaga:
While debugiging a test failure, I noticed our async httpclient tests are often messed up - in many tests it even gets executed in the caller thread after a blocking get, not on an async callback thread which is what we need to test.

I have ~~three~~ one PRs planned to fix this

1. Return status through callback instead of synchronous invocation (this one)
2. Add a knob to disable async tests. Disable them on libraries that don't support async invocation, there's nothing to test there
3. Split into `int executeRequest(method, headers)` and `void executeRequestAsync(method, headers, callback)` to make obvious that the async versions are supposed to be completely async.


instrumentation/akka-http-10.0/javaagent/src/test/groovy/AkkaHttpClientInstrumentationTest.groovy
```diff
@@ -25,20 +26,16 @@ class AkkaHttpClientInstrumentationTest extends HttpClientTest implements AgentT
...
   @Override
-  int doRequest(String method, URI uri, Map<String, String> headers, Closure callback) {
+  int doRequest(String method, URI uri, Map<String, String> headers, Consumer<Integer> callback) {
     def request = HttpRequest.create(uri.toString())
       .withMethod(HttpMethods.lookup(method).get())
       .addHeaders(headers.collect { RawHeader.create(it.key, it.value) })
 
     def response = Http.get(system)
       .singleRequest(request, materializer)
-    //.whenComplete { result, error ->
```
anuraaga, line 35: Removed this comment since this is in scope of my 3 PRs (and also an example of where the async test isn't testing)

instrumentation/play-ws/play-ws-1.0/javaagent/src/test/groovy/PlayWsClientTest.groovy
```diff
@@ -46,18 +46,17 @@ class PlayJavaStreamedWsClientTest extends PlayWsClientTestBase {
...
     StandaloneWSResponse wsResponse = wsRequest.setMethod(method).stream()
-      .whenComplete({ response, throwable ->
-        callback?.call()
-      }).toCompletableFuture().get(5, TimeUnit.SECONDS)
+      .toCompletableFuture().get(5, TimeUnit.SECONDS)
 
     // The status can be ready before the body so explicitly call wait for body to be ready
     wsResponse.getBodyAsSource().runFold("", { acc, out -> "" }, materializer)
       .toCompletableFuture().get(5, TimeUnit.SECONDS)
+    callback?.accept(wsResponse.getStatus())
```
anuraaga, line 59: While the play tests seem to have regressed in their async-ness, presumably they're already sort of broken since I'd expect this "body to be ready" caveat to be required for both sync and async. Will address when splitting out sync / async handlers.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -163,15 +165,18 @@ abstract class HttpClientTest extends InstrumentationSpecification {
     given:
     assumeTrue(testCallbackWithParent())
 
+    def status = new BlockingVariable<Integer>()
```
anuraaga, line 168: I started with this PR so the test framework is in charge of blocking, not the request executors. Now the latter can be switched to async since they don't need to block.

iNikem:
I think there is some confusion here. IMO we test 2 things in our HttpClientTests:
- That context propagation works correctly
- 2 tests verify that if particular library provides some kind of response callback mechanism, then that callback is executed in the correct context

This PR seems to change only the second point from above. And I am not quite sure how returning of the response status code is related to your concern of sync vs a sync request execution.

anuraaga:
> This PR seems to change only the second point from above. 

Yeah - I'm only concerned with the asynchronous tests, which are the ones that use callbacks, context propagation for the synchronous ones seems fine.

>  And I am not quite sure how returning of the response status code is related to your concern of sync vs a sync request execution.

This PR is mostly to unblock for step 3) in my list. As long as `executeRequest` has to synchronously return a status code, it makes it difficult, or at least confusing to me, to implement an asynchronous response handler in them - we effectively have to handle a response _both_ asynchronously and synchronously. Now that the base class is listening for the status via the callback I can split so asynchronous method doesn't expect a return value.

iNikem:
How do you to implements tests without callbacks for async clients?

O, and actually even in case of callback, you still want to wait for an answer in your test, right? So you are going to block anyway.

I think I understand your concern, but don't understand yet how are you going to solve it.

anuraaga:
> How do you to implements tests without callbacks for async clients?

Not sure I understand - I will implement the tests for async clients _with_ callbacks. They won't return any status code synchronously though.

> So you are going to block anyway.

Yeah but the test method blocks now, not the request execution which makes writing the async request execution more natural.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2675/files#r604653982

anuraaga:
@iNikem I thought splitting the PRs could make it easier to review but if not I can combine 1-3 into one too :)

iNikem:
As I am confused by this PR, probably it will be easier to review everything together indeed :)

instrumentation/akka-http-10.0/javaagent/src/test/groovy/AkkaHttpClientInstrumentationTest.groovy
```diff
@@ -25,21 +28,32 @@ class AkkaHttpClientInstrumentationTest extends HttpClientTest implements AgentT
...
-    // FIXME: Callback should be here instead.
-    //  callback?.call()
-    //}
-      .toCompletableFuture()
-      .get()
-    callback?.call()
-    return response.status().intValue()
+  }
+
+  // TODO(anuraaga): Context leak seems to prevent us from running two asynchronous tests in a row.
```
anuraaga, line 51: 😱 🙀

instrumentation/apache-httpasyncclient-4.1/javaagent/src/test/groovy/ApacheHttpAsyncClientCallbackTest.groovy
```diff
@@ -1,79 +0,0 @@
-/*
```
anuraaga, line 1: Callback and null callback are both handled in the main test now, so deleted these but check my work that they were indeed redundant (AFAICT, one of the already was even before my PR)

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/test/groovy/ApacheHttpClientResponseHandlerTest.groovy
```diff
@@ -1,55 +0,0 @@
-/*
```
anuraaga, line 1: I couldn't find a difference between this test and the `ResponseHandler` tests in the other file (now the `doRequestAsync` tests) so deleted, but check my work that it was indeed redundant.

instrumentation/apache-httpclient/apache-httpclient-5.0/javaagent/src/test/groovy/ApacheHttpClientResponseHandlerTest.groovy
```diff
@@ -1,96 +0,0 @@
-/*
```
anuraaga, line 1: Same here

anuraaga (commented):
@iNikem Finished the effort, hope this PR makes more sense now

instrumentation/akka-http-10.0/javaagent/src/test/groovy/AkkaHttpClientInstrumentationTest.groovy
anuraaga: By the way, the reason I know context leak is related is thanks to @jkwatson's awesome logging, showing a client span being suppressed.

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/src/test/groovy/JaxRsClientTest.groovy
```diff
@@ -81,6 +103,11 @@ abstract class JaxRsClientTest extends HttpClientTest implements AgentTestTrait
     "/client-error" | 400
     "/error"        | 500
   }
+
+  @Override
+  boolean testAsync() {
+    false
+  }
```
trask, line 110: I see doRequestAsync above
```suggestion
```

trask (approved):
nice

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/src/test/groovy/JaxRsClientTest.groovy
anuraaga: Ah yeah I was surprised to find the async pattern for a seemingly old library :)

instrumentation/vertx-reactive-3.5/javaagent/src/test/groovy/client/VertxRxCircuitBreakerWebClientTest.groovy
```diff
@@ -34,27 +36,34 @@ class VertxRxCircuitBreakerWebClientTest extends HttpClientTest implements Agent
...
 
   @Override
-  int doRequest(String method, URI uri, Map<String, String> headers, Closure callback) {
-    def request = client.request(HttpMethod.valueOf(method), uri.port, uri.host, "$uri")
-    headers.each { request.putHeader(it.key, it.value) }
-
-    def future = new CompletableFuture<Integer>()
+  int doRequest(String method, URI uri, Map<String, String> headers = [:]) {
+    // VertxRx doesn't seem to provide a synchronous API at all for circuit breaker. Bridge through
+    // a callback.
```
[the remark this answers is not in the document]
anuraaga: Perhaps it's a bit confusing, but the "sync" tests are also the baseline - they test most of the normal features of tracing, so even if a library only supports async we still need to run them. Unless we copied the tests into async versions but I don't think we need to.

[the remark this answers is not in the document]
trask: 👍 makes sense

iNikem:
Why are these two tests important only for async http clients? Even for sync clients it is important that CLIENT span has ended before calling provided callback.

Also I think we need some documentation to explain what implementors of `doRequest` and `doRequestAsync` should do or think about.

anuraaga:
> Even for sync clients it is important that CLIENT span has ended before calling provided callback.

@iNikem Sync request execution returns a result synchronously and has no callback, async does not return anything synchronously, instead through a callback. I think it is the definition of sync vs async - so I don't think this is a true statement since for sync statements, there is no provided callback at all :) 

Our test framework was providing a callback but calling it synchronously after request execution for sync clients - this doesn't test anything and more importantly doesn't reflect any sort of user code since there is no concept of a callback with such clients.

It's true that with the current pattern we do have an assumption that for the baseline, sync and async requests behave the same so we don't have to test everything with a callback. Perhaps this is not precise enough - but with this PR we can add such tests in the base class later without any more refactoring.

> Also I think we need some documentation

Yeah will add some docs.

iNikem:
@anuraaga It is not true, that only async clients have a notion of callback. The simplest example is Spring's `RestTemplate`: it allows users to specify both `ResponseExtractor` and `ResponseErrorHandler` callbacks. In both cases we want to verify that they are executed not in the context of the CLIENT span but of its parent. We want to test context "demarcation", not only context "propagation".

anuraaga:
@iNikem I think I see what you're saying, but I guess that's why I modeled those callbacks using `doRequestAsync`, but it's true that now the tests don't necessarily follow user code pattern. I'm not sure if it's an important point though, do you see any test coverage that has been lost with this PR? 

I don't think we've ever tested "span has ended before calling provided callback." in any test before, so we would need to add test coverage of that. The main purpose of this pr is to make sure we never have any more tests that call the callback after the request is made from test code instead of as a registered callback, since it's a broken test. Having different code for sync pattern and async pattern is important I think. If there's anything you think is missing other than docs let me know.

iNikem:
> I don't think we've ever tested "span has ended before calling provided callback." in any test before

I think that "trace request with callback and parent" tested exactly that in case of sync clients. And calling callback from the test code is correct for sync clients, because that is exactly what application code will do.

My point is that your desire to fix async http clients by ensuring that we hand over the provided callback to them to call it from (potentially) another thread is a good one. Also we probably should verify that if http client accepts callbacks, then we hand our callback to it and not call it ourselves. E.g. `SpringRestTemplateTest` should be fixed to use `ResponseExtractor` instead of calling callback directly. But both sync and async clients should be tested by "trace request with callback and parent" and "trace request with callback and no parent".

anuraaga:
> I think that "trace request with callback and parent" tested exactly that in case of sync clients. And calling callback from the test code is correct for sync clients, because that is exactly what application code will do.

But we don't ever check timings in test, so I guess it could be done before or after the callback, as long as it's done by `waitForTraces` right?

> E.g. SpringRestTemplateTest should be fixed to use ResponseExtractor instead of calling callback directly. 

Yeah in this PR that's already happening in `SpringRestTemplate.doRequestAsync`, maybe it's not obvious since it's a Groovy closure being automatically converted to ResponseExtractor.

> But both sync and async clients should be tested by "trace request with callback and parent" and "trace request with callback and no parent".

Do you mean for `HttpUrlConnectionTest` we should have `doRequestAsync` which looks like this?

```
HttpURLConnection connection = uri.toURL().openConnection()
    try {
      connection.setRequestMethod(method)
      headers.each { connection.setRequestProperty(it.key, it.value) }
      connection.setRequestProperty("Connection", "close")
      connection.useCaches = true
      connection.connectTimeout = CONNECT_TIMEOUT_MS
      def parentSpan = Span.current()
      def stream = connection.inputStream
      assert Span.current() == parentSpan
      stream.readLines()
      stream.close()
      callback.accept(connection.getResponseCode()) // Inserted here
    } finally {
      connection.disconnect()
    }
```

I don't think so, this callback isn't testing anything real since the callback isn't used by the library at all - what bug might we catch by calling `callback` in this way?

iNikem:
> But we don't ever check timings in test, so I guess it could be done before or after the callback, as long as it's done by `waitForTraces` right?

Sorry, I did not understand what is "it" in this sentence and how timings are related to our discussion :) For me **before** in "span has ended **before** calling provided callback" means not time, but the fact that callback span does NOT have CLIENT as its parent.

Now, let us take a step back. This PR seems to make a distinction between sync and async http clients. And this PR says that we should test callback handling only for async http clients. Thus `doRequestAsync` has callback parameter and `doRequest` has not. I am trying to say that we should test callback handling for all http clients which accept callbacks, regardless if they are sync or async.

anuraaga:
>  I am trying to say that we should test callback handling for all http clients which accept callbacks, regardless if they are sync or async.

Ok I think this makes sense - is it enough to rename `Async` to `WithCallback` in my usage in this PR?

iNikem:
> Ok I think this makes sense - is it enough to rename `Async` to `WithCallback` in my usage in this PR?

Then you have to rename `testAsync()` as well :) And then why should we have 2 separate `doRequest` methods? :) I in fact do want two separate methods, but playing Devil's advocate here.

anuraaga:
I renamed and added some docs, does it look ok? The main reason to have two methods is because with one method, it's too easy to have a successful, but meaningless callback test, like this one

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy#L44

with two separate methods, it's basically not possible for the test to pass without actually exercising the callback functionality correctly.

iNikem (approved):

PR #2680, opened 2021-03-31
Concurrent server test

iNikem:
Please read documentation for newly added "high concurrency test" test for the explanation


trask (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -99,6 +105,10 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
     true
   }
 
+  boolean testConcurrency() {
+    false
```
mateuszrzeszutek, line 109: Should it be true by default?

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -361,6 +372,70 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
...
+          def requestId = Integer.parseInt(rootSpan.name.substring("client ".length()))
+
+          basicSpan(it, 0, "client " + requestId, null, null) {
+            it."test.request.id" requestId
+          }
+          indexedServerSpan(it, span(0), requestId)
+          indexedControllerSpan(it, 2, span(1), requestId)
+        }
+      }
+    }
```
mateuszrzeszutek, line 436: Should we cleanup the pool?

```suggestion
    }
    
    cleanup:
    pool.shutdownNow()
```

mateuszrzeszutek (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
iNikem: This requires implementing support for the new endpoint, `INDEXED_CHILD`. Thus it cannot be default.

PR #2681, opened 2021-03-31
Cache config values in instrumentations

### LINKED ISSUE #1798: Cache config values in instrumentation
This is not possible currently because we sometimes test with different config values in the same JVM, but will be possible after #1643, when running with real javaagent and static config.

trask:
Closes #1798

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
@@ -98,7 +98,7 @@ public abstract class InstrumentationModule {
       throw new IllegalArgumentException("InstrumentationModules must be named");
     }
     this.instrumentationNames = new LinkedHashSet<>(instrumentationNames);
-    enabled = Config.get().isInstrumentationEnabled(this.instrumentationNames, DEFAULT_ENABLED);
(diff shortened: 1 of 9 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
```diff
@@ -354,8 +357,4 @@ public int getOrder() {
   protected Map<String, String> contextStore() {
     return Collections.emptyMap();
   }
-
-  protected boolean defaultEnabled() {
```
mateuszrzeszutek, line 358: Please leave this method -- it's useful to extend that in vendor distros.

mateuszrzeszutek: Thanks!

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/InstrumentationModule.java
trask: oh my, I just saw the build failure from when I had removed it, apparently it _is_ used, bad Intellij 😂

(I updated the comment just now to remove the part about it not being used in this repo)

anuraaga (approved):

PR #2682, opened 2021-03-31
Remove oshi-related code from shared agent code

### LINKED ISSUE #1565: Remove knowledge of oshi instrumentation from the shared agent code
Currently the oshi instrumentation is initialized in the system class loader, directly from `OpenTelemetryAgent`.

Ideally, the shared agent code shouldn't know about any individual instrumentation, interacting with the instrumentation instead only via SPI interface(s).

See https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1508#issuecomment-719213581 for one option that has been discussed specific to this oshi initialization.

trask:
Closes #1565

### DIFF
--- a/instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/OshiMetricsInstaller.java
+++ b/instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/OshiMetricsInstaller.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.oshi;
+
+import com.google.auto.service.AutoService;
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.javaagent.spi.ComponentInstaller;
+import java.lang.reflect.Method;
+import java.util.Collections;
+
+/**
+ * {@link ComponentInstaller} to enable oshi metrics during agent startup if oshi is present on the
+ * system classpath.
+ */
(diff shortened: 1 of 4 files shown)


mateuszrzeszutek (approved):

instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/OshiMetricsInstaller.java
```diff
@@ -0,0 +1,36 @@
...
+    if (config.isInstrumentationEnabled(Collections.singleton("oshi"), true)) {
+      try {
+        // Call oshi.SystemInfo.getCurrentPlatformEnum() to activate SystemMetrics.
+        // Oshi instrumentation will intercept this call and enable SystemMetrics.
+        Class<?> oshiSystemInfoClass =
+            ClassLoader.getSystemClassLoader().loadClass("oshi.SystemInfo");
+        Method getCurrentPlatformEnumMethod =
+            oshiSystemInfoClass.getMethod("getCurrentPlatformEnum");
+        getCurrentPlatformEnumMethod.invoke(null);
+      } catch (Throwable ex) {
```
iNikem, line 31: Catching Errors is fishy :(

iNikem (approved):

anuraaga (approved):

instrumentation/oshi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/oshi/OshiMetricsInstaller.java
trask: 🐠 🐟 🐠 🐟 will let this fish live for another day

PR #2689, opened 2021-04-01
Revert "Prevent duplicate telemetry when using both library and auto …

mateuszrzeszutek:
…instrumentation (#2661)"

Reverts #2661 because of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2661#issuecomment-811599122


anuraaga (approved):
Thanks, sorry for this churn

trask (approved):

PR #2692, opened 2021-04-01
Implement getContentLength for InternalJarUrlHandler

laurit:
Currently every time a class is loaded form `inst` there is a `NumberFormatException` which makes debugging a pain when you need to set an exception breakpoint for `IllegalArgumentException`


iNikem (approved):

mateuszrzeszutek (approved):

richardstartin:
I think you can also remove the prefix "inst" as it is a carry over from Datadog, where we bundle several independently enabled components into the same deployment and keep them separate from each other by prefix in the jar file (each component has its own prefix and its own isolated classloader which will only load from that prefix) as well as a classloader "shared" which loads classes common to all components (but keeps these classes otherwise isolated). A quick look at your build configuration suggests that maintaining the prefix isn't serving a purpose in this code base.

PR #2693, opened 2021-04-01
Enable debug logging for docker java client

laurit:
Maybe extra logging will show why windows smoke tests run slow.


iNikem:
It's fine for this debug PR, but I really don't want this in main branch. One of the main reasons why I tuned logging at all was to remove docker java logging :D

PR #2694, opened 2021-04-01
Update java in windows fake backend image

laurit:
My hypothesis is that to make windows tests faster our test images should use the windows base images that are cached on github action runners https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#cached-docker-images. These cached images are occasionally update. Fake backend image is base on a java image that is ~3 months old replacing it with a newer image will hopefully help.


mateuszrzeszutek (approved):

iNikem:
Should we update all other windows-based images as well?

iNikem (approved):

laurit:
@iNikem I didn't look too deeply into that but I suspect that those are fine for now

PR #2699, opened 2021-04-02
Fix using WithSpan annotation in java6 class

laurit:


instrumentation/opentelemetry-annotations-1.0/javaagent/src/test/groovy/WithSpanInstrumentationTest.groovy
```diff
@@ -370,4 +381,56 @@ class WithSpanInstrumentationTest extends AgentInstrumentationSpecification {
...
     }
   }
+
+  def "instrument java6 class"() {
+    setup:
+    /*
+     class GeneratedJava6TestClass implements Runnable {
+       @WithSpan
+       public void run() {
+       }
```
mateuszrzeszutek, line 391:
```suggestion
       public void run() {
         TraceUtils.runUnderTrace("intercept", {})
       }
```

kubawach (commented):
LGTM. We could probably make the java6 test more readable by moving it to a separate test module with maxJavaVersionForTests set to 6.

instrumentation/opentelemetry-annotations-1.0/javaagent/opentelemetry-annotations-1.0-javaagent.gradle
```diff
@@ -8,6 +8,7 @@ dependencies {
   compileOnly project(path: ':opentelemetry-ext-annotations-shaded-for-instrumenting', configuration: 'shadow')
 
   testImplementation deps.opentelemetryExtAnnotations
+  testImplementation "net.bytebuddy:byte-buddy-gradle-plugin:${versions.bytebuddy}"
```
mateuszrzeszutek, line 11: Why gradle plugin and not just byte buddy?

mateuszrzeszutek (approved):

instrumentation/opentelemetry-annotations-1.0/javaagent/opentelemetry-annotations-1.0-javaagent.gradle
laurit: copy-pasted it without paying attention, fixed.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/test/groovy/WithSpanInstrumentationTest.groovy
laurit: Fixed

mateuszrzeszutek:
> We could probably make the java6 test more readable by moving it to a separate test module with maxJavaVersionForTests set to 6.

Unfortunately I don't think this'd work: tests JVM process runs with our javaagent, which does not support Java <8.

instrumentation/opentelemetry-annotations-1.0/javaagent/src/test/groovy/WithSpanInstrumentationTest.groovy
```diff
@@ -370,4 +381,57 @@ class WithSpanInstrumentationTest extends AgentInstrumentationSpecification {
...
+        void intercept(@This Object o) {
+          TraceUtils.runUnderTrace("intercept", {})
+        }
+      }))
+      .visit(new MemberAttributeExtension.ForMethod()
+        .annotateMethod(AnnotationDescription.Builder.ofType(WithSpan).build())
+        .on(ElementMatchers.named("run")))
+      .make()
+      .load(getClass().getClassLoader())
+      .getLoaded()
```
trask, line 410: nice, I would have found a (random old) library that was compiled to Java 6 and tested instrumentation of that library, but this is better 👍

trask (approved):

Issue #2420, opened 2021-02-26
Broken JMS MessageProducer propagation


mateuszrzeszutek:
**Describe the bug**
`JmsMessageProducerInstrumentation` instruments two methods with two different advice classes: one of them injects the context into the message, the other does not.

**Additional context**
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java#L80
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java#L117

PR #2702, opened 2021-04-02
Fix JMS context propagation problems

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2420

### DIFF
--- a/instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java
+++ b/instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java
@@ -77,9 +77,7 @@ public class JmsMessageProducerInstrumentation implements TypeInstrumentation {
       MessageDestination messageDestination =
           tracer().extractDestination(message, defaultDestination);
       context = tracer().startProducerSpan(messageDestination, message);
-      // TODO: why are we propagating context only in this advice class? the other one does not
-      // inject current span context into JMS message
-      scope = tracer().startProducerScope(context, message);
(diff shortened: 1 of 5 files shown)


trask (approved):
is it easy to add a test for this?

iNikem (approved):

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
```diff
@@ -45,6 +45,13 @@ protected Boolean computeValue(Class<?> taskClass) {
             return false;
           }
 
+          // OrderedExecutor$1 is a worker class that processes tasks submitted to OrderedExecutor
```
anuraaga, line 48: Use `enclosingClass == OrderedExecutorFactory` similar to the below instead of this one since it's more readable than a lambda

mateuszrzeszutek:
> is it easy to add a test for this?

For hornetq and the tracer context leak fix - existing tests sort of cover that.

For the `ProducerWithDestinationAdvice` fix - that should be possible, I'll do that.

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageProducerInstrumentation.java
```diff
@@ -129,6 +127,7 @@ public static void stopSpan(
       }
       CallDepthThreadLocalMap.reset(MessageProducer.class);
 
+      scope.close();
```
mateuszrzeszutek, line 130: ...aaaand one more scope leak detected 😄 
It was worth adding that one test case

Issue #2674, opened 2021-03-09
HTTP Span Attributes: `http.url` must not contain username / password


pellared:
As is stated in the recent [specification change](https://github.com/open-telemetry/opentelemetry-specification/pull/1502) :

> `http.url` MUST NOT contain credentials passed via URL in form of `https://username:password@www.example.com/`. In such case the attribute's value should be `https://www.example.com/`

anuraaga:
Hi @Hangzhi - thanks for the help on this. You can find where we set the attribute, with no scrubbing of credentials I think, here

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java#L186

pellared:
@Hangzhi I think you can also look at this: https://github.com/open-telemetry/opentelemetry-java/issues/3118 which should be also simpler😉

Hangzhi:
@pellared Thanks! I'll work on that.

MALPI:
Hey folks,

I would be happy to contribute here.

mateuszrzeszutek:
Thanks! I'll assign this issue to you.

PR #2707, opened 2021-04-03
Prevent http.url containing credentials in HttpClientTracer

Hangzhi:
Fixes this #2674


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -182,9 +182,13 @@ private void setUrl(AttributeSetter setter, REQUEST request) {
     try {
       URI url = url(request);
       if (url != null) {
+        if (url.getUserInfo() != null) {
+          throw new IllegalArgumentException("http.url MUST NOT contain credentials!");
```
anuraaga, line 186: Hi @Hangzhi - thanks for the contribution! We don't want to cause an application error due to the presence of credentials, only that we don't record it. Below we call `url.toString` and we just need to make sure that is called with a url that has had the user info removed.

Also, we will need a unit test for the behavior. Here we have many different urls /expected url

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/HttpClientTracerTest.groovy#L73

So you can add a test which adds a url with credentials and an expected url without.

Hangzhi: Thanks for your review! I'll work on that.

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
Hangzhi: @aantono Hello aantono. I removed the user info and added a test according to your comment. I got this [failing test
](https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/2276486894?check_suite_focus=true#step:7:3066), but I did some search but still cannot figure out how to fix this. Could you please give me some more hints on how to solve this problem?

trask: hey @Hangzhi, this is a known sporadic error that will be fixed by #2728

> should suppress nested CLIENT span if already under parent CLIENT span

trask: I hit re-run on the tests for now

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -183,8 +183,14 @@ private void setUrl(AttributeSetter setter, REQUEST request) {
       URI url = url(request);
       if (url != null) {
         netPeerAttributes.setNetPeer(setter, url.getHost(), null, url.getPort());
-        setter.setAttribute(SemanticAttributes.HTTP_URL, url.toString());
+        if (url.getUserInfo() != null) {
+          setter.setAttribute(
+              SemanticAttributes.HTTP_URL, url.toString().replace(url.getUserInfo() + '@', ""));
```
mateuszrzeszutek, line 188: Instead of using string replacement (which can replace more than once) consider creating a new `URI` without the user info part:

```java
new URI(
    url.getScheme(),
    null,
    url.getHost(),
    url.getPort(),
    url.getPath(),
    url.getQuery(),
    url.getFragment());
```

mateuszrzeszutek (commented):
We also have a few more usages of this attribute: `HttpAttributesExtractor` from the instrumenter API (should `#url()` return `URI`? I think it should), the `HttpServerTracer`, `apache-camel` and `aws-lambda` instrumentations.

Can you fix all of them? Thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
Hangzhi: Thanks for your suggestion! I'll fix them all.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -183,7 +183,17 @@ private void setUrl(AttributeSetter setter, REQUEST request) {
...
+            SemanticAttributes.HTTP_URL,
+            new URI(
+                    url.getScheme(),
+                    null,
+                    url.getHost(),
+                    url.getPort(),
+                    url.getPath(),
+                    url.getQuery(),
+                    url.getFragment())
+                .toString());
```
trask, line 196: hi @Hangzhi! I think it would be a worthwhile optimization here to only create a new URI object when the existing URI has user-info

iNikem: @Hangzhi, are you willing to address this comment?

anuraaga: Thanks for the help @Hangzhi I've gone ahead and applied this suggestion to the PR

anuraaga (approved):

iNikem (approved):

PR #2710, opened 2021-04-03
Remove duplicate status verification

### LINKED ISSUE #608: Verify actual status instead of just errored
OTel spans have a status with different codes but many of our tests only test success vs error. We should migrate to asserting on the status instead.

trask:
I think this closes #608 now that there are only 3 status codes, and auto-instrumentation should never set `OK`, leaving only `ERROR` or `UNSET`.

### DIFF
--- a/instrumentation/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
+++ b/instrumentation/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
@@ -34,7 +34,7 @@ class TracerTest extends AgentInstrumentationSpecification {
           name "test"
           kind PRODUCER
           hasNoParent()
-          errored true
+          status ERROR
           attributes {
             "string" "1"
             "long" 2
--- a/instrumentation/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
+++ b/instrumentation/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
@@ -34,7 +34,7 @@ class TracerTest extends AgentInstrumentationSpecification {
           name "test"
           kind PRODUCER
           hasNoParent()
-          status ERROR
+          errored true
           attributes {
             "string" "1"
             "long" 2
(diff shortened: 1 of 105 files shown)


anuraaga (approved):

trask:
Closing and re-opening to re-trigger CLA check

instrumentation/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
```diff
@@ -34,7 +34,7 @@ class TracerTest extends AgentInstrumentationSpecification {
           name "test"
           kind PRODUCER
           hasNoParent()
-          status ERROR
+          errored true
```
iNikem, line 37: Are you sure that we don't want to make a reverse change? Spec does not define `errored`, it defines `status` :)

trask: ya, that makes sense to me, will update the PR tomorrow

iNikem: @trask Do you plan to do this?

trask: this is done now, cleared existing reviews

anuraaga (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -154,18 +150,12 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
+   *   client.executeAsync(request) {*     callback.accept(it.statusCode())
+   *}*
    *   // Hypothetical client returning a CompletableFuture
-   *   client.executeAsync(request).thenAccept {
-   *     callback.accept(it.statusCode())
-   *   }
-   * }
-   *
+   *   client.executeAsync(request).thenAccept {*     callback.accept(it.statusCode())
+   *}*}*
```
mateuszrzeszutek, line 158: Please revert those Javadoc changes

mateuszrzeszutek (approved):

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
trask: oh no thx 👍

PR #2714, opened 2021-04-04
Add AsyncSpanEndStrategy for Reactor 3.x instrumentation

HaloFour:
Adds an implementation of `AsyncSpanEndStrategy` which supports Reactor publishers returned from methods annotated with the `@WithSpan` annotation.

### DIFF
--- a/instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ReactorHooksAdvice.java
+++ b/instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ReactorHooksAdvice.java
@@ -5,6 +5,8 @@
 
 package io.opentelemetry.javaagent.instrumentation.reactor;
 
+import io.opentelemetry.instrumentation.api.tracer.async.AsyncSpanEndStrategies;
+import io.opentelemetry.instrumentation.reactor.ReactorAsyncSpanEndStrategy;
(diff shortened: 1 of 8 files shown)


instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ReactorHooksAdvice.java
```diff
@@ -5,12 +5,15 @@
...
+import io.opentelemetry.instrumentation.api.tracer.async.AsyncSpanEndStrategies;
+import io.opentelemetry.instrumentation.reactor.ReactorAsyncSpanEndStrategy;
 import io.opentelemetry.instrumentation.reactor.TracingOperator;
 import net.bytebuddy.asm.Advice;
 
 public class ReactorHooksAdvice {
   @Advice.OnMethodExit(suppress = Throwable.class)
   public static void postStaticInitializer() {
     TracingOperator.registerOnEachOperator();
+    AsyncSpanEndStrategies.getInstance().registerStrategy(ReactorAsyncSpanEndStrategy.INSTANCE);
```
mateuszrzeszutek, line 17: How about adding this to `TracingOperator`? In the RxJava2 instrumentation the async span strategy is also usable by the library instrumentation.

mateuszrzeszutek (approved):
Thanks! That looks great 👍

trask (approved):
thx!

PR #2719, opened 2021-04-05
Fix build and documentation typos related to autoconfigure SPI.

jack-berg:
Was exploring using the autoconfiguration SPI to configure the agent by building an initializer jar and specifying it via `otel.initializer.jar`.

Found that there was a typo in the documentation as `otel.initializer.jar` has been changed to `otel.javaagent.experimental.initializer.jar`.

Also found that various build's are relocating the `io.opentelemetry.spi` package which no longer exists after some recent refactoring of the `opentelemetry-java` project to move the SPI stuff into the SDK autoconfiguration module. I figured the right thing to do was to remove those relocations since there I didn't notice any other attempt to relocate SDK packages.

### DIFF
--- a/docs/contributing/javaagent-test-infra.md
+++ b/docs/contributing/javaagent-test-infra.md
@@ -10,9 +10,9 @@ There are a few key components that make this possible, described below.
 * shades the instrumentation
 * adds jvm args to the test configuration
   * -javaagent:[agent for testing]
-  * -Dotel.initializer.jar=[shaded instrumentation jar]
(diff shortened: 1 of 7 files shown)


trask (approved):
thx!

it looks like that relocation is used in two more files, would you mind removing those too?

* testing/agent-exporter.gradle
* instrumentation/instrumentation.gradle (which is different from gradle/instrumentation.gradle)

jack-berg:
All set fixing those other locations you noted @trask.

anuraaga (approved):
Thanks!

Issue #579, opened 2020-06-25
Review RequestDispatcherAdvice


iNikem:
It is quite confusing at the moment.

First, it seems to do something similar to `Servlet3Advice` rescoping check, but in very different way. Have to check if they can be unified.

Second, it replaces `CONTEXT_ATTRIBUTE` in the current request. This seems to beat the original purpose of that attribute: to server as a marker that this request is already being taken care of. And/or to hold `SERVER` span.

trask:
@iNikem RequestDispatcherAdvice looks much better [now](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/87950d1d3286bc58cd3c665db52b2b4465942842/instrumentation/servlet/servlet-javax-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/javax/dispatcher/RequestDispatcherAdvice.java#L19) compared to [when you opened this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/2bcd35a251d70809bc213ac6c871f7079a34121b/instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/auto/instrumentation/servlet/dispatcher/RequestDispatcherInstrumentation.java#L88), can this be closed?

iNikem:
> @iNikem RequestDispatcherAdvice looks much better [now](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/87950d1d3286bc58cd3c665db52b2b4465942842/instrumentation/servlet/servlet-javax-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/javax/dispatcher/RequestDispatcherAdvice.java#L19) compared to [when you opened this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/2bcd35a251d70809bc213ac6c871f7079a34121b/instrumentation/servlet/servlet-common/src/main/java/io/opentelemetry/auto/instrumentation/servlet/dispatcher/RequestDispatcherInstrumentation.java#L88), can this be closed?

I don't think so. It still does some funny business with `CONTEXT_ATTRIBUTE`

trask:
> I don't think so. It still does some funny business with `CONTEXT_ATTRIBUTE`

ya, true, tried to make some progress on this in #2724, but still left a couple of TODOs in the code to come back to (will leave this issue open until those are addressed)

PR #2724, opened 2021-04-05
Clean up request dispatcher instrumentation

trask:
Progress toward #579

### DIFF
--- a/instrumentation/servlet/servlet-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v5_0/dispatcher/RequestDispatcherAdvice.java
+++ b/instrumentation/servlet/servlet-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v5_0/dispatcher/RequestDispatcherAdvice.java
@@ -10,6 +10,7 @@ import static io.opentelemetry.javaagent.instrumentation.servlet.v5_0.dispatcher
 
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
+import io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge;
 import io.opentelemetry.javaagent.instrumentation.servlet.common.dispatcher.RequestDispatcherAdviceHelper;
 import jakarta.servlet.RequestDispatcher;
 import jakarta.servlet.ServletRequest;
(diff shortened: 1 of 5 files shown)


instrumentation/servlet/servlet-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v5_0/dispatcher/RequestDispatcherAdvice.java
```diff
@@ -22,48 +23,44 @@
...
+    Object requestContextAttr = request.getAttribute(CONTEXT_ATTRIBUTE);
+    requestContext = requestContextAttr instanceof Context ? (Context) requestContextAttr : null;
+
+    Context parentContext =
+        RequestDispatcherAdviceHelper.getStartParentContext(currentContext, requestContext);
+    if (parentContext == null) {
       return;
     }
 
-    try (Scope ignored = parent.makeCurrent()) {
```
anuraaga, line 38: This is one reason I made the decision to require `Context` as an argument to `start` in the new `Instrumenter` API. I wasn't too sure of that, since it makes writing instrumentation one step harder, but looking at this chance of messiness it seems like a good idea.

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/common/dispatcher/RequestDispatcherAdviceHelper.java
```diff
@@ -15,37 +15,37 @@
...
+  //  this could be simplified by always using currentContext, only falling back to requestContext
+  //  if currentContext does not have a valid span
+  public static Context getStartParentContext(Context currentContext, Context requestContext) {
+    Span currentSpan = Span.fromContext(currentContext);
+    SpanContext currentSpanContext = currentSpan.getSpanContext();
 
-    Context servletContext =
-        servletContextObject instanceof Context ? (Context) servletContextObject : null;
+    if (!currentSpanContext.isValid()) {
+      // this may be null
```
anuraaga, line 31: Add `@Nullable` instead? :)

anuraaga (approved):

iNikem (approved):
New code does look cleaner, but I still don't quite understand if it is needed. I mean `io.opentelemetry.instrumentation.servlet.ServletHttpServerTracer#attachServerContext` sets that attribute for all servlet requests without any regard for the old value. Why should we preserve old value here?

mateuszrzeszutek (approved):

laurit:
As far as I understand when we have only servlet instrumentation then dispatch can happen after server span has ended (tomcat forwarding to error page) then span from context isn't valid (invalid span is returned when there is no span) but request still has the context attached in `ServletHttpServerTracer#attachServerContext` which contains server span (this span has already ended though). To avoid creating a new trace this sets the ended server span as parent for the work done in dispatch.
I suspect that changing the request attribute isn't really necessary. I guess previously server span was always created by servlet integration and placed to request attribute but now there are server specific instrumentations that create server spans and they don't always carry context in the request attribute so anything that relies on context being set in request attribute is probably broken.

iNikem:
Yes, I understand _reading_ context from request attribute and using that as a parent. I am suspicious about _writing_ new context into the request and even more suspicious about _restoring_ old context. May be that is needed, but it is not obvious at all and should be either removed or documented.

trask:
> New code does look cleaner

👍 will go ahead and merge

> but I still don't quite understand if it is needed

will continue to work on making this clearer (or removing it) as part of #579

PR #2727, opened 2021-04-06
Don't create duplicate headers

### LINKED ISSUE #2594: HttpHeadersInjectAdapter should use 'setHeader' instead of 'addHeader'?
**Describe the bug**
If reuse http request, traceparent header will be more and more when use apache-httpclient4.x and 5.x

image

v1.0.1

trask:
Resolves #2594

Adds a new test to `HttpClientTest` that create a request, sends the request, and then reuses that request (sending it a second time).

Most of the http client libraries support this kind of reuse, and several of our instrumentations would end up adding the `traceparent` header multiple times (#2594).

`HttpTestServer` now throws an exception if there is more than one `traceparent` header.

To implement the new test, the `doRequest()` method that was implemented in each subclass of `HttpClientTest` has been split into two methods:
* `buildRequest()`
* `sendRequest()`

The normal `doRequest` (now final) is implemented as `buildRequest` + `sendRequest`, while the new `doReusedRequest` (also final) is implemented as `buildRequest` + `sendRequest` + `sendRequest`.


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/server/http/TestHttpServer.groovy
```diff
@@ -249,6 +249,10 @@ class TestHttpServer implements AutoCloseable {
         isTestServer = Boolean.parseBoolean(request.getHeader("is-test-server"))
       }
       if (isTestServer) {
+        def headers = req.getHeaders("traceparent")
```
mateuszrzeszutek, line 252: Should we use `GlobalOpenTelemetry.getPropagators().getTextMapPropagator().fields()` instead of hardcoding `traceparent`?

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -104,18 +129,18 @@ abstract class HttpClientTest extends InstrumentationSpecification {
...
    * @Override
-   * int doRequest(String method, URI uri, Map<String, String headers = [:]) {
+   * int sendRequest(REQUEST request, String method, URI uri, Map<String, String> headers) {
    *   CompletableFuture<Integer> result = new CompletableFuture<>()
    *   doRequestWithCallback(method, uri, headers) {
    *     result.complete(it.statusCode())
    *   }
    *   return result.join()
    * }
    */
```
mateuszrzeszutek, line 139: Can you update the whole Javadoc so that is does not mention `new Request()` or `doRequest()`?

Also, would it make sense to refactor `doRequestWithCallback()` in a similar way to `doRequest()`? Make it final and split into `buildRequest()` + `sendRequestWithCallback()`

instrumentation/akka-http-10.0/javaagent/src/test/groovy/AkkaHttpClientInstrumentationTest.groovy
```diff
@@ -9,43 +9,45 @@ import akka.actor.ActorSystem
...
+    return Http.get(system)
+      .singleRequest(request, materializer)
+      .toCompletableFuture()
+      .get()
+      .status()
+      .intValue()
   }
 
   @Override
   void doRequestWithCallback(String method, URI uri, Map<String, String> headers = [:], Consumer<Integer> callback) {
```
anuraaga, line 44: Can we pass request into this one too?

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/server/http/TestHttpServer.groovy
trask: oh good idea!

instrumentation/http-url-connection/javaagent/src/test/groovy/SpringRestTemplateTest.groovy
```diff
@@ -28,22 +27,28 @@ class SpringRestTemplateTest extends HttpClientTest implements AgentTestTrait {
   }
 
   @Override
-  int doRequest(String method, URI uri, Map<String, String> headers = [:]) {
+  Void buildRequest(String method, URI uri, Map<String, String> headers) {
+    return null
```
anuraaga, line 31: Should we set `testReusedRequest` to false to be explicit about it not being tested (I guess it's not being tested)?

anuraaga (approved):

instrumentation/http-url-connection/javaagent/src/test/groovy/SpringRestTemplateTest.groovy
trask: pushed a new commit that seems a bit better, lmk

mateuszrzeszutek (approved):

PR #2736, opened 2021-04-06
Exclude our Runnables

laurit:
Skip attaching state to Runnables loaded by agent class loader.


iNikem (approved):

mateuszrzeszutek (approved):

kubawach (approved):

trask (approved):
Can you add a comment about why this is not covered by `GlobalClassloaderIgnoresMatcher`? (mostly so someone reading it won't wonder why we don't do this everywhere)

PR #2737, opened 2021-04-06
Improve Akka instrumentation

laurit:
Attach context to Akka message and restore it before invoking target method to ensure it runs with correct parent context.

### DIFF
--- a/instrumentation/akka-actor-2.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkaactor/AkkaActorInstrumentationModule.java
+++ b/instrumentation/akka-actor-2.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkaactor/AkkaActorInstrumentationModule.java
@@ -25,7 +25,11 @@ public class AkkaActorInstrumentationModule extends InstrumentationModule {
 
   @Override
   public List<TypeInstrumentation> typeInstrumentations() {
-    return asList(new AkkaForkJoinPoolInstrumentation(), new AkkaForkJoinTaskInstrumentation());
+    return asList(
+        new AkkaForkJoinPoolInstrumentation(),
+        new AkkaForkJoinTaskInstrumentation(),
+        new AkkaDispatcherInstrumentation(),
+        new AkkaActorCellInstrumentation());
   }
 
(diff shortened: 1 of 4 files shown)


instrumentation/akka-actor-2.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkaactor/AkkaActorInstrumentationModule.java
```diff
@@ -34,6 +38,7 @@ public AkkaActorInstrumentationModule() {
     map.put(Runnable.class.getName(), State.class.getName());
     map.put(Callable.class.getName(), State.class.getName());
     map.put(AkkaForkJoinTaskInstrumentation.TASK_CLASS_NAME, State.class.getName());
+    map.put("akka.dispatch.Envelope", State.class.getName());
```
kubawach, line 41: Why not "Envelope.class.getName()" ? AFAIK the class is available for reference at this point. Not that it changes a lot ;)

kubawach (approved):

instrumentation/akka-actor-2.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkaactor/AkkaActorInstrumentationModule.java
laurit: Actually the class isn't available. `InstrumentationModule` instances are loaded by `AgentClassLoader` so they can only see what jdk has and the things that are bundled inside agent.

mateuszrzeszutek (approved):

instrumentation/akka-actor-2.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkaactor/AkkaActorInstrumentationModule.java
kubawach: That explains a lot ;)

trask (approved):

Issue #2713, opened 2021-04-03
Convert all instrumentation to the new Instrumenter API 


trask:
Tracking/discussion issue for converting new instrumentation to the Instrumenter API defined in #2596.

Please add your name next to any that you are working on to avoid duplication. If you don't ability to update this comment, please comment below and someone will update it (you don't need to wait until someone updates the comment to work on it).

- [x] akka-http-10.0 #4359
- [ ] apache-camel-2.20
- [x] apache-dubbo/apache-dubbo-2.7 #4059
- [x] apache-httpasyncclient-4.1 #3490
- [x] apache-httpclient/apache-httpclient-2.0 #3012
- [x] apache-httpclient/apache-httpclient-4.0 #3042
- [x] apache-httpclient/apache-httpclient-5.0 #3062
- [x] armeria-1.3 #2596
- [x] async-http-client/async-http-client-1.9 #3794
- [x] async-http-client/async-http-client-2.0 #3211
- [x] aws-lambda-1.0 #4362
- [x] aws-sdk/aws-sdk-1.11 #4315
- [x] aws-sdk/aws-sdk-2.2 #4341
- [x] cassandra/cassandra-3.0 #3066
- [x] cassandra/cassandra-4.0 #3126
- [x] couchbase/couchbase-2.0 #3802
- [x] dropwizard-views-0.7 #3784
- [x] elasticsearch/elasticsearch-rest-common #4153
- [x] elasticsearch/elasticsearch-transport-common #4252
- [x] external-annotations #3738 #4578
- [x] finatra-2.9 #4495 
- [x] geode-1.4 #3067
- [x] google-http-client-1.19 #3932
- [x] grails-3.0 #3795
- [x] grizzly-2.0 #4409
- [x] grpc-1.6 #3073
- [x] gwt-2.0 #3146
- [x] hibernate/hibernate-common #4369
- [x] http-url-connection #4044
- [x] hystrix-1.4 #3802
- [x] java-http-client #3790
- [x] jaxrs-client/jaxrs-client-1.1 #3934
- [x] jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common #3938
- [x] jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-resteasy-3.0 #3938
- [x] jaxrs/jaxrs-1.0 #3797
- [x] jaxrs/jaxrs-2.0/jaxrs-2.0-common #3796 
[...]

PR #2739, opened 2021-04-06
Use Instrumenter in JDBC instrumentation

mateuszrzeszutek:
Except jdbc-datasource, that one is going to be in a separate PR


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/db/DbSemanticConventions.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.db;
+
+import io.opentelemetry.instrumentation.api.instrumenter.SpanNameExtractor;
+
+public final class DbSemanticConventions {
```
mateuszrzeszutek, line 10: I put all DB-specific classes in a subpackage - let's evaluate whether this is a good idea

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/DbRequest.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.google.auto.value.AutoValue;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.Statement;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+@AutoValue
+public abstract class DbRequest {
+  public abstract DbInfo getDbInfo();
```
anuraaga, line 19: Do we use DbInfo to reduce the scope of this PR? I guess it could be inlined in

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/JdbcNetAttributesExtractor.java
```diff
@@ -0,0 +1,28 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+final class JdbcNetAttributesExtractor
+    extends InetSocketAddressNetAttributesExtractor<DbRequest, Void> {
+  @Override
+  protected @Nullable InetSocketAddress getAddress(DbRequest dbRequest, Void unused) {
+    String host = dbRequest.getDbInfo().getHost();
+    Integer port = dbRequest.getDbInfo().getPort();
+    if (host != null && port != null) {
+      return new InetSocketAddress(host, port);
```
anuraaga, line 19: We should either directly implement `NetAttributesExtractor` or at least create unresolved, we don't want to do DNS lookup to populate an attribute

anuraaga (commented):
Thanks looks pretty nice

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/db/DbSemanticConventions.java
```diff
@@ -0,0 +1,42 @@
...
+   * {@code <db.operation> <db.name><table>}. If {@code dbName} and {@code operation} are not
+   * provided then {@code defaultSpanNameExtractor} is used to compute the span name.
+   *
+   * @see DbAttributesExtractor#dbOperation(Object) used to extract {@code <db.operation>}.
+   * @see DbAttributesExtractor#dbName(Object) used to extract {@code <db.name>}.
+   * @see SqlAttributesExtractor#dbTable(Object) used to extract {@code <db.table>}.
+   */
+  public static <REQUEST> SpanNameExtractor<REQUEST> spanName(
+      DbAttributesExtractor<REQUEST> attributesExtractor,
+      SpanNameExtractor<REQUEST> defaultSpanNameExtractor) {
```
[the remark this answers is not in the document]
mateuszrzeszutek: I think that Cassandra uses `db.statement` as the default - maybe it can be avoided, but let's revisit this topic once we get to the Cassandra instrumentation. I'll remove this option in the meantime - let's keep this as simple as possible.

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/DbRequest.java
mateuszrzeszutek: I thought about inlining it first, but I noticed that it's being cached per `Connection`: see `JdbcMaps.connectionInfo`. I thought it would be simpler to just create another small class for request and reuse `DbInfo` instances.

anuraaga (approved):
Thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -110,6 +110,15 @@
     return newInstrumenter(InstrumenterConstructor.internal(), SpanKindExtractor.alwaysInternal());
   }
 
+  /**
+   * Returns a new {@link Instrumenter} which will create spans with kind determined by the passed
+   * {@code spanKindExtractor} and do no context propagation.
+   */
+  public Instrumenter<REQUEST, RESPONSE> newInstrumenter(
```
jkwatson, line 117: would `INPUTS, OUTPUTS` make more sense here, rather than request/response?

mateuszrzeszutek: It would, especially for non-HTTP instrumenters. @anuraaga WDYT?

anuraaga: I feel like request / response are fairly standard terms even in a general sense, even for DBs for example. INPUTS OUTPUTS just seems a bit too generic, I think the difference vs request / response is the latter implies being "remote" which is what we care about when instrumenting.

mateuszrzeszutek: Hmm, let's try a few more conventions first (RPC, messaging) - we can always revisit this topic a bit later.
In the meantime I'm going to merge this PR and take on the next semconv attribute group.

PR #2745, opened 2021-04-07
Build windows collector test image with otel-collector-contrib

mateuszrzeszutek:
Vendors will be able to reuse this image in their tests that depend on contrib components

### DIFF
--- a/smoke-tests/fake-backend/build.gradle
+++ b/smoke-tests/fake-backend/build.gradle
@@ -69,7 +69,7 @@ task windowsCollectorBinaryDownload(type: Download) {
     collectorDockerBuildDir.mkdirs()
   }
 
-  src("https://github.com/open-telemetry/opentelemetry-collector/releases/latest/download/otelcol_windows_amd64.exe")
+  src("https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/latest/download/otelcontribcol_windows_amd64.exe")
   dest(collectorDockerBuildDir)
 }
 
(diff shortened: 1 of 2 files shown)


iNikem (approved):

trask (approved):

anuraaga (commented):
Hmm - can we split out a dockerfile folder, otel-collector-windows or something? I guess this is just a windows image we want because there isn't any official one so it's confusing to tie this to fake backend. You wouldn't need gradle at all I think if just switching to `ADD https://github.com/open-telemetry/opentelemetry-collector/releases/latest/download/otelcol_windows_amd64.exe /otelcol_windows_amd64.exe`

anuraaga:
Though I guess even better would be to moving it to the `collector-contrib` repo, don't see why they'd be against having a windows image there

iNikem:
@anuraaga This PR just changes Collector for Collector Contrib in already existing procedure of building Collector Windows image. But there is https://github.com/open-telemetry/opentelemetry-collector/issues/2715, so in the future we definitely want to remove this from our repo.

anuraaga:
@iNikem Yeah - I don't mind us doing something temporary but still it'd be nice to remove this from the scope of "fake backend"? For example, I'm very confused that we would be changing this file but not changing the collector we use for Linux smoke tests

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/smoke-tests/src/test/java/io/opentelemetry/smoketest/LinuxTestContainerManager.java#L46

It sounds like this dockerfile has a different intent than our smoke tests :P

iNikem:
You first have to change image generation, merge, let GHA to build new image and then you can change the test to actually use new image...

anuraaga:
Ah so we will update both? Then it makes more sense - title mentioned "Windows smoke tests" and PR description mentions vendors (so not us?), so didn't know what the intent of this PR is :)

iNikem:
That is a good point :) This repo does not actually need contrib...

mateuszrzeszutek:
Hmm, that's true - let's close it PR then, I think I have a better idea what do do in our own distro. Thanks!

Issue #2123, opened 2021-01-26
Record internal metric for SQL cache misses.


breedx-splk:
In #2094 we introduced a cache for SQL sanitization/extraction to reduce the number of times that a familiar statement would need to undergo expensive parsing.  This was done in order to help reduce the overhead of the very common jdbc instrumentation.

It is expected that most common applications and frameworks will have a fixed/finite set of SQL statements used at runtime. There will be dynamically generated SQL statements that happen for some users, but we expect these exceptions to be rare.  Currently, when a SQL statement is not found in the cache and additional parsing has to happen, we log a message at `trace`, but this is likely to be unseen.

Let's create and increment an internal metric for parsed SQL cache misses.  This will allow users and operators troubleshooting jdbc instrumentation to determine how effective the cache is.  It could also allow tuning of this cache (size, lifetime, etc) for future users.

Before starting on this, we should agree what to call it and what its namespace should be.

mateuszrzeszutek:
We could use `SupportabilityMetrics` for this

PR #2747, opened 2021-04-07
Record internal metric for SQL cache misses

mateuszrzeszutek:
And use `SupportabilityMetrics` in `Instrumenter`

Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2123

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/db/SqlStatementSanitizer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/db/SqlStatementSanitizer.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.instrumentation.api.db;
 
 import static io.opentelemetry.instrumentation.api.db.StatementSanitizationConfig.isStatementSanitizationEnabled;
+import static io.opentelemetry.instrumentation.api.internal.SupportabilityMetrics.CounterNames.SQL_STATEMENT_SANITIZER_CACHE_MISS;
 
 import io.opentelemetry.instrumentation.api.caching.Cache;
 import io.opentelemetry.instrumentation.api.internal.SupportabilityMetrics;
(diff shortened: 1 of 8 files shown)


iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/db/SqlStatementSanitizer.java
```diff
@@ -28,7 +27,7 @@ public static SqlStatementInfo sanitize(String statement) {
     return sqlToStatementInfoCache.computeIfAbsent(
         statement,
         k -> {
-          log.trace("SQL statement cache miss");
+          supportability.incrementCounter("SqlStatementSanitizer cache miss");
```
jkwatson, line 30: It would be nice to have these counter names be constants in the SupportabilityMetrics class, so it's easy to see what the various counters are. It won't stop someone from adding custom counter names, but it will at least add a little bit to the self-documenting nature of the code.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/internal/SupportabilityMetricsTest.java
```diff
@@ -66,14 +73,17 @@ void resetsCountsEachReport() {
...
     metrics.recordSuppressedSpan(SpanKind.CLIENT, "favoriteInstrumentation");
+    metrics.incrementCounter("some counter");
 
     metrics.report();
     metrics.report();
 
     assertThat(reports)
         .isNotEmpty()
-        .hasSize(1)
+        .hasSize(2)
```
anuraaga, line 83: Just noticed now so no big deal, but we generally wouldn't call `isNotEmpty` just before a `hasSize` check

anuraaga (approved):

PR #2749, opened 2021-04-07
Implement concurrent test for Vert.x http server

iNikem:


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -393,7 +393,7 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
   def "high concurrency test"() {
     setup:
     assumeTrue(testConcurrency())
-    int count = 10
+    int count = 100
```
trask, line 396: 🚀

trask (approved):

PR #2753, opened 2021-04-08
Make dubbo and camel use loopback address

laurit:
Dubbo and camel use a random local ip address, sometimes connecting to it seems to hang. Perhaps these addresses are something that docker uses? idk. Hopefully using loopback address will reduce flakiness of these tests.

### DIFF
--- a/instrumentation/apache-camel-2.20/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/TwoServicesConfig.groovy
+++ b/instrumentation/apache-camel-2.20/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/TwoServicesConfig.groovy
@@ -28,7 +28,7 @@ class TwoServicesConfig {
           .log("Service One request: \${body}")
           .delay(simple("\${random(1000,2000)}"))
           .transform(simple("Service-One-\${body}"))
-          .to("http://0.0.0.0:{{service.two.port}}/serviceTwo")
+          .to("http://127.0.0.1:{{service.two.port}}/serviceTwo")
           .log("Service One response: \${body}")
       }
     }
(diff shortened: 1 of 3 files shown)


anuraaga (approved):

mateuszrzeszutek (approved):

kubawach (commented):
If I understood correctly, in Camel context the goal here is to avoid HTTP servers listening on random addresses. That's a great idea, but needs all "from" parts of configured routes to use the loopback instead of "0.0.0.0" ("from" sets up servers in Camel). So besides the config already included in the changeset, we'd need to include also SingleServiceConfig.

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/TwoServicesConfig.groovy
```diff
@@ -28,7 +28,7 @@ class TwoServicesConfig {
           .log("Service One request: \${body}")
           .delay(simple("\${random(1000,2000)}"))
           .transform(simple("Service-One-\${body}"))
-          .to("http://0.0.0.0:{{service.two.port}}/serviceTwo")
+          .to("http://127.0.0.1:{{service.two.port}}/serviceTwo")
```
kubawach, line 31: This should be reflected also in "from" part of the serviceTwoRoute().

laurit:
@kubawach it is not about listening but rather calling the service with a specific address. As the test passes I believe that changing the from part is not necessary `0.0.0.0` should match any local address. As this seems to be the only test that is occasionally failing I left other tests as they are.

kubawach:
> @kubawach it is not about listening but rather calling the service with a specific address. As the test passes I believe that changing the from part is not necessary `0.0.0.0` should match any local address. As this seems to be the only test that is occasionally failing I left other tests as they are.

Yeah, so for Camel "0.0.0.0" means "bind to all local addresses", therefore my idea was that since we change "call any" to "call loopback" it would make sense to also "bind to loopback only" ;)

trask:
@kubawach sorry I wasn't sure if your question was fully resolved, but I wanted to go ahead and merge this to hopefully help with the flakiness

PR #2755, opened 2021-04-08
resolved TODO - instrumentation names in suppressing doc

### LINKED ISSUE #2256: Address TODO in suppressing-instrumentation.md
In [suppressing-instrumentation.md](docs/suppressing-instrumentation.md) there is a `TODO` that says 

```
[TODO add table here with all instrumentation ids]
```

It's a good idea to have a reference for users to easily crib (copy) from in order to disable their least favorite instrumentation.  🙃 This does seem like a maintenance burden over time, so I'm wondering if there isn't a good way to automate that.

At worst case, we could expand the TODO to show a couple examples and then describe to the reader how to find the id of the thing they are looking for.

kubawach:
Resolves #2256

### DIFF
--- a/docs/suppressing-instrumentation.md
+++ b/docs/suppressing-instrumentation.md
@@ -75,7 +75,7 @@ You can suppress agent instrumentation of specific libraries by using
 | Play WS HTTP Client | play-ws|
 | RabbitMQ Client | rabbitmq|
 | Ratpack | ratpack|
-| ReactiveX RxJava | rxjava2 |
+| ReactiveX RxJava | rxjava2, rxjava3 |
 | Reactor | reactor|
 | Reactor Netty | reactor-netty|
 | Redis Jedis | jedis|
--- a/docs/suppressing-instrumentation.md
+++ b/docs/suppressing-instrumentation.md
@@ -50,8 +50,8 @@ You can suppress agent instrumentation of specific libraries by using
 | Java RMI | rmi|
 | Java Servlet | servlet|
 | java.util.concurrent | executor |
-| JAX-RS | jaxrs|
-| JAX-RS Client | jaxrs-client|
+| JAX-RS (Client) | jaxrs-client|
+| JAX-RS (Server) | jaxrs|
 | JAX-WS | jaxws|
 | JAX-WS Metro | metro|
 | Jetty | jetty|
--- a/docs/suppressing-instrumentation.md
+++ b/docs/suppressing-instrumentation.md
@@ -10,77 +10,79 @@ You can suppress agent instrumentation of specific libraries by using
 
 | Library/Framework | Instrumentation name |
 |-------------------|----------------------|
+| Additional methods tracing | methods |
+| Additional tracing annotations | external-annotations |
 | Akka Actor | akka-actor|
 | Akka HTTP | akka-http|
+| Apache Axis2 | axis2|
 | Apache Camel | apache-camel|
+| Apache Cassandra | cassandra|
+| Apache CXF | cxf|
 | Apache Dubbo | apache-dubbo|
+| Apache Geode | geode|
 | Apache HttpAsyncClient | apache-httpasyncclient|
 | Apache HttpClient | apache-httpclient|
+| Apache Kafka | kafka |
+| Apache RocketMQ | rocketmq-client|
+| Apache Tapestry | tapestry|
+| Apache Tomcat | tomcat|
+| Apache Wicket | wicket|
 | Armeria | armeria|
 | AsyncHttpClient (AHC) | async-http-client|
 | AWS Lambda | aws-lambda|
 | AWS SDK | aws-sdk|
-| Apache Axis2 | axis2|
-| Apache Cassandra | cassandra|
 | Couchbase | couchbase|
-| Apache CXF | cxf|
 | Dropwizard Views | dropwizard-views |
 | Eclipse OSGi | eclipse-osgi |
-| Elasticsearch REST client | elasticsearch-rest|
 | Elasticsearch client | elasticsearch-transport|
-| java.util.concurrent | executor |
-| Additional tracing annotations | external-annotations |
-| Twitter Finatra | finatra|
-| Apache Geode | geode|
-| Google HTTP client | google-http-client|
-| Grails | grails|
-| Java EE Grizzly | grizzly|
-| GRPC | grpc|
+| Elasticsearch REST client | elasticsearch-rest|
 | Google Guava | guava|
+| Google HTTP client | google-http-client|
 | Google Web Toolkit | gwt|
+| Grails | grails|
+| GRPC | grpc|
 | Hibernate | hibernate|
-| Java HTTP URL connection | http-url-connection |
+| Java EE Grizzly | grizzly|
 | Java HTTP Client | httpclient |
-| Netflix Hystrix | hystrix|
-| JAX-RS Clieny | jaxrs-client|
-| JAX-RS | jaxrs|
-| JAX-WS | jaxws|
+| Java `HttpURLConnection` | http-url-connection |
 | Java JDBC | jdbc |
-| Java JDBC data source | jdbc-datasource |
-| Redis Jedis | jedis|
+| Java JDBC `DataSource` | jdbc-datasource |
+| Java RMI | rmi|
+| Java Servlet | servlet|
+| java.util.concurrent | executor |
+| JAX-RS | jaxrs|
+| JAX-RS Client | jaxrs-client|
+| JAX-WS | jaxws|
+| JAX-WS Metro | metro|
 | Jetty | jetty|
 | JMS | jms|
+| JSF Mojarra | mojarra|
+| JSF MyFaces | myfaces|
 | JSP | jsp |
-| Apache Kafka | kafka |
+| K8s Client | kubernetes-client|
 | Kotlin HTTP (kHttp) | khttp |
 | kotlinx.coroutines | kotlinx-coroutines |
-| K8s Client | kubernetes-client|
-| Lettuce (Redis) | lettuce|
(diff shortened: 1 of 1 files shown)


docs/suppressing-instrumentation.md
```diff
@@ -6,11 +6,152 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
...
+| aws-lambda|aws-lambda-1.0 |
+| aws-sdk|aws-sdk-1.11 |
+| aws-sdk|aws-sdk-2.2 |
+| axis2|axis2-1.6 |
+| cassandra|cassandra-3.0 |
+| cassandra|cassandra-4.0 |
+| classloader |
+| couchbase|couchbase-2.0 |
+| couchbase|couchbase-2.6 |
+| couchbase|couchbase-3.1 |
```
trask, line 35: I'd suggest that we swap these, so that the "Main instrumentation name" is unique and matches the instrumentation module name, while the additional names have the shortcuts / version-agnostic names

kubawach: I followed what we have in the code (main vs additional). It would definitely make sense to have the main as unique, but probably better to change it in the code and then reflect in the docs. What do you think?

trask: 👍

```
public InstrumentationModule(
      String mainInstrumentationName, String... additionalInstrumentationNames) {
```

==>

```
public InstrumentationModule(String moduleName, String... additionalNames) {
```

where `moduleName` is the same as the gradle module name except when there are multiple InstrumentationModules in a single gradle module

mateuszrzeszutek: > where `moduleName` is the same as the gradle module name except when there are multiple InstrumentationModules in a single gradle module

Hmm, if we follow this won't we end up with main module names like `couchbase-2.0`? In that case whenever somebody wants to turn off a particular instrumentation they'll need to know 1) what version of the instrumented library they're using, and 2) what instrumentations it maps to (taking couchbase as an example: version 2.8 actually uses both 2.0 and 2.6 instrumentations).
I think I'd rather get rid of those versioned module names completely (with several exceptions of course; for example the `couchbase-2.6` module could use names `"couchbase", "couchbase-networking"`, with every other couchbase module using just `"couchbase"`)

kubawach: That's a very interesting discussion (unique vs generic) but I think we should move it to a separate issue. Naming is important from both disabling and usage (checking in logs what actually happens) points of view. Deciding this here (documentation task) would not seem the right thing to do :)

trask: I asked about this in the SIG meeting yesterday with @iNikem and @anuraaga, and consensus was to list the unversioned module names in this suppressing doc, and we can mention in troubleshooting about disabling a specific instrumentation module using the full module name, but that will just be a troubleshooting feature (e.g. for troubleshooting if muzzle is not doing its job) and we won't list all those versioned names in that doc. (and I think this means we don't need to change anything in the code)

kubawach: Updated :)

docs/suppressing-instrumentation.md
```diff
@@ -6,11 +6,100 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
...
+| apache-dubbo|
+| apache-httpasyncclient|
+| apache-httpclient|
+| armeria|
+| async-http-client|
+| aws-lambda|
+| aws-sdk|
+| axis2|
+| cassandra|
+| classloader |
```
mateuszrzeszutek, line 26: This module was recently renamed to `internal` (and it probably shouldn't be on this list)
```suggestion
```

kubawach: I see three names: `internal-class-loader`, `internal-proxy` and `internal-url-class-loader`. Why do you think these should not be on the list (ie user should not be able to disable them)?

iNikem: Yes, they are required for agent, and should not be disabled.

mateuszrzeszutek: Because without them the javaagent won't function properly (e.g. `helperResourceNames()` won't get injected).

kubawach: Sounds right :)

docs/suppressing-instrumentation.md
```diff
@@ -6,11 +6,99 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
 ## Suppressing specific agent instrumentation
 
 You can suppress agent instrumentation of specific libraries by using
-`-Dotel.instrumentation.[id].enabled=false`.
+`-Dotel.instrumentation.[name].enabled=false` where `name` is the instrumentation `name`
+(main or additional):
```
trask, line 10:
```suggestion
`-Dotel.instrumentation.[name].enabled=false` where `name` is the corresponding instrumentation name below:
```

docs/suppressing-instrumentation.md
```diff
@@ -6,11 +6,99 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
...
 
-where `id` is the instrumentation `id`:
-
-[TODO add table here with all instrumentation ids]
+| Instrumentation name |
+|---------------------------|
+| akka-actor|
```
trask, line 14: Single column table looks odd.

Maybe bullets?

Or maybe add "Library/Framework" column (matching https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/supported-libraries.md#libraries--frameworks), e.g.


```suggestion
| Library/Framework | Instrumentation name |
|-------------|--------------|
| Akka HTTP | akka-actor |
```

iNikem: +1 for library name

docs/suppressing-instrumentation.md
```diff
@@ -6,99 +6,96 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
...
+| Twitter Finatra | finatra|
+| Apache Geode | geode|
+| Google HTTP client | google-http-client|
+| Grails | grails|
+| Java EE Grizzly | grizzly|
+| GRPC | grpc|
+| Google Guava | guava|
+| Google Web Toolkit | gwt|
+| Hibernate | hibernate|
+| Java HTTP URL connection | http-url-connection |
```
mateuszrzeszutek, line 42: Consider just using the class name:
```suggestion
| Java `HttpURLConnection` | http-url-connection |
```

docs/suppressing-instrumentation.md
```diff
@@ -6,99 +6,96 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
...
+| Java HTTP Client | httpclient |
+| Netflix Hystrix | hystrix|
+| JAX-RS Clieny | jaxrs-client|
+| JAX-RS | jaxrs|
+| JAX-WS | jaxws|
+| Java JDBC | jdbc |
+| Java JDBC data source | jdbc-datasource |
```
mateuszrzeszutek, line 49:
```suggestion
| Java JDBC `DataSource` | jdbc-datasource |
```

docs/suppressing-instrumentation.md
```diff
@@ -6,99 +6,96 @@ You can disable the agent using `-Dotel.javaagent.enabled=false`
...
+| Ratpack | ratpack|
+| Reactor Netty | reactor-netty|
+| Reactor | reactor|
+| Rediscala | rediscala|
+| Java RMI | rmi|
+| Apache RocketMQ | rocketmq-client|
+| ReactiveX RxJava | rxjava2 |
+| Scala executors | scala-executors |
+| Java Servlet | servlet|
+| Apache Spark | spark|
```
mateuszrzeszutek, line 83: Actually it's not Apache Spark (big data processing), but [Spark Web Framework](http://sparkjava.com/)

```suggestion
| Spark Web Framework | spark|
```

docs/suppressing-instrumentation.md
kubawach: I find class name under "Library/Framework" confusing.

docs/suppressing-instrumentation.md
kubawach: I find class name under "Library/Framework" confusing.

docs/suppressing-instrumentation.md
trask: hm, tricky, I find the class name provides helpful specificity

docs/suppressing-instrumentation.md
trask: same here

trask (approved):
I think it would look better alphabetized by the first column

kubawach:
> I think it would look better alphabetized by the first column

Done ;)

docs/suppressing-instrumentation.md
```diff
@@ -10,76 +10,78 @@ You can suppress agent instrumentation of specific libraries by using
...
+| OpenLiberty | liberty |
 | OpenTelemetry Trace annotations | opentelemetry-annotations |
 | OSHI (Operating System and Hardware Information) | oshi |
-| Play WS HTTP Client | play-ws|
 | Play Framework | play|
+| Play WS HTTP Client | play-ws|
 | RabbitMQ Client | rabbitmq|
 | Ratpack | ratpack|
-| Reactor Netty | reactor-netty|
+| ReactiveX RxJava | rxjava2 |
```
mateuszrzeszutek, line 78: Oh, I almost forgot about it: recently RxJava3 instrumentation has been added:
```suggestion
| ReactiveX RxJava | rxjava2, rxjava3 |
```

mateuszrzeszutek (approved):

trask:
thx @kubawach!

kubawach:
> thx @kubawach!

It was longer march than I expected ;)

Issue #2644, opened 2021-03-26
Agent instrumentation fails on sybase jdbc instrumentation


pparapatics:
**Describe the bug**

Starting from version 0.17.0 the agent instrumentation fails on a tomcat 8 application
Error message:
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message transform method call failed at ./src/java.instrument/share/native/libinstrument/JPLISAgent.c line: 873

Versions failing: 0.17.0, 1.0.0, 1.0.1
Versions working: 0.10.1

**Steps to reproduce**
Start tomcat 8 with javagent 

**What did you expect to see?**
For version 0.10.1 the application starts up, agent works as expected (propagation, exporting, ...)

**What did you see instead?**
Version 0.17.0 fails with message above 
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message transform method call failed at ./src/java.instrument/share/native/libinstrument/JPLISAgent.c line: 873
After that no application logs are show anymore
Tomcat does not start up correctly

**What version are you using?**
Versions failing: 0.17.0, 1.0.0, 1.0.1
Versions working: 0.10.1

**Environment**
Tomcat version: Apache Tomcat/8.5.32

**Additional context**
The application running in tomcat is a quite complex monolithical application.

iNikem:
@pparapatics  What OS do you use? What version of JVM and from what vendor?

pparapatics:
Tried two setups:

1)
JVM:
openjdk version "11.0.8" 2020-07-14
OpenJDK Runtime Environment 18.9 (build 11.0.8+10)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.8+10, mixed mode)

Operating system:
NAME="Linux Mint"
VERSION="20 (Ulyana)"

2)
CentOS 8 with
openjdk 11.0.10 2021-01-19 LTS
OpenJDK Runtime Environment 18.9 (build 11.0.10+9-LTS)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.10+9-LTS, mixed mode, sharing)

trask:
@pparapatics can you post all the JVM command-line options you are using (including any other `-javaagent` flags if you have any)

also, is there a stack trace associated with the error you posted above?

pparapatics:
Sorry for the late reply. 
Below you find the parameters extracted from the logs. Unfortunately there is no stack trace for the error above. 
[...]

PR #2756, opened 2021-04-08
Fix JDBC instrumentation: recursive statements inside Connection#getMetaData

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2644

The recursive Statement call happens before we start the CLIENT span (`DbInfo dbInfo = extractDbInfo(connection)`), so the `shouldStartSpan()` cannot prevent this - it's too late.

### DIFF
--- a/instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/PreparedStatementInstrumentation.java
+++ b/instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/PreparedStatementInstrumentation.java
@@ -53,7 +53,13 @@ public class PreparedStatementInstrumentation implements TypeInstrumentation {
         @Advice.This PreparedStatement statement,
         @Advice.Local("otelContext") Context context,
         @Advice.Local("otelScope") Scope scope) {
-      // prevent recursive Statement calls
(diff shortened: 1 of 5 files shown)


instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/PreparedStatementInstrumentation.java
```diff
@@ -51,6 +53,11 @@ public static void onEnter(
         @Advice.This PreparedStatement statement,
         @Advice.Local("otelContext") Context context,
         @Advice.Local("otelScope") Scope scope) {
+      // prevent recursive Statement calls
```
trask, line 56: can you expand on this just a bit? at first glance I didn't understand why this was different from our nested CLIENT span suppression

trask (approved):
can you add a test for prepared statement too? just because this feels fragile (and an important fix 👍)

anuraaga (approved):

marcobjorge:
Hi @mateuszrzeszutek ,

With this fix if we have multi-layer JDBC driver we are never able to capture traces from the innermost layers (e.g. using Apache Calcite with underlying MySQL). Any ideias how we can work around it, even if as a workaround?

Thanks,

iNikem:
@marcobjorge can you please open a new issue for this? Otherwise it will be lost.

PR #2732, opened 2021-04-06
Cleanup mongodb tests

anuraaga:
- Use testcontainers to start mongodb to avoid hacky global file lock synchronization
- Use unique collection names to allow sharing mongodb across tests
- Move test code into base class and have subclasses only implement the mongo operations themselves based on that version's API

PR #2758, opened 2021-04-08
Mongo reactive context propagation

laurit:
Fixes flaky test where the number of traces varied.


mateuszrzeszutek (approved):

trask (approved):
can you add a more explicit test for the new instrumentation?

anuraaga (commented):
Also curious how you figured this out :) I started getting flakiness locally from #2732 but couldn't see any difference in the code or a reason to get flakiness so wasn't able to debug.

laurit:
@anuraaga For me it locally fails consistently and passes when other tests are commented out so I could compare the passing & failing runs and figure out what was different.

trask (commented):
nice test improvements!

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
```diff
@@ -0,0 +1,157 @@
...
+          .and(
+              declaresMethod(
+                  named("addCommandListener")
+                      .and(
+                          takesArguments(
+                              new TypeDescription.Latent(
+                                  "com.mongodb.event.CommandListener",
+                                  Modifier.PUBLIC,
+                                  null,
+                                  Collections.<TypeDescription.Generic>emptyList())))
```
mateuszrzeszutek, line 64: Why not just `named("com.mongodb.event.CommandListener")`?

mateuszrzeszutek (approved):

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/MongoClientInstrumentationModule.java
laurit: This is copy paste from mongo 3.7, I have no idea why it is done this way. I'll change it to use `named` for all mongo versions.

PR #2768, opened 2021-04-10
Introducing change log

trask:
We are planning to start follow the process being used in the opentelemetry-java project, of maintaining change log file, and updating it approx. weekly, in order to ease the release burden of doing it all at once.

(suggestions/edits are more than welcome)


CHANGELOG.md
```diff
@@ -0,0 +1,135 @@
...
+  ([#2629](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2629))
+- Make muzzle reference creation package(s) configurable
+  ([#2615](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2615))
+- New Instrumenter API
+  ([#2596](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2596))
+- Added Concurrent server test
+  ([#2680](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2680),
+   [#2750](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2750),
+   [#2749](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2749))
+- Added smoke tests for Tomcat 10 and Jetty 11 (Servlet API 5)
```
jkwatson, line 115: I wouldn't put this kind of thing in the changelog. end-users won't care about this, I don't think. I personally try to keep the changelog user-relevant, rather than inundating them with things they probably don't care about.

mateuszrzeszutek (approved):

CHANGELOG.md
iNikem: I kinda agree with John here. Probably only "New Instrumenter API" is worth mentioning here.

CHANGELOG.md
```diff
@@ -0,0 +1,135 @@
...
+- Reduce test flakiness
+  ([#2738](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2738),
+   [#2728](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2728),
+   [#2754](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2754),
+   [#2753](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2753),
+   [#2759](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2759))
+- Split HttpClientTest execution methods for sync and callback
+  ([#2675](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2675))
+
+### ☢️ Breaking changes
```
iNikem, line 129: Can we move them to the top? To make sure that a reader will notice them.

CHANGELOG.md
trask: @anuraaga I think you introduced the "Tooling" category in the release notes, so want to check what you think.

I thought the "Tooling" category was nice for contributors who don't follow every PR, but it's certainly less work not to include it, and I agree it isn't relevant to end users.

anuraaga: I think it's ok to have it since if it's at the end, easy enough to skip. The main motivation for having it was less about explaining to users and more about giving credit to the contributors, which I think is important, especially if they're non-maintainers. It could be only in release notes and not in the CHANGELOG, but I guess it's simpler if we just keep them matched up.

iNikem: I do agree with giving credit to contributors, but I think we need more clear understanding, who is the target group of release notes. Or having different sections for different groups.

CHANGELOG.md
```diff
@@ -0,0 +1,135 @@
+# Changelog
```
kubawach, line 1: That's a great idea, I missed something like this. I also very much like split on separate sections, with different target groups.

kubawach (approved):

anuraaga (approved):

iNikem (approved):

trask:
Trimmed down the Tooling section in this PR to match the published release notes https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.1.0

I started to split out separate "Instrumentation users" and "Instrumentation authors" sections, but "Instrumentation authors" turned out to be the same as "Tooling" (which is now significantly trimmed), so splitting out the two sections didn't seem to provide much value

mateuszrzeszutek (approved):

trask:
Going to merge this as-is, can always revisit/update.

PR #2772, opened 2021-04-12
Support URLClassLoader addURL

trask:
Ran into an issue with URLClassLoaders that dynamically add jar files later in their lifespan via `URLClassLoader.addURL()`.

The problem is that any instrumentation that has a `classLoaderOptimization` using `hasClassesNamed` (`ClassLoaderHasClassesNamedMatcher`) will cache the initial result that the URLClassLoader doesn't have that class, and won't re-test later on if that class is dynamically added.

This PR instruments `URLClassLoader.addURL()` and invalidates the cache for that class loader.


javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/ClassLoaderMatcherCacheHolder.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class ClassLoaderMatcherCacheHolder {
+
+  @GuardedBy("allCaches")
+  private static final List<Cache<ClassLoader, Boolean>> allCaches = new ArrayList<>();
+
+  private ClassLoaderMatcherCacheHolder() {}
+
+  public static void addCache(Cache<ClassLoader, Boolean> cache) {
+    allCaches.add(cache);
```
anuraaga, line 27: I think you still need `synchronized` to ensure memory updates are synced. Otherwise you want `CopyOnWriteArrayList`

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/ClassLoaderMatcherCacheHolder.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  public static void invalidateAllCachesForClassLoader(ClassLoader loader) {
+    List<Cache<ClassLoader, Boolean>> allCachesCopy;
+    synchronized (allCaches) {
```
anuraaga, line 32: If we synchronize on `add` (which we should), I don't think we need this copy, can just synchronize this method.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/bytebuddy/matcher/ClassLoaderMatcher.java
```diff
@@ -43,6 +44,7 @@ private ClassLoaderHasClassesNamedMatcher(String... classNames) {
       for (int i = 0; i < resources.length; i++) {
         resources[i] = resources[i].replace(".", "/") + ".class";
       }
+      ClassLoaderMatcherCacheHolder.addCache(cache);
```
anuraaga, line 47: Any thoughts on storing `URLClassLoader.getUrls().length` into the cache instead of `Boolean` and invalidating here if it doesn't match? The nice property of URLClassPath is it can only be added to, not removed from.

Pros
- Less instrumentation
- Less bootstrap classpath logic
- Doesn't split logic in two places

Cons
- Slower when matching, `URL[]` copy when calling `getURLs`. But benchmarks may not find it to be a big deal. 
  - If only URLClassLoader had a `size` method :(

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/ClassLoaderMatcherCacheHolder.java
trask: oops, thx!

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/ClassLoaderMatcherCacheHolder.java
trask: I get nervous calling anything under lock that I don't know exactly what it does (in this case `cache.remove()`), u know that code well though, so I'm good taking your recommendation

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/bytebuddy/matcher/ClassLoaderMatcher.java
trask: oh, that's a good idea, bummer about the `URL[]` copy, e.g. war file with 100 jars would have to make a 100-element copy on every `hasClassesNamed` match (which looks like there are ~100, and I think they all have to trigger for each class load)

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/ClassLoaderMatcherCacheHolder.java
anuraaga: Do you mean for performance reasons? Yeah I wouldn't worry about it for removing from a cache. Without the synchronized around the method, this code is harder to reason about - I think it actually is thread-safe only because we know `addURL` and `loadClass` are guarded by the same lock. If they weren't ,then a `loadClass` adding a new cache object could happen between copy and invalidate.

trask: my nervousness is about deadlocks 😁

> `loadClass` adding a new cache object could happen between copy and invalidate

oh, I hadn't thought about this, done 👍

anuraaga (approved):

instrumentation/classloaders/integration-test/src/test/groovy/AddUrlTest.groovy
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import io.opentelemetry.instrumentation.test.AgentInstrumentationSpecification
+import org.apache.commons.io.IOUtils
+import org.apache.commons.lang3.SystemUtils
+
+class AddUrlTest extends AgentInstrumentationSpecification {
+
+  def "should instrument class after it is loaded via addURL"() {
```
iNikem, line 12: I think that we have to test that instrumentation was NOT applied before `addURL`.

iNikem (approved):

instrumentation/classloaders/integration-test/src/test/groovy/AddUrlTest.groovy
trask: added verification 👍

Issue #1382, opened 2020-10-14
Embed java-classloader into javaagent-tooling


trask:
Benefits:

* It's not really an optional instrumentation
* It could then rely on internal api, without having to expose more than needed in `javaagent-api` (e.g. in #1380)

We don't need generated muzzle matcher for this instrumentation, so I _think_ it could just be moved into javaagent-tooling without much hassle (but haven't tried, could be wrong).

We could keep java-classloader-testing module with the `jboss`, `osgi` and `tomcat` class loader testing more or less as is.

iNikem:
If this is a required component for the functioning of the whole agent, then it makes sense.

trask:
It seems like good consensus for this change, turning this from a question into an issue...

trask:
From https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2773#issuecomment-817584256

> Alternatively, we could rename the `classloaders` module into something like `internal` - classloader instrumentation is pretty crucial and the javaagent won't work properly without it.

PR #2773, opened 2021-04-12
Avoid proxy IllegalArgumentException: repeated interface

trask:
Some bad code lives out in the world that dynamically scrapes interfaces from classes at runtime and feeds those into [Proxy.newProxyInstance()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/reflect/Proxy.html#newProxyInstance(java.lang.ClassLoader,java.lang.Class%5B%5D,java.lang.reflect.InvocationHandler)).

This can be a problem when introducing the javaagent, because we add `FieldBackedContextStoreAppliedMarker` to several classes, and then when this code scrapes classes at runtime (and is not careful to de-dup, which arguably it should), it can end up passing more than one `FieldBackedContextStoreAppliedMarker` to `Proxy.newProxyInstance()`, which then throws

```
IllegalArgumentException: repeated interface: io.opentelemetry.javaagent.bootstrap.FieldBackedContextStoreAppliedMarker
```

Putting this instrumentation into the `classloaders` module seems reasonable since it's low-level `java.lang` stuff, but I can also split it out if there's a preference for that.


mateuszrzeszutek (approved):

mateuszrzeszutek:
> Putting this instrumentation into the `classloaders` module seems reasonable since it's low-level java.lang stuff, but I can also split it out if there's a preference for that.

Alternatively, we could rename the `classloaders` module into something like `internal` - classloader instrumentation is pretty crucial and the javaagent won't work properly without it.

trask:
> Alternatively, we could rename the `classloaders` module into something like `internal` - classloader instrumentation is pretty crucial and the javaagent won't work properly without it.

👍 added this comment to #1382, will send a PR to get everyone's feedback on it.

PR #2774, opened 2021-04-12
First POC for external extension loading

iNikem:
Please start reading from `ExtensionClassLoader`.

This is the first attempt to address #2442. Manual testing confirmed that this works.


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,172 @@
...
+    } catch (MalformedURLException e) {
+      System.err.println("Filename could not be parsed: %s. Extension location is ignored");
+      e.printStackTrace();
+    }
+    return null;
+  }
+
+  public ExtensionClassLoader(URL url, ClassLoader parent) {
+    super(new URL[] {url}, parent);
+    this.manifest = getManifest(url);
```
trask, line 75: oh, I remember this, I'm not sure how to generalize this to more than one jar file under the constraint of using URLClassLoader

maybe we can pass URLs to URLClassLoader that have custom InputStreams that do the shading? If that can work somehow (big if) then URLClassLoader will take care of the class/package definitions for us and we don't need the manifest

trask (commented):
any thoughts on class loader isolation, so extensions can bring their own version of grpc/netty and not cause conflicts? or should we table that aspect for now

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,172 @@
...
+  // with transitive dependencies down the line.
+  private static final ShadingRemapper remapper =
+      new ShadingRemapper(
+          rule("#io.opentelemetry.api", "#io.opentelemetry.javaagent.shaded.io.opentelemetry.api"),
+          rule(
+              "#io.opentelemetry.context",
+              "#io.opentelemetry.javaagent.shaded.io.opentelemetry.context"),
+          rule(
+              "#io.opentelemetry.extension.aws",
+              "#io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.aws"),
```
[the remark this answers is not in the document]
anuraaga: This is the x-ray propagator - I think it's to be able to inject into user classloader for use from instrumentation, even if they have their own (potentially incompatible) one. Since extension instrumentation may use the propagator it seems appropriate to shade here too.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,172 @@
...
+          rule("#org.slf4j", "#io.opentelemetry.javaagent.slf4j"));
+
+  private final Manifest manifest;
+
+  public static ClassLoader getInstance(ClassLoader parent) {
+    // TODO add support for old deprecated properties, otel.exporter.jar and otel.initializer.jar
+    // TODO add support for system properties
+    URL extension = parseLocation(System.getenv("OTEL_JAVAAGENT_EXTENSIONS"));
+    if (extension != null) {
+      return new ExtensionClassLoader(extension, parent);
```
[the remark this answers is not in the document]
iNikem: I would support both single file and a directory.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: Hm... in fact we can do this, I think... I will give it a try.

iNikem:
What isolation of what from what are you thinking about? :) One extension jar from another? Extensions from AgentClassLoader?

trask:
> What isolation of what from what are you thinking about? :) One extension jar from another? Extensions from AgentClassLoader?

these are good questions 😁, ideally all of the above(?), but don't want to over-engineer it either

iNikem:
Extensions are already loaded from a separate classloader, which is a child of AgentClassLoader. Meaning nobody should see their classes. Extensions themselves should have access to agent classes, thus AgentClassLoader as a parent. This current state + possible separating extensions from each other - is it Ok or do we want more?

examples/extension/smoke-tests/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+}
+
+tasks.test {
+  useJUnitPlatform()
+
+  def shadowTask = project(":custom").tasks.jar
+  inputs.files(layout.files(shadowTask))
+
+  doFirst {
+    jvmArgs("-Dio.opentelemetry.smoketest.agent.shadowJar.path=/Users/nsalnikovtarnovski/Documents/workspace/opentelemetry-java-instrumentation/javaagent/build/libs/opentelemetry-javaagent-1.2.0-SNAPSHOT-all.jar")
```
mateuszrzeszutek, line 23: You probably need to pass some other javaagent path here 😄

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -34,21 +34,27 @@ private static synchronized void startAgent(Instrumentation inst, URL bootstrapU
       throws Exception {
     if (AGENT_CLASSLOADER == null) {
       ClassLoader agentClassLoader = createAgentClassLoader("inst", bootstrapUrl);
+      AGENT_CLASSLOADER = agentClassLoader;
+
       Class<?> agentInstallerClass =
           agentClassLoader.loadClass("io.opentelemetry.javaagent.tooling.AgentInstaller");
```
mateuszrzeszutek, line 40: If `agentClassLoader` is an `ExtensionClassLoader` then won't we put `AgentInstaller` in the same CL as extensions? And won't it apply shading transformations on all agent classes?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,172 @@
...
+      } catch (IOException e) {
+        e.printStackTrace();
+        //        log.debug(e.getMessage(), e);
+      }
+    }
+  }
+
+  @Override
+  public InputStream getResourceAsStream(String name) {
+    InputStream originalStream = super.getResourceAsStream(name);
```
mateuszrzeszutek, line 103: Missing `if (originalStream == null)`

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
iNikem: No, because `ExtensionClassLoader` first delegates to parent, as usual. And its parent is the old `AgentClassLoader`

Issue #2678, opened 2021-03-31
Muzzle can generate InstrumentationModule#contextStore()


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
This is another muzzle improvement that can be done to make javaagent instrumentation contributor's life a bit easier: it is possible to collect all class pairs by detecting `InstrumentationContext.get()` calls in the `ReferenceCollector` when it traverses all advice & helper classes.

anuraaga:
Just wondering, does this mean running muzzles in custom instrumentation is a requirement when using context store? Previously, missing muzzle might cause some unsafety, but IIUC now it lost a feature. If so, I guess it's a big regression until https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2778

mateuszrzeszutek:
You can implement `getMuzzleContextStoreClasses()` by yourself (just rename your `contextStore()` method), in the runtime it still works the same.

anuraaga:
@mateuszrzeszutek Ah great, thanks!

PR #2775, opened 2021-04-12
Generate InstrumentationModule#contextStore() with muzzle

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2678


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollectingClassVisitor.java
```diff
@@ -349,6 +360,28 @@ public void visitMethodInsn(
...
+          && methodType.getArgumentTypes().length == 2) {
+        // in case of invalid scenario (not using .class ref directly) don't store anything and
+        // clear the last LDC <class> stack
+        // note that FieldBackedProvider also check for an invalid context call in the runtime
+        if (lastTwoClassConstants.hasMaxSize()) {
+          String className = lastTwoClassConstants.pop();
+          String contextClassName = lastTwoClassConstants.pop();
+          contextStoreClasses.put(className, contextClassName);
+        } else {
+          lastTwoClassConstants.clear();
```
trask, line 381: can we throw exception here?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollectingClassVisitor.java
```diff
@@ -276,6 +281,10 @@ private static ManifestationFlag computeTypeManifestationFlag(int access) {
   private class AdviceReferenceMethodVisitor extends MethodVisitor {
     private int currentLineNumber = -1;
 
+    // this data structure will remember last two LDC <class> instructions before
+    // InstrumentationContext.get() call
+    private final FixedSizeQueue<String> lastTwoClassConstants = new FixedSizeQueue<>(2);
+
```
trask, line 287: it might work nice to extract this to a separate "TrackingMethodVisitor", and then `new AdviceReferenceMethodVisitor(new TrackingMethodVisitor(methodVisitor))`

trask (approved):
❤️

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/FixedSizeQueue.java
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.tooling.muzzle.collector;
+
+import java.util.ArrayDeque;
+import java.util.Iterator;
+
+final class FixedSizeQueue<T> implements Iterable<T> {
```
anuraaga, line 11: I think because of use of graph API, we already expect to not be able to remove Guava dependency from muzzle. In which case you might use `EvictingQueue`

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollectingClassVisitor.java
mateuszrzeszutek: That's a great idea!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/FixedSizeQueue.java
iNikem: Or just a ring buffer backed by fixed size array?

iNikem (commented):
You are a muzzle magician :)

PR #2782, opened 2021-04-12
Akka: clean up state on exception

laurit:
Instrumentations that use `ExecutorInstrumentationUtils.setupState` usually call `ExecutorInstrumentationUtils.cleanUpOnMethodExit` to clear state when there was an exception. Make akka dispatcher instrumentation do the same.


trask (approved):
👍

PR #2787, opened 2021-04-13
Rename (and split up) classloaders module

### LINKED ISSUE #1382: Embed java-classloader into javaagent-tooling

trask:
Before this PR:

* classloaders
  * javaagent
  * javaagent-integration-tests
  * javaagent-unit-tests

After this PR:

* internal
  * internal-class-loader
    * javaagent
  * internal-proxy
    * javaagent
    * javaagent-unit-tests
  * internal-url-class-loader
    * javaagent
    * javaagent-integration-tests

I think reasonable to say this closes #1382, with the idea that we can use internal javaagent apis from these internal instrumentation modules(?)


testing/agent-exporter/agent-exporter.gradle
```diff
@@ -24,7 +24,7 @@ dependencies {
   implementation project(':instrumentation:executors:javaagent')
   // FIXME: we should enable this, but currently this fails tests for google http client
   //testImplementation project(':instrumentation:http-url-connection:javaagent')
-  implementation project(':instrumentation:classloaders:javaagent')
+  implementation project(':instrumentation:internal:internal-class-loader:javaagent')
```
mateuszrzeszutek, line 27: Should we add all those internal modules to the test agent?

mateuszrzeszutek (approved):

PR #2789, opened 2021-04-13
Extract Mongo library instrumentation

anuraaga:


instrumentation/mongo/mongo-3.1/library/src/main/java/io/opentelemetry/instrumentation/mongo/v3_1/MongoTracing.java
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.mongo.v3_1;
+
+import com.mongodb.event.CommandListener;
+import io.opentelemetry.api.OpenTelemetry;
+
+/** Entrypoint to OpenTelemetry instrumentation of the MongoDB client. */
+public class MongoTracing {
```
mateuszrzeszutek, line 12: Can it be final?

instrumentation/mongo/mongo-3.1/testing/src/main/groovy/io/opentelemetry/instrumentation/mongo/v3_1/AbstractMongo31ClientTest.groovy
```diff
@@ -0,0 +1,174 @@
...
+    def options = MongoClientOptions.builder().description("some-description")
+    configureMongoClientOptions(options)
+    client = new MongoClient(new ServerAddress("localhost", port), options.build())
+  }
+
+  def cleanupSpec() throws Exception {
+    client?.close()
+    client = null
+  }
+  @Override
```
mateuszrzeszutek, line 40: Nit:

```suggestion

  @Override
```

mateuszrzeszutek (approved):

instrumentation/mongo/mongo-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v3_1/MongoClientInstrumentationModule.java
```diff
@@ -67,12 +66,7 @@ public MongoClientInstrumentationModule() {
     public static void injectTraceListener(
         @Advice.This MongoClientOptions.Builder builder,
         @Advice.FieldValue("commandListeners") List<CommandListener> commandListeners) {
-      for (CommandListener commandListener : commandListeners) {
-        if (commandListener instanceof TracingCommandListener) {
-          return;
-        }
-      }
```
trask, line 74: this may have been protecting against calling `build()` twice on the same builder

trask (approved):
👍

instrumentation/mongo/mongo-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v3_1/MongoClientInstrumentationModule.java
anuraaga: Ah good point, restored dedupe. Strangely a new test I added doesn't fail even without these though, haven't dug into why that might be.

instrumentation/mongo/mongo-3.1/testing/src/main/groovy/io/opentelemetry/instrumentation/mongo/v3_1/AbstractMongo31ClientTest.groovy
```diff
@@ -60,6 +60,15 @@ abstract class AbstractMongo31ClientTest extends AbstractMongoClientTest<MongoCo
...
   }
 
+  @Override
+  void createCollectionCallingBuildTwice(String dbName, String collectionName) {
+    def options = MongoClientOptions.builder().description("some-description")
+    configureMongoClientOptions(options)
+    options.build()
+    MongoDatabase db = new MongoClient(new ServerAddress("localhost", port), options.build()).getDatabase(dbName)
+    db.createCollection(collectionName)
+  }
```
trask, line 70: 👍

trask (approved):

instrumentation/mongo/mongo-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v3_1/MongoClientInstrumentationModule.java
trask: > a new test I added doesn't fail even without these though

oh no, maybe it's not needed after all(?)

anuraaga: I wonder if our workForTraces doesn't actually check for extra traces? I thought it did

trask: try putting a sleep before it to give the second trace time to get there

anuraaga: There isn't really anywhere to put a sleep - build just causes two listeners to be registered, which I could confirm in the debugger and Thread.dumpStack. Weird but not sure I'll dig too much into it

instrumentation/mongo/mongo-testing/src/main/groovy/io/opentelemetry/instrumentation/mongo/testing/AbstractMongoClientTest.groovy
```diff
@@ -120,6 +124,29 @@ abstract class AbstractMongoClientTest<T> extends AgentInstrumentationSpecificat
...
   }
 
+  def "test create collection calling build twice"() {
+    when:
+    runUnderTrace("parent") {
+      createCollectionCallingBuildTwice(dbName, collectionName)
+    }
+
+    then:
+    assertTraces(1) {
```
trask, line 134: try putting the sleep before this line, but agree not worth a lot of time, thx for adding the test and verifying manually 👍

PR #2794, opened 2021-04-13
Add instrumentation for RxJava 3

HaloFour:
Adds instrumentation for RxJava 3.0.  This is mostly a copypasta of the code for RxJava3 with updated dependencies and package names as the majority of the API surface is the same between RxJava 3 and RxJava 2.


mateuszrzeszutek (approved):
LGTM 👍 
Thanks!

instrumentation/rxjava/rxjava-3.0/library/src/main/java/io/opentelemetry/instrumentation/rxjava3/TracingAssembly.java
```diff
@@ -0,0 +1,279 @@
...
+import org.reactivestreams.Subscriber;
+
+/**
+ * rxjava3 library instrumentation.
+ *
+ * <p>In order to enable rxjava3 instrumentation one has to call the {@link
+ * TracingAssembly#enable()} method.
+ *
+ * <p>Instrumentation uses <code>on*Assembly</code> and <code>on*Subscribe</code> RxJavaPlugin hooks
+ * to wrap rxjava3 classes in their tracing equivalents.
```
trask, line 52:
```suggestion
 * RxJava3 library instrumentation.
 *
 * <p>In order to enable RxJava3 instrumentation one has to call the {@link
 * TracingAssembly#enable()} method.
 *
 * <p>Instrumentation uses <code>on*Assembly</code> and <code>on*Subscribe</code> RxJavaPlugin hooks
 * to wrap RxJava3 classes in their tracing equivalents.
```

trask (approved):
thx!

instrumentation/rxjava/rxjava-3.0/library/src/main/java/io/opentelemetry/instrumentation/rxjava3/TracingAssembly.java
HaloFour: oops, good catch!

PR #2796, opened 2021-04-13
Use isHelperClass over additionalHelperClassNames

trask:
Since `isHelperClass` is generally simpler and less error-prone compared to `additionalHelperClassNames`, I think it makes sense to use `isHelperClass` wherever possible, and (only) recommend `isHelperClass` in our `InstrumentationModule` docs (#2793).

There's only one place that I couldn't remove usage of `additionalHelperClassNames` completely: `MethodInstrumentationModule`.


anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):
💯 👍 

My thoughts exactly - I realized that those two methods do the same thing when I was writing the `InstrumentationModule` doc 😄

PR #2798, opened 2021-04-13
Bump smoke test timeout

trask:
Ran across a smoke test (windows-latest, wildfly) that timed out but wasn't really hung (just slower than usual), so bumping the timeout a bit to avoid unnecessary failures.

[GitHub Action log](https://github.com/open-telemetry/opentelemetry-java-instrumentation/files/6307262/21.txt) in case anyone is interested.


anuraaga (approved):

trask:
Another failure spotted on #2794

Issue #2600, opened 2021-03-19
Support sub-millisecond precision for start/end times on Java 9+


trask:
The OpenTelemetry SDK supports sub-millisecond precision for start/end times on Java 9+ via multi-release jar file.

But I checked under a debugger, while running Java 11, and the javaagent is not using `io.opentelemetry.sdk.internal.Java9VersionSpecific`, and so we are only getting millisecond precision on start/end times.

PR #2803, opened 2021-04-14
Use Instrumenter in JMS instrumentation

mateuszrzeszutek:
And introduce messaging semantic conventions

Part of #2713


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessageOperation.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Represents type of <a
+ * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#operation-names">operations</a>
+ * that may be used in a messaging system.
+ */
+public enum MessageOperation {
+  SEND,
+  RECEIVE,
+  PROCESS;
```
iNikem, line 16: I will ask my favourite question: why enum cannot be already in lower case?

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesExtractorTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        @Override
+        protected String messageId(Map<String, String> request, String response) {
+          return response;
+        }
+      };
+
+  @ParameterizedTest
+  @MethodSource("destinations")
+  void shouldExtractAllAvailableAttributes(
```
iNikem, line 97: I am a little bit sceptical about the value of this test. It _may_ be useful to test extracting operation name and behaviour in case of temporary destination. But otherwise, maybe it can be replaced with tests of actual implementations of `MessagingAttributesExtractor`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingSpanNameExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Returns a {@link SpanNameExtractor} that constructs the span name according to <a
+   * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-name>
+   * messaging semantic conventions</a>: {@code <destination name> <operation name>}.
+   *
+   * @see MessagingAttributesExtractor#destination(Object) used to extract {@code <destination
+   *     name>}.
+   * @see MessagingAttributesExtractor#operation(Object) used to extract {@code <operation name>}.
+   */
+  public static <REQUEST> SpanNameExtractor<REQUEST> create(
```
iNikem, line 21: Why does this class have a factory method, while `MessagingAttributesExtractor` does not? What is so different about this class that requires special construction?

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/MessageWithDestination.java
```diff
@@ -0,0 +1,108 @@
...
+
+  public String getDestinationKind() {
+    return destinationKind;
+  }
+
+  public boolean isTemporaryDestination() {
+    return temporaryDestination;
+  }
+
+  public static MessageWithDestination create(
```
iNikem, line 59: This method would benefit from test

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessageOperation.java
anuraaga: Are these missing from the generated conventions? If so let's fix it.

Until then we could also go with static final String which is what our generator produces since yesterday

mateuszrzeszutek: The semconv generated enum is missing the `send` value:

```java
  public enum MessagingOperationValues {
    /** receive. */
    RECEIVE("receive"),
    /** process. */
    PROCESS("process"),
    ;
  }
```

I suppose that's because setting `message.operation` to `send` is strictly forbidden in the spec. Still, having `send` available makes the instrumentation code a bit clearer...

> why enum cannot be already in lower case?

Good question - I don't really mind it, but most people tend to dislike anything other than CONSTANT_CASE.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingSpanNameExtractor.java
mateuszrzeszutek: `MessagingAttributesExtractor` is an abstract class that has to be extended, it cannot have a static factory method.

> What is so different about this class that requires special construction?

Hmm, there's nothing special - I've added it here because we use builders & factory methods (instead of plain constructor calls) in the instrumenter API. I guess that's more resistant to future changes? But I may be overthinking this, and maybe it just looks a bit cooler 😄

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesExtractorTest.java
mateuszrzeszutek: I think that the real value is testing the full semantic conventions themselves, not just one implementation of them.

Maybe if we replaced the inheritance with composition too those tests would make more sense: something like `MessagingAttributesExtractor.create(new JmsAttributesGetter())`

mateuszrzeszutek: I refactored the attributes extractor and made it not abstract: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2803/commits/73f848ae595f582b48d9bdfe7dd1240ac4cec019
Please tell me WDYT about it - the separation of concerns is nice, but on the other hand it makes setting up an `Instrumenter` a bit more bloated.

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java
```diff
@@ -49,28 +52,27 @@
...
+      Context parentContext = Java8BytecodeBridge.currentContext();
+      MessageWithDestination request =
+          MessageWithDestination.create(message, MessageOperation.receive, null);
 
-      if (throwable != null) {
-        tracer().endExceptionally(context, throwable);
-      } else {
-        tracer().end(context);
+      if (consumerInstrumenter().shouldStart(parentContext, request)) {
+        Context context = consumerInstrumenter().start(parentContext, request, startTime);
```
anuraaga, line 74: Any thoughts on adding `startTime` to `MessageWithDestination` and adding `StartTimeExtractor` to read it from something? If using Armeria's start time, which I would if we do end up needing this knob more generally :), it would just be reading the time off the request object. Current approach might be more general though, wondering what you think. I think `StartTimeExtractor` hides the complexity more than adding to our core API, something to consider if this is extremely rarely needed.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesGetter.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.messaging;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+// TODO: javadocs
```
anuraaga, line 10: Can we at least add the class doc like we've done for the others?

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesExtractorTest.java
anuraaga: Hmm - I guess I lean towards the abstract version if it keeps using the `Instrumenter` simpler. Not sure there was a huge problem with the test

anuraaga (commented):
Thanks a lot @mateuszrzeszutek mostly looks good and just some questions

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesGetter.java
mateuszrzeszutek: That was meant to be just a demonstration of the `AttributesGetter` idea - I'll revert it for now but we can always come back to this discussion later.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesExtractorTest.java
mateuszrzeszutek: I'll revert it for now, we can come back to this discussion in a separate issue/PR. Personally I don't really have a strong opinion about either solution, both have some good and bad sides.

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/JmsMessageConsumerInstrumentation.java
mateuszrzeszutek: Oh, that's much better than passing the argument -- will do.

mateuszrzeszutek: And I can add an `EndTimeExtractor` too while we're working on it - the `end()` method already has too many params.

mateuszrzeszutek:
Added `StartTimeExtractor` and `EndTimeExtractor`, those two should also be useful for metrics. And I've accidentally resolved https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2600 for Instrumenter-using instrumentations too 😂

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -56,22 +57,19 @@
...
+    this.instrumentationName = builder.instrumentationName;
+    this.tracer = builder.openTelemetry.getTracer(builder.instrumentationName);
+    this.spanNameExtractor = builder.spanNameExtractor;
+    this.spanKindExtractor = builder.spanKindExtractor;
+    this.spanStatusExtractor = builder.spanStatusExtractor;
+    this.extractors = new ArrayList<>(builder.attributesExtractors);
+    this.errorCauseExtractor = builder.errorCauseExtractor;
+    this.startTimeExtractor = builder.startTimeExtractor;
+    this.endTimeExtractor = builder.endTimeExtractor;
   }
```
mateuszrzeszutek, line 73: I refactored the Instrumenter constructor and used the same approach as OkHttp builder. I think it makes passing all those extractors a bit easier on the eyes

trask (approved):
> And I've accidentally resolved https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2600 for Instrumenter-using instrumentations too 😂

🎉

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/CurrentNanoTime.java
```diff
@@ -0,0 +1,79 @@
...
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.util.concurrent.TimeUnit;
+import java.util.function.LongSupplier;
+
+/**
+ * This utility class allows to retrieve an {@link Instant} with the current time measured with
+ * nanosecond precision.
+ */
+public final class CurrentNanoTime {
```
anuraaga, line 19: I think instead of this class defaults should return `-1` or `null` - we want to use the SDK's `Clock` as much as possible, only override as an exception. This is not quite as important for `start` but very improtant for `end` which should only be using nano time, not wall time

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -112,7 +112,8 @@ public Context start(Context parentContext, REQUEST request) {
         tracer
             .spanBuilder(spanNameExtractor.extract(request))
             .setSpanKind(spanKind)
-            .setParent(parentContext);
+            .setParent(parentContext)
+            .setStartTimestamp(startTimeExtractor.extract(request));
```
anuraaga, line 116: So e.g.,

```
if (startTimeExtractor != DEFAULT) {
  spanBuilder.setStartTimestamp(startTimeExtractor.extract(...));
}
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/CurrentNanoTime.java
trask: awwww, we lose the fix for #2600, oh well.. 😄

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessageOperation.java
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Represents type of <a
+ * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#operation-names">operations</a>
+ * that may be used in a messaging system.
+ */
+public enum MessageOperation {
+  send,
+  receive,
+  process
```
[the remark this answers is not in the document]
mateuszrzeszutek: 😂

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -83,23 +85,52 @@
   }
 
   /**
-   * Returns a new client {@link Instrumenter} which will create client spans and inject context
-   * into requests.
+   * Sets the {@link StartTimeExtractor} and the {@link EndTimeExtractor} to extract the timestamp
+   * marking the start and end of processing.
+   */
+  public InstrumenterBuilder<REQUEST, RESPONSE> setTimeExtractors(
+      StartTimeExtractor<REQUEST> startTimeExtractor, EndTimeExtractor<RESPONSE> endTimeExtractor) {
```
mateuszrzeszutek, line 92: These two should probably be always set together - start & end timestamps need to be comparable, so it does not make any sense to set just one and compare with the SDK timestamp.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/StartTimeExtractor.java
```diff
@@ -0,0 +1,33 @@
...
+   */
+  static <REQUEST> StartTimeExtractor<REQUEST> getDefault() {
+    return request -> null;
+  }
+
+  /**
+   * Returns the timestamp marking the start of the request processing. If the returned timestamp is
+   * {@code null} the OpenTelemetry SDK will use its internal clock to determine the start time.
+   */
+  @Nullable
```
anuraaga, line 31: Do you think there's a use case for a custom extractor to delegate to SDK conditionally? Otherwise maybe we don't need to make this default and can treat that as an implementation detail (we could even change the `null` check to `== getDefault()`. This is ok too though.

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/StartTimeExtractor.java
mateuszrzeszutek: > Do you think there's a use case for a custom extractor to delegate to SDK conditionally?

No, it doesn't make any sense for a custom extractor to delegate to SDK. I'll change that.
Thanks!

mateuszrzeszutek: I've removed the `getDefault()` methods from start/end extractors.

PR #2804, opened 2021-04-14
Move ContextPropagationDebug to internal package

mateuszrzeszutek:
And use it in the Instrumenter API


anuraaga (approved):
As much as I want to eliminate leaks and delete this that is going to take a long time so seems ok.

mateuszrzeszutek:
> As much as I want to eliminate leaks and delete this that is going to take a long time so seems ok.

Well, it might not be pretty, but it did help us detect some context leaks (JMS).

trask (approved):

PR #2809, opened 2021-04-15
Fix failing test on latest jersey client

laurit:
`JerseyInvocation` makes a copy of `ClientRequest` before passing it on because of that we see a different request in `ClientTracingFilter` where we start the span and `JerseyInvocation` where we end it in case of an exception.
I also change the handling for exceptions in async request, but unfortunately we don't have a test for it. I plan to add it later, but it will take some time because currently `HttpClientTest` doesn't handle async failures, adding it will require changes in about a million places.


instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/JerseyClientInstrumentationModule.java
```diff
@@ -35,6 +36,13 @@ public JerseyClientInstrumentationModule() {
     super("jaxrs-client", "jaxrs-client-2.0", "jersey-client", "jersey-client-2.0");
   }
 
+  @Override
+  public String[] additionalHelperClassNames() {
+    return new String[] {
+      "org.glassfish.jersey.client.ResponseCallbackWrapper",
+    };
+  }
```
mateuszrzeszutek, line 44: Can you use `isHelperClass()` instead? I think we'll get rid of `additionalHelperClassNames` soon, since `isHelperClass()` does the same thing but better.

```suggestion
  @Override
  public boolean isHelperClass(String className) {
    return className.equals("org.glassfish.jersey.client.ResponseCallbackWrapper");
  }
```

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/JerseyClientInstrumentationModule.java
```diff
@@ -66,8 +76,7 @@ public JerseyClientInstrumentationModule() {
 
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void handleError(
-        @Advice.FieldValue("requestContext") ClientRequest context,
-        @Advice.Thrown Throwable throwable) {
+        @Advice.Argument(0) ClientRequest context, @Advice.Thrown Throwable throwable) {
```
mateuszrzeszutek, line 79: Can you add a `takesArgument(...)` condition to the method matcher too?

mateuszrzeszutek (approved):

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/org/glassfish/jersey/client/ResponseCallbackWrapper.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.glassfish.jersey.client;
+
+import static io.opentelemetry.javaagent.instrumentation.jaxrsclient.v2_0.JerseyClientUtil.handleException;
+
+import javax.ws.rs.ProcessingException;
+import org.glassfish.jersey.process.internal.RequestScope;
+
+// implemented interface is package private so wrapper needs to be in the same package
```
iNikem, line 13: What about module system and split packages? Or we are agent and ignore all that?

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/org/glassfish/jersey/client/ResponseCallbackWrapper.java
```diff
@@ -0,0 +1,40 @@
...
+public class ResponseCallbackWrapper implements ResponseCallback {
```
trask, line 14: so that it's more obvious that this is a class we are injecting into this package, and doesn't look like a regular jersey class
```suggestion
public class OpenTelemetryResponseCallbackWrapper implements ResponseCallback {
```

trask (approved):

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-jersey-2.0/javaagent/src/main/java/org/glassfish/jersey/client/ResponseCallbackWrapper.java
laurit: I believe that split packages are not an issue, these are verified when module is defined, defining extra classes shouldn't really change anything in that regard. Sealed packages and singed jars could have issues with this though, but it seems really unlikely that someone would sign jersey jars.

trask: we do this in one other place that I'm aware of

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/d1ff1e12acdc6c239faea5e2100fb514c46618c2/instrumentation/rxjava/rxjava-1.0/library/src/main/java/rx/__OpenTelemetryTracingUtil.java#L6-L16

PR #2811, opened 2021-04-16
Async @WithSpan Instrumentation for Guava ListenableFuture

HaloFour:
Adds AsyncSpanEndStrategy that ends the span for a method annotated with `@WithSpan` when the returned Guava ListenableFuture is completed.


instrumentation/guava-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/guava/InstrumentationHelper.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.guava;
+
+import io.opentelemetry.instrumentation.api.tracer.async.AsyncSpanEndStrategies;
+import io.opentelemetry.instrumentation.guava.GuavaAsyncSpanEndStrategy;
+
+public class InstrumentationHelper {
```
mateuszrzeszutek, line 11: Nit: can you make it final and add a private constructor?

instrumentation/guava-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/guava/InstrumentationHelper.java
```diff
@@ -0,0 +1,17 @@
...
+  static {
+    AsyncSpanEndStrategies.getInstance().registerStrategy(GuavaAsyncSpanEndStrategy.INSTANCE);
+  }
+
+  public static void initialize() {}
```
mateuszrzeszutek, line 16: Please add a comment explaining why it is important to call this method even if it does nothin (static initializer on class load)

mateuszrzeszutek (approved):

iNikem (approved):

PR #2812, opened 2021-04-16
Test failing async http request

laurit:
Verifies that context is propagated to error callback of async request.


trask (approved):
nice 👍

mateuszrzeszutek (approved):

PR #2814, opened 2021-04-19
Better server span name for Grails and Wicket

trask:
The main change in this PR is adding a flag in `ServletSpanNaming` to prevent controller instrumentation from setting the SERVER span name more than once.

Otherwise, when the servlet request is forwarded to an error controller, the error controller is overwriting the SERVER span name that was set by the controller that handled the request.

This PR only applies this change to Spring MVC, Grails, Wicket and Struts instrumentation initially. Will follow up with issue or PR to apply this change to the remaining.


instrumentation/servlet/servlet-3.0/library/src/main/java/io/opentelemetry/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
```diff
@@ -116,16 +111,23 @@ private static MappingResolver getMappingResolver(
...
+  private static void updateServerSpanName(
+      Context context, Object servletOrFilter, HttpServletRequest request) {
+    Span serverSpan = ServerSpan.fromContextOrNull(context);
+    if (serverSpan != null) {
+      ServletSpanNaming servletSpanNaming = ServletSpanNaming.from(context);
+      if (servletSpanNaming.shouldServletUpdateServerSpanName()) {
+        String spanName = getSpanNameFromPath(servletOrFilter, request);
+        if (spanName != null) {
+          serverSpan.updateName(spanName);
+          servletSpanNaming.setServletUpdatedServerSpanName();
```
mateuszrzeszutek, line 127: This triple-if pattern can be seen in several different instrumentations. WDYT about adding the following method to `ServletSpanNaming`:

```java
void updateServerSpanNameByServlet(Supplier<String> serverSpanName) {
  Span serverSpan = ServerSpan.fromContextOrNull(context);
  if (serverSpan == null || !shouldServletUpdateServerSpanName()) {
    return;
  }
  if (serverSpanName != null) {
    serverSpan.updateName(serverSpanName);
    setServletUpdatedServerSpanName();
  }
}
```

and then just call:

```java
ServletSpanNaming.from(context)
    .updateServerSpanNameByServlet(() -> getSpanNameFromPath(servletOrFilter, request));
```

in the instrumentation code?

laurit: I think it is a good idea because every framework will need to make similar checks. Perhaps we could add a `kind` or `priority` to represent that controller can overwrite name from servlet but not vice versa?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServletSpanNaming.java
```diff
@@ -11,7 +11,7 @@
 /**
  * Helper container for tracking whether servlet integration should update server span name or not.
  */
-public class ServletSpanNaming {
+public abstract class ServletSpanNaming {
```
laurit, line 14: Maybe this class should be called `ServerSpanNaming`?

laurit (approved):
👍

instrumentation/servlet/servlet-3.0/library/src/main/java/io/opentelemetry/instrumentation/servlet/v3_0/Servlet3HttpServerTracer.java
mateuszrzeszutek: Well, I thought of copy-pasting this method with s/Servlet/Controller/, but a kind/priority would be fine too.

instrumentation/tomcat/tomcat-7.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/tomcat/v7_0/TomcatHandlerTest.groovy
```diff
@@ -30,6 +30,12 @@ class TomcatHandlerTest extends HttpServerTest<Tomcat> implements AgentTestTrait
     return "/app"
   }
 
+  @Override
+  boolean testNotFound() {
+    // currently span name is /notFound which indicates it won't be low-cardinality
```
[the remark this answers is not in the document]
trask: good point, I removed the servlet mapping for `/notFound` and brought this test back 👍

instrumentation/grails-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grails/GrailsTracer.java
```diff
@@ -23,14 +25,17 @@ public Context startSpan(Object controller, String action) {
...
+  public void updateServerSpanName(Context context, GrailsControllerUrlMappingInfo info) {
+    ServerSpanNaming.updateServerSpanName(
+        context,
+        CONTROLLER,
+        () -> {
+          String action =
+              info.getActionName() != null
+                  ? info.getActionName()
+                  : info.getControllerClass().getDefaultAction();
+          return ServletContextPath.prepend(context, "/" + info.getControllerName() + "/" + action);
```
mateuszrzeszutek, line 37: Nit: how about extracting a private method?

instrumentation/struts-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/struts2/Struts2Tracer.java
```diff
@@ -45,31 +44,29 @@ public Context startSpan(Context parentContext, ActionInvocation actionInvocatio
...
+            } else {
+              result = actionNamespace + "/" + result;
+            }
+          }
+
+          if (!result.startsWith("/")) {
+            result = "/" + result;
+          }
+
+          return ServletContextPath.prepend(context, result);
```
mateuszrzeszutek, line 68: Nit: WDYT about extracting a private method?

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -81,17 +81,11 @@ public Context startSpan(
...
 
     Context context = withServerSpan(parentContext, spanBuilder.startSpan());
-    context = customizeContext(context, request);
     attachServerContext(context, storage);
 
     return context;
   }
 
-  /** Override in subclass to customize context that is returned by {@code startSpan}. */
-  protected Context customizeContext(Context context, REQUEST request) {
```
laurit, line 91: The idea behind this was that it allows modifying context before it gets attached to the carrier so that `RequestDispatcherAdvice` that reads context from request attribute would get the same naming context as the request originally had.

trask: I pulled this code up into `startSpan()` so that `ServletHttpServerTracer` can call `ServerSpanNaming.init()` with `SERVLET` or `CONTAINER` based on whether it is called form a servlet or filter.

The tests haven't all passed yet so not sure if it was a success or no 😄.

trask: oh, ya, we need to add `ServletSpanNaming` to the context before attaching the context to the `STORAGE` param 👍

trask: brought back `customizeContext()` and added a `@Deprecated` method in `ServerSpanNaming` to allow `ServletHttpServerTracer` to update the `source` separately from calling `init` and we can revisit to see if we can do both at once after we migrate to the new Instrumenter API

mateuszrzeszutek (approved):

PR #2816, opened 2021-04-19
Remove forward and include spans

trask:
I don't see much value in these spans.

Proposing to remove them, or disable them by default.


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -554,30 +533,6 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
     }
   }
 
-  void forwardSpan(TraceAssert trace, int index, Object parent, String errorMessage = null, Class exceptionClass = Exception) {
```
anuraaga, line 557: Nice!

mateuszrzeszutek (approved):

trask:
last call, will merge this tomorrow if no other feedback 😄

anuraaga (approved):

mateuszrzeszutek (approved):

PR #2817, opened 2021-04-19
Add annotated parameters as attributes to spans created by Spring WithSpanAspect

HaloFour:
Extends the `WithSpanAspect` in the Spring Autoconfigure instrumentation to add attributes to the newly-created `Span` based on parameters annotated by the `WithSpanAttribute` annotation.

See: https://github.com/open-telemetry/opentelemetry-java/pull/3168

I've probably gone a bit overboard in attempting to support different parameter types and coercing those values into one of the possible attribute types.  Also, the `WithSpanAttributeBinding` instances are intended to support memoization to avoid having to constantly reflect over the method parameters but I have not added the implementation of that memoization.  As presumably this same functionality, if accepted, would be added to the `opentelemetry-annotations-1.0` module the plan should be that the binding code should be abstracted to the `instrumentation-api` module.


HaloFour:
Going to open a fresh PR

PR #2821, opened 2021-04-20
Move install component before BB to install method

pavolloffay:
Originally `installComponentsBeforeByteBuddy` was called from `public static ResettableClassFileTransformer installBytebuddyAgent(
      Instrumentation inst, Iterable<ComponentInstaller> componentInstallers) {` (reverts https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2569/files#diff-c8a0aa00335e8310e53c4f76f12c54d9187c9e3112b28c215525b31f8d0e7a93R86). This is useful as it allows easier workflow in tests - no need to register SPI for the component installer. 


mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -106,6 +105,9 @@ public static void installBytebuddyAgent(Instrumentation inst) {
   public static ResettableClassFileTransformer installBytebuddyAgent(
```
trask, line 105: are you calling this method from tests? I'm trying to understand how this PR is helping you (so that I don't accidentally break you again in the future)

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
pavolloffay: yes I call it from tests

PR #2830, opened 2021-04-20
Populate the InstrumentationVersion.VERSION from a file 

### LINKED ISSUE #2761: Library instrumentation should read its version from a file, not from the jar manifest
Library instrumentation will often get shaded into other jars, and the manifest information will get lost. In that case, the version number of the instrumentation will not end up correct...it will end up being populated with the version of the jar that it is shaded/fat-jarred into. For example, this happens when okhttp3 library instrumentation is used in an Android application.

It would be more robust for library instrumentation to include a resource file that contains the version number, rather than relying on the jar manifest.

breedx-splk:
Read a file within the jar to obtain the version number of instrumentation, instead of reading from the manifest.  The file is generated at build time and contains the project version number.  This works around the case where the agent jar has been shaded into another agent jar, which is the case for android and for vendor distributions.

This fixes #2761

### DIFF
--- a/instrumentation-api/instrumentation-api.gradle
+++ b/instrumentation-api/instrumentation-api.gradle
@@ -35,3 +35,11 @@ dependencies {
   testImplementation deps.awaitility
   testImplementation deps.opentelemetrySdkTesting
 }
+
+task createVersionFile {
+  def dir = 'src/main/resources/io/opentelemetry/instrumentation/api'
+  file(dir).mkdirs()
+  new File(projectDir, "${dir}/instrumentation-version.txt").text = "${project.version}"
+}
+
+jar.dependsOn createVersionFile
--- a/instrumentation-api/.gitignore
+++ b/instrumentation-api/.gitignore
@@ -1,0 +1,1 @@
+src/main/resources/instrumentation-version.txt
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/InstrumentationVersion.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/InstrumentationVersion.java
@@ -5,7 +5,29 @@
 
 package io.opentelemetry.instrumentation.api;
 
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class InstrumentationVersion {
-  public static final String VERSION =
-      InstrumentationVersion.class.getPackage().getImplementationVersion();
+  private static final Logger log = LoggerFactory.getLogger(InstrumentationVersion.class);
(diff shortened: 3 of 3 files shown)


instrumentation-api/instrumentation-api.gradle
```diff
@@ -35,3 +35,11 @@ dependencies {
   testImplementation deps.awaitility
   testImplementation deps.opentelemetrySdkTesting
 }
+
+task createVersionFile {
+  def dir = 'src/main/resources/io/opentelemetry/instrumentation/api'
+  file(dir).mkdirs()
+  new File(projectDir, "${dir}/instrumentation-version.txt").text = "${project.version}"
```
jkwatson, line 42: does this file get bundled into the instrumentation library that uses the API, somehow? Otherwise, this will only provide the version of the API that the instrumentation library is using, not the version of itself.

breedx-splk: This file gets bundled into the `opentelemetry-instrumentation-api-<version>.jar`.

All the instrumentation libraries have the same version, which is the same version as the project which is the same version as the API.  The instrumentation itself was naming tracers with `InstrumentationVersion.VERSION`.  The instrumentation will pick this up because the instrumentation depends on this api.

I don't think that instrumentation is versioned independently....at least not yet?

jkwatson: It's not, but someone could be using various versions of the library instrumentations, if they aren't keeping their versions in sync with each other (or something is coming in via a dependency). 

So, imagine I'm using library Y, which happens to use library X, along with some otel library instrumentation for X. That library instrumentation might be version 1.5.4 and depends transitively on version 1.5.4 of the instrumentation api.

I'm also using library Z, which happens to use library W, along with otel library instrumentation for W. That library instrumentation might be version 1.6.3 and depends transitively on version 1.6.3 of the instrumentation api.

At runtime, only one instrumentation API library will end up resolved onto the classpath, but both library instrumentation would now report with that version, rather than actual version. This will end up being extremely confusing for debugging issues based on the resulting telemetry.

Does that make sense?

breedx-splk: Your scenario makes sense, but reads to me like a different situation than was described in #2761.  

Is your expectation that both library X and library W, that were compiled against and depend on different versions of the telemetry, can run happily within the same application?  If so, is there a way for that to even be safe without the libs X and W shading the instrumentation?.

breedx-splk:
Sounds like this was not at all what was expected.  Closing until #2761 can be clarified.

instrumentation-api/instrumentation-api.gradle
jkwatson: If they're using compatible versions of the API (i.e. the same major version) then I think they should definitely be able to work together. Are you thinking there's some reason why they couldn't?

breedx-splk: I would expect the api to have changed between those versions and I would not expect a compatibility guarantee.

jkwatson: Once the API has been marked as stable, there will be compatibility guarantees. Major versions will be backwards compatible with themselves.

Issue #2820, opened 2021-04-19
smoke-test (windows-latest, wildfly) sporadically times out


trask:
Looks like we may need to bump the timeout more: https://gradle.com/s/npcmvqdj4e3a2

laurit:
I suspect that increasing timeout won't help for this.

Issue #2785, opened 2021-04-12
Glassfish smoke test sporadically hangs and times out


trask:
I've seen this a couple of times.

Gradle scan: https://scans.gradle.com/s/asjqhovfpsjw2

trask:
This one appears to be a different issue, but also leads to timeout: https://scans.gradle.com/s/xx26fgmlqp5ky

```

GlassFishSmokeTest > 5.2020.6 smoke test on JDK 8 FAILED |  
-- | --
  | java.net.ConnectException: Failed to connect to localhost/0:0:0:0:0:0:0:1:50616

```

laurit:
With debug logging https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/2410101447?check_suite_focus=true
Last lines on main thread 
```
main] DEBUG io.opentelemetry.javaagent.tooling.InstrumentationModule - Applying instrumentation: executor -- io.opentelemetry.javaagent.instrumentation.javaconcurrent.ExecutorInstrumentationModule on sun.misc.Launcher$AppClassLoader@18b4aac2
[main] DEBUG io.opentelemetry.javaagent.tooling.AgentInstaller$TransformLoggingListener - Transformed org.glassfish.hk2.utilities.cache.LRUHybridCache$OriginThreadAwareFuture$1 -- sun.misc.Launcher$AppClassLoader@18b4aac2
```
I guess we'll need to dump thread stacks to see whether it has deadlocked.

PR #2831, opened 2021-04-20
Retry smoke tests

trask:
Mitigates #2820 and #2785. I propose to also close them if this works, and there are no objections.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -84,7 +84,11 @@ jobs:
         env:
           S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.S3_BUILD_CACHE_ACCESS_KEY_ID }}
           S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
-        run: ./gradlew :smoke-tests:test -PsmokeTestSuite=${{ matrix.suite }}
+        uses: nick-invision/retry@v2.4.0
+        with:
+          timeout_minutes: 45
+          max_attempts: 3
+          command: ./gradlew :smoke-tests:test -PsmokeTestSuite=${{ matrix.suite }}
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


mateuszrzeszutek (approved):

anuraaga (approved):

laurit:
These tests are already retried. From https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2820 https://scans.gradle.com/s/npcmvqdj4e3a2/tests test is executed 5 times and still fails. Here is another scan where it fails the same way https://scans.gradle.com/s/6cc22rgkbl7pi/tests In both of these the failing server is 13.0.0.Final JDK 8 which makes me wonder whether this is an actual bug. It looks like no traces are received, but as there is very little info it seems to be impossible to tell what actually went wrong. Perhaps it would be better to add a bit more logging to these, `LoggingSpanExporter` output that is present in regular tests could tell us whether spans were created at all. To be honest I don't think we can get these tests to run stably by adding more retries, it might be better to temporarily remove the retries at least from the nightly build so we could get a better idea about which tests are flaky so we could fix them or add enough logging to reveal the cause of failures. If this does not work out well then we can always add the retry back. Initially we could keep retry for pr builds so that it wouldn't disrupt people. With the current retries we don't really know how likely this test is to fail, it might always require 3 retries to pass (if it were so then debugging it would be a lot easier). Also it might be good idea to make nightly build always run tests and not use the cached results.

trask:
> These tests are already retried.

When they are re-tried, does it restart the containers? I'm wondering if retrying via the github action will have a better chance of succeeding since it should restart more stuff(?)

> Also it might be good idea to make nightly build always run tests and not use the cached results.

👍👍

laurit:
From the test output it looks like it restarts the server. I think it reruns the whole test class, so when one server version fails all will be rerun. I have no idea whether retry from github action would help. I guess it restarts the gradle process, idk whether that helps.

trask:
Closing this in favor of the work @laurit is doing to find the root cause of the failures

Issue #2819, opened 2021-04-19
io.reactivex.exceptions.UndeliverableException using v1.1.0 auto instrumentation


necrolyte2:
**Describe the bug**
We are trying out the new `v1.1.0` auto instrumentation after using the `v1.0.1` and it is throwing this exception
```
io.reactivex.exceptions.UndeliverableException: The exception could not be delivered to the consumer because it has already canceled/disposed the flow or the exception has nowhere to go to begin with. Further reading: https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#error-handling | java.lang.NoSuchFieldError: qs
        at io.reactivex.plugins.RxJavaPlugins.onError(RxJavaPlugins.java:367)
        at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:69)
        at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
        at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.NoSuchFieldError: qs
        at io.opentelemetry.javaagent.shaded.instrumentation.rxjava2.TracingObserver.requestFusion(TracingObserver.java:67)
        at io.reactivex.internal.operators.observable.ObservableConcatMap$ConcatMapDelayErrorObserver.onSubscribe(ObservableConcatMap.java:322)
        at io.reactivex.internal.observers.BasicFuseableObserver.onSubscribe(BasicFuseableObserver.java:66)
        at io.reactivex.internal.operators.observable.ObservableFromArray.subscribeActual(ObservableFromArray.java:31)
        at io.reactivex.Observable.subscribe(Observable.java:12284)
        at io.reactivex.internal.operators.observable.ObservableConcatMap.subscribeActual(ObservableConcatMap.java:55)
        at io.reactivex.Observable.subscribe(Observable.java:12284)
[...]

PR #2836, opened 2021-04-21
Fix rxjava2 NoSuchFieldError

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2819


mateuszrzeszutek (approved):
`TracingConditionalSubscriber` and `TracingSubscriber` seem to suffer from the same problem - could you update them too?
Thanks!

laurit:
@mateuszrzeszutek they have a different super class where the field wasn't renamed. I guess this is because in there `qs` has type `QueueSubscription` but in `BasicFuseableObserver` it is `QueueDisposable`. Unless I missed something they should work as they are.

mateuszrzeszutek:
Oh, that's great then -- I just remembered that I had to change those two the last time I was dealing with field renames (`actual` -> `downstream`)

trask (approved):
thx!

PR #2845, opened 2021-04-22
Akka: shut down executor when test is done

laurit:
This should make `AkkaExecutorInstrumentationTest` pass with strict context check.


mateuszrzeszutek (approved):

anuraaga (approved):
💯

PR #2847, opened 2021-04-22
Prevent AgentClassLoader from loading ContextStorageOverride

laurit:
As far as I can tell `ContextStorageOverride` should be only used by library instrumentation so it might be cleaner to exclude that class from agent, but it would still be possible to add `ContextStorageOverride` to bootclasspath and that would affect grpc inside the agent. Perhaps `AgentClassLoader` should load in child first oder to prevent accidentally loading stuff from boot loader?


anuraaga (approved):
This seems nicer than stripping out of the jar, etc. Thanks

mateuszrzeszutek (approved):

PR #2849, opened 2021-04-22
Fix flaky smoke tests

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2820 Class transformations sometimes fail with `ClassCircularityError: java/lang/reflect/Proxy` due to which instrumentation is not applied and spans are not generated. I was able to reproduce `ClassCircularityError` by commenting out some agent code, but I wasn't able to reproduce the exact same behaviour as in failing tests. Maybe it requires something that isn't deterministic to happen in specific order (depends on iteration order of `HashMap` or something like that). It definitely depends of timing because `Proxy` is loaded from grpc background thread and this can happen before or after byte buddy has been set up.

Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2785 `JAVA_TOOL_OPTIONS` is also applied to a sub process that is started from glassfish launcher. My hypothesis is that due to our agent this sub process produces more output than it would normally. Subprocess blocks because it has filled its output buffer and as the output of this process is read only after it has finished it will hang until build times out. Enabling debug logging for our agent makes this issue reproducible.

This pr also enable debug logging for smoke tests so that such issues would be easier to debug in the future.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -25,6 +25,7 @@ import io.opentelemetry.javaagent.tooling.config.ConfigInitializer;
 import io.opentelemetry.javaagent.tooling.context.FieldBackedProvider;
 import io.opentelemetry.javaagent.tooling.matcher.GlobalClassloaderIgnoresMatcher;
 import java.lang.instrument.Instrumentation;
+import java.lang.reflect.Proxy;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -82,6 +83,9 @@ public class AgentInstaller {
     BootstrapPackagePrefixesHolder.setBoostrapPackagePrefixes(loadBootstrapPackagePrefixes());
     // this needs to be done as early as possible - before the first Config.get() call
     ConfigInitializer.initialize();
+    // ensure java.lang.reflect.Proxy is loaded, as transformation code uses it internally
+    // loading it after bytebuddy transformer is set up can cause a ClassCircularityError
+    Proxy.class.getName();
   }
 
   public static void installBytebuddyAgent(Instrumentation inst) {
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -84,7 +84,11 @@ public class AgentInstaller {
     // this needs to be done as early as possible - before the first Config.get() call
     ConfigInitializer.initialize();
     // ensure java.lang.reflect.Proxy is loaded, as transformation code uses it internally
-    // loading it after bytebuddy transformer is set up can cause a ClassCircularityError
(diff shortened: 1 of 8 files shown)


trask (approved):

iNikem (approved):

iNikem:
@laurit Do you think this will solve #2820?

laurit:
@iNikem Yes. Though with such issues it always depends on luck. It might be that there are more ways that this can happen or I misunderstood the issue. Only letting it run for a while will show whether this change really fixes this.

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -82,6 +83,9 @@ public static Instrumentation getInstrumentation() {
     BootstrapPackagePrefixesHolder.setBoostrapPackagePrefixes(loadBootstrapPackagePrefixes());
     // this needs to be done as early as possible - before the first Config.get() call
     ConfigInitializer.initialize();
+    // ensure java.lang.reflect.Proxy is loaded, as transformation code uses it internally
+    // loading it after bytebuddy transformer is set up can cause a ClassCircularityError
+    Proxy.class.getName();
```
trask, line 88: do you have a stack trace handy for the `ClassCircularityError`?

laurit: https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/2447546557
https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/2447546469

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -82,6 +83,9 @@ public static Instrumentation getInstrumentation() {
     BootstrapPackagePrefixesHolder.setBoostrapPackagePrefixes(loadBootstrapPackagePrefixes());
     // this needs to be done as early as possible - before the first Config.get() call
     ConfigInitializer.initialize();
+    // ensure java.lang.reflect.Proxy is loaded, as transformation code uses it internally
+    // loading it after bytebuddy transformer is set up can cause a ClassCircularityError
```
trask, line 87:
```suggestion
    // ensure java.lang.reflect.Proxy is loaded, as transformation code uses it internally
    // loading java.lang.reflect.Proxy after the bytebuddy transformer is set up causes
    // the internal-proxy instrumentation module to transform it, and then the bytebuddy
    // transformation code also tries to load it, which leads to a ClassCircularityError
    // loading java.lang.reflect.Proxy early here still allows it to be retransformed by the
    // internal-proxy instrumentation module after the bytebuddy transformer is set up
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
trask: thx for tracking this down! looks like I introduced the problem in #2773

trask (approved):

mateuszrzeszutek (approved):

Issue #2776, opened 2021-04-12
Auto-manual instrumentations integrations in "reactive" frameworks


iNikem:
We [have documentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/manual-instrumentation.md#adding-attributes-to-the-current-span) about how manual instrumentation can add extra attributes to a span, created by agent's auto-instrumentation. This works sufficiently well in simple, "single threaded" cases. But if I want to communicate with spans created for more "reactive" frameworks, such as Reactor Netty or Spring WebFlux, how can I be sure that `Span.current()` will return a correct span to me?

E.g. Reactor Netty provides a lot of callbacks (`doOnResolve`, `doAfterResolve`, `doAfterRequest`). It does not seem that currently those callbacks are executed with the right Otel context.

I can reach to the Netty connection and grab `AttributeKey.valueOf("io.opentelemetry.javaagent.shaded.instrumentation.netty.v4_1.AttributeKeys#client-context")` from that, if I know with agent's implementation details, but this is certainly a fragile and a bad way forward.

PR #2850, opened 2021-04-22
Expose correct Context.current() in reactive-netty callbacks

mateuszrzeszutek:
And make sure to clean up channel attributes in netty 4.1 instrumentation.

`doOnResolve`, `doAfterResolve`, `doOnResolveError`, `doOnConnected` and `doOnDisconnected` are not covered because they're called either too early or too late and getting any context from the connection in one of them is impossible.

I think it closes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2776 - I've checked all HTTP clients that we instrument and that seemed "reactive" to me and none of them allow to set listeners like reactor-netty does (or none at all).

### DIFF
--- a/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
+++ b/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
@@ -25,26 +25,32 @@ public class HttpClientResponseTracingHandler extends ChannelInboundHandlerAdapt
 
   @Override
   public void channelRead(ChannelHandlerContext ctx, Object msg) {
-    Context context = ctx.channel().attr(AttributeKeys.CLIENT_CONTEXT).get();
+    Attribute<Context> clientContextAttr = ctx.channel().attr(AttributeKeys.CLIENT_CONTEXT);
+    Context context = clientContextAttr.get();
     if (context == null) {
       ctx.fireChannelRead(msg);
       return;
     }
 
+    Attribute<Context> parentContextAttr = ctx.channel().attr(AttributeKeys.CLIENT_PARENT_CONTEXT);
+    Context parentContext = parentContextAttr.get();
+
     if (msg instanceof FullHttpResponse) {
       tracer().end(context, (HttpResponse) msg);
+      clientContextAttr.remove();
+      parentContextAttr.remove();
     } else if (msg instanceof HttpResponse) {
       // Headers before body have been received, store them to use when finishing the span.
       ctx.channel().attr(HTTP_RESPONSE).set((HttpResponse) msg);
     } else if (msg instanceof LastHttpContent) {
       // Not a FullHttpResponse so this is content that has been received after headers. Finish the
       // span using what we stored in attrs.
-      tracer().end(context, ctx.channel().attr(HTTP_RESPONSE).get());
+      tracer().end(context, ctx.channel().attr(HTTP_RESPONSE).getAndRemove());
+      clientContextAttr.remove();
+      parentContextAttr.remove();
     }
 
     // We want the callback in the scope of the parent, not the client span
-    Attribute<Context> parentAttr = ctx.channel().attr(AttributeKeys.CLIENT_PARENT_CONTEXT);
-    Context parentContext = parentAttr.get();
     if (parentContext != null) {
       try (Scope ignored = parentContext.makeCurrent()) {
         ctx.fireChannelRead(msg);
--- a/instrumentation/reactor-netty/reactor-netty-0.9/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
+++ b/instrumentation/reactor-netty/reactor-netty-0.9/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
@@ -3,10 +3,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+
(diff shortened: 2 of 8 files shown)


instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/v1_0/DecoratorFunctions.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static Context getChannelContext(ContextView contextView, @Nullable Channel channel) {
+    Context context = null;
+    // try to get the client span context from the channel if it's available
+    if (channel != null) {
+      context = channel.attr(AttributeKeys.CLIENT_CONTEXT).get();
+    }
+    // otherwise use the parent span context
+    if (context == null) {
+      context = contextView.get(ReactorNettyInstrumentationModule.MapConnect.CONTEXT_ATTRIBUTE);
```
jkwatson, line 67: how about just returning here, rather than assigning, then returning later? If you invert the if check, I also think it will read better. Like:

```
if (context != null) {
  return context;
} 

return contextView.get(...);
```

instrumentation/reactor-netty/reactor-netty-0.9/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
```diff
@@ -52,4 +58,51 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
...
+        def nettyClientSpan = span(1)
+
+        basicSpan(it, 0, "parent")
+        clientSpan(it, 1, parentSpan, "GET", server.address.resolve("/success"))
+        serverSpan(it, 2, nettyClientSpan)
+
+        assertSameSpan(parentSpan, onRequestSpan)
+        assertSameSpan(nettyClientSpan, afterRequestSpan)
+        assertSameSpan(nettyClientSpan, onResponseSpan)
+        assertSameSpan(nettyClientSpan, afterResponseSpan)
```
trask, line 97: I like the reactor-netty-1.0 behavior better where `afterResponseSpan` uses the `parentSpan` instead of the `nettyClientSpan`, as it seems to match our general practice of response handlers being outside of the client span, and running in the parent context

if that's not something easy to change in 0.9 that's ok, would be good to add comment somewhere about this difference

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
```diff
@@ -52,4 +61,87 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
...
+
+        assertSameSpan(parentSpan, onRequestSpan)
+        assertSameSpan(nettyClientSpan, afterRequestSpan)
+        assertSameSpan(nettyClientSpan, onResponseSpan)
+        assertSameSpan(parentSpan, afterResponseSpan)
+      }
+    }
+  }
+
+  def "should expose context to http request error callback"() {
```
trask, line 105: can this test be added to 0.9? (main reason i'm interested is to see which span it uses, client or parent)

trask (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientResponseTracingHandler.java
```diff
@@ -25,26 +25,32 @@
 
   @Override
   public void channelRead(ChannelHandlerContext ctx, Object msg) {
-    Context context = ctx.channel().attr(AttributeKeys.CLIENT_CONTEXT).get();
+    Attribute<Context> clientContextAttr = ctx.channel().attr(AttributeKeys.CLIENT_CONTEXT);
+    Context context = clientContextAttr.get();
```
iNikem, line 29: Can we use `getAndRemove` here?

mateuszrzeszutek: I can't unfortunately: we can't remove/end the context unless the whole response was received, see this part:

```java
    } else if (msg instanceof HttpResponse) {
      // Headers before body have been received, store them to use when finishing the span.
      ctx.channel().attr(HTTP_RESPONSE).set((HttpResponse) msg);
    }
```

The span is finished in another `HttpClientResponseTracingHandler` call (`LastHttpContent`) that comes later.

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
mateuszrzeszutek: Hmm, weird - I remember that I had some problems with that test on 0.9, I remember being convinced that this test made absolutely no sense in 0.9... but now that I copied it from 1.0 it just passed without any changes 😅

instrumentation/reactor-netty/reactor-netty-0.9/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
mateuszrzeszutek: Done.

Issue #2705, opened 2021-04-03
Netty instrumentation causes memory leak and more


raptium:
**Describe the bug**
`io.netty.channel.ChannelFuture#addListener()` is transformed to wrap the input listener with a helper class.
However the `removeListener()` `removeListeners()` counterparts are never transformed. Once a listener is added to the future, it can never be removed. This may cause memory leak and even logical incorrectness.

**Steps to reproduce**
```java
import io.netty.channel.embedded.EmbeddedChannel;
import io.netty.util.concurrent.Future;
import io.netty.util.concurrent.GenericFutureListener;

public class NettyChannelFutureDemo {
    public static void main(String[] args) {
        EmbeddedChannel channel = new EmbeddedChannel();
        for (int i = 0; i < 100; i++) {
            GenericFutureListener<Future<Object>> listener = future -> {
                System.out.println("channel closed");
            };
            channel.closeFuture().addListener(listener);
            channel.closeFuture().removeListener(listener); // not removed if netty instrumentation is on
        }
        try {
            channel.close().await();
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
    }
}
```

**What did you expect to see?**
`channel closed` should never be printed since listeners are removed before the future completes.

**What did you see instead?**
`channel closed` is printed 100 times.

**What version are you using?**
`v1.0.1`

**Environment**
Compiler: AdoptOpenJDK 11.0.10
OS: Microsoft Windows 10

**Additional context**
N/A

dkulig:
Hi, can I get some sort of an ETA when this will be released?

iNikem:
We plan to release new version of the agent during the next several days.

dkulig:
@iNikem thanks for the update. I can see that [v1.2.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.2.0) has been released however this guy seems to not be included in the Bug fixes. Does that mean it will be included in the next release?

mateuszrzeszutek:
Oh, the fix is there, but the note does not say anything about memleak explicitly:

> * Netty 4.1 listeners could not be removed by application (#2851)

dkulig:
thanks @mateuszrzeszutek

PR #2851, opened 2021-04-22
Fix netty 4.1 instrumentation not removing future listeners

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2705


instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/WrappedFutureListener.java
```diff
@@ -9,18 +9,41 @@
 import io.netty.util.concurrent.GenericFutureListener;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
+import io.opentelemetry.instrumentation.api.caching.Cache;
+
+public final class WrappedFutureListener<F extends Future<? super Void>>
```
laurit, line 14: Maybe a simpler signature like `WrappedFutureListener implements GenericFutureListener<Future<?>>` would also work.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureInstrumentation.java
```diff
@@ -35,18 +38,74 @@
...
   }
 
   public static class AddListenerAdvice {
     @Advice.OnMethodEnter
     public static void wrapListener(
-        @Advice.Argument(value = 0, readOnly = false) GenericFutureListener listener) {
-      listener = new WrappedFutureListener(Java8BytecodeBridge.currentContext(), listener);
+        @Advice.Argument(value = 0, readOnly = false)
+            GenericFutureListener<? extends Future<? super Void>> listener) {
+      listener = WrappedFutureListener.wrap(Java8BytecodeBridge.currentContext(), listener);
```
laurit, line 66: I also looked at this briefly a few days ago. One thing that seemed problematic was that netty internally does instanceof on the listener type for example `progressiveListeners` in https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/DefaultPromise.java checks for `GenericProgressiveFutureListener`. Perhaps we should create a special wrapper for these to make instanceof pass. Hopefully there aren't more of these special listener types.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/WrappedFutureListener.java
```diff
@@ -9,18 +9,41 @@
...
+    implements GenericFutureListener<F> {
+
+  private static final Cache<GenericFutureListener<?>, WrappedFutureListener<?>> wrappers =
```
anuraaga, line 17: This seems like it can be `InstrumentationContext` instead of a map right?

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureInstrumentation.java
mateuszrzeszutek: Oh, that's an excellent catch 👍  Will implement.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/WrappedFutureListener.java
mateuszrzeszutek: Yeah, I thought about it too - I don't know how it'd work with lambdas or method references, and how costly adding a field to lots of listener classes would be. (It's a good idea for a benchmark though)

laurit (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/WrappedFutureListener.java
anuraaga: But isn't the weakmap a subset of InstrumentationContext? We use weakmap behind the scenes when not being able to add a field, and I figure when we can add the field it should basically work better.

laurit: One problem with InstrumentationContext is that it is incapable of handling cases when the same object is used in different contexts e.g. same runnable is submitted to executor multiple times. With wrappers it should be possible to handle such cases, though the current solution doesn't.

mateuszrzeszutek: Okay, changed to `InstrumentationContext`.

> One problem with InstrumentationContext is that it is incapable of handling cases when the same object is used in different contexts e.g. same runnable is submitted to executor multiple times.

Hmm, I guess that it's a separate `InstrumentationContext` problem that applies to the whole javaagent. We'd probably like to get it solved separately.

anuraaga (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/WrappedFutureListener.java
trask: > One problem with InstrumentationContext is that it is incapable of handling cases when the same object is used in different contexts e.g. same runnable is submitted to executor multiple times.

Oh, that's an interesting problem, hadn't thought about that... have you seen this cause issues in real world apps (yet)?

PR #2858, opened 2021-04-23
fix db.operation in semantic-conventions.md

tydhot:
I have found all spans provided by JDBC instrumentation have the attribute `db.operation` and I check code found is actually provided . But according to the doc, JDBC is not support to add this attribute. Maybe the doc need to be fixed.

### DIFF
--- a/docs/semantic-conventions.md
+++ b/docs/semantic-conventions.md
@@ -76,7 +76,7 @@ not values defined by spec.
 | `db.mssql.instance_name`   | N | - |
 | `db.name`   | N | only set of JDBC, Mongo, Geode and MongoDB |
 | `db.statement`   | N | +, except for ElasticSearch and Memcached, see `db.operation` |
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):
👍

docs/semantic-conventions.md
```diff
@@ -76,7 +76,7 @@ not values defined by spec.
 | `db.mssql.instance_name`   | N | - |
 | `db.name`   | N | only set of JDBC, Mongo, Geode and MongoDB |
 | `db.statement`   | N | +, except for ElasticSearch and Memcached, see `db.operation` |
-| `db.operation`   | N | only set of ElasticSearch and Memcached |
+| `db.operation`   | N | only set of ElasticSearch, Memcached and JDBC |
```
trask, line 79:
```suggestion
| `db.operation`   | N | only set for ElasticSearch, Memcached and JDBC |
```

trask (approved):
thx!

PR #2860, opened 2021-04-23
Instrumenter API improvements

mateuszrzeszutek:
* Move HTTP & net classes to separate packages
* Remove `db` prefix from method names in `DbAttributesExtractor`
* Add request-only net attributes extractor (it'll be needed in spring-sleuth-otel once we decide to use Instrumenters there)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanNameExtractor.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.http;
+
+import io.opentelemetry.instrumentation.api.instrumenter.SpanNameExtractor;
+
+public final class HttpSpanNameExtractor<REQUEST> implements SpanNameExtractor<REQUEST> {
```
anuraaga, line 10: The tradeoff vs not having separate packages :) Now that the class is public let's add the javadoc

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetRequestAttributesExtractor.java
```diff
@@ -0,0 +1,66 @@
...
+ * attributes</a> from a {@link InetSocketAddress}. Most network libraries will provide access to a
+ * {@link InetSocketAddress} so this is a convenient alternative to {@link
+ * NetRequestAttributesExtractor}. There is no meaning to implement both in the same
+ * instrumentation.
+ *
+ * <p>This extractor implementation uses both {@code request} and {@code response} to extract
+ * network peer information. If those attributes are available in the request, it is recommended to
+ * use {@link InetSocketAddressNetAttributesExtractor} instead.
+ */
+public abstract class InetSocketAddressNetRequestAttributesExtractor<REQUEST, RESPONSE>
```
anuraaga, line 24: Didn't realize this would be needed already :)

I have considered an approach where there is only one extractor, and `@Nullable RESPONSE response` - if it's available when response is null (onStart), then it's gotten from request, otherwise gotten when response isn't null (onEnd). The advantage of two classes is it's easier to implement them. The disadvantage is having cognitive load deciding which class to use. I think I'm leaning towards the separate classes actually but wanted to get your thoughts too.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/NetAttributesExtractor.java
```diff
@@ -26,10 +31,6 @@ protected final void onStart(AttributesBuilder attributes, REQUEST request) {
   @Override
   protected final void onEnd(AttributesBuilder attributes, REQUEST request, RESPONSE response) {
     set(attributes, SemanticAttributes.NET_PEER_IP, peerIp(request, response));
-
-    // TODO(anuraaga): Clients don't have peer information available during the request usually.
```
anuraaga, line 30: But related to my above comment, actually this isn't quite solved since it means instrumentation defining separate ServerNetAttributesExtractor and ClientNetAttributesExtractor, which our pattern doesn't really support since it makes instrumentation more complex.  The approach with `@Nullable RESPONSE response` could solve it though.

mateuszrzeszutek: I've refactored the `NetAttributesExtractor` class so that it has a `@Nullable` response in its attribute methods.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanNameExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+  private HttpSpanNameExtractor(HttpAttributesExtractor<REQUEST, ?> attributesExtractor) {
+    this.attributesExtractor = attributesExtractor;
+  }
+
+  @Override
+  public String extract(REQUEST request) {
+    String route = attributesExtractor.route(request);
+    if (route != null) {
+      return route;
+    }
```
trask, line 38: 👍

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanNameExtractorTest.java
```diff
@@ -28,20 +29,20 @@
   void routeAndMethod() {
     when(extractor.route(anyMap())).thenReturn("/cats/{id}");
     when(extractor.method(anyMap())).thenReturn("GET");
-    assertThat(SpanNameExtractor.http(extractor).extract(Collections.emptyMap()))
+    Assertions.assertThat(HttpSpanNameExtractor.create(extractor).extract(Collections.emptyMap()))
```
trask, line 32:
```suggestion
    assertThat(HttpSpanNameExtractor.create(extractor).extract(Collections.emptyMap()))
```

trask (approved):

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanStatusExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+
+  private HttpSpanStatusExtractor(HttpAttributesExtractor<REQUEST, RESPONSE> attributesExtractor) {
+    this.attributesExtractor = attributesExtractor;
+  }
+
+  @Override
+  public StatusCode extract(REQUEST request, RESPONSE response, Throwable error) {
+    Long statusCode = attributesExtractor.statusCode(request, response);
+    if (statusCode != null) {
+      return HttpStatusConverter.statusFromHttpStatus((int) (long) statusCode);
```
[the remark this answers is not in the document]
mateuszrzeszutek: 🤦‍♂️ 
Eh, automatic refactorings 😂

Issue #2823, opened 2021-04-20
Muzzle incorrectly constructs field references


mateuszrzeszutek:
**Describe the bug**
I've noticed that in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2819 muzzle should have prevented the instrumentation from being applied, but it didn't: a `NoSuchFieldError` was thrown instead.

When you print out the muzzle references you can see that it incorrectly assigns the `qs` field to our helper class:

```
  io.opentelemetry.instrumentation.rxjava2.TracingObserver extends<io.reactivex.internal.observers.BasicFuseableObserver>
    Field: PRIVATE_OR_HIGHER NON_STATIC FieldRef:qsio/reactivex/internal/fuseable/QueueDisposable
  io.reactivex.internal.observers.BasicFuseableObserver
    Source: io.opentelemetry.instrumentation.rxjava2.TracingObserver:0
    Source: io.opentelemetry.instrumentation.rxjava2.TracingObserver:39
    Method: NON_STATIC PROTECTED_OR_HIGHER <init>(Lio/reactivex/Observer;)V
```

And the `BasicFuseableObserver` which actually owns this field only has a reference to its constructor. We don't check field access in our helper classes, so muzzle fails to recognize a mismatch in the runtime.

**What did you expect to see?**
Muzzle should assign the field to the class that actually defines it.

laurit:
For field access `this.qs` compiler generates bytecode where field owner is the type of `this` even when the field is in super class. This behaviour is the same for methods and static fields, owner is whatever the type is that access is performed on, actual field owner is resolved during runtime.
Are you suggesting that we somehow try to figure out the actual owner of field/method?

PR #2870, opened 2021-04-26
Muzzle should check whether used fields are declared somewhere

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2823

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/HelperReferenceWrapper.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/HelperReferenceWrapper.java
@@ -14,6 +14,7 @@ import io.opentelemetry.javaagent.tooling.muzzle.Reference.Flag.VisibilityFlag;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Stream;
+import net.bytebuddy.description.field.FieldDescription;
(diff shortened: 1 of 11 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/HelperReferenceWrapper.java
```diff
@@ -161,6 +200,21 @@ private Method toMethod(Reference.Method method) {
             method.getName(),
             method.getDescriptor());
       }
+
+      @Override
+      public Stream<Field> getFields() {
+        return reference.getFields().stream()
+            .filter(this::isDeclaredAndNotPrivate)
```
laurit, line 207: Why are private fields special? An optimization? Nestmates can access private fields within nest, would that be an issue?

mateuszrzeszutek: It's an optimization. This field check is mostly meant to detect changed/renamed fields when we're extending library classes - nested classes are not a valid scenario in this case. And if we use nested classes in our own instrumentation code the Java compiler will pick up all inconsistencies.

trask (approved):
nice 😅

laurit (approved):

PR #2874, opened 2021-04-26
Pass jvm arguments from env var on windows payara

laurit:
Linux payara image support passing jvm arguments in `JVM_ARGS` env variable, make windows image do the same. This is needed for https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2785 We can't use `JAVA_TOOL_OPTIONS` because during startup payara launches a java process to detect java version and parses its output. As this process also gets our agent from `JAVA_TOOL_OPTIONS` it produces more output than expected which can block that process when it fills its output buffer. The output of that process is read only after it has completed so if it creates too much output it will hang permanently.
Also update windows fake backend java version so that image would get rebuilt agains the latest windows image which should make windows tests a bit faster again.

### DIFF
--- a/smoke-tests/matrix/src/main/docker/payara/launch.bat
+++ b/smoke-tests/matrix/src/main/docker/payara/launch.bat
@@ -1,2 +1,1 @@
-java -jar glassfish/lib/client/appserver-cli.jar start-domain domain1
-powershell -command "Get-Content /server/glassfish/domains/domain1/logs/server.log -Wait"
+java %JVM_ARGS% -jar glassfish/lib/client/appserver-cli.jar start-domain --verbose domain1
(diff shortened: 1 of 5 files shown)


smoke-tests/matrix/src/main/docker/payara/launch.bat
```diff
@@ -1,2 +1 @@
-java -jar glassfish/lib/client/appserver-cli.jar start-domain domain1
-powershell -command "Get-Content /server/glassfish/domains/domain1/logs/server.log -Wait"
+java %JVM_ARGS% -jar glassfish/lib/client/appserver-cli.jar start-domain --verbose domain1
```
mateuszrzeszutek, line 1: `--verbose`?

mateuszrzeszutek (approved):

smoke-tests/matrix/src/main/docker/payara/launch.bat
laurit: `--verbose` basically means start in foreground, pretty much the same as `catalina.sh run`

PR #2877, opened 2021-04-27
Record HTTP server metrics in instrumenter API.

anuraaga:
Initial draft of recording metrics from Instrumenter. Let me know if you have any starter thoughts on this approach.

I found recording the metrics using the metrics API to be pretty annoying, especially having different set of labels for different metrics. I'm hoping the metrics conventions are moving in a direction of recording all the attributes, and letting the SDK deal with such details. That being said, it's still nice that all this is hidden from instrumentation authors, they just have to enable the metrics when constructing.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/RequestMetrics.java
```diff
@@ -0,0 +1,16 @@
...
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.context.Context;
+
+public interface RequestMetrics {
+
+  Context start(Context context, Attributes requestAttributes);
+
+  void end(Context context, Attributes responseAttributes);
```
anuraaga, line 15: My understanding is metrics will only ever be recorded with "request attributes" since metric attributes (labels) must be constant throughout a request.

But some metrics will need response attributes to get the actual value, for example, a "response content length" metric. In other words, some span attributes (usually response attributes) are actually metric values.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/RequestMetricsFactory.java
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.metrics.Meter;
+
+@FunctionalInterface
+public interface RequestMetricsFactory {
```
anuraaga, line 11: Possibly this class could be called `RequestMetrics` and the other `RequestListener`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private final LongUpDownCounter activeRequests;
+  private final DoubleValueRecorder duration;
+
+  private HttpServerRequestMetrics(Meter meter) {
+    activeRequests =
+        meter
+            .longUpDownCounterBuilder("http.server.active_requests")
+            .setUnit("requests")
```
jkwatson, line 39: This makes me think the semantic conventions need to have units in them as well, so we won't have to hard-code this stuff. Also, even though this is currently the semantic convention for these units, I suspect this will not last, as we will probably be standardizing on units in the future. See this issue for tracking: https://github.com/open-telemetry/opentelemetry-specification/issues/705

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+          }
+          switch (key.getKey()) {
+            case "http.method":
+            case "http.host":
+            case "http.scheme":
+            case "http.status_code":
+            case "http.flavor":
+            case "http.server_name":
+            case "net.host.name":
+            case "net.host.port":
```
richardstartin, line 113: I'm curious about OTel's approach to high cardinality metric attributes, having implemented similar functionality in at Datadog and paid a lot of attention to cardinality reduction prior to aggregation. I have a few questions which I hope are not disruptive to your review.
* I'm unfamiliar with OTel attributes but which of these, if any, is the request path? 
* Since the unsanitised request path can have unbounded cardinality, would that rule it out as a metric attribute?
* If this were a metric for a database query, would you sanitise the query here, or exclude it from the labels?

anuraaga: Thanks for reaching out @richardstartin. I doubt I'll have good answers as lots about OTel's approach to metrics is WIP but let's see ;)

> I'm unfamiliar with OTel attributes but which of these, if any, is the request path?

It's not in this list yet. There is a special note here

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#parameterized-labels

and I didn't want to deal with it yet :) In particular, it seems they are not very up to date as `http.route` is really the parameter to be using (it's the parameterized path)

> Since the unsanitised request path can have unbounded cardinality, would that rule it out as a metric attribute?

I think so, it would have to be a parameterized path of some sort. That being said, given the separation between API and SDK in OTel, I wonder if it's possible to pass raw paths in API and in the SDK either detect parameters in the path using some cool logic, or detect the cardinality and drop the attributes. Without that, it's definitely not appropriate. I wonder if your "cardinality reduction prior to aggregation" would be something we can incorporate in the SDK.

> If this were a metric for a database query, would you sanitise the query here, or exclude it from the labels?

If there were conventions for DB metrics I'd just follow them

https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/metrics/semantic_conventions

Lacking them, just from my sense, I've never added queries as metric labels. But I don't see a problem with adding prepared statements as they generally have low cardinality (10-20 different types of queries in an app, each with the same prepared form across invocations).

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
jkwatson: I was thinking the generator should provide them as well. Are they in the yaml yet?

anuraaga: Ah got it - nope there aren't any metric yaml files yet

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
richardstartin: Thanks for your answers. We track metrics at a much more granular level than here - so including the sanitised request path, sanitised database query, and so on - we call this the _resource name_ at Datadog. This is because latency distributions tend to depend on attributes like the query or the endpoint - e.g. database updates and inserts tend to perform differently, and the application's business logic is always slower than the health check endpoint. 

I will be interested to see the direction this API goes in - will higher cardinality attributes be sacrificed altogether to keep the number of points down, or will it find a way to share the sanitisation required to avoid the number of metric points exploding with tracing to reduce total overhead and promote consistency between metrics and tracing?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/RequestMetricsFactory.java
mateuszrzeszutek: Note: if we ever want to implement [this comment](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2596#discussion_r603924647): we could encapsulate all span operations in a `RequestListener`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+  @Override
+  public void end(Context context, Attributes responseAttributes) {
+    State state = context.get(HTTP_SERVER_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset HTTP request metrics.", context);
+      return;
+    }
+    activeRequests.add(-1, state.activeRequestLabels());
+    duration.record(System.nanoTime() - state.startTimeNanos(), state.durationLabels());
```
mateuszrzeszutek, line 73: The metric's unit is `milliseconds`, we probably should convert nanos to millis

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+          }
+        });
+    return labels.build();
+  }
+
+  private static Labels durationLabels(Attributes attributes) {
+    LabelsBuilder labels = Labels.builder();
+    attributes.forEach(
+        (key, value) -> {
+          if (key.getType() != AttributeType.STRING) {
```
mateuszrzeszutek, line 102: `http.status_code` and `net.host.port` are longs, not strings.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
```diff
@@ -0,0 +1,132 @@
...
+          "No state present when ending context {}. Cannot reset HTTP request metrics.", context);
+      return;
+    }
+    activeRequests.add(-1, state.activeRequestLabels());
+    duration.record(System.nanoTime() - state.startTimeNanos(), state.durationLabels());
+  }
+
+  private static Labels activeRequestLabels(Attributes attributes) {
+    LabelsBuilder labels = Labels.builder();
+    attributes.forEach(
```
mateuszrzeszutek, line 78: It'd be great if the SDK included some easy Attributes<->Labels conversion. Or if we could use Attributes directly (well, a subset of them) in metrics API...

anuraaga: Yeah that's what I'm hoping for.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/HttpServerRequestMetrics.java
anuraaga: Hmm - status code isn't ready anyways when I compute these labels at the beginning of the request. I guess depending on the metric, some need to be computed at beginning (active requests) and some at end (duration). But since there's no text in the spec on that yet I'll keep it as is for now.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/RequestMetricsFactory.java
anuraaga: Yeah that seems like a nice idea

anuraaga (commented):
Started going through comments will continue to clean this PR up

anuraaga:
Thanks for waiting think this is ready for review

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -118,14 +120,21 @@ public Context start(Context parentContext, REQUEST request) {
...
+    Context context = parentContext;
+
+    for (RequestListener metrics : requestListeners) {
+      context = metrics.start(context, attributes);
+    }
+
+    attributes.forEach((key, value) -> spanBuilder.setAttribute((AttributeKey) key, value));
     Span span = spanBuilder.startSpan();
-    Context context = parentContext.with(span);
+    context = parentContext.with(span);
```
trask, line 137:
```suggestion
    context = context.with(span);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -21,14 +23,17 @@
...
+  final Meter meter;
   final String instrumentationName;
   final SpanNameExtractor<? super REQUEST> spanNameExtractor;
 
   final List<AttributesExtractor<? super REQUEST, ? super RESPONSE>> attributesExtractors =
       new ArrayList<>();
+  final List<RequestListener> requestListeners = new ArrayList<>();
 
-  SpanKindExtractor<? super REQUEST> spanKindExtractor;
+  SpanKindExtractor<? super REQUEST> spanKindExtractor = null;
```
trask, line 36:
```suggestion
  SpanKindExtractor<? super REQUEST> spanKindExtractor;
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -118,14 +120,21 @@ public Context start(Context parentContext, REQUEST request) {
...
-      extractor.onStart(attributes, request);
+      extractor.onStart(attributesBuilder, request);
     }
-    attributes.build().forEach((key, value) -> spanBuilder.setAttribute((AttributeKey) key, value));
+    Attributes attributes = attributesBuilder.build();
 
+    Context context = parentContext;
+
+    for (RequestListener metrics : requestListeners) {
+      context = metrics.start(context, attributes);
```
trask, line 132:
```suggestion
    for (RequestListener requestListener: requestListeners) {
      context = requestListener.start(context, attributes);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -145,11 +154,17 @@ public Context start(Context parentContext, REQUEST request) {
...
+    AttributesBuilder attributesBuilder = Attributes.builder();
     for (AttributesExtractor<? super REQUEST, ? super RESPONSE> extractor : extractors) {
-      extractor.onEnd(attributes, request, response);
+      extractor.onEnd(attributesBuilder, request, response);
     }
-    attributes.build().forEach((key, value) -> span.setAttribute((AttributeKey) key, value));
+    Attributes attributes = attributesBuilder.build();
+
+    for (RequestListener metrics : requestListeners) {
+      metrics.end(context, attributes);
```
trask, line 164:
```suggestion
    for (RequestListener requestListener: requestListeners) {
      requestListener.end(context, attributes);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java
```diff
@@ -0,0 +1,152 @@
...
+ * {@link RequestListener} which keeps track of <a
+ * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server">HTTP
+ * server metrics</a>.
+ *
+ * <p>To use this class, you may need to add the {@code opentelemetry-api-metrics} artifact to your
+ * dependencies.
+ */
+public final class HttpServerMetrics implements RequestListener {
+
+  private static final double NANOS_IN_MS = TimeUnit.MILLISECONDS.toNanos(1);
```
trask, line 34:
```suggestion
  private static final double NANOS_PER_MS = TimeUnit.MILLISECONDS.toNanos(1);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java
```diff
@@ -0,0 +1,152 @@
...
+    return labels.build();
+  }
+
+  private static Labels durationLabels(Attributes attributes) {
+    LabelsBuilder labels = Labels.builder();
+    attributes.forEach(
+        (key, value) -> {
+          if (key.getType() != AttributeType.STRING) {
+            return;
+          }
```
trask, line 122: looks like there are long attributes below also

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java
```diff
@@ -0,0 +1,152 @@
...
+  public void end(Context context, Attributes attributes) {
+    State state = context.get(HTTP_SERVER_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset HTTP request metrics.", context);
+      return;
+    }
+    activeRequests.add(-1, state.activeRequestLabels());
+    duration.record(
+        (System.nanoTime() - state.startTimeNanos()) / NANOS_IN_MS, state.durationLabels());
```
trask, line 91: is it worth injecting a `Clock`?

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -74,6 +81,12 @@
     return addAttributesExtractors(Arrays.asList(attributesExtractors));
   }
 
+  /** Adds a {@link RequestMetrics} whose metrics will be recorded for request start and stop. */
+  public InstrumenterBuilder<REQUEST, RESPONSE> addRequestMetrics(RequestMetrics factory) {
```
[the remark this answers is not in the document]
anuraaga: Apparently I was happy enough having just written the docs on the annotation...

instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3Interceptors.java
```diff
@@ -14,4 +14,5 @@
 
   public static final Interceptor TRACING_INTERCEPTOR =
       OkHttpTracing.create(GlobalOpenTelemetry.get()).newInterceptor();
+  String a = Byte.toString((byte) 2);
```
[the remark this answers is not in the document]
anuraaga: 😱

mateuszrzeszutek (approved):

trask:
we have (some) metrics 🎉

PR #2880, opened 2021-04-28
Fix payara windows image

laurit:
Payara is started through a launcher process so adding agent to jvm arguments just adds it to the launcher not actual server.

### DIFF
--- a/smoke-tests/matrix/src/main/docker/payara/launch.bat
+++ b/smoke-tests/matrix/src/main/docker/payara/launch.bat
@@ -1,1 +1,8 @@
-java %JVM_ARGS% -jar glassfish/lib/client/appserver-cli.jar start-domain --verbose domain1
+:: server is run through a launcher process, to add jvm arguments we need to add them to configuration file
+:: firstly split JVM_ARGS environment variable by space character and put each argument into <jvm-options> tag
+:: after that place options into configuration xml after <jvm-options>-server</jvm-options>
+SET IN_CONF_FILE=/server/glassfish/domains/domain1/config/domain.xml
+SET OUT_CONF_FILE=/server/glassfish/domains/domain1/config/domain.xml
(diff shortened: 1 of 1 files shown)


smoke-tests/matrix/src/main/docker/payara/launch.bat
```diff
@@ -1 +1,8 @@
-java %JVM_ARGS% -jar glassfish/lib/client/appserver-cli.jar start-domain --verbose domain1
+:: server is run through a launcher process, to add jvm arguments we need to add them to configuration file
+:: firstly split JVM_ARGS environment variable by space character and put each argument into <jvm-options> tag
+:: after that place options into configuration xml after <jvm-options>-server</jvm-options>
+SET IN_CONF_FILE=/server/glassfish/domains/domain1/config/domain.xml
+SET OUT_CONF_FILE=/server/glassfish/domains/domain1/config/domain.xml
+powershell -command "$opts='<jvm-options>' + $env:JVM_ARGS + '</jvm-options>'; $opts=$opts -replace ' ', '</jvm-options><jvm-options>'; (gc $env:IN_CONF_FILE) -replace '<jvm-options>-server</jvm-options>', ('<jvm-options>-server</jvm-options>' + $opts) | sc $env:OUT_CONF_FILE"
```
iNikem, line 6: Are these `JVM_ARGS` dynamic? I mean can we already bake them into image itself?

iNikem (approved):

smoke-tests/matrix/src/main/docker/payara/launch.bat
laurit: `JVM_ARGS` currently contains just the path to our agent which really isn't dynamic, but as we can pass it as an environment variable for all other servers it doesn't make sense to bake this into image for only this server. Also sometimes it is useful to pass additional arguments like `-verbose:class` for debugging purposes.

PR #2881, opened 2021-04-28
First POC for external extension loading

iNikem:
Alternative approach to #2774


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,100 @@
...
+      delegateJarFile = new JarFile(new File(delegateJarFileLocation.toURI()), false);
+    } catch (URISyntaxException | IOException e) {
+      throw new IllegalStateException("Unable to read internal jar", e);
+    }
+
+    System.out.println("Delegate jar file is " + delegateJarFile.getName());
+  }
+
+  /** {@inheritDoc} */
+  protected URLConnection openConnection(URL url) {
```
laurit, line 55: are you sure that this is only called when class bytes are needed?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,54 @@
...
+// TODO find a way to initialize logging before using this class
+public class ExtensionClassLoader extends URLClassLoader {
+
+  public static ClassLoader getInstance(ClassLoader parent) {
+    // TODO add support for old deprecated properties, otel.exporter.jar and otel.initializer.jar
+    // TODO add support for system properties
+    URL extension = parseLocation(System.getenv("OTEL_JAVAAGENT_EXTENSIONS"));
+    if (extension != null) {
+      try {
+        URL wrappedUrl = new URL("otel", null, -1, "/", new RemappingUrlStreamHandler(extension));
```
laurit, line 28: instead of using a custom `URLStreamHandler` did you consider alternatives like doing the transform inside a `ClassFileTransformer` or `findClass`?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,100 @@
...
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    return new InternalJarUrlConnection(url, new ByteArrayInputStream(new byte[0]), 0);
+  }
+
+  private static byte[] remapClassBytes(InputStream in) throws IOException {
+    ClassWriter cw = new ClassWriter(0);
+    ClassReader cr = new ClassReader(in);
+    cr.accept(new ClassRemapper(cw, remapper), ClassReader.EXPAND_FRAMES);
```
laurit, line 97: out of curiosity is `EXPAND_FRAMES` really needed for remapping?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,100 @@
...
+        return new InternalJarUrlConnection(url, is, entry.getSize());
+      }
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    return new InternalJarUrlConnection(url, new ByteArrayInputStream(new byte[0]), 0);
+  }
+
+  private static byte[] remapClassBytes(InputStream in) throws IOException {
+    ClassWriter cw = new ClassWriter(0);
```
laurit, line 95: usually it is better to pass `ClassReader` to `ClassWriter` this way asm can copy methods that are not transformed directly from read to writer. Though I guess when remapping all methods get passed through a `MethodVisitor` and this optimization wouldn't work anyway.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
mateuszrzeszutek: The alternative approach https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2774 renames in the `findClass()` method

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
iNikem: Or resources. And we want to remap them as well, e.g. for Advices

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: We want to remap resources as well, so we cannot use class file transformers, I think...

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
iNikem: No idea, this was copy-paste from `ExporterClassLoader`

iNikem:
I think this PR is now ready for review and then merge. It is still an open question, should we go with URLHandler or move everything into classloader itself. But in anyway I think we should have one way to do that, both in this new classloader and in AgentClassloder. Thus discussion in #2912 is very relevant here as well (/cc @laurit )

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,77 @@
...
+  static {
+    ClassLoader.registerAsParallelCapable();
+  }
+
+  public static ClassLoader getInstance(ClassLoader parent) {
+    // TODO add support for old deprecated property otel.javaagent.experimental.exporter.jar
+    URL extension =
+        parseLocation(
+            System.getProperty(
+                "otel.javaagent.experimental.extensions",
```
laurit, line 37: System property is named extensions but seems like it only accepts 1 file. Should it be possible to only add 1 extension or multiple?
If multiple do they all live in the same class loader or would each get its own?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
laurit: perhaps it would make sense to push reading bytes and remapping into InternalJarUrlConnection and do it only when bytes are really needed

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: As class comment says, we want to support scanning the folder for several jars. In that case every jar should get a separate classloader, I think. This is a goal for a subsequent PR.

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -70,9 +77,18 @@ private static ClassLoader createAgentClassLoader(String innerJarFilename, URL b
...
-    Constructor constructor =
-        loaderClass.getDeclaredConstructor(URL.class, String.class, ClassLoader.class);
-    return (ClassLoader) constructor.newInstance(bootstrapUrl, innerJarFilename, agentParent);
+    Constructor<ClassLoader> constructor =
+        (Constructor<ClassLoader>)
+            loaderClass.getDeclaredConstructor(URL.class, String.class, ClassLoader.class);
+    ClassLoader agentClassLoader =
+        constructor.newInstance(bootstrapUrl, innerJarFilename, agentParent);
+
+    Class<?> extensionClassLoaderClass =
```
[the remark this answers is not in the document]
iNikem: I will touch this code again when doing support for multiple extension jars.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
```diff
@@ -0,0 +1,86 @@
...
+          rule(
+              "#io.opentelemetry.extension.aws",
+              "#io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.aws"),
+          rule("#java.util.logging.Logger", "#io.opentelemetry.javaagent.bootstrap.PatchLogger"),
+          rule("#org.slf4j", "#io.opentelemetry.javaagent.slf4j"));
+
+  private final JarFile delegateJarFile;
+  private final JarEntry entry;
+
+  private InputStream cachedResult;
```
anuraaga, line 41: Is only one `URLConnection` ever instantiated? Not sure when `openConnection` is called during classload

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,78 @@
...
+import java.net.URLStreamHandler;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+class RemappingUrlStreamHandler extends URLStreamHandler {
+
+  // We need to prefix the names to prevent the gradle shadowJar relocation rules from touching
+  // them. It's possible to do this by excluding this class from shading, but it may cause issue
+  // with transitive dependencies down the line.
+  private static final ShadingRemapper remapper =
```
anuraaga, line 26: Seems unused

examples/extension/smoke-tests/src/test/java/com/example/javaagent/smoketest/SmokeTest.java
```diff
@@ -0,0 +1,197 @@
...
+  protected Collection<ExportTraceServiceRequest> waitForTraces()
+      throws IOException, InterruptedException {
+    String content = waitForContent();
+
+    return StreamSupport.stream(OBJECT_MAPPER.readTree(content).spliterator(), false)
+        .map(
+            it -> {
+              ExportTraceServiceRequest.Builder builder = ExportTraceServiceRequest.newBuilder();
+              // TODO(anuraaga): Register parser into object mapper to avoid de -> re ->
+              // deserialize.
```
mateuszrzeszutek, line 161: Should we skip this TODO in an example?

examples/extension/custom/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+  id "java"
+}
+
+dependencies {
+  compileOnly("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
+  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${versions.opentelemetryAlpha}")
+
+  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-spi:${versions.opentelemetryJavaagentAlpha}")
+  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-api:${versions.opentelemetryJavaagentAlpha}")
+  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-tooling:${versions.opentelemetryJavaagentAlpha}")
```
mateuszrzeszutek, line 11: Shouldn't it be 
```suggestion
  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-api:${versions.opentelemetryJavaagentAlpha}")
  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api:${versions.opentelemetryJavaagentAlpha}")
```

?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,77 @@
...
+public class ExtensionClassLoader extends URLClassLoader {
+  // NOTE it's important not to use slf4j in this class, because this class is used before slf4j is
+  // configured, and so using slf4j here would initialize slf4j-simple before we have a chance to
+  // configure the logging levels
+
+  static {
+    ClassLoader.registerAsParallelCapable();
+  }
+
+  public static ClassLoader getInstance(ClassLoader parent) {
```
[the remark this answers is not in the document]
mateuszrzeszutek: It is used by reflection in `AgentInitializer` - `parent` is the `AgentClassLoader`

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,78 @@
...
+      if (file.startsWith("/")) {
+        file = file.substring(1);
+      }
+      JarEntry entry = delegateJarFile.getJarEntry(file);
+      if (entry == null) {
+        return null;
+      }
+
+      if (file.endsWith(".class")) {
+        return new RemappingUrlConnection(url, delegateJarFile, entry);
```
mateuszrzeszutek, line 68: I don't think that we have any SPIs that we shade, but maybe it's worth leaving a note here that says that `META-INF/services` won't be remapped?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
```diff
@@ -0,0 +1,86 @@
...
+  // with transitive dependencies down the line.
+  private static final ShadingRemapper remapper =
+      new ShadingRemapper(
+          rule("#io.opentelemetry.api", "#io.opentelemetry.javaagent.shaded.io.opentelemetry.api"),
+          rule(
+              "#io.opentelemetry.context",
+              "#io.opentelemetry.javaagent.shaded.io.opentelemetry.context"),
+          rule(
+              "#io.opentelemetry.extension.aws",
+              "#io.opentelemetry.javaagent.shaded.io.opentelemetry.extension.aws"),
```
mateuszrzeszutek, line 34: Should we add `io.opentelemetry.semconv` and `io.opentelemetry.instrumentation` too?

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,77 @@
...
+        parseLocation(
+            System.getProperty(
+                "otel.javaagent.experimental.extensions",
+                System.getenv("OTEL_JAVAAGENT_EXPERIMENTAL_EXTENSIONS")));
+
+    if (extension == null) {
+      extension =
+          parseLocation(
+              System.getProperty(
+                  "otel.javaagent.experimental.initializer.jar",
```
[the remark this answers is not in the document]
iNikem: When I add support for several jars, this will not be a big complication...

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
iNikem: Certainly not one, why do you ask?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
iNikem: Indeed, will do

examples/extension/custom/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+  id "java"
+}
+
+dependencies {
+  compileOnly("io.opentelemetry:opentelemetry-sdk:${versions.opentelemetry}")
+  compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${versions.opentelemetryAlpha}")
+
+  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api:${versions.opentelemetryJavaagentAlpha}")
+  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-api:${versions.opentelemetryJavaagentAlpha}")
+  compileOnly("io.opentelemetry.javaagent:opentelemetry-javaagent-tooling:${versions.opentelemetryJavaagentAlpha}")
```
mateuszrzeszutek, line 11: Is tooling still needed for the extension?

mateuszrzeszutek (approved):

examples/extension/custom/build.gradle
iNikem: I have tested at last, no need.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
anuraaga: If not once, I guess this cache is never used? Should it be a static cache?

anuraaga (approved):
A bit uncertain of the caching but basically looks good

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
iNikem: A, that's the catch :) it IS read/called at least 2 times by `URLClassoader`. Old `AgentClassLoader` also had an optimisation for this double reading.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,61 @@
...
+
+      if (file.startsWith("/")) {
+        file = file.substring(1);
+      }
+      JarEntry entry = delegateJarFile.getJarEntry(file);
+      if (entry == null) {
+        return null;
+      }
+
+      // That will NOT remap the content of files under META-INF/services
```
trask, line 49: is this a problem?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -0,0 +1,77 @@
...
+ * it compatible with shaded SDK used by the agent.
+ */
+// TODO find a way to initialize logging before using this class
+// TODO support scanning a folder for several extension jars and keep them isolated from each other
+// Used by AgentInitializer
+@SuppressWarnings("unused")
+public class ExtensionClassLoader extends URLClassLoader {
+  // NOTE it's important not to use slf4j in this class, because this class is used before slf4j is
+  // configured, and so using slf4j here would initialize slf4j-simple before we have a chance to
+  // configure the logging levels
```
trask, line 26: this restriction can be avoided by @laurit's comment https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2881/files#r628026661 (moving instantiation of ExtensionClassLoader inside of `AgentInstaller`)

trask (approved):
ExtensionClassLoader's `URLStreamHandler`/`URLConnection` approach has diverged from the new AgentClassLoader behavior implemented by @laurit in #2912, it's not clear if these differences are needed, or it's just drift since you started this work prior to his changes?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
iNikem: Not for now

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: I want to address all changes around this when I will add support for multiple jars. I already see that will change instantiation logic.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -241,7 +240,8 @@ private static IgnoreMatcherProvider loadIgnoreMatcherProvider() {
   }
 
   private static List<AgentExtension> loadAgentExtensions() {
-    return SafeServiceLoader.load(AgentExtension.class, AgentInstaller.class.getClassLoader())
+    return SafeServiceLoader.load(
+            AgentExtension.class, Thread.currentThread().getContextClassLoader())
```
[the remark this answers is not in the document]
iNikem: Done

examples/extension/settings.gradle
```diff
@@ -0,0 +1,5 @@
+rootProject.name = 'opentelemetry-java-instrumentation-extension-demo'
+
+include "custom"
```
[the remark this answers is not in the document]
iNikem: Looking at this now I think that we don't need multi-project at all. Can I do a follow up PR where I convert this to a usual gradle project? Together with removing mavenLocal :)

iNikem:
> ExtensionClassLoader's `URLStreamHandler`/`URLConnection` approach has diverged from the new AgentClassLoader behavior implemented by @laurit in #2912, it's not clear if these differences are needed, or it's just drift since you started this work prior to his changes?

We discussed this with Lauri and agreed that it is Ok for now to have different ways to do things. I will review it later.

examples/extension/settings.gradle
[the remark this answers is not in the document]
trask: 👍

trask:
> I will review it later.

👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private InputStream readAndRemap() {
+    try {
+      InputStream inputStream = delegateJarFile.getInputStream(entry);
+      byte[] remappedClass = remapClassBytes(inputStream);
+      return new ByteArrayInputStream(remappedClass);
+    } catch (IOException e) {
+      System.err.printf("Failed to remap bytes for %s: %s%n", url.toString(), e.getMessage());
+      return new ByteArrayInputStream(new byte[0]);
```
laurit, line 76: Perhaps it would make more sense to just throw an exception from here. If reading bytes fails when defining a class and we return an empty stream then won't the defining the class would still fail with classformaterror (i guess)?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
laurit: Caching `Inputstream` this way is weird. As streams can be read only once without using mark/reset then this caching would make sense only when someone opens the stream and doesn't use it at all and then reopens the stream and actually reads it. Perhaps this isn't really needed any more? If it is needed then caching byte array and returning a new `ByteArrayInputStream` from `getInputStream` would be nicer.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
```diff
@@ -0,0 +1,61 @@
...
+        // nullInputStream() is not available until Java 11
+        return new InputStreamUrlConnection(url, new ByteArrayInputStream(new byte[0]), 0);
+      }
+
+      if (file.startsWith("/")) {
+        file = file.substring(1);
+      }
+      JarEntry entry = delegateJarFile.getJarEntry(file);
+      if (entry == null) {
+        return null;
```
laurit, line 46: I suspect that this should throw an exception instead of returning null to be more in line with url stream handlers from jdk.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
iNikem: `handler.getInputStream()` is called twice:

1. In `jdk.internal.loader.URLClassPath.Loader#getResource(java.lang.String, boolean)`, where exactly "someone opens the stream and doesn't use it at all" happens
2. in `jdk.internal.loader.Resource#cachedInputStream`, when stream is actually used afterwards.

But I agree with you that caching byte arrays makes more sense, will do it.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlStreamHandler.java
iNikem: Throwing `IOException`

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
iNikem: Done.

examples/extension/smoke-tests/src/test/java/com/example/javaagent/smoketest/SmokeTest.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  private static GenericContainer backend;
+  private static GenericContainer collector;
+
+  @BeforeAll
+  static void setupSpec() {
+    backend =
+        new GenericContainer<>(
+            "open-telemetry-docker-dev.bintray.io/java/smoke-fake-backend:latest")
```
laurit, line 60: Smoketests use ghcr.io/open-telemetry/java-test-containers:smoke-fake-backend-20210324.684269693

laurit (approved):

trask:
🥳

PR #2900, opened 2021-05-03
Resteasy server span naming

laurit:
Currently jax-rs server span naming relies on instrumenting all jax-rs methods in `JaxRsAnnotationsInstrumentation` and when such a method is called `JaxRsAnnotationsTracer` tries to figure out a suitable server span name by finding `@Path` annotations on the called method and class. If `@Path` annotation is not found on called class it searches class hierarchy for `@Path` annotation and similarly if `@Path` is not found on called method it searches class hierarchy for a method that current method overrides.

Unfortunately current implementation does not detect all jax-rs methods. For example
```
@Path("test-resource-super")
class JaxRsSuperClassTestResource extends JaxRsSuperClassTestResourceSuper {
}

class JaxRsSuperClassTestResourceSuper {
  @GET
  Object call() {
    "ok"
  }
}
```
`JaxRsSuperClassTestResourceSuper.call` doesn't get instrumented and
```
class JaxRsInterfaceClassTestResource extends JaxRsInterfaceClassTestResourceSuper implements JaxRsInterface {
}

@Path("test-resource-interface")
interface JaxRsInterface {
  @Path("call")
  @GET
  Object call()
}

class JaxRsInterfaceClassTestResourceSuper {
  Object call() {
    "ok"
  }
}
```
`JaxRsInterfaceClassTestResourceSuper.call` doesn't get instrumented.

This pr implements resteasy specific integration for setting server span name. Resteasy invoker object is associated with the path for which it was created and when the invoker is used that path is set as server span name. This pr also adds support for subresource locators.


iNikem (approved):

mateuszrzeszutek (approved):

Sifat1998 (approved):

iNikem:
@laurit Please rebase

PR #2910, opened 2021-05-05
Fix Ratpack server context propagation and enable its concurrency test

### LINKED ISSUE #2648: Fix ratpack context propagation
Ratpack probably has the pattern of eagerly handling requests from the callback of a previous one, which causes context leak.

agoallikmaa:
Fixes #2648 

* Use context from Netty channel attributes as the parent context for `ratpack.handler` as executor instrumentation does not work correctly for it.
* Disable context propagation to Ratpack runnables and callables as the context they carry is not only unused, but also incorrect. As these are anonymous classes and created inside different classes, checking for exact classname does not work, so disabled it for the whole package.
* To do the above, made classname checking use the original task instead of wrapped task for anonymous classes.
* Enabled concurrency tests for Ratpack server.

### DIFF
--- a/instrumentation/executors/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaconcurrent/JavaExecutorInstrumentation.java
+++ b/instrumentation/executors/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaconcurrent/JavaExecutorInstrumentation.java
@@ -75,12 +75,11 @@ public class JavaExecutorInstrumentation extends AbstractExecutorInstrumentation
     public static State enterJobSubmit(
         @Advice.Argument(value = 0, readOnly = false) Runnable task) {
       if (ExecutorInstrumentationUtils.shouldAttachStateToTask(task)) {
-        Runnable newTask = RunnableWrapper.wrapIfNeeded(task);
-        task = newTask;
+        task = RunnableWrapper.wrapIfNeeded(task);
         ContextStore<Runnable, State> contextStore =
             InstrumentationContext.get(Runnable.class, State.class);
         return ExecutorInstrumentationUtils.setupState(
-            contextStore, newTask, Java8BytecodeBridge.currentContext());
+            contextStore, task, Java8BytecodeBridge.currentContext());
       }
       return null;
     }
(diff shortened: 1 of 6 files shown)


instrumentation/executors/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaconcurrent/JavaExecutorInstrumentation.java
```diff
@@ -74,8 +74,8 @@
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static State enterJobSubmit(
         @Advice.Argument(value = 0, readOnly = false) Runnable task) {
-      Runnable newTask = RunnableWrapper.wrapIfNeeded(task);
-      if (ExecutorInstrumentationUtils.shouldAttachStateToTask(newTask)) {
+      if (ExecutorInstrumentationUtils.shouldAttachStateToTask(task)) {
+        Runnable newTask = RunnableWrapper.wrapIfNeeded(task);
```
laurit, line 78: Could drop `newTask` and just assign to `task`.

laurit (approved):

instrumentation/executors/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaconcurrent/JavaExecutorInstrumentation.java
agoallikmaa: Done.

instrumentation/ratpack-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/ratpack/TracingHandler.java
```diff
@@ -23,9 +24,14 @@ public void handle(Context ctx) {
         ctx.getDirectChannelAccess().getChannel().attr(AttributeKeys.SERVER_SPAN);
     io.opentelemetry.context.Context serverSpanContext = spanAttribute.get();
 
-    // Relying on executor instrumentation to assume the netty span is in context as the parent.
+    // Must use context from channel, as executor instrumentation is not accurate - Ratpack
+    // internally queues events and then drains them in batches, causing executor instrumentation to
+    // attach the same context to a batch of events from different requests.
```
trask, line 29: 👍

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/ExecutorInstrumentationUtils.java
```diff
@@ -92,6 +92,15 @@ protected Boolean computeValue(Class<?> taskClass) {
             return false;
           }
 
+          if (taskClass.getName().startsWith("ratpack.exec.internal.")) {
+            // Context is passed through Netty channels in Ratpack as executor instrumentation is
+            // not suitable. As the context that would be propagated via executor would be
+            // incorrect, skip the propagation. Not checking for concrete class names as this covers
+            // anonymous classes from ratpack.exec.internal.DefaultExecution and
+            // ratpack.exec.internal.DefaultExecController.
```
trask, line 100: 👍

trask (approved):
nice

PR #2912, opened 2021-05-05
Agent class loader multi release jar

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2600
Treat `inst` directory in agent jar as a multi release jar.


iNikem:
@mateuszrzeszutek claimed that this is already solved: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2803#issuecomment-828516706

laurit:
As far as i understand @mateuszrzeszutek solved it for new instrumentation api, but later removed that change. This pr should allow us to use nanosecond precision clock from sdk with jdk9 (sdk is a multirelease jar).

iNikem:
I am very uncomfortable with these changes just for 1 class. And I have to think how this will play along with my work on extensions.

We in agent can detect which JVM we run on and install different Clocks anyway, right? We don't even have to use that `JavaVersionSpecific` from SDK...

anuraaga:
@iNikem We have an issue for adding another use of mrjar, which you suggested too :-)

https://github.com/open-telemetry/opentelemetry-java/issues/1564

That being said it's not that important for the agent to use the java 9+ version for that either. Using AgentClock with reflection seems OK too if the complication here is too much, it seems like there should be an easier way but maybe not :(

laurit:
Keep in mind that besides adding complexity to `AgentClassLoader` this change removes complexity by removing `InternalJarUrlHandler`. `InternalJarUrlHandler ` is a bit strange anyway. The caching in `openConnection` is weird and the fact that it opens stream to jar entry in `openConnection` is strange as jdk does this in `connect` of `URLConnection`. To me doing this in class loader seems more natural than doing it in url stream handler.
Currently we have 2 classes that reside under `META-INF/versions` I guess we could somehow replace the clock from sdk or do some unholy magic to make it work. To me it seems easier to just make our class loader handle mrjars so we wouldn't need to concern ourselves with stuff like that in the future.

iNikem:
But `InternalJarUrlHandler` does not need to handle e.g. Manifest, so I still argue that it was easier than doing everything in class loader. But I agree with you about caching...

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -92,6 +140,89 @@ public AgentClassLoader(
     return super.loadClass(name, resolve);
   }
 
+  @Override
+  protected Class<?> findClass(String name) throws ClassNotFoundException {
+    JarEntry jarEntry = findJarEntry(name.replace('.', '/') + ".class");
```
iNikem, line 145: Do you change the usual "parent-first" load order on purpose? Why?

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -92,6 +140,89 @@ public AgentClassLoader(
...
+    if (jarEntry != null) {
+      int size = (int) jarEntry.getSize();
```
iNikem, line 147: Please extract these low level details into a separate method

iNikem (commented):
How much of this custom code in AgentClassLoader is needed because of having `inst/` prefix inside jar file? I mean IF we could get read of it, how much simpler this code will be? Can we estimate?

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
laurit: Parent delegation is handled in `loadClass` and we use the default parent first strategy. Here I first search for the class in `inst/` and after that call `super.findClass` to find the class from agent initializer jar.
As a side note changing delegation to child first might be a good idea as currently when somebody adds something that we use in agent to bootclasspath we'll use it instead of what is in the agent. I guess this could be a feature or a bug depending on how you look it.

mateuszrzeszutek:
> How much of this custom code in AgentClassLoader is needed because of having `inst/` prefix inside jar file? I mean IF we could get read of it, how much simpler this code will be?

I think I remember @tylerbenson (or somebody else from DD) mentioning either in SIG or in some PR/issue that this prefix isn't really needed for otel javaagent and that we could remove that. If it makes our code simpler, why not?

richardstartin:
You don't need the prefix in this agent at all. It comes from Datadog where it serves the purpose of isolating various independent features which are bundled together, with sharing of classes required by more than one independent feature in a parent classloader.

laurit:
So what are our options here? We could drop the prefix and `AgentClassLoader` and shade all classes/resources under `io.opentelemetry.javaagent.shaded` to avoid our bundled libraries clashing with users libraries. Now this wouldn't actually solve the initial issue, as far as I know bootstrap class loader doesn't support multi release jars.
As there is only 1 class that needs to behave differently on jdk9+ we could just copy https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/CurrentJavaVersionSpecific.java from sdk modify it to return either JavaVersionSpecific or Java9VersionSpecific shade it together so that our copy overwrites the sdk one and hope that sdk authors never touch this file again or add any more mrjar usages that we would care about (this also applies to libraries that we use). Or we could nicely ask sdk authors to not use mrjar to make life easier for us. Or we could use some unorthodox reflection/unsafe trickery to replace the private static final value in https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/JavaVersionSpecific.java or we could use our awesome bytecode modification skils to rewrite CurrentJavaVersionSpecific to return jdk9 version of the JavaVersionSpecific class or replace the content of JavaVersionSpecific with Java9VersionSpecific when that class is loaded. So there are alternatives to making `AgentClassLoader` handle mrjars.
One reason we might choose to keep `AgentClassLoader` is that it isolates our classes from the rest of the world so we don't need to shade everything. Anybody who has had the pleasure of debugging shaded code knows that this can be a brain damaging process. A flag to not shade anything so that breakpoints would hit as expected could be handy. Though ease of debugging is a development feature even without using `AgentClassLoader` we could build agent without shading when needed. For second reason why we might wish to isolate our classes in `AgentClassLoader` consider serialization gadget vulnerabilities. See https://github.com/FasterXML/jackson-databind/blob/master/src/main/java/com/fasterxml/jackson/databind/jsontype/impl/SubTypeValidator.java apparently even bundling something as harmless looking as commons collections can be trouble. If you look closely you can see that the list of classes contains shaded classes from at least one agent, for them it was embedded-logback-core. A slightly higher chance of not appearing in lists like that should justify keeping `AgentClassLoader`.
Now we could keep `AgentClassLoader` and just drop the prefix. Bootstrap class loader can't load them because of the classdata suffix, but resources could still clash with user libraries. This could be solved by shading everything. When the coice is between shading everything or using a prefix then having a prefix doesn't sound that bad even when not having the prefix could let us leverage multi version handling from `JarFile` directly (because of prefix we need our own version of `JarFile.getVersionedEntry`). What I'm not sure about is when we use prefix should we also use a suffix for classes? Currently I fail to see how this helps. @richardstartin maybe you can shed some light on why this was added.
One more thing I'd like to point out is that the current version of this pr does the jar entry lookup the same way as `JarFile` which is by looping over current java version towards 8 searching for versioned entry. Testing whether entry is present is a hashtable lookup and should be fast, but as java version advances twice a year nowadays it could get excessive at some point.

mateuszrzeszutek:
> As there is only 1 class that needs to behave differently on jdk9+ we could just copy https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/internal/CurrentJavaVersionSpecific.java from sdk modify it to return either JavaVersionSpecific or Java9VersionSpecific shade it together so that our copy overwrites the sdk one and hope that sdk authors never touch this file again or add any more mrjar usages that we would care about (this also applies to libraries that we use). 

There's already an issue in otel-java that'll introduce mrjar usage in `ProcessResource`.

There's one more reason to keep the prefix: we have several instrumentations that injects `META-INF/services` files into the application classloader. If we were to put them into the bootstrap classloader, turning those instrumentation off would not be possible, since the SPI files would always be accessible in the instrumented application.

Also, I took a look what's inside the `inst` (and `inst/META-INF`) directory and we even have some *.so and *.dll files there - not exposing them to the bootstrap CL would be a safer choice.

richardstartin:
That's a wall of text with my handle in the middle of it 🤣 . I didn't read it all, but it's not the prefix that prevents the classes from being loaded by some other classloader but the `data` suffix. I don't think you need the prefix, we just use it for the parent/children classloaders I mentioned, I would get rid of it if I had the chance.

trask:
hey @laurit, I tested this on our agent distro (which is signed, unfortunately a requirement for us), and it's about 2 seconds slower to start up spring petclinic on Java 8 (unsigned agents don't appear to be affected).

[as a side note: signed agents are a disaster on Java 8 because the JIT compiler isn't initialized until after premain (this was fixed in Java 9, I lobbied to get it backported to Java 8, but it was not accepted), which makes jar file signature verification horribly slow]

I believe the reason for the additional slowness in this PR is because getResource()/findResource() now returns a normal jar-file URL, which doesn't bypass signature verification, where previously `InternalJarUrlHandler` bypassed signature verification.

trask:
> this approach will not work if we add several URLs into one AgentClassloader/ExtensionsClassloader

I didn't follow this, does it mean we won't be able to have multiple extension jar files?

laurit:
> hey @laurit, I tested this on our agent distro (which is signed, unfortunately a requirement for us), and it's about 2 seconds slower to start up spring petclinic on Java 8 (unsigned agents don't appear to be affected).
> 
> [as a side note: signed agents are a disaster on Java 8 because the JIT compiler isn't initialized until after premain (this was fixed in Java 9, I lobbied to get it backported to Java 8, but it was not accepted), which makes jar file signature verification horribly slow]
> 
> I believe the reason for the additional slowness in this PR is because getResource()/findResource() now returns a normal jar-file URL, which doesn't bypass signature verification, where previously `InternalJarUrlHandler` bypassed signature verification.

@trask could you test it again and see if it behaves better now. As far as I understand the expensive part is opening the jar with verification enabled. Tomcat disables caching for jar urls so a new jar gets opened every time someone tries to read bytes from the url. I replaced jar urls with a custom `URLStreamHandler` that avoids opening jar file multiple times. Urls returned from `BootstrapClassLoaderProxy` were also changed for the same reason.
I was aware that jit is enabled only after premain has completed on hotspot 8 but I didn't know that it was fixed in later versions :) On a previous agent I worked on we experimented with hooking into `LauncherHelper.checkAndLoadMain` and doing our startup logic from there to make it start a bit faster on jdk8.

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -79,14 +83,19 @@ public AgentClassLoader(
...
     jarEntryPrefix =
         internalJarFileName
             + (internalJarFileName.isEmpty() || internalJarFileName.endsWith("/") ? "" : "/");
     try {
-      jarBase = new URL("jar:" + bootstrapJarLocation + "!/");
+      // open jar with verification disabled
       jarFile = new JarFile(new File(bootstrapJarLocation.toURI()), false);
+      // base url for constructing jar entry urls
+      // we use a custom protocol instead of typical jar:file: because we don't want to be affected
+      // by user code disabling URLConnection caching for jar protocol e.g. tomcat does this
```
trask, line 96: 👍

trask (approved):
> @trask could you test it again and see if it behaves better now

yes, all good now, thx!!

> I was aware that jit is enabled only after premain has completed on hotspot 8 but I didn't know that it was fixed in later versions

I just learned all of this recently when trying to get the the bottom of our signed jar slowness issues on 8 😭

Issue #2651, opened 2021-03-29
Gradle configuration time too slow for development


anuraaga:
I've noticed lately that during development when I have to retry tests, it's slowed down a lot since configuration time in our project seems unusually slow. The output in IDE will include something like this

```
> Task :buildSrc:check UP-TO-DATE
> Task :buildSrc:build UP-TO-DATE
> Configure project :benchmark
Inferred project: opentelemetry-java-instrumentation, version: 1.1.0-SNAPSHOT
> Task :testing-common:processResources NO-SOURCE
> Task :instrumentation-api:sqlSanitizerJflex UP-TO-DATE
```

And it takes about 7s on my Macbook Pro 16 to move from 
```
> Configure project :benchmark
Inferred project: opentelemetry-java-instrumentation, version: 1.1.0-SNAPSHOT
```

to
```
> Task :testing-common:processResources NO-SOURCE
```

I don't think it's supposed to take this long, and dominates the entire time for small test runs (or even many agent tests).

ryandens:
I'd like to attempt some improvements here! I have two ideas

- Breaking the `conventions` included build into a few separate modules to reduce the impact of changes to plugins on every module. For example, the `otel.jmh-conventions` is only used in a few subprojects, but right now if any changes is made to `jmh-conventions` then every project needs to be reconfigured/rebuilt. This will have a relatively minor impact on day to day for most developers, but should have a significant impact when working on build changes
- Working towards enabling the Gradle Configuration Cache. This will take significant effort, but I'm excited by the potential savings here. There are a couple of issues I'm working on fixing, but mainly the problems are in upstream plugins. 
  - I've already made changes to the [jflex plugin](https://github.com/jprante/gradle-plugin-jflex/pull/21) to make it compatible with the configuration cache 
  - I have a [working prototype](https://github.com/unbroken-dome/gradle-xjc-plugin/compare/master...ryandens:gradle-xjc-plugin:ryandens/configuration-cache) of changes to the Gradle XJC plugin used by this project. I'm less certain the XJC plugin will get merged though, as the project hasn't had maintenance in some time.
  - There's a small Gradle bug with the Scala plugin with old versions of scala with the configuration cache enabled. I have a reproducer project [here](https://github.com/ryandens/scala-configuration-cache-reproducer) and am determining what if any workarounds are available. It's possible the build tool will not change to support this older version of Scala. 
  
[...]

PR #2916, opened 2021-05-06
Fix enabling gradle configure on demand

laurit:
Relates to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2651
See https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand
As this is an incubating feature strange things may happen, but it definitely does make running tests a bit less painful.

### DIFF
--- a/buildSrc/src/main/groovy/MuzzlePlugin.groovy
+++ b/buildSrc/src/main/groovy/MuzzlePlugin.groovy
@@ -31,6 +31,7 @@ import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
 import org.gradle.api.model.ObjectFactory
+
 /**
  * muzzle task plugin which runs muzzle validation against a range of dependencies.
  */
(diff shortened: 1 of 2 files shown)


anuraaga:
I've seen this flag be discussed on many Gradle issues as basically deprecated - not officially but it's not recommended since there are also unworkaroundable bugs that can cause builds to crash and the bugs won't be fixed. So I'd avoid this flag.

Our tests are painful probably due to a problem in our build scripts, they must be doing something heavy during configuration that shouldn't be happening since it really should never be that slow - back from vacation and I can spend some time now to find the problem if someone else doesn't first. But I'd prefer to do that over adding this potentially problematic flag.

iNikem:
@anuraaga can you point me to "this flag be discussed on many Gradle issues as basically deprecated"? I am curious...

laurit:
@iNikem For example https://github.com/gradle/gradle/issues/2868 they wanted to deprecate it but gave up on it as it was still useful for some projects.
Instead of enabling it by default I'll change this pr to only include a fix to muzzle plugin that unbreaks configure on demand so those who wish can add `org.gradle.configureondemand=true` to gradle.properties or use `--configure-on-demand` until a better solution emerges.

iNikem (approved):

mateuszrzeszutek (approved):

anuraaga (approved):
Thanks!

PR #2920, opened 2021-05-06
Fix context leak on jetty

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2915


trask (approved):
❤️ one more context leak down!

iNikem (approved):
@laurit you are a magician :D

PR #2925, opened 2021-05-07
InstrumentationModule cleanup

mateuszrzeszutek:
Removed a few deprecated methods (and an interface) - we already have lots of breaking changes, it's probably better to group as many of them as possible in a single release.

In the next PR I'll implement the `InstrumentationLoader` idea.


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/spi/ByteBuddyAgentCustomizer.java
```diff
@@ -1,34 +0,0 @@
...
- * {@link AgentBuilder} for vendor specific needs. For example install custom listeners or exclude
- * classes. Use this SPI carefully because it can change {@link net.bytebuddy.ByteBuddy} behaviour.
- *
- * <p>This is a service provider interface that requires implementations to be registered in {@code
- * META-INF/services} folder.
- *
- * @deprecated Use {@link AgentExtension} with correct {@link AgentExtension#order()} instead.
- */
-@Deprecated
-public interface ByteBuddyAgentCustomizer {
```
trask, line 24: cc @pavolloffay: this use case is now covered by `AgentExtension`

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModule.java
```diff
@@ -150,7 +176,7 @@ public boolean isHelperClass(String className) {
    * <p><b>This method is generated automatically</b>: if you override it, the muzzle compile plugin
    * will not generate a new implementation, it will leave the existing one.
    */
-  protected String[] getMuzzleHelperClassNames() {
+  public String[] getMuzzleHelperClassNames() {
```
trask, line 179: should we return `List<String>`, since it looks like we always wrap the return value in `asList` anyways? probably same too for `getMuzzleReferences` above (not suggesting either for this PR)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/ActualInstrumentationExtensionImplementation.java
```diff
@@ -68,7 +68,7 @@ AgentBuilder extend(
 
     ElementMatcher.Junction<ClassLoader> moduleClassLoaderMatcher =
         instrumentationModule.classLoaderMatcher();
-    MuzzleMatcher muzzleMatcher = new MuzzleMatcher(instrumentationModule);
+    MuzzleMatcher muzzleMatcher = new MuzzleMatcher(instrumentationModule, helperClassNames);
```
trask, line 71: is this purpose for passing helperClassNames here just an optimization to avoid calling `getMuzzleHelperClassNames()` multiple times? maybe ideal for `getMuzzleHelperClassNames()` (and other generated methods?) to return constant to avoid that concern (not a suggestion for this PR)

trask (approved):

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModule.java
mateuszrzeszutek: Yes, that can be easily done for `getMuzzleHelperClassNames`; it'll be a bit more complicated (but certainly doable) for `getMuzzleReferences` too. That could be done together with https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2925#discussion_r628691545 in one of the next PRs

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/ActualInstrumentationExtensionImplementation.java
mateuszrzeszutek: Yeah, it's just an optimization. I guess that helper class names could use the same pattern as references (lazy-init field).

PR #2928, opened 2021-05-07
Assert spans in grpc client error test

laurit:
This change makes grpc client error thrown test pass with strict context check enabled. The issue here is that the thread which calls `onError` is still running when the test exits and we detect a context leak. Asserting traces seems to consume enough time for the problematic thread to exit context.


trask (approved):
> This change makes grpc client error thrown test pass with strict context check enabled

🎉

mateuszrzeszutek (approved):

PR #2930, opened 2021-05-08
Uncomment and fix JAX-RS default method tests

trask:


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/AgentElementMatchers.java
```diff
@@ -36,8 +35,7 @@
 
   public static ElementMatcher.Junction<TypeDescription> safeHasSuperType(
       ElementMatcher<TypeDescription> matcher) {
-    return not(isInterface())
-        .and(new SafeHasSuperTypeMatcher(new SafeErasureMatcher<>(matcher), false));
+    return new SafeHasSuperTypeMatcher(new SafeErasureMatcher<>(matcher), false);
```
trask, line 38: Without this change, interface default methods are not instrumented. I looked through the callers and changed some of them to `extendsClass()`, and the remaining seem more correct taking interface default methods in to account.

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/AgentElementMatchers.java
```diff
@@ -25,8 +25,7 @@
 
   public static ElementMatcher.Junction<TypeDescription> implementsInterface(
       ElementMatcher<TypeDescription> matcher) {
-    return not(isInterface())
-        .and(new SafeHasSuperTypeMatcher(new SafeErasureMatcher<>(matcher), true));
+    return new SafeHasSuperTypeMatcher(new SafeErasureMatcher<>(matcher), true);
```
trask, line 28: this change was not needed for JAX-RS default methods, but I looked through callers and not sure why we wouldn't also want to instrument interface default methods, so removed `not(isInterface())` here too

iNikem (approved):

PR #2931, opened 2021-05-08
Clean up Java 7 references

trask:


instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/TracingSession.java
```diff
@@ -53,7 +52,7 @@ public Session apply(Session session) {
             return new TracingSession(session);
           }
         },
-        directExecutor());
```
trask, line 56: `directExecutor` was only introduced in guava 18.0, and cassandra 3.0.0 depends on guava 16.0, so muzzle (rightly) not happy with usage of it

instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/DirectExecutor.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.cassandra.v3_0;
+
+import java.util.concurrent.Executor;
+
+class DirectExecutor implements Executor {
```
anuraaga, line 10: Is there an advantage of class vs lambda?

anuraaga (approved):

iNikem (approved):

PR #2946, opened 2021-05-10
Instrument okhttp builder constructor instead of client

anuraaga:
@agoallikmaa Since we need to address #2886 for the release, I went ahead and sent this smaller PR to address https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2894#issuecomment-834080132

This does not change the behavior that the interceptor is somewhat exposed to users. I think the idea of making sure not to do so is interesting, but it does complicate the instrumentation quite a bit. Currently we don't do this in any other instrumentation, but it might only matter for okhttp, which actually has a getter for `interceptors` whereas most libraries I've seen have builders without such a getter. I think it makes sense to address this if we want separately from fixing the stack overflow bug we got.

### DIFF
--- a/instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3InstrumentationModule.java
+++ b/instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3InstrumentationModule.java
@@ -9,11 +9,11 @@ import static java.util.Collections.singletonList;
 import static java.util.Collections.singletonMap;
 import static net.bytebuddy.matcher.ElementMatchers.isConstructor;
 import static net.bytebuddy.matcher.ElementMatchers.named;
-import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
 import com.google.auto.service.AutoService;
 import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
 import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import io.opentelemetry.javaagent.instrumentation.api.CallDepthThreadLocalMap;
 import java.util.List;
 import java.util.Map;
 import net.bytebuddy.asm.Advice;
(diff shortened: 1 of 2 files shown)


instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3InstrumentationModule.java
```diff
@@ -37,20 +37,32 @@ public OkHttp3InstrumentationModule() {
...
+      callDepth = CallDepthThreadLocalMap.incrementCallDepth(OkHttpClient.Builder.class);
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void addTracingInterceptor(
+        @Advice.This OkHttpClient.Builder builder, @Advice.Local("callDepth") int callDepth) {
+      if (callDepth > 0) {
+        return;
+      }
+      CallDepthThreadLocalMap.reset(OkHttpClient.Builder.class);
```
iNikem, line 62: Why this `reset` is needed? `callDepth` is already 0 here, no?

iNikem (approved):

instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3InstrumentationModule.java
anuraaga: This is our standard pattern - `callDepth` is set to `0` on the first call in a tree, but is incremented in other ones It's global state so even though the local variable is 0, the global variable is greater than 0 and we reset.

instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3InstrumentationModule.java
```diff
@@ -37,20 +37,32 @@ public OkHttp3InstrumentationModule() {
...
+    public static void trackCallDepth(@Advice.Local("callDepth") int callDepth) {
+      callDepth = CallDepthThreadLocalMap.incrementCallDepth(OkHttpClient.Builder.class);
+    }
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void addTracingInterceptor(
+        @Advice.This OkHttpClient.Builder builder, @Advice.Local("callDepth") int callDepth) {
+      if (callDepth > 0) {
+        return;
+      }
```
trask, line 61: it's not clear why we need the call depth check (since already checking below if interceptor is present before adding), can you add comment explaining?

trask (approved):

instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3InstrumentationModule.java
anuraaga: Yeah added

Issue #2564, opened 2021-03-11
gRPC trace missing


asarkar:
Spin off from https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1550. gRPC client interceptor missing trace. Quoting @anuraaga's [comment](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1550#issuecomment-796453990):
>  have a feeling this is because of the switch off of using grpc context directly for OTel. So it's time to implement a proper context bridge it'll take some time.

Attached app can be used to reproduce the issue. Extract and run `./gradlew test` from the root directory.

[otel-grpc-tracing.zip](https://github.com/open-telemetry/opentelemetry-java-instrumentation/files/6126414/otel-grpc-tracing.zip)

asarkar:
@anuraaga Any idea which release this might go into?

anuraaga:
Hoping it gets in soon - I started with a prerequisite for it but have some tricky issues

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2601

ankitgaur-cer:
@anuraaga please provide us an update on this, it is blocking us from moving forward.

anuraaga:
Hi @ankitgaur-cer @asarkar - sorry for not getting to this until now. Actually 1.1.0 includes the context bridge, and I can run the repro and the test passes! @asarkar please confirm that the issue has been solved.

asarkar:
@anuraaga I see the trace, but also a lot of the following error:
```
Context was not attached when detaching java.lang.Throwable: null
	at i.o.i.g.v.i.ContextStorageBridge.detach(ContextStorageBridge.java:86)
	at i.g.o.ContextStorageOverride.detach(ContextStorageOverride.java:26)
	at io.grpc.Context.detach(Context.java:463)
	at i.g.k.GrpcContextElement.restoreThreadContext(GrpcContextElement.kt:36)
	at i.g.k.GrpcContextElement.restoreThreadContext(GrpcContextElement.kt:27)
	... 15 frames truncated

java.lang.Throwable: null
	at io.opentelemetry.instrumentation.grpc.v1_5.internal.ContextStorageBridge.detach(ContextStorageBridge.java:86)
	at io.grpc.override.ContextStorageOverride.detach(ContextStorageOverride.java:26)
	at io.grpc.Context.detach(Context.java:463)
[...]

PR #2948, opened 2021-05-10
Raise gRPC version floor to 1.6 and use new context bridge API

anuraaga:
This raises the gRPC minimum version just a little bit (still 4+ years old) and uses a newly introduced context storage API that can return a value, removing the need for the side queues. Should improve performance nicely, and unfortunately I haven't been able to reproduce the issues in #2564 but by controlling the return value which gets passed back to us more directly, I think it does fix a corner case where the reference gets changed (I can't think of a reason the app wouldn't start because of the log message still though). The [decoupling](https://github.com/grpc/grpc-java/blob/v1.5.0/context/src/main/java/io/grpc/Context.java#L384) of `current()` and `Storage.attach()` in <=1.5 would allow them to be out of sync


iNikem:
@anuraaga please update the list of supported libraries as well

PR #2949, opened 2021-05-10
Some Gradle optimizations

anuraaga:
- Use lazy configuration as much as possible (at least in the shared scripts didn't do full verification across projects)
- Don't iterate over classpath every time configuring muzzle task
- Move plugin version specification to `pluginManagement`

After (remaining created classes are muzzle which will address in followup)
image

Before
image


buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/bytebuddy/ByteBuddyPluginConfigurator.java
```diff
@@ -40,64 +40,73 @@
...
     // add build resources dir to classpath if it's present
     File resourcesDir = sourceSet.getOutput().getResourcesDir();
     this.inputClasspath =
-        resourcesDir == null
-            ? inputClasspath
-            : ImmutableList.<File>builder()
-                .addAll(inputClasspath)
-                .add(sourceSet.getOutput().getResourcesDir())
-                .build();
+        resourcesDir == null ? inputClasspath : inputClasspath.plus(project.files(resourcesDir));
```
anuraaga, line 54: Significant improvement since classpaths are full of files

settings.gradle
```diff
@@ -1,3 +1,19 @@
+pluginManagement {
```
anuraaga, line 1: I have a hypothesis that use of the `plugins` block allows caching of plugin dependency resolution that isn't happening right now - `Resolving :classpath` on the first project of this build always takes a second, while on `-java` project it takes no time after the first run. Switching to `plugins` block will require inverting build logic, instead of `apply from: java.gradle -> apply plugin: 'java-library'`, it will be `plugins { java-library } -> plugins.withId('java-library) { apply from: 'java.gradle'`. I've generally found this pattern to have less evaluation order problems and prefer it anyways but it will be a big change so not trying yet.

instrumentation/rmi/javaagent/rmi-javaagent.gradle
```diff
@@ -6,7 +6,9 @@ muzzle {
   }
 }
 
-task "rmic", dependsOn: testClasses {
+def rmic = tasks.register('rmic') {
```
anuraaga, line 9: Significant since the logic is I/O heavy

gradle/instrumentation-common.gradle
```diff
@@ -68,9 +68,13 @@ configurations {
...
   afterEvaluate {
-    def latestDepTest = tasks.findByName('latestDepTest')
-    if (latestDepTest) {
-      tasks.test.dependsOn latestDepTest
+    try {
+      def latestDepTest = tasks.named('latestDepTest')
+      tasks.named('test').configure {
+        dependsOn(latestDepTest)
+      }
+    } catch (UnknownDomainObjectException e) {
```
anuraaga, line 76: This seems to be the only way to configure a task lazily if it exists or skip otherwise

gradle/spotbugs.gradle
```diff
@@ -10,23 +10,4 @@ allprojects {
     effort = "max"
     excludeFilter = file("$rootDir/gradle/spotbugs-exclude.xml")
   }
-
-  spotbugsTest {
```
anuraaga, line 14: Oops - I still need to debug some issues I was having making this lazy configuration

PR #2950, opened 2021-05-10
Add separate nightly run which does not use any gradle caches

iNikem:


.github/workflows/nightly.yaml
```diff
@@ -31,28 +31,6 @@ jobs:
           S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.S3_BUILD_CACHE_SECRET_KEY }}
         run: ./gradlew build --stacktrace -x :smoke-tests:test
 
-  example-distro:
```
iNikem, line 34: This job does not need to populate remote cache. So it can run only on no-cache workflow.

.github/workflows/nightly-no-cache.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Set up JDK 11 for running checks
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: 11
+
+
+      - name: Build
+        run: ./gradlew build --stacktrace -x :smoke-tests:test --no-build-cache
```
anuraaga, line 25: Just curious is it possible to use top level `env` and `GRADLE_OPTS` for some DRY?

anuraaga (approved):

.github/workflows/nightly-no-cache.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Set up JDK 11 for running checks
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: 11
+
+      - name: Build
+        run: ./gradlew build --stacktrace
+        working-directory: examples/distro --no-build-cache
```
mateuszrzeszutek, line 42: Typo:

```suggestion
        run: ./gradlew build --stacktrace --no-build-cache
        working-directory: examples/distro
```

mateuszrzeszutek (approved):

iNikem:
> Why can't we add `--no-build-cache` to existing nightly?

Because we do want to populate remote cache so that PR checks were faster.

Issue #2929, opened 2021-05-07
Servlet instrumentation overwrites setStatus that was set manually earlier


ryanrupp:
**Describe the bug**
See Slack thread [here](https://cloud-native.slack.com/archives/C0150QF88FL/p1620416060116800). I have some custom instrumentation that is trying to call `span.setStatus(error)` on a span generated by servlet instrumentation - this works correctly. However, the servlet instrumentation ends up overwriting this status with `UNSET` [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java#L257) due to `HttpStatusConverter` returning unset for >= 100 && < 400 HTTP status codes. In the Slack thread it's mentioned the instrumentation should only ever set ERROR. Note my HTTP request actually does return a 200 (non-error) but this is the nature of the RPC protocol I'm instrumenting that it returns 200 with an error status in the response body instead of HTTP status code.

**Steps to reproduce**
Add an instrumentation that modifies the span that the servlet is going to operate on, calling `setStatus(error)`.

**What did you expect to see?**
The span should be reported with status=error

**What did you see instead?**
The span was reported with status=unset

**What version are you using?**
1.1.0

**Environment**
OpenJDK 1.8

ryanrupp:
There's probably some other areas that are using this similar pattern of setting to UNSET explicitly instead of skipping the method `span.setStatus` method call, general pattern:

```
span.setStatus(mapFromException(something))

mapFromException(something) {
     return isError(something) ? ERROR : UNSET;
}
```

Other examples if you look for `SpanStatus.UNSET` usage e.g.:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/apache-dubbo/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboHelper.java#L27
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/0b0516bd1a0599c8f536cc6b2151782000681f39/instrumentation/grpc-1.5/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_5/GrpcHelper.java#L36

iNikem:
[...]

PR #2956, opened 2021-05-11
Don't set span status to unset

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2929
An alternative would be to ignore `UNSET` status in sdk.


mateuszrzeszutek (approved):

instrumentation/apache-dubbo/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboTracer.java
```diff
@@ -46,7 +47,10 @@ public Context startClientSpan(String interfaceName, String methodName) {
 
   public void end(Context context, Result result) {
     Span span = Span.fromContext(context);
-    span.setStatus(DubboHelper.statusFromResult(result));
+    StatusCode statusCode = DubboHelper.statusFromResult(result);
+    if (statusCode != StatusCode.UNSET) {
+      span.setStatus(statusCode);
```
trask, line 52: can pull `Span span = Span.fromContext(context);` line down from above
```suggestion
      Span.fromContext(context).setStatus(statusCode);
```

trask (approved):

PR #2972, opened 2021-05-12
Simplify servlet advice

laurit:


instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
```diff
@@ -53,33 +53,22 @@ public static void onEnter(
...
-        scope = updatedContext.makeCurrent();
-      }
+      attachedContext =
+          tracer().updateContext(attachedContext, httpServletRequest, mappingResolver, servlet);
+      scope = attachedContext.makeCurrent();
       return;
     }
 
     Context currentContext = Java8BytecodeBridge.currentContext();
     if (currentContext != null
```
mateuszrzeszutek, line 65: By the way, `currentContext()` can never return null - you can simplify that bit too 😄

mateuszrzeszutek (approved):

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
laurit: Thanks, removed it.

mateuszrzeszutek (approved):

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
```diff
@@ -53,39 +53,29 @@ public static void onEnter(
...
     }
 
     Context currentContext = Java8BytecodeBridge.currentContext();
-    if (currentContext != null
-        && Java8BytecodeBridge.spanFromContext(currentContext).isRecording()) {
-      // We already have a span but it was not created by servlet instrumentation.
-      // In case it was created by app server integration we need to update it with info from
-      // current request.
+    if (attachedContext != null
+        || Java8BytecodeBridge.spanFromContext(currentContext).isRecording()) {
```
laurit, line 66: Won't using `isRecording` cause issues when running with sampling?

trask: ya, I think `isRecording` should only be used to short-circuit updating a given span's attributes (or name) and probably nothing else, so would support removing this check

trask (approved):
less code ✅ 
more comments ✅ 
tests pass ✅ 
🚀

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
laurit: I'll try to get rid of `isRecording` next week (my initial attempt failed on glassfish and I wasn't able to repro it locally so it requires a bit of debugging). As `isRecording` was already used before this pr I believe we can merge this if there aren't any objections.

PR #2990, opened 2021-05-14
Disable smoke tests in release build since we can't run them inline w…

anuraaga:
…ith other builds effectively.

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -69,7 +69,7 @@ jobs:
         with:
           job-id: jdk11
           remote-build-cache-proxy-enabled: false
-          arguments: build final closeAndReleaseRepository --stacktrace -Prelease.version=${{ github.event.inputs.version }}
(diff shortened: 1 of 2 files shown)


anuraaga:
This is good motivation to revisit using a github action to query the status of the `main-build` and just not run any tests in `release-build`. I had issues with a simple approach when I first tried it several months ago but I'm sure it must be possible

iNikem:
If we split release job into several steps, we then can run all the tests. But this requires making sure that all steps actually use the same commit.

I am still uneasy about building and publishing artifacts without testing them. Yes, even if commit is the same, build env may unexpectedly change from the time of `main` build.

anuraaga:
> If we split release job into several steps, we then can run all the tests. But this requires making sure that all steps actually use the same commit.

@iNikem Do you think we should block the next release on this or can temporarily disable?

iNikem:
We can disable for now

anuraaga:
@iNikem Cool - I filed #2993 so I guess we can merge this one?

iNikem (approved):

PR #2992, opened 2021-05-14
Change Async Servlet span end logic to fix race condition on Undertow

agoallikmaa:
After enabling concurrency test to RestEasy HTTP server tests, it turned out that test intermittently fails when it uses Undertow as the servlet engine. This was because of a race condition in Async Servlet instrumentations - the async listener events were fired before the method exit instrumentation added the listener, yet the `isAsyncStarted` method did not yet revert to returning false.

There does not seem to be a straightforward way to avoid races like this properly in the servlet service method exit instrumentation, therefore I moved adding the async listener to the instrumentation of the `startAsync` method itself, which should guarantee that that the listener is added to the `AsyncContext` before it is possible for them to be invoked.

The logic in service method exit itself was reduced to simply checking if an async listener has already been attached to this request - this information is saved in the request attributes at the time when the listener is added. In case the listener is not present, assume it is a synchronous request and end it immediately.


mateuszrzeszutek (approved):

agoallikmaa:
Added fix for Liberty - response object attached as an attribute to request because on Liberty, `AsyncContext#addListener(AsyncListener, ServletRequest, ServletResponse)` must be called instead of `AsyncContext#addListener(AsyncListener)`

mateuszrzeszutek (approved):

instrumentation/servlet/servlet-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v5_0/async/AsyncStartAdvice.java
```diff
@@ -0,0 +1,40 @@
...
+import io.opentelemetry.javaagent.instrumentation.api.CallDepthThreadLocalMap;
+import jakarta.servlet.AsyncContext;
+import jakarta.servlet.http.HttpServletRequest;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.implementation.bytecode.assign.Assigner;
+
+public class AsyncStartAdvice {
+  @Advice.OnMethodEnter(suppress = Throwable.class)
+  public static void startAsyncEnter() {
+    // This allows to detect the outermost invocation of startAsync in method exit
```
trask, line 19: why is this important? (since checking `isAsyncListenerAttached` before calling `attachAsyncListener` below already)

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3AsyncStartAdvice.java
```diff
@@ -0,0 +1,40 @@
...
+import io.opentelemetry.javaagent.instrumentation.api.CallDepthThreadLocalMap;
+import javax.servlet.AsyncContext;
+import javax.servlet.http.HttpServletRequest;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.implementation.bytecode.assign.Assigner;
+
+public class Servlet3AsyncStartAdvice {
+  @Advice.OnMethodEnter(suppress = Throwable.class)
+  public static void startAsyncEnter() {
+    // This allows to detect the outermost invocation of startAsync in method exit
```
trask, line 19: same question here about why this is important

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyStartSpanAdvice.java
```diff
@@ -26,5 +26,9 @@ public static void onEnter() {
 
     ctx.setContext(context);
     ctx.setScope(scope);
+
+    // For startAsync instrumentation to work on Liberty, the listener needs to be added with a
+    // response object included, which is not available in that instrumentation.
```
trask, line 31: it seems like we should always be passing request/response to AsyncContext.addListener(), since our listener calls [AsyncEvent.getSuppliedRequest()](https://docs.oracle.com/javaee/7/api/javax/servlet/ServletRequest.html#startAsync--) and [AsyncEvent.getSuppliedResponse()](https://docs.oracle.com/javaee/7/api/javax/servlet/AsyncEvent.html#getSuppliedResponse--).

-- or --

maybe our listener should call `event.getAsyncContext().getRequest()` and  `event.getAsyncContext().getResponse()` instead, and then maybe we don't ever need to pass request/response?

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/common/service/ServletAndFilterAdviceHelper.java
```diff
@@ -49,23 +46,33 @@
...
     }
+
+    // Possible scenarios:
+    // 1) synchronously handled request
+    // 2) asynchronously handled request, but startAsync instrumentation is broken on this server
+    //    (should be caught by tests for the specific server).
+    // In case of broken startAsync instrumentation, no fallback handling for asynchronous requests
+    // should be provided as handling it in the handler/service method is  prone to race conditions
+    // (seen happening on Undertow) that may make some tests pass which should fail due to the
+    // possible race.
```
trask, line 75: i'm hoping we can remove this part about broken startAsync instrumentation if the above comment turns out to be fruitful

instrumentation/servlet/servlet-common/library/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -142,6 +148,31 @@ protected int responseStatus(RESPONSE response) {
     return accessor;
```
trask, line 148: this method looks unused now

trask (commented):
> The logic in service method exit itself was reduced to simply checking if an async listener has already been attached to this request

👍 this looks like a solid approach

instrumentation/servlet/servlet-common/library/src/main/java/io/opentelemetry/instrumentation/servlet/ServletAccessor.java
```diff
@@ -52,7 +52,8 @@
 
   boolean isRequestAsyncStarted(REQUEST request);
 
-  void addRequestAsyncListener(REQUEST request, ServletAsyncListener<RESPONSE> listener);
+  void addRequestAsyncListener(
+      REQUEST request, ServletAsyncListener<RESPONSE> listener, Object response);
```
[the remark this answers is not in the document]
agoallikmaa: Since it's taken from a request attribute, that would require doing a `(RESPONSE) response` unchecked cast in the tracer, which is not really nice and shows a warning in IntelliJ unless a `//noinspection` is added. Since the accessor implementations work with specific classes, they can do the `instanceof` check properly.

instrumentation/servlet/servlet-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v5_0/async/AsyncStartAdvice.java
agoallikmaa: Working only with the outermost request makes sense because wrappers/subclasses can both wrap the `AsyncContext` from delegates or superclasses and intercept attribute handling. Only manipulating the final request and `AsyncContext` that the application code also sees is in my opinion the safest option. Since exit advices are invoked in reverse order, then omitting that check would make it try to add the listener to the innermost request instead.

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyStartSpanAdvice.java
agoallikmaa: `event.getAsyncContext().getResponse()` does not seem to solve the current issue with Liberty as it is not available there either (and the spec allows it to be missing). Passing the response would mean storing the response as a request attribute for all servlet engines - I am not sure if we want to do that preemptively for all of them.

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/common/service/ServletAndFilterAdviceHelper.java
agoallikmaa: I actually added this comment before I encountered the issue with Liberty - so its purpose is to be a hint for debugging and to make it clear to the reader why we even need a separate instrumentation in the first place.

instrumentation/servlet/servlet-common/library/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
agoallikmaa: Removed.

instrumentation/servlet/servlet-common/library/src/main/java/io/opentelemetry/instrumentation/servlet/ServletAccessor.java
```diff
@@ -52,7 +52,8 @@
 
   boolean isRequestAsyncStarted(REQUEST request);
```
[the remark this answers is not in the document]
agoallikmaa: Removed.

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3AsyncStartAdvice.java
agoallikmaa: Explained in the other comment.

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/LibertyStartSpanAdvice.java
trask: If we don't pass request/response to AsyncContext.addListener(), won't AsyncEvent.getSuppliedResponse() (which we call in our listener) always return null?

> If the AsyncListener was added via AsyncContext.addListener(AsyncListener), this method must return null.

https://docs.oracle.com/javaee/7/api/javax/servlet/AsyncEvent.html#getSuppliedResponse--

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/common/service/ServletAndFilterAdviceHelper.java
agoallikmaa: Rephrased the comment

instrumentation/jetty/jetty-11.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jetty/v11_0/Jetty11HandlerAdvice.java
```diff
@@ -31,6 +32,8 @@ public static void onEnter(
 
     context = tracer().startServerSpan(request);
     scope = context.makeCurrent();
+
+    tracer().setAsyncListenerResponse(request, response);
```
trask, line 36: do we need to add calls to `setAsyncListenerResponse` in each app server instrumentation, or can we do it once in servlet instrumentation?

agoallikmaa: Since it is possible for requests to be handled entirely within app server handlers without servlet advices ever triggering - yes.

trask: can those requests have servlet async listeners?

agoallikmaa: Jetty handlers and Tomcat valves both expose the `ServletRequest` and are allowed to generate the response without a servlet ever being invoked. I have not tested using `startAsync` in them, but the documentation does not say that either of them would only support synchronous responses.

trask: can you add comments to the code explaining this? otherwise worried these lines could be removed thinking they aren't really needed, and I don't think we have any tests that it would break

instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/ThreadLocalContext.java
```diff
@@ -8,18 +8,21 @@
...
   private static final ThreadLocal<ThreadLocalContext> local = new ThreadLocal<>();
 
   private final HttpServletRequest req;
+  private final HttpServletResponse response;
   private Context context;
   private Scope scope;
   private boolean started;
 
-  private ThreadLocalContext(HttpServletRequest req) {
+  private ThreadLocalContext(HttpServletRequest req, HttpServletResponse response) {
```
laurit, line 23: could you rename `req` -> `request` in this class to make it consistent with our style of using verbose names

agoallikmaa: Done.

instrumentation/jetty/jetty-11.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jetty/v11_0/Jetty11HandlerAdvice.java
agoallikmaa: Added those comments for all the appservers.

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3AsyncStartAdvice.java
```diff
@@ -0,0 +1,40 @@
...
+    CallDepthThreadLocalMap.incrementCallDepth(AsyncContext.class);
+  }
+
+  @Advice.OnMethodExit(suppress = Throwable.class)
+  public static void startAsyncExit(
+      @Advice.This(typing = Assigner.Typing.DYNAMIC) HttpServletRequest request) {
```
laurit, line 25: I think that instead of dynamic typing to `HttpServletRequest` it would be better to use `@Advice.This ServletRequest request` and use instanceof+cast to get `HttpServletRequest`.

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/common/service/ServletAndFilterAdviceHelper.java
```diff
@@ -49,23 +46,28 @@
...
+  public static <REQUEST, RESPONSE> boolean mustEndOnHandlerMethodExit(
+      ServletHttpServerTracer<REQUEST, RESPONSE> tracer, REQUEST request) {
+
+    if (tracer.isAsyncListenerAttached(request)) {
+      // This request is handled asynchronously and startAsync instrumentation has already attached
+      // the listener.
+      return false;
     }
+
+    // This means that startAsync was not called (assuming startAsync instrumentation works
```
laurit, line 68: I guess now that you don't use `isAsyncStarted` any more the servlet engine correctness part isn't really relevant

instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatServerHandlerAdviceHelper.java
```diff
@@ -57,35 +56,30 @@
...
-        } catch (IllegalStateException e) {
-          // thrown by newer tomcats if request was already handled while setting the listener.
-          tracer.end(context, response);
-          return;
-        }
-      }
-
-      // In case the request finished before adding the listener on older tomcats...
-      if (!servletRequest.isAsyncStarted() && responseHandled.compareAndSet(false, true)) {
+    if (servletRequest != null) {
```
laurit, line 61: could combine these if statements

laurit (approved):

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/common/service/ServletAndFilterAdviceHelper.java
agoallikmaa: This is just a debugging hint that if for some reason async tests should fail in some new appserver configuration, then one thing to check would be to make sure the `startAsync` advice is triggered and doesn't throw.

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3AsyncStartAdvice.java
agoallikmaa: Done.

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3AsyncStartAdvice.java
```diff
@@ -0,0 +1,40 @@
...
+
+    int callDepth = CallDepthThreadLocalMap.decrementCallDepth(AsyncContext.class);
+
+    if (callDepth != 0) {
+      // This is not the outermost invocation, ignore.
+      return;
+    }
+
+    if (request != null) {
```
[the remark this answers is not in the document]
agoallikmaa: Replaced with `instanceof` due to the typing change.

instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatServerHandlerAdviceHelper.java
agoallikmaa: Done.

trask (approved):
@agoallikmaa thanks for your patience, and great find btw 👍

PR #2994, opened 2021-05-14
Switch to colon notation for dependencies.

anuraaga:
- Less verbose
- Format provided by search.maven.org and official docs in most places I've seen
- Consistent with `opentelemetry-java`


examples/distro/build.gradle
```diff
@@ -17,12 +17,12 @@ subprojects {
...
-      bytebuddyagent      : dependencies.create(group: 'net.bytebuddy', name: 'byte-buddy-agent', version: versions.bytebuddy),
+      bytebuddy           : dependencies.create("net.bytebuddy:byte-buddy:${versions.bytebuddy}"),
+      bytebuddyagent      : dependencies.create("net.bytebuddy:byte-buddy-agent:${versions.bytebuddy}"),
       autoservice         : [
-        dependencies.create(group: 'com.google.auto.service', name: 'auto-service', version: '1.0-rc7'),
-        dependencies.create(group: 'com.google.auto', name: 'auto-common', version: '0.8'),
-        dependencies.create(group: 'com.google.guava', name: 'guava', version: "${versions.guava}"),
+        dependencies.create("com.google.auto.service:auto-service:1.0-rc7"),
+        dependencies.create("com.google.auto:auto-common:0.8"),
+        dependencies.create("com.google.guava:guava:${versions.guava}"),
```
mateuszrzeszutek, line 25: Nit: I think that we don't need to call `dependencies.create()` anymore, we can just pass the string.

mateuszrzeszutek (approved):

iNikem (approved):

examples/distro/build.gradle
anuraaga: Yay! Thanks!

buildSrc/src/main/groovy/MuzzlePlugin.groovy
```diff
@@ -324,7 +324,7 @@ class MuzzlePlugin implements Plugin<Project> {
     def config = instrumentationProject.configurations.create(taskName)
 
     if (!muzzleDirective.coreJdk) {
-      def dep = instrumentationProject.dependencies.create("$versionArtifact.groupId:$versionArtifact.artifactId:$versionArtifact.version") {
```
mateuszrzeszutek, line 327: Wait, you probably need those in this file 😄

anuraaga: Yikes bad regex :)

PR #2996, opened 2021-05-14
Start migrating Muzzle plugin to Java

anuraaga:
When working on lazy configuration, it was painful to not have good compile time checking of muzzle plugin. Since we'll want to publish it to plugin portal, switching to Java so its same code as all our others seems like a good direction. I've migrated the extension objects in this one and will follow up with migrating Plugin.

While migrating, I switched to Gradle's modern `Property` API


mateuszrzeszutek (approved):
🎉

PR #3006, opened 2021-05-16
Re-enable smoke tests in release build

### LINKED ISSUE #2993: Reenable smoke tests for release build
We are temporarily disabling smoke tests in release build to address that it hasn't been correctly configured to split out smoke tests like the main build.

#2990

trask:
Resolves #2993

`test` and `smoke-test` jobs copied from `ci.yaml` (only change is removing `fail-fast: false`)


iNikem (approved):

.github/workflows/release-build.yml
```diff
@@ -8,7 +8,67 @@ on:
         required: true
 
 jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 8, 15 ]
```
anuraaga, line 15: Let's add 11 here (do we need to test all versions on release build?)

.github/workflows/release-build.yml
```diff
@@ -22,7 +82,7 @@ jobs:
         with:
           job-id: jdk11
           remote-build-cache-proxy-enabled: false
-          arguments: build final closeAndReleaseSonatypeStagingRepository --stacktrace -x :smoke-tests:test -Prelease.version=${{ github.event.inputs.version }}
+          arguments: build final closeAndReleaseSonatypeStagingRepository --stacktrace -x test -Prelease.version=${{ github.event.inputs.version }}
```
anuraaga, line 85: Let's either rename `build` -> `assemble` unless we're comfortable relying on build cache after adding `11` above

trask: I like assemble 👍

anuraaga (approved):

PR #3008, opened 2021-05-16
Avoid loading advice classes in agent class loader

trask:
From https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation-module.md#transformers:

> You probably noticed in the example above that the advice class is being referenced in a slightly peculiar way:
>
> ```java
> this.getClass().getName() + "$MethodAdvice"
> ```
>
> Simply referring to the inner class and calling `getName()` would be easier to read and understand than this odd mix of string concatenation, but please note that **this is intentional** and should be maintained.
>
> Instrumentation modules are loaded by the agent's class loader, and this string concatenation is an optimization that prevents the actual advice class from being loaded into the agent's class loader.

Moved things around a bit where there were top-level Advice classes:

* `XInstrumentationModule`
* `AInstrumentation`
  * `AInstrumentation`'s advice classes are nested inside `AInstrumentation`
* `BInstrumentation`
   * `BInstrumentation`'s advice classes are nested inside `BInstrumentation`

No other changes in this PR other than moving things around to the above structure and renaming of Advice classes.

Commits are split out by module in case you prefer to review that way.


iNikem:
Why advices have to be inner classes? I am not sure such blanket rule helps readability. Even in case of top-level classes you reference them by `this.class.getPackage()+".Advice"`, no?

trask:
> Why advices have to be inner classes? I am not sure such blanket rule helps readability. Even in case of top-level classes you reference them by `this.class.getPackage()+".Advice"`, no?

This is the convention we use most commonly. I think it does help readability, at least for Advice that are not shared across multiple TypeInstrumentation classes, as it keeps the advice class close to the pointcut definition.

iNikem (approved):

anuraaga (approved):

PR #3011, opened 2021-05-16
Replace embedded Redis server with testcontainers in tests

iNikem:


instrumentation/lettuce/lettuce-5.0/javaagent/src/test/groovy/LettuceReactiveClientTest.groovy
```diff
@@ -201,7 +195,7 @@ class LettuceReactiveClientTest extends AgentInstrumentationSpecification {
           attributes {
             "$SemanticAttributes.DB_SYSTEM.key" "redis"
             "$SemanticAttributes.DB_STATEMENT.key" "COMMAND"
-            "lettuce.command.results.count" 157
+            "lettuce.command.results.count" 224
```
trask, line 198: this seems fragile, and the specific value is not really important
```suggestion
            "lettuce.command.results.count" { it > 100 }
```

trask (approved):
nice

instrumentation/lettuce/lettuce-4.0/javaagent/src/test/groovy/LettuceAsyncClientTest.groovy
```diff
@@ -71,18 +70,12 @@ class LettuceAsyncClientTest extends AgentInstrumentationSpecification {
     dbUriNonExistent = "redis://" + dbAddrNonExistent
     embeddedDbUri = "redis://" + dbAddr
 
-    redisServer = RedisServer.builder()
-    // bind to localhost to avoid firewall popup
-      .setting("bind " + HOST)
-    // set max memory to avoid problems in CI
-      .setting("maxmemory 128M")
-      .port(port).build()
+    redisServer = redisServer.withFixedExposedPort(port, 6379)
```
anuraaga, line 73: Do we need `withFixedExposedPort` or can we use normal testcontainers to get a random port?

iNikem: Some test classes have tests that kill this container. Thus they restart that in `setup`. Which means that normal container with a random port will get a new port for each test. This can be done but requires more changes to setups. I decided against that, at least in this PR

anuraaga (approved):

PR #3012, opened 2021-05-16
Update Apache HttpClient 2.0 to Instrumenter API

trask:
One small part of #2713


instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/ApacheHttpClientNetAttributesExtractor.java
```diff
@@ -0,0 +1,31 @@
...
+  protected String transport(HttpMethod httpMethod) {
+    return SemanticAttributes.NetTransportValues.IP_TCP;
+  }
+
+  @Override
+  @Nullable
+  protected InetSocketAddress getAddress(HttpMethod httpMethod, @Nullable Void unused) {
+    HostConfiguration hostConfiguration = httpMethod.getHostConfiguration();
+    return hostConfiguration != null
+        ? new InetSocketAddress(hostConfiguration.getHost(), hostConfiguration.getPort())
```
anuraaga, line 28: If we don't have access to a `SocketAddress` I think we'd usually use the other `NetAttributesExtractor` to avoid creating this every time.

instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/ApacheHttpClientNetAttributesExtractor.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  protected @Nullable Long peerPort(HttpMethod httpMethod, @Nullable Void unused) {
+    HostConfiguration hostConfiguration = httpMethod.getHostConfiguration();
+    return hostConfiguration != null ? (long) hostConfiguration.getPort() : null;
+  }
+
+  @Override
+  protected @Nullable String peerIp(HttpMethod httpMethod, @Nullable Void unused) {
+    return null;
+  }
```
trask, line 37: Should we detect if `host` is an IP address and use `net.peer.ip` instead of `net.peer.name` in that case?

anuraaga: This has always been tricky - I think if the framework doesn't provide us a resolved address (e.g., `SocketAddress`) then we'd just populate `net.peer.name`, even if it's an IP address. Not sure if that's great though.

instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/HttpHeaderSetter.java
```diff
@@ -9,9 +9,7 @@
 import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpMethod;
 
-public class HttpHeadersInjectAdapter implements TextMapSetter<HttpMethod> {
-
-  public static final HttpHeadersInjectAdapter SETTER = new HttpHeadersInjectAdapter();
+public final class HttpHeaderSetter implements TextMapSetter<HttpMethod> {
```
mateuszrzeszutek, line 12: Nit: can be package-private?

instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/ApacheHttpClientHttpAttributesExtractor.java
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  @Override
+  @Nullable
+  protected Long responseContentLength(HttpMethod httpMethod, Void unused) {
+    return null;
+  }
+
+  @Override
+  protected @Nullable Long responseContentLengthUncompressed(HttpMethod httpMethod, Void unused) {
```
mateuszrzeszutek, line 101:
```suggestion
  @Override
  @Nullable
  protected Long responseContentLengthUncompressed(HttpMethod httpMethod, Void unused) {
```

mateuszrzeszutek (approved):

anuraaga (approved):

PR #3013, opened 2021-05-16
Rename local variables in test assertions to improve some groovy assert messages

trask:
Not sure if it's just me, but I always get confused which side of the assertion is expected, and which side is actual.

With this PR, now output will look like:

```
assert span.name == expected
       |    |    |  |
       |    |    |  'HTTP POST'
       |    |    false
       |    'HTTP GET'
```

where previously it looked like:

```
assert span.name == name
       |    |    |  |
       |    |    |  'HTTP POST'
       |    |    false
       |    'HTTP GET'
```


anuraaga (commented):
Could you add to the PR description more concretely what we changed? Especially with the refactoring in the PR it's hard for me to find the real change.

anuraaga (approved):

anuraaga:
Updated the title since someone not familiar with Spock would still be confused. Best is removing Spock of course :)

trask:
> Could you add to the PR description more concretely what we changed? Especially with the refactoring in the PR it's hard for me to find the real change.

Sorry, there were some unrelated changes also in this PR, I will submit those separately.

anuraaga (approved):

mateuszrzeszutek (approved):

laurit (approved):
:+1: I also have trouble telling which is actual and which expected

Issue #2685, opened 2021-03-31
How to instrument static initializers?


trask:
Instrumenting static initializers is useful if we want to execute some code on class initialization (since we can't typically execute code during transformation itself since that can load more classes which will then not be instrumented since `transform()` isn't called on classes loaded inside of `transform()`).

Not sure why not able to instrument them via normal `TypeInstrumentation.transformers()` advice, suspect maybe because the underlying class doesn't have a static initializer to instrument. We may need to use `LoadedTypeInitializer`, but that wasn't working for me either, need to investigate further.

see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2672#discussion_r604672178

PR #3019, opened 2021-05-17
Refactor TypeInstrumentation#transformers() method

mateuszrzeszutek:
This PR was inspired by DataDog's `AdviceTransformation` interface introduced in https://github.com/DataDog/dd-trace-java/pull/2673 (FYI @tylerbenson)

I've changed the `TypeInstrumentation` interface and introduced `TypeTransformer`:

```java
interface TypeInstrumentation {
  // ...

  // removed the map-returning transformers()
  void transform(TypeTransformer transformer);
}

interface TypeTransformer {

  void applyAdviceToMethod(
      ElementMatcher<? super MethodDescription> methodMatcher, String adviceClassName);

  void applyTransformer(AgentBuilder.Transformer transformer);
}
```

This has several merits:
* we no longer need to create those transformer maps (which I believe was the main reason for the DD change);
* this allows us to introduce type transformations other than advices - e.g. adding the static initializer (see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2685)

The PR is has lots of changed files, but 98% of it is just changing all existing `TypeInstrumentation`s. The most interesting parts are in `javaagent-extension-api` and `javaagent-tooling`.


laurit:
If `TypeTransformer` had `applyAdviceToMethod` that takes class instead of className we could build a byte buddy plugin similar to `MuzzleCodeGenerationPlugin` that translates from class to className and calls the string version of `applyAdviceToMethod`.

tylerbenson:
I would suggest splitting this into at least two commits to make it easier to review.

mateuszrzeszutek:
> I would suggest splitting this into at least two commits to make it easier to review.

That's a good idea, done. The first commit in this PR contains all the API changes, the second one is just changing all `TypeInstrumentation`s in the project.

tylerbenson (approved):
Much better.

instrumentation/akka-actor-2.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkaactor/AkkaActorCellInstrumentation.java
```diff
@@ -30,8 +28,8 @@
   }
 
   @Override
-  public Map<? extends ElementMatcher<? super MethodDescription>, String> transformers() {
```
trask, line 33: I will not miss this method signature 😂

trask (approved):

anuraaga (approved):
Awesome!

PR #3040, opened 2021-05-19
Refactor muzzle Reference class

mateuszrzeszutek:
* split out inner classes into separate files
* and since they're public API I made them a little bit better: added javadocs, renamed classes/methods

Getting the `...extension.muzzle` package stable is the first step to extracting the muzzle compile time plugin - the plugin may have a bit different release lifecycle than the main project and getting those classes stable will help a lot with eliminating breaking changes.


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/muzzle/Flag.java
```diff
@@ -0,0 +1,159 @@
...
+    PACKAGE_OR_HIGHER {
+      @Override
+      public boolean matches(int asmFlags) {
+        return !VisibilityFlag.PRIVATE.matches(asmFlags);
+      }
+    },
+    PRIVATE_OR_HIGHER {
+      @Override
+      public boolean matches(int asmFlags) {
+        // you can't out-private a private
```
trask, line 95: 😁

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/MuzzleCodeGenerator.java
```diff
@@ -279,68 +284,67 @@ private void generateMuzzleReferencesMethod(ReferenceCollector collector) {
...
           mv.visitLdcInsn(references[i].getClassName());
           mv.visitMethodInsn(
-              Opcodes.INVOKESPECIAL,
-              referenceBuilderType.getInternalName(),
-              "<init>",
-              "(Ljava/lang/String;)V",
+              Opcodes.INVOKESTATIC,
+              referenceType.getInternalName(),
+              "newBuilder",
+              Type.getMethodDescriptor(referenceBuilderType, stringType),
```
trask, line 299: yay for removing constructor calling bytecode!

trask (approved):
nice

PR #3041, opened 2021-05-19
Add Nullable to Instrumenter API response args

trask:
@anuraaga @mateuszrzeszutek just confirming, did you want to call these methods with null response for symmetry, or alternatively can we skip calling them if there's no response (and then don't need to null check in the implementations)?


trask:
I'm wondering if these particular methods should only be called with non-null response?

* statusCode()
* responseContentLength()
* responseContentLengthUncompressed()

instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpAttributesExtractor.java
```diff
@@ -53,28 +53,36 @@ protected String userAgent(RequestContext ctx) {
...
+    if (status.equals(HttpStatus.UNKNOWN)) {
+      return null;
     }
-    return null;
+    return (long) status.code();
   }
 
   @Override
-  protected String flavor(RequestContext ctx, RequestLog requestLog) {
+  protected String flavor(RequestContext ctx, @Nullable RequestLog requestLog) {
```
anuraaga, line 85: Hmm - I'd like to not leak `@Nullable` in when it's not null. I guess what we want is to "only allow null if Void". But not sure we can model that, so maybe it's ok as it is, does it cause any nullability issues right now?

trask: I think most instrumentations (other than armeria) will also pass `null` response to `end()` when exception occurs

anuraaga: Ah thanks - in that case yeah skipping the invocation if null seems easier for instrumentation

trask: added skipping of calls to

* statusCode()
* responseContentLength()
* responseContentLengthUncompressed()

when response is null

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpAttributesExtractor.java
```diff
@@ -85,27 +90,81 @@ protected final void onEnd(AttributesBuilder attributes, REQUEST request, RESPON
...
-  protected abstract Long requestContentLength(REQUEST request, RESPONSE response);
-
+  protected abstract Long requestContentLength(REQUEST request, @Nullable RESPONSE response);
+
+  /**
+   * Extracts the {@code http.request_content_length_uncompressed} span attribute.
+   *
+   * <p>This is called from both {@link Instrumenter#start(Context, Object)} and {@link
+   * Instrumenter#end(Context, Object, Object, Throwable)}. When it is called from {@code start},
+   * {@code response} will be {@code null}.
```
mateuszrzeszutek, line 108: Those methods aren't currently called in `start()`

anuraaga (approved):

mateuszrzeszutek (approved):

PR #3042, opened 2021-05-19
Update Apache HttpClient 4.0 to Instrumenter API

trask:
Another small part of #2713


instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientNetAttributesExtractor.java
```diff
@@ -0,0 +1,48 @@
...
+    if (port != -1) {
+      return (long) port;
+    }
+    switch (uri.getScheme()) {
+      case "http":
+        return 80L;
+      case "https":
+        return 443L;
+    }
+    return null;
```
trask, line 41: do you think I should extract `UriNetAttributesExtractor` similar to `InetSocketAddressNetAttributesExtractor`?

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientHelper.java
```diff
@@ -5,18 +5,21 @@
...
 
-import static io.opentelemetry.javaagent.instrumentation.apachehttpclient.v4_0.ApacheHttpClientTracer.tracer;
+import static io.opentelemetry.javaagent.instrumentation.apachehttpclient.v4_0.ApacheHttpClientInstrumenters.instrumenter;
 
 import io.opentelemetry.context.Context;
 import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpUriRequest;
 
-public class ApacheHttpClientHelper {
+// needs to be public to be accessed from inside of advice
```
trask, line 14: I guess public doesn't matter inside of javaagent module, may remove these comments

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientNetAttributesExtractor.java
anuraaga: Is it correct to say URI will be more common in automatic instrumentation vs library instrumentation since it's a result of not having access to the internals? If so, possibly not, to encourage library authors adding OTel support to expose the more useful type.

trask: I'm not sure how common yet, will know soon 😁 (won't extract anything common for now and we can revisit later)

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientHttpAttributesExtractor.java
```diff
@@ -0,0 +1,137 @@
...
+  @Override
+  protected String url(HttpUriRequest request) {
+    return request.getURI().toString();
+  }
+
+  @Override
+  protected String target(HttpUriRequest request) {
+    URI uri = request.getURI();
+    String queryString = uri.getQuery();
+    return queryString != null ? uri.getPath() + "?" + queryString : uri.getPath();
```
mateuszrzeszutek, line 39: According to javadocs `URI#getPath()` can return nulls:

```suggestion
    String pathString = uri.getPath();
    String queryString = uri.getQuery();
    if (pathString != null && queryString != null) {
      return pathString + "?" + queryString;
    } else if (queryString != null) {
      return "?" + queryString;
    } else {
      return pathString;
    }
```

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/HttpHeaderSetter.java
```diff
@@ -8,9 +8,8 @@
 import io.opentelemetry.context.propagation.TextMapSetter;
 import org.apache.http.client.methods.HttpUriRequest;
 
-class HttpHeadersInjectAdapter implements TextMapSetter<HttpUriRequest> {
-
-  public static final HttpHeadersInjectAdapter SETTER = new HttpHeadersInjectAdapter();
+enum HttpHeaderSetter implements TextMapSetter<HttpUriRequest> {
+  INSTANCE;
```
mateuszrzeszutek, line 12: Nit: we don't really need to have setters/getters be enums, we can just create a new instance and pass to the instrumenter.

mateuszrzeszutek (approved):

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientNetAttributesExtractor.java
mateuszrzeszutek: There's one advantage to using URI in the instrumenter API: if we used it everywhere then we could have encapsulate the userInfo sanitization logic in the attribute extractor (well, we still can do it right now, but with a `String url()` method we'd have to have one more type conversion)

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/HttpHeaderSetter.java
trask: will send a follow-up PR to change in all (few) existing places so @anuraaga can weigh in

trask: #3054

PR #3043, opened 2021-05-19
Change a couple of Longs to Integers in Instrumenter API

trask:
These two methods feel like they should return `Integer`, and it also makes implementing them a bit easier:

```
protected abstract Long peerPort(REQUEST request, RESPONSE response);

protected abstract Long statusCode(REQUEST request, RESPONSE response);
```

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpAttributesExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpAttributesExtractor.java
@@ -43,7 +43,10 @@ public abstract class HttpAttributesExtractor<REQUEST, RESPONSE>
         attributes,
         SemanticAttributes.HTTP_REQUEST_CONTENT_LENGTH_UNCOMPRESSED,
         requestContentLengthUncompressed(request, response));
-    set(attributes, SemanticAttributes.HTTP_STATUS_CODE, statusCode(request, response));
+    Integer statusCode = statusCode(request, response);
+    if (statusCode != null) {
+      set(attributes, SemanticAttributes.HTTP_STATUS_CODE, (long) statusCode);
+    }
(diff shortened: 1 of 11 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpAttributesExtractor.java
```diff
@@ -43,7 +43,10 @@ protected final void onEnd(AttributesBuilder attributes, REQUEST request, RESPON
         attributes,
         SemanticAttributes.HTTP_REQUEST_CONTENT_LENGTH_UNCOMPRESSED,
         requestContentLengthUncompressed(request, response));
-    set(attributes, SemanticAttributes.HTTP_STATUS_CODE, statusCode(request, response));
+    Integer statusCode = statusCode(request, response);
+    if (statusCode != null) {
```
iNikem, line 47: I wonder if we should use `OptionalInt` instead?

anuraaga: I think there's still reasonable chance of someone returning `null` even with `OptionalInt`, I guess nullable return has less chance of NPE.

anuraaga (approved):
While it adds grokkability for the type to match the attribute type, it does seem to significantly improve the instrumentation code

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpAttributesExtractor.java
mateuszrzeszutek: And we're never using any `Optional`s anywhere in the agent, so even if `OptionalInt` is a bit nicer and prettier ` @Nullable Integer` is more consistent with the codebase.

mateuszrzeszutek (approved):

PR #3048, opened 2021-05-20
Share some code between netty 4.0 and 4.1 instrumentation

laurit:
Move classes that are identical in netty 4.0 and 4.1 instrumentation into a common module. Implement removing our tracing handler together with original handler for netty 4.0. Copy `ChannelFutureTest` and `ChannelPipelineTest` to netty 4.0.


mateuszrzeszutek (approved):

trask (approved):
❤️

Issue #778, opened 2020-07-24
Support for subclassing instrumentation Tracers in custom distributions


trask:
Use some kind of lookup (SPI) to instantiate the instrumentation Tracer singletons so that they can be subclassed in custom distributions.

This will enable people to capture additional attributes from standard Request/Response type objects, e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/566#issuecomment-655942298, https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/776#issuecomment-663334381

A similar hook has proven useful in Brave per @anuraaga 👍.

pavolloffay:
I can work on this, but first I would like to clarify what would be the best approach.

Would each instrumentation define SPI to get its tracer or just an SPI to load the BaseTracer? Ideally the tracer subclasses should access typed request/response objects.

trask:
Great!

My first thought would be SPI lookup for each tracer, e.g. `io.opentelemetry.instrumentation.auto.grpc.server.GrpcServerTracer`, to see if there's a subclass available to use instead.

It's a bit odd to use SPI lookup for a (non-abstract) class though.

And a complication is that we'll need to inject the subclass as a "helper class", and any other classes that it brings along with it (e.g. anonymous inner classes).

So maybe a separate SPI makes sense for loading these kinds of customizations, that can return the normal tracer class name, the subclass to use instead, and a list of helper classes to inject along with the subclass. We could load all of these SPI classes at agent init (they can live in AgentClassLoader) and build a map with an instrumentation API on top of the map that all the instrumentation use to inject extra helper classes and instantiate their tracers (which will happen in the user class loader).

pavolloffay:
> It's a bit odd to use SPI lookup for a (non-abstract) class though.

Yes I wanted to avoid doing this, but I cannot think of a better approach. So we will define an SPI for all tracer classes. We can add support gradually, one by one.

trask:
Oh, there's a second proposal above that wasn't clear. E.g. create a single SPI interface something like:

```
interface CustomTracer {
  String getTracerClassName();
  String getCustomTracerClassName();
  List<String> getHelperClassNames();
}
```

And use ServiceLoader to load as many of these as are present during agent init and build a map keyed on tracer class name.

And add a method to instrumentation api, e.g.:

```
static <T> createTracer(Class<T> clazz)
```

[...]

PR #3050, opened 2021-05-20
[WIP] Add peer.service to Instrumenter API

mateuszrzeszutek:
We should add `peer.service` extractor to Instrumenter API as soon as possible so that we don't lose existing functionality when migrating tracers.
This attribute is only used in the agent - the problem is, `NetAttributesExtractor` implementations are package-private in library instrumentations and agent code currently has no way to inject another extractor that depends on the private net one.
For example:
```
private static final ArmeriaTracing TRACING = ArmeriaTracing.create(GlobalOpenTelemetry.get());
```

I thought of several options on how to handle this:
1. Just add the attribute to `NetAttributesExtractor`;
2. Create a separate extractor just for `peer.service` and keep it in instrumentation-api so that it's accessible to library instrumentation code; the library instrumentations have to add it even though it's useless without javaagent (this PR implements this solution);
3. Add a getter to the builder: `InstrumenterBuilder#attributeExtractors()` (and library instrumentations tracing builders...) and in the agent code find net attributes extractor using `instanceof` and if it's there, add the peer service one. This is hacky, and you have to do it in each javaagent instrumentation...
4. Add a `Function<InstrumenterBuilder, InstrumenterBuilder>` hook to instrumentation-api that'll be implemented in the javaagent (similar to e.g. rxjava hooks that we use in our instrumentations) that'll do the `instanceof` check. This option sort of implements https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/778

WDYT? I'd like to get your opinion on this matter as I don't have any strong preference towards any of these options (and maybe there's another one that I didn't think of)

TODO: tests, javadocs


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/PeerServiceAttributesExtractor.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.net;
```
anuraaga, line 6: I guess this should be in javaagent-api since it's only for auto instrumentation?

instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/ApacheHttpClientInstrumenters.java
```diff
@@ -27,13 +28,16 @@
...
+    ApacheHttpClientNetAttributesExtractor netAttributesExtractor =
+        new ApacheHttpClientNetAttributesExtractor();
 
     INSTRUMENTER =
         Instrumenter.<HttpMethod, Void>newBuilder(
                 GlobalOpenTelemetry.get(), INSTRUMENTATION_NAME, spanNameExtractor)
             .setSpanStatusExtractor(spanStatusExtractor)
             .addAttributesExtractor(httpAttributesExtractor)
-            .addAttributesExtractor(new ApacheHttpClientNetAttributesExtractor())
+            .addAttributesExtractor(netAttributesExtractor)
```
anuraaga, line 39: This is the approach I had in mind. I guess it's a bit tedious, but we could do it for now as a quick fix, luckily being agent code we're never stuck with it long term.

anuraaga (commented):
>Create a separate extractor just for peer.service and keep it in instrumentation-api so that it's accessible to library instrumentation code; the library instrumentations have to add it even though it's useless without javaagent (this PR implements this solution);

This doesn't seem to be the approach in this PR since you add it in javaagent instrumentation not library instrumentation, right?

4) Seems like an interesting option if we know other real use cases wonder if anything comes to mind?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/PeerServiceAttributesExtractor.java
mateuszrzeszutek: That's the problem: if I put it inside javaagent-api then I'd have to make library instrumentations (armeria, okhttp, ...) depend on javaagent-api to use that (because `NetAttributeExtractor` implementations are package-private and not visible from the javaagent code).

mateuszrzeszutek:
> 4. Seems like an interesting option if we know other real use cases wonder if anything comes to mind?

Adding custom attributes/metrics to built-in javaagent instrumentations?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/PeerServiceAttributesExtractor.java
anuraaga: Oops package names are so long so got confused. Hmm - yeah we definitely shouldn't use this code from library instrumentation, libraries are just supposed to be called with something like `.addAttributeExtractor(FixedAttributeExtractor.of(PEER_SERVICE, "my-service"))`, there's no point of the map approach outside of agent and it shouldn't be supported.

Should we just access the package private `ArmeriaNetAttributesExtractor` from the agent reflectively?

mateuszrzeszutek: > Should we just access the package private `ArmeriaNetAttributesExtractor` from the agent reflectively?

Hmm, that would work too -- I'll implement that option in this PR and we'll see how it looks.

instrumentation/armeria-1.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/armeria/v1_3/ArmeriaDecorators.java
```diff
@@ -6,20 +6,58 @@
...
-  public static final Function<? super HttpClient, ? extends HttpClient> CLIENT_DECORATOR =
-      TRACING.newClientDecorator();
+  public static final Function<? super HttpService, ? extends HttpService> SERVER_DECORATOR;
 
-  public static final Function<? super HttpService, ? extends HttpService> SERVER_DECORATOR =
-      TRACING.newServiceDecorator();
+  static {
+    ArmeriaTracingBuilder builder = ArmeriaTracing.newBuilder(GlobalOpenTelemetry.get());
+
+    Constructor<? extends NetAttributesExtractor<RequestContext, RequestLog>> constructor = null;
```
anuraaga, line 35: Should we go ahead and add a helper for this?

anuraaga (commented):
This seems to do a good job of keeping the agent-specific logic to the agent. It's too bad it makes the net attributes not follow the pattern of the others with `protected` methods. Maybe we should just go with `public` for all of them though, since they're user-implemented classes the visibility doesn't matter that much I guess.

mateuszrzeszutek:
> Maybe we should just go with public for all of them though, since they're user-implemented classes the visibility doesn't matter that much I guess.

+ 100
That'll even make testing a bit easier, since you'll be able to call `AttributesExtractor` methods directly:
```java
AttributesExtractor<?> ae = new MyExtractor<>();
// onStart is now public, can be called by other packages
ae.onStart(...);
```

mateuszrzeszutek:
I ended up making only `NetAttributesExtractor` public - I'll change all other extractors in a separate PR, to minimize merge conflicts in this one.

trask (approved):
I'm ok with this, thanks for getting the functionality back. Hopefully we'll see some patterns emerge in the future and come up with something nicer (or realize that this is just a weird one-off and we can live with it).

anuraaga (approved):

Issue #3027, opened 2021-05-18
java.lang.OutOfMemoryError: Java heap space occured in api-gateway of the Spring Petclinic Microservice variant


matthias-huber:
**Describe the bug**
I used the Spring Pet Clinic ([Microservice variant (commit 8818cc9)](https://github.com/spring-petclinic/spring-petclinic-microservices)) as basis. Added opentelemetry agent 1.2.0 to all microservices (api-gateway, vets-,visits-,customer-service). Traces are exporter to OTel collector. Everything runs with docker-compose. I also changed from internal db to mysql setup. However, I don't think that it is important as the api-gateway doesn't use the database.

I ran a small loadtest with around ~10 req/s and the api-gateway ran out of memory => 2021-05-18 12:58:53.080  WARN 19 --- [or-http-epoll-5] i.n.c.AbstractChannelHandlerContext      : An exception 'java.lang.OutOfMemoryError: Java heap space' [enable DEBUG level for full stacktrace] was thrown by a user handler's exceptionCaught() method while handling the following exception:

VisualVM shows the CPU usage of the api-gateway. It shows the typical behaviour of a (memory leak) that less and less memory can be freed after each GC:

![api-gateway-cpu](https://user-images.githubusercontent.com/13692153/118682744-00090000-b801-11eb-9e2e-8c21b891026d.png)

I took a heap dump of the api-gateway after the OutOfMemoryError occured.

Leak Suspect shows:
![petclinic_leak_suspect](https://user-images.githubusercontent.com/13692153/118683198-77d72a80-b801-11eb-92c9-29624086c8b6.png)

Histogram shows a lot of io.opentelemetry.javaagent.shaded.instrumentation.reactor.TracingSubscriber Instances

![TracingSubscriber](https://user-images.githubusercontent.com/13692153/118684570-a0135900-b802-11eb-85f3-81482db18772.png)

Merge Shortest Path to GC root (with all references) shows the leak suspect "io.opentelemetry.javaagent.bootstrap.instrumentation.context.FieldBackedProvider$ContextStore$io$netty$util$concurrent$GenericFutureListener$io$netty$util$concurrent$GenericFutureListener"

![TracingSubscriber_merge_shortest_root_gc](https://user-images.githubusercontent.com/13692153/118684592-a43f7680-b802-11eb-89b5-bc46f79a7f3e.png)

**Steps to reproduce**
[...]

PR #3059, opened 2021-05-21
Fix memory leak on netty

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3027
Instead of `ContextStore` use `Cache` with weak keys and weak values to store link between original listener and wrapper. `ContextStore` works fine when wrapper is stored in a field on original listener, but when listener class is a lambda instead of field it gets stored in a map with weak keys where original listener is key and wrapper is value. As wrapper has a strong reference to original this causes a memory leak.

### DIFF
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/FutureListenerWrappers.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/FutureListenerWrappers.java
@@ -19,6 +19,9 @@ public final class FutureListenerWrappers {
   // listener, but when listener class is a lambda instead of field it gets stored in a map with
   // weak keys where original listener is key and wrapper is value. As wrapper has a strong
   // reference to original listener this causes a memory leak.
+  // Also note that it's ok if the value is collected prior to the key, since this cache is only
+  // used to remove the wrapped listener from the netty future, and if the value is collected prior
+  // to the key, that means it's no longer used (referenced) by the netty future anyways.
   private static final Cache<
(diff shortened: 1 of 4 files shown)


mateuszrzeszutek (approved):

mateuszrzeszutek:
Can you add a comment to the `FutureListenerWrappers` class why a cache is used instead of context store? (basically what you wrote in the PR description)

mateuszrzeszutek (approved):

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/FutureListenerWrappers.java
```diff
@@ -11,85 +11,83 @@
...
 import io.opentelemetry.context.Scope;
-import io.opentelemetry.javaagent.instrumentation.api.ContextStore;
+import io.opentelemetry.instrumentation.api.caching.Cache;
 
 public final class FutureListenerWrappers {
+  // Instead of ContextStore use Cache with weak keys and weak values to store link between original
+  // listener and wrapper. ContextStore works fine when wrapper is stored in a field on original
+  // listener, but when listener class is a lambda instead of field it gets stored in a map with
+  // weak keys where original listener is key and wrapper is value. As wrapper has a strong
+  // reference to original listener this causes a memory leak.
```
trask, line 21:
```suggestion
  //
  // also note that it's ok if the value is collected prior to the key, since this cache is only
  // used to remove the wrapped listener from the netty future, and if the value is collected prior
  // to the key, that means it's no longer used (referenced) by the netty future anyways
```

trask (approved):

PR #3061, opened 2021-05-21
Vert.x Reactive HTTP server and client concurrency tests

agoallikmaa:
Enables the following tests:

* For `VertxRxHttpServerTest` and `VertxRxCircuitBreakerHttpServerTest`
  * "high concurrency test"
* For `VertxRxWebClientTest` and `VertxRxCircuitBreakerWebClientTest`
  * "high concurrency test"
  * "high concurrency test on single connection"
* For `VertxReactivePropagationTest`
  * Adds "should propagate context correctly over vert.x rx-java framework with high concurrency" which uses the same endpoint as "should propagate context over vert.x rx-java framework", but performs multiple concurrent requests similar to "high concurrency test" of server tests

### DIFF
--- a/instrumentation/vertx-reactive-3.5/javaagent/src/test/groovy/VertxReactivePropagationTest.groovy
+++ b/instrumentation/vertx-reactive-3.5/javaagent/src/test/groovy/VertxReactivePropagationTest.groovy
@@ -3,16 +3,26 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import static VertxReactiveWebServer.TEST_REQUEST_ID_ATTRIBUTE
+import static VertxReactiveWebServer.TEST_REQUEST_ID_PARAMETER
 import static io.opentelemetry.api.trace.SpanKind.CLIENT
 import static io.opentelemetry.api.trace.SpanKind.SERVER
 import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.SUCCESS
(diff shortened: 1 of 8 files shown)


instrumentation/vertx-reactive-3.5/javaagent/src/test/groovy/VertxReactivePropagationTest.groovy
```diff
@@ -85,5 +95,76 @@ class VertxReactivePropagationTest extends AgentInstrumentationSpecification {
...
+
+    when:
+    count.times { index ->
+      def job = {
+        latch.await()
+        Request.Builder builder = new Request.Builder().url("$baseUrl?$TEST_REQUEST_ID_PARAMETER=$index").get()
+
+        runUnderTrace("client " + index) {
+          Span.current().setAttribute(TEST_REQUEST_ID_ATTRIBUTE, index)
+          propagator.inject(Context.current(), builder, setter)
```
iNikem, line 118: Why do we need this manual propagation? Looks very suspicious to me...

agoallikmaa: The request is performed with OkHttp, which is not the library being tested here (as this is a server test), thus there is no `testInstrumentation` for it. This is the same logic used in `HttpServerTest`. This approach still makes sure the server instrumentation respects propagated headers without requiring instrumentation of the client.

iNikem (approved):

mateuszrzeszutek (approved):

PR #3064, opened 2021-05-21
Update jedis-1.4 to Instrumenter API

trask:


instrumentation/jedis/jedis-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v1_4/JedisDbAttributesExtractor.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  protected String name(JedisRequest request) {
+    return null;
+  }
+
+  @Override
+  protected String connectionString(JedisRequest request) {
+    Connection connection = request.getConnection();
+    return connection.getHost() + ":" + connection.getPort();
```
anuraaga, line 34: Since `JedisRequest` is our own class how about inlining these two business logic into it?

instrumentation/jedis/jedis-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v1_4/JedisRequest.java
```diff
@@ -0,0 +1,33 @@
...
+import static java.util.Collections.emptyList;
+
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import redis.clients.jedis.Connection;
+import redis.clients.jedis.Protocol;
+
+@AutoValue
+public abstract class JedisRequest {
+  private static final byte[][] NO_ARGS = new byte[0][];
```
mateuszrzeszutek, line 17: Unused?

mateuszrzeszutek (approved):

PR #3066, opened 2021-05-22
Update cassandra-3.0 to Instrumenter API

trask:


instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/TracingSession.java
```diff
@@ -48,97 +48,110 @@ public Session init() {
...
     try (Scope ignored = context.makeCurrent()) {
       resultSet = session.execute(query);
     } catch (Throwable t) {
-      tracer().endExceptionally(context, t);
+      instrumenter().end(context, request, null, t);
       throw t;
     }
-    tracer().end(context, resultSet.getExecutionInfo());
+    request.setExecutionInfo(resultSet.getExecutionInfo());
+    instrumenter().end(context, request, null, null);
```
trask, line 61: do you think this justifies adding `RESPONSE` type to `DbAttributesExtractor`?

anuraaga: `executionInfo` seems to only be used in net attributes, so it can have a `RESPONSE` I think?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/db/DbSpanNameExtractor.java
```diff
@@ -43,7 +43,8 @@ public String extract(REQUEST request) {
     if (dbName != null || table != null) {
       name.append(' ');
     }
-    if (dbName != null) {
+    // skip db name if table already has a db name prefixed to it
```
anuraaga, line 46: Can we add a unit test?

instrumentation/cassandra/cassandra-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v3_0/TracingSession.java
anuraaga: While working on gRPC instrumentation I understand what you're saying now :) Yeah I think it should, we should just have RESPONSE on all extractors it seems

mateuszrzeszutek: > do you think this justifies adding `RESPONSE` type to `DbAttributesExtractor`?

👍 
I think that none of our "default" attribute extractors should impose any type restrictions on the user; we should definitely use `RESPONSE` instead of hard-coded `Void`.

trask: I added issue #3076 and will do this as separate PR, since it will conflict with a few PRs currently in flight.

mateuszrzeszutek (approved):

anuraaga (approved):

PR #3068, opened 2021-05-22
Update lettuce-4.0 to Instrumenter API

trask:


instrumentation/lettuce/lettuce-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v4_0/LettuceConnectAttributesExtractor.java
```diff
@@ -0,0 +1,27 @@
...
+import com.lambdaworks.redis.RedisURI;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.instrumentation.api.instrumenter.AttributesExtractor;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+
+final class LettuceConnectAttributesExtractor extends AttributesExtractor<RedisURI, Void> {
+
+  @Override
+  protected void onStart(AttributesBuilder attributes, RedisURI redisUri) {
+    attributes.put(SemanticAttributes.DB_SYSTEM, SemanticAttributes.DbSystemValues.REDIS);
```
anuraaga, line 17: Don't think we need this

instrumentation/lettuce/lettuce-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v4_0/LettuceDbAttributesExtractor.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.lambdaworks.redis.protocol.RedisCommand;
+import io.opentelemetry.instrumentation.api.instrumenter.db.DbAttributesExtractor;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+final class LettuceDbAttributesExtractor extends DbAttributesExtractor<RedisCommand<?, ?, ?>> {
+  @Override
+  protected String system(RedisCommand<?, ?, ?> request) {
+    return SemanticAttributes.DbSystemValues.REDIS;
```
anuraaga, line 16: Or rather it looks like this class isn't used. Seems that we are seeing a pattern of annoyance in DB instrumentation since URI and statement are almost always different classes, unlike non-DB

anuraaga: Nevermind, I didn't notice we have two INSTRUMENTER but see it now

anuraaga (approved):

mateuszrzeszutek (approved):

PR #3069, opened 2021-05-23
Update redisson-3.0 to Instrumenter API

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2713](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2713)

Signed-off-by: dengliming <liming.d.pro@gmail.com>

### DIFF
--- a/instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
+++ b/instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
@@ -9,7 +9,6 @@ import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.channel.Channel;
 import io.opentelemetry.instrumentation.api.db.RedisCommandSanitizer;
 import io.opentelemetry.instrumentation.api.instrumenter.db.DbAttributesExtractor;
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
@@ -30,20 +29,21 @@ final class RedissonDbAttributesExtractor extends DbAttributesExtractor<Redisson
     return SemanticAttributes.DbSystemValues.REDIS;
   }
 
+  @Nullable
   @Override
-  protected @Nullable String user(RedissonRequest request) {
+  protected String user(RedissonRequest request) {
     return null;
   }
 
+  @Nullable
   @Override
-  protected @Nullable String name(RedissonRequest request) {
+  protected String name(RedissonRequest request) {
     return null;
   }
 
   @Override
   protected String connectionString(RedissonRequest request) {
-    Channel channel = request.getConnection().getChannel();
-    InetSocketAddress remoteAddress = (InetSocketAddress) channel.remoteAddress();
+    InetSocketAddress remoteAddress = request.getAddress();
     return remoteAddress.getHostString() + ":" + remoteAddress.getPort();
   }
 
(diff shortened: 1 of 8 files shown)


instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
```diff
@@ -11,34 +11,59 @@
...
+    return null;
+  }
 
-  private RedissonClientTracer() {
-    super(NetPeerAttributes.INSTANCE);
+  @Override
+  protected String connectionString(RedissonRequest request) {
+    Channel channel = request.getConnection().getChannel();
+    InetSocketAddress remoteAddress = (InetSocketAddress) channel.remoteAddress();
+    return remoteAddress.getHostString() + ":" + remoteAddress.getPort();
```
anuraaga, line 47: How about doing this business logic when creating `RedissonRequest`, especially to make sure we only sanitize once

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
```diff
@@ -11,34 +11,59 @@
...
+      default:
+        return String.join(";", sanitizedStatements);
+    }
   }
 
   @Override
-  protected String spanName(
-      RedisConnection connection, Object ignored, List<String> sanitizedStatements) {
+  protected String operation(RedissonRequest request) {
+    List<String> sanitizedStatements = sanitizeStatement(request);
```
anuraaga, line 66: Do you know why we use the sanitized statements here instead of directly reading from `Command.getName()` or plural?

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
```diff
@@ -11,34 +11,59 @@
...
   private static final String UNKNOWN_COMMAND = "Redis Command";
 
-  private static final RedissonClientTracer TRACER = new RedissonClientTracer();
+  @Override
+  protected String system(RedissonRequest request) {
+    return SemanticAttributes.DbSystemValues.REDIS;
+  }
+
+  @Override
+  protected @Nullable String user(RedissonRequest request) {
```
mateuszrzeszutek, line 34:
```suggestion
  @Nullable 
  @Override
  protected String user(RedissonRequest request) {
```

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
mateuszrzeszutek: I suppose it's a remnant of old tracer implementation - it should be fine to use `command.getName()` for operation.

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonDbAttributesExtractor.java
```diff
@@ -52,18 +77,8 @@ protected String spanName(
     }
```
mateuszrzeszutek, line 77: Let's just return null when size != 1 -- that's what [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes) says about `db.operation`:

> If the SQL statement has an ambiguous operation, or performs more than one operation, this value may be omitted.

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonNetAttributesExtractor.java
```diff
@@ -0,0 +1,26 @@
...
+    extends InetSocketAddressNetAttributesExtractor<RedissonRequest, Void> {
+
+  @Override
+  protected InetSocketAddress getAddress(RedissonRequest request, @Nullable Void unused) {
+    Channel channel = request.getConnection().getChannel();
+    return (InetSocketAddress) channel.remoteAddress();
+  }
+
+  @Override
+  protected @Nullable String transport(RedissonRequest request) {
```
mateuszrzeszutek, line 23:
```suggestion
  @Nullable
  @Override
  protected String transport(RedissonRequest request) {
```

mateuszrzeszutek (approved):

anuraaga (approved):

trask (approved):
thanks @dengliming!

PR #3070, opened 2021-05-23
Update methods-javaagent to Instrumenter API

dengliming:
for [https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2713](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2713)

Signed-off-by: dengliming <liming.d.pro@gmail.com>


instrumentation/methods/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/methods/MethodSpanNameExtractor.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.methods;
+
+import static io.opentelemetry.instrumentation.api.tracer.BaseTracer.spanNameForMethod;
```
anuraaga, line 8: Instead of using the method from `BaseTracer`, let's add to `instrumentation` API when needed. I think generating a span name from a method is common so we could have a util class / unit tests for it.

dengliming: I'm a little confused why is `BaseTracer.spanNameForMethod` method declared `public static`. How about moving these common methods(eg. spanNameForMethod) to ` io.opentelemetry.javaagent.tooling.Utils `. WDYT? @anuraaga : )

anuraaga: It being public static is legacy, it should probably have a module but we just added it to BaseTracer organically.

We don't want any instrumentation utils in `javaagent-tooling` since they're used outside the agent. It would go in `instrumentation-api`. How about adding a `SpanNames` class there fo rnow

anuraaga: By the way for some context, while ideally we just have a `SpanNameExtractor<Method>` in that module itself, I'm not sure we'd be able to handle cases like this one without a util class anyways

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jaxrs/jaxrs-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/JaxRsAnnotationsTracer.java#L50

But maybe others have some thoughts on that too

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
```diff
@@ -0,0 +1,53 @@
...
+    String className = clazz.getName();
+    if (clazz.getPackage() != null) {
+      String pkgName = clazz.getPackage().getName();
+      if (!pkgName.isEmpty()) {
+        className = className.substring(pkgName.length() + 1);
+      }
+    }
+    return className;
+  }
+}
```
mateuszrzeszutek, line 53: Nit: please add a private default constructor:
```suggestion
  private SpanNames() {}
}
```

instrumentation/methods/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/methods/MethodInstrumenters.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.reflect.Method;
+
+public final class MethodInstrumenters {
+  private static final String INSTRUMENTATION_NAME =
+      "io.opentelemetry.javaagent.external-annotations";
+
+  private static final Instrumenter<Method, Void> INSTRUMENTER;
+
+  static {
+    SpanNameExtractor<Method> spanName = new MethodSpanNameExtractor();
```
mateuszrzeszutek, line 21: You don't need a separate class, you can use a method reference:
```suggestion
    SpanNameExtractor<Method> spanName = SpanNames::spanNameForMethod;
```

mateuszrzeszutek (approved):

instrumentation/methods/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/methods/MethodSpanNameExtractor.java
mateuszrzeszutek: > By the way for some context, while ideally we just have a `SpanNameExtractor<Method>` in that module itself, I'm not sure we'd be able to handle cases like this one without a util class anyways

We'll have to think of something once we get there 😅 
Ideally I'd like us to have `SpanNameExtractor`s for class/method named spans (and `AttributeExtractor` for `code.namespace` & `code.function`) in instrumentation-api, but we can think of how to implement them in one of the next PRs.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  public static String spanNameForMethod(Method method) {
+    return spanNameForMethod(method.getDeclaringClass(), method.getName());
+  }
+
+  /**
+   * This method is used to generate an acceptable span (operation) name based on a given method
+   * reference. Anonymous classes are named based on their parent.
+   */
+  public static String spanNameForMethod(Class<?> clazz, @Nullable Method method) {
```
anuraaga, line 24: Not for this PR, just for everyone's thought I feel like this method probably isn't necessary, does / should `clazz` not always be `method.getDeclaringClass()`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  public static String spanNameForMethod(Class<?> cl, String methodName) {
+    return spanNameForClass(cl) + "." + methodName;
+  }
+
+  /**
+   * This method is used to generate an acceptable span (operation) name based on a given class
+   * reference. Anonymous classes are named based on their parent.
+   */
+  public static String spanNameForClass(Class<?> clazz) {
```
anuraaga, line 40: Not for this PR, just for everyone's thought, I feel like this method shouldn't be needed (publicly) either we should audit its usage, spans are methods, not classes.

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
dengliming: yeah. you were right. let't remove this and use `method.getDeclaringClass()`

trask (approved):
thanks @dengliming!

PR #3073, opened 2021-05-24
Migrate gRPC to Instrumenter API

anuraaga:


instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingServerInterceptor.java
```diff
@@ -127,29 +128,28 @@ public void onCancel() {
...
     @Override
     public void onComplete() {
       try {
         delegate().onComplete();
       } catch (Throwable e) {
-        tracer.endExceptionally(context, e);
+        instrumenter.end(context, request, null, e);
         throw e;
       }
-      tracer.end(context);
```
anuraaga, line 144: I think ending in `close` handles this, either that or we don't have any unit test covering this case

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcSpanStatusExtractor.java
```diff
@@ -0,0 +1,33 @@
...
+
+final class GrpcSpanStatusExtractor implements SpanStatusExtractor<GrpcRequest, Status> {
+  @Override
+  public StatusCode extract(GrpcRequest grpcRequest, Status status, @Nullable Throwable error) {
+    if (status == null) {
+      if (error instanceof StatusRuntimeException) {
+        status = ((StatusRuntimeException) error).getStatus();
+      } else if (error instanceof StatusException) {
+        status = ((StatusException) error).getStatus();
+      }
```
mateuszrzeszutek, line 23: Don't these exceptions get unwrapped by `GrpcErrorCauseExtractor`?

anuraaga: Not if there's no cause, which is a common usage for these too, e.g. just `throw Status.INVALID_ARGUMENT.asRuntimeException()`

anuraaga: Hmm - I think it means one corner case for `Status.OK.withCause(fatalException).asRuntimeException()`, while `Status` is OK, the OTel `StatusCode` would be error for this... Do you think it could be better to pass the wrapped exception to this instead of unwrapped (only use unwrapped for `Span.recordException`)?

mateuszrzeszutek: Hmm, originally I thought that it'd be better to always unwrap the exception and pass the unwrapped one everywhere: `span.recordException()`, `SpanStatusExtractor` and `RequestListener#end()` once we add `Throwable` arg there. The `ErrorCauseExtractor` was meant to remove those meaningless exception wrappers like `ExecutionException`.
In the gRPC instrumentation maybe it'd be better not to unwrap `StatusException`/`StatusRuntimeException`? They seem to carry some information with them (the status) that should be available when setting span status.

anuraaga: The main reason I added the cause extractor was to get the existing tests to pass :) But I guess `recordException` with the status exceptions could be considered better, yeah the exception message will have some information about the status.

instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcTest.groovy
```diff
@@ -276,15 +288,15 @@ abstract class AbstractGrpcTest extends InstrumentationSpecification {
               "message.id" 1
             }
           }
-          if (grpcStatus.cause != null) {
-            errorEvent grpcStatus.cause.class, grpcStatus.cause.message, 1
-          }
+          errorEvent grpcStatus.asRuntimeException().class, grpcStatus.asRuntimeException().message, 1
```
anuraaga, line 291: Note the change in telemetry here. But I think it does make sense to record the StatusRuntimeException (otherwise user could just throw a normal RuntimeException)

mateuszrzeszutek (approved):

instrumentation/grpc-1.6/library/grpc-1.6-library.gradle
```diff
@@ -12,4 +12,7 @@ dependencies {
 
   testImplementation deps.assertj
   testImplementation project(':instrumentation:grpc-1.6:testing')
+
+  compileOnly deps.autoValueAnnotations
+  annotationProcessor deps.autoValue
```
trask, line 17: looks like didn't end up using this?
```suggestion
```

trask (approved):

PR #3077, opened 2021-05-24
Enable Spring WebFlux concurrency tests

agoallikmaa:
Enables "high concurrency test on single connection" for `SpringWebfluxHttpClientTest`.

WebFlux previously had only its own custom server tests and did have any implementations of `HttpServerTest`. However there is no reason to not have it for WebFlux, so added 4 different configurations of `HttpServerTest`:

* `ImmediateControllerSpringWebFluxServerTest` - uses controller methods, throws exceptions and creates "controller" span inside controller method scope.
* `ImmediateHandlerSpringWebFluxServerTest` - uses route handlers, throws exceptions and creates "controller" span inside route handler method scope.
* `DelayedControllerSpringWebFluxServerTest` - uses controller methods, throws exceptions and creates "controller" span in a `Mono` map step (this step is also artificially delayed).
* `DelayedHandlerSpringWebFluxServerTest` - uses route handlers, throws exceptions and creates "controller" span in a `Mono` map step (this step is also artificially delayed).

Each of these also has the "high concurrency test" enabled.

### DIFF
--- a/instrumentation/spring/spring-webflux-5.0/javaagent/src/test/groovy/client/SpringWebFluxSingleConnection.groovy
+++ b/instrumentation/spring/spring-webflux-5.0/javaagent/src/test/groovy/client/SpringWebFluxSingleConnection.groovy
@@ -1,0 +1,67 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package client
+
+import io.netty.channel.ChannelOption
+import io.opentelemetry.instrumentation.test.base.HttpClientTest
+import io.opentelemetry.instrumentation.test.base.SingleConnection
+import org.springframework.http.HttpMethod
+import org.springframework.http.client.reactive.ReactorClientHttpConnector
+import org.springframework.web.reactive.function.client.WebClient
+import reactor.ipc.netty.http.client.HttpClientOptions
+import reactor.ipc.netty.resources.PoolResources
+
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.TimeoutException
+
+class SpringWebFluxSingleConnection implements SingleConnection {
+  private final ReactorClientHttpConnector connector
+  private final String host
+  private final int port
+
+  SpringWebFluxSingleConnection(boolean isOldVersion, String host, int port) {
+    if (isOldVersion) {
+      connector = new ReactorClientHttpConnector({ HttpClientOptions.Builder clientOptions ->
+        clientOptions.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, HttpClientTest.CONNECT_TIMEOUT_MS)
+        clientOptions.poolResources(PoolResources.fixed("pool", 1, HttpClientTest.CONNECT_TIMEOUT_MS))
+      })
+    } else {
+      def httpClient = reactor.netty.http.client.HttpClient.create(reactor.netty.resources.create("pool", 1)).tcpConfiguration({ tcpClient ->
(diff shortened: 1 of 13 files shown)


laurit (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/java/server/base/ServerTestController.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @GetMapping("/query")
+  public Mono<String> query_param(ServerHttpRequest request, ServerHttpResponse response) {
+    ServerEndpoint endpoint = ServerEndpoint.QUERY_PARAM;
+
+    return wrapControllerMethod(
+        endpoint,
+        () -> {
+          setStatus(response, ServerEndpoint.QUERY_PARAM);
```
mateuszrzeszutek, line 39: Nit:
```suggestion
          setStatus(response, endpoint);
```

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/java/server/base/ServerTestController.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @GetMapping("/path/{id}/param")
+  Mono<String> path_param(ServerHttpResponse response, @PathVariable("id") String id) {
+    ServerEndpoint endpoint = ServerEndpoint.PATH_PARAM;
+
+    return wrapControllerMethod(
+        endpoint,
+        () -> {
+          setStatus(response, ServerEndpoint.PATH_PARAM);
```
mateuszrzeszutek, line 87: Nit:
```suggestion
          setStatus(response, endpoint);
```

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/java/server/base/ServerTestController.java
```diff
@@ -0,0 +1,112 @@
...
+  Mono<String> indexed_child(ServerHttpRequest request, ServerHttpResponse response) {
+    ServerEndpoint endpoint = ServerEndpoint.INDEXED_CHILD;
+
+    return wrapControllerMethod(
+        endpoint,
+        () -> {
+          Span.current()
+              .setAttribute(
+                  "test.request.id", Long.parseLong(request.getQueryParams().getFirst("id")));
+          setStatus(response, ServerEndpoint.INDEXED_CHILD);
```
mateuszrzeszutek, line 102: Nit:
```suggestion
          setStatus(response, endpoint);
```

mateuszrzeszutek (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/java/server/base/ServerTestController.java
agoallikmaa: Done.

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/java/server/base/ServerTestController.java
agoallikmaa: Done.

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/java/server/base/ServerTestController.java
agoallikmaa: Done.

PR #3081, opened 2021-05-25
Configure kotlin directly in only project that uses it and enable tes…

anuraaga:
…tLatestDeps

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -19,8 +19,6 @@ ext {
     logback           : "1.2.3",
     bytebuddy         : "1.10.18", // Also explicitly specified in buildSrc
     scala             : "2.11.12",
-    kotlin            : "1.4.0",
-    coroutines        : "1.3.0",
     springboot        : "2.3.1.RELEASE",
     junit5            : "5.7.1",
     checkerFramework  : "3.6.1",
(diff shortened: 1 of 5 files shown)


gradle/dependencies.gradle
```diff
@@ -19,8 +19,6 @@ ext {
     logback           : "1.2.3",
     bytebuddy         : "1.10.18", // Also explicitly specified in buildSrc
     scala             : "2.11.12",
-    kotlin            : "1.4.0",
-    coroutines        : "1.3.0",
```
anuraaga, line 23: `coroutines` is the library we instrument so it's more consistent for the version to be in the project

settings.gradle
```diff
@@ -7,6 +7,7 @@ pluginManagement {
     id "io.github.gradle-nexus.publish-plugin" version "1.0.0"
     id "me.champeau.jmh" version "0.6.4"
     id "net.ltgt.errorprone" version "1.3.0"
+    id 'org.jetbrains.kotlin.jvm' version '1.5.10'
```
anuraaga, line 10: Kotlin version needs to match gradle plugin version, so figured best to just manage it same as other plugins

instrumentation/kotlinx-coroutines/javaagent/kotlinx-coroutines-javaagent.gradle
```diff
@@ -18,6 +21,13 @@ dependencies {
...
 
   testImplementation deps.opentelemetryKotlin
-  testImplementation deps.kotlin
-  testImplementation deps.coroutines
+  testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+  // Use first version with flow support since we have tests for it.
+  testLibrary "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.0"
+}
+
+tasks.named('compileTestGroovy').configure {
```
anuraaga, line 29: I can move this again if we end up with more projects with kotlin plugin

iNikem (approved):

mateuszrzeszutek (approved):

PR #3084, opened 2021-05-26
Don't change snapshot version setting and use shared repository decla…

anuraaga:
…ration

We don't use snapshots much anymore and generally used the precise snapshot version anyways. I think I notice a speedup in configuration time by removing this, but could be placebo. Think it's fine to remove though.

Also moved repository declaration to top level using newer Gradle feature.


trask (approved):

mateuszrzeszutek (approved):

instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-arquillian-testing/jaxrs-2.0-arquillian-testing.gradle
```diff
@@ -5,7 +5,9 @@ apply from: "$rootDir/gradle/java.gradle"
 
 // add repo for org.gradle:gradle-tooling-api which org.jboss.shrinkwrap.resolver:shrinkwrap-resolver-gradle-depchain depends on
 repositories {
+  mavenCentral()
```
iNikem, line 8: Why do we need to add this to individual projects? I thought that `settings.gradle` will take care of that?

anuraaga: Unfortunately it seems `repositories` overrides what's in `settings.gradle`, not add to it. So for the few times we need custom repos, we need to define them all. It's not too common so I guess it's OK

PR #3086, opened 2021-05-26
Fix merge conflict

anuraaga:


dependencyManagement/build.gradle.kts
```diff
@@ -0,0 +1,118 @@
...
+fun isNonStable(version: String): Boolean {
+  val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.toUpperCase().contains(it) }
+  val regex = "^[0-9,.v-]+(-r)?$".toRegex()
+  val isGuava = version.endsWith("-jre")
+  val isStable = stableKeyword || regex.matches(version) || isGuava
+  return isStable.not()
+}
+
+tasks {
+  named<DependencyUpdatesTask>("dependencyUpdates") {
```
mateuszrzeszutek, line 110: Nice!

By the way, are we switching the instrumentation project gradle files to kotlin too?

mateuszrzeszutek (approved):

dependencyManagement/build.gradle.kts
anuraaga: Doh! I accidentally added this into the wrong PR, still working on it :) I need to verify the impact of all of our projects and performance of kotlin but I am hoping to do so. But before that I want to make the dependency management consistent with SDK repo

laurit (approved):

PR #3099, opened 2021-05-27
Akka propagation fix and concurrency tests

agoallikmaa:
* We previously propagated context to an Akka `Mailbox` which caused context leak and broke concurrency tests due to the validation in `ContextPropagationDebug`. The class name of the `Mailbox` is `akka.dispatch.Dispatcher$$anon$1` as it is created in `akka.dispatch.Dispatcher#createMailbox` using `with <trait>` syntax which actually creates an anonymous class. Excluded it based on that name for now as for the versions currently being tested, this is the only anonymous class present in that file.
* Enabled `high concurrency test` for Akka `HttpServerTest` and `HttpClientTest` implementations.
* Changed `ContextPropagationDebug` to record not just the stack trace of each propagation, but the name of the class that carried the context. This made it much simpler to debug what classes we actually attach the context to.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
@@ -20,7 +20,7 @@ public final class ContextPropagationDebug {
 
   // locations where the context was propagated to another thread (tracking multiple steps is
   // helpful in akka where there is so much recursive async spawning of new work)
-  private static final ContextKey<List<StackTraceElement[]>> THREAD_PROPAGATION_LOCATIONS =
+  private static final ContextKey<List<Propagation>> THREAD_PROPAGATION_LOCATIONS =
       ContextKey.named("thread-propagation-locations");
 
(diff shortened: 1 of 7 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
```diff
@@ -67,18 +68,20 @@ public static void debugContextLeakIfEnabled() {
...
+    if (propagations != null) {
       StringBuilder sb = new StringBuilder();
-      Iterator<StackTraceElement[]> i = locations.iterator();
+      Iterator<Propagation> i = propagations.iterator();
       while (i.hasNext()) {
-        for (StackTraceElement ste : i.next()) {
-          sb.append("\n");
+        Propagation entry = i.next();
+        sb.append("\ncarrier of type: ").append(entry.carrierClassName);
+        for (StackTraceElement ste : entry.location) {
```
mateuszrzeszutek, line 83: Nice! 👍

mateuszrzeszutek (approved):

iNikem (approved):

laurit (approved):

trask (approved):

PR #3100, opened 2021-05-27
Make TypeInstrumentations top-level classes (and Advice classes nested in them)

trask:
When navigating around all of the instrumentations, I find it helpful for this to be consistent, instead of sometimes type instrumentations being nested inside of instrumentation module classes, and sometimes advice being top-level.

This seems to be the pattern we have settled on, so would would like to apply it to remaining locations.

This PR doesn't hit all the remaining locations, just wanted to float this PR first to check if any objections first, and if none, will follow up with another PR to do the rest. Will try to avoid conflicts with any inflight PRs, since moving things around like this creates horrific diffs.


anuraaga (approved):
I've been quite annoyed when we have advice classes not nested so this looks great to me. The reason is we can't ctrl-click to the definition with IntelliJ, while with instrumentation classes we can so no problem having at top level

mateuszrzeszutek (approved):
💯 👍

iNikem (approved):

Issue #3049, opened 2021-05-20
Flaky test ArmeriaHttpClientTest trace request with callback and parent


laurit:
https://scans.gradle.com/s/gbcpfr5wicl5e
```
assert traces.size() == expectedSize	
       |      |      |  |	
       |      2      |  1	
       |             false	
[...]

PR #3108, opened 2021-05-27
Propagate context to armeria callbacks

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3049

### DIFF
--- a/instrumentation/armeria-1.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/armeria/v1_3/AbstractStreamMessageSubscriptionInstrumentation.java
+++ b/instrumentation/armeria-1.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/armeria/v1_3/AbstractStreamMessageSubscriptionInstrumentation.java
@@ -9,10 +9,8 @@ import static net.bytebuddy.matcher.ElementMatchers.isConstructor;
 import static net.bytebuddy.matcher.ElementMatchers.named;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
-import io.opentelemetry.context.Context;
 import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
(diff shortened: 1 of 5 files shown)


anuraaga (approved):
Will eventually add proper context bridge but this seems good to fix build for now.

instrumentation/armeria-1.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/armeria/v1_3/AbstractStreamMessageSubscriptionInstrumentation.java
```diff
@@ -0,0 +1,47 @@
...
+            .and(takesArgument(1, named("org.reactivestreams.Subscriber"))),
+        AbstractStreamMessageSubscriptionInstrumentation.class.getName() + "$WrapSubscriberAdvice");
+  }
+
+  public static class WrapSubscriberAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void attachContext(
+        @Advice.Argument(value = 1, readOnly = false) Subscriber subscriber) {
+      Context context = Java8BytecodeBridge.currentContext();
+      if (context != Context.root()) {
```
mateuszrzeszutek, line 42: `Context.root()` is a static interface method too, you probably need a new bridge method for that one.

mateuszrzeszutek (approved):

instrumentation/armeria-1.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/armeria/v1_3/AbstractStreamMessageSubscriptionInstrumentation.java
laurit: moved the whole wrapping logic to a static method in `SubscriberWrapper`

PR #3112, opened 2021-05-27
Change smoketests to use testcontainers httpclient5 transport

laurit:
Sometimes smoke tests timeout after 45 minutes. From the logs server is started, ready check doesn't time out but no tests seem to be executed as if it got stuck between starting the server and executing tests. Similar behaviour is reported in https://github.com/testcontainers/testcontainers-java/issues/3531 though as we don't have a thread dump it is not possible to say whether we hit the same issue. Anyway using httpclient5 transport is reported to fix this.


mateuszrzeszutek (approved):

smoke-tests/src/test/resources/testcontainers.properties
```diff
@@ -0,0 +1 @@
+transport.type=httpclient5
```
trask, line 1:
```suggestion
# https://github.com/testcontainers/testcontainers-java/issues/3531
transport.type=httpclient5
```

trask (approved):

PR #3113, opened 2021-05-27
Use dependency management pattern for dependency versions

anuraaga:
Matches pattern we use in SDK.

`dependencies.gradle` is run against every single project, despite being just a bunch of constants. Not the slowest script but it all adds up to hundreds of ms during configuration time. There are a few approaches to make it happen only one time but decided to take the opportunity to match the style to SDK for dependency declarations.

That was the impetus - even more important benefit is probably applying version management to transitive dependencies. We don't want to force libraries we instrument to new versions since we always compile against old and test against new. But transitive we don't have to do that. I will follow up with more management in the future but added jackson as a starting point for huge bang for the buck. Compare this to your IntelliJ ;)

image

One big pain has been the time it takes to first index the project, or reindex on IntelliJ updates. I think more management will help a lot with it. With some care, we should actually be able to use management for instrumented libraries too will try later - I'm looking at you Netty.


bom-alpha/bom-alpha.gradle
```diff
@@ -9,7 +9,7 @@ group = "io.opentelemetry.instrumentation"
 archivesBaseName = "opentelemetry-instrumentation-bom-alpha"
 
 rootProject.subprojects.forEach { subproject ->
-  if (!project.name.startsWith("bom")) {
+  if (!subproject.name.startsWith("bom")) {
```
anuraaga, line 12: This was a bug the BOM may have been missing some of the top level projects as Gradle evaluates BFS, alphabetical (BOM users generally wouldn't have noticed though I think)

instrumentation/dropwizard-testing/dropwizard-testing.gradle
```diff
@@ -10,5 +10,8 @@ dependencies {
   // First version with DropwizardTestSupport:
   testImplementation "io.dropwizard:dropwizard-testing:0.8.0"
   testImplementation "javax.xml.bind:jaxb-api:2.2.3"
-  testImplementation "com.fasterxml.jackson.module:jackson-module-afterburner:2.9.10"
+  testImplementation "com.fasterxml.jackson.module:jackson-module-afterburner"
 }
+
+// Requires old Guava. Can't use enforcedPlatform since predates BOM
```
anuraaga, line 16: Sometimes forcing to older versions is a reasonable tradeoff IMO.

instrumentation/spring/spring-webflux-5.0/javaagent/src/test/groovy/client/SpringWebFluxSingleConnection.groovy
```diff
@@ -29,7 +31,8 @@ class SpringWebFluxSingleConnection implements SingleConnection {
         clientOptions.poolResources(PoolResources.fixed("pool", 1, HttpClientTest.CONNECT_TIMEOUT_MS))
       })
     } else {
-      def httpClient = reactor.netty.http.client.HttpClient.create(reactor.netty.resources.create("pool", 1)).tcpConfiguration({ tcpClient ->
```
anuraaga, line 32: I confirmed on `main` that this branch never runs even for latest deps test. No clue why. Also no clue why my PR fixed it (Dr. Murphy probably involved). All the changes in this package were to get real latest deps test to work.

gradle/java.gradle
```diff
@@ -69,20 +69,40 @@ repositories {
...
+  compileOnly "org.checkerframework:checker-qual"
+
+  testImplementation "org.junit.jupiter:junit-jupiter-api"
+  testImplementation "org.junit.jupiter:junit-jupiter-params"
+  testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine"
+  testRuntimeOnly "org.junit.vintage:junit-vintage-engine"
+
+  testImplementation "org.objenesis:objenesis"
+  testImplementation "org.spockframework:spock-core"
+  testImplementation "ch.qos.logback:logback-classic"
```
anuraaga, line 100: I didn't bother with deduping a lot of repetition related to logging, and also other libraries. Will follow up

trask (approved):
🎉🎉

Issue #3103, opened 2021-05-27
Netty client spans don't capture any span attributes on connection failure


trask:
Not sure if it's even possible, but would be nice if Netty instrumentation could capture the same span attributes on connection failure as other http client instrumentation.

Currently it captures a span named `CONNECT` with no attributes:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/0a34b2b0562dc7b2a6fd4aab18990f25587c4d66/instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/ChannelFutureListenerInstrumentation.java#L66-L67
And the netty tests suppress the normal test verification of span attributes for these requests:

```
  @Override
  boolean hasClientSpanAttributes(URI uri) {
    switch (uri.toString()) {
      case "http://localhost:61/": // unopened port
      case "http://www.google.com:81/": // dropped request
      case "https://192.0.2.1/": // non routable address
        return false
      default:
        return true
    }
  }
```

PR #3115, opened 2021-05-27
Add attributes to netty connection failure span

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3103
Adds net peer attributes to netty connection failure span. Http attributes are not added because they are not available during connect.


trask (approved):
nice 👍

PR #3117, opened 2021-05-28
Use gradle-nullaway-plugin for nullaway.

anuraaga:
This will allow me to followup by decoupling error prone and nullaway, allowing us to replace spotbugs with error prone.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ plugins {
   id "com.diffplug.spotless"
   id "com.github.spotbugs" apply false
   id "net.ltgt.errorprone" apply false
+  id "net.ltgt.nullaway" apply false
 }
 
 release {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -71,6 +72,31 @@ allprojects {
...
+
+    nullaway {
+      annotatedPackages.addAll("io.opentelemetry", "com.linecorp.armeria,com.google.common")
+    }
+
+    tasks.withType(JavaCompile).configureEach {
+      if (!name.toLowerCase().contains("test")) {
+        options.errorprone {
+          nullaway {
+            severity = net.ltgt.gradle.errorprone.CheckSeverity.ERROR
```
anuraaga, line 91: I had a similar problem before, but AFAICT, there's no way to access plugin classes from a Gradle "script plugin", which is what `apply from: *.gradle` type of stuff is. So I don't see a way to do this sort of configuration from `java.gradle` and just moved it here.

trask (approved):

PR #3120, opened 2021-05-28
Spring Integration library instrumentation

mateuszrzeszutek:
Spring-integration is based on spring-messaging and serves as a base for spring-cloud-stream (and probably many other spring libs). It's basically an implementation of the Enterprise Integration Patterns idea (like Camel).
The central interface in spring-messaging/integration is `MessageChannel` - it connects different components/pieces of logic/cloud applications (e.g. connects RabbitMQ listener to underlying impl agnostic spring `MessageHandler`). Fortunately Spring provides a `ChannelInterceptor` interface that allows us to inject some code into various phases of `MessageChannel` lifecycle.

Initially I wanted to implement just the context propagation without any tracing (spans would be created by e.g. RabbitMQ or Kafka instrumentations), but once I started working on that and dug into the spring code I realized that I'll have to create at least one span that'll hold a link to the context received from a remote message. It also became apparent to me that instrumenting just the `send()` method is enough and `receive()` interceptor callbacks won't propagate anything.
I think that this instrumentation is probably mostly correct (it preserves and propagates the context), but I haven't tested it with a real spring-cloud-stream scenario -- will do that in one of the next PRs.

This PR includes a new addition to the Instrumenter API: the `SpanLinkExtractor`.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -119,8 +122,12 @@ public Context start(Context parentContext, REQUEST request) {
       spanBuilder.setStartTimestamp(startTimeExtractor.extract(request));
     }
 
+    for (SpanLinkExtractor<? super REQUEST> extractor : spanLinkExtractors) {
+      spanBuilder.addLink(extractor.extract(parentContext, request));
```
iNikem, line 126: Can extractor NOT return any link? What will happen if no link is found?

iNikem: We should have tests for both cases

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/messaging/MessageHeadersSetter.java
```diff
@@ -0,0 +1,33 @@
...
+
+// Native headers logic inspired by
+// https://github.com/spring-cloud/spring-cloud-sleuth/blob/main/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessageHeaderPropagatorSetter.java
+enum MessageHeadersSetter implements TextMapSetter<MessageHeaderAccessor> {
+  INSTANCE;
+
+  @Override
+  public void set(MessageHeaderAccessor carrier, String key, String value) {
+    carrier.setHeader(key, value);
+    setNativeHeader(carrier, key, value);
```
iNikem, line 24: Why do we need to set both? At least add a comment, please.

iNikem (commented):
I don't understand why we don't do anything on `receive`? Who is going to "restore" the context that we propagated?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
mateuszrzeszutek: > Can extractor NOT return any link? What will happen if no link is found?

It can return `SpanContext.getInvalid()` - which will be ignored by the SDK. I'll add a comment explaining that.

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @Override
+  public void postSend(Message<?> message, MessageChannel messageChannel, boolean sent) {}
+
+  @Override
+  public void afterSendCompletion(
+      Message<?> message, MessageChannel messageChannel, boolean sent, Exception e) {
+    ContextAndScope contextAndScope =
+        message.getHeaders().get(CONTEXT_AND_SCOPE_KEY, ContextAndScope.class);
```
trask, line 61: is `afterSendCompletion` guaranteed to run in same thread as `preSend`?

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+    Scope scope = context.makeCurrent();
+    MessageHeaderAccessor messageHeaderAccessor = MessageHeaderAccessor.getMutableAccessor(message);
+    messageHeaderAccessor.setHeader(SCOPE_KEY, scope);
+    return createMessageWithHeaders(message, messageHeaderAccessor);
+  }
+
+  @Override
+  public void afterMessageHandled(
+      Message<?> message, MessageChannel channel, MessageHandler handler, Exception ex) {
+    Scope scope = message.getHeaders().get(SCOPE_KEY, Scope.class);
```
trask, line 100: same question about `beforeHandle` and `afterMessageHandled`

trask (approved):
👍

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
mateuszrzeszutek: Spring docs only say that 

> `afterSendCompletion()` and `afterReceiveCompletion()` are invoked after `send()` and `receive()` calls, regardless of any exception that is raised, which allow for resource cleanup.
 
t does not mention that they always run on the same thread, but this is in fact how it works: all existing implementations call `preSend`/`afterSendCompletion` directly before/after the internal `doSend` method.

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
mateuszrzeszutek: Spring Javadocs:

> Invoked inside the Runnable submitted to the Executor after calling the target MessageHandler regardless of the outcome (i.e. Exception raised or not) thus allowing for proper resource cleanup.

In addition to that, these two methods actually seem to be sort of deprecated in spring integration/cloud-stream, as they're only used once in one `MessageChannel` implementation from spring-messaging that is not used anywhere in the spring code.

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/MessageSpanLinkExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+    SpanContext spanFromMessage = delegate.extract(parentContext, messageWithChannel);
+    SpanContext parentSpan = Span.fromContext(parentContext).getSpanContext();
+    if (referencesSameSpan(spanFromMessage, parentSpan)) {
+      return SpanContext.getInvalid();
+    }
+    return spanFromMessage;
+  }
+
+  // SpanContext#equals() includes e.g. remote flag, which we don't really care about here
+  // we just want to avoid adding links to spans with the same id, flags don't matter at all
```
iNikem, line 31: Is it normal situation when `referencesSameSpan==true` but `spanFromMessage.equals(parentSpan) == false`? I am not sure that we should work around such situation...

mateuszrzeszutek:
> I don't understand why we don't do anything on `receive`? Who is going to "restore" the context that we propagated?

One reason why I didn't touch the receive interceptor methods is because they all execute before `receive()` returns, and thus cannot set context. `receive()` is also only used in `PollableChannel`s, never in `SubscribableChannel`s.

It's a bit unintuitive, but the way spring-integration works is `send()` instrumentation both extracts the context from the message and injects the new one (with the newly created span) into it - this way if this channel calls `MessageHandler`s directly on `send()` (like `DirectChannel` does) handlers will see the current context based on the one extracted from the message; if the channel transmits the message somewhere else (an external queue, like Kafka) there will always be another `MessageChannel` whose `send()` will be called by that "somewhere else" (a Kafka consumer, for example).

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
iNikem: Then maybe we don't have to pass context and scope via headers? And can rely on thread locals/fields?

instrumentation/spring/spring-integration-4.1/library/src/test/groovy/ComplexPropagationTest.groovy
```diff
@@ -0,0 +1,160 @@
...
+          name "application.sendChannel"
+          childOf span(0)
+          hasNoLinks()
+        }
+      }
+      trace(1, 2) {
+        span(0) {
+          name "application.receiveChannel"
+          hasNoParent()
+          hasLink sendChannelSpan
```
iNikem, line 84: Why is this a link and not a parent? E.g. in Kafka instrumentation we use context from the message as a parent. And I believe that current semantic conventions also dictate this.

iNikem:
Have you tested

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
mateuszrzeszutek: Hmm, thread local should work too. By fields do you mean `ContextStore`? It's the library instrumentation, so I can't use that here

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/MessageSpanLinkExtractor.java
mateuszrzeszutek: `SpanContext.equals()` also includes the remote flag, which is always set to true when you extract the context from the message. So in a completely "local" scenario, when you have one message channel sending a message to another you'd end up with a link to the parent span in the second channel's span.

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
iNikem: Is this interceptor reusable? Can we just use its fields?

mateuszrzeszutek: It's supposed to be used as a global one, one per application.

Anyway, I tried using thread local but it looks like it breaks one of the test scenarios: if you have multiple message channels (`DirectChannel`) chained together, the latter channels overwrite the thread local and parent spans are lost. I'd have to use something like `ThreadLocal<Map<MessageChannel, ContextAndStore>>` to make it work.

instrumentation/spring/spring-integration-4.1/library/src/test/groovy/ComplexPropagationTest.groovy
mateuszrzeszutek: @anuraaga @trask What do you think about it?

Short version: right now the parent of the `MessageChannel` is always the "local" current span, and whatever comes with the message is added as a link. When there's no local span only link is added.
Another option is to always use the span context extracted from the message (and fall back to `Context.current()` if there's none).
Which one does make more sense as the default behavior? Is there any other option?

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
iNikem: yeah.. map is too complex... forget about it

instrumentation/spring/spring-integration-4.1/library/src/test/groovy/ComplexPropagationTest.groovy
iNikem: We discussed this today during "special topics" meeting and it seems that using propagated context as parent makes more sense than using current or implicit context. When you use message listener (and in case of `SubscribableChannel` that is what you do, right), then your code will be called by the framework code. And you usually don't care about framework loops. So it still makes sense to have PRODUCER span as your parent. You have only "processing" part, not "receiving" one.

mateuszrzeszutek: 👍

mateuszrzeszutek: Done - thanks for the discussion!

iNikem (approved):

PR #3122, opened 2021-05-28
Migrate from spotbugs to errorprone

anuraaga:
Consistent with SDK, prints errors to console instead of file, and easier to SuppressWarnings


instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/TracedDelegatingConsumer.java
```diff
@@ -62,35 +62,18 @@ public void handleRecoverOk(String consumerTag) {
...
-    Scope scope = null;
-    try {
-      context = tracer().startDeliverySpan(queue, envelope, properties, body);
-      scope = context.makeCurrent();
+    Context context = tracer().startDeliverySpan(queue, envelope, properties, body);
+    Scope scope = context.makeCurrent();
 
-    } catch (Exception e) {
-      log.debug("Instrumentation error in tracing consumer", e);
+    try {
```
mateuszrzeszutek, line 68: Nit: we can use try-with-resources here

instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/TracedDelegatingConsumer.java
```diff
@@ -62,35 +62,18 @@ public void handleRecoverOk(String consumerTag) {
...
+      // Call delegate.
+      delegate.handleDelivery(consumerTag, envelope, properties, body);
+      tracer().end(context);
+    } catch (Throwable throwable) {
+      tracer().endExceptionally(context, throwable);
+      throw throwable;
     } finally {
-      // TODO this is very unusual code structure for this repo
-      // We have to review it
```
mateuszrzeszutek, line 75: 👍

mateuszrzeszutek (approved):

trask (approved):
> Consistent with SDK

❤️ 

And great clean-ups 👍

instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/TracedDelegatingConsumer.java
trask: done (so I can merge this before #3118)

BigGuy573 (commented):
Tremendous Gain!  👍

PR #3125, opened 2021-05-29
Update lettuce-5.0 to Instrumenter API

trask:


instrumentation/lettuce/lettuce-5.0/javaagent/src/test/groovy/LettuceAsyncClientTest.groovy
```diff
@@ -151,10 +149,8 @@ class LettuceAsyncClientTest extends AgentInstrumentationSpecification {
           errorEvent AbstractChannel.AnnotatedConnectException, String
           attributes {
             "$SemanticAttributes.NET_PEER_NAME.key" PEER_NAME
-            "$SemanticAttributes.NET_PEER_IP.key" PEER_IP
             "$SemanticAttributes.NET_PEER_PORT.key" incorrectPort
             "$SemanticAttributes.DB_SYSTEM.key" "redis"
-            "$SemanticAttributes.DB_STATEMENT.key" "CONNECT"
```
anuraaga, line 157: Can we hard code at least operation == CONNECT in the instrumentation?

anuraaga: Or perhaps we're fine with it being the span name, given there's no Redis command named connect. That makes sense actually

anuraaga (approved):

mateuszrzeszutek (approved):

PR #3126, opened 2021-05-29
Update cassandra-4.0 to Instrumenter API

trask:


instrumentation/cassandra/cassandra-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v4_0/CassandraNetAttributesExtractor.java
```diff
@@ -0,0 +1,37 @@
...
+  public @Nullable InetSocketAddress getAddress(
+      CassandraRequest request, @Nullable ExecutionInfo response) {
+    if (response == null) {
+      return null;
+    }
+    Node coordinator = response.getCoordinator();
+    if (coordinator == null) {
+      return null;
+    }
+    SocketAddress address = coordinator.getEndPoint().resolve();
```
anuraaga, line 34: Wasn't happy to see the word resolve but looks like it's free...

https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java#L39

Might be good to add a comment that most implementations of Endpoint should not incur a cost here

anuraaga (approved):

instrumentation/cassandra/cassandra-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/cassandra/v4_0/CassandraNetAttributesExtractor.java
trask: ya, I had the same reaction when I saw `resolve`, comment added

mateuszrzeszutek (approved):

PR #3130, opened 2021-05-29
Move logic to redisson request object

trask:
Based on @anuraaga's comment https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3064#discussion_r637676791, it generally makes sense to move logic from extractors into request objects when those request objects are our own objects.

### DIFF
--- a/instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonRequest.java
+++ b/instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonRequest.java
@@ -5,17 +5,91 @@
 
 package io.opentelemetry.javaagent.instrumentation.redisson;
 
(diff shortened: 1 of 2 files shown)


instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonRequest.java
```diff
@@ -5,17 +5,91 @@
...
+      }
+    }
+    return null;
+  }
+
+  public String getStatement() {
+    List<String> sanitizedStatements = sanitizeStatement();
+    switch (sanitizedStatements.size()) {
+      case 0:
+        return null;
```
trask, line 49: this used to return the text "Redis Command", but since `db.statement` is not required, I think makes sense to return `null` and not set it instead

anuraaga (approved):

mateuszrzeszutek (approved):

PR #3134, opened 2021-05-30
Fix sporadic failing spring webflux tests

trask:
This simplifies the webflux instrumentation, and seems to resolve the sporadic test failures, but a couple of tests no longer pass, so need to look into it more...


trask:
this was fixed by #3150, and observation about webflux handler span recorded in #3144

PR #3137, opened 2021-05-30
Use AtomicLongUpdater for gRPC message ID

anuraaga:
Should always use an updater for a non-static atomic to reduce number of allocations.


anuraaga:
For reference http://normanmaurer.me/blog/2013/10/28/Lesser-known-concurrent-classes-Part-1/

trask (approved):

PR #3145, opened 2021-05-31
Sync yaml files

trask:
`pr.yaml`, `ci.yaml`, `nightly.yaml`, `nightly-no-cache.yaml`, `release-build.yml`, and `patch-release-build.yml` now all run:

* test
* testLatestDep (except for `pr.yaml` due to it causing contributors headaches sometimes when a new breaking library version is released)
* smoke-tests
* example-distro


.github/workflows/pr.yaml
```diff
@@ -77,6 +78,10 @@ jobs:
           path: /tmp/deadlock-detector-*
           if-no-files-found: ignore
 
+  # testLatestDep is intentionally not included in the PR workflow
+  # because it can break randomly when a new library version is released
+  # which can be confusing for contributors
```
trask, line 83: I'm wondering how this is different from running muzzle in the PR workflow..

.github/workflows/nightly-no-cache.yaml
```diff
@@ -50,6 +50,22 @@ jobs:
       - name: Test
         run: ./gradlew test -PtestJavaVersion=${{ matrix.java }} --stacktrace -x :smoke-tests:test -Porg.gradle.java.installations.paths=${{ steps.setup-test-java.outputs.path }} -Porg.gradle.java.installations.auto-download=false --no-build-cache
 
+  testLatestDep:
```
iNikem, line 53: I did not include this job into this workflow on purpose. Assuming that latestDep test does not use cache anyway.

.github/workflows/patch-release-build.yml
```diff
@@ -23,39 +23,155 @@ jobs:
...
+
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 8, 11, 15 ]
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          ref: ${{ needs.prepare-release-branch.outputs.release-branch-name }}
```
iNikem, line 57: If you reference `needs` here, shouldn't this job actually have a `needs` section?

.github/workflows/pr.yaml
mateuszrzeszutek: Muzzle only runs for the instrumentation code; latestDepTest often detects problems in the test code (or slight changes in emitted telemetry)

trask: 👍 updated comment to reflect

.github/workflows/nightly-no-cache.yaml
trask: > Assuming that latestDep test does not use cache anyway.

I'll test that out 👍

.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
+
+      - uses: gradle/wrapper-validation-action@v1.0.4
         with:
           fetch-depth: 0
```
mateuszrzeszutek, line 19: Isn't `fetch-depth` a checkout (git clone) option?

trask: 😱 thanks!

mateuszrzeszutek (approved):

.github/workflows/nightly-no-cache.yaml
trask: It looks like `testLatestDeps` does use the cache?

```
$ ./gradlew :instrumentation:spring:spring-webflux-5.0:javaagent:test -PtestLatestDeps=true

...

BUILD SUCCESSFUL in 1m 19s
68 actionable tasks: 5 executed, 63 up-to-date
S3 cache 300ms wasted on misses, reads: 3, elapsed: 300ms

$ ./gradlew :instrumentation:spring:spring-webflux-5.0:javaagent:test -PtestLatestDeps=true

...

BUILD SUCCESSFUL in 6s
68 actionable tasks: 68 up-to-date
```

iNikem: Interesting... I wonder if Gradle invalidates cache if new version is detected... But anyway it does make sense to include it in this workflow 👍

iNikem (approved):

PR #3146, opened 2021-05-31
Migrate GWT to Instrumenter API

anuraaga:
This started as me wanting to remove a usage of `SpanNames`, but realized this can just be an RPC instrumentation.


instrumentation/gwt-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/gwt/GwtSingletons.java
```diff
@@ -0,0 +1,29 @@
+package io.opentelemetry.javaagent.instrumentation.gwt;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.api.instrumenter.rpc.RpcAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.rpc.RpcSpanNameExtractor;
+import java.lang.reflect.Method;
+
+public final class GwtSingletons {
```
anuraaga, line 10: I think I've used every pattern possible for the name of this sort of class. I wonder if we should just converge on `*Singletons` so it can fit instrumenters, but also any other singletons that come our way.

trask (commented):
It looks like the [RPC semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#common-remote-procedure-call-conventions) specifically target CLIENT/SERVER spans

though I can definitely see a case for RPC conventions applied to RPC systems designed on top of servlets

or maybe that's what a "handler span" is? 😂

anuraaga:
I guess there's a difference between REST and RPC - I think we're only supposed to have one span here since this seems to be similar to gRPC. So I guess we should be updating any HTTP span with the RPC span name and attributes. A mode the instrumenter API doesn't yet support, and it ties into the general nested span stuff as usual :)

instrumentation/gwt-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/gwt/GwtRpcInstrumentation.java
```diff
@@ -52,23 +53,23 @@ public void transform(TypeTransformer transformer) {
...
-        @Advice.Thrown Throwable throwable,
+        @Advice.Argument(1) Method method,
         @Advice.Local("otelContext") Context context,
-        @Advice.Local("otelScope") Scope scope) {
+        @Advice.Local("otelScope") Scope scope,
+        @Advice.Thrown Throwable throwable) {
       scope.close();
 
-      tracer().endSpan(context, throwable);
+      INSTRUMENTER.end(context, method, null, throwable);
```
mateuszrzeszutek, line 72: Do we want to use an `instrumenter()` static method here?

mateuszrzeszutek (approved):

trask (approved):
can you add a TODO somewhere about the GWT RPC span currently being an INTERNAL span?

PR #3150, opened 2021-05-31
Fix flaky spring webflux tests

laurit:
Hopefully fixes flaky spring webflux tests. As far as I understand the problem is that reactive flow can get canceled (cancellation happens because request is already considered to be completed) before `SpanFinishingSubscriber` is run so that span remains unfinished. This pr adds handling of cancellation.


trask (approved):
❤️❤️❤️

PR #3153, opened 2021-05-31
End span on cancellation of subscription to reactive publishers

HaloFour:
Updates the AsyncSpanEndStrategy for Reactor, RxJava 2 and RxJava 3 so that the Span is ended if the subscription to the publisher is canceled by the consumer.  Optionally a semantic attribute will also be added to the Span to indicate that the subscription was canceled.

### DIFF
--- a/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
+++ b/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
@@ -5,18 +5,34 @@
 
 package io.opentelemetry.instrumentation.reactor;
 
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
 import io.opentelemetry.instrumentation.api.tracer.async.AsyncSpanEndStrategy;
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-public enum ReactorAsyncSpanEndStrategy implements AsyncSpanEndStrategy {
-  INSTANCE;
+public final class ReactorAsyncSpanEndStrategy implements AsyncSpanEndStrategy {
+  private static final AttributeKey<Boolean> CANCELED_ATTRIBUTE_KEY =
+      AttributeKey.booleanKey("reactor.canceled");
+
+  public static ReactorAsyncSpanEndStrategy create() {
+    return newBuilder().build();
+  }
+
+  public static ReactorAsyncSpanEndStrategyBuilder newBuilder() {
+    return new ReactorAsyncSpanEndStrategyBuilder();
+  }
+
+  private final boolean captureExperimentalSpanAttributes;
+
+  ReactorAsyncSpanEndStrategy(boolean captureExperimentalSpanAttributes) {
+    this.captureExperimentalSpanAttributes = captureExperimentalSpanAttributes;
+  }
 
   @Override
   public boolean supports(Class<?> returnType) {
@@ -46,7 +62,7 @@ public enum ReactorAsyncSpanEndStrategy implements AsyncSpanEndStrategy {
    * OnError notifications are received. Multiple notifications can happen anytime multiple
    * subscribers subscribe to the same publisher.
    */
-  private static final class EndOnFirstNotificationConsumer extends AtomicBoolean
+  private final class EndOnFirstNotificationConsumer extends AtomicBoolean
       implements Runnable, Consumer<Throwable> {
 
     private final BaseTracer tracer;
@@ -63,7 +79,12 @@ public enum ReactorAsyncSpanEndStrategy implements AsyncSpanEndStrategy {
     }
 
     public void onCancel() {
-      accept(new CancellationException());
+      if (compareAndSet(false, true)) {
+        if (captureExperimentalSpanAttributes) {
+          Span.fromContext(context).setAttribute(CANCELED_ATTRIBUTE_KEY, true);
+        }
+        tracer.end(context);
+      }
     }
 
     @Override
--- a/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
+++ b/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
@@ -8,6 +8,7 @@ package io.opentelemetry.instrumentation.reactor;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
 import io.opentelemetry.instrumentation.api.tracer.async.AsyncSpanEndStrategy;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import org.reactivestreams.Publisher;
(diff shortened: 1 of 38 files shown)


instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
```diff
@@ -57,6 +62,10 @@ public EndOnFirstNotificationConsumer(BaseTracer tracer, Context context) {
       accept(null);
     }
 
+    public void onCancel() {
+      accept(new CancellationException());
```
trask, line 66: The grpc instrumentation adds a "canceled" attribute instead, which seems like a good compromise between failure and success:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f755b3f52c5873da01a715bbc805da33e6d0d5d6/instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingServerInterceptor.java#L130-L141

Unfortunately there isn't a semantic attribute for such a canceled attribute, so we'd need to go with something like:

* `reactor.canceled`
* `rxjava.canceled`

And unfortunately we hide non-semantic attributes behind "experimental" flags (search for `captureExperimentalSpanAttributes` and `CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES`), which would probably mean not using singletons for the strategies anymore so that flag can be injected.

trask (commented):
thanks, and great tests as always!

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
HaloFour: That approach makes sense and I can change the strategies to adopt that pattern.  It does force a breaking change but it's probably better (and more future-proof) to adopt the same create/builder strategy anyway.  I'll amend the PR.

HaloFour: Should we do the same with the Java 8 and Guava strategies?  They do propagate cancellation through completing exceptionally but that is something we can detect and end the span successfully with a similar optional semantic attribute.  It'd be an excuse to change the singleton implementations of those strategies to also follow the create/builder pattern.  I could also submit that as a separate PR if we want to keep this one from getting too complicated.

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -44,13 +45,20 @@
...
   public static void registerOnEachOperator() {
     Hooks.onEachOperator(TracingSubscriber.class.getName(), tracingLift());
-    AsyncSpanEndStrategies.getInstance().registerStrategy(ReactorAsyncSpanEndStrategy.INSTANCE);
+    AsyncSpanEndStrategies.getInstance()
+        .registerStrategy(
+            ReactorAsyncSpanEndStrategy.newBuilder()
+                .setCaptureExperimentalSpanAttributes(
+                    Config.get()
+                        .getBooleanProperty(
+                            "otel.instrumentation.reactor.experimental-span-attributes", false))
```
mateuszrzeszutek, line 54: Ideally we should only use `Config` in the javaagent... How about adding the builder & flag to `TracingOperator` and making `registerOnEachOperator` an instance method? You won't need a builder for `ReactorAsyncSpanEndStrategy` anymore, you'll be able to use a package-private constructor.

Then in the javaagent code the correct property value will be passed:
```java
TracingOperator.newBuilder()
  .setCaptureExperimentalSpanAttributes(Config.....)
  .build()
  .registerOnEachOperator();
```

HaloFour: Got it, will refactor.

mateuszrzeszutek: Thanks! 🙏

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
```diff
@@ -14,8 +16,23 @@
...
+  private static final AttributeKey<Boolean> CANCELED_ATTRIBUTE_KEY =
+      AttributeKey.booleanKey("reactor.canceled");
+
+  public static ReactorAsyncSpanEndStrategy create() {
+    return newBuilder().build();
+  }
+
+  public static ReactorAsyncSpanEndStrategyBuilder newBuilder() {
+    return new ReactorAsyncSpanEndStrategyBuilder();
+  }
```
mateuszrzeszutek, line 29: Hmm, do we need the builder for this class? Is it supposed to be possible to use it without `TracingOperator`?
If we can we can treat it as an internal part of the library instrumentation then we can make it package-private and remove the builder and use a simple `ReactorAsyncSpanEndStrategy(boolean)` constructor instead.

instrumentation/rxjava/rxjava-3.0/javaagent/src/test/groovy/RxJava3WithSpanInstrumentationTest.groovy
```diff
@@ -619,6 +737,40 @@ class RxJava3WithSpanInstrumentationTest extends AgentInstrumentationSpecificati
     }
   }
 
+  def "should capture span for eventually errored Flowable"() {
```
mateuszrzeszutek, line 740:
```suggestion
  def "should capture span for eventually canceled Flowable"() {
```

instrumentation/rxjava/rxjava-2.0/javaagent/src/test/groovy/RxJava2WithSpanInstrumentationTest.groovy
```diff
@@ -756,6 +913,42 @@ class RxJava2WithSpanInstrumentationTest extends AgentInstrumentationSpecificati
     }
   }
 
+  def "should capture span for eventually errored ParallelFlowable"() {
```
mateuszrzeszutek, line 916:
```suggestion
  def "should capture span for eventually canceled ParallelFlowable"() {
```

?

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
mateuszrzeszutek: > Should we do the same with the Java 8 and Guava strategies?

If it's possible -- sure, why not.

>  I could also submit that as a separate PR if we want to keep this one from getting too complicated.

💯 👍

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
HaloFour: Hm, I had not considered them as being internal to the instrumentation.  Would it be problematic to allow them to be used directly?  They were already public even though they were singletons.

As of right now I am using the ReactorAsyncSpanEndStrategy without the TracingOperator in my Spring WebFlux projects.  I've been slowly migrating away from our own instrumentation but I've not yet considered replacing our own Reactor operators, partially because they do more than propagate the OTel Context.  I've combined several operations into a single operator because every individual operator effectively doubles the stack trace as they are added for every reactive operation.

instrumentation/rxjava/rxjava-3.0/javaagent/src/test/groovy/RxJava3WithSpanInstrumentationTest.groovy
HaloFour: ugh, copy/paste fail.  I'm surprised it doesn't error when two methods have identical names like that.

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncSpanEndStrategy.java
mateuszrzeszutek: > Hm, I had not considered them as being internal to the instrumentation. Would it be problematic to allow them to be used directly? They were already public even though they were singletons.

Hmm, it would be a bit better to expose less API surface, and they seemed like they could be internal. I wouldn't say that it's a hard requirement for them to be internal; I feel like they should be, but not have to. If you're using them right now then it's probably fine to keep them public for a while.

mateuszrzeszutek (approved):

trask (approved):
thanks!

PR #3155, opened 2021-06-01
Enable all errorprone checks

anuraaga:
Consistent with SDK


iNikem:
So now even here we have to have a perfect code before we can run any single code-experiment? I find this annoying in SDK repo, to be honest.

anuraaga:
@iNikem Are you mostly concerned with tests / benchmarks? I think we can aggressively disable more checks on them as we find annoying ones, and don't need to disable all of them. For anything merged that's in main we should do our diligence.

iNikem:
I am concerned with everything :D I think what I want is a simple way to run any code/task without errorprone. E.g. can we introduce project property to disable errorprone checks completely? Then I can give it on command line and be happy :)

anuraaga:
Sure - added a flag `disableErrorProne=true`

iNikem (approved):

PR #3159, opened 2021-06-01
Refactor HttpClientTest

mateuszrzeszutek:
* Remove unused `extraClientSpans()` method
* Merge `extraAttributes()` and `hasClientSpanHttpAttributes(uri)` into `httpAttributes(uri)`: this will allow to choose extracted attributes per URI (for example, OkHttp won't set flavor on connection exception once it's rewritten to `Instrumenter` because it's only available in the response)


laurit (approved):

trask (approved):
nice

anuraaga (approved):

PR #3160, opened 2021-06-01
Moved jdbc-datasource instrumentation to Instrumenter API

mateuszrzeszutek:
* Changed span type from CLIENT to INTERNAL (we're creating a span for opening a connection, which isn't the same as sending a request; besides, sometimes opening a SQL connections sends a validation query - we'd end up with nested CLIENT spans this way)
* Implemented code semantic conventions (`code.namespace`, `code.function`)


instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/datasource/DataSourceInstrumenters.java
```diff
@@ -0,0 +1,36 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.api.instrumenter.SpanNameExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.code.CodeAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.code.CodeSpanNameExtractor;
+import javax.sql.DataSource;
+
+public final class DataSourceInstrumenters {
+  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.javaagent.jdbc";
+
+  private static final Instrumenter<DataSource, Void> INSTRUMENTER;
```
trask, line 18: would `DataSource` -> `Method` make sense?

trask (approved):

anuraaga (approved):

instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/datasource/DataSourceInstrumenters.java
mateuszrzeszutek: I tried that first - but I got different class names in tests (`method.getDeclaringClass()` vs `dataSource.getClass()`). I thought that it's probably a bit better to use the actual data source type in the attribute value, not the abstract/base class type.

trask: makes sense 👍

PR #3166, opened 2021-06-02
Run gRPC tests with Netty instrumentation

anuraaga:
Since gRPC uses Netty we should run tests with Netty instrumentation on. I'm actually not sure why it works, Netty should be making server spans I thought. Seems ok for some reason though (something about gRPC's Netty implementation that doesn't trigger our matchers? 🤷 )


anuraaga:
Well thinking again, I don't know if our Netty instrumentation supports HTTP/2. If not, then that's why it doesn't interfere here - a bit lucky :)

iNikem (approved):

mateuszrzeszutek (approved):

PR #3171, opened 2021-06-02
Inject ObjectProvider into BeanPostProcessors to avoid premature initialization of OpenTelemetry

HaloFour:
When a BeanPostProcessor takes a dependency on a bean directly it forces that bean and it's dependencies to be immediately instantiated.  This can create a situation where those dependencies might not be post-processed by other BeanPostProcessors that have not yet been initialized.  By injecting `ObjectProvider<T>` into the BeanPostProcessors the initialization of those dependencies can be deferred avoiding this issue.


HaloFour:
I had been bitten by this as I have a custom `SpanExporter` implementation that uses `MeterRegistry` to report span metrics to prometheus.  These BeanPostProcessors were causing `OpenTelemetry` and my `SpanExporter` to be instantiated which was causing `MeterRegistry` to also be initialized before `MeterRegistryPostProcessor` was available.  That was causing my `MeterRegistryCustomizer` to not be processed resulting in custom tags not being applied to any metrics.

I've not found any written documentation that states that BeanPostProcessors should always use `ObjectProvider<T>` or `@Lazy` but it seems like the safest route to take.

trask (approved):

mateuszrzeszutek (approved):

PR #3173, opened 2021-06-02
Split akka actor instumentation

laurit:
Split akka actor instrumentation into actor instrumentation and fork join instrumentation. Fork join code is gone from the latest akka and this makes muzzle fail the whole instrumentation because of mismatched references.
This pr also enables akka instrumentation by default. If anybody knows the reason why it was disabled then please speak up.


instrumentation/akka-actor-2.5/javaagent/akka-actor-2.5-javaagent.gradle
```diff
@@ -7,16 +7,28 @@ muzzle {
...
+  pass {
+    group = 'com.typesafe.akka'
+    module = 'akka-actor_2.12'
+    versions = "[2.5.0,)"
+  }
+  pass {
+    group = 'com.typesafe.akka'
+    module = 'akka-actor_2.13'
+    versions = "(,)"
+  }
```
trask, line 19: 👍

instrumentation/akka-actor-2.5/javaagent/akka-actor-2.5-javaagent.gradle
```diff
@@ -7,16 +7,28 @@ muzzle {
...
   testImplementation "com.typesafe.akka:akka-actor_2.11:2.5.0"
+
+  latestDepTestLibrary "com.typesafe.akka:akka-actor_2.13:+"
 }
 
-tasks.withType(Test).configureEach {
-  jvmArgs '-Dotel.instrumentation.akka-actor.enabled=true'
+if (findProperty('testLatestDeps')) {
+  configurations {
+    // akka artifact name is different for regular and latest tests
```
trask, line 31: 👍

trask (approved):
> Split akka actor instrumentation into actor instrumentation and fork join instrumentation. Fork join code is gone from the latest akka and this makes muzzle fail the whole instrumentation because of mismatched references.

👍 and we can move it over to contrib when we move over old instrumentation versions

> This pr also enables akka instrumentation by default. If anybody knows the reason why it was disabled then please speak up.

IIRC (which I may not) we disabled it by default in #911 because it was causing context leaks in the akka http tests. I'm good with enabling it.

mateuszrzeszutek (approved):

PR #3179, opened 2021-06-03
Remove no longer needed NameMatchers class

mateuszrzeszutek:
ByteBuddy's `ElementMatchers` already contains a `namedOneOf` method.


trask (approved):
👍

PR #3181, opened 2021-06-03
Update to Errorprone 2.7

anuraaga:


instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/ChannelTraceContext.java
```diff
@@ -74,4 +75,16 @@ public boolean equals(Object obj) {
   public int hashCode() {
     return Objects.hash(connectionContext, clientParentContext, context);
   }
+
+  @Override
+  public String toString() {
```
mateuszrzeszutek, line 80: Is that `toString()` needed?

instrumentation/vertx-reactive-3.5/javaagent/src/test/java/VertxReactiveWebServer.java
```diff
@@ -126,7 +126,6 @@ private static void handleListProducts(RoutingContext routingContext) {
           .rxQuery("SELECT id" + queryInfix + ", name, price, weight FROM products")
           .flatMap(
               result -> {
-                Thread.dumpStack();
```
mateuszrzeszutek, line 129: 😱

mateuszrzeszutek (approved):

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/ChannelTraceContext.java
anuraaga: There's a check that toString is added if equals/hashcode are. I can disable it if it seems to much but figure having a toString is generally better than not.

mateuszrzeszutek: Hmm, it's not really a big deal or anything, but those methods will probably be never called - they're just dead code. If we need a `toString()` method we usually add it.

trask: I was going to say +1 for disabling this rule, then went to read why the rule exists, and didn't realize `Object.toString()` was affected by overriding `hashCode()` 😱, I guess I thought it was always based on `System.identityHashCode()`, so now at least the rule makes sense to me. I think I'm probably still _slightly_ +1 for disabling the rule for the same reasons as @mateuszrzeszutek mentioned, but also good with keeping it if you think it's useful.

trask (approved):
nice

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/ChannelTraceContext.java
anuraaga: IntelliJ calls toString when debugging which is the only reason really to have toString on types that won't be called by code. Not that debugging works well for us :) But it's why I slightly lean towrads keeping it

PR #3182, opened 2021-06-03
Rename ComponentInstaller to AgentListener and add #order() method

mateuszrzeszutek:
* I also moved `AgentExtension` from the `...extension.spi` to `...extension` package -- we already have SPIs in other packages (`...instrumentation`) and there's also a note about it in the class Javadoc
* Added `Ordered` interface, now implemented by several SPIs


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.javaagent.instrumentation.api.InstrumentationContext;
+import java.lang.instrument.Instrumentation;
+import java.util.Map;
+import net.bytebuddy.agent.builder.AgentBuilder;
+
+/**
+ * {@link AgentListener} can be used to execute code before/after Java agent installation, for
+ * example to install any implementation providers that are used by * instrumentations. For
```
trask, line 16:
```suggestion
 * example to install any implementation providers that are used by instrumentations. For
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeServiceLoader.java
```diff
@@ -3,12 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.instrumentation.api;
+package io.opentelemetry.javaagent.tooling;
```
trask, line 6: 👍

trask (approved):
I wasn't sure about using "safe" service loader for everything, since the safety is only needed currently for instrumentation modules that are compiled with newer Java versions, but I could imagine other SPI implementations also only applying to newer Java versions, so seems good 👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeServiceLoader.java
```diff
@@ -40,5 +43,15 @@
...
   }
 
+  /**
+   * Same as {@link #load(Class)}, but also orders the returned implementations by comparing their
+   * {@link Ordered#order()}.
+   */
+  public static <T extends Ordered> List<T> loadOrdered(Class<T> serviceClass) {
+    return load(serviceClass).stream()
+        .sorted(Comparator.comparingInt(Ordered::order))
+        .collect(Collectors.toList());
```
iNikem, line 53: Why not just use `List.sort` method? Without any streams.

mateuszrzeszutek: Good point, done

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/SafeServiceLoader.java
```diff
@@ -40,5 +42,15 @@
...
   }
 
+  /**
+   * Same as {@link #load(Class)}, but also orders the returned implementations by comparing their
+   * {@link Ordered#order()}.
+   */
+  public static <T extends Ordered> List<T> loadOrdered(Class<T> serviceClass) {
+    List<T> impls = load(serviceClass);
+    impls.sort(Comparator.comparing(Ordered::order));
+    return impls;
```
iNikem, line 52:
```suggestion
    List<T> result = load(serviceClass);
    result.sort(Comparator.comparing(Ordered::order));
    return result;
```

iNikem (approved):

PR #3188, opened 2021-06-03
Adds binding of attributes in methods annotated with @WithSpan

HaloFour:
Implementation of binding method parameters to Span attributes with integration with Byte Buddy and Spring aspects.

Todo:

- [x] Memoize `AttributeBindings` instances by `Method` so reflection only has to happen once per instrumented method and attribute keys can be reused.
- [x] Update instrumentation to use separate advice only if the method has at least one parameter annotated with `@SpanAttribute` to avoid the overhead of copying/boxing the arguments.
    * Applies to ByteBuddy instrumentation to avoid copying/boxing of arguments into `Object[]` unnecessarily.
    * I don't think this will be possible until `@SpanAttribute` is [merged into the annotation extensions module](https://github.com/open-telemetry/opentelemetry-java/pull/3168).
- [x] Add instrumentation for Spring advice as well.
- [x] Expand the mapping of parameter types to attribute types
    - [x] Scalar types: `String`, `long`, `Long`, `double`, `Double`, `boolean`, `Boolean`, `int`, `Integer`, `float`, `Float`
    - [x] Simple wrappable arrays: `String[]`, `Long[]`, `Double[]`, `Boolean[]`
        * Wraps array using `Arrays#asList`
    - [x] Simple `List<T>`s: `List<String>,` `List<Long>`, `List<Double>`, `List<Boolean>`
    - [x] Convertible arrays: `int[]`, `long[]`, `float[]`, `double[]`, `boolean[]`, `Integer[]`, `Float[]`, `T[]`
        * Wraps array using anonymous subclass of `AbstractList<T>` which converts each element on demand to avoid copying or additional work
    - [x] Convertible `List<T>`s:  `List<Integer>`, `List<Float>`
    - [x] ~`Set<? extends Enum>`~
        * Figure an `enum` is unlikely to have a lot of possible values
    - [x] Subtypes of `List<T>`?
        * Java reflection over generics is maddening, wrote reflection helper class `ParameterizedClass` to facilitate identifying subclasses of `List<T>` and to map the actual type argument, value vs. complexity is certainly up for debate
    - [x] Default (calls `toString()`)
- [x] More tests


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ParameterizedClass.java
```diff
@@ -0,0 +1,197 @@
...
+import java.lang.reflect.TypeVariable;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Helper class for reflecting over the type hierarchy of parameterized types where the generic type
+ * arguments are reified and/or mapped to the type parameters of the implemented superclasses or
+ * interfaces.
+ */
+public final class ParameterizedClass implements ParameterizedType {
```
mateuszrzeszutek, line 19: Is it necessary to have this class implement `ParameterizedType`? I think there's a lot of logic around owner types that's only used in the `getTypeName()` - which is only used in tests.

mateuszrzeszutek (commented):
Sorry for being late, it took me a while to review this PR - I can't really say that I understand everything what happens in `ParameterizedClass`, but I think I get the general idea how it's supposed to work. WDYT about adding some more comments to that class? It could benefit from some additional explanations.

Also, how about moving all those binding-related classes to a package `annotationsupport`? Together with the async operation code (which could be moved to the same package) it'd be a good candidate for extraction to a separate module (`instrumentation-annotation-support` maybe?(

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanInstrumentation.java
```diff
@@ -56,9 +65,26 @@
...
+        tracedMethods, WithSpanInstrumentation.class.getName() + "$WithSpanAttributesAdvice");
+
+    // TODO: To avoid copying/boxing parameters only use WithSpanAttributesAdvice
+    //      if method has any parameters annotated with @SpanAttribute
+    /*
+    transformer.applyAdviceToMethod(tracedMethodsWithoutParameters,
         WithSpanInstrumentation.class.getName() + "$WithSpanAdvice");
+
+    transformer.applyAdviceToMethod(tracedMethodsWithParameters,
+        WithSpanInstrumentation.class.getName() + "$WithSpanAttributesAdvice");
```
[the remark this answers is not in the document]
HaloFour: I couldn't get it to work with the `SpanAttribute` annotation as a part of the `javaagent` module.  Now that `SpanAttribute` was merged into the annotated extension module I can revisit this.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/AttributeBindings.java
```diff
@@ -0,0 +1,25 @@
...
+
+/** Represents the bindings of method parameters to attributes of a traced method. */
+public interface AttributeBindings {
+
+  /**
+   * Indicates that the traced method has no parameters bound to attributes.
+   *
+   * @return {@code true} if the traced method has no bound parameters; otherwise {@code false}
+   */
+  boolean isEmpty();
```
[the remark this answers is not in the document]
HaloFour: It's an optimization to avoid copying the method parameters if none of the parameters are going to be traced as attributes.  This is primarily for the Spring instrumentation where the pointcut syntax isn't sophisticated enough (I believe) to apply based on annotations applied to any of the parameters and where `ProceedingJoinPoint#getArgs` results in allocating a new `Object[]` and copying the actual parameter values into it.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/AttributeBindingFactory.java
```diff
@@ -0,0 +1,366 @@
...
+      return (setter, arg) -> setter.setAttribute(key, ((Float) arg).doubleValue());
+    }
+
+    if (isArrayType(type)) {
+      return arrayBinding(name, type);
+    }
+
+    if (isEnumSetType(type)) {
+      return enumSetBinding(name);
+    }
```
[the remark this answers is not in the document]
HaloFour: Probably.  I was trying to handle these cases fairly narrowly to avoid situations where copying the parameter value into the attribute would be expensive.  If we want to support any `Collection<T>` as an attribute of one of the array types that would work, and we could optimize for the cases that don't require copying, e.g. `List<Long>` or `List<String>`.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ParameterizedClass.java
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  private static boolean equalsTypeVariable(TypeVariable<?> left, TypeVariable<?> right) {
+    if (left == right) {
+      return true;
+    } else if (left == null || right == null) {
+      return false;
+    }
+    return left.getGenericDeclaration().equals(right.getGenericDeclaration())
+        && left.getName().equals(right.getName());
```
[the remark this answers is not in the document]
HaloFour: Can do.  This was ported from Guava TypeToken API and this is the comment they use on the `TypeVariableKey` wrapper type they use to compare `TypeVariable<?>` identity:

```java
  /**
   * Wraps around {@code TypeVariable<?>} to ensure that any two type variables are equal as long as
   * they are declared by the same {@link java.lang.reflect.GenericDeclaration} and have the same
   * name, even if their bounds differ.
   *
   * <p>While resolving a type variable from a {@code var -> type} map, we don't care whether the
   * type variable's bound has been partially resolved. As long as the type variable "identity"
   * matches.
   *
   * <p>On the other hand, if for example we are resolving {@code List<A extends B>} to {@code
   * List<A extends String>}, we need to compare that {@code <A extends B>} is unequal to {@code <A
   * extends String>} in order to decide to use the transformed type instead of the original type.
   */
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/AttributeBindingFactory.java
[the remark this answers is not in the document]
mateuszrzeszutek: > If we want to support any `Collection<T>` as an attribute of one of the array types that would work, and we could optimize for the cases that don't require copying, e.g. `List<Long>` or `List<String>`.

Hmm, honestly I'm not sure whether we need to support those - I think I'd like to keep it as simple as possible until someone requests that explicitly. Do we need the enum set use case? Is that something you find useful in your code?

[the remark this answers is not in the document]
HaloFour: I nixed EnumSet and related types and kept the array attribute types to only `List<T>`.

HaloFour:
I've split the attribute binding to a separate module, `instrumentation-annotation-support`.  I agree that it makes sense to plan on moving the async stuff here as well to unclutter `instrumentation-api` in a followup PR.

mateuszrzeszutek:
> I've split the attribute binding to a separate module, `instrumentation-annotation-support`. I agree that it makes sense to plan on moving the async stuff here as well to unclutter `instrumentation-api` in a followup PR.

That sounds great, thanks!

mateuszrzeszutek (approved):
LGTM 👍 

Just one more question: we're going to replace those `SpanAttribute` annotation with the one that'll be introduced in the 1.4.0 SDK release, right?

HaloFour:
> Just one more question: we're going to replace those SpanAttribute annotation with the one that'll be introduced in the 1.4.0 SDK release, right?

Yes, I'm going to try to wrap that up today.  As the new annotation is in the 1.4.0-SNAPSHOT release is there a recommended way to pull that in temporarily?

mateuszrzeszutek:
> As the new annotation is in the 1.4.0-SNAPSHOT release is there a recommended way to pull that in temporarily?

Hmm, I thought of creating a new issue in this project to resolve that.

HaloFour:
With the version bump the tests for the otel annotations and spring boot instrumentations pass in isolation (at least locally).  Given a number of projects define their own set of repositories they are overriding the ones from `settings.gradle` so those projects are not resolving the snapshot version.

instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
```diff
@@ -5,6 +5,8 @@ plugins {
 val versions: Map<String, String> by project
 
 dependencies {
+  implementation(project(":instrumentation-annotation-support"))
```
mateuszrzeszutek, line 8: Oh, one more thing that I missed earlier: to ensure that lib ends up in the bootstrap classloader please add it as `implementation` dependency to the `javaagent-bootstrap` module and use `compileOnly` in javaagent instrumentations:

```suggestion
  compileOnly(project(":instrumentation-annotation-support"))
```

dependencyManagement/build.gradle.kts
```diff
@@ -33,7 +36,7 @@ val DEPENDENCY_BOMS = listOf(
   "com.fasterxml.jackson:jackson-bom:2.12.3",
   "com.google.guava:guava-bom:30.1.1-jre",
   "org.codehaus.groovy:groovy-bom:${groovyVersion}",
-  "io.opentelemetry:opentelemetry-bom:${otelVersion}",
+  "io.opentelemetry:opentelemetry-bom:${otelVersionSnapshot}",
```
mateuszrzeszutek, line 39: To avoid snapshot problems how about you copy the annotation to the new `instrumentation-annotation-support` module and add an issue for replacing it with the real one once SDK 1.4.0 is released? I'll add that issue to 1.4.0 milestone and make sure it's done before instrumentation repo releases.

trask (approved):
Just a few comments, looks great 👍

trask:
Probably a couple of mentions about these features in the docs will help people find them, e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/manual-instrumentation.md#creating-spans-around-methods-with-withspan

and

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/spring/spring-boot-autoconfigure/README.md#manual-instrumentation-support---withspan

trask:
hey @HaloFour! OTel SDK repo is targeting 1.4.0 release for Friday, and this repo is targeting 1.4.0 release for the following week. Do you have time this week or early next week to finish the last things here? If not, no worries, in that case we'll probably just temporarily remove/hide the `@SpanAttribute` annotation in OTel SDK repo until after 1.4.0 and plan to release everything with 1.5.0.

instrumentation-annotation-support/src/main/java/io/opentelemetry/instrumentation/annotation/support/AttributeBindingFactory.java
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  private static AttributeBinding defaultListBinding(String name) {
+    AttributeKey<List<String>> key = AttributeKey.stringArrayKey(name);
+    return mappedListBinding(Object.class, key, Object::toString);
+  }
+
+  private static AttributeBinding defaultBinding(String name) {
+    AttributeKey<String> key = AttributeKey.stringKey(name);
+    return (setter, arg) -> setter.setAttribute(key, arg.toString());
```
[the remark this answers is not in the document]
HaloFour: I have the consuming class perform the `null` check so that I could keep the `AttributeBinding` implementations/lambdas simple:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/56e72c5ac69700ce9149daeba48a718e8ae26eca/instrumentation-annotation-support/src/main/java/io/opentelemetry/instrumentation/annotation/support/BaseAttributeBinder.java#L100

I could move the `null` check to the `AttributeBinding` itself if that would be preferred.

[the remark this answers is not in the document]
trask: Oh I missed that, makes sense where you have it 👍

instrumentation/spring/spring-boot-autoconfigure/build.gradle.kts
```diff
@@ -8,6 +8,8 @@ group = "io.opentelemetry.instrumentation"
 val versions: Map<String, String> by project
 
 dependencies {
+  compileOnly(project(":instrumentation-annotation-support"))
```
mateuszrzeszutek, line 11: Hmm, should we use `implementation` here? It probably should be built in the autoconfigure module by default, we probably shouldn't force users to add this dependency by themselves.

donbeave: Maybe `api` instead of `implementation`?

mateuszrzeszutek: Hm, `api` would work too, but do we expect anybody to use those classes directly? If not then `implementation` would suffice

HaloFour:
The `check` task succeeds for the `:instrumentation:spring-boot-autoconfigure` and `:instrumentation:opentelemetry-annotations-1.0:javaagent` projects but it appears that there are some breaking changes with `opentelemetry-metrics`.

instrumentation-api-annotation-support/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+  id("otel.java-conventions")
+  id("otel.jacoco-conventions")
+  id("otel.publish-conventions")
+}
+
+group = "io.opentelemetry.instrumentation"
+
+dependencies {
+  implementation(project(":instrumentation-api"))
+  implementation(project(":instrumentation-api-caching"))
```
mateuszrzeszutek, line 11: You don't need to depend on the caching module anymore, it's now built in the `instrumentation-api` module

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

instrumentation-api-annotation-support/src/test/groovy/io/opentelemetry/instrumentation/api/annotation/support/AttributeBindingFactoryTest.groovy
```diff
@@ -0,0 +1,307 @@
...
+    String toString() {
+      return "TestClass{value = " + value + "}"
+    }
+  }
+
+  enum TestEnum { FOO, BAR, BAZ }
+
+  class TestFields {
+    EnumSet<TestEnum> enumSet
+    Set<TestEnum> setOfEnum
```
trask, line 297: are these still used?

trask (approved):

instrumentation-api-annotation-support/src/test/groovy/io/opentelemetry/instrumentation/api/annotation/support/AttributeBindingFactoryTest.groovy
HaloFour: Good catch!  Removed

trask:
Thanks @HaloFour 🥳

PR #3191, opened 2021-06-04
Fix netty muzzle

### LINKED ISSUE #3189: Netty muzzle configuration of vertx may be incorrect
I think I see randomly a Netty muzzle failure where it accepts failure but succeeds for VertX 2.0.0-CR1

trask:
Resolves #3189

I think it broke because vertx-core `2.0.0-CR1` depends on netty `4.0.0.CR5` and so muzzle was expected to fail, but the new dependency version management is causing the netty dependency to resolve to `4.0.0.Final` and so it was passing.

I don't think we need the vertx-core muzzle constraints in netty instrumentation modules anyways, so just removing them.

Not sure if we also want to (can?) disable dependency management for muzzle?

### DIFF
--- a/instrumentation/netty/netty-4.0/javaagent/netty-4.0-javaagent.gradle
+++ b/instrumentation/netty/netty-4.0/javaagent/netty-4.0-javaagent.gradle
@@ -19,12 +19,6 @@ muzzle {
     module = "netty"
     versions = "[,]"
   }
-  pass {
-    group = "io.vertx"
-    module = "vertx-core"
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
I agree that checking vertx doesn't seem so important. I thought I fixed applying management to muzzle (`configurations.configureEach` -> `configurations.each`) but maybe for some odd reason it fixed the direct `netty` but not the transitive one from `vertx`. Will check it out, there should be a way to help on the Muzzle side which is good since the plugin should theoretically be independent of such management in any build.

PR #3192, opened 2021-06-04
Add http client concurrency test with callback

laurit:
Add http client concurrency test with callback which is based on the existing test. And fix it for
- okhttp2.2
- java-http-client
- play-ws
Play-ws also required some changes in akka instrumentation.


trask (approved):
👍

iNikem (approved):

mateuszrzeszutek (approved):

Issue #2892, opened 2021-04-29
Release stable version of opentelemetry-instrumentation-api


trask:
This is a tracking issue for releasing stable version of the `opentelemetry-instrumentation-api` artifact.

Will post updates from time to time on this issue for anyone who wants to watch/follow it.

You may also be interested in following #2713 which is the current major work in progress towards this goal.

In the meantime, if you require a stable artifact, we recommend using `opentelemetry-api` directly, or shading `opentelemetry-instrumentation-api`.

trask:
I realized, another gate on a stable release of the opentelemetry-instrumentation-api is a stable release of the semantic conventions

trask:
Something else to consider is whether to wait for otel-wide spec of "instrumentation api" (discussion at https://github.com/open-telemetry/oteps/pull/165)

iNikem:
Or just have all attributes extractor in a separate unstable module

iNikem:
But anyway I totally support the idea of somehow making possible to declare stable a subset of a module

mateuszrzeszutek:
> Or just have all attributes extractor in a separate unstable module

We've already introduced `@UnstableApi` for metrics API, the semantic conventions situations is kind of similar to that.
But yeah, a separate module would be fine too.

ThomasVitale:
Is it correct to follow the backlog in https://github.com/open-telemetry/opentelemetry-java-instrumentation/projects/10#card-70538476 to know what tasks are left to do before opentelemetry-instrumentation-api can go GA?

iNikem:
@ThomasVitale yes

trask:
for those watching this issue, we are planning for the next release of the opentelemetry-instrumentation-api (1.17.0) to be a Release Candidate

PR #3199, opened 2021-06-06
Remove cassandra db.name attribute

trask:
from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#cassandra:

> `db.cassandra.keyspace` - The name of the keyspace being accessed. To be used instead of the generic db.name attribute.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/db/DbAttributesExtractor.java
```diff
@@ -25,7 +26,10 @@
   protected void onStart(AttributesBuilder attributes, REQUEST request) {
     set(attributes, SemanticAttributes.DB_SYSTEM, system(request));
     set(attributes, SemanticAttributes.DB_USER, user(request));
-    set(attributes, SemanticAttributes.DB_NAME, name(request));
+    AttributeKey<String> nameAttribute = dbNameAttribute();
```
anuraaga, line 29: Hmm this weirdness seems like a good reason to go to the spec and remove the keyspace attribute? Not filling in a generic attribute seems like a bug, non-cassandra aware tracing systems would still get benefit from it.

anuraaga: Definitely see where #2892 came from ;)

trask: 👍 can I block my own PR 😂

PR #3205, opened 2021-06-06
Fix reactor netty latest dep test

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3204
I have no idea why this started failing now because latest version was already released in may. The issue seems to be that when connecting fails there is a retry with different address (surprisingly this doesn't happen for all tests that expect connection to fail) which creates extra spans and changes expected exception. This pr changes name resolver used in test to only return one address for each host which seems to stop retry from happening.

### DIFF
--- a/instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
+++ b/instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
@@ -7,6 +7,11 @@ import static io.opentelemetry.instrumentation.test.utils.PortUtils.UNUSABLE_POR
 import static io.opentelemetry.instrumentation.test.utils.TraceUtils.basicSpan
 import static io.opentelemetry.instrumentation.test.utils.TraceUtils.runUnderTrace
 
+import io.netty.resolver.AddressResolver
+import io.netty.resolver.AddressResolverGroup
+import io.netty.resolver.InetNameResolver
+import io.netty.util.concurrent.EventExecutor
+import io.netty.util.concurrent.Promise
 import io.opentelemetry.api.common.AttributeKey
 import io.opentelemetry.api.trace.Span
 import io.opentelemetry.api.trace.SpanKind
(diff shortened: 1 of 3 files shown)


instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
```diff
@@ -179,4 +188,40 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
     assert expectedSpanContext.traceId == actualSpanContext.traceId
     assert expectedSpanContext.spanId == actualSpanContext.spanId
   }
+
+  // custom address resolver that returns at most one address for each host
```
trask, line 192:
```suggestion
  // custom address resolver that returns at most one address for each host
  // (because reactor-netty will retry on connection failure across multiple resolved addresses
  // which creates spans that are not expected by the http client tests)
```

trask (approved):

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
```diff
@@ -179,4 +188,42 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
...
+  // (because reactor-netty will retry on connection failure across multiple resolved addresses
+  // which creates spans that are not expected by the http client tests)
+  // adapted from io.netty.resolver.DefaultAddressResolverGroup
+  static class CustomNameResolverGroup extends AddressResolverGroup<InetSocketAddress> {
```
anuraaga, line 196: I'm assuming this is because of our test that hits `google.com`. I got burned by this a little while ago for not being able to pass tests in offline mode when developing in a waiting room so am planning on rewriting this test relatively soon. No big deal just FYI

anuraaga (approved):

iNikem (approved):

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/AbstractReactorNettyHttpClientTest.groovy
laurit: Actually for this test it was localhost that resolved both to ipv4 and ipv6 address.

mateuszrzeszutek (approved):

iNikem:
@laurit Please rebase to fix muzzle check

PR #3208, opened 2021-06-06
Remove last muzzleCheck methods

trask:
🎉


anuraaga (approved):

iNikem:
Seems like I missed something :D Why we don't need muzzle checks anymore?

mateuszrzeszutek (approved):

mateuszrzeszutek:
> Why we don't need muzzle checks anymore?

Probably because muzzle collects more references and we have `classLoaderMatcher()`

PR #3209, opened 2021-06-06
Pre- update apache-httpasyncclient-4.1 to Instrumenter API

trask:
This is pre-cursor to updating apache-httpasyncclient-4.1 to Instrumenter API.

I split out this PR separately to confirm that we want to go this route: postponing http client span creation until the request is available.


anuraaga (approved):
I think in basically all our other instrumentation we generate a span when a request is available so this seems OK. It only generates the head not the body so timing wise seems fine.

mateuszrzeszutek (approved):

PR #3214, opened 2021-06-07
Always include server span in HTTP client tests.

anuraaga:
Conditionally enabling server spans makes the tests harder to reason about I think and seems to provide little benefit in writing assertions.


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -341,34 +341,14 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
 
   //FIXME: add tests for POST with large/chunked data
 
-  def "trace request without propagation"() {
```
anuraaga, line 344: I think this only tested that `TestHttpServer` respects is-test-server, nothing about the client itself.

mateuszrzeszutek (approved):

instrumentation/http-url-connection/javaagent/src/test/groovy/HttpUrlConnectionTest.groovy
```diff
@@ -157,7 +157,6 @@ class HttpUrlConnectionTest extends HttpClientTest<HttpURLConnection> implements
     runUnderTrace("someTrace") {
       HttpURLConnection connection = url.openConnection()
       connection.useCaches = useCaches
-      connection.addRequestProperty("is-test-server", "false")
```
trask, line 160: let's delete this whole test ("trace request without propagation") and rename the test above "trace request with propagation" -> "trace request"

trask (approved):

PR #3215, opened 2021-06-07
Use Ratpack core API for test server.

anuraaga:
When trying to understand the Ratpack instrumentation, I found it easier to follow the tests using the core API instead of the groovy one. In particular, this doesn't use Guice which makes sure we don't have a requirement on Guice for the instrumentation to load. The syntax penalty seems minor enough.


instrumentation/ratpack-1.4/javaagent/src/test/groovy/server/RatpackAsyncHttpServerTest.groovy
```diff
@@ -97,23 +97,22 @@ class RatpackAsyncHttpServerTest extends RatpackHttpServerTest {
...
           }
         }
-        prefix("path/:id/param") {
-          all {
+        it.prefix("path/:id/param") {
+          it.all {context ->
             Promise.sync {
               PATH_PARAM
-            }.fork().then { endpoint ->
+            } then { endpoint ->
```
anuraaga, line 104: Also fixed typo of forking in this non-forking test

laurit (approved):

mateuszrzeszutek (approved):

PR #3217, opened 2021-06-07
Cancel in-progress PR builds when new commit is pushed.

anuraaga:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency


laurit (approved):

iNikem (approved):

trask (approved):
😎

PR #3079, opened 2021-05-25
Jetty-9 httpclient instrumentation, comments please

robododge:
This is my first PR attempt here,  I am focusing on  Jetty-9 instrumentation only.  I am pushing this PR only for help resolving my questions, then I can build the Jetty 8 client.

_Test status:_ **16 passed tests , 2 fails and 5 ignores**

1.  Need feedback on the 2 failing tests, they are for callback style http client.  
2. Jetty 9.0 client had much API drift going into 9.1 and some breaking changes, so I starting the Liberary version at 9.1.x.  How to compensate for the hole in Jetty 9.0?
3. Jetty 9.1 onward to Jetty 10 should be usable, but what features do the tests have to push through testing the various versions from 9.1 through 10?

I am using a different variation of this Jetty9 client in a live scenario, so I know it mostly works.   I appreciate your feedback.

resolves: #2837

EDIT: (6/2/21:  All tests pass now, callbacks wrapped properly)

PR #3225, opened 2021-06-08
Migrate HttpClientTest test server to Armeria

anuraaga:
My main motivation for this is to follow up with migrating the HTTPS test from an external request to internal to work offline, and also add HTTP/2 tests for libraries that support it. I was sad when I found that Netty instrumentation tests wouldn't work without introducing shading, but in the end I think it's good given we've found some issues with Jetty conflicts between instrumentation and testing in #3079 too, I guess it's best practice to shade the server we use for tests.

If this PR is merged, I will follow up with removing remaining usage of `TestHttpServer` and delete it.

The build won't pass though, reactor-netty fails mysteriously - as far as I can tell the connection context is not getting cleaned up so only the first test passes, and all other client spans get suppressed after that. Armeria presumably has triggered some connection reuse behavior that wasn't there before, my attempt to try to prevent it with `maxRequestsPerCnonection(1)` didn't help. But I guess the instrumentation has some bug - @iNikem could you help troubleshoot?


instrumentation/akka-http-10.0/javaagent/akka-http-10.0-javaagent.gradle
```diff
@@ -51,16 +45,9 @@ dependencies {
...
-  version101TestImplementation "com.typesafe.akka:akka-http_2.11:10.1.0"
-  version101TestImplementation "com.typesafe.akka:akka-stream_2.11:2.5.11"
-}
-
-test.dependsOn version101Test
-
-compileVersion101TestGroovy {
-  classpath = classpath.plus(files(compileVersion101TestScala.destinationDir))
-  dependsOn compileVersion101TestScala
+  // 10.0.x has some bugs making it incompatible with our test server.
```
anuraaga, line 48: I tried latest version of 10.0.x too but it still didn't sure. Not sure what the bug is but hopefully this is OK.

instrumentation/akka-http-10.0/javaagent/akka-http-10.0-javaagent.gradle
```diff
@@ -51,16 +45,9 @@ dependencies {
...
   library "com.typesafe.akka:akka-stream_2.11:2.4.14"
 
-  // There are some internal API changes in 10.1 that we would like to test separately for
-  version101TestImplementation "com.typesafe.akka:akka-http_2.11:10.1.0"
-  version101TestImplementation "com.typesafe.akka:akka-stream_2.11:2.5.11"
-}
-
-test.dependsOn version101Test
-
-compileVersion101TestGroovy {
```
anuraaga, line 61: I couldn't find any difference between this pattern and the standard latestTestDeps, we generally are fine with it vs testing the baseline of a new minor I think.

anuraaga: Ah to clarify, even if I didn't have to change to `testLibrary` for the mysterious 10.0.x failure, I would have still made this change

instrumentation/java-http-client/javaagent/src/test/groovy/JdkHttpClientTest.groovy
```diff
@@ -64,40 +62,4 @@ class JdkHttpClientTest extends HttpClientTest<HttpRequest> implements AgentTest
...
-          kind CLIENT
-          attributes {
-            "${SemanticAttributes.NET_TRANSPORT.key}" IP_TCP
-            "${SemanticAttributes.NET_PEER_NAME.key}" uri.host
-            "${SemanticAttributes.NET_PEER_IP.key}" { it == null || it == "127.0.0.1" }
-            // Optional
-            "${SemanticAttributes.NET_PEER_PORT.key}" uri.port > 0 ? uri.port : { it == null || it == 443 }
-            "${SemanticAttributes.HTTP_URL.key}" { it == "${uri}" || it == "${removeFragment(uri)}" }
-            "${SemanticAttributes.HTTP_METHOD.key}" method
-            "${SemanticAttributes.HTTP_FLAVOR.key}" "2.0"
```
anuraaga, line 92: This test was only here to override the base class for this assertion. I forced the client to HTTP1 instead and will add proper HTTP2 tests in the future.

iNikem (commented):
What reactor-netty? :) I see other failed tests: https://scans.gradle.com/s/niu4tnujvnxsu/tests

anuraaga:
@inikem Ah looks like a difference of localhost vs 127.0.0.1 caused by the execution environment difference, will tweak that. Still would help if you could try this branch's reactor-netty tests since those seem like real breakage

instrumentation/java-http-client/javaagent/src/test/groovy/JdkHttpClientTest.groovy
anuraaga: Sure filed #3227

iNikem:
> @iNikem Ah looks like a difference of localhost vs 127.0.0.1 caused by the execution environment difference, will tweak that. Still would help if you could try this branch's reactor-netty tests since those seem like real breakage (and reactor netty instrumentation is extremely hard to grok :(

Still don't see failing reactor-netty tests :D https://scans.gradle.com/s/iw4tlx733ld2g I see just netty failure

anuraaga:
@iNikem Hmm, test (8) is a testcontainers random flake, build looks like a flakiness of some sort though will see if there is extra flakiness because of this PR.

But test (15) shows it (as does running locally every time :P)

https://scans.gradle.com/s/5naapz3zi7hrg

I think it's because the client span remains in the channel through the tests and couldn't figure out where that's supposed to get cleared out.

instrumentation/reactor-netty/reactor-netty-0.9/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/reactornetty/v0_9/AbstractReactorNettyHttpClientTest.groovy
```diff
@@ -40,19 +42,27 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
...
+      .baseUrl(resolveAddress("").toString())
       ."${method.toLowerCase()}"()
       .uri(uri.toString())
   }
 
   @Override
   int sendRequest(HttpClient.ResponseReceiver request, String method, URI uri, Map<String, String> headers) {
-    return request.response().block().status().code()
+    return request.responseSingle {resp, content ->
+      // Make sure to consume content since that's when we close the span.
```
anuraaga, line 53: This monstrosity is the only way I found to have reactor-netty present us a status code after consuming the response.

instrumentation/akka-http-10.0/javaagent/akka-http-10.0-javaagent.gradle
anuraaga: Was able to fix issue by discarding entity

anuraaga:
Looks like issues were related to Armeria flushing headers before content, causing tests to also see headers before content was consumed. Fixed the tests to consume explicitly

iNikem (approved):

PR #3226, opened 2021-06-08
Support for multiple extension jars by scanning the given folder

iNikem:
Documentation will follow later when we are ready to declare this stable.


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
```diff
@@ -0,0 +1,88 @@
...
+    List<URL> result = new ArrayList<>();
+
+    if (locationName == null) {
+      return result;
+    }
+
+    File location = new File(locationName);
+    if (location.isFile()) {
+      addFileUrl(result, location);
+    } else {
```
mateuszrzeszutek, line 68: Nit: it'll probably be a bit safer to exclude sockets, pipes and other UNIX file types:

```suggestion
    } else if (file.isDirectory()) {
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -490,7 +490,7 @@ public void onComplete(
 
     private static boolean isIgnored(Class<?> c) {
       ClassLoader cl = c.getClassLoader();
-      if (cl instanceof AgentClassLoader || cl instanceof ExtensionClassLoader) {
+      if (cl instanceof AgentClassLoader) {
```
mateuszrzeszutek, line 493: What's the reason for this change? Do we want to redefine extension classes?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
```diff
@@ -0,0 +1,88 @@
...
+            System.getProperty(
+                "otel.javaagent.experimental.initializer.jar",
+                System.getenv("OTEL_JAVAAGENT_EXPERIMENTAL_INITIALIZER_JAR"))));
+    // TODO when logging is configured add warning about deprecated property
+
+    List<ClassLoader> delegates = new ArrayList<>(extension.size());
+    for (URL url : extension) {
+      delegates.add(getDelegate(parent, url));
+    }
+    return new MultipleParentClassLoader(parent, delegates);
```
mateuszrzeszutek, line 51: Nit: it probably could be:
```suggestion
    return new MultipleParentClassLoader(delegates);
```
Each of the `delegates` already calls `parent`

iNikem: I actually don't know about this. We use this `MultipleParentClassLoader` to load ALL our classes, including agent's ones (this `parent` here is `AgentClassLoader`). It somehow seems strange to me to "proxy" all those load requests through the first extension classloader.

O, while writing the sentence above, I have realised that we will situation when there is NO extensions. Then we still have to have a way to delegate to `AgentClassLoader`, thus I think we need this parent here.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
iNikem: The reason for this change is there is no more `ExtensionClassLoader`. I can change for `MultipleParentClassLoader`, but there may be different instance of it, not created by us.

And I don't know if there is any actual implication of this check.

examples/extension/build.gradle
```diff
@@ -30,6 +30,7 @@ ext {
 }
 
 repositories {
+  mavenLocal()
```
anuraaga, line 33: Remove or order at bottom?

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
iNikem: But do we actually need this?

examples/extension/build.gradle
iNikem: Will remove in later PR when we have suitable snapshots published.

mateuszrzeszutek (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
trask: It's an optimization to avoid retransforming classes that were loaded by the extension before we set up the bytebuddy transformers, @laurit's idea seems simple enough to be worth keeping.

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
iNikem: Done.

mateuszrzeszutek (approved):

PR #3234, opened 2021-06-10
Remove test for dropped request to google.com

anuraaga:
I want to remove requests to external domains. I think this test is the same as (unopened port) and can just be deleted


iNikem (approved):
I am not sure, but don't have any actual counter-arguments

anuraaga:
Just noticed that DD also seems to have come to a similar conclusion

https://github.com/DataDog/dd-trace-java/pull/2387

PR #3236, opened 2021-06-10
Use local server to test HTTPS instead of google.com

anuraaga:


gradle/java.gradle
```diff
@@ -246,6 +246,14 @@ tasks.withType(Test).configureEach {
   jvmArgs "-Dio.opentelemetry.context.enableStrictContext=${rootProject.findProperty('enableStrictContext') ?: false}"
   // TODO(anuraaga): Have agent map unshaded to shaded.
   jvmArgs "-Dio.opentelemetry.javaagent.shaded.io.opentelemetry.context.enableStrictContext=${rootProject.findProperty('enableStrictContext') ?: false}"
+
+  def trustStore = project(":testing-common").file("src/misc/testing-keystore.p12")
+  inputs.file(trustStore)
+  // Work around payara not working when this is set for some reason.
```
anuraaga, line 252: @laurit Would you be able to help debug this? All the arquilian tests were fine, just not payara which is extremely baffling to me. I'm not familiar with the shrinkwrap stuff which I assume is having an effect.

anuraaga: Ah thanks for the explanation, that helps! Yeah let's leave this temporarily I'm working on a PR to migrate off of OkHttp for tests and I think can fix it then.

laurit (approved):

trask (approved):
nice

PR #3237, opened 2021-06-10
Embedded extension

iNikem:
Builds on top of #3226 allowing to embed extension jar right inside javaagent jar for simplified deployment.

@anuraaga can you take a look at `examples/extension/src/test/java/com/example/javaagent/smoketest/SpringBootIntegrationTest.java`? It often time fails for me. Looks like later tests continue receiving spans from the previous tests. `/clear-requests` on fake backend does not work as expected or?


examples/extension/build.gradle
```diff
@@ -92,17 +92,43 @@ dependencies {
...
+    }
+  }
+}
+
 tasks {
   test {
     useJUnitPlatform()
 
-    def extensionJar = tasks.shadowJar
+    def extensionJar = tasks.extendedAgent
```
mateuszrzeszutek, line 124: Should we have two test cases/tasks here? One for extension placed inside the javaagent jar, one for external one

iNikem: Yeah... Wanted to get some feedback before I start messing with gradle test tasks

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
```diff
@@ -0,0 +1,120 @@
...
+  private static void addFileUrl(List<URL> result, File file) {
+    try {
+      URL wrappedUrl = new URL("otel", null, -1, "/", new RemappingUrlStreamHandler(file));
+      result.add(wrappedUrl);
+    } catch (MalformedURLException ignored) {
+      System.err.println("Ignoring " + file);
+    }
+  }
+
+  public static void extractFile(URL url, File outputFile) throws IOException {
```
laurit, line 113: Could be private?

anuraaga (commented):
@iNikem Do you see the flakiness only locally or also on CI? Either way can you file an issue?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
```diff
@@ -0,0 +1,120 @@
...
+    List<URL> result = new ArrayList<>();
+
+    if (locationName == null) {
+      return result;
+    }
+
+    File location = new File(locationName);
+    if (location.isFile()) {
+      addFileUrl(result, location);
+    } else if (location.isDirectory()) {
```
laurit, line 92: Reading all jars from a directory seems a bit risky, someone will eventually point it to a directory with a million random jars. Perhaps instead of accepting a directory it would be sufficient to accept a comma separated list of paths to extension jars?

iNikem:
> @iNikem Do you see the flakiness only locally or also on CI? Either way can you file an issue?

We don't run examples' tests on CI. So only locally. Will file

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
```diff
@@ -0,0 +1,120 @@
...
+
+    List<ClassLoader> delegates = new ArrayList<>(extensions.size());
+    for (URL url : extensions) {
+      delegates.add(getDelegate(parent, url));
+    }
+    return new MultipleParentClassLoader(parent, delegates);
+  }
+
+  private static void includeEmbeddedExtensionsIfFound(ClassLoader parent, List<URL> extensions) {
+    URL embeddedExtension = parent.getResource("otel-extensions.jar");
```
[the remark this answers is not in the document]
iNikem: That is a great idea! Will do.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
iNikem: That comment should go to #3226 :) But no, I think that from usability point of view it is much easier to point to a directory than forcing user list several files. In any way this should a conscious decision from a person and it will their responsibility to make a correct choice.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
iNikem: Done

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionLoader.java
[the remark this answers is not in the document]
iNikem: Done, PTAL

anuraaga:
@iNikem I didn't confirm the fake backend image referenced in the examples smoketest is new enough for it, but we changed the path to `/clear` at some point, that might be the problem.

iNikem:
> @iNikem I didn't confirm the fake backend image referenced in the examples smoketest is new enough for it, but we changed the path to `/clear` at some point, that might be the problem.

Yes, that's it! Thank you! Fixed now.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -86,8 +136,13 @@ private static void addFileUrl(List<URL> result, File file) {
     }
   }
 
-  private static URLClassLoader getDelegate(ClassLoader parent, URL extensionUrl) {
-    return new ExtensionClassLoader(new URL[] {extensionUrl}, parent);
+  private static void extractFile(JarFile jarFile, JarEntry jarEntry, File outputFile)
```
anuraaga, line 139: `outputDir` since we're extracting a jar?

trask (commented):
Is there any possibility to read directly from the embedded jar? Otherwise we limit possibility to run on read-only file system, a request [we've gotten](https://github.com/microsoft/ApplicationInsights-Java/issues/1673), and @tylerbenson has mentioned previously that they support.

iNikem:
> Is there any possibility to read directly from the embedded jar? Otherwise we limit possibility to run on read-only file system, a request [we've gotten](https://github.com/microsoft/ApplicationInsights-Java/issues/1673), and @tylerbenson has mentioned previously that they support.

I did not find it on the first try. But I don't think this is an actual blocker: on read-only system one can use external extensions jar. I can file a follow-up issue for that.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: No, that parameter is actually a file

anuraaga: Oh oops - so the output file is a JAR file itself. `extractFile` made me think the embedded JAR itself is getting extracted into its contents. I don't think `unpackFile` had the same impression for me so may be a clearer name for people like me

https://github.com/spring-projects/spring-boot/blob/v2.4.3/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/archive/JarFileArchive.java#L140

anuraaga:
FWIW, spring-boot has been extracting embedded JARs to file system for a loong time so it's tried and true :)

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: We extract a file from jarEntry into an outputFile. Like "extract a file from a jar". How can this translate to a double extraction? :) But I can rename if you wish, no problem

examples/extension/build.gradle
iNikem: Still one test task and test class, but now we actually test both with extended agent and the original one.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
anuraaga: There's two "Jar" variables there :) My brain processed it as extracting the second one. Maybe it's silly, so don't mind either way

anuraaga (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -29,33 +38,83 @@
...
+  private static File ensureTempDirectoryExists(File tempDirectory) throws IOException {
+    if (tempDirectory == null) {
+      tempDirectory = Files.createTempDirectory("otel-extensions").toFile();
+      tempDirectory.deleteOnExit();
+    }
+    return tempDirectory;
+  }
+
+  private static URLClassLoader getDelegate(ClassLoader parent, URL extensionUrl) {
+    return new URLClassLoader(new URL[] {extensionUrl}, parent);
```
mateuszrzeszutek, line 114: Shouldn't it return a new `ExtensionClassLoader`?
```suggestion
    return new ExtensionClassLoader(new URL[] {extensionUrl}, parent);
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
```diff
@@ -29,33 +38,83 @@
...
+        if (jarEntry.getName().startsWith(prefix) && !jarEntry.isDirectory()) {
+          tempDirectory = ensureTempDirectoryExists(tempDirectory);
+
+          File tempFile = new File(tempDirectory, jarEntry.getName().substring(prefix.length()));
+          if (tempFile.createNewFile()) {
+            tempFile.deleteOnExit();
+            extractFile(jarFile, jarEntry, tempFile);
+            addFileUrl(extensions, tempFile);
+          } else {
+            System.out.println("Failed to create temp file " + tempFile);
```
mateuszrzeszutek, line 96: Print out to stderr?

```suggestion
            System.err.println("Failed to create temp file " + tempFile);
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ExtensionClassLoader.java
iNikem: oops :)

mateuszrzeszutek (approved):

examples/extension/build.gradle
```diff
@@ -92,19 +92,36 @@ dependencies {
...
+//Extracts manifest from OpenTelemetry Java agent to reuse it later
+task agentManifest(type: Copy) {
+  from zipTree(configurations.otel.singleFile).matching {
+    include 'META-INF/MANIFEST.MF'
+  }
+  into buildDir
+}
+
+//Produces a copy of upstream javaagent with this extension jar included inside it
+//The location of extension jar inside agent jar is hard-coded in the agent source code
```
trask, line 104:
```suggestion
//The location of extension directory inside agent jar is hard-coded in the agent source code
```

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
```diff
@@ -53,38 +52,29 @@ public static void premain(String agentArgs, Instrumentation inst) {
...
-      URL bootstrapUrl = installBootstrapJar(inst);
-
-      Class<?> agentInitializerClass =
-          ClassLoader.getSystemClassLoader()
-              .loadClass("io.opentelemetry.javaagent.bootstrap.AgentInitializer");
-      Method startMethod =
-          agentInitializerClass.getMethod("initialize", Instrumentation.class, URL.class);
-      startMethod.invoke(null, inst, bootstrapUrl);
+      File javaagentFile = installBootstrapJar(inst);
+      AgentInitializer.initialize(inst, javaagentFile);
```
trask, line 56: 👍

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -74,21 +67,15 @@ private static ClassLoader createAgentClassLoader(String innerJarFilename, URL b
...
 
-    Class<?> loaderClass =
-        ClassLoader.getSystemClassLoader()
-            .loadClass("io.opentelemetry.javaagent.bootstrap.AgentClassLoader");
-    Constructor<ClassLoader> constructor =
-        (Constructor<ClassLoader>)
-            loaderClass.getDeclaredConstructor(URL.class, String.class, ClassLoader.class);
     ClassLoader agentClassLoader =
-        constructor.newInstance(bootstrapUrl, innerJarFilename, agentParent);
+        new AgentClassLoader(javaagentFile, innerJarFilename, agentParent);
```
trask, line 71: 👍

trask (approved):
Nice tests.

> FWIW, spring-boot has been extracting embedded JARs to file system

I assumed that spring-boot used a class loader to read the embedded jars directly, given the requirement that the embedded jars have to be uncompressed?

PR #3240, opened 2021-06-10
Migrate HttpServerTest to Armeria

anuraaga:
I randomly noticed the okhttp instrumentation tests never run against the baseline version because it picks up the higher version from testing-common. Like we did for the server, we should shade the client too to prevent testing components from interacting with instrumentation. Since Armeria's already shading, switching the client is simplest I guess.

I'll follow up to migrate remaining usages.

### DIFF
--- a/instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ServerTest.groovy
+++ b/instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ServerTest.groovy
@@ -71,15 +71,17 @@ class Netty41ServerTest extends HttpServerTest<EventLoopGroup> implements AgentT
                       response = new DefaultFullHttpResponse(HTTP_1_1, HttpResponseStatus.valueOf(endpoint.status), content)
                       break
                     case INDEXED_CHILD:
+                      content = Unpooled.EMPTY_BUFFER
                       endpoint.collectSpanAttributes { new QueryStringDecoder(uri).parameters().get(it).find() }
-                      response = new DefaultFullHttpResponse(HTTP_1_1, HttpResponseStatus.valueOf(endpoint.status))
+                      response = new DefaultFullHttpResponse(HTTP_1_1, HttpResponseStatus.valueOf(endpoint.status), content)
                       break
                     case QUERY_PARAM:
                       content = Unpooled.copiedBuffer(uri.query, CharsetUtil.UTF_8)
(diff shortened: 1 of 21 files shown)


instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ServerTest.groovy
```diff
@@ -71,15 +71,17 @@ class Netty41ServerTest extends HttpServerTest<EventLoopGroup> implements AgentT
                       response = new DefaultFullHttpResponse(HTTP_1_1, HttpResponseStatus.valueOf(endpoint.status), content)
                       break
                     case INDEXED_CHILD:
+                      content = Unpooled.EMPTY_BUFFER
```
anuraaga, line 74: @trustin @minwoox @ikhoon OkHttp client worked fine with the old code but Armeria seems to require the content length to be set. Is it expected?

anuraaga (commented):
@trustin @minwoox @ikhoon If you have time to skim this to make sure I didn't miss any more recent Armeria idioms that would be nice. Thanks!

instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ServerTest.groovy
anuraaga: Ah to clarify the behavior is that Armeria client times out (timeout is set to 1 minute). Just doesn't seem to treat the response as complete

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -430,22 +394,21 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
...
     when:
     count.times { index ->
       def job = {
         latch.await()
-        def url = HttpUrl.get(endpoint.resolvePath(address)).newBuilder()
-          .query("${ServerEndpoint.ID_PARAMETER_NAME}=$index")
-          .build()
-        Request.Builder builder = request(url, "GET", null)
+        HttpRequestBuilder request = HttpRequest.builder()
+          .get(endpoint.resolvePath(address).toString().replace("http://", "h1c://"))
```
trask, line 406:
```suggestion
          // Force HTTP/1 via h1c so upgrade requests don't show up as traces
          .get(endpoint.resolvePath(address).toString().replace("http://", "h1c://"))
```

trask (approved):

minwoox:
Looks perfect! 👍

PR #3241, opened 2021-06-10
Test spring-integration instrumentation with spring-cloud-stream example

mateuszrzeszutek:
Adds a test that deploys two apps (producer, consumer) using a RabbitMQ queue, verifies that context is propagated.

### DIFF
--- a/instrumentation/spring/spring-integration-4.1/library/src/test/groovy/SpringCloudStreamRabbitTest.groovy
+++ b/instrumentation/spring/spring-integration-4.1/library/src/test/groovy/SpringCloudStreamRabbitTest.groovy
@@ -1,0 +1,131 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.runInternalSpan
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.runUnderTrace
+
+import io.opentelemetry.instrumentation.test.LibraryInstrumentationSpecification
+import java.time.Duration
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.SpringApplication
+import org.springframework.boot.SpringBootConfiguration
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration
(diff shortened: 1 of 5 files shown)


instrumentation/spring/spring-integration-4.1/library/src/test/groovy/SpringCloudStreamRabbitTest.groovy
```diff
@@ -0,0 +1,131 @@
...
+          name "producer"
+        }
+        span(1) {
+          name "testProducer.output"
+          childOf span(0)
+        }
+        span(2) {
+          name "testConsumer.input"
+          childOf span(1)
+        }
```
trask, line 90: sorry I forget, is the plan for these spans to have messaging semantic attributes and suppress the underlying rabbitmq spans?

trask (approved):

instrumentation/spring/spring-integration-4.1/library/src/test/groovy/SpringCloudStreamRabbitTest.groovy
```diff
@@ -0,0 +1,131 @@
...
+  def cleanupSpec() {
+    rabbitMQContainer?.stop()
+    producerContext?.close()
+    consumerContext?.close()
+  }
+
+  def "should propagate context through RabbitMQ"() {
+    when:
+    producerContext.getBean("producer", Runnable).run()
+    consumerContext.getBean("latch", CountDownLatch).await(10, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
mateuszrzeszutek: True, it's not necessary -- will remove. Thanks!

instrumentation/spring/spring-integration-4.1/library/src/test/groovy/SpringCloudStreamRabbitTest.groovy
mateuszrzeszutek: No, these were supposed to just propagate the context and coexist with actual underlying messaging instrumentations.
Though, now that I think of it, the reason why I added these spans was storing links; since we've decided on always getting the parent from the message they might not be necessary - the only thing that justifies their existence is adding custom `AttributesExtractor`s in the library instrumentation... I'll try removing the span in a separate PR.

Also, I'm planning to add a test that verifies how this instrumentation works with the rabbitmq instrumentation in the next PR: javaagent instrumentation.

PR #3250, opened 2021-06-11
Migrate remaining test client calls to Armeria and remove testing-com…

anuraaga:
…mon dependency.


instrumentation/log4j/log4j-2.13.2/library/log4j-2.13.2-library.gradle
```diff
@@ -3,5 +3,9 @@ apply from: "$rootDir/gradle/instrumentation-library.gradle"
 dependencies {
   library "org.apache.logging.log4j:log4j-core:2.13.2"
 
+  // Library instrumentation cannot be applied to 2.13.2 due to a bug in Log4J. The agent works
```
anuraaga, line 6: Didn't expect to find a second case of dependency conflict caused by testing-common, heh

instrumentation/tapestry-5.4/javaagent/src/test/groovy/TapestryTest.groovy
```diff
@@ -54,6 +54,25 @@ class TapestryTest extends AgentInstrumentationSpecification implements HttpServ
     return "/jetty-context"
   }
 
+  WebClient client
+
+  def setup() {
+    client = WebClient.builder(address)
+      .decorator(new DecoratingHttpClientFunction() {
+        // https://github.com/line/armeria/issues/2489
```
anuraaga, line 62: I decided to copy-paste this redirect logic because there are only two locations. If a third, I'd extract - I am hoping the issue gets fixed before that though. @trustin @minwoox @ikhoon

minwoox: Thanks for reminding me. 😅
Yeah, we will do. 😄

laurit (approved):

PR #3255, opened 2021-06-11
Replace Guava cache with Caffeine in agent type pool.

anuraaga:
- Caffeine doesn't expose concurrency level, presumably since we wouldn't expect it to be a good idea to ever set it (?)
- Our cache API doesn't expose `softValues`. I could add it, but it doesn't seem worth it just for a cache of 64 tiny objects.
- Similarly, could add a size method to cache API but tests using it seem to mostly just test the implementation detail of the cache (semantics are handled by the reference comparisons)


mateuszrzeszutek (approved):

laurit:
> * Caffeine doesn't expose concurrency level, presumably since we wouldn't expect it to be a good idea to ever set it (?)
I believe this is because modern `ConcurrentHashMap` doesn't seem to care much about concurrency level, old version used to divide map into concurrency level segments which were also maps.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/bytebuddy/AgentCachingPoolStrategy.java
```diff
@@ -61,23 +50,13 @@
    * </ul>
    */
   final Cache<ClassLoader, WeakReference<ClassLoader>> loaderRefCache =
-      CacheBuilder.newBuilder()
-          .weakKeys()
-          .concurrencyLevel(CONCURRENCY_LEVEL)
-          .initialCapacity(LOADER_CAPACITY / 2)
-          .maximumSize(LOADER_CAPACITY)
-          .build();
+      Cache.newBuilder().setWeakKeys().setMaximumSize(LOADER_CAPACITY).build();
```
laurit, line 53: How many class loaders could there possibly be? Does it make sense to limit maximum size here at all?

trask: and it's only caching a very small object per ClassLoader, so little downside if there are a lot of class loaders

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/bytebuddy/AgentCachingPoolStrategy.java
anuraaga: Removed the classloader limit - curious if 64 is too small a number for the other one but leaving it for now.

anuraaga:
@ben-manes Just curious, how did you happen to find this PR? :D

ben-manes:
@anuraaga github's global search is neat. I have a list of keywords that I search on and review what has been recently updated. That's nice because (a) its motivating to see others use my work, (b) I can help guide users from becoming frustrated, (c) I might learn something. Sometimes that's lead to minor improvements or increased adoption, which is great. But mostly it's to retain some project enthusiasm as doing OSS as a hobby is hard to sustain after you've done the fun parts. I think it would help other maintainers if something like [Google Alerts](https://en.wikipedia.org/wiki/Google_Alerts) was natively part of Github ot highlight usage.

anuraaga:
Neat - thanks for such diligence for Caffeine!

PR #3256, opened 2021-06-11
Extract agent shadow configuration to conventions script.

anuraaga:
We have been desparately in need of making this common logic for a long time. And for fixing the -all jar, I am expecting yet another project that contains the bundle of exporters with the shadow configuration. So finally did it.

I am using this relatively new pattern of conventions plugins

https://docs.gradle.org/current/samples/sample_convention_plugins.html

Eventually we can migrate all the `gradle/*.gradle` files to conventions to allow using the `plugins` block everywhere.


mateuszrzeszutek (approved):

iNikem (approved):

trask (approved):
> We have been desparately in need of making this common logic for a long time

💯

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/ReferenceCollector.java
```diff
@@ -48,6 +49,14 @@
  * plugin.
  */
 public class ReferenceCollector {
+
+  // ReferenceCollector's classloader has a parent including the Gradle classpath, such as buildSrc
```
anuraaga, line 53: It turns out we were previously loading SPI helper resources from buildSrc before our own classpath. Adding shadow to buildSrc, for use in the convention plugin, broke log4j instrumentation since shadow (probably incorrectly, it seems to be the log4j default) bundles in a log4j service file.

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+  implementation("org.eclipse.aether:aether-connector-basic:1.1.0")
+  implementation("org.eclipse.aether:aether-transport-http:1.1.0")
+  implementation("org.apache.maven:maven-aether-provider:3.3.9")
+
+  implementation("com.google.guava:guava:30.1-jre")
+  implementation("gradle.plugin.com.github.jengelman.gradle.plugins:shadow:7.0.0")
+  implementation("org.ow2.asm:asm:7.0-beta")
+  implementation("org.ow2.asm:asm-tree:7.0-beta")
+  implementation("org.apache.httpcomponents:httpclient:4.5.10")
+  // When updating, also update dependencyManagement/dependencyManagement.gradle.kts
```
anuraaga, line 47: /cc @iNikem I bundled in an update to the byte buddy gradle plugin into this PR (it didn't fix the problem I was seeing but was my first try :P)

Issue #3221, opened 2021-06-08
default-enabled=false fails with NoClassDefFoundError on internal instrumentation


anuraaga:
I think we need internal instrumentation enabled all the time regardless of that flag, or `default-enabled` flag becomes too hard to use in practice.

https://cloud-native.slack.com/archives/C0150QF88FL/p1623072246099100

PR #3257, opened 2021-06-11
Internal instrumentations should always be enabled by default

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3221

It's still possible to disable one of those instrumentations if somebody wants to do it for some reason, but they have to do it one by one using the `otel.instrumentation.<name>.enabled` property.


laurit (approved):

trask (approved):
that was easier than I thought it would be 👍

PR #3258, opened 2021-06-11
Enable https tests

laurit:


mateuszrzeszutek (approved):

instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/client/NettyHttpClientTracer.java
```diff
@@ -62,24 +62,24 @@ public void connectionFailure(Context parentContext, Channel channel, Throwable
...
-  protected URI url(HttpRequest request) throws URISyntaxException {
-    URI uri = new URI(request.getUri());
+  protected URI url(NettyRequestWrapper request) throws URISyntaxException {
+    URI uri = new URI(request.uri());
     if ((uri.getHost() == null || uri.getHost().equals("")) && request.headers().contains(HOST)) {
-      return new URI("http://" + request.headers().get(HOST) + request.getUri());
-    } else {
-      return uri;
+      String protocol = request.isHttps() ? "https://" : "http://";
+      uri = new URI(protocol + request.headers().get(HOST) + request.uri());
```
trask, line 80: 👍

trask (approved):

instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ClientTest.groovy
```diff
@@ -39,31 +42,50 @@ import spock.lang.Unroll
...
+    Bootstrap bootstrap = new Bootstrap()
+    bootstrap.group(eventLoopGroup)
       .channel(getChannelClass())
       .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, CONNECT_TIMEOUT_MS)
       .handler(new ChannelInitializer<SocketChannel>() {
         @Override
         protected void initChannel(SocketChannel socketChannel) throws Exception {
           ChannelPipeline pipeline = socketChannel.pipeline()
+          if (https) {
+            SslContext sslContext = SslContextBuilder.forClient().trustManager(InsecureTrustManagerFactory.INSTANCE).build()
```
anuraaga, line 67: If possible we should avoid the insecure trust manager in tests too. Is the system property trustStore not enough here? If so it would be nice to read it and set up the trust manager manually using it.

laurit: For some reason it was needed when we used google for https test. I didn't try without it after google -> localhost change.

PR #3262, opened 2021-06-11
Port AsyncSpanEndStrategy to Instrumenter API

mateuszrzeszutek:
This PR ports https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2530 to Instrumenter API - we need this to be able to port `opentelemetry-annotations` and `spring-boot-autoconfigure` to Instrumenter API. 

CC @HaloFour (I can't add you as a reviewer for some reason)

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndStrategies.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndStrategies.java
@@ -5,34 +5,36 @@
 
 package io.opentelemetry.instrumentation.api.asyncannotationsupport;
 
-import static java.util.Objects.requireNonNull;
-
-import java.util.List;
-import java.util.concurrent.CopyOnWriteArrayList;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /** A global registry of {@link AsyncOperationEndStrategy} implementations. */
-public final class AsyncOperationEndStrategies {
-  private static final List<AsyncOperationEndStrategy> STRATEGIES = new CopyOnWriteArrayList<>();
+public abstract class AsyncOperationEndStrategies {
+  private static volatile AsyncOperationEndStrategies instance;
 
-  static {
-    STRATEGIES.add(Jdk8AsyncOperationEndStrategy.INSTANCE);
+  /**
+   * Sets the actual strategies' registry implementation. The javaagent uses weak references to make
+   * unloading strategy classes possible.
+   *
+   * <p>This is supposed to be only called by the javaagent. <b>Instrumentation must not call
+   * this.</b>
+   */
+  public static void internalSetStrategiesStorage(AsyncOperationEndStrategies strategies) {
+    instance = strategies;
+  }
+
+  /** Obtain instance of the async strategy registry. */
+  public static AsyncOperationEndStrategies instance() {
+    if (instance == null) {
+      instance = new AsyncOperationEndStrategiesImpl();
+    }
+    return instance;
   }
 
   /** Add the passed {@code strategy} to the registry. */
-  public static void registerStrategy(AsyncOperationEndStrategy strategy) {
-    STRATEGIES.add(requireNonNull(strategy));
-  }
+  public abstract void registerStrategy(AsyncOperationEndStrategy strategy);
 
   /** Remove the passed {@code strategy} from the registry. */
-  public static void unregisterStrategy(AsyncOperationEndStrategy strategy) {
-    STRATEGIES.remove(strategy);
-  }
-
-  /** Remove all strategies of type {@code strategyClass} from the registry. */
-  public static void unregisterStrategy(Class<? extends AsyncOperationEndStrategy> strategyClass) {
-    STRATEGIES.removeIf(strategy -> strategy.getClass() == strategyClass);
-  }
+  public abstract void unregisterStrategy(AsyncOperationEndStrategy strategy);
 
   /**
    * Returns an {@link AsyncOperationEndStrategy} that is able to compose over {@code returnType},
@@ -40,12 +42,5 @@ public final class AsyncOperationEndStrategies {
    * registry.
    */
   @Nullable
-  public static AsyncOperationEndStrategy resolveStrategy(Class<?> returnType) {
-    for (AsyncOperationEndStrategy strategy : STRATEGIES) {
-      if (strategy.supports(returnType)) {
-        return strategy;
-      }
-    }
-    return null;
-  }
+  public abstract AsyncOperationEndStrategy resolveStrategy(Class<?> returnType);
 }
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndStrategies.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndStrategies.java
@@ -12,7 +12,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /** A global registry of {@link AsyncOperationEndStrategy} implementations. */
-public final class AsyncEndStrategies {
+public final class AsyncOperationEndStrategies {
   private static final List<AsyncOperationEndStrategy> STRATEGIES = new CopyOnWriteArrayList<>();
 
   static {
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndSupport.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndSupport.java
@@ -11,32 +11,39 @@ import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 /**
  * A wrapper over {@link Instrumenter} that is able to defer {@link Instrumenter#end(Context,
  * Object, Object, Throwable)} until asynchronous computation finishes.
- *
- * <p>This class is not able to extract {@code RESPONSE} from the asynchronous computation value it
- * receives in the {@link #asyncEnd(Context, Object, Object, Throwable)} call, so it will always
- * pass {@code null} as the response to the wrapped {@link Instrumenter}.
  */
-public final class AsyncOperationEndSupport<REQUEST> {
+public final class AsyncOperationEndSupport<REQUEST, RESPONSE> {
 
   /**
    * Returns a new {@link AsyncOperationEndSupport} that wraps over passed {@code syncInstrumenter},
-   * configured for usage with asynchronous computations that are instances of {@code asyncType}.
+   * configured for usage with asynchronous computations that are instances of {@code asyncType}. If
+   * the result of the async computation ends up being an instance of {@code responseType} it will
+   * be passed as the response to the {@code syncInstrumenter} call; otherwise {@code null} value
+   * will be used as the response.
    */
-  public static <REQUEST> AsyncOperationEndSupport<REQUEST> create(
-      Instrumenter<REQUEST, ?> syncInstrumenter, Class<?> asyncType) {
+  public static <REQUEST, RESPONSE> AsyncOperationEndSupport<REQUEST, RESPONSE> create(
+      Instrumenter<REQUEST, RESPONSE> syncInstrumenter,
+      Class<RESPONSE> responseType,
+      Class<?> asyncType) {
     return new AsyncOperationEndSupport<>(
-        syncInstrumenter, asyncType, AsyncOperationEndStrategies.resolveStrategy(asyncType));
+        syncInstrumenter,
+        responseType,
+        asyncType,
+        AsyncOperationEndStrategies.instance().resolveStrategy(asyncType));
   }
 
-  private final Instrumenter<REQUEST, ?> instrumenter;
+  private final Instrumenter<REQUEST, RESPONSE> instrumenter;
+  private final Class<RESPONSE> responseType;
   private final Class<?> asyncType;
   private final AsyncOperationEndStrategy asyncOperationEndStrategy;
 
   private AsyncOperationEndSupport(
-      Instrumenter<REQUEST, ?> instrumenter,
+      Instrumenter<REQUEST, RESPONSE> instrumenter,
+      Class<RESPONSE> responseType,
       Class<?> asyncType,
       AsyncOperationEndStrategy asyncOperationEndStrategy) {
     this.instrumenter = instrumenter;
+    this.responseType = responseType;
     this.asyncType = asyncType;
     this.asyncOperationEndStrategy = asyncOperationEndStrategy;
   }
@@ -64,7 +71,8 @@ public final class AsyncOperationEndSupport<REQUEST> {
 
     // use the configured strategy to compose over the asyncValue
     if (asyncOperationEndStrategy != null && asyncType.isInstance(asyncValue)) {
-      return (ASYNC) asyncOperationEndStrategy.end(instrumenter, context, request, asyncValue);
+      return (ASYNC)
+          asyncOperationEndStrategy.end(instrumenter, context, request, asyncValue, responseType);
     }
 
     // fall back to sync end() if asyncValue type doesn't match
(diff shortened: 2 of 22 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
```diff
@@ -0,0 +1,84 @@
...
+      Instrumenter<REQUEST, ?> instrumenter,
+      Class<?> asyncType,
+      AsyncOperationEndStrategy asyncOperationEndStrategy) {
+    this.instrumenter = instrumenter;
+    this.asyncType = asyncType;
+    this.asyncOperationEndStrategy = asyncOperationEndStrategy;
+  }
+
+  /** Same as calling {@link Instrumenter#shouldStart(Context, Object)}. */
+  public boolean shouldStart(Context parentContext, REQUEST request) {
```
HaloFour, line 45: Are `shouldStart`/`start` expected to ever be called through `AsyncInstrumenter`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** A global registry of {@link AsyncOperationEndStrategy} implementations. */
+public final class AsyncEndStrategies {
+  private static final List<AsyncOperationEndStrategy> STRATEGIES = new CopyOnWriteArrayList<>();
```
laurit, line 16: shouldn't this be something weak

HaloFour:
I think it looks good.

As a minor refactor of the async span strategies would you want me to follow this up with a refactor of the implementations of those strategies in the other instrumentation modules?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.async;
```
anuraaga, line 6: Does this need to be in instrumentation API? It seems to only really be needed by one instrumentation, the annotations Instrumentation, isn't it?

HaloFour: And springboot autoconfigure

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
mateuszrzeszutek: Good question -- it probably won't ever be called in this codebase. To be honest, this class is probably one that I like the least in this PR. Maybe I should have named it `AsyncInstrumenterEnd` and only keep `asyncEnd()` method

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
mateuszrzeszutek: Actually no, not this one - these strategies shouldn't expire and get removed automatically, once installed they should be always available until unregistered (or the app shuts down)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
mateuszrzeszutek: Yep, we need to keep it in some shared place.

> I did suggest at that time that maybe this was a little too specialized for instrumentation-api and maybe it would make sense to have a separate shared module for AOP concerns?

Hmm, I wonder about that - I fear that introducing yet another module may only increase the cognitive load of implementing an instrumentation.

@anuraaga WDYT about marking the whole package as `@UnstableApi`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
laurit: What if the application that contains the library with async strategy gets undeployed? Won't keeping a strong ref here prevent gc of that applications class loader?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
anuraaga: Cognitive load is precisely why we might want to find some split for it. I think async return value instrumentation is far less common than others so having a package in our instrumentation API could be too discoverable for our generally expected user. Just because a server framework is async, returning CompletableFuture, doesn't mean it needs to care about this package but someone might think so just by the word async, I would have I think.

Marking as UnstableApi seems like a good idea, can we maybe name the package annotationsasyncsupport or something similar to tie it closer to the instrumentation?

HaloFour: > Hmm, I wonder about that - I fear that introducing yet another module may only increase the cognitive load of implementing an instrumentation.

I guess it depends on what instrumentation we expect to rely on these classes.  If it's just the AOP instrumentation (`opentelemetry-annotations-1.0` and `spring-boot-autoconfigure`) as well as the different reactive frameworks (`guava`,`rxjava2`, `rxjava3` and `reactor`, with maybe a couple more being added over time), I don't think it's too problematic to spit it into a separate module as it's use will be fairly specialized.

However, if it's expected that instrumentation for various reactive client libraries (e.g. `lettuce`, `spring-webflux-5.0`) might also want to reuse the strategies, then it probably makes more sense for them to live in `instrumentation-api`.  That said I don't think the strategies are geared towards that use, especially now that some of them are configuration driven and there's an expectation that they are registered with the reactive framework they represent.  Although nothing would stop that instrumentation module from creating its own instance of the strategy rather than bouncing through `AsyncEndStrategies`.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
mateuszrzeszutek: > What if the application that contains the library with async strategy gets undeployed? Won't keeping a strong ref here prevent gc of that applications class loader?

Oh, that's an excellent point. `AsyncOperationEndStrategy` instances are created just to be registered in this class, they're not being referenced in any other context - a weak-keys CL->strategy map wouldn't work, since the strategy references the CL; a weak-values one wouldn't work either, because strategies could disappear even while they're still needed... Any ideas?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
mateuszrzeszutek: > having a package in our instrumentation API could be too discoverable for our generally expected user. Just because a server framework is async, returning CompletableFuture, doesn't mean it needs to care about this package but someone might think so just by the word async, I would have I think.

That's very true 👍 💯 

> we maybe name the package annotationsasyncsupport or something similar to tie it closer to the instrumentation?

Done.

You've completely convinced me about the split (it'd make instrumentation-api much cleaner!), now I'm starting to think if we have any other scenarios like this one that can benefit from it 😄

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
laurit: The api could say that it is users responsibility to keep strategy alive e.g. strategy instances could be singletons. Or we could build an utility that allows inserting strong reference from class loader to object and use that to bind `AsyncOperationEndStrategy` instance to the class loader that defined it so that we could use weak refs without worrying that the strategy instance gets collected prematurely.

HaloFour: Would that work with the ways that `AsyncSpanEndStrategy` is implemented now?  What in the different reactive modules would hold onto the root reference of the strategy?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
HaloFour: _Bikeshedding:_  `AsyncOperationEndInstrumenter`, to sync the name with the strategy interface name?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
mateuszrzeszutek: I think `AsyncSpanEndStrategy` has the same problem - the non-enum strategies don't have any singleton instance and putting them in a weak reference would make them disappear.

> What in the different reactive modules would hold onto the root reference of the strategy?

In our javaagent instrumentations we can just create static singleton instances of them and that would solve the problem of weak references, but I have no idea what to do in library instrumentations, like spring-boot-autoconfigure.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncInstrumenter.java
mateuszrzeszutek: 👍 
I ended up with `AsyncOperationEndSupport`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndSupport.java
```diff
@@ -0,0 +1,74 @@
...
+package io.opentelemetry.instrumentation.api.asyncannotationsupport;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+
+/**
+ * A wrapper over {@link Instrumenter} that is able to defer {@link Instrumenter#end(Context,
+ * Object, Object, Throwable)} until asynchronous computation finishes.
+ *
+ * <p>This class is not able to extract {@code RESPONSE} from the asynchronous computation value it
```
anuraaga, line 15: I don't quite follow - why is this? Is it because some return types don't have one response (e.g., a reactive stream that isn't a mono)? Can we at least resolve it for the mono-like cases (especially CompletableFuture)?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndStrategies.java
```diff
@@ -0,0 +1,51 @@
...
+    STRATEGIES.add(requireNonNull(strategy));
+  }
+
+  /** Remove the passed {@code strategy} from the registry. */
+  public static void unregisterStrategy(AsyncOperationEndStrategy strategy) {
+    STRATEGIES.remove(strategy);
+  }
+
+  /** Remove all strategies of type {@code strategyClass} from the registry. */
+  public static void unregisterStrategy(Class<? extends AsyncOperationEndStrategy> strategyClass) {
```
anuraaga, line 33: Do we need this? It seems to not be symmetrical

HaloFour: I believe this was also inherited from `AsyncSpanEndStrategy`.  The strategies were initially singletons so "unregistering" them by instance was sufficient but when they became configurable I wanted a way to unregister by type so that I didn't have to track strategy instances in the instrumentation, which generally has no state.

I'm not arguing for keeping that method or porting it to the new API, just providing context as to why that method was originally added.

mateuszrzeszutek: On the other hand: if those strategies were stores as fields/constants, there would be no need for this method; we could unregister them by reference.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndSupport.java
mateuszrzeszutek: > Is it because some return types don't have one response (e.g., a reactive stream that isn't a mono)?

That was one reason; the other "reason" (well, "excuse" would be more correct 😅 ) was that all those strategies pretty much operate on raw/unchecked types, before passing the response to the Instrumenter I'd have to check if it's an instance of RESPONSE type - I'd need to pass `Class<RESPONSE>` somewhere to do that, and I wanted to keep those strategies as simple as possible.
I'll work on that; multi-result reactive types should be the only thing that's not passed to the instrumenter.

HaloFour: Having gone down the Java generics and reflection rabbit hole recently this sounds like it would be very tricky to really get right. Would the strategy itself be responsible for trying to negotiate the actual type parameters of the returned generic type?

mateuszrzeszutek: I thought to make it as simple as possible, just use a `Class<RESPONSE>` instance to check type and cast, without delving into Java generics weirdness. All `Instrumenter`s in this project that would use the async support would use `Void` as the response type, so we'd never actually need to pass anything there - at least in this repo.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
mateuszrzeszutek: Yeah, switching the implementation sounds like a reasonable idea. I'll try it out.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndSupport.java
HaloFour: Maybe I'm not understanding.  Is the goal to be able to derive that `RESPONSE` is `String` and to pass a `Class<String>` to the instrumentation if the annotated method returns `CompletableFuture<String>` or `Mono<String>` or the like?  Or is this for non-AOP instrumentation?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/async/AsyncEndStrategies.java
mateuszrzeszutek: Done 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/asyncannotationsupport/AsyncOperationEndSupport.java
mateuszrzeszutek: I went with the simplest possible approach:

```java
  public static <REQUEST, RESPONSE> AsyncOperationEndSupport<REQUEST, RESPONSE> create(
      Instrumenter<REQUEST, RESPONSE> syncInstrumenter,
      Class<RESPONSE> responseType,
      Class<?> asyncType)
```

You have to manually pass the expected response type. Extracting it from the passed `Instrumenter` is impossible, it will never have type parameters in the runtime.

mateuszrzeszutek:
> As a minor refactor of the async span strategies would you want me to follow this up with a refactor of the implementations of those strategies in the other instrumentation modules?

Oh, sorry @HaloFour , I haven't noticed your comment before, it got overshadowed by all the other ones 🙇 

If you only find some time for that, yes please! That'd be really helpful.

HaloFour:
@mateuszrzeszutek 

> If you only find some time for that, yes please! That'd be really helpful.

Can do, should I start a branch off of this one or would it be better to wait until this is approved/merged?

mateuszrzeszutek:
> Can do, should I start a branch off of this one or would it be better to wait until this is approved/merged?

@anuraaga will you find some time to look at this PR this week? If not, then it'll probably be better for @HaloFour to branch off.

anuraaga (approved):

PR #3263, opened 2021-06-11
Akka http: propagate context to callbacks

laurit:
Additionally adds `PoolMasterActorInstrumentation` which prevents context leaks be setting context to root context during library intialization. The idea for this was stolen from datadog.


trask:
I extracted the DataDog portion out to #3264 in order to merge directly and maintain original author/attribution in git history

trask (approved):

PR #3278, opened 2021-06-13
Consistent github workflows extension

trask:
There were more `.yaml` already, if anyone prefers them all to be `.yml` I can update in that direction instead.


anuraaga (approved):

anuraaga:
I have not verified if GitHub itself is consistent, but the first filename in their docs I found is `.yml`

https://docs.github.com/en/actions/learn-github-actions/introduction-to-github-actions#create-an-example-workflow

So perhaps that's better

trask:
Renamed yaml -> yml. If others chime in and want yaml, will just revert the last commit 😄.

iNikem (approved):

PR #3281, opened 2021-06-13
Refactor span names class

trask:
A few things in this PR:

* Remove already deprecated `BaseTracer.spanName*` methods
* Split out `ClassNames` from `SpanNames`
  * Rename `SpanNames.spanNameForMethod` --> `SpanNames.fromMethod`
  * Move `SpanNames.spanNameForClass` --> `ClassNames.simpleName`

Based on @anuraaga's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3070#discussion_r640303721:

> Not for this PR, just for everyone's thought, I feel like this method shouldn't be needed (publicly) either we should audit its usage, spans are methods, not classes.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/ClassNames.java
```diff
@@ -0,0 +1,27 @@
...
+
+package io.opentelemetry.instrumentation.api.tracer;
+
+public class ClassNames {
+
+  /**
+   * This method is used to generate a simple name based on a given class reference, e.g. for use in
+   * span names and span attributes. Anonymous classes are named based on their parent.
+   */
+  public static String simpleName(Class<?> clazz) {
```
anuraaga, line 14: Add `ClassValue`? Of course fine in another PR :-)

anuraaga (approved):
I'm hoping as we migrate more to instrumenter API, we'll use a `CodeSpanNameExtractor.create(codeAttributes)` pattern instead of this class directly

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/ClassNames.java
```diff
@@ -7,21 +7,29 @@
...
 
+  private static final ClassValue<String> simpleNames =
+      new ClassValue<String>() {
+        @Override
+        protected String computeValue(Class<?> type) {
+          if (!type.isAnonymousClass()) {
+            return type.getSimpleName();
+          }
+          String className = type.getName();
+          if (type.getPackage() != null) {
```
laurit, line 18: Packages are optional. Perhaps it would be better to use the same strategy as jdk uses
```
className = className.substring(className.lastIndexOf('.') + 1);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/ClassNames.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.tracer;
+
+public class ClassNames {
```
mateuszrzeszutek, line 8: Nit: make it final and add a private constructor

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/code/CodeSpanNameExtractor.java
```diff
@@ -33,7 +33,7 @@ private CodeSpanNameExtractor(CodeAttributesExtractor<REQUEST, ?> attributesExtr
   public String extract(REQUEST request) {
     Class<?> cls = attributesExtractor.codeClass(request);
     // TODO: avoid using SpanNames, encapsulate the logic here
```
mateuszrzeszutek, line 35: I guess we can remove this TODO

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

PR #3283, opened 2021-06-14
Remove unnecessary final on params and local vars

trask:
This is not the style we follow (anymore).

Also thanks to effectively final in Java 8.

### DIFF
--- a/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/muzzle/MuzzlePlugin.java
+++ b/buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/muzzle/MuzzlePlugin.java
@@ -212,7 +212,7 @@ public class MuzzlePlugin implements Plugin<Project> {
       Artifact versionArtifact,
       Project instrumentationProject,
       TaskProvider<?> runAfter) {
-    String taskName;
+    final String taskName;
     if (muzzleDirective.getCoreJdk().get()) {
       taskName = "muzzle-Assert" + muzzleDirective;
     } else {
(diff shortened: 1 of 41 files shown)


buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/muzzle/MuzzlePlugin.java
```diff
@@ -212,7 +212,7 @@ private static TaskProvider addMuzzleTask(
       Artifact versionArtifact,
       Project instrumentationProject,
       TaskProvider<?> runAfter) {
-    final String taskName;
```
anuraaga, line 215: Still recommend `final` when variable isn't initialized inline

anuraaga: Applies to many of the changes in this PR :P

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/MessagePropertyGetter.java
```diff
@@ -23,7 +23,7 @@
   @Override
   public String get(MessageWithDestination carrier, String key) {
     String propName = key.replace("-", MessagePropertySetter.DASH);
-    Object value;
+    final Object value;
```
anuraaga, line 26: :+1:

anuraaga (approved):

buildSrc/src/main/java/io/opentelemetry/instrumentation/gradle/muzzle/MuzzlePlugin.java
trask: oh yes, reverted them all, though I'm not sure I understand the purpose for marking these vars final?

anuraaga: When initialization happens deep in some control structure, accidentally double-initializing is a relatively common bug (or at least that's what I was taught when learning Java a long time ago :P)

iNikem (approved):

PR #3286, opened 2021-06-14
Only include exporters in -all jar.

### LINKED ISSUE #3082: opentelemetry-javaagent-1.2.0.jar includes all dependencies and all exporters
**Describe the bug**
`opentelemetry-javaagent-1.2.0.jar` includes all dependencies. While that is expected for `opentelemetry-javaagent-1.2.0-all.jar`, the one *without* `-all` should not include the exporters.

**Steps to reproduce**
Go to e.g. [Maven Central](https://repo1.maven.org/maven2/io/opentelemetry/javaagent/opentelemetry-javaagent/1.2.0/) and download the non-all version. 

A)
```
unzip -v opentelemetry-javaagent-1.2.0.jar  | grep -F " inst/"
```

B)
```
unzip -v opentelemetry-javaagent-1.2.0.jar  | grep -F "inst/io/opentelemetry/exporter/"
```

A) I *believe* this should return an empty response
B) I'm sure this should return an empty response.

**What did you see instead?**
A) approx 15'000 files in
B) 142 files

`v1.2.0`

**Environment**
Not relevant.

**Additional context**

[Slack thread](https://cloud-native.slack.com/archives/C0150QF88FL/p1621928228041500)

In the above slack thread @anuraaga wrote:
```
The agent doesn’t use code from the app so both versions are supposed to have dependencies, while the -all is supposed to in 
addition include all exporters. If even the non -all had no dependencies included it wouldn’t work as a javaagent.
```

anuraaga:
Fixes #3082


javaagent-exporters/javaagent-exporters.gradle
```diff
@@ -0,0 +1,21 @@
+// Project to collect and shade exporter dependencies included in the agent's full distribution.
+
+plugins {
+  id "otel.shadow-conventions"
```
trask, line 4: this is so great

trask (approved):

iNikem (approved):

PR #3288, opened 2021-06-14
Disable default resource providers in tests.

anuraaga:
Adds spam to assertion failures / log messages with no benefit

### DIFF
--- a/javaagent-extension-api/src/test/groovy/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModuleTest.groovy
+++ b/javaagent-extension-api/src/test/groovy/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModuleTest.groovy
@@ -11,11 +11,6 @@ import spock.lang.Specification
 
 class InstrumentationModuleTest extends Specification {
 
-  def setup() {
-    assert System.getenv().findAll { it.key.startsWith("OTEL_") }.isEmpty()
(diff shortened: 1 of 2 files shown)


javaagent-extension-api/src/test/groovy/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModuleTest.groovy
```diff
@@ -11,11 +11,6 @@ import spock.lang.Specification
 
 class InstrumentationModuleTest extends Specification {
 
-  def setup() {
-    assert System.getenv().findAll { it.key.startsWith("OTEL_") }.isEmpty()
```
anuraaga, line 15: Couldn't figure out what these were here for

iNikem: Probably to make sure that we don't override default conf that we test below

iNikem (approved):

javaagent-extension-api/src/test/groovy/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModuleTest.groovy
anuraaga: Yeah I thought it could be a reason - I'm comfortable with deleting since I don't think we do this in other tests

trask (approved):

PR #3291, opened 2021-06-14
Make kotlin stdlib dependency compileOnly

anuraaga:
https://kotlinlang.org/docs/gradle.html#dependency-on-the-standard-library

### DIFF
--- a/instrumentation/kotlinx-coroutines/javaagent/gradle.properties
+++ b/instrumentation/kotlinx-coroutines/javaagent/gradle.properties
@@ -1,0 +1,1 @@
+kotlin.stdlib.default.dependency=false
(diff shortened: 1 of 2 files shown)


instrumentation/kotlinx-coroutines/javaagent/gradle.properties
```diff
@@ -0,0 +1 @@
+kotlin.stdlib.default.dependency=false
```
iNikem, line 1: Who uses this property?

anuraaga: kotlin gradle plugin, can check out from the link

iNikem (approved):

PR #3293, opened 2021-06-14
Fix ForkJoinPool sometimes not instrumented

### LINKED ISSUE #3292: ExecutorInstrumentationTest is flaky, appears to be sporadic propagation failure across ForkJoinPool
See https://scans.gradle.com/s/6flvnsy3ajxi4/tests/flaky

I was able to repro locally (sometimes) and add some printlns, and found that this line is never reached in the bad state:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/e9a8beb2316b0fe35f139fb249437ada96ab572c/instrumentation/executors/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/javaconcurrent/JavaExecutorInstrumentation.java#L78

There are a couple possibilities from here:

* `shouldAttachStateToTask` returned false above (not sure why)
* `ForkJoinPool` failed to get instrumented (maybe loaded too early?)

Having a hard time getting it to fail again though to get further details..

trask:
Resolves #3292

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
@@ -36,6 +36,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ForkJoinPool;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


anuraaga (approved):
Thanks!

iNikem (approved):

laurit (approved):

mateuszrzeszutek (approved):

PR #3295, opened 2021-06-14
Spring Integration javaagent instrumentation

mateuszrzeszutek:


instrumentation/spring/spring-integration-4.1/javaagent/spring-integration-4.1-javaagent.gradle
```diff
@@ -0,0 +1,50 @@
+apply from: "$rootDir/gradle/instrumentation.gradle"
+
+ext {
+  // context "leak" here is intentional: spring-integration instrumentation will always override
+  // "local" span context with one extracted from the incoming message
+  doNotFailOnContextLeak = true
+}
```
mateuszrzeszutek, line 7: This is a consequence of always using the context extracted from the incoming message; we completely ignore the "local" current context, but that's fine in that case.

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
```diff
@@ -0,0 +1,68 @@
...
+        }
+        span(4) {
+          name ~/testTopic.anonymous.[-\w]+ process/
+          childOf span(3)
+          kind CONSUMER
+        }
+        span(5) {
+          name "testConsumer.input"
+          childOf span(3)
+        }
```
mateuszrzeszutek, line 53: This test runs both RabbitMQ and spring-integration instrumentation - the interesting thing here is that the "consumer" span created by spring-integration instrumentation is the child of the actual PRODUCER span (and sibling of CONSUMER) - that's another consequence of choosing the context extracted from the message.

Removing the integration span would be even worse here, since now the parent of whatever comes next would be the PRODUCER span, not the CONSUMER one.

instrumentation/spring/spring-integration-4.1/testing/src/main/groovy/AbstractComplexPropagationTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.event.EventListener
+import org.springframework.integration.channel.DirectChannel
+import org.springframework.integration.channel.ExecutorChannel
+import org.springframework.messaging.Message
+import org.springframework.messaging.SubscribableChannel
+import org.springframework.messaging.support.MessageBuilder
+import spock.lang.Shared
+
+abstract class AbstractComplexPropagationTest extends InstrumentationSpecification {
```
mateuszrzeszutek, line 30: FYI: these tests haven't really changed, I just moved them to the `testing` module and made them abstract. I don't understand why git chose to show them as completely new if 95% of content is the same 🤷

gradle/instrumentation.gradle
```diff
@@ -100,7 +100,9 @@ afterEvaluate {
     jvmArgs "-javaagent:${project(":testing:agent-for-testing").tasks.shadowJar.archiveFile.get().asFile.absolutePath}"
     jvmArgs "-Dotel.javaagent.experimental.initializer.jar=${shadowJar.archiveFile.get().asFile.absolutePath}"
     jvmArgs "-Dotel.javaagent.testing.additional-library-ignores.enabled=false"
-    jvmArgs "-Dotel.javaagent.testing.fail-on-context-leak=true"
+    if (!findProperty('doNotFailOnContextLeak')) {
+      jvmArgs "-Dotel.javaagent.testing.fail-on-context-leak=true"
+    }
```
iNikem, line 105: Just inline this boolean into the value of system property?

instrumentation/spring/spring-integration-4.1/javaagent/spring-integration-4.1-javaagent.gradle
iNikem: Hmm... our `debugContextLeakIfEnabled` should not be called at all in case of `CONSUMER` spans. This is not spring integration specific problem.

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
iNikem: Which of the spans above is created by spring-integration? 4 or 5?

instrumentation/spring/spring-integration-4.1/javaagent/spring-integration-4.1-javaagent.gradle
mateuszrzeszutek: It's called for all spans, span kind does not matter.

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
mateuszrzeszutek: Number 5; number 4 (the CONSUMER one) is created by rabbit instrumentation.

instrumentation/spring/spring-integration-4.1/javaagent/spring-integration-4.1-javaagent.gradle
iNikem: Yes, but the logic of not failing the test should be the same for all `CONSUMER` spans.

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
iNikem: I don't like it... I am not sure I want to block this PR, but at some time in near future we have to review span modelling here...

mateuszrzeszutek: I have mixed feeling about it too - this situation was the reason why my first PR had initially used links for spans extracted from incoming messages.

There's one more idea that comes to my mind, a bit more complex behavior:
* if there's no "local" current span, use the span extracted from the message as the parent;
* otherwise use the "local" span as the parent and link the span context extracted from the message (if it's a different span than parent)

iNikem: We can discuss this tomorrow face to face, if you wish

iNikem (approved):

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
mateuszrzeszutek: Yep, definitely! 👍

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
```diff
@@ -0,0 +1,68 @@
...
+        span(2) {
+          name "exchange.declare"
+          childOf span(1)
+          kind CLIENT
+        }
+        span(3) {
+          name "testTopic -> testTopic send"
+          childOf span(1)
+          kind PRODUCER
+        }
```
trask, line 44: can you add attribute verification? I think that will help me understand the span structure a bit better

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
```diff
@@ -0,0 +1,78 @@
...
+        span(4) {
+          name ~/testTopic.anonymous.[-\w]+ process/
+          childOf span(3)
+          kind CONSUMER
+        }
+        span(5) {
+          name "testConsumer.input"
+          childOf span(3)
+          kind CONSUMER
+        }
```
mateuszrzeszutek, line 63: Two CONSUMER spans here are a necessity: turns out spring-cloud-stream for rabbitmq uses a listener that puts all incoming messages into a `BlockingQueue` immediately after receiving them; there's no way to propagate context through that... (BTW, this explains why `@RabbitListener` did not work for some people)

So when spring-integration instrumentation kicks in, it sees no CONSUMER span, so it creates a new one - a child of the remote PRODUCER span.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -89,10 +90,11 @@ public boolean shouldStart(Context parentContext, REQUEST request) {
     SpanKind spanKind = spanKindExtractor.extract(request);
     switch (spanKind) {
       case SERVER:
-        suppressed = ServerSpan.fromContextOrNull(parentContext) != null;
+      case CONSUMER:
+        suppressed = ServerSpan.exists(parentContext) || ConsumerSpan.exists(parentContext);
```
mateuszrzeszutek, line 94: SERVER and CONSUMER spans are now both treated as "local-root" type of spans; by default, you can't have both of them in a single process, in a single trace, since they both represent the receiving side.
This is a pretty significant change, please tell me WDYT.

iNikem (approved):

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+          attributes {
+            "${SemanticAttributes.MESSAGING_SYSTEM.key}" "rabbitmq"
+            "${SemanticAttributes.MESSAGING_DESTINATION.key}" "testTopic"
+            "${SemanticAttributes.MESSAGING_DESTINATION_KIND.key}" "queue"
+            "${SemanticAttributes.MESSAGING_OPERATION.key}" "process"
+            "${SemanticAttributes.MESSAGING_MESSAGE_PAYLOAD_SIZE_BYTES.key}" Long
+          }
+        }
+        span(5) {
+          name "testConsumer.input"
```
trask, line 86:
```suggestion
          name "testConsumer.input process"
```

trask (approved):

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+          attributes {
+            "${SemanticAttributes.NET_PEER_NAME.key}" { it == null || it == "localhost" }
+            "${SemanticAttributes.NET_PEER_IP.key}" "127.0.0.1"
+            "${SemanticAttributes.NET_PEER_PORT.key}" { it == null || it instanceof Long }
+            "${SemanticAttributes.MESSAGING_SYSTEM.key}" "rabbitmq"
+            "${SemanticAttributes.MESSAGING_DESTINATION_KIND.key}" "queue"
+          }
+        }
+        span(3) {
+          name "testTopic -> testTopic send"
```
trask, line 57: I think messaging semantic conventions suggest this should be just `testTopic send`

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+          kind SERVER
+          attributes {}
+        }
+        span(1) {
+          name "producer"
+          childOf span(0)
+          attributes {}
+        }
+        span(2) {
+          name "exchange.declare"
```
trask, line 45: can you comment in the test which spans are from spring integration instrumentation and which are from rabbitmq instrumentation?

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
mateuszrzeszutek: That span comes from rabbit instrumentation, do you mind if I create an issue for this?

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
mateuszrzeszutek: Good point, will do.

instrumentation/spring/spring-integration-4.1/javaagent/src/test/groovy/SpringIntegrationAndRabbitTest.groovy
iNikem: It seems that we should review rabbitmq integration's usage of semantic convention anyway. There are some other strange things here

mateuszrzeszutek: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3337

PR #3298, opened 2021-06-14
Rewrite global ignores matcher to IgnoredTypesConfigurer

mateuszrzeszutek:
Continuation of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3219

`IgnoreMatcherProvider` is going to disappear in the next PR by the way 😛


trask (approved):
👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ignore/IgnoredTypesMatcher.java
```diff
@@ -5,21 +5,66 @@
...
 package io.opentelemetry.javaagent.tooling.ignore;
 
+import io.opentelemetry.javaagent.spi.IgnoreMatcherProvider;
 import io.opentelemetry.javaagent.tooling.ignore.trie.Trie;
+import java.util.regex.Pattern;
 import net.bytebuddy.description.type.TypeDescription;
 import net.bytebuddy.matcher.ElementMatcher;
 
-class IgnoredTypesMatcher extends ElementMatcher.Junction.AbstractBase<TypeDescription> {
+public class IgnoredTypesMatcher extends ElementMatcher.Junction.AbstractBase<TypeDescription> {
```
anuraaga, line 14: IIUC, this was made public for access from the benchmark. Better to move the benchmark into the same package

mateuszrzeszutek: It's also used in the `AgentInstaller`. I thought that it'd be a bit more useful to expose the constructed trie from the builder, the instrumented classes check could use that one in one of the later PRs.

anuraaga (approved):

PR #3304, opened 2021-06-15
Directly initialize OkHttp without using test util.

anuraaga:
A generic HTTP client is too much API for us to expose in our testing util. The main job is base classes like `HttpServerTest`, `HttpClientTest`, some small utils like `findOpenPort` is probably OK, but users can bring in their favorite HTTP client instead of us providing it.

I'll move Armeria to an `implementation` dependency in a separate PR too.

### DIFF
--- a/examples/distro/instrumentation/servlet-3/build.gradle
+++ b/examples/distro/instrumentation/servlet-3/build.gradle
@@ -11,7 +11,7 @@ dependencies {
     exclude group: 'org.eclipse.jetty', module: 'jetty-server'
   }
 
-  testImplementation "com.squareup.okhttp3:okhttp:4.9.1"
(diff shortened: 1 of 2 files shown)


anuraaga:
Interestingly the smoke test already did this :)

trask (approved):

iNikem:
I am not sure I understand. `OkHttpUtils` is part of the published artifact `testing-common`. Why cannot external users (such as custom distros) use it?

anuraaga:
Yeah but is deleted now anyways - I don't think we should expose such API in general, low value for huge explosion in API surface. We haven't really put any time into making sure this testing-common has a low surface, high quality API but it's one step in that direction.

iNikem (approved):

mateuszrzeszutek (approved):

PR #3305, opened 2021-06-15
Move shaded testing dependencies to internal package.

anuraaga:
The frameworks we use for HTTP in testing is an implementation detail of our testing-common libraries and our repo's tests, they shouldn't be used directly by users. It's easy enough to bring in a favorite HTTP framework if a downstream repo needs it I think.

I tried moving the dependency from `api` to `implementation` at first but remembered Groovy doesn't let you compile a subclass without the parent's dependencies on the compile classpath (grr).


instrumentation/apache-camel-2.20/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/SingleServiceCamelTest.groovy
```diff
@@ -11,7 +11,7 @@ import io.opentelemetry.instrumentation.test.AgentInstrumentationSpecification
 import io.opentelemetry.instrumentation.test.RetryOnAddressAlreadyInUseTrait
 import io.opentelemetry.instrumentation.test.utils.PortUtils
 import io.opentelemetry.semconv.trace.attributes.SemanticAttributes
-import io.opentelemetry.testing.armeria.client.WebClient
+import io.opentelemetry.testing.internal.armeria.client.WebClient
```
trask, line 14: what's the purpose of moving these to `internal` package if they still have external usages? or is this a first step?

anuraaga: By external usage do you mean the usages in this repo? `internal` is fine to use in this repo, similar to the `internal` package being used across artifacts in the `opentelemetry-java` repo. The purpose is to make them unsupported for use from consumers of the `testing-common` package

trask (approved):

mateuszrzeszutek (approved):

PR #3307, opened 2021-06-15
Move min/max java properties to an extension class

anuraaga:
Currently, defining properties with `ext` has a large limitation in preventing the use of the `plugins` block for defining plugins used, which can allow Gradle to optimize the build better. We could change the way we handle `ext` to remove the ordering constraint, but it also still has the limitation of offering no type safety or discoverability. An extension class is the idiom in Gradle to allow configuring a plugin from projects so I've added an extension to start with modeling with min/max java versions.


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -35,14 +38,9 @@ if (applyCodeCoverage) {
...
 
-val minJavaVersionSupported = project.findProperty("minJavaVersionSupported")?.let(JavaVersion::toVersion)
-  ?: JavaVersion.VERSION_1_8
-val maxJavaVersionForTests = project.findProperty("maxJavaVersionForTests")?.let(JavaVersion::toVersion)
-
 java {
   toolchain {
-    languageVersion.set(JavaLanguageVersion.of(
-      Math.max(minJavaVersionSupported.majorVersion.toInt(), DEFAULT_JAVA_VERSION.majorVersion.toInt())))
+    languageVersion.set(otelJava.minJavaVersionSupported.map { JavaLanguageVersion.of(Math.max(it.majorVersion.toInt(), DEFAULT_JAVA_VERSION.majorVersion.toInt())) })
```
iNikem, line 43: Should this be encapsulated in a method in `OtelJavaExtension`?

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -53,19 +51,21 @@ java {
 
 tasks.withType<JavaCompile>().configureEach {
   with(options) {
-    release.set(minJavaVersionSupported.majorVersion.toInt())
+    release.set(otelJava.minJavaVersionSupported.map { it.majorVersion.toInt() })
```
iNikem, line 54: why `map` and not `get`?

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -53,19 +51,21 @@ java {
...
-tasks.withType<GroovyCompile>().configureEach {
-  sourceCompatibility = minJavaVersionSupported.majorVersion
-  targetCompatibility = minJavaVersionSupported.majorVersion
-}
-tasks.withType<ScalaCompile>().configureEach {
-  sourceCompatibility = minJavaVersionSupported.majorVersion
-  targetCompatibility = minJavaVersionSupported.majorVersion
+afterEvaluate {
+  tasks.withType<GroovyCompile>().configureEach {
+    sourceCompatibility = otelJava.minJavaVersionSupported.get().majorVersion
```
iNikem, line 62: Just a thought: can you use `map` here, to assign a provider to `sourceCompatibility` and remove `afterEvaluate`?

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -172,16 +172,16 @@ normalization {
 }
 
 fun isJavaVersionAllowed(version: JavaVersion): Boolean {
```
iNikem, line 174: Can this whole method be encapsulated into `OtelJavaExtension`?

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
anuraaga: Not in `afterEvaluate` - where possible we want to map without relying on evaluation order which results in less fragility in build scripts. Because of Gradle legacy it's unfortunately not always possible though so is a bit of a lost cause in some sense...

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
anuraaga: Unfortunately not, `sourceCompatibility` is old and is not a `Provider` (with kotlin you can easily tell the difference between provider and variable, there's no ability to use `=` with a provider). Hopefully someday

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
anuraaga: I think this is specifically the semantics of how we define the build toolchain so is appropriate here (can't come up with a name of a method that would be appropriate without directly defining this implementation detail)

buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
anuraaga: I guess it could - not sure there's a real reason to unless we can see ourselves calling this from the projects themselves. Here, it would just cause the below invocations to become more verbose, `otelJava.isJavaVersionAllowed` so don't think it's a win.

iNikem (approved):

mateuszrzeszutek (approved):

trask (approved):

PR #3310, opened 2021-06-15
Migrate spotless, checkstyle, codenarc to convention plugin

anuraaga:


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -250,3 +250,15 @@ afterEvaluate {
     }
   }
 }
+
+codenarc {
```
anuraaga, line 254: These seemed tiny enough to not extract into files. But happy to do so anyways if anyone prefers it.

trask (approved):

iNikem (approved):

PR #3330, opened 2021-06-16
Migrate jacoco.gradle to plugin

anuraaga:


buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts
```diff
@@ -27,17 +27,6 @@ afterEvaluate {
 // Version to use to compile code and run tests.
 val DEFAULT_JAVA_VERSION = JavaVersion.VERSION_11
 
-val applyCodeCoverage = !project.path.run {
```
anuraaga, line 30: Given the current state of affairs, I figured opt-in is simpler than opt-out?

mateuszrzeszutek (approved):

trask (approved):

PR #3338, opened 2021-06-16
Hide Config#create() method and use builder everywhere

mateuszrzeszutek:
`Config#create()` does not normalize property names, which could cause some confusion when writing tests. Now it was made private and `ConfigBuilder` is used everywhere to create `Config` instances.

Also added a `Config#getLongProperty()` that might be useful for vendor distros.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
@@ -176,6 +176,12 @@ public abstract class Config {
     return anyEnabled;
   }
 
+  /**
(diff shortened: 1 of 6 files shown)


iNikem (approved):

trask (approved):
Can we remove/deprecate `Config.asJavaProperties()` now that `getAllProperties()` is public?

mateuszrzeszutek:
> Can we remove/deprecate `Config.asJavaProperties()` now that `getAllProperties()` is public?

Unfortunately I can't remove it (it's still used in e.g. `SpanExporterFactory`, which is deprecated by the way), but it definitely should be deprecated 👍

Issue #2822, opened 2021-04-20
Auto-Instrumentation across Message Queue (RabbitMQ): traceId lost in subsequent spans


Johnny850807:
**Describe the bug**

Say, `A` passes a message to `B` via a message queue (RabbitMQ)
The message's consumer `B` invokes `C` and then `C` invokes `D` to handle the message. (`C` and `D` are annotated with `@WithSpan`)
Then I should see the trace has `A -> B -> C -> D`, instead I only see `A -> B`

(Furthermore, in `C` and `D`'s logs, the `trace_id` value is lost)

I can confirm that I can find the `traceparent` header included as the message's header, but
the auto-instrumentation doesn't use the same `traceparent` as the traceId for the subsequent spans (`C, D`)

Is this a feature or a bug?
How do I have `A --(message queue)--> B -> C -> D` traced so that I can see more clearly into asynchronous flows?

What version are you using?
---

opentelemetry-api: 1.0.0
opentelemetry-extension-annotations: 1.0.0
opentelemetry-javaagent: 1.0.0
org.springframework.boot: 2.4.3
spring-boot-starter-amqp: 2.3.2.RELEASE

Environment
---

Compiler/Runtime: AdoptOpenJDK 11.0.9
OS: macOS 10.15.5

mateuszrzeszutek:
Javaagent 0.14.0 only supports OTel API & annotations 0.14.0. Can you try using javaagent with the same version as the annotations extension?

Johnny850807:
> Javaagent 0.14.0 only supports OTel API & annotations 0.14.0. Can you try using javaagent with the same version as the annotations extension?

Sorry, it's Javaagent 1.0.0, I mistyped the version.

mateuszrzeszutek:
Can you share the code responsible for consuming the message (`B` and the call to `C` ideally)? Or, if it's feasible, create a small example app that reproduces this issue?

Johnny850807:
> Can you share the code responsible for consuming the message (`B` and the call to `C` ideally)? Or, if it's feasible, create a small example app that reproduces this issue?

Sorry for the late reply, here's the small app that can reproduce the issue.
https://github.com/Johnny850807/Spring-Boot-Demo-Observability-with-Open-Telemetry

It's a simple sign-up/sign-in demo app. After a sign-up request succeeds, the log shows:

[...]

Issue #2436, opened 2021-02-28
MQ Boundaries are not crossed, Redis call and MQ consumer not auto instrumented 


mnadeem:
**Describe the bug**

**Three things**

1) MQ (Rabbit MQ) calls are not crossing the component boundary as as in rest calls (Trace Ids are getting changed)

![image](https://user-images.githubusercontent.com/476145/109412537-95e57980-79ce-11eb-8248-a5418ff961d9.png)

2) Redis calls are not instrumented

![image](https://user-images.githubusercontent.com/476145/109413368-07272b80-79d3-11eb-86d4-fb028bd571c4.png)

3) RabbitMQ Consumer is not auto instrumented, I had to use  `@WithSpan`

![image](https://user-images.githubusercontent.com/476145/109414113-29bb4380-79d7-11eb-9312-82486f6066f4.png)

**Steps to reproduce**

https://github.com/mnadeem/boot-opentelemetry-tempo ==> look for **complex** branches, same behavior is across the branches.

````
mvn clean package docker:build
````
and then

````
docker-compose up
````

![image](https://user-images.githubusercontent.com/476145/109413416-481f4000-79d3-11eb-906c-bf66a6e70259.png)

**What did you expect to see?**
A clear and concise description of what you expected to see.

**What did you see instead?**
A clear and concise description of what you saw instead.

**What version are you using?**
(e.g., `v0.9.0`, `393e4a2`, etc) : 0.15.0, 0.16.0 and 0.17.0

**Environment**
Compiler: (e.g., "AdoptOpenJDK 11.0.6") : Oracle jdk 8
OS: (e.g., "Ubuntu 20.04") : Widows 10
Runtime (if different from JDK above): (e.g., "Oracle JRE 8u251")
OS (if different from OS compiled on): OS is same as that of compiled one, Widows 10

**Additional context**

mnadeem:
Note: Redis is instrumented with Lettuce 5 and not with 6

mnadeem:
For rabbitMQ consumer I doing the following way as a work around for the time being

https://github.com/mnadeem/boot-opentelemetry-tempo/blob/0.15.0_complex/boot-otel-tempo-mq-consumer/src/main/java/io/opentelemetry/example/flight/messaging/FlightMQConsumer.java#L38

mnadeem:
I see that,  RabbitMQ Client , 2.7+ is supported

And I am using 5.9

[...]

PR #3339, opened 2021-06-16
Make @RabbitListener propagate context properly

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2822
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2436

It turns out that the `BlockingQueueConsumer` class that I've discovered while working on the spring-integration instrumentation was the cause behind this. Similar to spring-integration, I've added another CONSUMER span that will now propagate its context forward to the actual user listener code.


instrumentation/spring/spring-amqp-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/amqp/SpringAmqpMessageAttributesExtractor.java
```diff
@@ -0,0 +1,74 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessageOperation;
+import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttributesExtractor;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.springframework.amqp.core.Message;
+
+final class SpringAmqpMessageAttributesExtractor
+    extends MessagingAttributesExtractor<Message, Void> {
+  @Override
+  protected String system(Message message) {
+    return "rabbitmq";
```
trask, line 17: maybe we should rename this module to `spring-amqp-rabbitmq`?

trask (approved):

laurit (approved):

instrumentation/spring/spring-amqp-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/amqp/SpringAmqpMessageAttributesExtractor.java
mateuszrzeszutek: Hmm, I'll rename it to `spring-rabbit` - this is the name of the library that actually contains `AbstractMessageListenerContainer`

tpgirishn: @mateuszrzeszutek I am still facing this issue. Trace Ids differ at RabbitListener annotated method. Btw I am using manual Otel instrumentation. Please guide me if I am missing anything obvious.

PR #3340, opened 2021-06-16
Fix shadow publishing of javaagent instrumentation

trask:
Previously only the `-all` artifact was being published:

```
  10456 Jun 16 11:45 opentelemetry-javaagent-google-http-client-1.19-1.3.0-alpha-SNAPSHOT-all.jar
   1497 Jun 16 11:45 opentelemetry-javaagent-google-http-client-1.19-1.3.0-alpha-SNAPSHOT.pom
```

I don't completely understand it, but after a bunch of trial and error this change seems to be working.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
@@ -8,7 +8,7 @@ plugins {
 publishing {
   publications {
     register<MavenPublication>("maven") {
-      if (tasks.names.contains("shadowJar") && findProperty("noShadowPublish") != "true") {
+      if (tasks.names.contains("shadowJar") && findProperty("noShadowPublish") != true) {
         the<ShadowExtension>().component(this)
       } else {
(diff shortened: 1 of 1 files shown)


trask:
Hm, it's still not quite right, opentelemetry-javaagent and opentelemetry-armeria-shaded-for-testing do not publish javadoc or sources artifact..

trask:
ok I reverted and went with small fix, let's see how it goes...

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -8,7 +8,7 @@ plugins {
 publishing {
   publications {
     register<MavenPublication>("maven") {
-      if (tasks.names.contains("shadowJar") && findProperty("noShadowPublish") != "true") {
```
anuraaga, line 11: Doh! Another motivation to move this and the other flags to the Java extension to make it typesafe

anuraaga (approved):

PR #3343, opened 2021-06-17
Remove testLatestDeps from release build.

anuraaga:
I noticed that testLatestDeps took forever in the release build probably for sporadic reasons. But while this may be ok in general, we actually wouldn't want a release to be blocked because of a latest deps failure that showed up on the release day.


laurit (approved):

mateuszrzeszutek (approved):

iNikem:
@trask may have different opinion: #3145

trask (approved):
> @trask may have different opinion: #3145

thanks for the ping, I'm good with this change since we are staying on top of the nightly build failures when `testLatestDeps` is failing

PR #3354, opened 2021-06-18
Extract build plugin for instrumentation testing projects that aren't…

anuraaga:
… muzzled / published.

Removes the `skipPublish` property, making testing projects a bit less verbose and reduces the amount of conditionals needed in the build logic.

### DIFF
--- a/buildSrc/src/main/kotlin/otel.javaagent-instrumentation.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.javaagent-instrumentation.gradle.kts
@@ -1,142 +1,23 @@
-import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
-import io.opentelemetry.instrumentation.gradle.bytebuddy.ByteBuddyPluginConfigurator
-
 plugins {
-  id("net.bytebuddy.byte-buddy")
-
-  id("otel.java-conventions")
-  id("otel.shadow-conventions")
+  id("otel.javaagent-testing")
+  id("otel.publish-conventions")
 }
+
+// TODO(anuraaga): Migrate MuzzlePlugin to kotlin script since Java plugins don't seem to be usable" +
+// with plugins block. It will be good to add a property for bootstrapRuntime to that plugin at the
+// time instead of defining it here and reading it in a non-obvious way.
+apply(plugin = "muzzle")
 
 extra["mavenGroupId"] = "io.opentelemetry.javaagent.instrumentation"
-// Shadow is only for testing, not publishing.
-extra["noShadowPublish"] = true
 
-val skipPublish: Boolean? by extra
+base.archivesName.set(projectDir.parentFile.name)
 
-if (skipPublish != true) {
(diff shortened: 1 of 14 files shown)


buildSrc/src/main/kotlin/otel.javaagent-instrumentation.gradle.kts
```diff
@@ -1,142 +1,23 @@
...
-    }
-    compileOnly("net.bytebuddy:byte-buddy")
-    annotationProcessor("com.google.auto.service:auto-service")
-    compileOnly("com.google.auto.service:auto-service")
-    compileOnly("org.slf4j:slf4j-api")
-
-    testImplementation("io.opentelemetry:opentelemetry-api")
-
-    testImplementation(project(":testing-common"))
-    testAnnotationProcessor("net.bytebuddy:byte-buddy")
```
anuraaga, line 65: Deleted this byte-buddy dependency while migrating since I couldn't find it being used

buildSrc/src/main/kotlin/otel.javaagent-testing.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+  compileOnly("net.bytebuddy:byte-buddy")
+  annotationProcessor("com.google.auto.service:auto-service")
+  compileOnly("com.google.auto.service:auto-service")
+  compileOnly("org.slf4j:slf4j-api")
+
+  testImplementation(project(":testing-common"))
+
+  testImplementation("org.testcontainers:testcontainers")
+
+  toolingRuntime(project(path = ":javaagent-tooling", configuration = "instrumentationMuzzle"))
```
anuraaga, line 42: I'm going to try to rework the bytebuddy / muzzle plugins to move this to the other plugin since they shouldn't be needed for tests (tests just need to muzzle the test source sets for generating helper class names)

buildSrc/src/main/kotlin/otel.javaagent-instrumentation.gradle.kts
```diff
@@ -1,142 +1,23 @@
...
+base.archivesName.set(projectDir.parentFile.name)
 
+// Used by muzzle
 val bootstrapRuntime by configurations.creating {
   isCanBeConsumed = false
   isCanBeResolved = true
 }
 
-afterEvaluate {
-  // TODO(anuraaga): There is no obvious reason this needs to be in afterEvaluate but it breaks
```
anuraaga, line 40: Was a fun one - `afterEvaluate` allowed testing-common -> armeria-shaded to be at the end of the classpath, so the accidentally bundled in Jackson was winning in priority

iNikem:
So how do we now tell if some module should not be published?

anuraaga:
We publish all our instrumentation libraries by convention anything with publish plugin applied (opt-in) and nothing else. There shouldn't be any need for opt out of these conventions right? (We haven't done it so far)

trask (approved):

PR #3359, opened 2021-06-18
Fix spring jms listener instrumentation

trask:


laurit (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ignore/AdditionalLibraryIgnoredTypesConfigurer.java
```diff
@@ -146,8 +156,8 @@ public void configure(Config config, IgnoredTypesBuilder builder) {
 
     builder
         .ignoreClass("org.springframework.jms.")
-        .ignoreClass("org.springframework.jms.listener.")
-        .ignoreClass(
+        .allowClass("org.springframework.jms.listener.")
+        .allowClass(
```
mateuszrzeszutek, line 160: Oops 🙈

mateuszrzeszutek (approved):

Issue #889, opened 2020-08-04
Fix all test failures on Windows


trask:
There are a few tests that fail on Windows (e.g. different socket related things). This is confusing for contributors who develop on Windows.

trask:
Ideally we would run github actions on both `ubuntu-latest` and `windows-latest`, but looks like we can't run linux docker containers on `windows-latest` (yet), which some of our tests require (https://github.com/actions/virtual-environments/issues/1143)

trask:
a reasonable option here may be to have a github action on `windows-latest` that runs all the tests that don't use testcontainers

trask:
Since there has been no activity on this enhancement for the past year we are closing it to help maintain our backlog. Anyone who would like to work on it is still welcome to do so, and we can re-open it at that time.

PR #3360, opened 2021-06-18
Fix tests on windows

trask:
Add updates the `build` github actions to run on both linux and windows.


anuraaga:
In my experience windows Github runner is extremely slow. Not sure if we want to run all the tests on it though we can try.

trask:
> In my experience windows Github runner is extremely slow. Not sure if we want to run all the tests on it though we can try.
> 
> We might want to opt-in to some important libraries instead of opt out.

It's more for windows users contributor experience than for test coverage. I fix the tests on windows from time to time, so hoping to find a way to keep them from continual drift. I'd be happy with adding it only to the nightly builds if it causes any issues for non-nightly builds.

trask:
oops, docker containers are all failing, went googling, found actions/virtual-environments#1143, and saw the linked issue from our repo #889 😅

laurit (approved):

mateuszrzeszutek (approved):

PR #3367, opened 2021-06-21
OpenTelemetry JDBC instrumentation library

donbeave:
Hello!

This is the first preview of JDBC instrumentation as a library, which started from this discussion: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/3309. It's not the final implementation yet, there are not any tests and documentation at the moment. I just want to confirm I'm going in the right direction. Most of this work is based on the code provided by https://github.com/opentracing-contrib/java-jdbc and utilizes the current JDBC javaagent implementation API classes, that's why I have moved them to the `library` module.

I have few questions here.

1) Maybe it's better to rename package for utilities classes, which now located in `library` module? My suggestion rename `io.opentelemetry.javaagent.instrumentation.jdbc` to `io.opentelemetry.instrumentation.jdbc`.
2) I didn't port `Slow Query` and `Fast Query` from the original OpenTracing Instrumentation for JDBC. Do I need to do it here or not? It looks like for me, if port this functionality, it must be ported to javaagent implementation as well. If so I think it will be better to use `io.opentelemetry.instrumentation.api.config.Config` to configure these features.
3) The current `otel.library.name` is `io.opentelemetry.javaagent.jdbc`. Maybe it's time to drop `javaagent` prefix and rename it to `io.opentelemetry.jdbc`. 

Please let me know if anything needs to be changed or improved here.


anuraaga:
Thanks @donbeave

> Maybe it's better to rename package for utilities classes, which now located in library module?

Yup nothing in that module should have `javaagent` in the package name. For any helpers that need to be accessed from both library and javaagent code, move them to an `internal` subpackage

> I didn't port Slow Query and Fast Query from the original OpenTracing Instrumentation for JDBC.

Let's avoid adding any new features, we'd probably want to go through opentelemetry specification for this if we want to

> Maybe it's time to drop javaagent prefix and rename it to io.opentelemetry.jdbc

Sounds good (@trask et al maybe we should rename all our instrumentation names that currently include javaagent, even if we don't have library instrumentation for them yet)

mateuszrzeszutek:
> > Maybe it's time to drop javaagent prefix and rename it to io.opentelemetry.jdbc
> 
> Sounds good (@trask et al maybe we should rename all our instrumentation names that currently include javaagent, even if we don't have library instrumentation for them yet)

👍 

There's issue https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1436 that's about that, wecan continue this discussion there.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,223 @@
...
+
+    final String realUrl = extractRealUrl(url);
+
+    // find the real driver for the URL
+    final Driver wrappedDriver = findDriver(realUrl);
+
+    final Connection connection = wrappedDriver.connect(realUrl, info);
+
+    final DbInfo dbInfo = JdbcConnectionUrlParser.parse(realUrl, info);
+    JdbcMaps.connectionInfo.put(connection, dbInfo);
```
mateuszrzeszutek, line 182: There should be no need to use `JdbcMaps`in a library instrumentation: you can just add a `DbInfo` field to `OpenTelemetryConnection`.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,223 @@
...
+    try {
+      Enumeration<Driver> enumeration = DriverManager.getDrivers();
+      List<Driver> drivers = new ArrayList<>();
+      for (int i = 0; enumeration.hasMoreElements(); ++i) {
+        Driver driver = enumeration.nextElement();
+        if (i == 0 && driver == INSTANCE) {
+          // the first driver is OTEL driver, skip all this verification
+          return;
+        }
+        if (driver instanceof OpenTelemetryDriver) {
```
mateuszrzeszutek, line 72: Shouldn't it be the opposite?
```suggestion
        if (!(driver instanceof OpenTelemetryDriver)) {
```

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,223 @@
...
+   * @throws IllegalStateException if the driver is already registered
+   * @throws SQLException if registering the driver fails
+   */
+  public static void register() throws SQLException {
+    if (isRegistered()) {
+      throw new IllegalStateException(
+          "Driver is already registered. It can only be registered once.");
+    }
+    DriverManager.registerDriver(INSTANCE);
+    OpenTelemetryDriver.registered = true;
```
mateuszrzeszutek, line 113: `registered` should probably be an `AtomicBoolean` - for thread correctness

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,223 @@
...
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Properties;
+import java.util.logging.Logger;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class OpenTelemetryDriver implements Driver {
```
mateuszrzeszutek, line 40: Consider making it final and adding a private constructor

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/OpenTelemetryPreparedStatement.java
```diff
@@ -0,0 +1,397 @@
...
+    // Connection#getMetaData() may execute a Statement or PreparedStatement to retrieve DB info
+    // this happens before the DB CLIENT span is started (and put in the current context), so this
+    // instrumentation runs again and the shouldStartSpan() check always returns true - and so on
+    // until we get a StackOverflowError
+    // using CallDepth prevents this, because this check happens before Connection#getMetadata()
+    // is called - the first recursive Statement call is just skipped and we do not create a span
+    // for it
+    if (CallDepthThreadLocalMap.getCallDepth(Statement.class).getAndIncrement() > 0) {
+      return callable.call();
+    }
```
mateuszrzeszutek, line 68: I wonder if this is needed for the library instrumentation: this bug only manifested itself when instrumented applications were creating JDBC connections directly (`new SomeDbConnection(...).getMetadata()`), without calling a driver. Since this instrumentation wraps over everything there may be no need for this code (and that'll help getting rid of the `javaagent-api` dependency which shouldn't be present in a library instrumentation)

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Yeah, fixed.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Tried. But it doesn't work with HikariCP:

```
Caused by: java.lang.RuntimeException: Failed to load driver class io.opentelemetry.instrumentation.jdbc.OpenTelemetryDriver in either of HikariConfig class loader or Thread context classloader
	at com.zaxxer.hikari.HikariConfig.setDriverClassName(HikariConfig.java:486)
	at io.micronaut.configuration.jdbc.hikari.$DatasourceConfigurationDefinition.injectBean(Unknown Source)
	at io.micronaut.configuration.jdbc.hikari.$DatasourceConfigurationDefinition.doBuild(Unknown Source)
	at io.micronaut.context.AbstractParametrizedBeanDefinition.build(AbstractParametrizedBeanDefinition.java:118)
	at io.micronaut.context.BeanDefinitionDelegate.build(BeanDefinitionDelegate.java:149)
	at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:1979)
	... 21 common frames omitted
```

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/OpenTelemetryPreparedStatement.java
donbeave: Removed it.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Fixed. Thank you!

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Sorry, not this one. Can make it final, but can not add a private constructor because of this exception:

```
Caused by: java.lang.RuntimeException: Failed to instantiate class io.opentelemetry.instrumentation.jdbc.OpenTelemetryDriver
	at com.zaxxer.hikari.HikariConfig.setDriverClassName(HikariConfig.java:494)
Caused by: java.lang.RuntimeException: Failed to instantiate class io.opentelemetry.instrumentation.jdbc.OpenTelemetryDriver

	at io.micronaut.configuration.jdbc.hikari.$DatasourceConfigurationDefinition.injectBean(Unknown Source)
	at io.micronaut.configuration.jdbc.hikari.$DatasourceConfigurationDefinition.doBuild(Unknown Source)
	at io.micronaut.context.AbstractParametrizedBeanDefinition.build(AbstractParametrizedBeanDefinition.java:118)
	at io.micronaut.context.BeanDefinitionDelegate.build(BeanDefinitionDelegate.java:149)
	at io.micronaut.context.DefaultBeanContext.doCreateBean(DefaultBeanContext.java:1979)
	... 21 common frames omitted
Caused by: java.lang.NoSuchMethodException: io.opentelemetry.instrumentation.jdbc.OpenTelemetryDriver.<init>()
	at java.base/java.lang.Class.getConstructor0(Class.java:3517)
Caused by: java.lang.NoSuchMethodException: io.opentelemetry.instrumentation.jdbc.OpenTelemetryDriver.<init>()

	at java.base/java.lang.Class.getConstructor(Class.java:2238)
	at com.zaxxer.hikari.HikariConfig.setDriverClassName(HikariConfig.java:490)
	... 26 common frames omitted
```

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Done.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
mateuszrzeszutek: Okay, let's leave it like it is then. Thanks for checking this!

donbeave:
This PR is ready for review. I have added some tests and small documentation.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,227 @@
...
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Logger;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public final class OpenTelemetryDriver implements Driver {
+
+  public static final OpenTelemetryDriver INSTANCE = new OpenTelemetryDriver();
```
anuraaga, line 42: Can this be package private?

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/datasource/OpenTelemetryDataSource.java
```diff
@@ -0,0 +1,124 @@
...
+import io.opentelemetry.instrumentation.jdbc.internal.DbInfo;
+import io.opentelemetry.instrumentation.jdbc.internal.OpenTelemetryConnection;
+import java.io.PrintWriter;
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.SQLFeatureNotSupportedException;
+import java.util.logging.Logger;
+import javax.sql.DataSource;
+
+public class OpenTelemetryDataSource implements DataSource, AutoCloseable {
```
anuraaga, line 39: Can this be package private and final?

anuraaga: Oh looks like it's another package, we could probably move it if it'll allow hiding it

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/CheckedCallable.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.jdbc.internal;
+
+@FunctionalInterface
+public interface CheckedCallable<T, E extends Throwable> {
```
anuraaga, line 9: Isn't normal Java 8 `Callable` checked (declares Exception in signature)?

anuraaga:
@donbeave FYI I merged in #3409 and renamed the files into this PR so probably want to git pull

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: I think so, it's public just for testing purposes, I'll change to package scope.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/datasource/OpenTelemetryDataSource.java
donbeave: Actually not, this one is another way to use OpenTelemetry. One way is by using the Driver and another way is to wrap the DataSource, same solution as to how P6Spy works: https://p6spy.readthedocs.io/en/latest/integration.html#datasource-way

I will update the README with covering this use case.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/CheckedCallable.java
donbeave: Actually, it's bad naming. In my case, I need a Callable with a generic error. If I change to the JVM's `Callable` I will need to wrap with additional try/catch in `wrapCall` method or in all other methods.

Example:

```java
  // change to use JVM's Callable instead of CheckedCallable
  private <T> T wrapCall(Callable<T> callable) throws Exception {
    Context parentContext = Context.current();

    if (!Span.fromContext(parentContext).getSpanContext().isValid()) {
      // this instrumentation is already very noisy, and calls to getConnection outside of an
      // existing trace do not tend to be very interesting
      return callable.call();
    }

    Context context = instrumenter().start(parentContext, delegate);
    T result;
      result = callable.call();
    } catch (Exception t) {
      instrumenter().end(context, delegate, null, t);
      throw t;
    }
    instrumenter().end(context, delegate, null, null);
    return result;
  }
```

As result, I will need to add try/catch to other methods which use `wrapCall`, but this catch will never executed:

CleanShot 2021-06-25 at 15 59 17@2x

Another way is to change `wrapCall` method to catch `Exception`:

```java
  private <T> T wrapCall(Callable<T> callable) throws SQLException {
    Context parentContext = Context.current();

    if (!Span.fromContext(parentContext).getSpanContext().isValid()) {
      // this instrumentation is already very noisy, and calls to getConnection outside of an
      // existing trace do not tend to be very interesting
      try {
        return callable.call();
      } catch (SQLException e) {
        throw e;
      } catch (Exception e) {
        // must never happens
        throw new SQLException(e);
      }
    }

    Context context = instrumenter().start(parentContext, delegate);
    T result;
      result = callable.call();
    } catch (Exception e) {
      instrumenter().end(context, delegate, null, e);
      if (e instanceof SQLException) {
        throw (SQLException) e;
      }
      // must never happens
      throw new SQLException(e);
    }
    instrumenter().end(context, delegate, null, null);
    return result;
  }
```

For me both of these approaches look a little bit ugly, that's why I chose the `CheckedCallable` way.

But I don't mind to change if you think so.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: > @donbeave FYI I merged in #3409 and renamed the files into this PR so probably want to git pull

Thank you! I'll do it now.

instrumentation/jdbc/library/jdbc-library.gradle
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+apply plugin: "otel.library-instrumentation"
```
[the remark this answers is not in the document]
donbeave: Done.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Done. Changed to package scope.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/CheckedCallable.java
mateuszrzeszutek: Nitpicking: you can rename it to [`ThrowingSupplier`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/util/ThrowingSupplier.java)

donbeave: oh, @mateuszrzeszutek, big thanks for this!

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/datasource/OpenTelemetryDataSource.java
donbeave: Readme updated.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/CheckedCallable.java
donbeave: Done.

instrumentation/jdbc/library/src/test/groovy/io/opentelemetry/instrumentation/jdbc/TestConnection.groovy
```diff
@@ -0,0 +1,281 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.jdbc
+
+import java.sql.*
+import java.util.concurrent.Executor
+
+class TestConnection implements Connection {
```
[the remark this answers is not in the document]
donbeave: Agree, will do it.

instrumentation/jdbc/library/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+plugins {
+  id("otel.library-instrumentation")
+}
+
+dependencies {
+  implementation project(':instrumentation-api')
+  implementation project(':javaagent-api')
```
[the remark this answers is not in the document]
donbeave: Ok, sorry I forgot to do it.

donbeave:
@mateuszrzeszutek thank you for improving documentation!

mateuszrzeszutek:
You're welcome! Glad to have been of help 😄

instrumentation/jdbc/library/build.gradle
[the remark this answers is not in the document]
donbeave: Done.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,236 @@
...
+      throw new ExceptionInInitializerError(e);
+    }
+  }
+
+  /**
+   * Ensure {@code TracingDriver} be the first driver of {@link DriverManager} to make sure
+   * "interceptor mode" works. WARNING: Driver like Oracle JDBC may fail since it's destroyed
+   * forever after deregistration.
+   */
+  public static synchronized void ensureRegisteredAsTheFirstDriver() {
```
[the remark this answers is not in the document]
donbeave: To prevent other simultaneous calls. Because normally driver will be registered during static initialization.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,236 @@
...
+    REGISTERED.set(false);
+  }
+
+  /** Returns {@code true} if the driver is registered against {@link DriverManager}. */
+  public static boolean isRegistered() {
+    return REGISTERED.get();
+  }
+
+  private static Driver findDriver(String realUrl) {
+    if (realUrl == null || realUrl.trim().length() == 0) {
```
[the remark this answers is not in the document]
donbeave: You are right, removed.

instrumentation/jdbc/library/src/test/groovy/io/opentelemetry/instrumentation/jdbc/TestConnection.groovy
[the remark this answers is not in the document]
donbeave: Done.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/PeerServiceAttributesExtractor.java
```diff
@@ -3,11 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.instrumentation.api.instrumenter;
+package io.opentelemetry.instrumentation.api.instrumenter;
```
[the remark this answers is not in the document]
donbeave: Done.

donbeave:
@marcingrzejszczak @anuraaga please let me know if anything else needs to be improved or fixed. Btw, there are two build steps are failed `accept-pr` and `muzzle (:instrumentation:spring:spring-rabbit-1.0:javaagent)`, but both seems not related to my changes:

1)
CleanShot 2021-06-28 at 13 10 38@2x

2) 
CleanShot 2021-06-28 at 13 12 09@2x

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,230 @@
...
+  private static String extractRealUrl(String url) {
+    return url.startsWith(INTERCEPTOR_MODE_URL_PREFIX)
+        ? url.replace(INTERCEPTOR_MODE_URL_PREFIX, "jdbc:")
+        : url;
+  }
+
+  @Nullable
+  @Override
+  public Connection connect(String url, Properties info) throws SQLException {
+    if (url == null || url.trim().length() == 0) {
```
anuraaga, line 168:
```suggestion
    if (url == null || url.trim().isEmpty()) {
```

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,230 @@
...
+    }
+    if (url.startsWith(INTERCEPTOR_MODE_URL_PREFIX)) {
+      return true;
+    }
+    return interceptorMode && url.startsWith("jdbc:") && url.length() > 5;
+  }
+
+  @Override
+  public DriverPropertyInfo[] getPropertyInfo(String url, Properties info) throws SQLException {
+    if (url == null || url.trim().length() == 0) {
```
anuraaga, line 201:
```suggestion
    if (url == null || url.trim().isEmpty()) {
```

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -0,0 +1,230 @@
...
+      throw new IllegalArgumentException("url is required");
+    }
+
+    final String realUrl = extractRealUrl(url);
+    final Driver wrappedDriver = findDriver(realUrl);
+    return wrappedDriver.getPropertyInfo(realUrl, info);
+  }
+
+  @Override
+  public int getMajorVersion() {
```
anuraaga, line 211: I guess these are supposed to match the version we released. Can you parse these out of `InstrumentationVersion.VERSION`?

anuraaga (commented):
Just found one point about the driver version we report which I think is blocking, but other then that looks great

trask:
> there are two build steps are failed `accept-pr` and `muzzle (:instrumentation:spring:spring-rabbit-1.0:javaagent)`, but both seems not related to my changes

👍 I opened #3417 to track/fix the muzzle failure. `accept-pr` is an aggregating build step, so fails when any other step fails. I'll hit retry on the build, good chance it will pick up different set of spring-rabbit versions to check.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
donbeave: Yeah, of course.

donbeave: Done.

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/OpenTelemetryDriver.java
```diff
@@ -162,6 +163,22 @@ private static String extractRealUrl(String url) {
         : url;
   }
 
+  private static int[] parseInstrumentationVersion() {
+    String[] parts = InstrumentationVersion.VERSION.split("\\.");
+    if (parts.length >= 2) {
+      try {
+        int majorVersion = Integer.parseInt(parts[0]);
```
anuraaga, line 170: It may not be a big deal but it would be nice if you can parse into private static final fields `MAJOR_VERSION` `MINOR_VERSION` in a static initializer block instead of every time the methods are called

donbeave: Done. Please check.

anuraaga (approved):
Thanks!

anuraaga:
@donbeave I've merged in #3414 to this branch so you may need to `git pull` if any other changes are needed

donbeave:
@anuraaga Got it! Thank you!

mateuszrzeszutek (approved):

instrumentation/jdbc/library/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Calling `OpenTelemetryDriver.ensureRegisteredAsTheFirstDriver()` method along with
+`setInterceptorMode(true)` will fix that problem.
+
+Please note drivers like Oracle JDBC may fail since they're destroyed forever after deregistration.
+
+### Datasource way
+
+If your application uses a DataSource, simply wrap your current DataSource object with
+`OpenTelemetryDataSource`. `OpenTelemetryDataSource` has a constructor method that accepts the
```
trask, line 95: Let's put this option first, since it seems preferred

instrumentation/jdbc/library/README.md
```diff
@@ -0,0 +1,120 @@
...
+}
+```
+
+Sometimes it's also required to register the OpenTelemetry driver as a first driver in the
+DriverManager list, otherwise another driver can be selected at the time of initialization.
+
+Calling `OpenTelemetryDriver.ensureRegisteredAsTheFirstDriver()` method along with
+`setInterceptorMode(true)` will fix that problem.
+
+Please note drivers like Oracle JDBC may fail since they're destroyed forever after deregistration.
```
trask, line 90: This option worries me from support perspective. Do we think it's a needed option? Or can we start without it and see?

instrumentation/jdbc/library/README.md
donbeave: Ok, let me do it.

instrumentation/jdbc/library/README.md
donbeave: Let's remove it and the whole idea of interceptor mode. I think setting driver with `jdbc:otel:...` prefixed URL is a much better and more clear solution. So let's remove it?

instrumentation/jdbc/library/README.md
donbeave: Done.

instrumentation/jdbc/library/README.md
donbeave: Done.

mateuszrzeszutek (approved):

trask:
thanks @donbeave! this is a great addition! ❤️

PR #3374, opened 2021-06-21
support logs in the fake test backend

breedx-splk:
Interest in logs is heating up. I think we might find it helpful to have this wired up.

I just followed the pattern that existed for logs and metrics.


smoke-tests/fake-backend/src/main/java/io/opentelemetry/smoketest/fakebackend/FakeBackendMain.java
```diff
@@ -69,25 +70,38 @@ public void serialize(
           }
         });
     module.setSerializers(serializers);
+    serializers.addSerializer(
```
anuraaga, line 73: Extreme nit, looks a bit nicer to do this before `setSerializers`

anuraaga (approved):

smoke-tests/fake-backend/src/main/java/io/opentelemetry/smoketest/fakebackend/FakeBackendMain.java
breedx-splk: Whoops, thanks, this was a mistake to have it in there twice.  Copypasta.  Will fix.  🎈

trask (approved):

mateuszrzeszutek (approved):

PR #3375, opened 2021-06-22
Remove extra spotless config

trask:
I think the spotless config in `buildSrc/build.gradle.kts` is not needed since we have `otel.spotless-conventions.gradle.kts`.

`spotless-groovy.properties` seems unused(?)


mateuszrzeszutek (approved):

buildSrc/build.gradle.kts
```diff
@@ -5,14 +5,6 @@ plugins {
   id("com.diffplug.spotless") version "5.13.0"
 }
 
-spotless {
```
anuraaga, line 8: `buildSrc` is esssentially a completely independent project, so it doesn't actually get to use anything defined here. It's why we even need to keep the spotless versions in sync between the `plugins` block above (used by buildSrc) and the `dependencies` block below (used by the main build) :(

mateuszrzeszutek (approved):

anuraaga (approved):

PR #3378, opened 2021-06-22
Fix PeerServiceAttributesExtractor#createUsingReflection()

mateuszrzeszutek:
`PeerServiceAttributesExtractor` is located in the bootstrap classloader, all `NetAttributesExtractor` implementations are in the application classloaders - we need to pass the application classloader (from any helper class that was already injected) to make loading extractors correctly possible.

@anuraaga WDYT about moving `NetAttributeExtractor` library implementations to their respective `internal` packages and just using those classes normally, without this hack?


anuraaga (commented):
Heh yeah `internal` package seems like a good idea

anuraaga (approved):

trask (approved):

laurit (approved):

PR #3380, opened 2021-06-22
Ignore task classes using IgnoredTypesConfigurer

mateuszrzeszutek:
I also tried to split the global config a bit and introduced several lib-specific `IgnoredTypesConfigurer`s that ignore unwanted task classes.


trask (approved):

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/InstrumentedTaskClasses.java
```diff
@@ -0,0 +1,64 @@
...
+            return false;
+          }
+          return true;
+        }
+      };
+
+  private static volatile Trie<Boolean> ignoredTaskClasses;
+
+  /**
+   * Sets the configured ignored tasks trie. This method is called internally from the agent
```
anuraaga, line 44: Isn't this whole class / mechanism basically internal?

anuraaga (approved):

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/concurrent/InstrumentedTaskClasses.java
mateuszrzeszutek: Hmm, that's true, I guess that it's very similar to `BootstrapPackagePrefixesHolder` in that regard. I'll move it to the `internal` package. Thanks!

PR #3385, opened 2021-06-23
Migrate muzzle data classes to kotlin.

anuraaga:
When migrating from Groovy to Java I expected this would eventually come anyways :) These are almost fully automated conversions by IntelliJ, just hit "Reformat Code" after, changed e.g. `emptySet()` to `setOf()` and moved `init` to top as per kotlin conventions https://kotlinlang.org/docs/coding-conventions.html#class-layout. At first I moved `companion` to top too but then found that page. Also switched to Kotlin's Regex helper


anuraaga:
image

Reduction but not massive, I think generally Java still holds up ok :-)

trask (approved):

mateuszrzeszutek (approved):

PR #3386, opened 2021-06-23
Use plugins block everywhere for applying plugins

anuraaga:


instrumentation/lettuce/lettuce-common/library/lettuce-common-library.gradle
```diff
@@ -1,9 +1,3 @@
-// not applying $rootDir/gradle/instrumentation.gradle because that brings running tests with agent
-// infrastructure, and this module only wants to run unit tests
```
trask, line 2: this isn't a problem pulling in `otel.javaagent-instrumentation` below?

trask (approved):

instrumentation/lettuce/lettuce-common/library/lettuce-common-library.gradle
anuraaga: Don't think so - we have the same pattern in some other `-common` files, it'll slow down the test a bit but not a big deal

anuraaga: Ah but realized that lettuce-5.1 library instrumentation depends on this so it's supposed to have been library instrumentation anyways

mateuszrzeszutek (approved):

PR #3388, opened 2021-06-23
Migrate muzzle plugin to kotlin script.

anuraaga:


buildSrc/src/main/kotlin/otel.muzzle-check.gradle.kts
```diff
@@ -0,0 +1,351 @@
...
+plugins {
+  `java-library`
+}
+
+// Select a random set of versions to test
+val RANGE_COUNT_LIMIT = 10
+
+val muzzleConfig = extensions.create<MuzzleExtension>("muzzle")
+
+val muzzleTooling by configurations.creating {
```
anuraaga, line 37: Creating the configurations in the plugin instead of the project like before as plugins normally do. I haven't quite been able to reason out the difference between these configurations yet (intuitively I'd expect us to only need one classpath when muzzling) but have reproduced the split for now.

mateuszrzeszutek (approved):

buildSrc/src/main/kotlin/otel.muzzle-check.gradle.kts
mateuszrzeszutek: I think one of them (`muzzleTooling`) is supposed to be used as the agent classpath source (to be able to load `InstrumentationModule`), the other (`muzzleBootstrap`) should be a base for the "instrumented application" classpath - add the library jar for a given version to our bootstrap classes and pass it to the check method.

buildSrc/src/main/kotlin/otel.muzzle-check.gradle.kts
```diff
@@ -0,0 +1,351 @@
...
+
+  runAfter.configure { finalizedBy(muzzleTask) }
+  return muzzleTask
+}
+
+fun createClassLoaderForTask(muzzleTaskConfiguration: Configuration): ClassLoader {
+  val userUrls = objects.fileCollection()
+  logger.info("Creating task classpath")
+  userUrls.from(muzzleTaskConfiguration.resolvedConfiguration.files)
+  return classpathLoader(userUrls.plus(muzzleTooling), ClassLoader.getPlatformClassLoader())
```
mateuszrzeszutek, line 264: Shouldn't that be the bootstrap conf?
```suggestion
  return classpathLoader(userUrls.plus(muzzleBootstrap), ClassLoader.getPlatformClassLoader())
```

buildSrc/src/main/kotlin/otel.muzzle-check.gradle.kts
anuraaga: Ah thanks that makes sense!

buildSrc/src/main/kotlin/otel.muzzle-check.gradle.kts
anuraaga: Oops - guess it still passed since bootstrap gets included transitively

trask (approved):

PR #3395, opened 2021-06-23
Remove unnecessary and complex logic

### LINKED ISSUE #3322: Javaagent interferes with custom log manager on J9 Java 8 JVM
Reported in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3316#issuecomment-861568334.

Also need to review the comment below for accuracy:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/c7ac235614b0a1c4bb8431041dab9a8efda9779e/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java#L86-L91

trask:
Related to #3322

Based on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3316#issuecomment-862087497:

> I have never seen a problem with getting the agent jar from the code source: https://github.com/glowroot/glowroot/blob/main/agent/core/src/main/java/org/glowroot/agent/AgentPremain.java.
>
> Doesn't look like NewRelic has either: https://github.com/newrelic/newrelic-java-agent/blob/main/newrelic-agent/src/main/java/com/newrelic/bootstrap/BootstrapAgent.java

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/OpenTelemetryAgent.java
@@ -9,16 +9,10 @@ import io.opentelemetry.javaagent.bootstrap.AgentInitializer;
 import java.io.File;
 import java.io.IOException;
 import java.lang.instrument.Instrumentation;
-import java.lang.management.ManagementFactory;
-import java.lang.reflect.Field;
 import java.net.URISyntaxException;
 import java.security.CodeSource;
-import java.util.Arrays;
-import java.util.List;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
-import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

mateuszrzeszutek (approved):

laurit (approved):

laurit:
I'm not sure whether this will really resolve https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3322 as getVmArgumentsThroughReflection would only be run when getting agent location from code source fails (at least I can't see why it should). I wasn't able to repro https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3322 https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3316 mentions that it should fail on Adopt J9 9 which is not available for mac, it also mentions IBM J9 7 which must be a typo?

PR #3401, opened 2021-06-24
Migrate codegen to be a plugin

anuraaga:


buildSrc/src/main/kotlin/otel.javaagent-testing.gradle.kts
```diff
@@ -39,15 +34,10 @@ dependencies {
 
   testImplementation("org.testcontainers:testcontainers")
 
-  toolingRuntime(project(path = ":javaagent-tooling", configuration = "instrumentationMuzzle"))
-  toolingRuntime(project(path = ":javaagent-extension-api", configuration = "instrumentationMuzzle"))
+  add("codegen", project(path = ":javaagent-tooling", configuration = "instrumentationMuzzle"))
```
mateuszrzeszutek, line 37: Idea for future PRs: if we moved `MuzzleCodeGenerationPlugin` and its related classes to gradle plugin code then we'd be able to get rid of the tooling dependency. Ideally we should only depend on javaagent-extension-api

mateuszrzeszutek (approved):

PR #3414, opened 2021-06-25
Migrate instrumentation gradle files to kotlin

anuraaga:
Mostly regex and then fixing about 100 issues manually.

Was hoping that having a commit with just the renames would help GitHub render side by side but seems like it didn't :( So looks like directly looking at the Convert commit is the way to go

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3414/commits/5c2ef766300d6217767725160011b7d36eb5949b


instrumentation/aws-sdk/aws-sdk-1.11/javaagent/build.gradle.kts
```diff
@@ -26,79 +26,91 @@ testSets {
   // in 1.11.106 than 1.11.84.
   // We test older version in separate test set to test newer version and latest deps in the 'default'
   // test dir. Otherwise we get strange warnings in Idea.
-  test_before_1_11_106 {
-    filter {
```
anuraaga: These blocks move to Tests block since apparently groovy was resolving the `filter` method on the test task named `test_before_1_11_106` despite being nested in a configuration named `test_before_1_11_106` :O

instrumentation/aws-sdk/aws-sdk-1.11/javaagent/build.gradle.kts
```diff
@@ -26,79 +26,91 @@ testSets {
...
-  // TODO run tests both with and without experimental span attributes
-  jvmArgs "-Dotel.instrumentation.aws-sdk.experimental-span-attributes=true"
+  if (!(findProperty("testLatestDeps") as Boolean)) {
+    named("check") {
+      dependsOn(test_before_1_11_106)
+      dependsOn(testSqs)
+    }
+  }
+
+  named<Test>("test") {
```
anuraaga, line 108: Actually `test.configure {` at the top level is equivalent to this and arguably less verbose, but I figured consistency with other tests, which don't get such DSL, is simpler

instrumentation/build.gradle.kts
```diff
@@ -1,58 +0,0 @@
-// this project will run in isolation under the agent's classloader
```
anuraaga: I had accidentally renamed this without the intention of migrating and renamed back in this commit so looks like a deletion when looking at commits

anuraaga (commented):
Let me know if it helps to split this up. I haven't done it yet since it requires some effort and I wasn't sure it helps the review that much given the files are all unrelated to each other, so just line-by-line comparison of either 1 PR or multiple.

trask (approved):
omg that's a lot of gradle files 😅

PR #3425, opened 2021-06-28
Fixed Semantic conventions in RabbitMQ

### LINKED ISSUE #3337: Review RabbitMQ instrumentation's semantic conventions usage
RabbitMQ does not seem to implement messaging semantic conventions correctly, e.g. it uses `testTopic -> testTopic send` instead of `testTopic send` as span names.

**Additional context**
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3295#discussion_r652381304

RashmiRam:
1. Constructed producer span name according to spec rule. (Removed routing key as part of span name)
2. Renamed routing key attribute(part of experiment span attributes feature) according to semantic convention

Fixes #3337

### DIFF
--- a/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitTracer.java
+++ b/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitTracer.java
@@ -65,11 +65,9 @@ public class RabbitTracer extends BaseTracer {
       spanBuilder.setAttribute(
           SemanticAttributes.MESSAGING_DESTINATION,
           normalizeExchangeName(response.getEnvelope().getExchange()));
-      if (CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES) {
-        spanBuilder.setAttribute(
-            SemanticAttributes.MESSAGING_RABBITMQ_ROUTING_KEY,
-            response.getEnvelope().getRoutingKey());
-      }
+      spanBuilder.setAttribute(
+          SemanticAttributes.MESSAGING_RABBITMQ_ROUTING_KEY,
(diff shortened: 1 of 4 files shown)


RashmiRam:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3337

instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitTracer.java
```diff
@@ -66,7 +66,9 @@ public Context startGetSpan(
           SemanticAttributes.MESSAGING_DESTINATION,
           normalizeExchangeName(response.getEnvelope().getExchange()));
       if (CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES) {
-        spanBuilder.setAttribute("rabbitmq.routing_key", response.getEnvelope().getRoutingKey());
+        spanBuilder.setAttribute(
+            SemanticAttributes.MESSAGING_RABBITMQ_ROUTING_KEY,
```
iNikem, line 70: If this attribute is already a part of the "official" semantic convention, then it is not experimental anymore and should be moved outside this `if`.

iNikem (approved):

instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitTracer.java
RashmiRam: Done.

laurit (approved):

mateuszrzeszutek (approved):

iNikem (approved):

PR #3426, opened 2021-06-29
Remove noShadowPublish

anuraaga:
`noShadowPublish` wasn't being set. Actually, I realized that shadowJar with the default seems to be part of `components["java"]` anyways, I didn't know that. So there wasn't a huge point in having the branch. This brings the publishing back to where instrumentation libraries publish both `all` and not all, which is better than now. I'll try to see how to get rid of the `all`, it's not obvious from the documentation but I'm guessing any shadowJar with a configuration including `runtimeClasspath` ends up in there strangely enough.

Checked armeria-javaagent, javaagent, instrumentation-api-caching, armeria-for-testing and all seemed to publish what we expect

### DIFF
--- a/instrumentation-api-caching/build.gradle
+++ b/instrumentation-api-caching/build.gradle
@@ -41,3 +41,12 @@ jar {
 
   dependsOn shadowJar
 }
+
+// Because shadow does not use default configurations
+publishing {
+  publications {
+    maven {
+      project.shadow.component(it)
+    }
+  }
+}
--- a/javaagent/build.gradle
+++ b/javaagent/build.gradle
@@ -53,6 +53,7 @@ CopySpec isolateSpec(Collection<Project> projectsWithShadowJar) {
 
 //Includes everything needed for OOTB experience
 shadowJar {
+  archiveClassifier.set("all")
   def projectsWithShadowJar = [project(':instrumentation'), project(":javaagent-exporters")]
   projectsWithShadowJar.each {
     dependsOn("${it.path}:shadowJar")
@@ -62,10 +63,14 @@ shadowJar {
 
 //Includes instrumentations, but not exporters
 task lightShadow(type: ShadowJar) {
+  archiveClassifier.set("")
   dependsOn ':instrumentation:shadowJar'
   def projectsWithShadowJar = [project(':instrumentation')]
   with isolateSpec(projectsWithShadowJar)
 }
+
+// lightShadow is the default classifier we publish so disable the default jar.
+jar.enabled = false
 
 publishing {
   publications {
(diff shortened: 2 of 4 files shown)


instrumentation-api-caching/build.gradle
```diff
@@ -41,3 +41,11 @@ jar {
 
   dependsOn shadowJar
 }
+
+publishing {
```
anuraaga, line 45: Here `configurations` doesn't include runtimeclasspath which is why it's not in `java["components]` (I guess...)

javaagent/build.gradle
```diff
@@ -53,6 +53,7 @@ CopySpec isolateSpec(Collection<Project> projectsWithShadowJar) {
 
 //Includes everything needed for OOTB experience
 shadowJar {
+  archiveClassifier.set("all")
```
anuraaga, line 56: No-op, just to document more clearly since was confusing to me

javaagent/build.gradle
```diff
@@ -62,11 +63,15 @@ shadowJar {
...
 //Includes instrumentations, but not exporters
 task lightShadow(type: ShadowJar) {
+  archiveClassifier.set("")
   dependsOn ':instrumentation:shadowJar'
   def projectsWithShadowJar = [project(':instrumentation')]
   with isolateSpec(projectsWithShadowJar)
 }
 
+// lightShadow is the default classifier we publish so disable the default jar.
+jar.enabled = false
```
anuraaga, line 73: I have a feeling there was some luck involved in not overwriting lightShadow with this default jar and publishing it. By removing the `noShadowPublish` branch it became disallowed to have two tasks publishing the same file.

instrumentation-api-caching/build.gradle
iNikem: This is needed here because we disable `jar` task.

I don't exactly like how this module is now published/used. I mean when `instrumentation-api` module declares dependency on `api project(":instrumentation-api-caching")` what exactly does it want? Just jar (which is disabled) or full shadow jar? If the latter we should do this more explicit.

iNikem (approved):
I am Ok with this patch, but we have to review our big picture. When reviewing this PR I was reminded that I don't quite understand the purpose of some of our modules, why we sometimes use shadow plugin the way we do etc etc

instrumentation-api-caching/build.gradle
anuraaga: Yeah I think we could consider the technique we found for jctools in SDK

https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace-shaded-deps/build.gradle.kts
https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace/build.gradle.kts#L73

We don't publish the shadowed thing separately but embed it into the SDK's normal JAR, so here we could embed into instrumentation-api.

PR #3430, opened 2021-06-29
Exclude duplicates from -all.jar

### LINKED ISSUE #3405: the javaagent-all.jar file contains duplicate files
**Describe the bug**
There are multiple files with the exact same name in the opentelemetry-javaagent-*-all.jar file. This doesn't seem to have any negative effect other than enlarging the jar file distributable. None of the class files seems to have different size nor crc.

**Steps to reproduce**

Run the following to get a count and a list of the files.
```bash
unzip -v otel-all.jar | cut -c59- | sort | uniq -d
```

Empty result because every file should only be included once.

**What did you see instead?**
389 duplicate files, ie files that are added to the jar twice. 

`1.3.1`

**Environment**
Not relevant

**Additional context**

anuraaga:
Fixes #3405

I tried excluding duplicate dependencies from the javaagent-exporters project but found it quite complicated - in particular, while it's easy to exclude the declared dependencies (opentelemetry-sdk), excluding transitives (opentelemetry-sdk-trace) requires resolving configurations and adding lots of hairyness. This one line seems fine instead, since we apply dependency management I think it's unlikely we'd attempt to package different versions and smoke tests should catch anything fishy that could result from ignoring dupes.

### DIFF
--- a/javaagent/build.gradle
+++ b/javaagent/build.gradle
@@ -59,6 +59,7 @@ shadowJar {
     dependsOn("${it.path}:shadowJar")
   }
   with isolateSpec(projectsWithShadowJar)
+  duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 }
 
 //Includes instrumentations, but not exporters


mateuszrzeszutek (approved):

mateuszrzeszutek:
Just wondering: did you run the bash command posted in that issue? Are there no duplicates left after this change?

anuraaga:
@mateuszrzeszutek Yeah confirmed no duplicates with the change

PR #3431, opened 2021-06-29
Simplify jflex configuration

anuraaga:
JFlex plugin automatically sets itself up. The problem was that it requires files to have extension `.jflex`

https://github.com/jprante/gradle-plugin-jflex/blob/master/src/main/groovy/org/xbib/gradle/plugin/JFlexPlugin.groovy#L42
https://github.com/jprante/gradle-plugin-jflex/blob/master/src/main/groovy/org/xbib/gradle/plugin/JFlexPlugin.groovy#L72


instrumentation-api/build.gradle
```diff
@@ -1,34 +1,22 @@
...
-}
-
-tasks.named("sourcesJar").configure {
-  dependsOn(sqlSanitizerJflex)
+sourceSets {
+  main {
+    java {
+      // gradle-jflex-plugin has a bug in that it always looks for the last srcDir in this source
+      // set to generate into. By default it would be the src/main directory itself.
+      srcDir("${buildDir}/generated/sources/jflex")
```
anuraaga, line 14: I went ahead and changed the directory a bit to match what gradle-jflex-plugin wanted to use as a default

https://github.com/jprante/gradle-plugin-jflex/blob/master/src/main/groovy/org/xbib/gradle/plugin/JFlexPlugin.groovy#L71

Even though that code can basically never use that default as the java sourceset will always have a srcDir, the src/main dir.

laurit (approved):

mateuszrzeszutek (approved):
Nice 👍

Issue #3403, opened 2021-06-24
Otel api/context/semconv classes in inst/


tobias-:
**Describe the bug**
It seems that some classes from opentelemetry/opentelemetry api/context/semconv classes are leaked into `inst/` in the javaagent `-all.jar`

**Steps to reproduce**
Listing the contents of the jar file. E.g. via
```bash
unzip -v otel-all.jar | egrep 'inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/(api|context|semconv)/'
```

**What did you expect to see?**
According to @mateuszrzeszutek in [this thread](https://cloud-native.slack.com/archives/C0150QF88FL/p1624532996147600), there should be no `io.opentelemetry.{api,context,semconv}`.

I.e. I was expecting to see an empty result.

**What did you see instead?**
338 files were found , and among them the following:
```
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/semconv/resource/attributes/ResourceAttributes.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/context/propagation/ContextPropagators.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/context/Context.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/trace/Span.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/metrics/Meter.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/common/AttributeKey.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/baggage/Baggage.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/GlobalOpenTelemetry.classdata
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/api/OpenTelemetry.classdata
```

**What version are you using?**
Tested with:
* `1.3.0` and `1.3.1` these contains 338 extra files.
* `1.2.0` and `1.1.0` contains 156 extra files.
* `1.0.0` contains 152 extra files.
* `0.13.1` contains 52 extra files.

**Environment**
Not Applicable

**Additional context**
Many, probably all, of these classes appear in the root of the jar as well, e.g. both of these files exist in the jar:
```
inst/io/opentelemetry/javaagent/shaded/io/opentelemetry/semconv/resource/attributes/ResourceAttributes.classdata
[...]

PR #3432, opened 2021-06-29
Exclude bootstrap dependencies from inst

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3403

### DIFF
--- a/instrumentation/build.gradle
+++ b/instrumentation/build.gradle
@@ -33,6 +33,8 @@ configurations {
   // exclude bootstrap dependencies from shadowJar
   implementation.exclude group: 'org.slf4j'
   implementation.exclude group: 'io.opentelemetry', module: 'opentelemetry-api'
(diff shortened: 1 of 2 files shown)


anuraaga (approved):
Ah - while working on the other PR I got to a point of reducing 393 duplicates to 129 or so. I guess those must have been these API ones.

To confirm, we need both, one to exclude bootstrap and one to prevent duplicate SDK files?

mateuszrzeszutek (approved):

laurit:
@anuraaga I believe there are two different kinds of duplicates. In the other PR you took care of duplicate jar entries for the exact same file. This PR is for duplicate files in jar root and `inst`. If a file is present in jar root `AgentClassLoader` will never load the one from `inst` so it shouldn't be needed.

iNikem (approved):

PR #3435, opened 2021-06-29
Refactor/rename BootstrapPackagesProvider and PropertySource

mateuszrzeszutek:
* Refactored `BootstrapPackagesProvider` to `BootstrapPackagesConfigurer`/`BootstrapPackagesBuilder`
* Added Javadocs that clearly state that `BootstrapPackagesConfigurer` can't add new packages to the bootstrap CL, it just defines ones that already are there
* Renamed `PropertySource` to `ConfigPropertySource`

After this PR is merged the only classes left in the old `spi` package are deprecated span/metrics exporter factories.


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/bootstrap/BootstrapPackagesConfigurer.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package io.opentelemetry.javaagent.extension.bootstrap;
+
+import io.opentelemetry.instrumentation.api.config.Config;
+
+/**
+ * This SPI can be used to define which packages/classes belong to the bootstrap classloader: all
+ * packages configured here will always be loaded by the bootstrap classloader, even if it is not a
+ * parent of the current context classloader.
```
laurit, line 13: > even if it is not a parent of the current context classloader.

Imo it would be better to rephrase this a bit. The first thing that comes to mind when context classloader is mentioned is thread context class loader which isn't really relevant here. Perhaps even if class loader that initiated loading of the class does not normally delegate loading of classes from given package to bootstrap class loader.
I'd also replace `classloader` with `class loader` though we use both.

laurit (approved):

trask (approved):

PR #3439, opened 2021-06-29
Read correct manifest

iNikem:


javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -94,7 +94,7 @@ public AgentClassLoader(File javaagentFile, String internalJarFileName, ClassLoa
       jarBase =
           new URL("x-internal-jar", null, 0, "/", new AgentClassLoaderUrlStreamHandler(jarFile));
       codeSource = new CodeSource(javaagentFile.toURI().toURL(), (Certificate[]) null);
-      manifest = getManifest(jarFile, jarEntryPrefix + META_INF_MANIFEST_MF);
```
mateuszrzeszutek, line 97: Can `META_INF_MANIFEST_MF` constant be removed too?

mateuszrzeszutek (approved):

anuraaga (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
trask: I'm surprised errorprone didn't complain about this 😁

trask (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
iNikem: removed

Issue #3427, opened 2021-06-29
Trace ID was not passed correctly in dubbo scenario


zmapleshine:
**Describe the bug**
Trace ID was not passed correctly in dubbo.

**Steps to reproduce**

Start four services（named consumer,service1,service2,provider） that  use Dubbo 2.7.x
Their invocation relationship is as follows：
  
consumer -> service1 -> service2 -> provider.

"consumer service" use dubbo client to invoke the dubbo interface provided by the "service1 service",
and "service1 service" use dubbo client to invoke the dubbo interface provided by the"service2 service",
and "service2 service" use dubbo client to invoke the dubbo interface provided by the "provider service".

**What did you expect to see?**

Trace passes correctly between service invocations.

**What did you see instead?**

Use the Jaeger trace analysis tool to check, found that the "service2 service" span, "provider service" span is also under the "consumer service",like this:

![image](https://user-images.githubusercontent.com/12667610/123757384-82044280-d8f0-11eb-8089-60fba8a1da0d.png)

**What version are you using?**

1.3.1,1.2.0

**Environment**

Compiler: jdk11

OS: MacOS

I seem to have found the root cause of the problem. How can I help fix this problem?

mateuszrzeszutek:
Hey @zmapleshine,

> I seem to have found the root cause of the problem. How can I help fix this problem?

An example that reproduces this issue would be very valuable for us - if you can send us a link to the repo with the setup that you've outlined in the post above that would be great. You can also submit a PR that fixes this issue.
Thanks!

zmapleshine:
> An example that reproduces this issue would be very valuable for us - if you can send us a link to the repo with the setup that you've outlined in the post above that would be great. 

I built a test project to help us test this problem well. 
[https://github.com/zmapleshine/dubbo-invocation-example](url)
The PR that fixes this issue #3442

PR #3442, opened 2021-06-30
Fix dubbo trace/span cross-process propagation 

zmapleshine:
It is inappropriate to use the Invocation for cross-process propagation. When the Dubbo Client makes an RPC call, No matter how to set up the invocation's attachement  in Filter, ``AbstractInvoker`` will eventually get the attachement from the RPCContext and backfill it to the invocation.

[OpenTelemetryFilter](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/OpenTelemetryFilter.java)

```java
if (kind.equals(CLIENT)) {
      context = tracer.startClientSpan(interfaceName, methodName);
      tracer.inject(context, (RpcInvocation) invocation, DubboInjectAdapter.SETTER);
    } else {
      context = tracer.startServerSpan(interfaceName, methodName, (RpcInvocation) invocation);
    }
```

``DubboInjectAdapter``
```java
@Override
  public void set(RpcInvocation rpcInvocation, String key, String value) {
    rpcInvocation.setAttachment(key, value);
  }
```

[AbstractInvoker](https://github.com/apache/dubbo/blob/dubbo-2.7.12/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/protocol/AbstractInvoker.java)

```java
RpcInvocation invocation = (RpcInvocation) inv;
        invocation.setInvoker(this);
        if (attachment != null && attachment.size() > 0) {
            invocation.addAttachmentsIfAbsent(attachment);
        }
        Map<String, String> contextAttachments = RpcContext.getContext().getAttachments();
        if (contextAttachments != null && contextAttachments.size() != 0) {
            /**
             * invocation.addAttachmentsIfAbsent(context){@link RpcInvocation#addAttachmentsIfAbsent(Map)}should not be used here,
             * because the {@link RpcContext#setAttachment(String, String)} is passed in the Filter when the call is triggered
             * by the built-in retry mechanism of the Dubbo. The attachment to update RpcContext will no longer work, which is
             * a mistake in most cases (for example, through Filter to RpcContext output traceId and spanId and other information).
             */
            invocation.addAttachments(contextAttachments);
        }
```

As a Dubbo server, the attachements of the ``RpcContext`` are set in the ``ContextFilter``, which is derived from the invocation. ,and this Filter is a high priority.

[ContextFilter](https://github.com/apache/dubbo/blob/dubbo-2.7.12/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/filter/ContextFilter.java)

```java
RpcContext.getContext()
                .setInvoker(invoker)
                .setInvocation(invocation)
//                .setAttachments(attachments)  // merged from dubbox
                .setLocalAddress(invoker.getUrl().getHost(),
                        invoker.getUrl().getPort());

        // merged from dubbox
        // we may already added some attachments into RpcContext before this filter (e.g. in rest protocol)
        if (attachments != null) {
            if (RpcContext.getContext().getAttachments() != null) {
                RpcContext.getContext().getAttachments().putAll(attachments);
            } else {
                RpcContext.getContext().setAttachments(attachments);
            }
        }
```

About this issue see #3427

### DIFF
--- a/instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboInjectAdapter.java
+++ b/instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboInjectAdapter.java
@@ -6,9 +6,9 @@
 package io.opentelemetry.instrumentation.apachedubbo.v2_7;
 
 import io.opentelemetry.context.propagation.TextMapSetter;
-import org.apache.dubbo.rpc.RpcInvocation;
+import org.apache.dubbo.rpc.RpcContext;
 
-class DubboInjectAdapter implements TextMapSetter<RpcInvocation> {
+class DubboInjectAdapter implements TextMapSetter<RpcContext> {
 
   static final DubboInjectAdapter SETTER = new DubboInjectAdapter();
 
--- a/instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboInjectAdapter.java
+++ b/instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboInjectAdapter.java
@@ -13,7 +13,7 @@ class DubboInjectAdapter implements TextMapSetter<RpcInvocation> {
   static final DubboInjectAdapter SETTER = new DubboInjectAdapter();
 
   @Override
-  public void set(RpcInvocation rpcInvocation, String key, String value) {
-    rpcInvocation.setAttachment(key, value);
+  public void set(RpcContext rpcContext, String key, String value) {
+    rpcContext.setAttachment(key, value);
   }
 }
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek (commented):
Would it be possible to change the tests so that they verify that passing context via `RpcInvocation` is broken?

tydhot (approved):
Yes, it should choose RpcContext here, thanks!

zmapleshine:
> Would it be possible to change the tests so that they verify that passing context via `RpcInvocation` is broken?

Well,I think this problem is difficult to verify in general tests because it occurs inside the dubbo invoke and there is no reference value to prove that the trace is not correct when dubbo server side extract trace context.

It seems that the problem can only be discovered by looking at it manually.

laurit (approved):

mateuszrzeszutek (approved):

mateuszrzeszutek:
> It seems that the problem can only be discovered by looking at it manually.

Too bad -- I hoped that it'd be possible to add a one-liner that does something with the `RpcContext` to reproduce that.

Still, thanks for fixing that! 👍

zmapleshine:
> Too bad -- I hoped that it'd be possible to add a one-liner that does something with the `RpcContext` to reproduce that.

I will pay attention to this issue, and if there is a better way to test it in the future, I will start a issue to discuss it.

PR #3443, opened 2021-06-30
Draft: suppress instrumentation flag

lmolkova:
See https://github.com/open-telemetry/opentelemetry-specification/issues/1767

> #### Suppressing instrumentation implementations in OTel
>  **Please note that context suppression is not really possible in client libraries as it requires dependency on instrumentation packages (to export suppress key)**.

>  - [pyhton](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/63e7561931cd2d5e8edca9281c5156b9852eb348/instrumentation/opentelemetry-instrumentation-requests/src/opentelemetry/instrumentation/requests/__init__.py#L60)
>  - [.net](https://github.com/open-telemetry/opentelemetry-dotnet/blob/6b7f2dd77cf9d37260a853fcc95f7b77e296065d/src/OpenTelemetry/SuppressInstrumentationScope.cs) 
>  - [js](https://github.com/open-telemetry/opentelemetry-js/blob/c85fbe6682b9a8fd52e4b99bdbb079833b04e845/packages/opentelemetry-core/src/trace/suppress-tracing.ts)
>  - [java suppresses nested client spans by default](https://github.com/trask/opentelemetry-java-instrumentation/blob/9c1083b5410d2a1f2ece899bbcd0c933ac7f6bbd/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java#L109)
>  - [seems ruby does it](https://github.com/open-telemetry/opentelemetry-specification/issues/530#issuecomment-605462917)

### DIFF
--- a/instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/BaseTracerTest.groovy
+++ b/instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/BaseTracerTest.groovy
@@ -8,11 +8,15 @@ package io.opentelemetry.instrumentation.api.tracer
 import io.opentelemetry.api.trace.Span
 import io.opentelemetry.api.trace.SpanKind
 import io.opentelemetry.context.Context
+import io.opentelemetry.context.ContextKey
 import spock.lang.Shared
 import spock.lang.Specification
 
+import static org.junit.Assume.assumeTrue
+
 // TODO add tests for BaseTracer
 class BaseTracerTest extends Specification {
+
   @Shared
   def tracer = newTracer()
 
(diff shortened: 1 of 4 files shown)


instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/BaseTracerTest.groovy
```diff
@@ -21,6 +25,10 @@ class BaseTracerTest extends Specification {
   @Shared
   def root = Context.root()
 
+  @Shared
+  def ContextKey<Boolean> suppressContextKey = ContextKey.named("otel.suppress_instrumentation")
```
laurit, line 29: is this used somewhere?

lmolkova: nope, this is a draft PR, I'm not even sure I want to proceed with it, let me close it to not confuse anyone.

PR #3452, opened 2021-06-30
Set propagators in opentelemetry-spring-boot-autoconfigure

### LINKED ISSUE #2771: opentelemetry-spring-boot-autoconfigure module is not setting up any propagators
When the opentelemetry-spring-boot-autoconfigure module builds the `OpenTelemetrySdk` in `OpenTelemetryAutoConfiguration`, it doesn't call `setPropagators`, so no propagators are set up and no context propagation takes place.

Related to #2624 (though in that issue the javaagent is also being used, so things are a bit different there).

Enkelian:
Resolves #2771 - OpenTelemetryAutoConfiguration now has context propagators set while configuring `OpenTelemetrySdk` builder. 

Propagators are configurable, using the approach from
https://github.com/spring-cloud-incubator/spring-cloud-sleuth-otel/blob/main/spring-cloud-sleuth-otel/src/main/java/org/springframework/cloud/sleuth/otel/propagation/CompositeTextMapPropagator.java

CLA required: Anna Nosek anosek@splunk.com


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
```diff
@@ -48,6 +50,10 @@ public OpenTelemetry openTelemetry(
         tracerProviderBuilder
             .setSampler(Sampler.traceIdRatioBased(samplerProperties.getProbability()))
             .build();
-    return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).buildAndRegisterGlobal();
+
+    return OpenTelemetrySdk.builder()
+        .setTracerProvider(tracerProvider)
+        .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
```
mateuszrzeszutek, line 56: Can you make it configurable? You could make it similar to how exporters are currently configured (some propagators, like B3, are not included in the API and are present in the `opentelemetry-extension-trace-propagators` jar)

Enkelian: I'll try my best. To to make sure I got that correctly, this would mean creating AutoConfigurations for different kinds of propagators (as seen in case of exporters). The configuration should then be passed as an argument. Correct me if I'm wrong :)

mateuszrzeszutek: Hmm, after thinking about it for a bit my previous idea is a bit lacking: opposite to span exporters, it makes sense to use multiple propagators (the detaults should be `w3c` and `baggage` after all), so implementing this exactly the same as exporters and using things like `@ConditionalOnMissingBean` (because there can be more than one) or `@ConditionalOnClass` (because once you add the propagators extension all of them will be on classpath) is far from ideal.

How about implementing this similarly to spring-sleuth-otel?
In particular see https://github.com/spring-cloud-incubator/spring-cloud-sleuth-otel/blob/main/spring-cloud-sleuth-otel-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/otel/OtelPropagationConfiguration.java#L97 and https://github.com/spring-cloud-incubator/spring-cloud-sleuth-otel/blob/main/spring-cloud-sleuth-otel/src/main/java/org/springframework/cloud/sleuth/otel/propagation/CompositeTextMapPropagator.java#L50. 
You could use a single property: a list of propagator names (with default value `baggage,w3c`) and build a composite one depending on what's on that list. WDYT?

Enkelian: Makes sense, I'll try implementing it that way. So both `w3c` and `baggage` should be added regardless of the value of the property?

mateuszrzeszutek: No, they should be used as the default value - if user doesn't specify any propagator at all these two should be set.

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
```diff
@@ -30,12 +33,14 @@
  */
 @Configuration
 @EnableConfigurationProperties(SamplerProperties.class)
+@Import(PropagationAutoConfiguration.class)
```
mateuszrzeszutek, line 36: Is that annotation needed? All other configurations use `@AutoConfigureBefore` and that seems to be enough to guarantee the correct order

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/** Configures {@link ContextPropagators} bean for propagation. */
+@Configuration
+@EnableConfigurationProperties(PropagationProperties.class)
+@AutoConfigureBefore(OpenTelemetryAutoConfiguration.class)
+@ConditionalOnProperty(prefix = "otel.propagation", name = "enabled", matchIfMissing = true)
+public class PropagationAutoConfiguration {
+
+  @Bean
```
mateuszrzeszutek, line 28: Consider adding `@ConditionalOnMissingBean` in case the user wants to configure that themselves

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+  ContextPropagators contextPropagators(ObjectProvider<List<TextMapPropagator>> propagators) {
+    List<TextMapPropagator> mapPropagators = propagators.getIfAvailable(ArrayList::new);
```
mateuszrzeszutek, line 30:
```suggestion
    List<TextMapPropagator> mapPropagators = propagators.getIfAvailable(Collections::emptyList);
```

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationAutoConfigurationTest.java
```diff
@@ -0,0 +1,91 @@
...
+    this.contextRunner
+        .withPropertyValues("otel.propagation.type=B3")
+        .run(
+            context -> {
+              CompositeTextMapPropagator compositePropagator =
+                  context.getBean("compositeTextMapPropagator", CompositeTextMapPropagator.class);
+
+              assertThat(compositePropagator.fields()).contains("b3");
+              assertThat(compositePropagator.fields())
+                  .doesNotContainAnyElementsOf(Arrays.asList("baggage", "traceparent"));
```
mateuszrzeszutek, line 88: You can use AssertJ's fluent API:
```suggestion
              assertThat(compositePropagator.fields())
                  .contains("b3")
                  .doesNotContain("baggage", "traceparent");
```

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagator.java
```diff
@@ -0,0 +1,102 @@
...
+    mapping.put(PropagationType.BAGGAGE, W3CBaggagePropagator.getInstance());
+    mapping.put(PropagationType.NOOP, TextMapPropagator.noop());
+  }
+
+  private static boolean isOnClasspath(String clazz) {
+    return ClassUtils.isPresent(clazz, null);
+  }
+
+  @Override
+  public Collection<String> fields() {
```
mateuszrzeszutek, line 70: Instead of implementing `fields()`, `inject()` and `extract()` yourself you can use `TextMapPropagator.composite()`; convert this class into some kind of factory class (or put the creation code in `PropagatorsConfiguration`), pass the configured propagators to the `composite()` method. WDYT?

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationType.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.spring.autoconfigure.propagators;
+
+/** Types of supported propagators. */
+public enum PropagationType {
```
mateuszrzeszutek, line 9: Would it be possible to use the same names as [OTel SDK](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#propagator) uses? (i.e. lowercase)

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationType.java
```diff
@@ -0,0 +1,28 @@
...
+  BAGGAGE,
+
+  /** JaegerPropagator */
+  JAEGER,
+
+  /** OtTracePropagator */
+  OT_TRACER,
+
+  /** NoOp propagator */
+  NOOP
```
mateuszrzeszutek, line 27: I don't think you need that option, if you want no-op you can just set `enabled = false`.

iNikem:
> [CLA Not Signed](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/210933087/3452/#/?version=1)
> 
> 
> * [❌](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/210933087/3452/#/?version=1) - Anna Nosek The commit ([fae683b](https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/fae683bb33c43919ecd37a1c0a6e37941fb31c14) ,c5989a44b83d2f213311ad2a08e32898bf709bd6 ,2a1db36c37f1affa6760d3e3e03e4c1b3b2a2061 ,14fe57327989c621c22931efe07cd90c493d1c8b ,c55bd33faf740a1733d794b70e4e95202dbb925b) is not authorized under a signed CLA. [Please click here to be authorized](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/9522512/210933087/3452/#/?version=1). For further assistance with EasyCLA, [please submit a support request ticket](https://jira.linuxfoundation.org/servicedesk/customer/portal/4).

@Enkelian Please follow the link in this comment and sign the CLA

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
Enkelian: You're right - I lacked the right configuration in the tests and fixing it made this line obsolete, thanks!

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagator.java
Enkelian: Sure, that's a way better approach. I can change `CompositeTextMapPropagator` to a factory class so as to not clutter the `PropagatorsConfiguration`.

PR #3453, opened 2021-06-30
Remove jetty-server exclude

laurit:
As `testing-common` doesn't depend on `jetty-server` any more this exclude isn't needed.


mateuszrzeszutek (approved):

anuraaga (approved):
Thanks!

PR #3454, opened 2021-06-30
Port AsyncSpanEndStrategy implementations to AsyncOperationEndStrategy

HaloFour:
Port the existing AsyncSpanEndStrategy implementations to the newer AsyncOperationEndStrategy that works with the `Instrumentation<REQUEST, RESPONSE>` API.

To keep the port relatively simple I have the existing strategies implement both `AsyncSpanEndStrategy` and `AsyncSpanOperationStrategy`.  Once the annotation instrumentations are ported to the new Instrumentation API then the implementation of the deprecated interface can be removed with any abstractions between `BaseTracer` and `Instrumenter<REQUEST, RESPONSE>`.


HaloFour:
/cc @mateuszrzeszutek 

Wanted some feedback on the direction before I went down the path of applying this to the other implementations.

Rather than copypasta the implementation I thought it would be easier to have the strategy temporarily implement both interfaces, since the implementation would be very similar between them.

mateuszrzeszutek:
> Rather than copypasta the implementation I thought it would be easier to have the strategy temporarily implement both interfaces, since the implementation would be very similar between them.

👍 
That's a great idea! It should make those changes much easier to review too.

mateuszrzeszutek (approved):

PR #3455, opened 2021-07-01
Build Java 16 smoke test images

trask:
I'm not sure how this is going to work out, so leaving the Java 15 image generation in place for now.


anuraaga:
I don't think Spock will run at all on Java 16. So good to migrate the smoke tests to junit :-)

anuraaga (approved):

PR #3456, opened 2021-07-01
Add prepare-release-branch to release needs since output is used.

anuraaga:
We reference `needs.prepare-release-branch`

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -160,7 +160,7 @@ jobs:
 
   release:
     runs-on: ubuntu-latest
-    needs: [ test, smoke-test, example-distro ]
+    needs: [ prepare-release-branch, test, smoke-test, example-distro ]
     steps:
       - uses: actions/checkout@v2.3.4
         with:


.github/workflows/patch-release-build.yml
```diff
@@ -160,7 +160,7 @@ jobs:
 
   release:
     runs-on: ubuntu-latest
-    needs: [ test, smoke-test, example-distro ]
+    needs: [ prepare-release-branch, test, smoke-test, example-distro ]
```
trask, line 163: Oops, I think I removed this when I added `test`, `smoke-test` and `example-distro`. are needs not transitive?

trask (approved):

.github/workflows/patch-release-build.yml
anuraaga: The dependency itself is but probably not that interpolation syntax which is reading this yaml block. May be able to replace `needs.` below with `jobs.` but not sure

trask: oh! yes, I see now, thx

PR #3464, opened 2021-07-01
Allow disabling thread details and disable for most tests.

anuraaga:
I think there is always a use case for disabling a global procesor that adds data, similar to how we support disabling ResourceProvider, so tried adding here. We have a use case here in that the thread details add noise to assertion messages and it's nicer if we don't have to worry.

### DIFF
--- a/testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/AttributesAssert.groovy
+++ b/testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/AttributesAssert.groovy
@@ -58,12 +58,6 @@ class AttributesAssert {
     Set<String> unverifiedAttributes = new TreeSet(allAttributes)
     unverifiedAttributes.removeAll(assertedAttributes)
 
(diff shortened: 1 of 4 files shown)


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/asserts/AttributesAssert.groovy
```diff
@@ -58,12 +58,6 @@ class AttributesAssert {
     Set<String> unverifiedAttributes = new TreeSet(allAttributes)
     unverifiedAttributes.removeAll(assertedAttributes)
 
-    // Don't need to verify thread details.
-    assertedAttributes.add("thread.id")
```
anuraaga, line 62: These were added by me when migrating the tests to run with the java agent, before we somehow skipped the thread processor. It's a pretty mysterious hack so glad they're gone.

mateuszrzeszutek (approved):

iNikem (approved):

trask (approved):

PR #3471, opened 2021-07-02
Fix maven packaging for instrumentation-api-caching

### LINKED ISSUE #3498: Caused by: java.lang.NoClassDefFoundError: io/opentelemetry/instrumentation/api/caching/Cache after switch to use the latest snapshot
**Describe the bug**
I started to get an error `Caused by: java.lang.NoClassDefFoundError: io/opentelemetry/instrumentation/api/caching/Cache` after switching to use snapshot version of `opentelemetry-java-instrumentation` from Sonatype. It works well with the forked version, where I have implemented JDBC instrumentation: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3367.
Originally I thought about the problem with the dependencies, but found that my application already includes all required libraries and I even can see instrumentation-caching library in the Gradle report:

```
|    +--- io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:1.4.0-alpha-SNAPSHOT
|    |    +--- io.opentelemetry:opentelemetry-api-metrics:1.3.0-alpha (*)
|    |    +--- org.slf4j:slf4j-api:1.7.30
|    |    +--- io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-caching:1.4.0-alpha-SNAPSHOT
```

But for some unknown reason classes from this dependency are not included in the classpath or not visible. I found that the problem started from this commit: https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/d53e62b4dcd2ffcce9b7fee0ea8f4539b7b55e53. And everything works perfectly with the previous commit: https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/47c8a9bfa56d77934df75ba1aaff2d47992db783. Probably @anuraaga you can help to find the reason why.

I have tried to compare Maven pom files for this dependency and looks like the previous version doesn't have `<packaging>pom</packaging>` attribute:

1) old pom version:

old

2) new pom version:

new

But seems it's probably not the reason why it's failed.

The problem only actually for this particular package, I have tried to resolve classes from `opentelemetry-instrumentation-api`, it works.

**Steps to reproduce**

I have added a project to reproduce this error: https://github.com/donbeave/opentelemetry-java-instrumentation-cache-class-not-found-error

Please check the CI build log: https://github.com/donbeave/opentelemetry-java-instrumentation-cache-class-not-found-error/runs/2991829149

The `test1` task is a success because it's using a class from the `opentelemetry-instrumentation-api` library, but `test2` task failed because it's using a class from the `opentelemetry-instrumentation-api-caching` library.

But if uncomment `mavenLocal()` in `build.gradle.kts` files and publish to Maven local this commit: https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/47c8a9bfa56d77934df75ba1aaff2d47992db783, everything will compile successfully.

`1.4.0-alpha-SNAPSHOT`

trask:
Currently the instrumentation-api-caching pom is being published with `<packaging>pom</packaging>`.

Somehow removing stuff seems to fix it, but probably breaks something else(?)

Resolves #3498


trask:
Converted PR back to Draft, it did break something else..

trask:
@anuraaga this almost worked (and seems to pass locally, but maybe also completely wrong), let me know if you have any ideas..

instrumentation-api-caching/build.gradle.kts
```diff
@@ -36,19 +36,4 @@ tasks {
 
     minimize()
   }
-
-  named("jar") {
-    enabled = false
```
anuraaga, line 41: This was to avoid publishing both a normal jar and the -all one. But it's added complication without much benefit so the change seems fine to me.

instrumentation-api-caching/build.gradle.kts
```diff
@@ -36,19 +36,4 @@ tasks {
...
-
-    dependsOn(shadowJar)
-  }
-}
-
-// Because shadow does not use default configurations
-publishing {
```
anuraaga, line 48: I think this might still be needed to make sure `-all` is published

mateuszrzeszutek:
Would it make sense to use the same method as in the SDK, where [trace-shaded-deps](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace-shaded-deps/build.gradle.kts) is not published but integrated into `:sdk:trace`? `instrumentation-api-caching` would be just a part of the `instrumentation-api` (same for `armeria-shaded-for-testing` and `testing-common`)

trask:
> Would it make sense to use the same method as in the SDK, where [trace-shaded-deps](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace-shaded-deps/build.gradle.kts) is not published but integrated into `:sdk:trace`? `instrumentation-api-caching` would be just a part of the `instrumentation-api` (same for `armeria-shaded-for-testing` and `testing-common`)

oh, I will try it 👍

trask:
> > Would it make sense to use the same method as in the SDK, where [trace-shaded-deps](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/trace-shaded-deps/build.gradle.kts) is not published but integrated into `:sdk:trace`? `instrumentation-api-caching` would be just a part of the `instrumentation-api` (same for `armeria-shaded-for-testing` and `testing-common`)
> 
> oh, I will try it 👍

IT WORKED ❗

iNikem:
I am confused now :) First, why do we need this `api-caching` as a separate module?

trask:
> I am confused now :) First, why do we need this `api-caching` as a separate module?

Because shadowJar. Help wanted 😭.

iNikem:
> > I am confused now :) First, why do we need this `api-caching` as a separate module?
> 
> Because shadowJar. Help wanted 😭.

I am afraid I don't quite understand. Can you explain, please?

trask:
> > > I am confused now :) First, why do we need this `api-caching` as a separate module?
> > 
> > 
> > Because shadowJar. Help wanted 😭.
> 
> I am afraid I don't quite understand. Can you explain, please?

I don't really understand either 😭. I only know that this is the only incantation of shadowJar that I've been able to get to work and publish correct maven artifacts.

anuraaga:
The gradle shadow plugin doesn't seem to give a good way to solve the use case "I want to relocate a dependency and package it into this artifact without other changes". The closest it comes is to do so while publishing a `-all` artifact, and our workarounds to try to publish without the `-all` seem to have not worked well so far. I've actually seen strange behavior with Gradle metadata before too

https://github.com/open-telemetry/opentelemetry-lambda/blob/main/java/layer-javaagent/build.gradle.kts#L12

Currently it seems the best pattern for shadowing a library without having it affect publishing in any way is to keep the shadowing in a separate module and embedding that into another one.

mateuszrzeszutek (approved):

Issue #2778, opened 2021-04-12
Extract the muzzle codegen gradle plugin


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
The aim of this issue is to extract a gradle plugin that'll contain the muzzle compile-time code generation functionality (i.e. the byte buddy plugin). It should be possible to use the new gradle plugin in external instrumentations. The new plugin should be published independently from the main project.

iNikem:
Done.

PR #3474, opened 2021-07-02
Split up build logic to plugins that can eventually be published and …

anuraaga:
…this-repo-specific stuff.

This creates several plugins prefixed `io.opentelemetry.instrumentation` (our maven coordinate). It means we can then separate them into a separate project, apply the gradle plugin publish plugin, and publish them to Gradle plugin portal as part of our release build. Of course, that would first require testing that a distro can actually be built with just those plugins which I'm hoping @iNikem will help with as part of #2778 

We still have `otel.` plugins to use these publishable plugins as well as add our repository's specific opinions (errorprone, etc)


buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
```diff
@@ -0,0 +1,10 @@
+plugins {
+  id("io.opentelemetry.instrumentation.javaagent-testing")
+  id("io.opentelemetry.instrumentation.muzzle-check")
+}
+
+dependencies {
+  add("muzzleBootstrap", "io.opentelemetry.javaagent:opentelemetry-javaagent-bootstrap")
```
anuraaga, line 7: As @mateuszrzeszutek said, `instrumentationMuzzle` seemed to be equivalent to `runtimeClasspath`, which is equivalent to just having a normal dependency like this one

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
```diff
@@ -1,7 +1,16 @@
 /** Common setup for manual instrumentation of libraries and javaagent instrumentation. */
 
 plugins {
-  id("otel.java-conventions")
+  `java-library`
+}
+
+afterEvaluate {
```
anuraaga, line 7: Compared to previous version, this publishable one removes the (javaagent, library, testing) folder scheme handling from this file (moved to otel.java-conventions) and adds this application of the BOM

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -0,0 +1,116 @@
...
+
+  id("io.opentelemetry.instrumentation.base")
+  id("io.opentelemetry.instrumentation.javaagent-codegen")
+  id("io.opentelemetry.instrumentation.javaagent-shadowing")
+}
+
+dependencies {
+  // Integration tests may need to define custom instrumentation modules so we include the standard
+  // instrumentation infrastructure for testing too.
+  compileOnly("io.opentelemetry.instrumentation:opentelemetry-instrumentation-api")
```
anuraaga, line 16: Compared to the previous version, these have been changed from project dependencies (which are subbed back in by otel.java-conventions)

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -0,0 +1,116 @@
...
+  isCanBeResolved = true
+}
+
+tasks.named<ShadowJar>("shadowJar").configure {
+  configurations = listOf(project.configurations.runtimeClasspath.get(), testInstrumentation)
+
+  archiveFileName.set("agent-testing.jar")
+}
+
+val agentForTesting by configurations.creating {
```
anuraaga, line 49: Instead of directly referencing task, I made this configuration and resolve via it to allow resolving the external dependency (or the project dependency that's subbed in in otel.java-conventions)

buildSrc/src/main/kotlin/otel.javaagent-testing.gradle.kts
```diff
@@ -1,100 +1,20 @@
...
-    // OpenTelemetry SDK is not needed for compilation
-    exclude(group = "io.opentelemetry", module = "opentelemetry-sdk")
-    exclude(group = "io.opentelemetry", module = "opentelemetry-sdk-metrics")
-  }
-  compileOnly(project(":javaagent-tooling")) {
-    // OpenTelemetry SDK is not needed for compilation
-    exclude(group = "io.opentelemetry", module = "opentelemetry-sdk")
-    exclude(group = "io.opentelemetry", module = "opentelemetry-sdk-metrics")
-  }
   compileOnly("net.bytebuddy:byte-buddy")
```
anuraaga, line 12: Hoping this won't need to be added for external dependencies (wonder if we even need it internally I didn't try deleting it yet)

anuraaga: If this (and slf4j-api) are needed, we should probably just make them `api` dependencies on `javaagent-api`

mateuszrzeszutek: I don't think that `testing-common` and `javaagent-api` need byte buddy, it should be an `api` dependency in `javaagent-extension-api`

mateuszrzeszutek (approved):

instrumentation/apache-camel-2.20/javaagent/build.gradle.kts
```diff
@@ -47,21 +54,9 @@ dependencies {
...
 
 tasks {
-
   withType<Test>().configureEach {
     // TODO run tests both with and without experimental span attributes
     jvmArgs("-Dotel.instrumentation.apache-camel.experimental-span-attributes=true")
     jvmArgs("-Dotel.instrumentation.aws-sdk.experimental-span-attributes=true")
   }
-
-  named<Javadoc>("javadoc") {
```
anuraaga, line 57: This was ported over from the Groovy before

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.3.1/instrumentation/apache-camel-2.20/javaagent/apache-camel-2.20-javaagent.gradle#L55

But was originally incorrect because dependencies is on `project`, having it under `javadoc` didn't do anything, but apparently it could still work in some cases based on the evaluation order of the project based on other configuration. This PR seems to have triggered a subtle change in order that didn't evaluate `javadoc` in time

buildSrc/src/main/kotlin/otel.javaagent-testing.gradle.kts
anuraaga: Cool we already had it

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// need to run this after evaluate because testSets plugin adds new test tasks
+afterEvaluate {
+  tasks.withType<Test>().configureEach {
+    val shadowJar = tasks.shadowJar.get()
+    val agentShadowJar = agentForTesting.resolve().first()
+
+    inputs.files(agentForTesting)
+    inputs.file(shadowJar.archiveFile)
```
iNikem, line 65: I think you can and should depend on the task, not explicitly on its output file.

anuraaga: You mean with `dependsOn(shadowJar)` as on line 67? I thought so too, I think @trask added this later finding it wasn't enough. I haven't tried investigating past that but for the purpose of this PR, it's copied over from before any kotlin migrations.

iNikem (approved):

PR #3477, opened 2021-07-02
Remove dependency on javaagent-bootstrap from javaagent-extension-api

mateuszrzeszutek:


javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/internal/ClassLoaderMatcherCacheHolder.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.bootstrap;
+package io.opentelemetry.javaagent.instrumentation.api.internal;
```
iNikem, line 6: It still lives in bootstrap classloader, right?

mateuszrzeszutek: Yes - but at least it's not in the (internal) `javaagent-bootstrap` module.

iNikem: Why exactly do we consider bootstrap module as "internal"?

iNikem (approved):

javaagent-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/internal/ClassLoaderMatcherCacheHolder.java
mateuszrzeszutek: There've been mentions of making it internal: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/2779#discussioncomment-601705

trask (approved):
👍

PR #3490, opened 2021-07-03
Update Apache HttpAsyncClient to Instrumenter API

trask:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanStatusExtractor.java
```diff
@@ -40,7 +40,10 @@ public StatusCode extract(REQUEST request, @Nullable RESPONSE response, Throwabl
     if (response != null) {
       Integer statusCode = attributesExtractor.statusCode(request, response);
       if (statusCode != null) {
-        return HttpStatusConverter.statusFromHttpStatus(statusCode);
+        StatusCode statusCodeObj = HttpStatusConverter.statusFromHttpStatus(statusCode);
+        if (statusCodeObj == StatusCode.ERROR) {
+          return statusCodeObj;
+        }
```
trask, line 46: Please note this change in behavior.

What prompted this here is that on circular redirect, apache-httpasyncclient calls `Instrumenter.end()` with both a 302 response object AND and exception.

And our http client tests think that circular redirects should result in `StatusCode.ERROR`.

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpClientRequest.java
```diff
@@ -0,0 +1,137 @@
...
+
+public final class ApacheHttpClientRequest {
+
+  private static final Logger logger = LoggerFactory.getLogger(ApacheHttpClientRequest.class);
+
+  @Nullable private final URI uri;
+
+  private final HttpRequest delegate;
+
+  @SuppressWarnings("SystemOut")
```
mateuszrzeszutek, line 28: There's no `System.out` calls here anymore?

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/HttpHeaderSetter.java
```diff
@@ -6,14 +6,13 @@
 package io.opentelemetry.javaagent.instrumentation.apachehttpasyncclient;
 
 import io.opentelemetry.context.propagation.TextMapSetter;
-import org.apache.http.HttpRequest;
 
-public class HttpHeadersInjectAdapter implements TextMapSetter<HttpRequest> {
+public class HttpHeaderSetter implements TextMapSetter<ApacheHttpClientRequest> {
 
-  public static final HttpHeadersInjectAdapter SETTER = new HttpHeadersInjectAdapter();
+  public static final HttpHeaderSetter SETTER = new HttpHeaderSetter();
```
mateuszrzeszutek, line 12: Is `SETTER` used anywhere?

PR #3492, opened 2021-07-05
Remove usages of agent runtime specific code from reusable classes

iNikem:
I am not so sure in these changes, but reusable `ReferenceMatcher` cannot depend on agent runtime specific `Utils.getBootstrapProxy()`

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
@@ -7,7 +7,6 @@ package io.opentelemetry.javaagent.tooling.muzzle.matcher;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
-import static net.bytebuddy.dynamic.loading.ClassLoadingStrategy.BOOTSTRAP_LOADER;
 
 import io.opentelemetry.instrumentation.api.caching.Cache;
 import io.opentelemetry.javaagent.extension.muzzle.ClassRef;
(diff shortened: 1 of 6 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
```diff
@@ -57,9 +55,6 @@ public ReferenceMatcher(
    * @return true if all references match the classpath of loader
    */
   public boolean matches(ClassLoader userClassLoader) {
-    if (userClassLoader == BOOTSTRAP_LOADER) {
```
anuraaga, line 60: Nice

anuraaga (commented):
At a glance this seems like a nice change

mateuszrzeszutek (approved):
Nit: `LocationStrategy` and `PoolStrategy` could be injected into `ReferenceMatcher` - this way you'd avoid depending on agent-internal bytebuddy component implementations.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
```diff
@@ -87,6 +94,10 @@ private boolean doesMatch(ClassLoader loader) {
   }
 
   private static TypePool createTypePool(ClassLoader loader) {
+    Objects.requireNonNull(loader); // "proxy bootstrap" must be passed instead of null
+
+    // it's ok to use locationStrategy() without fallback bootstrap proxy here since the class
+    // loader cannot be null
     return AgentTooling.poolStrategy()
         .typePool(AgentTooling.locationStrategy().classFileLocator(loader), loader);
```
trask, line 102: this took me some time to convince myself that it was ok to use locationStrategy without  fallback bootstrap proxy here, so thought these comments / `requireNonNull` validations may help others (this module would definitely benefit from `nullaway`, then we wouldn't need these `requireNonNull` -- hm, maybe I'll turn these `requireNonNull` just into method comment

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/ReferenceMatcher.java
iNikem: I actually like `Objects.requireNonNull` in general. As long as this will not end in customer JVM crash :)

PR #3493, opened 2021-07-05
Don't publish per-instrumentation testing -all jar.

anuraaga:
Found that shadow plugin automatically adds its elements to the java component

https://github.com/johnrengelman/shadow/blob/master/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/ShadowJavaPlugin.groovy#L54

And found an example of how to skip publishing of an element added by a plugin

https://docs.gradle.org/current/userguide/publishing_customization.html#sec:adding-variants-to-existing-components


mateuszrzeszutek (approved):

trask (approved):
👍

PR #3494, opened 2021-07-05
JDBC instrumentation: switch to use api instead of implementation for instrumentation-api.

donbeave:
This fixes some issues with compilation, based on official Gradle documentation:

> Dependencies appearing in the `api` configurations will be transitively exposed to consumers of the library, and as such will appear on the compile classpath of consumers.
> 
> Dependencies found in the `implementation` configuration will, on the other hand, not be exposed to consumers, and therefore not leak into the consumers' compile classpath.


mateuszrzeszutek (approved):

iNikem (approved):

anuraaga (approved):

instrumentation/jdbc/library/build.gradle.kts
```diff
@@ -8,7 +8,7 @@ plugins {
 }
 
 dependencies {
-  implementation(project(":instrumentation-api"))
+  api(project(":instrumentation-api"))
```
anuraaga, line 11: I think we add this with our plugin - can we just delete?

donbeave: Yeah, I think can delete it. Actually, I found another problem and thought this fix can fix it, but eventually found that problem is not here. Please check this issue: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3498

mateuszrzeszutek (approved):

instrumentation/jdbc/library/build.gradle.kts
donbeave: You are right. Removed.

Issue #1957, opened 2020-12-23
Instrumentation-specific bootstrap classes


trask:
@iNikem introduced a nice feature in #1917 to allow instrumentation to have their own bootstrap classes.

I removed it in #1643 (moving the `KeyHolder` class from #1917 into `javaagent-api`), but would be nice to figure out how to make that feature work with the new javaagent testing infra introduced in #1643.

PR #3495, opened 2021-07-05
Instrumentation-specific bootstrap classes

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1957

As a part of stabilizing `javaagent-(advice|runtime|instrumentation)-api` I thought it'd be pretty good to move instrumentation-specific classes out of this module. This PR does that just for Undertow, but if this approach is okay I'll move all others in the next one.


javaagent-bootstrap/build.gradle.kts
```diff
@@ -25,6 +25,7 @@ dependencies {
 
   implementation(project(":instrumentation-api"))
   implementation(project(":javaagent-api"))
+  implementation(project(":instrumentation", configuration = "bootstrap"))
```
iNikem, line 28: I am very suspicious about this dependency... `javaagent-bootstrap` depending on `instrumentation` project is not something I would expect... Also, this will not work with extensions, right? I mean, instrumentation from extension will not be able to add bootstrap classes this way?

mateuszrzeszutek: > Also, this will not work with extensions, right? I mean, instrumentation from extension will not be able to add bootstrap classes this way?

No, it won't. Currently there's no way to add something to bootstrap with an extension, it has to be built into the javaagent jar.

> I am very suspicious about this dependency... `javaagent-bootstrap` depending on `instrumentation` project is not something I would expect...

Well, it depends on a specific configuration of `instrumentation` - but if you take that part away it might be confusing.

laurit: For extensions we could package classes that are supposed to be loaded by agent class loader under some prefix like `inst/` in agent and use `Instrumentation.appendToBootstrapClassLoaderSearch` on extension jar if we find that extensions also need to add stuff too boot classpath.

laurit (approved):

javaagent-bootstrap/build.gradle.kts
mateuszrzeszutek: > For extensions we could package classes that are supposed to be loaded by agent class loader under some prefix like `inst/` in agent and use `Instrumentation.appendToBootstrapClassLoaderSearch` on extension jar if we find that extensions also need to add stuff too boot classpath.

That or create a new SPI (or extend `BootstrapPackagesConfigurer`) to enable marking whole packages as "should inject to bootstrap"

trask (approved):

trask:
> I am very suspicious about this dependency... javaagent-bootstrap depending on instrumentation project is not something I would expect...

`javaagent-bootstrap` has a dual nature
* code
* aggregator of things that go in the bootstrap class loader

maybe it would help if we split this a part?

iNikem:
> * aggregator of things that go in the bootstrap class loader

Can `javaagent` module be such an aggregator? It brings everything together and should know what goes where.

mateuszrzeszutek:
> Can `javaagent` module be such an aggregator? It brings everything together and should know what goes where.

That's a good idea, I'll try it out.

mateuszrzeszutek:
After trying out moving things to `javaagent` I've got mixed feelings about it: if we want to move the "aggregator" functionality to `javaagent` from `javaagent-bootstrap` then we'd have to duplicate it in the `agent-for-testing` module - which is far from ideal.
I guess we could create another module that serves just as an aggregator; or just accept the fact that `javaagent-bootstrap` (`javaagent-internal-bootstrap`?) does the aggregation. WDYT?

mateuszrzeszutek:
Merging this for now; I'll move the dependency aggregation to `javaagent` in a separate PR.

PR #3497, opened 2021-07-05
Reduce CallDepth classes' API surface

mateuszrzeszutek:
This PR makes `CallDepthThreadLocalMap` package-private and refactors all classes so that they always use `CallDepth` methods. Also adds some javadocs to `CallDepth`. 

I've noticed that in most places we're getting `CallDepth` twice: once when entering the method, the second time when exiting. This means we're retrieving the value from a `ThreadLocal` twice - WDYT about always putting the `CallDepth` into a local var when it's retrieved the first time?


trask:
> WDYT about always putting the `CallDepth` into a local var when it's retrieved the first time?

I like this pattern (or something similar) 👍

```
    public static void methodEnter(
        ...
        @Advice.Local("otelCallDepth") CallDepth callDepth) {

      callDepth = CallDepthThreadLocalMap.getCallDepth(HttpURLConnection.class);
      if (callDepth.getAndIncrement() > 0) {
        return;
      }

      ...
    }

    public static void methodExit(
        ...
        @Advice.Local("otelCallDepth") CallDepth callDepth) {

      if (callDepth.decrementAndGet() > 0) {
        return;
      }

      ...
    }
```

mateuszrzeszutek:
> I like this pattern (or something similar) 👍

Yep, that's exactly what I thought of

anuraaga (approved):

instrumentation/hibernate/hibernate-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/SessionMethodUtils.java
```diff
@@ -43,7 +43,7 @@
       return null; // No state found. We aren't in a Session.
     }
 
-    int depth = CallDepthThreadLocalMap.incrementCallDepth(SessionMethodUtils.class);
+    int depth = CallDepth.forClass(SessionMethodUtils.class).getAndIncrement();
```
iNikem, line 46: Should we have more succinct API? Like `CallDepth.increment(Class)`?

mateuszrzeszutek: Hmm, `CallDepth.increment(Class)` would return an int, correct?
I think that if we use the pattern that Trask posted in this PR there'll be no need for static `increment`/`decrement` methods - we will always have to obtain a `CallDepth` instance exactly once and use it both before and after the original method code. (I'll implement that pattern everywhere if this PR gets merged)

iNikem: Ok, with that pattern the shorter API indeed does not make sense.

iNikem (approved):

PR #3499, opened 2021-07-05
Style guide section for Nullable annotation usage

trask:


docs/contributing/style-guideline.md
```diff
@@ -96,3 +96,22 @@ Method parameters should never be declared `final`.
...
+
+### `@Nullable` annotation usage
+
+[Note: this section is aspirational, as opposed to a reflection of the current codebase]
+
+All parameters and fields which can be `null` should be annotated with `@Nullable`
+(specifically `org.checkerframework.checker.nullness.qual.Nullable`, which is included by the
+`otel.java-conventions` gradle plugin as a `compileOnly` dependency).
+
+The default is assumed to be non-null, so there is no need to use `@NonNull`.
```
iNikem, line 108: This is not enough for Idea to highlight when `null` is passed to non-null parameter.

mateuszrzeszutek: You need `@ParametersAreNonnullByDefault` in `package-info.java` for that. Should we start adding those? At the very least to the *-api modules.

iNikem: First I would like to make it clear: does this mean that we don't need to defensively check passed attributes for null? Because SDK is full of those, where they check if `ParametersAreNonnullByDefault` parameter is actually not null.

trask: @iNikem PTAL I think I addressed your questions

docs/contributing/style-guideline.md
```diff
@@ -105,10 +105,22 @@ All parameters and fields which can be `null` should be annotated with `@Nullabl
...
-nullable usage validation:
+```java
+@DefaultQualifier(value = NonNull.class)
+package io.opentelemetry.instrumentation.api;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.framework.qual.DefaultQualifier;
+```
+
+Public APIs should still defensively check for `null` parameters, even if the parameter is not
```
iNikem, line 119: 😭 I understand the need, but Idea will be all yellow! :(

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/package-info.java
```diff
@@ -0,0 +1,5 @@
+@DefaultQualifier(value = NonNull.class)
```
anuraaga, line 1: This is still one point where all three repos aren't consistent and is on my list of things to address eventually. SDK uses `javax.annotation.Nullable`, which is less powerful than `checkerframework` since it doesn't support type annotations. Even if we don't use type annotations proactively, I'd like to converge on checkerframework version to at least have the ability for it without any real downside otherwise.

For the DefaultQualifier, we probably want to define our own meta-annotation that is something like

https://github.com/open-telemetry/opentelemetry-java-contrib/blob/8944db8459a51a5cb0500ecaf6f2903ec7546982/aws-xray/src/main/java/io/opentelemetry/contrib/awsxray/package-info.java#L7

Marking everything as NonNull by default is too much I've found (in particular, wildcard bounds become generally wrong).

docs/contributing/style-guideline.md
trask: Ya, that is annoying. I found disabling the inspection below gets rid of that yellow. May not be so bad to disable it once we have enforcement via nullaway.

image

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/package-info.java
trask: 👍 added those same locations restrictions

anuraaga (approved):

mateuszrzeszutek (approved):

PR #3501, opened 2021-07-05
Add temporary warning about memory leak to readme

trask:
Based on contributor feedback in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3027#issuecomment-871939174

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ if you are looking for documentation on using those.
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


trask:
I think this raises an interesting question about when to do patch releases. I have added this to today's SIG meeting agenda https://docs.google.com/document/d/1WK9h4p55p8ZjPkxO75-ApI9m0wfea6ENZmMoFRvXSCw.

mateuszrzeszutek (approved):

PR #3502, opened 2021-07-06
Cancel in-progress CI build on main branch

iNikem:


mateuszrzeszutek (approved):

anuraaga (approved):

trask:
there's some benefit to not canceling on main branch, since there's a history on the commit page of the build status for each commit

https://github.com/open-telemetry/opentelemetry-java-instrumentation/commits/main

anuraaga:
Ah good call out - keeping seems a bit better. I guess the downside is of multiple main builds hog our workers preventing accept-pr but maybe it's ok still?

iNikem:
> there's some benefit to not canceling on main branch, since there's a history on the commit page of the build status for each commit
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/commits/main

Ok, makes sense.

PR #3506, opened 2021-07-06
Springboot propagators

### LINKED ISSUE #2771: opentelemetry-spring-boot-autoconfigure module is not setting up any propagators

Enkelian:
Resolves #2771 - OpenTelemetryAutoConfiguration now has context propagators set while configuring OpenTelemetrySdk builder.

Propagators are configurable, using similar approach as spring-cloud-sleuth-otel.

CLA required: Anna Nosek anosek@splunk.com


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationType.java
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.spring.autoconfigure.propagators;
+
+/** Types of supported propagators. */
+public enum PropagationType {
+
+  /** B3 Single */
+  b3,
```
mateuszrzeszutek, line 12: We follow the [Google Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) that recommends using CONSTANT_CASE for enums. 
Is it possible to use some kind of converter for those values? If not, I guess using plain strings (like `"b3"`) without an enum would be fine too.

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
```diff
@@ -0,0 +1,88 @@
...
+          break;
+        case tracecontext:
+          propagators.add(W3CTraceContextPropagator.getInstance());
+          break;
+        case baggage:
+          propagators.add(W3CBaggagePropagator.getInstance());
+          break;
+      }
+
+      propagators.add(TextMapPropagator.noop());
```
mateuszrzeszutek, line 79: Is this needed?

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    Set<TextMapPropagator> propagators = new HashSet<>();
+
+    for (PropagationType type : types) {
+      switch (type) {
+        case b3:
+          if (isOnClasspath("io.opentelemetry.extension.trace.propagation.B3Propagator")) {
+            propagators.add(
+                beanFactory
+                    .getBeanProvider(B3Propagator.class)
+                    .getIfAvailable(B3Propagator::injectingSingleHeader));
```
mateuszrzeszutek, line 36: Does this mean that you have to have a bean of type `B3Propagator` in your application?
Why not just
```suggestion
            propagators.add(B3Propagator.injectingSingleHeader());
```
?

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationType.java
Enkelian: I think using just strings would make sense, thanks

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
Enkelian: In case user supplies a propagator that we don't support (or simply a wrong one). I guess now that I'll be converting enums to strings it would make sense to move that line to a default case in the switch, correct me if I'm wrong.

mateuszrzeszutek: That should be fine even without that - if you look at `TextMapPropagator#composite()` implementation you'll see that it handles an empty list of propagators correctly. I think it's fine to rely on that.

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
Enkelian: The current approach uses the `B3Propagator` bean if it's present, but if it's not it creates an instance in `getIfAvailable` (similarly as in https://github.com/spring-cloud-incubator/spring-cloud-sleuth-otel/blob/main/spring-cloud-sleuth-otel/src/main/java/org/springframework/cloud/sleuth/otel/propagation/CompositeTextMapPropagator.java)

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
Enkelian: Right, it does the `isEmpty` check and then supplies noop if needed. Thanks!

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
mateuszrzeszutek: Oh, I see - that's pretty cool 👍

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
```diff
@@ -0,0 +1,88 @@
...
+          }
+          break;
+        case "tracecontext":
+          propagators.add(W3CTraceContextPropagator.getInstance());
+          break;
+        case "baggage":
+          propagators.add(W3CBaggagePropagator.getInstance());
+          break;
+        default:
+          break;
```
mateuszrzeszutek, line 78: How about logging a warning about an unsupported value?

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationAutoConfigurationTest.java
```diff
@@ -0,0 +1,85 @@
...
+        .run(
+            context -> {
+              TextMapPropagator compositePropagator =
+                  context.getBean("compositeTextMapPropagator", TextMapPropagator.class);
+
+              assertThat(compositePropagator.fields())
+                  .contains("b3")
+                  .doesNotContain("baggage", "traceparent");
+            });
+  }
```
mateuszrzeszutek, line 84: Can you add a test that verifies what happens when you pass an unsupported propagator value?

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
Enkelian: Good idea. This means creating logger for the class and then `log.warn()`?

mateuszrzeszutek: Yep, exactly

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/PropagationAutoConfigurationTest.java
Enkelian: Should the test check if the warning is logged?

mateuszrzeszutek: Nah, just verify that no propagator at all is used.

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/propagators/CompositeTextMapPropagatorFactory.java
```diff
@@ -75,6 +80,7 @@ static TextMapPropagator getCompositeTextMapPropagator(
           propagators.add(W3CBaggagePropagator.getInstance());
           break;
         default:
+          log.warn("Unsupported type of propagator: " + type);
```
mateuszrzeszutek, line 83: Nit: you can use slf4j's built-in message formatting facilities instead of gluing strings yourself:

```suggestion
          log.warn("Unsupported type of propagator: {}", type);
```

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

trask (approved):
thx @Enkelian!

Issue #3503, opened 2021-07-06
Test if publishing works as expected


iNikem:
Recently we had several problems when modules publish unexpected artifacts. That lead to failures in downstream projects which depend on them, such as custom distributions.

We need to have automated tests for at least two separate scenarios:

- [ ] External extension can consume required artifacts to successfully build and test itself
- [ ] External library dependency can consume required artifacts to successfully build and test itself

Both external modules can live in our `examples` folder. The test itself should build the main project, publish it to local maven repository and then try to build-test example module, consuming dependencies from local maven repo.

PR #3507, opened 2021-07-06
Nightly build examples against locally published agent snapshots

iNikem:
Part of #3503.

As examples serve as, well, examples, I chose not to spoil their build scripts with this conditional "let us choose snapshot" logic. Instead I did a small shell script that makes required changes in extensions example build scripts externally.

It does fail currently, because our examples were not yet updated to latest snapshot changes.

### DIFF
--- a/.github/workflows/nightly-no-cache.yml
+++ b/.github/workflows/nightly-no-cache.yml
@@ -115,7 +115,7 @@ jobs:
       - name: Test
         run: ./gradlew :smoke-tests:test -PsmokeTestSuite=${{ matrix.suite }} --no-build-cache
 
-  example-distro:
+  examples:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.3.4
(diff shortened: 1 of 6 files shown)


.github/workflows/nightly-no-cache.yml
```diff
@@ -134,13 +134,47 @@ jobs:
...
+        working-directory: examples/extension
+
+      - name: Build
+        run: ./gradlew build --no-build-cache
+        working-directory: examples/extension
+
   issue:
     name: Open issue on failure
-    needs: [ build, test, testLatestDeps, smoke-test, example-distro ]
+    needs: [ build, test, testLatestDeps, smoke-test, examples, examples-snapshot ]
```
trask, line 177: how will we fix `examples-snapshot` failures while there are still breaking changes between latest release and snapshot?

iNikem: Doh, this again. You are right. Is it enough if I remove creating new issue or do you suggest something more?

trask: How about a copy of the examples directory `examples-snapshot` that we keep up-to-date with the latest snapshot (until we have stable apis)?

iNikem: Just to be sure: are we sure that we don't want to ALWAYS build examples against snapshot?

iNikem: @anuraaga @mateuszrzeszutek your opinion?

iNikem: Especially if we do #3516

anuraaga: I think this is just something we live with until the APIs are stable, which is getting good activity now. Until then we can probably treat this as a best effort check rather than anything too noisy, and we'll get it eventually.

iNikem: Sorry, @anuraaga , I did not understand which if the proposed options do you support :) 

1. Always build against snapshot and fix ASAP
2. Have a separate copy of examples to build against snapshot and main copy against last released version
3. Accept nightly workflow having failing job when breaking changes are introduced

anuraaga: > Accept nightly workflow having failing job when breaking changes are introduced

I was leaning towards this. 1) seems like a no go, they're not examples if they can't be used against released code. 2) seems too tedious. 3) isn't great of course but seems best of the three to me

iNikem: Why do you think building against snapshot is no-go? Snapshots are still published to sonatype, so everybody can build and try them. And if tags show to a state, where examples use this exact version, then we have stable points in time as well.

anuraaga: The examples are supposed to be like tutorials - tutorials don't point at snapshots since users are supposed to not use snapshots in their apps, even if it's possible. I guess having examples in a release branch pointing at non-snapshots is OK, but I do think such examples have to exist somewhere

iNikem: Agreed. Examples tagged with release version must point to that release version.

mateuszrzeszutek: Hm, let's assume that our APIs are stable for a moment.

We want to verify that we have no publishing issues, right? Like a unit test for our gradle publishing scripts/conventions. Would it make sense to create an empty project that has no source code at all, just a gradle script that depends on our snapshots? The examples could always use the stable version, as @anuraaga said they're like tutorials - I'm guessing that most people will only look at the main branch for them (still, examples in the tagged release must have correct versions).
This way examples are just examples - sure, they must build and be valid, but they have exactly one use. For the other scenario (publishing test) we have a dedicated test module.

Going back to the API stability: if we worry about this we could setup API diffs like the otel-java has - our examples can't possibly cover all cases anyway.

anuraaga: @mateuszrzeszutek I believe we do need some source code, the packaging=POM publishing seems to have not caused obvious Gradle resolution failure but `NoClassDefFoundError`. If we were to have a separate package it would still need to exercise our APIs to some extent I guess. It seems like we are converging on an idea for using the examples for both in #3516

mateuszrzeszutek: Oh, I see - I thought it failed on the gradle script level. Then I guess it does make sense to use examples for that, having a separate module that runs as many of our public classes as possible would be a major pain.

iNikem:
Will do another way

PR #3515, opened 2021-07-07
Use consistent logger field name

### LINKED ISSUE #3364: Make logger variable name consistent
Looks like we currently have LOG, log, logger throughout the codebase. We should make them consistent, suggest `logger`

trask:
Resolves #3364


anuraaga (approved):
Wonder if there are any checkstyle approaches for this

iNikem (approved):

mateuszrzeszutek (approved):

trask:
> Wonder if there are any checkstyle approaches for this

I'll send a separate PR with a possible option

PR #3519, opened 2021-07-07
 Vertx4 instrumentation

laurit:
vertx-web instrumentation is split into two, vertx-http-client and vertx-web for the server. For vertx-web instrumentation for versions 3 & 4 is the same, only tests are a bit different. For vertx-http-client versions 3 & 4 instrumentation and tests are different.
This pr removed netty 4.0 & 4.1 `ChannelFutureListenerInstrumentation` which was responsible for creating connection failure span and instead uses `FutureListenerWrappers` for the same purpose. The reason behind this is that instrumentations using a matcher like `implementsInterface(named("io.netty.channel.ChannelFutureListener"))` don't apply to lambdas. For vert.x the listener we want to intercept happens to be a lambda.


iNikem:
@laurit Can you split this PR into independent chunks? 1K lines of code does not encourage a review :)

anuraaga:
@laurit Do you know if we can use the official instrumentation? I guess one of the goals of @iNikem's proposal for the external instrumentation registry is for us to use them too :)

https://github.com/eclipse-vertx/vertx-tracing/tree/master/vertx-opentelemetry

laurit:
@anuraaga I don't think it is going to be easy. While they seem to trace some extra things like sql client (https://github.com/eclipse-vertx/vertx-sql-client) and eventbus (https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/eventbus/EventBus.java) that we don't there are also some things that we do and they don't like propagating context to callbacks (didn't test just looked at code). Another issue is that currently instrumentation for all http clients produces spans with the same naming scheme and same attributes (ok this isn't really true, but at least we try to make them the same) their instrumentation will use a slightly different name and fills less attributes. Our instrumentation sanitizes sql queries and strips user info from urls, theirs doesn't.
Additionally their opentelmetry instrumentation was added in `4.1.0` I don't know whether it would work with `4.0.0`. When we start using their instrumentation we would also need to figure out how we are going to keep it up to date. Currently with couchbase, where we also use instrumentation provided by framework, we only update instrumentation when something breaks.

anuraaga:
These are all good and hard questions. But I think it goes to the core of open-telemetry to use upstream instrumentation when it's available and even encourage it. If we rewrite instrumentation, I think we're never going to achieve it. Is it possible to contribute changes upstream if we find things that don't work properly here? I think this is something we need to think properly about and not just put in code.

trask:
> These are all good and hard questions. But I think it goes to the core of open-telemetry to use upstream instrumentation when it's available and even encourage it. If we rewrite instrumentation, I think we're never going to achieve it. Is it possible to contribute changes upstream if we find things that don't work properly here? I think this is something we need to think properly about and not just put in code.

I added this topic to the 7/20 SIG agenda 👍

PR #3521, opened 2021-07-07
Trying to extract codegen plugin

iNikem:


buildSrc/build.gradle.kts
```diff
@@ -27,6 +27,8 @@ dependencies {
   implementation(gradleApi())
   implementation(localGroovy())
 
+  implementation("io.opentelemetry.instrumentation.gradle:codegen:1.4.0-SNAPSHOT")
```
anuraaga, line 30: Just in case it wasn't obvious, the dance we need to perform is 1) comment out this line 2) rename .gradle.kts.bak to .gradle.kts 3) ./gradlew :codegen:publishToMavenLocal 4) Undo 1+2

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
```diff
@@ -1,10 +1,15 @@
...
+  id("io.opentelemetry.instrumentation.javaagent-codegen")
 }
 
 dependencies {
   add("muzzleBootstrap", "io.opentelemetry.javaagent:opentelemetry-javaagent-bootstrap")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
+
+  add("codegen", "io.opentelemetry.javaagent:opentelemetry-javaagent-bootstrap")
+  add("codegen", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
```
mateuszrzeszutek, line 13: Is the `codegen` configuration still needed? The codegen plugin should only require the `extension-api` module, is it possible for it to depend on it directly?

codegen/src/main/java/io/opentelemetry/javaagent/muzzle/generation/collector/AdviceClassNameCollector.java
```diff
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.tooling.muzzle.collector;
+package io.opentelemetry.javaagent.muzzle.generation.collector;
```
mateuszrzeszutek, line 6: Nit:
```suggestion
package io.opentelemetry.javaagent.muzzle.codegen;
```
?

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModule.java
```diff
@@ -145,7 +145,7 @@ public boolean isHelperClass(String className) {
    * advices, grouped by {@link ClassRef#getClassName()}.
    *
    * <p>The actual implementation of this method is generated automatically during compilation by
-   * the {@code io.opentelemetry.javaagent.tooling.muzzle.collector.MuzzleCodeGenerationPlugin}
+   * the {@code io.opentelemetry.javaagent.muzzle.generation.collector.MuzzleCodeGenerationPlugin}
    * ByteBuddy plugin.
```
mateuszrzeszutek, line 149: WDYT about 
```suggestion
   * the {@code io.opentelemetry.instrumentation.javaagent-codegen} gradle plugin.
```
instead? It might be a better idea to point to the gradle plugin instead of one of its internal classes.

muzzle/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+
+group = "io.opentelemetry.javaagent"
+
+//val instrumentationMuzzle by configurations.creating {
+//  isCanBeConsumed = true
+//  isCanBeResolved = false
+//  extendsFrom(configurations.implementation.get())
+//}
+
+dependencies {
```
mateuszrzeszutek, line 14: This module is just for `InstrumentationClassPredicate`, right? Can we put it in the `extension-api` module instead?

iNikem: For now. When I extract MuzzleCheck it may host more. And I am not sure about putting them into client-facing module.

codegen/src/main/java/io/opentelemetry/javaagent/muzzle/generation/collector/AdviceClassNameCollector.java
iNikem: Using `io.opentelemetry.javaagent.muzzle.generation` so far

muzzle/build.gradle.kts
mateuszrzeszutek: Hmm, right, the muzzle check plugin will bring a lot of shared stuff...

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
iNikem: Removed

iNikem:
Asking for specific feedback:

- Name of the plugin module. Currently `javaagent-muzzle-generation`. Ok/Nok/better ideas?
- Id of the plugin. Currently `io.opentelemetry.instrumentation.javaagent-codegen`. Ok/Nok/better ideas?

javaagent-muzzle-generation/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+  id("com.gradle.plugin-publish")
+}
+
+group = "io.opentelemetry.instrumentation.gradle"
+
+dependencies {
+  implementation("com.google.guava:guava")
```
anuraaga, line 12: Sorry had a pending comment but forgot to hit submit. On the topic of a `codegen` configuration

I think we may still want the configuration

1) To add less to Gradle classpath. If we add too much it can cause dependency hell, a bit worried about Guava
2) To be able to substitute in the project dependencies in our project, not sure we'd have any way of doing it for a plugin transitive dependency which would be resolved way too early for a substitution to kick in I think.

I guess 1 could be resolved with shading anyways though. But 2) seems important and a requirement to allow using an older version of the plugin even as we work on the javaagent-extension-api.

javaagent-muzzle-generation/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+plugins {
```
anuraaga, line 1: Do we want the word muzzle in these projects? Assuming muzzle refers to our safety mechanism for applying instrumentation across versions, a lot of the codegen is independent of muzzle

javaagent-muzzle-generation/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+  implementation(project(":muzzle"))
+  implementation("net.bytebuddy:byte-buddy-gradle-plugin")
+  implementation("io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
+}
+
+pluginBundle {
+  website = "https://opentelemetry.io"
+  vcsUrl = "https://github.com/open-telemetry/opentelemetry-java-instrumentation"
+  tags = listOf("opentelemetry", "instrumentation")
+}
```
anuraaga, line 22: We'll need that `plugins` snippet from my draft PR when publishing to portal so may as well add it now right?

javaagent-muzzle-generation/src/main/java/io/opentelemetry/javaagent/muzzle/generation/Utils.java
```diff
@@ -0,0 +1,20 @@
+package io.opentelemetry.javaagent.muzzle.generation;
+
+class Utils {
```
anuraaga, line 3:
```suggestion
final class Utils {
```

javaagent-muzzle-generation/src/main/java/io/opentelemetry/javaagent/muzzle/generation/Utils.java
```diff
@@ -0,0 +1,20 @@
...
+  /** com.foo.Bar to com/foo/Bar */
+  static String getInternalName(Class<?> clazz) {
+    return clazz.getName().replace('.', '/');
+  }
+
+  /** com.foo.Bar to com/foo/Bar.class */
+  static String getResourceName(String className) {
+    return className.replace('.', '/') + ".class";
+  }
+
```
anuraaga, line 19:
```suggestion

  private Utils() {}
```

javaagent-muzzle-generation/build.gradle.kts
iNikem: At the same time self-contained plugin is good for external users. They don't need to know about "codegen" dependency and that something has to be added to it.

javaagent-muzzle-generation/build.gradle.kts
iNikem: This plugin generates bodies for `InstrumentationModule` methods with word `muzzle` in method names. So currently it is certainly muzzle-related.

javaagent-muzzle-generation/build.gradle.kts
anuraaga: Yeah they wouldn't need to with the configuration either since we configure it with defaults here. Before this PR our repo is substituting to use our project dependencies but it's not required to care otherwise.

iNikem: I am afraid I don't quite understand what do you suggest? Do you mean to create and use `codegen` configuration already here, right in the module? Or how?

anuraaga: I propose using the same pattern we have

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts#L33

This as all the io.openteletry.* ones are meant to be published for downstream. Though most will get this one transitively via io.opentelemetry.instrumentation.javaagent-instrumentation.

The configuration is set here when the plugin is included so most users won't need to worry about it.

anuraaga: Sorry by here I meant there :P I mean the config is set in the javaagent-testing plugin

anuraaga: Also makes me realize this project shouldn't be called muzzle or codegen or anything - let's call it `gradle-plugins`. We have a few plugins to publish.

iNikem: > Also makes me realize this project shouldn't be called muzzle or codegen or anything - let's call it `gradle-plugins`. We have a few plugins to publish.

Which ones? I am fine with publishing muzzle generation and checking as one plugin. But I am not sure we should bundle all our plugins into one artifact?

javaagent-muzzle-generation/build.gradle.kts
iNikem: Which PR? :) Sorry, you are too laconic for me recently :(

javaagent-muzzle-generation/build.gradle.kts
iNikem: > I propose using the same pattern we have
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts#L33
> 
> This as all the io.openteletry.* ones are meant to be published for downstream. Though most will get this one transitively via io.opentelemetry.instrumentation.javaagent-instrumentation.
> 
> The configuration is set here when the plugin is included so most users won't need to worry about it.

Do I understand you correctly?

- Published `io.opentelemetry.instrumentation.javaagent-codegen` plugin does not have any declared runtime dependencies
- Actual dependencies, such as guava and `javaagent-extension-api` are added by `io.opentelemetry.instrumentation.javaagent-instrumentation` plugin
- End users are supposed to use `io.opentelemetry.instrumentation.javaagent-instrumentation` plugin when it gets published

anuraaga: Yeah that understanding sounds correct to me.

>  But I am not sure we should bundle all our plugins into one artifact?

I don't think there's any real downside - anyways we'll publish to Gradle Plugin Portal so users just add the plugin to their project and don't care how the artifact is structured. So it allows us to keep all plugin publishing type of logic together.

javaagent-muzzle-generation/build.gradle.kts
anuraaga: Sorry for my smartphone replies, somehow GitHub app still doesn't support linking to lines of code. And now that I look for it I can't find it in the PR I sent, that's definitely too laconic...

```
gradlePlugin {
    plugins { 
        named("io.opentelemetry.instrumentation.javaagent-codegen") { 
            displayName = "<short displayable name for plugin>" 
            description = "<Good human-readable description of what your plugin is about>" 
        }
    }
}
```

https://docs.gradle.org/current/userguide/publishing_gradle_plugins.html#configure_the_plugin_publishing_plugin

One thing I found is for precompiled script plugin, we leave out `id` and `implementationClass` and it seems to work ok

javaagent-muzzle-generation/build.gradle.kts
iNikem: First, we don't actually HAVE to publish to Gradle Plugin Portal. Plugins can be consumed just from Maven Central as well.

Second, ok, let's give it a try...

iNikem:
For the reference: here is an example of using this new plugin in an external project: https://github.com/signalfx/splunk-otel-java/compare/muzzle-codegen?expand=1. Only one module was verified to work for now.

I am concerned about using `codegen` configuration in that module's build script.

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
```diff
@@ -1,10 +1,13 @@
 plugins {
   id("io.opentelemetry.instrumentation.javaagent-testing")
   id("io.opentelemetry.instrumentation.muzzle-check")
+  id("io.opentelemetry.instrumentation.javaagent-codegen")
```
anuraaga, line 4: Seems to come in transitively (though doesn't hurt at all either)

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -6,8 +6,8 @@ plugins {
   id("net.bytebuddy.byte-buddy")
 
   id("io.opentelemetry.instrumentation.base")
-  id("io.opentelemetry.instrumentation.javaagent-codegen")
   id("io.opentelemetry.instrumentation.javaagent-shadowing")
+  id("io.opentelemetry.instrumentation.javaagent-codegen")
```
anuraaga, line 10: Did you have an ordering issue? We can fix it later, but good not to have ordering constraints so just checking

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/MuzzleCodeGenerationPlugin.java
```diff
@@ -0,0 +1,62 @@
...
+import net.bytebuddy.dynamic.DynamicType;
+
+/**
+ * This class is a ByteBuddy build plugin that is responsible for generating actual implementation
+ * of some {@link InstrumentationModule} methods. Auto-generated methods have the word "muzzle" in
+ * their names.
+ *
+ * <p>This class is used in the gradle build scripts, referenced by each instrumentation module.
+ */
+public class MuzzleCodeGenerationPlugin implements Plugin {
```
anuraaga, line 23:
```suggestion
public final class MuzzleCodeGenerationPlugin implements Plugin {
```

Unless you know a use case for extending it

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/MuzzleCodeGenerationPlugin.java
```diff
@@ -0,0 +1,62 @@
...
+  public MuzzleCodeGenerationPlugin(URLClassLoader classLoader) {
+    this.classLoader = classLoader;
+  }
+
+  @Override
+  public boolean matches(TypeDescription target) {
+    if (target.isAbstract()) {
+      return false;
+    }
+    // AutoService annotation is not retained at runtime. Check for InstrumentationModule supertype
```
anuraaga, line 39: Nit: Nor can we know if a downstream instrumentation is using AutoService anyways, so this comment feels like unneeded noise

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/MuzzleCodeGenerator.java
```diff
@@ -0,0 +1,554 @@
...
+
+/**
+ * This class generates the actual implementation of the {@link
+ * InstrumentationModule#getMuzzleReferences()} method. It collects references from all advice
+ * classes defined in an instrumentation and writes them as Java bytecode in the generated {@link
+ * InstrumentationModule#getMuzzleReferences()} method.
+ *
+ * <p>This class is run at compile time by the {@link MuzzleCodeGenerationPlugin} ByteBuddy plugin.
+ */
+class MuzzleCodeGenerator implements AsmVisitorWrapper {
```
anuraaga, line 46:
```suggestion
final class MuzzleCodeGenerator implements AsmVisitorWrapper {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/MuzzleCompilationException.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.muzzle.generation;
+
+class MuzzleCompilationException extends RuntimeException {
```
anuraaga, line 8:
```suggestion
final class MuzzleCompilationException extends RuntimeException {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollectingClassVisitor.java
```diff
@@ -0,0 +1,567 @@
...
+import net.bytebuddy.jar.asm.Opcodes;
+import net.bytebuddy.jar.asm.Type;
+
+/** Visit a class and collect all references made by the visited class. */
+// Additional things we could check
+// - annotations on class
+// - outer class
+// - inner class
+// - cast opcodes in method bodies
+class ReferenceCollectingClassVisitor extends ClassVisitor {
```
anuraaga, line 37:
```suggestion
final class ReferenceCollectingClassVisitor extends ClassVisitor {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * {@link LinkedHashMap} is used for reference map to guarantee a deterministic order of iteration,
+ * so that bytecode generated based on it would also be deterministic.
+ *
+ * <p>This class is only called at compile time by the {@link MuzzleCodeGenerationPlugin} ByteBuddy
+ * plugin.
+ */
+class ReferenceCollector {
```
anuraaga, line 49:
```suggestion
final class ReferenceCollector {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+
+  private final Map<String, ClassRef> references = new LinkedHashMap<>();
+  private final MutableGraph<String> helperSuperClassGraph = GraphBuilder.directed().build();
+  private final Map<String, String> contextStoreClasses = new LinkedHashMap<>();
+  private final Set<String> visitedClasses = new HashSet<>();
+  private final InstrumentationClassPredicate instrumentationClassPredicate;
+  private final ClassLoader resourceLoader;
+
+  // only used by tests
+  public ReferenceCollector(Predicate<String> libraryInstrumentationPredicate) {
```
anuraaga, line 59:
```suggestion
 ReferenceCollector(Predicate<String> libraryInstrumentationPredicate) {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+    this(libraryInstrumentationPredicate, ReferenceCollector.class.getClassLoader());
+  }
+
+  public ReferenceCollector(
```
anuraaga, line 63:
```suggestion
 ReferenceCollector(
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * If passed {@code resource} path points to an SPI file (either Java {@link
+   * java.util.ServiceLoader} or AWS SDK {@code ExecutionInterceptor}) reads the file and adds every
+   * implementation as a reference, traversing the graph of classes until a non-instrumentation
+   * (external) class is encountered.
+   *
+   * @param resource path to the resource file, same as in {@link ClassLoader#getResource(String)}
+   * @see InstrumentationClassPredicate
+   */
+  public void collectReferencesFromResource(String resource) {
```
anuraaga, line 79:
```suggestion
 void collectReferencesFromResource(String resource) {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+   * Traverse a graph of classes starting from {@code adviceClassName} and collect all references to
+   * both internal (instrumentation) and external classes.
+   *
+   * <p>The graph of classes is traversed until a non-instrumentation (external) class is
+   * encountered.
+   *
+   * @param adviceClassName Starting point for generating references.
+   * @see InstrumentationClassPredicate
+   */
+  public void collectReferencesFromAdvice(String adviceClassName) {
```
anuraaga, line 124:
```suggestion
 void collectReferencesFromAdvice(String adviceClassName) {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+      helperSuperClassGraph.addNode(helperClass);
+    }
+    if (!isAdviceClass) {
+      for (String helperSuperClass : helperSuperClasses) {
+        helperSuperClassGraph.putEdge(className, helperSuperClass);
+      }
+    }
+  }
+
+  public Map<String, ClassRef> getReferences() {
```
anuraaga, line 209:
```suggestion
 Map<String, ClassRef> getReferences() {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+    return references;
+  }
+
+  public void prune() {
```
anuraaga, line 213:
```suggestion
 void prune() {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+    for (String type : reference.getInterfaceNames()) {
+      if (hasLibrarySuperType(type)) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  // see https://en.wikipedia.org/wiki/Topological_sorting#Kahn's_algorithm
+  public List<String> getSortedHelperClasses() {
```
anuraaga, line 303:
```suggestion
 List<String> getSortedHelperClasses() {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
```diff
@@ -0,0 +1,339 @@
...
+    Queue<String> helpersWithNoDeps = new LinkedList<>();
+    for (String helperClass : dependencyGraph.nodes()) {
+      if (dependencyGraph.predecessors(helperClass).isEmpty()) {
+        helpersWithNoDeps.add(helperClass);
+      }
+    }
+    return helpersWithNoDeps;
+  }
+
+  public Map<String, String> getContextStoreClasses() {
```
anuraaga, line 336:
```suggestion
 Map<String, String> getContextStoreClasses() {
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/ReferenceCollector.java
anuraaga: Guess class went from public to package private

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/collector/MuzzleCompilationException.java
```diff
@@ -5,8 +5,8 @@
 
 package io.opentelemetry.javaagent.tooling.muzzle.collector;
 
-public class MuzzleCompilationException extends RuntimeException {
-  public MuzzleCompilationException(String message) {
+class MuzzleCompilationException extends RuntimeException {
```
anuraaga, line 8:
```suggestion
final class MuzzleCompilationException extends RuntimeException {
```

anuraaga (approved):
Just nits, thanks!

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
iNikem: Transitivity via `javaagent-testing` is very obscure. I prefer to have obvious dependencies.

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
iNikem: No issues. I first removed that but it turned out still required so I put it back.

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
```diff
@@ -1,10 +1,13 @@
...
 }
 
 dependencies {
   add("muzzleBootstrap", "io.opentelemetry.javaagent:opentelemetry-javaagent-bootstrap")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
+
+  add("codegen", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
```
mateuszrzeszutek, line 12: Is tooling needed for codegen?

examples/extension/build.gradle
```diff
@@ -11,6 +11,8 @@ plugins {
   See https://imperceptiblethoughts.com/shadow/ for more details about Shadow plugin.
    */
   id "com.github.johnrengelman.shadow" version "6.1.0"
+
+  id "io.opentelemetry.instrumentation.javaagent-codegen" version "0.1.0"
```
mateuszrzeszutek, line 15: Is 0.1.0 some temporary released version? Are we going to version gradle plugins the same way we do all other artficats?

mateuszrzeszutek (approved):

examples/extension/build.gradle
iNikem: Currently this module is totally independent from the main build. I don't have a strong opinion on its versioning atm

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
iNikem: Apparently not right now...

PR #3527, opened 2021-07-08
Clean up AgentElementMatchers

mateuszrzeszutek:
I realized that one thing I didn't look at all when extracting `javaagent-extension-api` was our matchers. This PR cleans up that class, adds some Javadocs, move some matcher implementations used internally to the tooling module, and moved all static factory method (for classloader matchers too) to `AgentElementMatchers`.


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/ClassLoaderHasClassesNamedMatcher.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private final Cache<ClassLoader, Boolean> cache =
+      Cache.newBuilder().setWeakKeys().setMaximumSize(25).build();
+
+  private final String[] resources;
+
+  ClassLoaderHasClassesNamedMatcher(String... classNames) {
+    resources = classNames;
+    for (int i = 0; i < resources.length; i++) {
+      resources[i] = resources[i].replace(".", "/") + ".class";
```
iNikem, line 23: There is `io.opentelemetry.javaagent.tooling.Utils#getResourceName` if you wish

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/ClassLoaderHasClassesNamedMatcher.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+    ClassLoaderMatcherCacheHolder.addCache(cache);
+  }
+
+  private boolean hasResources(ClassLoader cl) {
```
iNikem, line 28: Move it below its usage?

iNikem (approved):

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/ClassLoaderHasClassesNamedMatcher.java
mateuszrzeszutek: `Utils` is in `javaagent-tooling`, it's not accessible here.

trask (approved):

PR #3532, opened 2021-07-09
Merge Java TraceUtils into extension and switch to Instrumenter API.

anuraaga:
TraceUtils needs to use global OpenTelemetry though we'd prefer to use the test runner's.


mateuszrzeszutek (approved):

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -96,11 +96,11 @@ afterEvaluate {
 
       // If agent depends on some shared instrumentation module that is not a testing module, it will
       // be packaged into the testing jar so we need to make sure to exclude from the test classpath.
-      val libPath = it.absolutePath
-      val instrumentationPath = file("${rootDir}/instrumentation/").absolutePath
```
anuraaga, line 100: Didn't have trailing slash so `instrumentation-api` got excluded too. Finally hit this bug because javaagent test classloads the instrumenter API after this change.

PR #3533, opened 2021-07-09
Add a small test utility that mimics Spock's 'cleanup:' block

mateuszrzeszutek:
I thought that it might be useful to have a thing like that, instead of giant try-finally blocks spanning whole test methods.


testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/AutoCleanupExtension.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Queue;
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * A small utility extension that allows deferring executing cleanup code in {@code @Test} methods
+ * until the test has finished. All cleanup callbacks added during the test will always be executed
+ * after it finishes, no matter the outcome. Inspired by Spock's {@code cleanup:} block.
+ */
+public final class AutoCleanupExtension implements AfterEachCallback {
```
anuraaga, line 20: I don't think we'd want to publish a helper that's not really related to instrumentation to end users, but we could split out, or maybe inline, later. What do you think?

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/AutoCleanupExtension.java
```diff
@@ -0,0 +1,56 @@
...
+  private final Queue<AutoCloseable> thingsToCleanUp = new LinkedList<>();
```
anuraaga, line 21:
```suggestion
  private final Queue<AutoCloseable> thingsToCleanUp = new ConcurrentLinkedQueue<>();
```

mateuszrzeszutek: Hm, I think JUnit creates a new test class instance for each test method and there's hardly any need to care about thread safety, but this can't hurt anyway.

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/AutoCleanupExtension.java
mateuszrzeszutek: Hmm, good point - I'll move it to an `internal` package for now, we could split out a "shared internal testing code" module later if there's any need.

anuraaga (approved):

iNikem (approved):

PR #3534, opened 2021-07-09
Use async requests for high concurrency test.

anuraaga:
This can allow for even more concurrency than 4, as many as the server will take. What do you think?


mateuszrzeszutek (approved):

laurit (approved):

iNikem:
@anuraaga Is there now a guarantee that requests will actually be executed in parallel and not sequentially due to some quirk of the client library?

anuraaga:
@iNikem Yeah it's how non-blocking clients work. Possibly even more expected to be parallel without having to rely on OS thread scheduling.

Looking at the request logs saw a good amount of req req req resp resp resp type of logging

PR #3535, opened 2021-07-09
Inline one usage of basicSpan

anuraaga:
While thinking about assertions, I was thinking about what to do with `basicSpan`. I want to propose removing it - there is no concept of a "basic span" in OpenTelemetry so it's not obvious what it's actually testing. It also seems to have a very small effect on boilerplate, basicSpan is almost all just if statements wrapping optional arguments. What do you think?


mateuszrzeszutek (approved):
I agree - it only seems to check name & kind (sometimes the parent span), inlining it does not add much code. And I'd even argue that it makes the assertion a little bit clearer, because you don't have to remember which param means what but can clearly see  it in the assert.

trask (approved):
👍

Issue #3531, opened 2021-07-09
RocketMQ consumer‘s trace is lost when other methods are called 


EricGreen239:
When I call another method after receiving a message in RocketMQ's consumer listener method, the trace information is not carried, resulting in two Traces exported to Jaeger.

opentelemetry-java-instrumentation javaagent version: 1.2.0
RockerMQ version: 4.9
RocketMQ client maven:
        <dependency>
            <groupId>org.apache.rocketmq</groupId>
            <artifactId>rocketmq-client</artifactId>
            <version>4.8.0</version>
        </dependency>

Consumer code:

```
@Service
@Slf4j
public class ConsumerService {

    @Autowired
    private DubboHealthService healthService;

    @PostConstruct
    public void start (){
            try {
                DefaultMQPushConsumer consumer = new DefaultMQPushConsumer("test-group1");
                // consumer.setConsumeFromWhere(ConsumeFromWhere.CONSUME_FROM_FIRST_OFFSET);
                consumer.setNamesrvAddr("myRocketMQIp:9876");

                consumer.subscribe("TopicTest", "*");

                consumer.registerMessageListener(new MessageListenerConcurrently() {

                    @Override
                    public ConsumeConcurrentlyStatus consumeMessage(List<MessageExt> msgs,
                                                                    ConsumeConcurrentlyContext context) {
                        MessageExt messageExt = msgs.get(0);
                        log.info("consumer message "+new String(messageExt.getBody()));
                        healthService.redis();
                        return ConsumeConcurrentlyStatus.CONSUME_SUCCESS;
                    }
                });

                consumer.start();

                log.info("Consumer Started.%n");
            }catch (Exception e){
                e.printStackTrace();
            }
    }
}
```

DubboHealthService.redis() code:

```
@Override
    public String redis() {
        // Query the data of redis whose key is name through jedis 
        log.info("redis");
        redis.set("name", "lucy");
        return redis.get("name");
    }
```

![image](https://user-images.githubusercontent.com/19650475/125025228-dfe51700-e0b4-11eb-87ac-757274f96df4.png)

After receiving the message, this trace has only one consumer span on Jaeger, and no Redis-related span, which is already on a new trace, indicating that this is a request, but there are two traces

![image](https://user-images.githubusercontent.com/19650475/125025500-597d0500-e0b5-11eb-93b6-105e3ff5f573.png)

[...]

PR #3537, opened 2021-07-09
Put RocketMQ's consumer span in current context

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3531

### DIFF
--- a/instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
+++ b/instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
@@ -28,7 +28,12 @@ final class TracingConsumeMessageHookImpl implements ConsumeMessageHook {
       return;
     }
     Context otelContext = tracer.startSpan(Context.current(), context.getMsgList());
-    context.setMqTraceContext(ContextAndScope.create(otelContext));
+
+    // it's safe to store the scope in the rocketMq trace context, both before() and after() methods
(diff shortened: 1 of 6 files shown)


instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/TracingConsumeMessageHookImpl.java
```diff
@@ -28,17 +28,18 @@ public void consumeMessageBefore(ConsumeMessageContext context) {
       return;
     }
     Context otelContext = tracer.startSpan(Context.current(), context.getMsgList());
-    context.setMqTraceContext(otelContext);
+    context.setMqTraceContext(ContextAndScope.create(otelContext));
```
anuraaga, line 31: I have some recollection of this possibly hopping threads in before / after but could be mixing up with some other instrumentation. Can you check the rocketmq code to verify? I think anytime we make a scope current in this sort of pattern that doesn't follow a try/resources form, we might do best to actually link to the library's implementation code to give a pointer to the safety of the threadlocal.

I would also not `makeCurrent` in the `create` method to show the threadlocal activation and deactivation more clearly in one file at least

mateuszrzeszutek: Good point; I've checked that in the sources and it always calls both before() and after() listener methods from the same thread (same method, even).

anuraaga (approved):

iNikem (approved):

PR #3538, opened 2021-07-09
Convert OkHttp 3 to Instrumenter API

mateuszrzeszutek:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -129,6 +129,11 @@
         InstrumenterConstructor.propagatingToDownstream(setter), SpanKindExtractor.alwaysClient());
   }
 
+  /** Returns a new {@link Instrumenter} which will create client spans. */
+  public Instrumenter<REQUEST, RESPONSE> newClientInstrumenter() {
```
anuraaga, line 133: I think `newInstrumenter(SpanKindExtractor)` is fine for the rare case of an instrumenter that only generates client spans without propagation, or otherwise the semantics of the overloads for `newClientInstrumenter` become too different

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/OkHttpAttributesExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpAttributesExtractor;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import okhttp3.Request;
+import okhttp3.Response;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+final class OkHttpAttributesExtractor extends HttpAttributesExtractor<Request, Response> {
+  @Override
+  protected @Nullable String method(Request request) {
```
anuraaga, line 16: Think some of these aren't `@Nullable` (sure about `request.url.toString()` at least :)

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/OkHttpAttributesExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  protected @Nullable String scheme(Request request) {
+    return request.url().scheme();
+  }
+
+  @Override
+  protected @Nullable String userAgent(Request request) {
+    return request.header("User-Agent");
```
anuraaga, line 47:
```suggestion
    return request.header("user-agent");
```

Probably a good pattern to use lowercase to make sure it's not case sensitive

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
mateuszrzeszutek: Excellent point, removed.

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/OkHttpAttributesExtractor.java
mateuszrzeszutek: Ouch, auto-generated method stubs... removed, thanks!

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/OkHttpAttributesExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  protected String scheme(Request request) {
+    return request.url().scheme();
+  }
+
+  @Override
+  protected @Nullable String userAgent(Request request) {
+    return request.header("user-agent");
```
trask, line 47:
```suggestion
    // using lowercase header name intentionally to ensure extraction is not case-sensitive
    return request.header("user-agent");
```

instrumentation/okhttp/okhttp-3.0/testing/src/main/groovy/io/opentelemetry/instrumentation/okhttp/v3_0/AbstractOkHttp3Test.groovy
```diff
@@ -68,6 +70,24 @@ abstract class AbstractOkHttp3Test extends HttpClientTest<Request> {
...
 
+  @Override
+  Set<AttributeKey<?>> httpAttributes(URI uri) {
+    Set<AttributeKey<?>> extra = [
+      SemanticAttributes.HTTP_HOST,
+      SemanticAttributes.HTTP_SCHEME
+    ]
+    def attributes = super.httpAttributes(uri) + extra
+
+    // flavor is extracted from the response, and those URLs cause exceptions (= null response)
```
trask, line 81: 👍

trask (approved):

instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3Singletons.java
```diff
@@ -0,0 +1,25 @@
...
+import io.opentelemetry.javaagent.instrumentation.api.instrumenter.PeerServiceAttributesExtractor;
+import okhttp3.Interceptor;
+
+/** Holder of singleton interceptors for adding to instrumented clients. */
+public final class OkHttp3Singletons {
+
+  public static final Interceptor TRACING_INTERCEPTOR =
+      OkHttpTracing.newBuilder(GlobalOpenTelemetry.get())
+          .addAttributesExtractor(
+              PeerServiceAttributesExtractor.create(new OkHttpNetAttributesExtractor()))
```
jkwatson, line 20: Is there a reason why we wouldn't want these to be included by default, in library instrumentation usages?

anuraaga (approved):

instrumentation/okhttp/okhttp-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3Singletons.java
jkwatson: Is that the mapping between ip and name? It might be nice to expose that option to the library instrumentation as well in the future.

trask: I think the idea was that library instrumentation users have more convenient ways to set `peer.service` than using a global map, since they can customize interceptors/extractors at different usage points

jkwatson: true, but they would have to understand the otel convention around peer.service to even know that it's a thing to set.

trask: opened #3560 to track since this PR is just following existing convention

PR #3539, opened 2021-07-09
Add noop sdk configuration option

### LINKED ISSUE #3053: Provide a configuration option to install a true no-op SDK implementation.
Performance overhead testing: it would be very nice to be able to distinguish overhead between overhead caused by the OpenTelemetry SDK & the Instrumentation provided by the agent. 

If there were a configuration option to utilize the newly minted "true no-op" SDK (https://github.com/open-telemetry/opentelemetry-java/tree/main/extensions/noop-api) rather than the full SDK, then overhead could be compared when running in the no-op mode vs. the full mode, and hence get the overhead introduced by the instrumentation code.

Furthermore, if there were also an option to not install an SDK at all, but use the standard API implementation, then you could get an idea of the overhead due to the API vs the usage of the API.

I guess you could build a custom agent for these purposes, but having this as an option for the agent would be useful.

**Additional context**
It would also be great to run the iguanodon in the various modes to automate the overhead measurements.

Enkelian:
Resolves #3053
Added new configuration option: `otel.javaagent.experimental.sdk.noop` that allows using no-op SDK. If set true, no traces or metrics are being exported.

When the dependency on `io.opentelemetry:opentelemetry-extension-noop-api` is present in `javaagent-tooling`, `NoopContextStorageProvider` is registered and used, regardless of its configuration. To avoid such behaviour, enforcing the `default` value of `io.opentelemetry.context.contextStorageProvider` is necessary.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
@@ -6,6 +6,8 @@
 package io.opentelemetry.javaagent.tooling;
 
 import com.google.auto.service.AutoService;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.extension.noopapi.NoopOpenTelemetry;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.javaagent.extension.AgentListener;
 import io.opentelemetry.javaagent.instrumentation.api.OpenTelemetrySdkAccess;
@@ -23,6 +25,7 @@ public class OpenTelemetryInstaller implements AgentListener {
   private static final Logger logger = LoggerFactory.getLogger(OpenTelemetryInstaller.class);
 
   static final String JAVAAGENT_ENABLED_CONFIG = "otel.javaagent.enabled";
+  static final String JAVAAGENT_NOOP_CONFIG = "otel.javaagent.experimental.sdk.noop";
 
   @Override
   public void beforeAgent(Config config) {
@@ -37,16 +40,24 @@ public class OpenTelemetryInstaller implements AgentListener {
   @SuppressWarnings("unused")
   public static synchronized void installAgentTracer(Config config) {
     if (config.getBooleanProperty(JAVAAGENT_ENABLED_CONFIG, true)) {
+
+      System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
+
       copySystemProperties(config);
 
-      OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
-      OpenTelemetrySdkAccess.internalSetForceFlush(
-          (timeout, unit) -> {
-            CompletableResultCode traceResult = sdk.getSdkTracerProvider().forceFlush();
-            CompletableResultCode flushResult = IntervalMetricReader.forceFlushGlobal();
-            CompletableResultCode.ofAll(Arrays.asList(traceResult, flushResult))
-                .join(timeout, unit);
-          });
+      if (config.getBooleanProperty(JAVAAGENT_NOOP_CONFIG, false)) {
+        GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
+      } else {
+        OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
+        OpenTelemetrySdkAccess.internalSetForceFlush(
+            (timeout, unit) -> {
+              CompletableResultCode traceResult = sdk.getSdkTracerProvider().forceFlush();
+              CompletableResultCode flushResult = IntervalMetricReader.forceFlushGlobal();
+              CompletableResultCode.ofAll(Arrays.asList(traceResult, flushResult))
+                  .join(timeout, unit);
+            });
+      }
+
     } else {
       logger.info("Tracing is disabled.");
     }
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
@@ -25,7 +25,7 @@ public class OpenTelemetryInstaller implements AgentListener {
   private static final Logger logger = LoggerFactory.getLogger(OpenTelemetryInstaller.class);
 
   static final String JAVAAGENT_ENABLED_CONFIG = "otel.javaagent.enabled";
-  static final String JAVAAGENT_NOOP_CONFIG = "otel.javaagent.experimental.sdk.noop";
+  static final String JAVAAGENT_NOOP_CONFIG = "otel.javaagent.experimental.use-noop-api";
 
   @Override
   public void beforeAgent(Config config) {
@@ -41,13 +41,13 @@ public class OpenTelemetryInstaller implements AgentListener {
   public static synchronized void installAgentTracer(Config config) {
     if (config.getBooleanProperty(JAVAAGENT_ENABLED_CONFIG, true)) {
 
-      System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
-
       copySystemProperties(config);
 
       if (config.getBooleanProperty(JAVAAGENT_NOOP_CONFIG, false)) {
         GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
       } else {
+        System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
+
         OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
         OpenTelemetrySdkAccess.internalSetForceFlush(
             (timeout, unit) -> {
--- a/javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/tooling/OpenTelemetryInstallerTest.groovy
+++ b/javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/tooling/OpenTelemetryInstallerTest.groovy
@@ -33,7 +33,7 @@ class OpenTelemetryInstallerTest extends Specification {
     .build()
 
     when:
-    def otelInstaller = new OpenTelemetryInstaller();
+    def otelInstaller = new OpenTelemetryInstaller()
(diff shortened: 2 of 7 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -34,11 +37,19 @@ public void beforeAgent(Config config) {
   @SuppressWarnings("unused")
   public static synchronized void installAgentTracer(Config config) {
     if (config.getBooleanProperty(JAVAAGENT_ENABLED_CONFIG, true)) {
+
+      System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
+
       copySystemProperties(config);
```
mateuszrzeszutek, line 43: Shouldn't that be moved to the `!= noop` branch? There's no need to copy SDK config if you don't use the SDK at all

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -20,6 +22,7 @@
   private static final Logger logger = LoggerFactory.getLogger(OpenTelemetryInstaller.class);
 
   static final String JAVAAGENT_ENABLED_CONFIG = "otel.javaagent.enabled";
+  static final String JAVAAGENT_NOOP_CONFIG = "otel.javaagent.experimental.sdk.noop";
```
mateuszrzeszutek, line 25: How about
```suggestion
  static final String JAVAAGENT_NOOP_CONFIG = "otel.javaagent.experimental.use-noop-api";
```
? It's noop API, not SDK (see the module name), so we can reflect that in the property name.

javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/tooling/OpenTelemetryInstallerTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+  def "should initialize noop"(){
+
+    given:
+    def config = Mock(Config)
+    config.getBooleanProperty(OpenTelemetryInstaller.JAVAAGENT_NOOP_CONFIG, false) >> true
+    config.getBooleanProperty(OpenTelemetryInstaller.JAVAAGENT_ENABLED_CONFIG, true) >> true
+    config.getAllProperties() >> ["otel.javaagent.enabled":"true", "otel.javaagent.experimental.sdk.noop":"true"]
+
+    when:
+    OpenTelemetryInstaller.installAgentTracer(config)
```
mateuszrzeszutek, line 34: Consider creating a new `OpenTelemetryInstaller` instance and calling the `beforeAgent()` instance method.
I'm not even sure why there's a static public method here, it shouldn't be needed at all.

javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/tooling/OpenTelemetryInstallerTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+
+  def "should initialize noop"(){
+
+    given:
+    def config = Mock(Config)
+    config.getBooleanProperty(OpenTelemetryInstaller.JAVAAGENT_NOOP_CONFIG, false) >> true
+    config.getBooleanProperty(OpenTelemetryInstaller.JAVAAGENT_ENABLED_CONFIG, true) >> true
+    config.getAllProperties() >> ["otel.javaagent.enabled":"true", "otel.javaagent.experimental.sdk.noop":"true"]
```
mateuszrzeszutek, line 31: Nit: `Config` is a fairly simple type, you can create a new instance of it instead of mocking it:
```suggestion
    def config = Config.newBuilder()
        .readProperties([
            (OpenTelemetryInstaller.JAVAAGENT_NOOP_CONFIG): "false",
            (OpenTelemetryInstaller.JAVAAGENT_ENABLED_CONFIG): "true"
        ])
        .build()
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -34,11 +37,19 @@ public void beforeAgent(Config config) {
   @SuppressWarnings("unused")
   public static synchronized void installAgentTracer(Config config) {
     if (config.getBooleanProperty(JAVAAGENT_ENABLED_CONFIG, true)) {
+
+      System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
```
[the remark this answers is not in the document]
Enkelian: Adding a dependency on `io.opentelemetry:opentelemetry-extension-noop-api` resulted in `NoopContextStorageProvider` being used instead of the default provider. 
I know it's not the best way to fix it, but it's the only way I could make that work so far.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/NoopSdkSmokeTest.groovy
```diff
@@ -0,0 +1,58 @@
...
+class NoopSdkSmokeTest extends SmokeTest {
+
+  protected String getTargetImage(String jdk) {
+    "ghcr.io/open-telemetry/java-test-containers:smoke-springboot-jdk$jdk-20210218.577304949"
+  }
+
+  @Override
+  protected Map<String, String> getExtraEnv() {
+    return [
+      "otel.javaagent.experimental.sdk.noop": "true"
```
mateuszrzeszutek, line 25: Hmm, these are environment variables, so 
```suggestion
      "OTEL_JAVAAGENT_EXPERIMENTAL_SDK_NOOP": "true"
```
would be the preferred naming pattern

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/LibraryTestRunner.java
```diff
@@ -36,6 +36,8 @@
   private static final LibraryTestRunner INSTANCE = new LibraryTestRunner();
 
   static {
+    System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
```
mateuszrzeszutek, line 39: Hmm, why is it needed here? Does the noop-api get picked somehow?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
Enkelian: yep, it's only useful there - thanks!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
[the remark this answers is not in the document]
iNikem: @jkwatson Is this side-effect expected?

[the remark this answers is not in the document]
jkwatson: @anuraaga might be a better person to answer this, but the intent with the true no-op API/SDK was indeed to have it be a true no-op, so the context would also be a no-op. Whether that's desired here or not is a separate question, I guess.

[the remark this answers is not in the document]
iNikem: @Enkelian Why do want default context storage with noop API/SDK then? :) Do you want "no-op" or "some-op"? :)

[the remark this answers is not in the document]
Enkelian: Default configuration of context storage is used when noop is not enabled (see current version, I've moved this line after @mateuszrzeszutek review). This line is necessary because the presence of noop extension dependency in gradle resulted in `NoopContextStorageProvider` being used in every case - even when noop configuration was not being set up.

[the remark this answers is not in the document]
iNikem: > the presence of noop extension dependency in gradle resulted in `NoopContextStorageProvider` being used in every case - even when noop configuration was not being set up.

A! That sounds like a bug. Or at least an unintended side effect that needs fixing.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -34,11 +37,19 @@ public void beforeAgent(Config config) {
...
+
       copySystemProperties(config);
 
-      OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
-      OpenTelemetrySdkAccess.internalSetForceFlush(
-          (timeout, unit) -> sdk.getSdkTracerProvider().forceFlush().join(timeout, unit));
+      if (config.getBooleanProperty(JAVAAGENT_NOOP_CONFIG, false)) {
+        GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
+      } else {
+        System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
```
iNikem, line 46: Please create a new issue about this unwanted side effect.

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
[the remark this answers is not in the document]
anuraaga: Yeah we might need that.

Just for reference, the noop API artifact is still alpha precisely in case we need to make any changes to support its use cases so we can make any fixes we need to it.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
Enkelian: Created #3566 :)

mateuszrzeszutek (approved):

mateuszrzeszutek:
@Enkelian can you rebase with the `main` branch? It seems you have some conflicts

Issue #3517, opened 2021-07-07
agent opentelemetry-all 1.3.1 Hibernate auto instrumentation not work


rbelfils:
I run my petclinic with this command

java -Dotel.metrics.exporter=none -Dotel.traces.exporter=jaeger -Dotel.exporter.jaeger.endpoint=http://localhost:14250 -Dotel.resource.attributes=service.name=petclinic -javaagent:"${HOME}/.tennaxia/opentelemetry-javaagent-all.jar" -jar spring-petclinic-2.4.5.jar

**Steps to reproduce**
Add a Owner

**What did you expect to see?**
instrumentation of hibernate Query / Session / Transaction

**What did you see instead?**
instrumentation of jdbc otel.library.name = io.opentelemetry.javaagent.jdbc
![image](https://user-images.githubusercontent.com/23735835/124750663-058fe600-df26-11eb-9a8e-bd56a215b10c.png)

**What version are you using?**
opentelemetry all 1.3.1

**Environment**
Compiler: (e.g., "AdoptOpenJDK 11.0.11+9")

PR #3540, opened 2021-07-09
Hibernate: handle jpa EntityManager

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3517

### DIFF
--- a/instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/EntityManagerTest.groovy
+++ b/instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/EntityManagerTest.groovy
@@ -1,0 +1,214 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import static io.opentelemetry.api.trace.SpanKind.CLIENT
+import static io.opentelemetry.api.trace.SpanKind.INTERNAL
+import static io.opentelemetry.instrumentation.test.utils.TraceUtils.runUnderTrace
+
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes
+import javax.persistence.EntityManager
+import javax.persistence.EntityManagerFactory
+import javax.persistence.EntityTransaction
+import javax.persistence.LockModeType
+import javax.persistence.Persistence
+import javax.persistence.Query
+import spock.lang.Shared
+import spock.lang.Unroll
+
+class EntityManagerTest extends AbstractHibernateTest {
+
+  @Shared
+  EntityManagerFactory entityManagerFactory = Persistence.createEntityManagerFactory("test-pu")
+
+  @Unroll
+  def "test hibernate action #testName"() {
+    setup:
+
+    EntityManager entityManager = entityManagerFactory.createEntityManager()
+    EntityTransaction entityTransaction = entityManager.getTransaction()
+    entityTransaction.begin()
+
+    def entity = prepopulated.get(0)
+    if (attach) {
+      entity = runUnderTrace("setup") {
+        entityManager.merge(prepopulated.get(0))
+      }
+      ignoreTracesAndClear(1)
+    }
+    try {
+      sessionMethodTest.call(entityManager, entity)
+    } catch (Exception e) {
+      // We expected this, we should see the error field set on the span.
+    }
+
+    entityTransaction.commit()
+    entityManager.close()
+
+    expect:
+    boolean isPersistTest = "persist" == testName
+    assertTraces(1) {
+      trace(0, 4 + (isPersistTest ? 1 : 0)) {
+        span(0) {
+          name "Session"
+          kind INTERNAL
+          hasNoParent()
+          attributes {
+          }
+        }
+        span(1) {
+          name ~/Session.$methodName $resource/
+          kind INTERNAL
+          childOf span(0)
+          attributes {
+          }
+        }
+
+        def offset = 0
+        if (isPersistTest) {
+          // persist test has an extra query for getting id of inserted element
+          offset = 1
+          span(2) {
+            childOf span(1)
+            kind CLIENT
+            attributes {
+              "${SemanticAttributes.DB_SYSTEM.key}" "h2"
+              "${SemanticAttributes.DB_NAME.key}" "db1"
+              "${SemanticAttributes.DB_USER.key}" "sa"
+              "${SemanticAttributes.DB_CONNECTION_STRING.key}" "h2:mem:"
+              "${SemanticAttributes.DB_STATEMENT.key}" String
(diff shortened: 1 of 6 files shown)


instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/SpringJpaTest.groovy
```diff
@@ -26,17 +25,33 @@ class SpringJpaTest extends AgentInstrumentationSpecification {
...
-      trace(0, 1) {
+      trace(0, 4) {
         span(0) {
+          name "Session"
+          kind INTERNAL
+          hasNoParent()
+          attributes {
+          }
+        }
+        span(1) {
```
iNikem, line 44: Why do we need this span? Why do we create it at all?

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/EntityManagerTest.groovy
```diff
@@ -0,0 +1,214 @@
...
+    testName                                  | methodName   | resource | attach | flushOnCommit | sessionMethodTest
+    "lock"                                    | "lock"       | "Value"  | true   | false         | { em, val ->
+      em.lock(val, LockModeType.PESSIMISTIC_READ)
+    }
+    "refresh"                                 | "refresh"    | "Value"  | true   | false         | { em, val ->
+      em.refresh(val)
+    }
+    "get"                                     | "(get|find)" | "Value"  | false  | false         | { em, val ->
+      em.find(Value, val.getId())
+    }
```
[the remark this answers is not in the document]
laurit: Changed to `find`. Jpa `EntityManager` interface only has `find` but the implementation will call hibernate `SessionImpl` method `get` or `find`, depending on the version of hibernate, which we trace.

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/EntityManagerTest.groovy
```diff
@@ -0,0 +1,214 @@
...
+        entityManager.merge(prepopulated.get(0))
+      }
+      ignoreTracesAndClear(1)
+    }
+    try {
+      sessionMethodTest.call(entityManager, entity)
+    } catch (Exception e) {
+      // We expected this, we should see the error field set on the span.
+    }
+
```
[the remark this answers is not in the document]
laurit: Change to `when:` `then:`

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/EntityManagerTest.groovy
```diff
@@ -0,0 +1,214 @@
...
+          childOf span(0)
+          attributes {
+          }
+        }
+
+        def offset = 0
+        if (isPersistTest) {
+          // persist test has an extra query for getting id of inserted element
+          offset = 1
+          span(2) {
```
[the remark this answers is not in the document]
laurit: Name of this span is a normalized sql query. I added a name for this one because it is used only by one test and always has the same value.
This test is based on a test from `SessionTest` which also does not assert names for sql queries presumably this is because the sql queries differ between testcases. To me asserting them doesn't really seem that important because this test is about hibernate instrumentation not jdbc. If you want I can add the asserts.

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/SpringJpaTest.groovy
laurit: This span is created for the duration of hibernate session. What is a bit weird is that this span is used as a parent for all other hibernate spans.

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/EntityManagerTest.groovy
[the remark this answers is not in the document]
iNikem: The reason I asked for name is that without name it is hard to understand why all these span exist. This indeed is probably more about span modelling in general than this PR in particular.

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/SpringJpaTest.groovy
```diff
@@ -4,18 +4,17 @@
...
+import org.hibernate.Version
 import org.springframework.context.annotation.AnnotationConfigApplicationContext
 import spock.lang.Shared
 import spring.jpa.Customer
 import spring.jpa.CustomerRepository
 import spring.jpa.PersistenceConfig
 
-/**
- * Unfortunately this test verifies that our hibernate instrumentation doesn't currently work with Spring Data Repositories.
- */
```
trask, line 18: 👍

instrumentation/hibernate/hibernate-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/SessionMethodUtils.java
```diff
@@ -100,5 +100,12 @@ public static void end(
     targetContextStore.putIfAbsent(target, sessionContext);
   }
 
+  public static String getSessionMethodSpanName(String methodName) {
+    if ("fireLock".equals(methodName)) {
+      methodName = "lock";
```
trask, line 105: reducing allocation seems worth small code duplication
```suggestion
      return "Session.lock";
```

trask (approved):
nice

instrumentation/hibernate/hibernate-4.0/javaagent/src/test/groovy/SpringJpaTest.groovy
iNikem: Wait, how long is "the duration of hibernate session" usually? Does it span multiple transactions/user requests or what is its lifecycle?

trask: @iNikem good questions, I opened #3552 to discuss

laurit: See https://docs.jboss.org/hibernate/core/4.1/devguide/en-US/html/ch02.html#d5e626 typically there is one session per request but it is possible to have long sessions that span multiple requests.

mateuszrzeszutek: Some projects manage sessions manually, and sometimes fail miserably at this (been there, done that...)

mateuszrzeszutek (approved):

Issue #1726, opened 2020-11-21
CONTEXT_*_SPAN_KEY need to be moved to an unshaded location?


anuraaga:
If I'm not mistaken, these keys will not function correctly when using both the agent and library instrumentation together since the key will be shaded in one and not compatible. I think these need to be moved to a location in the bootstrap classloader, with methods to get / set these. The opentelemetry-api instrumentation will need to also instrument these methods to perform bridging for these operations since they accept a `Span`.

trask:
> I think these need to be moved to a location in the bootstrap classloader, with methods to get / set these.

these are in `instrumentation-api` which is (shaded) in the bootstrap classloader already

> The opentelemetry-api instrumentation will need to also instrument these methods to perform bridging for these operations in the app classloader since they accept a `Span`.

👍

trask:
I think this is needed for a common use case for end users too: adding attributes to the `SERVER` span, but finding yourself under a spring controller `INTERNAL` span so you can't use `Span.current()`.

Maybe `ServerSpan.current()` or something similarly friendly?

mateuszrzeszutek:
As an additional improvement, we could also set the library `Config` so that it contains exactly the same configuration as the javaagent: config file & SPI included.

PR #3547, opened 2021-07-12
Extract logic for bridging recognized context keys

anuraaga:
For #1726 figured it's good to start with making some common logic. Yet somehow the common logic takes more lines of code... I think it's still winning but not sure


gradle/enforcement/checkstyle.xml
```diff
@@ -222,24 +222,24 @@
     </module>
     <module name="ClassTypeParameterName">
       <!-- modified from default Google Java Style -->
-      <property name="format" value="^[A-Z]+[0-9]?$"/>
+      <property name="format" value="(^[A-Z][0-9]?)$|([A-Z][a-zA-Z0-9]*$)"/>
```
anuraaga, line 225: I've never quite understood the standard styles related to type parameters - a type is a type. Can we loosen to that?

anuraaga:
With https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3547/commits/075c27dec3b0a4cacdbc06f6ad0b0509446db1e4 I abstracted a lot more. A bit more overhead due to array / virtual method call, but guess it shouldn't be significant.

gradle/enforcement/checkstyle.xml
mateuszrzeszutek: I'm fine with both CamelCase and CONSTANT_CASE, as long as you can separate words in the type name - the current MORETHANONEWORDINATYPENAME makes two- or three-word type param names really hard to read.

mateuszrzeszutek (approved):

trask (approved):

instrumentation/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/context/AgentContextStorage.java
```diff
@@ -54,66 +55,19 @@
...
+        new ContextKeyBridge<Span, io.opentelemetry.api.trace.Span>(
+            "application.io.opentelemetry.api.trace.SpanContextKey",
+            "io.opentelemetry.api.trace.SpanContextKey",
+            Bridging::toApplication,
+            Bridging::toAgentOrNull),
+        new ContextKeyBridge<Baggage, io.opentelemetry.api.baggage.Baggage>(
+            "application.io.opentelemetry.api.baggage.BaggageContextKey",
+            "io.opentelemetry.api.baggage.BaggageContextKey",
+            BaggageBridging::toApplication,
+            BaggageBridging::toAgent),
```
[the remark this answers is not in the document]
anuraaga: Hmm - looked at it but not too sure how to change it so leaving it as is for now

PR #3554, opened 2021-07-12
Skeleton for overhead testing subproject.

breedx-splk:
Starting the ball rolling on a series of contributions around overhead testing of the agent.  There will be a series of follow-up PRs that continue adding in functionality.

This first one just sets up the directory and README along with a dockerfile for building a `spring-petclinic-rest` base image that also supports ssh.  Ideally we wouldn't have to build petclinic and we could just layer on top of the existing docker image -- but it seems that it hasn't been updated in > 2 years.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "testing-overhead/spring-petclinic-rest"]
-	path = testing-overhead/spring-petclinic-rest
-	url = https://github.com/spring-petclinic/spring-petclinic-rest.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "testing-overhead/spring-petclinic-rest"]
+	path = testing-overhead/spring-petclinic-rest
+	url = https://github.com/spring-petclinic/spring-petclinic-rest.git
--- a/testing-overhead/Dockerfile-petclinic-base
+++ b/testing-overhead/Dockerfile-petclinic-base
@@ -11,25 +11,9 @@ WORKDIR /app/spring-petclinic-rest
 RUN ./mvnw package -Dmaven.test.skip=true
 RUN cp target/spring-petclinic-rest*.jar /app/spring-petclinic-rest.jar
 
-
 FROM adoptopenjdk:11-jdk
-
-# Allow the test runner to ssh into our container to start and control
-# the application under test.
-
-RUN apt update && \
-    apt install -y openssh-server sudo && \
-    mkdir /run/sshd && \
-    sed -i 's/^#PasswordAuthentication.*/PasswordAuthentication yes/' /etc/ssh/sshd_config && \
-    sed -i 's/^#PermitUserEnvironment.*/PermitUserEnvironment yes/' /etc/ssh/sshd_config && \
-    sed -i 's/^#PermitEmptyPasswords.*/PermitEmptyPasswords yes/' /etc/ssh/sshd_config && \
-    sed -i 's/secure_path="/secure_path="\/opt\/java\/openjdk\/bin:/' /etc/sudoers && \
-    sed -i 's/PATH="/PATH="\/opt\/java\/openjdk\/bin:/' /etc/environment && \
-    useradd -m -d /app -s /bin/bash -u 5001 remote && \
-    passwd -d remote && \
-    mkdir -p /app/.ssh && chmod 700 /app/.ssh && \
-    ln -s /etc/environment /app/.ssh/environment && \
-    echo "%remote ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 COPY --from=app-build /app/spring-petclinic-rest.jar /app/spring-petclinic-rest.jar
 WORKDIR /app
+EXPOSE 9966
+CMD ["java", "-jar", "spring-petclinic-rest.jar"]
--- a/testing-overhead/Dockerfile-petclinic-base
+++ b/testing-overhead/Dockerfile-petclinic-base
@@ -11,7 +11,7 @@ WORKDIR /app/spring-petclinic-rest
 RUN ./mvnw package -Dmaven.test.skip=true
 RUN cp target/spring-petclinic-rest*.jar /app/spring-petclinic-rest.jar
 
-FROM adoptopenjdk:11-jdk
+FROM adoptopenjdk:11-jre
 
 COPY --from=app-build /app/spring-petclinic-rest.jar /app/spring-petclinic-rest.jar
 WORKDIR /app
--- a/testing-overhead/Dockerfile-petclinic-base
+++ b/testing-overhead/Dockerfile-petclinic-base
@@ -1,14 +1,24 @@
-FROM adoptopenjdk:11-jdk
+FROM adoptopenjdk:11-jdk as app-build
 
 # This is the base image that will contain a built version of the spring-petclinic-rest
 # application. Installing the dependencies and maven compiling the application is time
 # consuming, so we do it in a base image to save time nightly.
 
+RUN apt update && apt install -y git
+WORKDIR /app
+RUN git clone http://github.com/spring-petclinic/spring-petclinic-rest.git
+WORKDIR /app/spring-petclinic-rest
+RUN ./mvnw package -Dmaven.test.skip=true
+RUN cp target/spring-petclinic-rest*.jar /app/spring-petclinic-rest.jar
+
+
+FROM adoptopenjdk:11-jdk
+
 # Allow the test runner to ssh into our container to start and control
 # the application under test.
 
 RUN apt update && \
-    apt install -y openssh-server sudo git && \
+    apt install -y openssh-server sudo && \
     mkdir /run/sshd && \
     sed -i 's/^#PasswordAuthentication.*/PasswordAuthentication yes/' /etc/ssh/sshd_config && \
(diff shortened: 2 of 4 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "testing-overhead/spring-petclinic-rest"]
```
jkwatson, line 1: wat

.gitmodules
```diff
@@ -0,0 +1,3 @@
...
+	path = testing-overhead/spring-petclinic-rest
+	url = https://github.com/spring-petclinic/spring-petclinic-rest.git
```
iNikem, line 3: How about using https://github.com/spring-petclinic/spring-petclinic-microservices ? That project already has a jmeter test script that we can reuse. And it seems to be more actively developed.

testing-overhead/Dockerfile-petclinic-base
```diff
@@ -0,0 +1,29 @@
...
+
+# This is the base image that will contain a built version of the spring-petclinic-rest
+# application. Installing the dependencies and maven compiling the application is time
+# consuming, so we do it in a base image to save time nightly.
+
+# Allow the test runner to ssh into our container to start and control
+# the application under test.
+
+RUN apt update && \
+    apt install -y openssh-server sudo && \
```
iNikem, line 11: Why do we need ssh here?

breedx-splk:
Ok let's see if we can do this without a submodule.

jkwatson:
> Ok let's see if we can do this without a submodule.

opentelemetry-java just pulls in the repo as a part of the build, rather than relying on git's crazy submodule system.

.gitmodules
laurit: For start spring-petclinic-rest which starts only once service is probably easier to manage than spring-petclinic-microservices which starts multiple services.

.gitmodules
breedx-splk: Yeah, good call...we don't have to do it this way.  We can just clone it into the base docker image instead.

.gitmodules
breedx-splk: I think it's too many components for our usage, and the service itself is fragmented across multiple services. I like petclinic-rest because you can do all the common workflows (create a pet, schedule an appointment, etc) with one app.

testing-overhead/Dockerfile-petclinic-base
breedx-splk: ssh will be used to start/stop the application under test (petclinic-rest) and also to run `jcmd` on the same container.

iNikem: - Usual "Spring Boot in a Docker" package starts application right away.
- You don't need `jcmd` just to get a jfr recording
- `docker-java` probably allows for command execution inside container in a cleaner way that ssh from java :)

testing-overhead/Dockerfile-petclinic-base
```diff
@@ -0,0 +1,29 @@
...
+    passwd -d remote && \
+    mkdir -p /app/.ssh && chmod 700 /app/.ssh && \
+    ln -s /etc/environment /app/.ssh/environment && \
+    echo "%remote ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
+WORKDIR /app
+RUN git clone http://github.com/spring-petclinic/spring-petclinic-rest.git
+WORKDIR /app/spring-petclinic-rest
+
+RUN ./mvnw package -Dmaven.test.skip=true
```
iNikem, line 28: That is one more argument in favour of microservice version: it already can build itself into docker images

iNikem: This way will leave the full maven local repo inside the container

testing-overhead/Dockerfile-petclinic-base
breedx-splk: * Each test pass wants to run the application twice -- typically with agent and without agent (or possibly with two different agent versions). The test runner needs a mechanism to be able to shut down the running application so that it can restart it in the other mode within the same contianer. SSH seemed like a reasonable way to do this, but I'm open to other ideas. I don't think that "just let docker run the jvm for you" is a good answer here.
* I had also expected to use `jcmd` to take a heap dump of the running application before shutting it down.
* Hmmm...so you think we should create yet another java project that requires compile/build and a runtime JDK just to talk to docker in order to run commands? I don't think we want to be communicating with the docker daemon from CI/docker.

SSH is also a convenient way for pulling output files out of the target app's container...but if we hate ssh now I can look into other options for remote command execution and file transfer. A shared docker-compose volume might work for the latter.

testing-overhead/Dockerfile-petclinic-base
breedx-splk: Does it really matter that much? Would you prefer that I remove the source code and just keep the build output jar?

testing-overhead/Dockerfile-petclinic-base
iNikem: > can restart it in the other mode within the same contianer

Why is it important to have the same container?

testing-overhead/Dockerfile-petclinic-base
iNikem: If we don't plan to share this image outside the job - it does not matter. If we do, it does :)

mateuszrzeszutek: You could also use a multi-stage dockerfile to avoid keeping the trash produced by the build (and build tools themselves) in the image.

testing-overhead/Dockerfile-petclinic-base
breedx-splk: In a cloud-based CI environment we have no guarantees about where a given container will be running. If we don't run the test app in the same container instance, then we run the risk of spinning up on different hardware, on a different network segment, with different computational neighbors. By reusing the same container we reduce the thrash and improve the comparability of the two runs.

iNikem: Wait, how do you plan to run that container? I assumed so far that we run some GHA job, which runs some script or java application which starts containers as much as needed. In this case everything happens on the same VM, does not it?

testing-overhead/Dockerfile-petclinic-base
breedx-splk: Ok, reworked to be a multi-stage dockerfile that no longer leaves the source tree behind. Looks like the image size goes from 839675046 to 636291599, so I guess that's a good thing.

testing-overhead/Dockerfile-petclinic-base
breedx-splk: Your understanding is correct. The end goal is to just run `docker-compose up --exit-code-from testrunner` from a github action. 

There will be 3 containers -- collector, application, and testrunner. The testrunner will start/stop the application runs inside the application container via ssh (or another mechanism if that's rejected). The same application container instance will be (re)used for multiple app runs.

iNikem: But that means we don't need to use one docker container for all runs?

breedx-splk: Why? I thought I explained above why I think it's important/useful to re-use the same container for successive runs of the application.

breedx-splk: Or maybe you're trying to say that a docker-compose environment started by a github action will run all containers on the same VM and therefore we don't need to reuse the container?  If so, how do you know that? Is that documented somewhere?

iNikem: Docker compose does not distribute containers across machines. That is what docker swarm for.

If you assume that containers can spawn on different hosts, then how can you assume that ssh from the coordinator script into a container will work?

anuraaga: IIUC, the brunt of the logic is implemented in shell, like this one?

https://github.com/breedx-splk/iguanodon/blob/main/bin/run-remote-sequence.sh

Can we use testcontainers instead? We use it for a lot of things, and it makes it pretty easy to run commands in container, copy files, etc I think.

breedx-splk: Yeah, the thought was that shell stuff might be more likely to benefit other language agents, but I have been convinced to continue using testcontainers for this. Makes sense.

mateuszrzeszutek: You probably still can test other language agents with testcontainers - only the testrunner has to use JVM

breedx-splk: Yup, this is a great point too. 🤘🏻

iNikem (approved):

testing-overhead/Dockerfile-petclinic-base
```diff
@@ -0,0 +1,19 @@
...
+# consuming, so we do it in a base image to save time nightly.
+
+RUN apt update && apt install -y git
+WORKDIR /app
+RUN git clone http://github.com/spring-petclinic/spring-petclinic-rest.git
+WORKDIR /app/spring-petclinic-rest
+RUN ./mvnw package -Dmaven.test.skip=true
+RUN cp target/spring-petclinic-rest*.jar /app/spring-petclinic-rest.jar
+
+FROM adoptopenjdk:11-jdk
```
anuraaga, line 14: Since it's multistage we can make this JRE so it's quite a bit smaller still :)

breedx-splk: Great idea, thanks!

mateuszrzeszutek (approved):

anuraaga (approved):

Issue #3044, opened 2021-05-19
OpenTelemetry API instrumentation (bridge) tests are only run against latest OTel API


trask:
And no muzzle either.

We want to run muzzle against 1.0.0+, and we want to run tests at least against 1.0.0 and latest.

mateuszrzeszutek:
For muzzle to work we'd have to have multiple versions of `application.io...` shaded libs - the easiest way would probably be to publish `opentelemetry-api-shaded-for-instrumenting`

trask:
Let's keep this open to track publishing of the old versions and running the tests against 1.0.0 + latestDep

anuraaga:
Is everyone comfortable with just changing the dependency version in the `build.gradle.kts` for the opentelemetry-api-shaded-for-instrumenting project instead of checking out old tags?

This actually made my brain realize - should we move this project to the `opentelemetry-java` repo for internal consumption? This repo's versions don't necessarily sync with that one at the patch level

trask:
> Is everyone comfortable with just changing the dependency version in the `build.gradle.kts` for the opentelemetry-api-shaded-for-instrumenting project instead of checking out old tags?

👍

> This actually made my brain realize - should we move this project to the `opentelemetry-java` repo for internal consumption? This repo's versions don't necessarily sync with that one at the patch level

oh, great thought, this totally makes sense

trask:
Closing, the situation is much better now, can revisit later if anyone has renewed interest.

PR #3555, opened 2021-07-13
Publish shaded OpenTelemetry API and annotations for muzzle

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3044 (well, a part of it - we still need to publish old versions)

### DIFF
--- a/instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
+++ b/instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
@@ -2,13 +2,7 @@ plugins {
   id("otel.javaagent-instrumentation")
 }
 
-muzzle {
-  pass {
-    group.set("io.opentelemetry.javaagent")
-    module.set("opentelemetry-ext-annotations-shaded-for-instrumenting")
-    versions.set("[1,)")
-  }
-}
+// TODO: add muzzle once 1.4.0 is released
 
 val versions: Map<String, String> by project
 
--- a/instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
+++ b/instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
@@ -2,6 +2,14 @@ plugins {
   id("otel.javaagent-instrumentation")
 }
 
+muzzle {
+  pass {
+    group.set("io.opentelemetry.javaagent")
+    module.set("opentelemetry-ext-annotations-shaded-for-instrumenting")
+    versions.set("[1,)")
+  }
+}
+
 val versions: Map<String, String> by project
 
 dependencies {
--- a/opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
+++ b/opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
@@ -8,40 +8,16 @@ plugins {
 description = "opentelemetry-api shaded for internal javaagent usage"
 group = "io.opentelemetry.javaagent"
 
-val shadowInclude by configurations.creating {
-  isCanBeResolved = true
-  isCanBeConsumed = false
-}
-
 dependencies {
-  shadowInclude("io.opentelemetry:opentelemetry-api")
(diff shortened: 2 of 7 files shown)


opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
```diff
@@ -2,18 +2,46 @@ plugins {
...
+
+  publishing {
+    publications {
+      named<MavenPublication>("maven") {
+        // shadow does not use default configurations
+        project.shadow.component(this)
+        // force jar packaging
+        pom {
+          packaging = "jar"
+        }
```
mateuszrzeszutek, line 43: I found a hack that fixes pom packaging issues 🙈  😂

instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
```diff
@@ -2,6 +2,14 @@ plugins {
   id("otel.javaagent-instrumentation")
 }
 
+muzzle {
```
anuraaga, line 5: Should we add muzzle after publishing? Just in case it fails, it doesn't happen because of a release but because of a PR

opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
anuraaga: Would it be worth letting muzzle configuration specify a classifier instead of having to do this?

trask: > I found a hack that fixes pom packaging issues 🙈 😂

wow. just wow. 😂

> Would it be worth letting muzzle configuration specify a classifier instead of having to do this?

I do prefer not having an unnecessary classifier on shaded artifacts, though very flexible for "anything that works" when it comes to gradle shadow plugin. How I miss maven-shade-plugin...

anuraaga: Yeah for these that aren't meant for end users to consume, I would be ok on going with simpler build scripts if it makes sense. For end user artifacts, no classifier is very worth it though

opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
```diff
@@ -2,18 +2,46 @@ plugins {
   id("com.github.johnrengelman.shadow")
 
   id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+description = "opentelemetry-api shaded for internal javaagent usage"
+group = "io.opentelemetry.javaagent"
```
trask, line 9: any interest in `io.opentelemetry.javaagent.build` or something similar? `agent-for-testing` could go there too. `testing-common` could go in `io.opentelemetry.instrumentation.build` or similar. do we have anything else to justify new sub-groupIds?

trask (approved):

opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
anuraaga: I don't think we need separate groups just for these artifacts that end users consume. One group for all possible artifacts library ones will (testing is part of the deal) and one for agent.

A separate one for this sort of implementation detail could make sense but is probably overkill.

instrumentation/opentelemetry-annotations-1.0/javaagent/build.gradle.kts
mateuszrzeszutek: Hmm, good point. I'll remove that for now and add a TODO to add it back after the release.

opentelemetry-api-shaded-for-instrumenting/build.gradle.kts
mateuszrzeszutek: Okay, I simplified the shadow build and added classifier property to muzzle check plugin. That should work for these internal artifacts.

anuraaga (approved):

PR #3557, opened 2021-07-13
Clean up javaagent-instrumentation-api and javaagent-extension-api de…

mateuszrzeszutek:
…ps a bit

### DIFF
--- a/javaagent-instrumentation-api/build.gradle.kts
+++ b/javaagent-instrumentation-api/build.gradle.kts
@@ -7,10 +7,8 @@ plugins {
 group = "io.opentelemetry.javaagent"
 
 dependencies {
-  api("io.opentelemetry:opentelemetry-api")
-  compileOnly("io.opentelemetry:opentelemetry-sdk")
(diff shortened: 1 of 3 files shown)


javaagent-instrumentation-api/build.gradle.kts
```diff
@@ -7,10 +7,8 @@ plugins {
 group = "io.opentelemetry.javaagent"
 
 dependencies {
-  api("io.opentelemetry:opentelemetry-api")
-  compileOnly("io.opentelemetry:opentelemetry-sdk")
+  api(project(":instrumentation-api"))
```
mateuszrzeszutek, line 10: `javaagent-instrumentation-api` directly extends `instrumentation-api` with a few javaagent-specific usecases, so it makes sense to expose `instrumentation-api` as API dependency here.

trask (approved):

anuraaga (approved):

iNikem (approved):

PR #3559, opened 2021-07-13
Move caffeine patching to instrumentation-api-caching

mateuszrzeszutek:
We never use a `ForkJoinPool` anyway (we always use `Runnable::run` as the executor), so we might as well patch the library as early as possible, just before it's integrated into `instrumentation-api`.


trask (approved):
👍

anuraaga (approved):

PR #3569, opened 2021-07-14
Netty connection failure handling

laurit:
This pr removes netty 4.0 & 4.1 `ChannelFutureListenerInstrumentation` which was responsible for creating connection failure span and instead uses `FutureListenerWrappers` for the same purpose. The reason behind this is that instrumentations using a matcher like `implementsInterface(named("io.netty.channel.ChannelFutureListener"))` don't apply to lambdas. For vert.x 4 the listener we want to intercept happens to be a lambda.


instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
```diff
@@ -0,0 +1,72 @@
...
+      return;
+    }
+
+    if (!(future instanceof ChannelFuture)) {
+      return;
+    }
+    // If first call to GenericFutureListener#operationComplete has an exception then we
+    // treat it as the cause of connection failure and create a special span for it
+    ChannelFuture channelFuture = (ChannelFuture) future;
+    AttributeKey<Context> ak = null;
```
mateuszrzeszutek, line 46:
```suggestion
```

mateuszrzeszutek (approved):

PR #3570, opened 2021-07-14
Clean up bootstrap dependencies and build scripts

mateuszrzeszutek:
This PR moves the bootstrap projects aggregation to the `javaagent` module (and `agent-for-testing` too); `javaagent-bootstrap` is now a plain, standard library. I also reorganized both javaagent modules a bit to keep configurations/dependencies and tasks together.


iNikem (approved):

trask (approved):
nice

PR #3571, opened 2021-07-14
Publish gradle plugins to Sonatype

iNikem:


mateuszrzeszutek (approved):

.github/workflows/gradle-plugins.yml
```diff
@@ -0,0 +1,39 @@
+name: CI build
```
trask, line 1:
```suggestion
name: Publish Gradle plugin snapshots
```

trask (commented):
It looks like groupId is `io.opentelemetry.instrumentation.gradle`, should we change that to `io.opentelemetry.instrumentation` based on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3555#discussion_r669317787

trask:
or `io.opentelemetry.javaagent`, since it's javaagent-specific..

iNikem:
> or `io.opentelemetry.javaagent`, since it's javaagent-specific..

Done

mateuszrzeszutek (approved):

PR #3579, opened 2021-07-15
Disable SystemOut errorprone check when running locally

trask:
this solves my primary errorprone complaint

### DIFF
--- a/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
@@ -8,7 +8,8 @@ dependencies {
   errorprone("com.google.errorprone:error_prone_core")
 }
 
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

buildSrc/src/main/kotlin/otel.errorprone-conventions.gradle.kts
```diff
@@ -20,6 +21,10 @@ tasks {
 
       excludedPaths.set(".*/build/generated/.*")
 
+      if (System.getenv("CI") == null) {
```
anuraaga, line 24: I had thought it a necessary evil since the prevelance of printf debugging for us actually increases the chance of us accidentally merging it. But this is a great solution.

iNikem (approved):

mateuszrzeszutek (approved):

PR #3580, opened 2021-07-15
Fix test metrics conversion from OTLP

anuraaga:
Also migrated MeterTest to Java.

- (Main reason) Allows easy copy paste of assertions from SDK repo like https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/metrics/src/test/java/io/opentelemetry/sdk/metrics/LongCounterSdkTest.java#L74
- Metrics will continue to break and fixing breaks in Java is much easier
- The heavy branching / priority on reduction of lines of code in spock tests made them very hard to reason about and maintain. Now the tests are independent and easy to read, at the expense of some duplication of structure which is OK IMO


instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/test/groovy/MeterTest.groovy
```diff
@@ -1,271 +0,0 @@
...
-    metricData != null
-    metricData.description == "d"
-    metricData.unit == "u"
-    metricData.type == expectedType
-    metricData.instrumentationLibraryInfo.name == instrumentationName
-    metricData.instrumentationLibraryInfo.version == "1.2.3"
-    points(metricData).size() == 1
-    def point = points(metricData).iterator().next()
-    if (bind) {
-      point.attributes == Attributes.of(AttributeKey.stringKey("w"), "x", AttributeKey.stringKey("y"), "z")
```
anuraaga, line 57: Bind binds with empty labels above. I have no idea what's going on with this line and am glad the new tests are easier to understand

trask (approved):

PR #3581, opened 2021-07-15
Don't use cache for distro build

iNikem:
One of the reasons our builds are so slow recently is because gradle dependencies cache is being polluted by `example-distro` job. If finishes fast and publishes _its_ gradle dependencies to the cache, which is very minimal. All subsequent jobs try to use that, and thus have to actually download a lot of dependencies.

See https://github.com/burrunan/gradle-cache-action/issues/45

This PR removes gradle cache from `example-distro` job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,11 +171,6 @@ jobs:
           distribution: adopt
           java-version: 11
 
-      - name: Restore cache
-        uses: burrunan/gradle-cache-action@v1.10
-        with:
-          job-id: jdk11
-
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
         with:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -171,11 +171,6 @@ jobs:
           distribution: adopt
           java-version: 11
 
-      - name: Restore cache
-        uses: burrunan/gradle-cache-action@v1.10
-        with:
-          job-id: jdk11
-
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
         with:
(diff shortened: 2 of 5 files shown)


anuraaga:
Any thoughts on `save-gradle-dependencies-cache: false`?

anuraaga:
Actually there's a ` gradle-dependencies-cache-key:`!

anuraaga:
If `gradle-dependencies-cache-key:` is usable, then looks like we can try 

- primary
- testLatestDeps
- examples
- smoke-tests (not sure on this one, guess it might overlap well with primary)

iNikem:
`gradle-dependencies-cache-key` takes a list of files, not an arbitrary string.

If `save-gradle-dependencies-cache: false`, then what benefit we have by using this action anyway?

anuraaga:
@iNikem Yeah looks like we'd need to `echo` into a file for now, I filed an issue for that. I suspect there is thrash between normal and testLatestDeps builds though so it would be nice to try it. 

> If save-gradle-dependencies-cache: false, then what benefit we have by using this action anyway?

Figured for examples builds that don't change much, the build cache will still work well to save GitHub some resources, which we use so much of. I know the examples build is small, but adding that flag seems as simple as removing the cache. But `echo` into a file for the dependencies cache key seems like a nice approach to try if you want

iNikem:
"echo into a file for the dependencies cache key" seems like a hack to me :D Anyway, I suggest merging this PR and then monitoring `example-distro` job. If it takes more than a couple of minutes, we can take a look at how to optimise it.

anuraaga (approved):

anuraaga:
Yeah let's also look at testLatestDeps - a while ago I was looking at the logs and confused why there would be so much time on `post restore cache` spent some times. But thanks to this PR realize it's because of the thrashing - I believe we used to only run testLatestDeps on nightly build and had much less thrashing, I think I've seen some slowness ever since we added it to main build.

mateuszrzeszutek (approved):

PR #3584, opened 2021-07-15
Specialize dependencies cache key for testLatestDeps

anuraaga:
Smoke tests only assemble the agent and don't need any of the testing dependencies, some massive like glassfish app server, gwt. If a smoke test happens to upload cache after a build job, it could remove a lot of dependencies from the cache.

First just running to see the output of post restore cache


iNikem (approved):

iNikem:
@anuraaga Let's merge it and see how it will play out :)

iNikem:
You removed `smoke-test` key? :) Do you want to rename the PR as well then?

anuraaga:
@iNikem Thanks - after #3593 the only build I see with any chance of benefit is testLatestDeps so switched to it for this PR

anuraaga:
FYI just for reference, I considered putting my hack file into `.gradle` but found it's always excluded

https://github.com/burrunan/gradle-cache-action/blob/main/layered-cache/src/main/kotlin/com/github/burrunan/gradle/cache/dependenciesCache.kt#L78

mateuszrzeszutek (approved):

PR #3586, opened 2021-07-15
Call correct tasks for publishing

iNikem:
Using SNAPSHOT version to not expose this yet too much on Maven Central

### DIFF
--- a/.github/workflows/gradle-plugins.yml
+++ b/.github/workflows/gradle-plugins.yml
@@ -35,5 +35,5 @@ jobs:
           SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
-        run: ../gradlew build snapshot
+        run: ../gradlew build publishToSonatype closeAndReleaseStagingRepository
         working-directory: gradle-plugins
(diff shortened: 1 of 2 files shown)


.github/workflows/gradle-plugins.yml
```diff
@@ -35,5 +35,5 @@ jobs:
           SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
           GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
-        run: ../gradlew build snapshot
+        run: ../gradlew build publishToSonatype closeAndReleaseStagingRepository
```
mateuszrzeszutek, line 38: What's wrong with `snapshot`?

mateuszrzeszutek (approved):

.github/workflows/gradle-plugins.yml
anuraaga: This will publish to oss.sonatype right? No need to close repo for it (don't know if the API works).

iNikem: It is task provided by `nebula` plugin, which is not used in this module

iNikem: > This will publish to oss.sonatype right? No need to close repo for it (don't know if the API works).

I have no idea. I was unable to find any explanation what does which task do and how to just publish public artifact to `https://oss.sonatype.org/content/repositories/snapshots/`.

https://central.sonatype.org/publish/publish-guide/#releasing-to-central seems to say that after deployment the repo is private until it is closed.

anuraaga: It's the nexus publish plugin that controls where things go. SNAPSHOT will go to oss.sonatype snapshots by default. No need to close it.

Otherwise Maven Central would reject them if they went there.

I'm not a huge fan of the separate settings.gradle and having to reproduce all our publishing logic, so couldn't approve the other PR. Perhaps this is motivation to avoid it? ;)

iNikem: Sure, as soon as we understand how to properly re-use all the existing code. I was unable to do so :)

anuraaga (approved):

PR #3587, opened 2021-07-15
 Fixed potential NullPointerException if AMQP.Properties::getHeaders returns null

bnikster:
In some applications we are not providing headers at the AMQP.Properties, which results in the following exception:
java.lang.NullPointerException\n at io.opentelemetry.javaagent.instrumentation.rabbitmq.TextMapExtractAdapter.get(TextMapExtractAdapter.java:22)\n at io.opentelemetry.javaagent.instrumentation.rabbitmq.TextMapExtractAdapter.get(TextMapExtractAdapter.java:11)\n 
.... 

After the error RabbitMQ stopped working, these fix prevents the NullPointerException

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -191,7 +191,7 @@ jobs:
         with:
           job-id: jdk11
           remote-build-cache-proxy-enabled: false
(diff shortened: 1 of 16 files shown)


iNikem:
@bnikster looks like you have unrelated changes in this PR. Can you clean that up, please?

bnikster:
will do the fix in main again, and start a new pull request

PR #3593, opened 2021-07-15
Only one main build should persist gradle dependencies to GHA cache

iNikem:


.github/workflows/ci.yml
```diff
@@ -97,6 +98,7 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: latestDepTest
```
anuraaga, line 100: It would be nice to try to save just this one to another key but can try it later.

anuraaga (approved):

PR #3594, opened 2021-07-16
Update gradle plugin artifact groupId

trask:
An alternative to this PR would be to publish two separate gradle plugin artifacts:
* `io.opentelemetry.instrumentation:gradle-plugins` for library instrumentation gradle plugins
* `io.opentelemetry.javaagent:gradle-plugins` for javaagent instrumentation gradle plugins

### DIFF
--- a/gradle-plugins/build.gradle.kts
+++ b/gradle-plugins/build.gradle.kts
@@ -8,7 +8,9 @@ plugins {
   id("io.github.gradle-nexus.publish-plugin")
 }
 
-group = "io.opentelemetry.javaagent"
+// this is the groupId for the gradle plugin artifact (the groupId for each gradle plugin is
+// determined by the gradle plugin file name)
+group = "io.opentelemetry.instrumentation"
 version = "0.1.0-SNAPSHOT"
 
 repositories {


anuraaga (approved):

mateuszrzeszutek (approved):

iNikem:
Wait, why `instrumentation`? You said before that `javaagent` is more suitable because it generates methods for javaagent to use.

trask:
> Wait, why `instrumentation`? You said before that `javaagent` is more suitable because it generates methods for javaagent to use.

Ya, so the question is whether we want to publish one gradle plugin artifact that has both `instrumentation` and `javaagent` gradle plugins, or if we want to publish two gradle plugin artifacts, one for `instrumentation` and one for `javaagent`

If we only publish one gradle plugin artifact, I think it's better for it to be under `instrumentation`, because in general writing `javaagent` stuff depends on `instrumentation` artifacts, but not the other way around.

The alternative to this PR is to publish two separate gradle plugin artifacts.

I support either alternative, though I probably lean towards separate gradle plugin artifacts, since we generally try to keep javaagent-specific stuff out of instrumentation artifacts.

iNikem:
In any way current plugin that we publish, and several subsequent (`javaagent-instrumentation`, `javaagent-testing` and muzzle check) are certainly `javaagent` plugins. If at some point in time we are ready to publish "library" plugins, we can do that under separate group.

Also please not forget that the only thing that our end-users have to care about is plugin id, which is completely separate from the groupId of the jar file that implements that plugin.

anuraaga:
> Also please not forget that the only thing that our end-users have to care about is plugin id, which is completely separate from the groupId of the jar file that implements that plugin.

This is precisely the reason why I think we shouldn't bother with having two artifacts we publish of gradle plugins. End users don't care that much, so having an artifact for library instrumentation gradle plugin (just a thin wrapper around `.base` which introduces `testLatestDeps`, yes this is why we do need to publish the library instrumentation plugin too) is too much. All the gradle plugins for the java instrumentation project could just go in the `instrumentation` group ID.

iNikem:
@anuraaga Do I understand you correctly, that you propose plugin id to be like `io.opentelemetry.javaagent.muzzle.generation` (in `javaagent` namespace), but implementation jars to be in `io.opentelemetry.instrumentation` group?

anuraaga:
@iNikem I guess I still sort of like what we currently have in `buildSrc`

`io.opentelemetry.instrumentation.javaagent-instrumentation`
`io.opentelemetry.instrumentation.library-instrumentation`

Other plugins could have any ID, even `io.opentelemetry.instrumentation.internal.muzzle-generation` potentially.

While I prefer the single artifact, if everyone else wants to split though, going with `io.opentelemetry.javaagent.instrumentation` and `io.opentelemetry.instrumentation` plugins I guess, I'm ok with it, it just doesn't seem like great bang-for-buck vs the additional build scripts.

iNikem:
`muzzle-generation` is certainly not internal, we do want to publish it and encourage custom distros/extensions to use it. :)

iNikem:
So, for `pluginId` you don't want different namespaces/groups for agent vs non-agent use-cases? Just `io.opentelemetry.instrumentation` for everything?

anuraaga:
`muzzle-generation` - when would we expect this to be used instead of a top level `javaagent-instrumentation`? Can't think of a use case but could be I'm missing it.

The features that are needed by any javaagent instrumentation, external or internal, are

- muzzle codegen
- muzzle check (we should provide the feature, it's up to external users whether they actually call the task in their build)
- run tests with javaagent
- testLatestDeps (we should provide the feature, it's up to external users whether they actually call the task in their build)

I don't believe any of these are individually needed but not sure.

> Just io.opentelemetry.instrumentation for everything?

I think so, mainly since I suspect there should only be 2 or 3 practical entry point plugins, as opposed to hundreds like our other artifacts. Separating doesn't seem to be that important.

iNikem:
> `muzzle-generation` - when would we expect this to be used instead of a top level `javaagent-instrumentation`? Can't think of a use case but could be I'm missing it.

You are right. I always forget about this extra level :)

> 
> The features that are needed by any javaagent instrumentation, external or internal, are
> 
> * muzzle codegen
> * muzzle check (we should provide the feature, it's up to external users whether they actually call the task in their build)
> * run tests with javaagent
> * testLatestDeps (we should provide the feature, it's up to external users whether they actually call the task in their build)
> 

These are certainly needed, agreed.

> I don't believe any of these are individually needed but not sure.

If there is little overhead, we can indeed apply them all at once. Then pluginId `io.opentelemetry.instrumentation.javaagent-instrumentation` probably makes sense as "all-included".

> 
> > Just io.opentelemetry.instrumentation for everything?
> 
> I think so, mainly since I suspect there should only be 2 or 3 practical entry point plugins, as opposed to hundreds like our other artifacts. Separating doesn't seem to be that important.

Ok, works for me :)

trask:
Closing in favor of #3620

PR #3595, opened 2021-07-16
Help out Intellij

### LINKED ISSUE #3551: Intellij not finding io.opentelemetry.instrumentation.api.caching.Cache
This started from #3471

Any ideas how to resolve? Should I make `opentelemetry-api-caching` a `compileOnly` dependency of modules that use it, just to avoid this issue in Intellij workspace?

image

trask:
Resolves #3551

@anuraaga thought that @mateuszrzeszutek's pom->jar magic might be an alternative if you can't stand this 😭


anuraaga (approved):
The main point being anything Gradle can build but IntelliJ can't resolve should be a bug in IntelliJ so we'd hope for that to be fixed eventually

iNikem (approved):

Issue #3566, opened 2021-07-14
`NoopContextStorageProvider` side-effect


Enkelian:
Adding gradle dependency on `io.opentelemetry:opentelemetry-extension-noop-api` in `javaagent-tooling` module (refer to #3539)  results in an unwanted side-effect - `NoopContextStorageProvider` gets picked up by SPI and is used instead of the default provider in every agent configuration.

Currently, this problem is solved by using `System.setProperty("io.opentelemetry.context.contextStorageProvider", "default")` in `OpentelemetryInstaller`. This line should be reconsidered once noop API provides sufficient configuration.

mateuszrzeszutek:
@anuraaga or @jkwatson can you move this issue to opentelemetry-java?

Also, quoting my idea:

> `NoopContextStorageProvider` is an SPI implementation that automatically gets picked up whenever it's on classpath - that makes perfect sense when you use the library in your app, but for the agent it looks like an unintented effect.
> Just thinking out loud: maybe something like `otel.java.disabled.context.storage.providers`?

jkwatson:
> @anuraaga or @jkwatson can you move this issue to opentelemetry-java?
> 
> Also, quoting my idea:
> 
> > `NoopContextStorageProvider` is an SPI implementation that automatically gets picked up whenever it's on classpath - that makes perfect sense when you use the library in your app, but for the agent it looks like an unintented effect.
> > Just thinking out loud: maybe something like `otel.java.disabled.context.storage.providers`?

I'm not sure why the current system property isn't sufficient for this? I don't think you'd ever want to specify more than one storage provider, or include more than one in your SPI classpath, so it seems strange to not just have the existing system property to specifically pick out the one you want.

trask:
@laurit @Enkelian thanks for delving into this! we discussed (at length 😭) in SIG yesterday and agreed that current system property approach is a reasonable compromise

PR #3599, opened 2021-07-16
Exclude service provider configuration for NoopContextStorageProvider

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3566

### DIFF
--- a/javaagent-tooling/opentelemetry-ext-noop-api-for-instrumenting/build.gradle.kts
+++ b/javaagent-tooling/opentelemetry-ext-noop-api-for-instrumenting/build.gradle.kts
@@ -1,0 +1,25 @@
+plugins {
+  id("com.github.johnrengelman.shadow")
+
+  id("otel.java-conventions")
+  id("otel.publish-conventions")
+}
+
+group = "io.opentelemetry.javaagent"
+
+dependencies {
+  implementation("io.opentelemetry:opentelemetry-extension-noop-api")
+}
+
+tasks {
+  shadowJar {
+    dependencies {
+      include(dependency("io.opentelemetry:opentelemetry-extension-noop-api"))
+    }
+    exclude("META-INF/services/io.opentelemetry.context.ContextStorageProvider")
+  }
+
+  assemble {
+    dependsOn(shadowJar)
+  }
+}
--- a/javaagent-tooling/opentelemetry-ext-noop-api-for-instrumenting/build.gradle.kts
+++ b/javaagent-tooling/opentelemetry-ext-noop-api-for-instrumenting/build.gradle.kts
@@ -14,6 +14,7 @@ dependencies {
 tasks {
   shadowJar {
     dependencies {
+      // including only opentelemetry-extension-noop-api excludes its transitive dependencies
       include(dependency("io.opentelemetry:opentelemetry-extension-noop-api"))
     }
     exclude("META-INF/services/io.opentelemetry.context.ContextStorageProvider")
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
@@ -46,8 +46,6 @@ public class OpenTelemetryInstaller implements AgentListener {
       if (config.getBooleanProperty(JAVAAGENT_NOOP_CONFIG, false)) {
         GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
       } else {
-        System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
-
         OpenTelemetrySdk sdk = OpenTelemetrySdkAutoConfiguration.initialize();
         OpenTelemetrySdkAccess.internalSetForceFlush(
             (timeout, unit) -> {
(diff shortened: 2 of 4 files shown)


javaagent-tooling/opentelemetry-ext-noop-api-for-instrumenting/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+group = "io.opentelemetry.javaagent"
+
+dependencies {
+  implementation("io.opentelemetry:opentelemetry-extension-noop-api")
+}
+
+tasks {
+  shadowJar {
+    dependencies {
+      include(dependency("io.opentelemetry:opentelemetry-extension-noop-api"))
```
iNikem, line 17: ShadowJar automatically uses runtimeClasspath, including `implementation` dependencies. Why do we need to list it here again?

laurit: To exclude transitive dependencies of `opentelemetry-extension-noop-api`.

iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -46,8 +46,6 @@ public static synchronized void installAgentTracer(Config config) {
       if (config.getBooleanProperty(JAVAAGENT_NOOP_CONFIG, false)) {
         GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
       } else {
-        System.setProperty("io.opentelemetry.context.contextStorageProvider", "default");
```
anuraaga, line 49: After taking a step back, I'm sort of wondering how bad this approach was actually.

- We force the context storage in the app classloader to be the bridge into the agent's context storage
- Any attempt to set this by the user will not work as expected since it's only used in the agent classloader

So just hardcoding this property doesn't seem that bad anymore as regardless of whether we have the noop config or not, it seems like it could have been a good idea. Any thoughts?

laurit: I agree that since even the name of system property is renamed by shading it is unlikely to clash with whatever user is doing.
I don't like `System.setProperty` approach much because seeing it immediately makes me wonder why would you do something like that. In that sense making `NoopContextStorageProvider` non discoverable if we don't want it to be found feels like a more natural solution, but considering that `System.setProperty` only needs 1 line it doesn't feel that bad any more.

anuraaga: I agree with the hesitance around setProperty for control flow. But indeed lacking any one other identifiable problem, the 1 line aspect is very appealing here.

trask: 👍

iNikem: @anuraaga @trask @laurit I am confused about what is our final decision :)

trask: can excluding the META-INF/services file be one line also, by adding the exclusion directly when building the javaagent, e.g. the `==>` line below

```
fun isolateSpec(projectsWithShadowJar: Collection<Project>): CopySpec = copySpec {
  from(projectsWithShadowJar.map { zipTree(it.tasks.getByName<ShadowJar>("shadowJar").archiveFile) }) {
    // important to keep prefix "inst" short, as it is prefixed to lots of strings in runtime mem
    into("inst")
    rename("""(^.*)\.class$""", "$1.classdata")
    // Rename LICENSE file since it clashes with license dir on non-case sensitive FSs (i.e. Mac)
    rename("""^LICENSE$""", "LICENSE.renamed")
==> exclude("META-INF/services/io.opentelemetry.javaagent.shaded.io.opentelemetry.context.ContextStorageProvider")
  }
}
```

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -46,8 +46,6 @@ public static synchronized void installAgentTracer(Config config) {
       if (config.getBooleanProperty(JAVAAGENT_NOOP_CONFIG, false)) {
         GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
```
mateuszrzeszutek, line 47: If the noop context implementation is no longer auto-discoverable, should we set it explicitly here?

I wonder if we're just trading one `System.setProperty()` for another, in other words...

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
laurit: @trask this is what I did initially. You actually need to do this for both testing agent and real agent. As this file also gets shaded into intermediate dependencies like `opentelemetry-instrumentation-1.5.0-SNAPSHOT-all.jar` it gets hard to track where it comes from. This is why I chose to eliminate it as close to the source as possible.

trask:
@laurit @Enkelian thanks for delving into this! we discussed (at length 😭) in SIG yesterday and agreed that current system property approach is a reasonable compromise

PR #3600, opened 2021-07-16
Refactor agent-for-testing build

mateuszrzeszutek:
This PR removes  the shadowing logic from the `agent-exporter` project; it's now a "normal" library. All shadowing/relocation code is moved to `agent-for-testing`.
The test agent build script has two configurations that are used to assemble the javaagent:
* `bootstrapLibs` for dependencies/modules that are to be placed in the final jar directly (i.e. loadable by the bootstrap CL)
* `javaagentLibs` for dependencies/modules that are to be placed in the `inst/` dir (loadable by the agent CL)

The testing agent had some duplications too, I've removed them all by excluding bootstrap libs from the `javaagentLibs` configuration.

### DIFF
--- a/testing/agent-exporter/build.gradle.kts
+++ b/testing/agent-exporter/build.gradle.kts
@@ -1,47 +1,19 @@
 plugins {
-  id("io.opentelemetry.instrumentation.javaagent-shadowing")
-
   id("otel.java-conventions")
 }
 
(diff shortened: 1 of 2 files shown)


iNikem (approved):

trask (approved):
nice 👍

PR #3603, opened 2021-07-16
Hibernate: set span name only on method entry

laurit:
This pr removes updating span name on method exit, we don't need to inspect the return value of method to know what entity is being operated on. Also instead of trying to figure out whether the first argument of a method is String representing requested entity name or an Object representing entity make things a bit more explicit by using method descriptor. If we know that the first argument of method is a String use it as entity name, if it is an Object then it must be an entity and ask hibernate what its name is.


instrumentation/hibernate/hibernate-3.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v3_3/CriteriaInstrumentation.java
```diff
@@ -72,7 +78,6 @@ public static void startMethod(
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void endMethod(
         @Advice.Thrown Throwable throwable,
-        @Advice.Return(typing = Assigner.Typing.DYNAMIC) Object entity,
         @Advice.Origin("#m") String name,
```
trask, line 81:
```suggestion
```

trask (approved):
👍

iNikem (approved):

PR #3610, opened 2021-07-17
Revert "Publish shaded OpenTelemetry API and annotations for muzzle"

trask:
Reverts open-telemetry/opentelemetry-java-instrumentation#3555

Based on discussions in last Thu and Fri SIG meetings, it sounds like we probably want a different solution (location) for publishing these artifacts, so let's hold off on publishing these to maven central.


jkwatson (approved):

iNikem:
Are we sure it is not enough to remove `id("otel.publish-conventions")`? Other changes in muzzle plugin seems useful anyway

PR #3612, opened 2021-07-18
Fix bounds check

trask:
Looking at the code, I don't believe it was even possible to end up out-of-bounds here, but a static analysis tool (https://lgtm.com/) we run against our fork found this, so may as well fix.


instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/aspects/WithSpanAspectAttributeBinder.java
```diff
@@ -51,7 +51,7 @@ private static String attributeName(Parameter parameter, String[] parameterNames
     if (!value.isEmpty()) {
       return value;
     }
-    if (parameterNames != null && parameterNames.length >= index) {
+    if (parameterNames != null && parameterNames.length > index) {
```
iNikem, line 54:
```suggestion
    if (parameterNames != null && index  < parameterNames.length) {
```

iNikem (approved):

anuraaga (approved):

PR #3616, opened 2021-07-19
Remove remaining usages of basic span asserts

anuraaga:


instrumentation/gwt-2.0/javaagent/src/testapp/webapp/WEB-INF/web.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <web-app xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xmlns="http://java.sun.com/xml/ns/javaee" xmlns:web="http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
```
anuraaga, line 3: I think running optimize imports on the project caused this. Test still passed but let me know if I should go ahead and revert

laurit: I think it is better this way, `xmlns:web` isn't needed.

iNikem (approved):

trask (approved):

PR #3617, opened 2021-07-19
Revert save-gradle-dependencies-cache and try setting cache key for smoke tests

anuraaga:
Not sure if a bug or confusingly named property, but it seems like we now don't even download gradle dependencies cache at all for these jobs

image

### DIFF
--- a/.github/workflows/build-grpc-smoke-dist.yml
+++ b/.github/workflows/build-grpc-smoke-dist.yml
@@ -27,7 +27,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: grpc-smoke
-          save-gradle-dependencies-cache: false
 
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          save-gradle-dependencies-cache: false
 
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
@@ -153,7 +152,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: smokeTests
-          save-gradle-dependencies-cache: false
 
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
@@ -208,7 +206,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk11
-          save-gradle-dependencies-cache: false
 
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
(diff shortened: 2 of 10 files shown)


.github/workflows/build-grpc-smoke-dist.yml
```diff
@@ -27,7 +27,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: grpc-smoke
-          save-gradle-dependencies-cache: false
```
iNikem, line 30: It is confusingly named property, I have confirmed it with Vladimir. In this case of `build-smoke` I think we should use `read-only: true` instead

.github/workflows/ci.yml
```diff
@@ -67,7 +67,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          save-gradle-dependencies-cache: false
```
iNikem, line 70: `read-only: true`

anuraaga:
Given Gradle automatically clears out old cache entries periodically, I wonder if we might consider just sticking to normal cache action to make things more easy to reason about. That being said, unfortunately there doesn't seem to still be a way to configure the cleanup time and it's a pretty long 30 days

https://docs.gradle.org/current/userguide/directory_layout.html#dir:gradle_user_home:cache_cleanup

.github/workflows/ci.yml
```diff
@@ -208,7 +213,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk11
-          save-gradle-dependencies-cache: false
```
iNikem, line 211: `read-only: true`

.github/workflows/nightly.yml
```diff
@@ -67,7 +67,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          save-gradle-dependencies-cache: false
```
iNikem, line 70: `read-only: true`

.github/workflows/patch-release-build.yml
```diff
@@ -79,7 +79,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          read-only: true
```
iNikem, line 82: We don't want this task to override gradle dependencies.

.github/workflows/pr.yml
```diff
@@ -75,7 +75,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          save-gradle-dependencies-cache: false
```
iNikem, line 78: `read-only: true`

.github/workflows/release-build.yml
```diff
@@ -38,7 +38,6 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          read-only: true
```
iNikem, line 41: Please restore :)

iNikem (commented):
I agree with your usage of cache key. But all other cases should remain read-only

anuraaga:
Ah thanks! Let me start over

PR #3618, opened 2021-07-19
Replace save-gradle-dependencies: false with read-only: true

anuraaga:
Don't think we need the cache key change I tried in #3617 as long as read-only: true works well

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          save-gradle-dependencies-cache: false
+          read-only: true
 
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
@@ -153,7 +153,7 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: smokeTests
-          save-gradle-dependencies-cache: false
+          read-only: true
 
       - name: Cache Gradle Wrapper
         uses: actions/cache@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         uses: burrunan/gradle-cache-action@v1.10
         with:
           job-id: jdk${{ matrix.test-java-version }}
-          save-gradle-dependencies-cache: false
+          read-only: true
```
anuraaga, line 70: @iNikem Hmm - doesn't this mean our tests will never have `FROM-CACHE` on these builds since Java version is in cache key? Probably still faster than what we have right now though with the dependency downloading

anuraaga: If the plugin supports being defined multiple times in a job, one workaround that comes to mind is to have two cache steps, one with `save-local-build-cache: false` along with `read-only: true` and one with `save-gradle-dependencies-cache: false`

anuraaga: I don't think I've seen 5min builds in a long time so we seem to be on the right track :-) But this will only be true for the next couple of builds.

Let me try applying my idea and see what happens

iNikem: job-id is used to download local gradle build cache from GHA cache. I would expect that e.g. `javaCompile` still will be taken from remote Gradle build cache on S3. Test results will not be cached in this scenario.

anuraaga: Ah we have our S3 cache forgot about it. So yeah I think we're OK to rely on it. To confirm, what do you mean by "Test results will not be cached in this scenario."? They will be cached with S3 right? Caching the test results is critical I guess

iNikem: > They will be cached with S3 right?

I actually don't know :) https://github.com/burrunan/gradle-cache-action#how-does-github-actions-based-gradle-remote-build-cache-work

anuraaga: I guess let's try - since we're not using the remote caching feature I'm thinking that isn't applying to our build

iNikem (approved):

PR #3623, opened 2021-07-19
Add library instrumentation for Apache HTTPClient 4.3

anuraaga:
I referred to both opentracing and zipkin to help with corner cases but mostly wrote from scratch. Both of them have and lack features vs the other one with regards to apache.

For now I have copied much from 4.0 javaagent instrumentation, in a future PR will try to share in a common library.


instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/HostAndRequestAsHttpUriRequest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.http.HttpHost;
+import org.apache.http.HttpRequest;
+import org.apache.http.ProtocolVersion;
+import org.apache.http.RequestLine;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.message.AbstractHttpMessage;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** Wraps HttpHost and HttpRequest into a HttpUriRequest for tracers and injectors. */
+final class HostAndRequestAsHttpUriRequest extends AbstractHttpMessage implements HttpUriRequest {
```
anuraaga, line 19: I need to adopt @trask's new pattern for this class but otherwise think this PR is good for an initial look

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientRequest.java
```diff
@@ -0,0 +1,130 @@
...
+    if (major == 1 && minor == 0) {
+      return SemanticAttributes.HttpFlavorValues.HTTP_1_0;
+    }
+    if (major == 1 && minor == 1) {
+      return SemanticAttributes.HttpFlavorValues.HTTP_1_1;
+    }
+    if (major == 2 && minor == 0) {
+      return SemanticAttributes.HttpFlavorValues.HTTP_2_0;
+    }
+    logger.debug("unexpected http protocol version: " + protocolVersion);
```
laurit, line 102:
```suggestion
    logger.debug("unexpected http protocol version: {}", protocolVersion);
```

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/TracingProtocolExec.java
```diff
@@ -0,0 +1,172 @@
...
+    CloseableHttpResponse response = null;
+    Throwable error = null;
+    try (Scope ignored = context.makeCurrent()) {
+      response = exec.execute(route, request, httpContext, httpExecutionAware);
+      return response;
+    } catch (Throwable e) {
+      error = e;
+      throw e;
+    } finally {
+      if (!pendingRedirect(context, httpContext, request, instrumenterRequest, response)) {
```
laurit, line 117: It is unfortunate that we can't wrap around redirect and automatic retry. Perhaps we should just let this instrumentation create multiple spans when there is a redirect? If I understand correctly if retry is enable then we only record the first failing request even when retry succeeds?

anuraaga: Nah, the normal redirect tests in HttpClientTest are passing so we're recording it as a success. And the timing is around all the redirects since we end the span after the last one.

The code is sort of complex but I think the modeling is consistent with our other instrumentation so guess it's not too bad.

laurit: Yeah redirects look ok. There is also an option to retry request when it fails, imo it would need similar handling as redirects to make sure that we don't end span on the first attempt. We don't have a test for that.

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+  ApacheHttpClientTracingBuilder(OpenTelemetry openTelemetry) {
+    this.openTelemetry = openTelemetry;
+  }
+
+  /**
+   * Adds an additional {@link AttributesExtractor} to invoke to set attributes to instrumented
+   * items. The {@link AttributesExtractor} will be executed after all default extractors.
+   */
+  public ApacheHttpClientTracingBuilder addAttributeExtractor(
+      AttributesExtractor<? super ApacheHttpClientRequest, ? super HttpResponse>
```
mateuszrzeszutek, line 40: Should `ApacheHttpClientRequest` be public? We're exposing a possibility to add a custom attributes extractor, but you can't really use request in it.

instrumentation/apache-httpclient/apache-httpclient-4.3/testing/src/main/groovy/io/opentelemetry/instrumentation/apachehttpclient/v4_3/HttpUriRequest.groovy
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.apachehttpclient.v4_3
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
mateuszrzeszutek, line 10:
```suggestion
```

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/TracingProtocolExec.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  private boolean pendingRedirect(
+      Context context,
+      HttpClientContext httpContext,
+      HttpRequestWrapper request,
+      ApacheHttpClientRequest instrumenterRequest,
+      @Nullable CloseableHttpResponse response)
+      throws ProtocolException {
```
mateuszrzeszutek, line 129: This method can throw and is called in a `finally` block - would it make sense to catch whatever it throws and `addSuppressed()` to `error`? Or is it a case when you have to declare `throws` but nothing should ever fail here?

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/TracingProtocolExec.java
anuraaga: @laurit We discussed a bit and figure it's good to try this code first and iterate later, since we don't have any tests for retries yet. We'd probably need to check everything in the future more broadly

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/TracingProtocolExec.java
anuraaga: Thanks for noticing! Tightened up the below catch to prevent it from leaking, which would be bad for many reasons not just being in a finally (well technically being in the finally would save the app from crashing maybe :)

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
anuraaga: @trask This is an interesting case for wanting to go back to the HttpUriRequest interface perhaps :-)

For now I just made public with a single method returning the delegate.

mateuszrzeszutek (approved):

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/TracingProtocolExec.java
laurit: I just wanted to point out that with library instrumentation we have limited options to force the framework to do things as we like so we could instead create spans in a way that fits best with the integration points the framework provides us. This way there should be less corner cases we have to worry about. Sure it will be bad that different library instrumentations don't behave in a consistent way and even worse would be testing them, but instrumentation would be simpler and the chance of something breaking because http client is configured slightly differently or customized a bit should be smaller.

laurit (approved):

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
trask: i'm not sure wrangling into `HttpUriRequest` is ideal for custom attribute extractors either, I think ideally they would want access to `HttpHost` + `HttpRequest` in the case where an `HttpHost` + `HttpRequest` overload is used

PR #3625, opened 2021-07-19
Extension snapshot

### LINKED ISSUE #3503: Test if publishing works as expected

iNikem:
Fixes #3503

After this PR we always build examples against locally built and published snapshot.

In this PR examples does not use Sonatype snapshot repo, only `mavenCentral` and `mavenLocal`. This makes sure that when we want to test that the _current_ version of the our main repo publishes artifacts correctly, then we actually will test locally published versions and not something from Sonatype. The downside of that is that one cannot build examples until there is locally published version of the main project. This may be pretty big downside.

The better approach would be to use Sonatype snapshot repo by default but during publishing test somehow remove it and use only local repo. I see only one way to achieve that: via some shell script manipulating `build.gradle`, which is quite fragile IMO. All ideas are welcomed.


iNikem:
I solved the previous problem. Now all builds except nightly will consume snapshots from Sonatype repo. Nightly will be forced to use locally published one.

examples/distro/build.gradle
```diff
@@ -29,10 +29,8 @@ subprojects {
   }
 
   repositories {
-    maven {
-      url = uri("https://oss.sonatype.org/content/repositories/snapshots")
```
mateuszrzeszutek, line 33: Can we start using SNAPSHOTs in distro example too? (Separate PR?)

mateuszrzeszutek (approved):

examples/distro/build.gradle
iNikem: Absolutely, will fix right away

iNikem: Turns out, this cannot be done until we fix publishing of `agent-for-testing`.

anuraaga (approved):

examples/distro/build.gradle
iNikem: Done

mateuszrzeszutek (approved):

.github/scripts/local.init.gradle.kts
```diff
@@ -0,0 +1,10 @@
+addListener(object : BuildAdapter() {
+  override fun projectsEvaluated(gradle: Gradle) {
+    gradle.rootProject {
+      repositories {
+        mavenLocal()
+        removeIf { it.name == "sonatype" }
+      }
+    }
+  }
+})
```
trask, line 10: wow

trask (commented):
nice, this is awesome

PR #3627, opened 2021-07-19
Overhead testing foundations

breedx-splk:
~~This should be blocked by #3554 and #3640.~~ Good to review now.

It sets up the basic project which uses testcontainers and a set of test configurations to determine which test runs to perform. It stands up a collector, then a petclinic container with the given agent (which might be `none`) and then runs k6 against the running app. It repeats this for each configuration.

While I think the configuration stuff is pretty flexible, it might be a little overengineered and I'm looking for feedback. One of the goals is to eventually support direct http/file URLs to agents (vendors might benefit from this). 

The results processing bits are intended to come in a follow-up PR (there's a TODO). Right now, the results are just output to the workspace in the form of k6 json output and jfr files.


testing-overhead/src/test/java/io/opentelemetry/LatestAgentSnapshotResolver.java
```diff
@@ -0,0 +1,83 @@
...
+    String latestFilename = fetchLatestFilename(version);
+    String url = BASE_URL +  "/" + version + "/" + latestFilename;
+    byte[] jarBytes = fetchBodyBytesFrom(url);
+    Path path = Paths.get(".", "opentelemetry-javaagent-SNAPSHOT-all.jar");
+    FileOutputStream out = new FileOutputStream(path.toFile());
+    out.write(jarBytes);
+    return Optional.of(path);
+  }
+
+  private String fetchLatestFilename(String version) throws IOException {
```
iNikem, line 42: Please take a look at Eclipse Aether and how we use it `io.opentelemetry.instrumentation.muzzle-check.gradle.kts`. It may be easier/more robust than writing this code manually.

testing-overhead/src/test/java/io/opentelemetry/OverheadTests.java
```diff
@@ -0,0 +1,76 @@
...
+    GenericContainer<?> k6 = K6Container.builder()
+        .withNetwork(NETWORK)
+        .withConfig(config)
+        .withAgentNamed(agent)
+        .build();
+    k6.start();
+
+    // This is required to get a graceful exit of the VM before testcontainers kills it forcibly.
+    // Without it, our jfr file will be empty.
+    petclinic.execInContainer("kill", "1");
```
iNikem, line 69: Spring applications can have `/shutdown` endpoint to exit them gracefully.

testing-overhead/src/test/java/io/opentelemetry/PetClinicRestContainer.java
```diff
@@ -0,0 +1,110 @@
...
+    result.add("-jar");
+    result.add("/app/spring-petclinic-rest.jar");
+    return result.toArray(new String[] {});
+  }
+
+  static Builder builder() {
+    return new Builder();
+  }
+
+  static class Builder {
```
iNikem, line 78: Why do we need builders? We have all parameters mandatory, no? Just constructors will do then.

iNikem (commented):
Higher level feedback:

- Splitting into packages may be good. Currently I somehow got confused what classes are related. But maybe this is just GH :)
- Please don't forget to polish file operations. Currently they assume too much of the green path.

testing-overhead/src/test/java/io/opentelemetry/LatestAgentSnapshotResolver.java
breedx-splk: Thanks, I'll take a look. Writing this code definitely felt bad.

testing-overhead/src/test/java/io/opentelemetry/OverheadTests.java
breedx-splk: This allows testing of non spring apps too?  😁 
Since we have testcontainers, is there a reason _not_ to do this?

iNikem: Ok, fair. If we can assume that `kill 1` is a "valid" way to stop application :)

testing-overhead/src/test/java/io/opentelemetry/PetClinicRestContainer.java
breedx-splk: I'll nuke em.

testing-overhead/src/test/java/io/opentelemetry/LatestAgentSnapshotResolver.java
breedx-splk: Oof, [aether is archived](https://projects.eclipse.org/projects/technology.aether). Are we sure that we'd want to add another dependency on an archived (dead) project?  Last release was 6.5 years ago.  😢

iNikem: :( Probably not :(

breedx-splk: Well, if we find a better contemporary way of resolving latest/snapshot versions of jar files we can overhaul the resolver at a future time. At least it is abstracted.

testing-overhead/k6/basic.js
```diff
@@ -0,0 +1,70 @@
+import http from "k6/http";
```
laurit, line 1: perhaps it would be better to place test specific scripts to a common subdirectory so it would be easier to add more tests

testing-overhead/src/test/java/io/opentelemetry/config/TestConfig.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+
+/**
+ * Defines a test config.
+ */
+public class TestConfig {
+
+  private final static int DEFAULT_MAX_REQUEST_RATE = 0;  // none
+  private final static int DEFAULT_CONCURRENT_CONNECTIONS = 5;
+  private final static int DEFAULT_TOTAL_ITERATIONS = 500;
```
laurit, line 19: 500 iterations seems too little. Ideally there should be a warmup iteration before measurement iteration so that we would measure application in steady state where it has already jit compiled everything it can.

breedx-splk: That's reasonable -- I think I agree that 500 is too little and we should probably bump it up once things settle down and we're actually persisting data. For development, 500 is nice because you don't have to wait _too_ long.

As far as warmup -- I'm torn, because if the agent impacts startup and jit (with more code, the jitter will probably have to work harder, right?) -- maybe we _do_ want to account for that?  Does allowing the app to reach steady state seem slightly unfair? Maybe we're also talking about two different types of tests, so I could see this maybe becoming a parameter of the configuration.  🤔

trask: > Maybe we're also talking about two different types of tests

👍

testing-overhead/k6/basic.js
breedx-splk: that's the entire purpose of the `k6` subdir.

testing-overhead/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+plugins {
+  id("java")
```
anuraaga, line 2: Can we avoid the yet another gradle wrapper and use `otel.java-conventions`? Any particular reason for the separation?

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class AgentResolver {
+
+  private final LatestAgentSnapshotResolver snapshotResolver = new LatestAgentSnapshotResolver();
+
+  public Optional<Path> resolve(Agent agent) throws Exception {
+    if(Agent.NONE.equals(agent)){
+      return Optional.empty();
+    }
+    if(Agent.LATEST_SNAPSHOT.equals(agent)){
```
anuraaga, line 25: Instead of snapshot, which includes the annoying resolver, how about just building from this repo and using that agent, similar to our smoke tests? The snapshot can get old due to main build failures such as testLatestDeps so that would be easier to reason about which agent commit is actually teste

testing-overhead/src/test/java/io/opentelemetry/config/TestConfig.java
laurit: Our users are likely to care less about startup overhead than steady state performance.

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
```diff
@@ -0,0 +1,55 @@
...
+      return snapshotResolver.resolve();
+    }
+    if(agent.hasUrl()){
+      return Optional.of(downloadAgent(agent));
+    }
+    throw new IllegalArgumentException("Unknown agent: " + agent);
+  }
+
+  Path downloadAgent(Agent agent) throws Exception {
```
mateuszrzeszutek, line 34: Would it make sense to pass just `URL` here? You've already checked that `agent.hasUrl()` and you're only using the URL

testing-overhead/src/test/java/io/opentelemetry/agents/LatestAgentSnapshotResolver.java
```diff
@@ -0,0 +1,83 @@
...
+
+  Optional<Path> resolve() throws IOException {
+    String version = fetchLatestSnapshotVersion();
+    logger.info("Latest snapshot version is {}", version);
+    String latestFilename = fetchLatestFilename(version);
+    String url = BASE_URL +  "/" + version + "/" + latestFilename;
+    byte[] jarBytes = fetchBodyBytesFrom(url);
+    Path path = Paths.get(".", "opentelemetry-javaagent-SNAPSHOT-all.jar");
+    FileOutputStream out = new FileOutputStream(path.toFile());
+    out.write(jarBytes);
```
mateuszrzeszutek, line 38:
```suggestion
    try (OutputStream out = Files.newOutputStream(path)) {
      out.write(jarBytes);
    }
```

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
```diff
@@ -0,0 +1,55 @@
...
+    byte[] raw = response.body().bytes();
+    Path path = Paths.get(".", "opentelemetry-javaagent-all.jar");
+    FileOutputStream out = new FileOutputStream(path.toFile());
+    out.write(raw);
+    out.flush();
+    out.close();
+    return path;
+  }
+
+  // TODO: Download latest snapshot
```
mateuszrzeszutek, line 53: Isn't that already done in another class?

mateuszrzeszutek (commented):
WDYT about renaming the module to `benchmark-overhead`? Usually our `testing...` modules are for correctness testing, we could keep all performance testing in `benchmark...` modules (also see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3672#issuecomment-886575259)

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
breedx-splk: Yup, I'll remove that dead comment.

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+    Request request = new Request.Builder().url(agent.getUrl()).build();
+    OkHttpClient client = new OkHttpClient();
+    Response response = client.newCall(request).execute();
+    byte[] raw = response.body().bytes();
+    Path path = Paths.get(".", "opentelemetry-javaagent-all.jar");
+    FileOutputStream out = new FileOutputStream(path.toFile());
+    out.write(raw);
+    out.flush();
+    out.close();
```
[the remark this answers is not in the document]
breedx-splk: Even better, I think since all the bytes are in memory we can just `Files.write(path, bytes)`.  Simple!

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
breedx-splk: Sure. It's funny, I had it that way and factored it back. Not sure why. Agree that url makes more sense here, since this method should be private.

testing-overhead/src/test/java/io/opentelemetry/config/TestConfig.java
breedx-splk: Ok. @laurit would you be ok with deferring a warmup enhancement for a future PR?

testing-overhead/build.gradle.kts
breedx-splk: Would that imply adding this subdir as a submodule of the main project? 

My desire for keeping this stand-alone is primarily motivated by the massive size of the base project and the fact that it commonly takes 9000 years to open in IDEA and index.  It's *MUCH* easier/simpler/faster to work in this project when the rest of the instrumentation is not in scope.

I also understand/disgust in having so many gradle wrappers in the repo...but is it really worse than having everything in one MASSIVE gradle project? I dunno. Curious what others think.

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
breedx-splk: I had really hoped to avoid this and to keep this test process decoupled from the main build.  It's going to slow things down considerably.

Is the thought that regressions would be harder to track back to commits or something?

breedx-splk:
> WDYT about renaming the module to `benchmark-overhead`? Usually our `testing...` modules are for correctness testing, we could keep all performance testing in `benchmark...` modules (also see [#3672 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3672#issuecomment-886575259))

I'm fine wit that changes, I wasn't aware of the subtle semantic difference...so that's cooll.  *However* I do have a pile of commits stacked up in #3648 and I'm skeptical of my abilities to cleanly rebase/pick them after a name change. My preference would be to do the rename in a future PR.

trask:
> I do have a pile of commits stacked up in #3648 and I'm skeptical of my abilities to cleanly rebase/pick them after a name change. My preference would be to do the rename in a future PR.

sounds good 👍

testing-overhead/build.gradle.kts
iNikem: If adding one gradle dependency to this separate project requires re-importing of all our 100s of instrumentation - that is much larger overkill to me than yet another gradle wrapper.

Also, don't forget that you don't actually _need_ a separate wrapper for every independent project. `../gradlwe build` works just fine.

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
iNikem: I support Jason in his reluctance to build local project every time.

iNikem (approved):
I am OK with this as POC to get MVP running. We then can iterate.

trask (commented):
> I am OK with this as POC to get MVP running. We then can iterate.

👍 

@breedx-splk can you open an issue to (briefly) summarize open questions from above? then let's merge

breedx-splk:
Thanks everyone. I captured the open things as 3 issues and assigned them to me for follow-up.

PR #3637, opened 2021-07-20
Don't use daemon for muzzle tasks

iNikem:
There are almost a minute long pause between gradle logs of starting the daemon and the next line:

```
Tue, 20 Jul 2021 08:11:52 GMT
Starting a Gradle Daemon (subsequent builds will be faster)
22
Tue, 20 Jul 2021 08:12:52 GMT
> Task :buildSrc:extractPrecompiledScriptPluginPlugins FROM-CACHE
```

Let's see if this change makes any difference


iNikem:
Did not make any difference

PR #3639, opened 2021-07-20
Use published muzzle generation plugin

iNikem:


trask:
this would be amazing if true 😂
image

anuraaga:
@inikem I thought you were going to apply this plugin to the distro example. Do we have any hope for applying it to our main build without the substitutions stuff?

iNikem:
> @iNikem I thought you were going to apply this plugin to the distro example. 

I am applying it to extensions example. May do the same with distro as well.

> Do we have any hope for applying it to our main build without the substitutions stuff?

Sorry, what substitution?

anuraaga:
> I am applying it to extensions example. May do the same with distro as well.

Ah thought you're applying to the main build since see changes to buildSrc like https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3639/files#diff-7b5c40eb5522b5216cfe8c75612e579a8db0daee9ded80df9ce6a56e8361a962L30

> Sorry, what substitution?

Substitution of e.g. `io.opentelemetry.javaagent:opentelemetry-muzzle` from published version to project version

iNikem:
I am applying to both, main project and extension as external example.

I can try to remove the substitution, but why? I thought you did want to use local projects when possible?

anuraaga:
> I can try to remove the substitution, but why? I thought you did want to use local projects when possible?

Yup we do - I didn't know we confirmed we can actually do it with the build file dependencies like

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/gradle-plugins/build.gradle.kts#L24

But it it already works then it's great! Just want to make sure you're not stuck on it and can help if needed

iNikem:
The last commit is quite dirty. I had to comment out some tests and duplicate several test classes. I will be able to clean up that a lot after I move muzzle checking plugin as well. Which I will do next.

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
```diff
@@ -1,12 +1,15 @@
...
 }
 
 dependencies {
   add("muzzleBootstrap", "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api-annotation-support")
   add("muzzleBootstrap", "io.opentelemetry.javaagent:opentelemetry-javaagent-instrumentation-api")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
+  add("muzzleTooling", "org.slf4j:slf4j-simple")
+
+  add("codegen", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
```
anuraaga, line 14: Can you add some comments to describe the dependencies and where they are used? In particular helpful to describe the difference between this dependency and the one in build.gradle, which I guess are related somehow

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/gradle-plugins/build.gradle.kts#L24

iNikem: PTAL

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
```diff
@@ -1,12 +1,21 @@
...
   add("muzzleBootstrap", "io.opentelemetry.javaagent:opentelemetry-javaagent-instrumentation-api")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-extension-api")
   add("muzzleTooling", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
+  add("muzzleTooling", "org.slf4j:slf4j-simple")
+
+  /*
+    Dependencies added to this configuration will be found by the muzzle gradle plugin during code
+    generation phase. They differ from the implementation dependencies declared in plugin's build
+    script, because these classes become part of the code that plugin inspects and traverses during
+    references collection phase. They are not part of the observer, they are part of the observation.
```
anuraaga, line 18: Is javaagent-extension-api needed in the buildSrc dependencies? It seems like something the observer wouldn't need, only the observation.

anuraaga (approved):

buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts
iNikem: Are you talking about `muzzleTooling` dependency above? That is for muzzle checking, I have not touched it yet.

anuraaga: Sorry I didn't mean buildSrc, but gradle-plugins (it feels like buildSrc :P) Is this one needed?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/gradle-plugins/build.gradle.kts#L25

iNikem: Yes, `MuzzleCodeGenerator` uses it:

```
import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
```

Take a look at this method's content: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/85b7c0569e7fa10aaa97ca9acc07f830753dfb0f/gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/generation/MuzzleCodeGenerator.java#L161

PR #3640, opened 2021-07-20
Add gh action to publish base image for overhead testing

breedx-splk:
In support of overhead testing, we want to have this petclinic-rest base image handy, so this creates an action to build it and publish to ghcr.

I don't expect needing to change this very often, so it's just left as a manual trigger for now.

### DIFF
--- a/.github/workflows/build-petclinic-overhead-base.yml
+++ b/.github/workflows/build-petclinic-overhead-base.yml
@@ -1,0 +1,25 @@
+name: publish-petclinic-base-image
+on: workflow_dispatch
+jobs:
+  push_to_registry:
+    name: make spring-petclinic-rest base image for overhead testing
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+      contents: read
+    steps:
+      - name: check out the repo
+        uses: actions/checkout@v2
+      - uses: docker/setup-buildx-action@v1
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: push to gh packages
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: testing-overhead/Dockerfile-petclinic-base
+          tags: ghcr.io/open-telemetry/opentelemetry-java-instrumentation/petclinic-rest-base:latest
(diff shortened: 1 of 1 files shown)


jkwatson (approved):
Looks ok to me

.github/workflows/build-petclinic-overhead-base.yml
```diff
@@ -0,0 +1,25 @@
+name: publish-petclinic-base-image
+on: workflow_dispatch
```
anuraaga, line 2: How about adding a path trigger?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/.github/workflows/build-smoke-dist-fakebackend.yml#L4

breedx-splk: I love it!  Thanks for the idea!

.github/workflows/build-petclinic-overhead-base.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: push to gh packages
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          file: testing-overhead/Dockerfile-petclinic-base
+          tags: ghcr.io/open-telemetry/opentelemetry-java-instrumentation/petclinic-rest-base:latest
```
anuraaga, line 30: I might be missing a new feature in GitHub which would be cool! But if not, we only have one image we use, `java-test-containers` and just use a tag per different image

https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/docker-test-containers.yml#L33

ghcr is scoped to organization, or at least it was, so rather than asking for a new image each time we just do this.

breedx-splk: Yeah, it's scoped to org, but I _think_ this makes a package called `opentelemetry-java-instrumentation/petclinic-rest-base` - like the `/` is part of the name?  I can change it, but calling it `testcontainers` when it's really just petclinic seems misleading. 

How would you feel about trying the original approach and if it doesn't work I'll fix it?

iNikem: The problem is that we cannot automatically create new packages in otel org. Only org admins can do that.

iNikem: But we can try :)

iNikem (approved):

iNikem:
@breedx-splk please rebase to fix muzzle failure

PR #3648, opened 2021-07-21
Overhead results collector

breedx-splk:
This builds on top of #3627 and will need a rebase after that is ready.

This is now able to gather up the results into a data object that can then be handled by a persistence mechanism. Eventually that will be the repo itself, but for now it just outputs to the console.


testing-overhead/src/test/java/io/opentelemetry/containers/K6Container.java
```diff
@@ -14,22 +15,25 @@
...
+  public K6Container(Network network, Agent agent, TestConfig config, NamingConvention namingConvention) {
     this.network = network;
     this.agent = agent;
     this.config = config;
+    this.namingConvention = namingConvention;
   }
 
   public GenericContainer<?> build(){
-    String k6OutputFile = "/results/k6_out_" + agent + ".json";
+    Path k6OutputFile = namingConvention.k6Results(agent);
```
mateuszrzeszutek, line 36: Nit: it'd be pretty good to hide all paths behind `NamingConvention`, for example

```java
.withFileSystemBind(localNamingConvention.rootDir(), containerNamingConvention.rootDir())
```

testing-overhead/src/test/java/io/opentelemetry/results/AppPerfResults.java
```diff
@@ -0,0 +1,179 @@
...
+    private TestConfig config;
+    private double iterationAvg;
+    private double iterationP95;
+    private double requestAvg;
+    private double requestP95;
+    private long totalGCTime;
+    private long totalAllocated;
+    private MinMax heapUsed;
+    private float maxThreadContextSwitchRate;
+    public long peakThreadCount;
```
mateuszrzeszutek, line 108: `public`?

testing-overhead/src/test/java/io/opentelemetry/util/JFRUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public static MinMax findMinMax(Path jfrFile, String eventName, String valueKey) throws IOException {
+    return reduce(jfrFile, eventName, valueKey, new MinMax(), (MinMax acc, Long value) -> {
+      if (value > acc.max) {
+          acc.max = value;
+        }
+        if (value < acc.min) {
+          acc.min = value;
+        }
+        return acc;
```
mateuszrzeszutek, line 33: It'd be more functional-style to return a new `MinMax` each time (since you're already using functional programming concepts like `reduce()` function 😛 ); that'd also solve one minor problem with `AppPerfResults`: it appears to be an immutable value class but in fact isn't, because `MinMax` is mutable.

testing-overhead/src/test/java/io/opentelemetry/results/ResultsCollector.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class ResultsCollector {
+
+  private final NamingConvention namingConvention;
+
+  public ResultsCollector(NamingConvention namingConvention) {this.namingConvention = namingConvention; }
+
+  public Map<Agent, AppPerfResults> collect(TestConfig config) {
```
mateuszrzeszutek, line 25: `AppPerfResults` already contains the `Agent`, why not return `List<AppPerfResults>` instead?

Or you could go a bit further and return sth like this:

```java
class AppPerfResults {
  TestConfig config;
  List<AgentResults> agentResults;
}

class AgentResults {
  Agent agent;
  // ... and all other fields from current AppPerfResults
}
```

mateuszrzeszutek (commented):
Also, is that `.gitmodules` file needed?

testing-overhead/src/test/java/io/opentelemetry/results/AppPerfResults.java
breedx-splk: LOL thanks.

testing-overhead/src/test/java/io/opentelemetry/util/JFRUtils.java
breedx-splk: ooh, I like this. Lemme do that. 👍🏻

testing-overhead/src/test/java/io/opentelemetry/results/ResultsCollector.java
breedx-splk: Changing the Map -> List worked out very nicely. I'm not ready to commit to the latter change yet tho 😁 , but I do see where you're coming from (the config is redundant).

breedx-splk:
> Also, is that `.gitmodules` file needed?

Nope, I think that was an artifact from when I first had petclinic as a submodule. I've removed it again, hopefully this time for good...it might be a special file tho, so we'll see. 🤞🏻

testing-overhead/src/test/java/io/opentelemetry/containers/K6Container.java
breedx-splk: I'm not sure about _all_, since there are a lot of moving parts and I could see it exploding pretty quickly, but I'm going to make this change (and the corresponding one in the petclinic container too) and we improve upon this over time if it feels useful.

mateuszrzeszutek (approved):

testing-overhead/src/test/java/io/opentelemetry/OverheadTests.java
```diff
@@ -71,4 +84,10 @@ void runAppOnce(TestConfig config, Agent agent) throws Exception {
 
     //TODO: Parse and aggregate the test results.
   }
+
+  private void writeStartupTimeFile(Agent agent, long start) throws IOException {
```
iNikem, line 88: Startup time of Spring applications can be parsed from the logs. Current approach of relying on testcontainers check depends on the frequency of polling, so has higher granularity

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
```diff
@@ -8,8 +8,8 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
+import java.nio.file.StandardCopyOption;
```
iNikem, line 12: That is certainly not an alphabetical order :)

testing-overhead/src/test/java/io/opentelemetry/results/AppPerfResults.java
```diff
@@ -0,0 +1,200 @@
...
+  private final double iterationAvg;
+  private final double iterationP95;
+  private final double requestAvg;
+  private final double requestP95;
+  private final long totalGCTime;
+  private final long totalAllocated;
+  private final MinMax heapUsed;
+  private final float maxThreadContextSwitchRate;
+  private final long startupDurationMs;
+  private final long peakThreadCount;
```
iNikem, line 23: `public final` and remove all those getters?

testing-overhead/src/test/java/io/opentelemetry/util/JFRUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.function.BiFunction;
+import io.opentelemetry.results.AppPerfResults.MinMax;
+import jdk.jfr.consumer.RecordedEvent;
+import jdk.jfr.consumer.RecordingFile;
+
+public class JFRUtils {
+
+  public static long totalLongEvents(Path jfrFile, String eventName,
```
iNikem, line 16:
```suggestion
  public static long countLongEvents(Path jfrFile, String eventName,
```

testing-overhead/src/test/java/io/opentelemetry/util/NamingConvention.java
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package io.opentelemetry.util;
+
+import io.opentelemetry.agents.Agent;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class NamingConvention {
```
iNikem, line 11: Javadoc please

testing-overhead/src/test/java/io/opentelemetry/util/NamingConventions.java
```diff
@@ -0,0 +1,15 @@
+package io.opentelemetry.util;
+
+public class NamingConventions {
```
iNikem, line 3: Javadoc please

iNikem (approved):

testing-overhead/src/test/java/io/opentelemetry/results/AppPerfResults.java
breedx-splk: I did use direct field access on the `MinMax` class, but getters seem way more idiomatic. I'm totally fine with direct field access for immutable things, but I thought it would be discouraged. :). What do other folks think?

testing-overhead/src/test/java/io/opentelemetry/agents/AgentResolver.java
breedx-splk: 🙈 
This is the best comment I will read this year.  😝

testing-overhead/src/test/java/io/opentelemetry/OverheadTests.java
breedx-splk: True. I'm not currently pulling in the logs, so the time isn't readily available for collection. I could look into that, but I honestly don't think we should be too concerned with subsecond granularity. We could also possibly make testcontainers poll more than 1s. 

If you think it's important can you please open an issue to track it?  Thanks!

testing-overhead/src/test/java/io/opentelemetry/util/NamingConvention.java
breedx-splk: For a test utility class?  Ok, I suppose I can do that.....

iNikem: Utility class is irrelevant here. We need to explain what is this naming convention and why do we need it.

testing-overhead/src/test/java/io/opentelemetry/results/AppPerfResults.java
mateuszrzeszutek: Whatever, as long as it is immutable 😄

testing-overhead/src/test/java/io/opentelemetry/util/NamingConventions.java
breedx-splk: ok.

testing-overhead/src/test/java/io/opentelemetry/util/JFRUtils.java
breedx-splk: Count is incorrect, it does a **sum** of the values, not just count them. Maybe you like `sumLongEvents` better? Very subtle distinction between summing and totaling, but I can change it if you like one better.

PR #3651, opened 2021-07-22
Handle port and IPv6 in forwarded headers

trask:


iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java
```diff
@@ -239,9 +235,31 @@ static String extractForwardedFor(String forwarded) {
...
+  // VisibleForTesting
+  static String extractForwardedFor(String forwarded) {
+    return extractIpAddress(forwarded, 0);
+  }
+
+  private static String extractIpAddress(String forwarded, int start) {
+    if (forwarded.length() == start) {
+      return null;
+    }
+    if (forwarded.charAt(start) == '[') {
```
laurit, line 250: Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded and https://www.rfc-editor.org/rfc/rfc7239 it seems to me that ipv6 addresses need to be quoted, for example `for="[2001:db8:cafe::17]"`. Similarly when ipv4 address includes a port it should be quoted `"192.0.2.43:47011"`

trask: good catch, fixed, and yet more tests added...

laurit (approved):

mateuszrzeszutek:
Should we add the same thing to the Instrumenter API?

PR #3652, opened 2021-07-22
Convert HttpClientTest to JUnit

anuraaga:
While there are some issues with verbosity of Java assertions, I think we do have the goal of presenting our test framework, the base classes, for use from Java tests, especially since we can expect external Java instrumentation to need them. Hopefully it's ok.

The Groovy API is mostly preserved, but the big difference is the assertion failures. The assertions print for each layer in the stack (traces, trace, span, attributes for example) and tend to have the most important information at the bottom. This isn't so terrible except that if there is an exception event in the data, then a huge exception stack trace gets printed multiple times. I'd like to continue to improve on this in the sdk-testing package (and am glad to have fixed a few gaps through the process of migrating this test). On the bright side, at least for me, an assertion failure like "NET_PEER_PORT expected to be <443> but was <-1>" is much easier to understand how to fix than the Spock assertions that dump a lot of information and have dashes trying to point out what is wrong.

A big cleanup I'll follow up with is replacing all the Test subclass method knobs with a single `HttpClientOptions` / Builder class which will reduce the forwarding, and also improve the readability of the subclasses I think.

I'll also look through the TestRunner code a bit, I moved some code around in this PR but think there is still a bit more shuffling to make it consistent (some logic that could have been in runner was duplicated in the JUnit and Spock calling classes).


testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -305,97 +241,26 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-  Set<AttributeKey<?>> httpAttributes(URI uri) {
-    [
-      SemanticAttributes.HTTP_URL,
-      SemanticAttributes.HTTP_METHOD,
-      SemanticAttributes.HTTP_FLAVOR,
-      SemanticAttributes.HTTP_USER_AGENT
-    ]
-  }
-
   def "basic #method request #url"() {
```
trask, line 244:
```suggestion
  def "basic GET request #path"() {
```

trask (approved):
> While there are some issues with verbosity of Java assertions, I think we do have the goal of presenting our test framework, the base classes, for use from Java tests, especially since we can expect external Java instrumentation to need them. Hopefully it's ok.

👍 

> I'd like to continue to improve on this in the sdk-testing package (and am glad to have fixed a few gaps through the process of migrating this test).

all of these improvements in sdk-testing is great 👍

anuraaga:
Thanks @trask. @mateuszrzeszutek would be great to get your review too since this touches the test runner stuff :)

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/AbstractHttpClientTest.java
```diff
@@ -0,0 +1,1220 @@
...
+                        .hasAttributesSatisfying(
+                            attrs -> assertThat(attrs).containsEntry("test.request.id", requestId)),
+                span ->
+                    span.hasName("child")
+                        .hasKind(SpanKind.INTERNAL)
+                        .hasParentSpanId(rootSpan.getSpanId()));
+          });
+    }
+
+    testing.waitAndAssertTraces(assertions.toArray(new Consumer[0]));
```
mateuszrzeszutek, line 846: Should we add a `waitAndAssertTraces(Collection< TraceAssert>)` variant?

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/AbstractHttpClientTest.java
```diff
@@ -0,0 +1,1220 @@
...
+    Map<String, String> headers = new HashMap();
+    for (String field :
+        testing.getOpenTelemetry().getPropagators().getTextMapPropagator().fields()) {
+      headers.put(field, "12345789");
+    }
+    REQUEST request = buildRequest(method, uri, headers);
+    return sendRequest(request, method, uri, headers);
+  }
+
+  // ideally private, but then groovy closures in this class cannot find them
```
mateuszrzeszutek, line 1186: Comment no longer applies?
```suggestion
```

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/AbstractHttpClientTest.java
```diff
@@ -0,0 +1,1220 @@
...
+  private RequestResult doRequestWithCallback(String method, URI uri, Runnable callback) {
+    return doRequestWithCallback(method, uri, Collections.emptyMap(), callback);
+  }
+
+  // ideally private, but then groovy closures in this class cannot find them
```
mateuszrzeszutek, line 1191:
```suggestion
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -405,222 +275,52 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-      }
-    }
-
-    where:
-    method = "GET"
+    expect:
+    junitTest.requestWithCallbackAndNoParent()
   }
 
   def "basic #method request with 1 redirect"() {
```
mateuszrzeszutek, line 290:
```suggestion
  def "basic request with 1 redirect"() {
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -405,222 +275,52 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-      }
-    }
-
-    where:
-    method = "GET"
+    expect:
+    junitTest.basicRequestWith1Redirect()
   }
 
   def "basic #method request with 2 redirects"() {
```
mateuszrzeszutek, line 296:
```suggestion
  def "basic request with 2 redirects"() {
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -405,222 +275,52 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-      }
-    }
-
-    where:
-    method = "GET"
+    expect:
+    junitTest.basicRequestWith2Redirects()
   }
 
   def "basic #method request with circular redirects"() {
```
mateuszrzeszutek, line 302:
```suggestion
  def "basic request with circular redirects"() {
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -405,222 +275,52 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-
-    where:
-    method = "GET"
+    assumeTrue(testRedirects())
+    assumeTrue(testCircularRedirects())
+    expect:
+    junitTest.circularRedirects()
   }
 
   def "redirect #method to secured endpoint copies auth header"() {
```
mateuszrzeszutek, line 309:
```suggestion
  def "redirect to secured endpoint copies auth header"() {
```

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -814,270 +377,63 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-        }
-      }
-    }
+    expect:
+    junitTest.highConcurrency()
   }
 
   def "high concurrency test with callback"() {
-    setup:
     assumeTrue(testCausality())
```
mateuszrzeszutek, line 386: Is it necessary to duplicate assumptions? The JUnit delegate test already calls them.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -814,270 +377,63 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
...
-    }
+    expect:
+    junitTest.highConcurrencyOnSingleConnection()
   }
 
-  //This method should create either a single connection to the target uri or a http client
-  //which is guaranteed to use the same connection for all requests
-  SingleConnection createSingleConnection(String host, int port) {
-    return null
+  // ideally private, but then groovy closures in this class cannot find them
```
mateuszrzeszutek, line 405: Is this comment still valid? This class doesn't call `doRequest()`, tests extending this one call it.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
```diff
@@ -1139,11 +495,34 @@ abstract class HttpClientTest<REQUEST> extends InstrumentationSpecification {
     return true
   }
 
-  URI removeFragment(URI uri) {
-    return new URI(uri.scheme, null, uri.host, uri.port, uri.path, uri.query, null)
+  Throwable clientSpanError(URI uri, Throwable exception) {
+    return exception
+  }
+
+  // parent span must be cast otherwise it breaks debugging classloading (junit loads it early)
```
mateuszrzeszutek, line 502: Is this comment still valid? Now that we have a separate testing java agent there should be no problem with loading that stuff early (probably...)

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/InstrumentationTestRunner.java
```diff
@@ -34,6 +43,38 @@
 
   boolean forceFlushCalled();
 
+  /** Return a list of all captured traces, where each trace is a sorted list of spans. */
+  default List<List<SpanData>> traces() {
+    return TelemetryDataUtil.groupTraces(getExportedSpans());
+  }
+
```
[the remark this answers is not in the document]
anuraaga: I lean towards not bothering for these alpha modules - I figure there is lots of similar cruft that could be marked deprecated but we'll just remove :P

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/HttpClientLibraryInstrumentationExtension.java
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.testing.junit;
+
+import io.opentelemetry.instrumentation.testing.LibraryTestRunner;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+public final class HttpClientLibraryInstrumentationExtension extends InstrumentationExtension {
```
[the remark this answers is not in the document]
anuraaga: Nice! Will do the same for the current extensions in a different PR

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpClientTest.groovy
anuraaga: Unfortunately, the jupiter assumption exception is not recognized by junit4 / spock :(

mateuszrzeszutek (approved):

PR #3663, opened 2021-07-23
Don't wait in busy loop

laurit:


instrumentation/finatra-2.9/javaagent/src/test/groovy/FinatraServerTest.groovy
```diff
@@ -43,6 +43,7 @@ class FinatraServerTest extends HttpServerTest<HttpServer> implements AgentTestT
       if (System.currentTimeMillis() > startupDeadline) {
         throw new TimeoutException("Timed out waiting for server startup")
       }
+      Thread.sleep(1000)
```
trask, line 46: just to save a little waiting
```suggestion
      Thread.sleep(100)
```

trask (approved):

instrumentation/finatra-2.9/javaagent/src/test/groovy/FinatraServerTest.groovy
anuraaga: Can we switch to awaitility?

laurit: switched to awaitility

anuraaga (approved):

PR #3669, opened 2021-07-25
Okhttp3: fix concurrency test with callback

laurit:


trask (approved):
💯

I opened #3671 to track the issue for OkHttp library instrumentation

anuraaga (approved):

PR #3672, opened 2021-07-25
Add basic jmh overhead test

trask:
I did a bit of profiling, and planning to submit a couple of (micro) optimizations, so want to first submit the test I'm using for profiling/measurement.

```
Benchmark                                                               Score      Error   Units
ServletBenchmark.execute                                              139.343 ±    0.138   us/op
ServletBenchmark.execute:execute·p0.50                                125.696              us/op
ServletBenchmark.execute:·gc.alloc.rate.norm                        29111.284 ± 7057.135    B/op
ServletWithAgentDisabledBenchmark.execute                             119.724 ±    0.112   us/op
ServletWithAgentDisabledBenchmark.execute:execute·p0.50               110.464              us/op
ServletWithAgentDisabledBenchmark.execute:·gc.alloc.rate.norm       25580.861 ± 5803.009    B/op
ServletWithOnePercentSamplingBenchmark.execute                        138.331 ±    0.135   us/op
ServletWithOnePercentSamplingBenchmark.execute:execute·p0.50          124.928              us/op
ServletWithOnePercentSamplingBenchmark.execute:·gc.alloc.rate.norm  28508.222 ± 6843.209    B/op
```


buildSrc/src/main/kotlin/otel.jmh-conventions.gradle.kts
```diff
@@ -0,0 +1,36 @@
+plugins {
+  id("me.champeau.jmh")
+  id("io.morethan.jmhreport")
+}
+
```
trask, line 5: this file is identical copy from opentelemetry-java repo

testing-overhead-jmh/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+    options.errorprone {
+      isEnabled.set(false)
+    }
+  }
+
+  named<me.champeau.jmh.JMHTask>("jmh") {
+    dependsOn(copyAgent)
+
+    // TODO(trask) is this ok? if it's ok, move to otel.jmh-conventions?
+    outputs.upToDateWhen { false }
```
trask, line 52: gradle experts: is this ok? each time I run `jmh` task I actually want it to run. using `--rerun-tasks` takes a long time because it doesn't only rerun this one task

testing-overhead-jmh/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+tasks {
+  val copyAgent by registering(Copy::class) {
+    into(file("$buildDir/agent"))
+    from(configurations.named("agentJar"))
+    rename { file: String ->
+      file.replace("-$version", "")
+    }
+
+    dependsOn(":javaagent:shadowJar")
+  }
```
trask, line 37: gradle experts: let me know if there is a more preferred way to get the opentelemetry-javaagent-all into a location that the jmh benchmark can depend on

testing-overhead-jmh/src/jmh/java/io/opentelemetry/javaagent/benchmark/ServletWithAgentBenchmark.java
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.benchmark;
+
+import org.openjdk.jmh.annotations.Fork;
+
+@Fork(
+    jvmArgsPrepend = {
+      "-javaagent:build/agent/opentelemetry-javaagent-all.jar",
```
anuraaga, line 12: Is it better to set these in the Gradle file instead of in the annotation? Then you wouldn't need the copyAgent stuff

trask: I gave it a try, I think it's better now(?), at least it matches smoke-tests, which is good

buildSrc/src/main/kotlin/otel.jmh-conventions.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+jmh {
+  failOnError.set(true)
+  resultFormat.set("JSON")
+  // Otherwise an error will happen:
+  // Could not expand ZIP 'byte-buddy-agent-1.9.7.jar'.
+  includeTests.set(false)
+  profilers.add("gc")
+  val jmhIncludeSingleClass: String? by project
+  if (jmhIncludeSingleClass != null) {
+    includes.add(jmhIncludeSingleClass as String)
```
iNikem, line 23: Just curious: why this `as String` is needed? It is already clear that this is not null.

testing-overhead-jmh/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+  // TODO(trask) without disabling errorprone, jmh task fails with
+  //  Task :testing-overhead-jmh:jmhCompileGeneratedClasses FAILED
+  //   error: plug-in not found: ErrorProne
+  withType<JavaCompile>().configureEach {
+    options.errorprone {
+      isEnabled.set(false)
+    }
+  }
+
+  named<me.champeau.jmh.JMHTask>("jmh") {
```
iNikem, line 30: Can you use import here?

iNikem (approved):

buildSrc/src/main/kotlin/otel.jmh-conventions.gradle.kts
anuraaga: IIRC, the `by` syntax seems to result in something "dynamic". Seems like it might be a Gradle bug, but it makes kotlin not treat the null check as a type cast.

testing-overhead-jmh/build.gradle.kts
anuraaga: Ditto for ShadowJar

anuraaga (approved):

testing-overhead-jmh/src/jmh/java/io/opentelemetry/javaagent/benchmark/servlet/ServletWithAgentDisabledBenchmark.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.benchmark.servlet;
+
+import org.openjdk.jmh.annotations.Fork;
+
+@Fork(jvmArgsAppend = "-Dotel.javaagent.enabled=false")
```
mateuszrzeszutek, line 10: Just an idea: we could also run a benchmark with `-Dotel.javaagent.experimental.use-noop-api=true` to see how much overhead is added when no SDK at all is used.

testing-overhead-jmh/src/jmh/java/io/opentelemetry/javaagent/benchmark/servlet/ServletBenchmark.java
```diff
@@ -0,0 +1,46 @@
...
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+
+@BenchmarkMode(Mode.SampleTime)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Thread)
+public class ServletBenchmark {
+
+  static {
+    HelloWorldApplication.main();
```
mateuszrzeszutek, line 26: Just wondering: why do we initialize the app in `static {}` and not `@Setup`?

mateuszrzeszutek (approved):

mateuszrzeszutek:
Just thought about it now, but should we name the dir/module `benchmark-overhead-jmh`? Usually our `testing...` modules contain code that tests for correctness, we could keep everything that tests for performance in `benchmark...` modules.

testing-overhead-jmh/src/jmh/java/io/opentelemetry/javaagent/benchmark/servlet/ServletWithAgentDisabledBenchmark.java
trask: added 👍

testing-overhead-jmh/src/jmh/java/io/opentelemetry/javaagent/benchmark/servlet/ServletBenchmark.java
trask: added comment to explain

```
    // using static initializer instead of @Setup since only want to initialize the app under test
    // once regardless of @State and @Threads
```

trask:
> Just thought about it now, but should we name the dir/module `benchmark-overhead-jmh`? Usually our `testing...` modules contain code that tests for correctness, we could keep everything that tests for performance in `benchmark...` modules.

ya, I like this, updated

trask:
here's a re-run, including the no-op api test:

```
Benchmark                                                               Score      Error   Units
ServletBenchmark.execute                                              138.941 ±    0.137   us/op
ServletBenchmark.execute:execute·p0.50                                125.824              us/op
ServletBenchmark.execute:·gc.alloc.rate.norm                        29090.156 ± 7065.042    B/op
ServletWithAgentDisabledBenchmark.execute                             122.255 ±    0.107   us/op
ServletWithAgentDisabledBenchmark.execute:execute·p0.50               112.768              us/op
ServletWithAgentDisabledBenchmark.execute:·gc.alloc.rate.norm       25585.117 ± 5811.971    B/op
ServletWithNoopApiBenchmark.execute                                   134.359 ±    0.127   us/op
ServletWithNoopApiBenchmark.execute:execute·p0.50                     121.344              us/op
ServletWithNoopApiBenchmark.execute:·gc.alloc.rate.norm             27834.075 ± 6578.200    B/op
ServletWithOnePercentSamplingBenchmark.execute                        136.299 ±    0.130   us/op
ServletWithOnePercentSamplingBenchmark.execute:execute·p0.50          123.264              us/op
ServletWithOnePercentSamplingBenchmark.execute:·gc.alloc.rate.norm  28484.639 ± 6846.715    B/op
```

PR #3675, opened 2021-07-26
Very crude fix for publishing problem with agent-for-testing

iNikem:


buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
```diff
@@ -9,8 +9,10 @@ publishing {
       plugins.withId("java-platform") {
         from(components["javaPlatform"])
       }
-      plugins.withId("java-library") {
-        from(components["java"])
+      if(project.name != "agent-for-testing") {
```
mateuszrzeszutek, line 12: Instead of hard-coding the project name, would it make sense to do it similarly to `mavenGroupId`? You could add some `skipJavaComponentPublish` property and only use it in the `agent-for-testing` project.

mateuszrzeszutek (approved):

buildSrc/src/main/kotlin/otel.publish-conventions.gradle.kts
iNikem: I am not sure I want to add yet another configuration flag. Ultimate solution should be something different.

trask (approved):

PR #3604, opened 2021-07-16
Move instrumentation specific classes out of javaagent-instrumentatio…

mateuszrzeszutek:
…n-api

PR #3682, opened 2021-07-27
Refactor javaagent build

mateuszrzeszutek:
Depends on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3604

This PR removes shadowing from `:instrumentation` and `:javaagent-exporters` and moves all the build logic to the `:javaagent` project.

### DIFF
--- a/javaagent/build.gradle.kts
+++ b/javaagent/build.gradle.kts
@@ -11,12 +11,32 @@ plugins {
 }
 
 description = "OpenTelemetry Javaagent"
-
 group = "io.opentelemetry.javaagent"
 
+// this configuration collects libs that will be placed in the bootstrap classloader
 val bootstrapLibs by configurations.creating {
   isCanBeResolved = true
   isCanBeConsumed = false
+}
+// this configuration collects libs that will be placed in the agent classloader, isolated from the instrumented application code
+val javaagentLibs by configurations.creating {
+  isCanBeResolved = true
+  isCanBeConsumed = false
+}
+// this configuration collects just exporter libs (also placed in the agent classloader & isolated from the instrumented application)
+val exporterLibs by configurations.creating {
+  isCanBeResolved = true
+  isCanBeConsumed = false
+}
+
+// exclude dependencies that are to be placed in bootstrap from agent libs - they won't be added to inst/
+listOf(javaagentLibs, exporterLibs).forEach {
+  it.run {
+    exclude("org.slf4j")
+    exclude("io.opentelemetry", "opentelemetry-api")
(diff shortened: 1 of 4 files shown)


javaagent/build.gradle.kts
```diff
@@ -80,28 +100,42 @@ tasks {
...
+    archiveFileName.set("exporterLibs-relocated.jar")
   }
 
-  withType<ShadowJar>().configureEach {
+  //Includes instrumentations, but not exporters
+  val lightShadow by registering(ShadowJar::class) {
     configurations = listOf(bootstrapLibs)
 
+    dependsOn(relocateJavaagentLibs)
+    isolateClasses(relocateJavaagentLibs.get().outputs.files)
```
iNikem, line 135: Can we just pass configuration here? Without the intermediate shadow task.

mateuszrzeszutek: Unfortunately not: the intermediate shadow task does the relocation, if we were to pass the configuration here we'd get not relocated files (because that'd be a separate `CopySpec` that does not share the relocation settings of the main shadow jar task).

iNikem (approved):

testing/agent-for-testing/build.gradle.kts
```diff
@@ -59,11 +62,9 @@ val javaagentDependencies = dependencies
 project(":instrumentation").subprojects {
   val subProj = this
 
-  plugins.withId("java") {
-    if (subProj.name == "bootstrap") {
-      javaagentDependencies.run {
-        add(bootstrapLibs.name, project(subProj.path))
-      }
+  plugins.withId("otel.javaagent-bootstrap") {
```
trask, line 65: nice

testing/agent-for-testing/build.gradle.kts
```diff
@@ -35,6 +37,7 @@ dependencies {
   javaagentLibs(project(":testing:agent-exporter"))
   javaagentLibs(project(":javaagent-extension-api"))
   javaagentLibs(project(":javaagent-tooling"))
+  javaagentLibs(project(":muzzle"))
```
trask, line 40: just curious why this needed here now?

trask (approved):
nice 👍

anuraaga (approved):

testing/agent-for-testing/build.gradle.kts
mateuszrzeszutek: Hm, mostly for visibility/being super-explicit I guess. It's transitively included by `javaagent-tooling` so it would work even without that, but I wanted to explicitly include all our top-level agent modules just for readability I guess

Issue #1499, opened 2020-10-27
Mask DB statement in Apache Camel instrumentation


kubawach:
Once #1356 is merged, we need to review and mask any DB statement values logged.

PR #3683, opened 2021-07-27
Sanitize SQL in Apache Camel instrumentation

Enkelian:
Related to #1499
* DB statement span attributes generated by Apache Camel instrumentation are now sanitized using `SqlStatementSanitizer`
* Mongo statement attribute is no longer added by Camel instrumentation - sanitizing it here would be cumbersome (and mongo instrumentation should handle it anyway)
* Disclaimer: cql case will be handled in a separate PR as the way we're retrieving statements from URI in the current implementation needs to be properly tested


breedx-splk:
@Enkelian Thanks!  Looks like the build is failing due to checkstyle violations in the `DBSpanDecorator`.  

You can run these yourself with `./gradlew checkstyleMain` but if it's just a minor formatting thing you can often just quick fix it with `./gradlew spotlessApply`

Enkelian:
@breedx-splk
Turns out there's some inconsistency with `spotlessApply` and `checkstyleMain` when TODO is right above `case:` line. `spotlessApply` adds indents, which causes `checkstyleMain` to fail :thinking: 
Ended up just moving the line.

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/SanitizationTest.groovy
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.camel.Message
+import org.springframework.boot.SpringApplication
+import org.springframework.context.ConfigurableApplicationContext
+import spock.lang.Shared
+import spock.lang.Specification
+
+class SanitizationTest extends Specification {
+
+  @Shared
+  ConfigurableApplicationContext server
```
mateuszrzeszutek, line 18: Is this needed? You're not calling this app anywhere, you're using mocks in all test cases

mateuszrzeszutek:
> Turns out there's some inconsistency with `spotlessApply` and `checkstyleMain` when TODO is right above `case:` line. `spotlessApply` adds indents, which causes `checkstyleMain` to fail 🤔

🤯 
Can you create an issue (bug) for that? And add an example that makes those two plugins conflict with each other.

instrumentation/apache-camel-2.20/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/SanitizationTest.groovy
Enkelian: I was getting `java.lang.NoClassDefFoundError: io.opentelemetry.javaagent.instrumentation.apachecamel.decorators.DbSpanDecorator` without that :/

mateuszrzeszutek: Oh, I see - that's because this class is not normally available in `javaagent` module tests. We exclude instrumentation classes from the test classpath and the agent injects them when it instruments the code - that's why it was only working when you ran the app. When you removed it, the instrumentation wasn't applied and helper classes weren't injected.

Since this is a unit test, how about you place it in the camel's `javaagent-unit-tests` module? That one is an ordinary, "standard" unit test project that doesn't have any javaagent-specific weird behavior.

Enkelian: I just tried moving the test to `javaagent-unit-tests`, but I'm still getting `NoClassDefFoundError` for some reason

Enkelian: Works after adding `instrumentation-api` dependency, thanks!

instrumentation/apache-camel-2.20/javaagent-unit-tests/src/test/groovy/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/SanitizationTest.groovy
```diff
@@ -0,0 +1,58 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.apachecamel.decorators
+
+import org.apache.camel.Exchange
+import org.apache.camel.Message
+import spock.lang.Specification
+
+class SanitizationTest extends Specification {
```
mateuszrzeszutek, line 12: I've just noticed it now, but if you're using parameterized test case names you should add `@Unroll` on the class/test methods

Enkelian: Thanks, I'll add it right away 👍

mateuszrzeszutek (approved):

trask (approved):
thx!

PR #3685, opened 2021-07-27
Extract some muzzle checking logic from plugin util into standalone class

iNikem:
As it turned out that having `MuzzleGradlePluginUtil` as it is in gradle plugin's classloader does not actually work, I have started a series of PRs to extract all important parts from it into classes in `muzzle` module. In the end this will allow for easier bridge between plugin's and agent's classloaders.

Trying to avoid huge big-bang PR, thus splitting into smaller ones. Hoping for fast reviews and merges :D

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
@@ -59,27 +59,30 @@ public final class MuzzleGradlePluginUtil {
     ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
     Thread.currentThread().setContextClassLoader(agentClassLoader);
 
-    Map<String, List<Mismatch>> allMismatches = ClassLoaderMatcher.matchesAll(userClassLoader);
+    Map<String, List<Mismatch>> allMismatches;
+    try {
+      allMismatches = ClassLoaderMatcher.matchesAll(userClassLoader);
 
-    allMismatches.forEach(
-        (moduleName, mismatches) -> {
-          boolean passed = mismatches.isEmpty();
+      allMismatches.forEach(
+          (moduleName, mismatches) -> {
+            boolean passed = mismatches.isEmpty();
 
-          if (passed && !assertPass) {
-            System.err.println("MUZZLE PASSED " + moduleName + " BUT FAILURE WAS EXPECTED");
-            throw new IllegalStateException(
-                "Instrumentation unexpectedly passed Muzzle validation");
-          } else if (!passed && assertPass) {
-            System.err.println("FAILED MUZZLE VALIDATION: " + moduleName + " mismatches:");
+            if (passed && !assertPass) {
+              System.err.println("MUZZLE PASSED " + moduleName + " BUT FAILURE WAS EXPECTED");
+              throw new IllegalStateException(
+                  "Instrumentation unexpectedly passed Muzzle validation");
+            } else if (!passed && assertPass) {
+              System.err.println("FAILED MUZZLE VALIDATION: " + moduleName + " mismatches:");
 
-            for (Mismatch mismatch : mismatches) {
-              System.err.println("-- " + mismatch);
+              for (Mismatch mismatch : mismatches) {
+                System.err.println("-- " + mismatch);
+              }
+              throw new IllegalStateException("Instrumentation failed Muzzle validation");
             }
-            throw new IllegalStateException("Instrumentation failed Muzzle validation");
-          }
-        });
-
-    Thread.currentThread().setContextClassLoader(contextClassLoader);
+          });
+    } finally {
+      Thread.currentThread().setContextClassLoader(contextClassLoader);
+    }
 
     // run helper injector on all instrumentation modules
     if (assertPass) {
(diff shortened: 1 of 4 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -54,45 +55,32 @@
...
-          System.err.println("-- classloader mismatch");
-        }
 
-        for (Mismatch mismatch : mismatches) {
-          System.err.println("-- " + mismatch);
-        }
-        throw new IllegalStateException("Instrumentation failed Muzzle validation");
-      }
+    ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+    Thread.currentThread().setContextClassLoader(agentClassLoader);
```
trask, line 60: setting context class loader always feels mysterious, can you add a comment here explaining, or pass `agentClassLoader` to `matchesAll` and avoid?

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ClassLoaderMatcher.java
```diff
@@ -0,0 +1,63 @@
...
+              result.put(module.getClass().getName(), matches(module, classLoader));
+            });
+    return result;
+  }
+
+  /**
+   * Returns a list of {@link Mismatch}s between expectations of the given {@link
+   * InstrumentationModule} and what the given ClassLoader can provide.
+   */
+  public static List<Mismatch> matches(
```
trask, line 44:
```suggestion
  private static List<Mismatch> matches(
```

trask (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
iNikem: The main reason I decided not to pass `agentClassLoader` is the simplicity/clarity of the resulting method signature. I admit that relying on context class loader may be perceived as mysterious, but to me this is a "default" mode of operation. I would even prefer to change https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8c6adb481aa04862acdcf0a7b72bdc4d4d5060a3/buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts#L225 and to set `agentClassloader` already there instead of the bogus one.

trask: > I would even prefer to change
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8c6adb481aa04862acdcf0a7b72bdc4d4d5060a3/buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts#L225
> 
> and to set `agentClassloader` already there instead of the bogus one.

oh, I would like that too, does that not work for some reason? (I can try in a separate PR if you want to merge and move on)

iNikem: I have a couple of more PRs in the pipeline :) Will certainly try that.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -54,45 +55,32 @@
...
+            System.err.println("FAILED MUZZLE VALIDATION: " + moduleName + " mismatches:");
+
+            for (Mismatch mismatch : mismatches) {
+              System.err.println("-- " + mismatch);
+            }
+            throw new IllegalStateException("Instrumentation failed Muzzle validation");
+          }
+        });
+
+    Thread.currentThread().setContextClassLoader(contextClassLoader);
```
anuraaga, line 82: Can you put this in a finally block?

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ClassLoaderMatcher.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.ServiceLoader;
+
+/**
+ * This class verifies that a given {@link ClassLoader} satisfies all expectations of a given {@link
+ * InstrumentationModule}. It is used to make sure than module's transformations can be safely
```
anuraaga, line 16: make sure that a module's

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -304,14 +304,14 @@ private static void collectMethodsFromTypeHierarchy(
   }
 
   // optimization to avoid ArrayList allocation in the common case when there are no mismatches
-  private static List<Mismatch> lazyAdd(List<Mismatch> mismatches, Mismatch mismatch) {
+  static List<Mismatch> lazyAdd(List<Mismatch> mismatches, Mismatch mismatch) {
```
anuraaga, line 307: `add` for these methods that are not private any more is better name since lazy is an implementation detail.

anuraaga (approved):

PR #3690, opened 2021-07-27
Overhead external db

breedx-splk:
This builds on top of #3648 and will need to be rebased once that goes in.

This migrates the database out of the JVM and into an external postgresql. It seems unfair to have an in-memory database for benchmarks, because it's an uncommon scenario in production. The in-memory database would likely inflate GC pressure and allocation rates. This PR moves it to an external postgres container, which petclinic readily supports.


anuraaga:
@breedx-splk Did you happen to find any numbers for the change before and after in-memory? Gut feeling is that there would be more variance introduced from extra network hops, more so than in-memory db handling.

breedx-splk:
This got lost at sea. I will recreate.

PR #3691, opened 2021-07-27
Client span keys: suppressing same instrumentation

lmolkova:
This is a draft implementation of client span keys per instrumentations that suppresses spans with the same semantics from being created. 
E.g.
```
DB (fine)
  HTTP (fine)
    HTTP (suppressed)
       user custom (fine)
```    

- Supports HTTP, RPC, DB and Messaging

Features:
- when instrumentation type is set (affects client spans only), suppresses duplicate layers
- nested server/consumer spans are suppressed - no change here
- Generic instrumentation is never suppressed
- if Instrumenter is built with no type (configurable with system property), preserves current behavior - suppresses any nested spans. Going forward we can remove current behavior and switch to suppression based on types
- Instrumenter builder can figure out instrumentation type from attribute extractors (i.e. we can potentially make `IntrumentationType` package-private.

### DIFF
--- a/instrumentation-api/build.gradle.kts
+++ b/instrumentation-api/build.gradle.kts
@@ -1,6 +1,5 @@
 plugins {
   id("org.xbib.gradle.plugin.jflex")
-
   id("otel.java-conventions")
   id("otel.jacoco-conventions")
   id("otel.publish-conventions")
--- a/instrumentation-api/build.gradle.kts
+++ b/instrumentation-api/build.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
   id("org.xbib.gradle.plugin.jflex")
+
   id("otel.java-conventions")
   id("otel.jacoco-conventions")
   id("otel.publish-conventions")
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
@@ -99,7 +99,7 @@ public class Instrumenter<REQUEST, RESPONSE> {
    */
   public boolean shouldStart(Context parentContext, REQUEST request) {
     SpanKind spanKind = spanKindExtractor.extract(request);
-    boolean suppressed = spanSuppressionStrategy.shouldSuppress(spanKind, parentContext);
+    boolean suppressed = spanSuppressionStrategy.shouldSuppress(parentContext, spanKind);
 
     if (suppressed) {
       supportability.recordSuppressedSpan(spanKind, instrumentationName);
@@ -146,7 +146,7 @@ public class Instrumenter<REQUEST, RESPONSE> {
     Span span = spanBuilder.startSpan();
     context = context.with(span);
 
-    return spanSuppressionStrategy.storeInContext(spanKind, context, span);
+    return spanSuppressionStrategy.storeInContext(context, spanKind, span);
   }
 
   /**
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
@@ -15,9 +15,6 @@ import io.opentelemetry.api.trace.Tracer;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.InstrumentationVersion;
 import io.opentelemetry.instrumentation.api.internal.SupportabilityMetrics;
-import io.opentelemetry.instrumentation.api.tracer.ClientSpan;
-import io.opentelemetry.instrumentation.api.tracer.ConsumerSpan;
-import io.opentelemetry.instrumentation.api.tracer.ServerSpan;
 import java.util.ArrayList;
 import java.util.List;
 import org.checkerframework.checker.nullness.qual.Nullable;
@@ -76,6 +73,7 @@ public class Instrumenter<REQUEST, RESPONSE> {
   private final ErrorCauseExtractor errorCauseExtractor;
   @Nullable private final StartTimeExtractor<REQUEST> startTimeExtractor;
   @Nullable private final EndTimeExtractor<RESPONSE> endTimeExtractor;
+  private final SpanSuppressionStrategy spanSuppressionStrategy;
 
   Instrumenter(InstrumenterBuilder<REQUEST, RESPONSE> builder) {
     this.instrumentationName = builder.instrumentationName;
@@ -90,6 +88,7 @@ public class Instrumenter<REQUEST, RESPONSE> {
     this.errorCauseExtractor = builder.errorCauseExtractor;
     this.startTimeExtractor = builder.startTimeExtractor;
     this.endTimeExtractor = builder.endTimeExtractor;
+    this.spanSuppressionStrategy = builder.getSpanSuppressionStrategy();
   }
 
   /**
@@ -99,21 +98,9 @@ public class Instrumenter<REQUEST, RESPONSE> {
    * without calling those methods.
    */
   public boolean shouldStart(Context parentContext, REQUEST request) {
-    boolean suppressed = false;
     SpanKind spanKind = spanKindExtractor.extract(request);
-    switch (spanKind) {
-      case SERVER:
-        suppressed = ServerSpan.exists(parentContext);
-        break;
-      case CONSUMER:
-        suppressed = ConsumerSpan.exists(parentContext);
-        break;
-      case CLIENT:
-        suppressed = ClientSpan.exists(parentContext);
-        break;
-      default:
-        break;
-    }
+    boolean suppressed = spanSuppressionStrategy.shouldSuppress(spanKind, parentContext);
+
     if (suppressed) {
       supportability.recordSuppressedSpan(spanKind, instrumentationName);
     }
@@ -158,16 +145,8 @@ public class Instrumenter<REQUEST, RESPONSE> {
     spanBuilder.setAllAttributes(attributes);
     Span span = spanBuilder.startSpan();
     context = context.with(span);
-    switch (spanKind) {
-      case SERVER:
-        return ServerSpan.with(context, span);
-      case CLIENT:
-        return ClientSpan.with(context, span);
-      case CONSUMER:
-        return ConsumerSpan.with(context, span);
-      default:
-        return context;
-    }
+
+    return spanSuppressionStrategy.storeInContext(spanKind, context, span);
   }
 
   /**
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
@@ -32,7 +32,6 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  */
 public final class InstrumenterBuilder<REQUEST, RESPONSE> {
 
-  // TODO move this constant out of api and only use system property from javaagent?
   /** Instrumentation type suppression configuration property key. */
   private static final boolean ENABLE_SPAN_SUPPRESSION_BY_TYPE =
       Config.get()
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
@@ -5,19 +5,27 @@
 
 package io.opentelemetry.instrumentation.api.instrumenter;
 
+import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
 
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.metrics.GlobalMeterProvider;
 import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.context.propagation.TextMapGetter;
 import io.opentelemetry.context.propagation.TextMapSetter;
 import io.opentelemetry.instrumentation.api.annotations.UnstableApi;
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.instrumentation.api.instrumenter.db.DbAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.rpc.RpcAttributesExtractor;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
 /**
  * A builder of {@link Instrumenter}. Instrumentation libraries should generally expose their own
@@ -25,6 +33,12 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  * {@link Instrumenter}.
  */
 public final class InstrumenterBuilder<REQUEST, RESPONSE> {
+
+  // TODO move this constant out of api and only use system property from javaagent?
+  /** Instrumentation type suppression configuration property key. */
+  private static final boolean ENABLE_SPAN_SUPPRESSION_BY_TYPE =
+      Config.get()
+          .getBooleanProperty("otel.instrumentation.experimental.outgoing-span-suppression-by-type", false);
 
   final OpenTelemetry openTelemetry;
   final Meter meter;
@@ -42,6 +56,8 @@ public final class InstrumenterBuilder<REQUEST, RESPONSE> {
   ErrorCauseExtractor errorCauseExtractor = ErrorCauseExtractor.jdk();
   @Nullable StartTimeExtractor<REQUEST> startTimeExtractor = null;
   @Nullable EndTimeExtractor<RESPONSE> endTimeExtractor = null;
+
+  private boolean enableSpanSuppressionByType = ENABLE_SPAN_SUPPRESSION_BY_TYPE;
 
   InstrumenterBuilder(
       OpenTelemetry openTelemetry,
@@ -120,6 +136,38 @@ public final class InstrumenterBuilder<REQUEST, RESPONSE> {
     return this;
   }
 
+  // visible for tests
+  /**
+   * Enables suppression based on client instrumentation type.
+   *
+   * <p><strong>When enabled, suppresses nested spans depending on their {@link SpanKind} and
+   * type</strong>.
+   *
+   * <ul>
+   *   <li>CLIENT and PRODUCER nested spans are suppressed based on their type (HTTP, RPC, DB,
+   *       MESSAGING) i.e. if span with the same type is on the context, new span of this type will
+   *       not be started.
+   * </ul>
+   *
+   * <p><strong>When disabled:</strong>
+   *
+   * <ul>
+   *   <li>CLIENT and PRODUCER nested spans are always suppressed
+   * </ul>
+   *
+   * <p><strong>In both cases:</strong>
+   *
+   * <ul>
+   *   <li>SERVER and CONSUMER nested spans are always suppressed
+   *   <li>INTERNAL spans are never suppressed
+   * </ul>
+   */
+  InstrumenterBuilder<REQUEST, RESPONSE> enableInstrumentationTypeSuppression(
+      boolean enableInstrumentationType) {
+    this.enableSpanSuppressionByType = enableInstrumentationType;
+    return this;
+  }
+
   /**
    * Returns a new {@link Instrumenter} which will create client spans and inject context into
    * requests.
@@ -187,6 +235,34 @@ public final class InstrumenterBuilder<REQUEST, RESPONSE> {
       SpanKindExtractor<? super REQUEST> spanKindExtractor) {
     this.spanKindExtractor = spanKindExtractor;
     return constructor.create(this);
+  }
+
+  SpanSuppressionStrategy getSpanSuppressionStrategy() {
+    if (!enableSpanSuppressionByType) {
+      // if not enabled, preserve current behavior, not distinguishing types
+      return SpanSuppressionStrategy.from(singletonList(SpanKey.OUTGOING));
+    }
+
+    List<SpanKey> spanKeys = spanKeysFromAttributeExtractor(this.attributesExtractors);
+    return SpanSuppressionStrategy.from(spanKeys);
+  }
+
+  private static List<SpanKey> spanKeysFromAttributeExtractor(
+      List<? extends AttributesExtractor<?, ?>> attributesExtractors) {
+
+    List<SpanKey> spanKeys = new ArrayList<>();
+    for (AttributesExtractor<?, ?> attributeExtractor : attributesExtractors) {
+      if (attributeExtractor instanceof HttpAttributesExtractor) {
+        spanKeys.add(SpanKey.HTTP);
+      } else if (attributeExtractor instanceof RpcAttributesExtractor) {
+        spanKeys.add(SpanKey.RPC);
+      } else if (attributeExtractor instanceof DbAttributesExtractor) {
+        spanKeys.add(SpanKey.DB);
+      } else if (attributeExtractor instanceof MessagingAttributesExtractor) {
+        spanKeys.add(SpanKey.MESSAGING);
+      }
+    }
+    return spanKeys;
   }
 
   private interface InstrumenterConstructor<RQ, RS> {
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
@@ -22,7 +22,9 @@ import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttr
 import io.opentelemetry.instrumentation.api.instrumenter.rpc.RpcAttributesExtractor;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
@@ -241,14 +243,14 @@ public final class InstrumenterBuilder<REQUEST, RESPONSE> {
       return SpanSuppressionStrategy.SUPPRESS_ALL_NESTED_OUTGOING_STRATEGY;
     }
 
-    List<SpanKey> spanKeys = spanKeysFromAttributeExtractor(this.attributesExtractors);
+    Set<SpanKey> spanKeys = spanKeysFromAttributeExtractor(this.attributesExtractors);
     return SpanSuppressionStrategy.from(spanKeys);
   }
 
-  private static List<SpanKey> spanKeysFromAttributeExtractor(
+  private static Set<SpanKey> spanKeysFromAttributeExtractor(
       List<? extends AttributesExtractor<?, ?>> attributesExtractors) {
 
-    List<SpanKey> spanKeys = new ArrayList<>();
+    Set<SpanKey> spanKeys = new HashSet<>();
     for (AttributesExtractor<?, ?> attributeExtractor : attributesExtractors) {
       if (attributeExtractor instanceof HttpAttributesExtractor) {
         spanKeys.add(SpanKey.HTTP_CLIENT);
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.instrumentation.api.instrumenter;
 
-import static java.util.Collections.singletonList;
 import static java.util.Objects.requireNonNull;
 
 import io.opentelemetry.api.OpenTelemetry;
@@ -39,7 +38,6 @@ public final class InstrumenterBuilder<REQUEST, RESPONSE> {
       Config.get()
           .getBooleanProperty("otel.instrumentation.experimental.outgoing-span-suppression-by-type", false);
 
-  private static final SpanSuppressionStrategy ALL_CLIENT_SUPPRESSION_STRATEGY = SpanSuppressionStrategy.from(singletonList(SpanKey.OUTGOING));
   final OpenTelemetry openTelemetry;
   final Meter meter;
   final String instrumentationName;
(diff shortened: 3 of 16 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -58,13 +59,18 @@
...
+    return new InstrumenterBuilder<>(openTelemetry, instrumentationName, instrumentationType, spanNameExtractor);
   }
 
   private static final SupportabilityMetrics supportability = SupportabilityMetrics.instance();
 
   private final String instrumentationName;
+  // TODO we should be able to figure it out from the schema_url?
+  // but then we'd have multiple versions with different urls
+  // and we want to suppress multiple versions of the same instrumentation
+  private final InstrumentationType instrumentationType;
```
trask, line 73: I think schema_url is something like `https://opentelemetry.io/schemas/1.5.0` and doesn't include the semantic convention

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -58,13 +59,18 @@
   public static <REQUEST, RESPONSE> InstrumenterBuilder<REQUEST, RESPONSE> newBuilder(
       OpenTelemetry openTelemetry,
       String instrumentationName,
+      InstrumentationType instrumentationType,
       SpanNameExtractor<? super REQUEST> spanNameExtractor) {
-    return new InstrumenterBuilder<>(openTelemetry, instrumentationName, spanNameExtractor);
+    return new InstrumenterBuilder<>(openTelemetry, instrumentationName, instrumentationType, spanNameExtractor);
```
trask, line 64: Since `instrumentationType` is only for client spans(?), it would be nice to scope it to just client instrumenters. Let's see if @anuraaga and @mateuszrzeszutek have any thoughts how to do this.

trask (commented):
> `InstrumentationType` class is created to achieve extensibility and we may also use it to improve the registration of custom types. But I'm also happy to shrink it back to enum.

I sent a small PR to your PR, to simplify InstrumentationType a bit (short of making it an enum). I also like the idea of simplifying it all the way to an enum.

It would be good to have a configuration setting to suppress all nested CLIENT spans (preserving current behavior), since some backends don't handle nested CLIENT spans well yet.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
lmolkova: I recall we thought it would be useful to have a server span key as well, but as a way to enrich it rather than suppress it. Is there a need in java, e.g. in Spring to ever suppress server spans? I remember such cases in .NET world.  
Anyway, agree that beyond client spans it's not that important

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
lmolkova: I think http and db schemas may be versioned differently and in the same app you may have instrumentations with different versions (e.g. auto and native or manual instrumentations).
Also `schema_url` is an arbitrary url that you get `Tracer` with, so ideally it may be somehow associated with just this instrumentation type (e.g. `https://opentelemetry.io/schemas/http/1.5.0`). Anyway, that's not how it works now.

iNikem:
> It would be good to have a configuration setting to suppress all nested CLIENT spans (preserving current behavior), since some backends don't handle nested CLIENT spans well yet.

Also the opposite: not suppress anything.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
trask: oh that makes sense, we do suppress nested SERVER spans in Java instrumentation today. i c now, since `ClientSpan` creates it's own context keys, those wouldn't overlap with context keys created by `ServerSpan` 👍

trask:
> > It would be good to have a configuration setting to suppress all nested CLIENT spans (preserving current behavior), since some backends don't handle nested CLIENT spans well yet.
> 
> Also the opposite: not suppress anything.

In SIG last week I thought your use case was capturing HTTP spans under Database spans (and network spans under HTTP spans in the future). And that you didn't need to capture "duplicate" HTTP client spans under another HTTP client span. Though the logic seems nicely centralized with this approach and so adding such an option seems reasonable if someone really wants that capability.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/InstrumentationType.java
```diff
@@ -0,0 +1,31 @@
+package io.opentelemetry.instrumentation.api.tracer;
+
+public class InstrumentationType {
```
mateuszrzeszutek, line 3: It seems to me that `InstrumentationType` is a more generic version of `ServerSpan`/`ClientSpan` - perhaps we should get rid of those? Instead of having two mechanisms that work in a very similar way we'd have just one.

mateuszrzeszutek:
Hey @lmolkova ,
I tried to implement something that solves the problems I've found: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3734
Please take a look at it, maybe you'll find some part that'll be useful in your PR.

lmolkova:
Thanks everyone for comments and suggestions! 

@mateuszrzeszutek, I followed your idea : #3734 in this version, please take a look

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SuppressingSpanWrapper.java
```diff
@@ -0,0 +1,63 @@
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+interface SuppressingSpanWrapper {
```
lmolkova, line 8: this is what's left from `ClientSpan`, `ServerSpan` and `ConsumerSpan`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -189,6 +201,41 @@
...
 
+  InstrumentationType getInstrumentationType() {
+    if (instrumentationType != null) {
+      return instrumentationType;
+    }
+
+    return typeFromAttributeExtractor(this.attributesExtractors);
+  }
+
+  private static InstrumentationType typeFromAttributeExtractor(
```
lmolkova, line 212: I think we should either 
1. do this magic (and make `InstrumentationType` package-private) 
2. or let `InstrumentationType` be explicitly set (and remove the magic)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
```diff
@@ -81,6 +82,9 @@ protected BaseTracer(OpenTelemetry openTelemetry) {
    */
   protected abstract String getInstrumentationName();
 
+  protected InstrumentationType getInstrumentationType() {
+    return InstrumentationType.NONE;
```
lmolkova, line 86: we have to onboard Tracers to `InstrumentationType` to make `WithSpan` work well with instrumenter's suppression at least until `WithSpan` uses Tracers

lmolkova:
@trask @mateuszrzeszutek this one is ready for review, can you please take a look.

It needs #3771 to distinguish server/consumer and pass tests  (thanks @mateuszrzeszutek for creating it).

trask (commented):
I sent a PR to your PR, with an idea to rename/restructure InstrumentationType/SuppressingSpanWrapper into SpanKey and SpanSuppressionStrategy.

I think the explicit `ContextKey` constants in the `SpanKey` class will make bridging easier (between user/library instrumentation and javaagent instrumentation).

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -189,6 +227,37 @@
...
+        if (attributeExtractor instanceof HttpAttributesExtractor) {
+          return InstrumentationType.HTTP;
+        } else if (attributeExtractor instanceof RpcAttributesExtractor) {
+          return InstrumentationType.RPC;
+        } else if (attributeExtractor instanceof DbAttributesExtractor) {
+          return InstrumentationType.DB;
+        } else if (attributeExtractor instanceof MessagingAttributesExtractor) {
+          return InstrumentationType.MESSAGING;
+        }
+      }
```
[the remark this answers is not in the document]
trask: although I guess if a span implements both HTTP and RPC semantic conventions, then it should suppress both HTTP and RPC? I sent a second commit to my PR on your PR to see how that might look: https://github.com/lmolkova/opentelemetry-java-instrumentation/pull/2/commits/da56732284de6d2d6853c15eba11888645bf7c74

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
```diff
@@ -0,0 +1,65 @@
...
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+// TODO make class (and fields/methods below) package-private after tracers are gone and no need for
+//  bridging from ClientSpan/ServerSpan/ConsumerSpan
```
mateuszrzeszutek, line 14: I think something like `ServerSpan#fromContextOrNull()` will always be needed - there are many instrumentations that use it to update the server span name.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final ContextKey<Span> RPC_KEY =
+      ContextKey.named("opentelemetry-traces-span-key-rpc");
+  private static final ContextKey<Span> DB_KEY =
+      ContextKey.named("opentelemetry-traces-span-key-db");
+  private static final ContextKey<Span> MESSAGING_KEY =
+      ContextKey.named("opentelemetry-traces-span-key-messaging");
+
+  // this is used instead of above, depending on the configuration value for
+  // otel.instrumentation.experimental.span-suppression-by-type
+  // named outgoing because it covers both client and producer spans
```
mateuszrzeszutek, line 34: Previously we didn't suppress PRODUCER spans at all - should we maintain that behavior? Having a PRODUCER -> CLIENT relationship happens a few times in our messaging instrumentations (a producer calls a http/rpc client that actually sends the message to the broker)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final ContextKey<Span> OUTGOING_KEY =
+      ContextKey.named("opentelemetry-traces-span-key-client");
+
+  public static final SpanKey SERVER = new SpanKey(SERVER_KEY);
+  public static final SpanKey CONSUMER = new SpanKey(CONSUMER_KEY);
+
+  static final SpanKey HTTP = new SpanKey(HTTP_KEY);
+  static final SpanKey RPC = new SpanKey(RPC_KEY);
+  static final SpanKey DB = new SpanKey(DB_KEY);
+  static final SpanKey MESSAGING = new SpanKey(MESSAGING_KEY);
```
mateuszrzeszutek, line 44: We could name that one `PRODUCER` - the `CONSUMER` is also messaging related after all

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
```diff
@@ -0,0 +1,65 @@
...
+  // otel.instrumentation.experimental.span-suppression-by-type
+  // named outgoing because it covers both client and producer spans
+  private static final ContextKey<Span> OUTGOING_KEY =
+      ContextKey.named("opentelemetry-traces-span-key-client");
+
+  public static final SpanKey SERVER = new SpanKey(SERVER_KEY);
+  public static final SpanKey CONSUMER = new SpanKey(CONSUMER_KEY);
+
+  static final SpanKey HTTP = new SpanKey(HTTP_KEY);
+  static final SpanKey RPC = new SpanKey(RPC_KEY);
```
mateuszrzeszutek, line 42: Should we name those two `HTTP_CLIENT` and `RPC_CLIENT`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+          return consumerStrategy.storeInContext(spanKind, context, span);
+        case INTERNAL:
+          return context;
+      }
+      return context;
+    }
+
+    @Override
+    boolean shouldSuppress(SpanKind spanKind, Context parentContext) {
+      switch (spanKind) {
```
mateuszrzeszutek, line 110: I'd love to remove the association between span kind and span suppression at some point in the future - but I guess that's still impossible at this point. Maybe when we split our `AttributesExtractor` implementations into server/client? We've been thinking of doing that for HTTP, we could as well do the same thing for other conventions.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+        new SuppressIfSameType(clientSpanKeys),
+        SERVER_STRATEGY,
+        CONSUMER_STRATEGY);
+  }
+
+  abstract Context storeInContext(SpanKind spanKind, Context context, Span span);
+
+  abstract boolean shouldSuppress(SpanKind spanKind, Context parentContext);
+
+  static final class SuppressIfSameType extends SpanSuppressionStrategy {
```
mateuszrzeszutek, line 38: Nit: what do you think about `SuppressIfSameSpanKey`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    boolean shouldSuppress(SpanKind spanKind, Context parentContext) {
+        for (SpanKey outgoingSpanKey : outgoingSpanKeys) {
+          if (outgoingSpanKey.fromContextOrNull(parentContext) == null) {
+            return false;
+          }
+        }
+        return true;
```
mateuszrzeszutek, line 61: There's a bit too much indentation here 🦅 👁️ 

```suggestion
      for (SpanKey outgoingSpanKey : outgoingSpanKeys) {
        if (outgoingSpanKey.fromContextOrNull(parentContext) == null) {
          return false;
        }
      }
      return true;
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+import io.opentelemetry.context.Context;
+import java.util.Arrays;
+import java.util.List;
+
+abstract class SpanSuppressionStrategy {
+
+  private static final SpanSuppressionStrategy SERVER_STRATEGY = new SuppressIfSameType(
+      Arrays.asList(SpanKey.SERVER));
+  private static final SpanSuppressionStrategy CONSUMER_STRATEGY = new SuppressIfSameType(
+      Arrays.asList(SpanKey.CONSUMER));
```
mateuszrzeszutek, line 19: Nit

```suggestion
      Collections.singletonList(SpanKey.SERVER));
  private static final SpanSuppressionStrategy CONSUMER_STRATEGY = new SuppressIfSameType(
      Collections.singletonList(SpanKey.CONSUMER));
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -189,6 +240,34 @@
     return constructor.create(this);
   }
 
+  SpanSuppressionStrategy getSpanSuppressionStrategy() {
+    if (!enableSpanSuppressionByType) {
+      // if not enabled, preserve current behavior, not distinguishing types
+      return SpanSuppressionStrategy.from(singletonList(SpanKey.OUTGOING));
```
mateuszrzeszutek, line 246: WDYT about making that a constant in `SpanSuppressionStrategy`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+
+  static SpanSuppressionStrategy from(List<SpanKey> clientSpanKeys) {
+    if (clientSpanKeys.isEmpty()) {
+      return new CompositeStrategy(
+          NeverSuppress.INSTANCE,
+          SERVER_STRATEGY,
+          CONSUMER_STRATEGY);
```
mateuszrzeszutek, line 26: Nit: that could be a constant too

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+    return new CompositeStrategy(
+        new SuppressIfSameType(clientSpanKeys),
+        SERVER_STRATEGY,
+        CONSUMER_STRATEGY);
+  }
+
+  abstract Context storeInContext(SpanKind spanKind, Context context, Span span);
+
+  abstract boolean shouldSuppress(SpanKind spanKind, Context parentContext);
```
mateuszrzeszutek, line 36: Nit: we usually keep `Context` as the first parameter in all instrumentation-api methods (or at the very least we aim to do that)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,124 @@
...
+      return false;
+    }
+  }
+
+  static final class CompositeStrategy extends SpanSuppressionStrategy {
+    private final SpanSuppressionStrategy clientStrategy;
+    private final SpanSuppressionStrategy serverStrategy;
+    private final SpanSuppressionStrategy consumerStrategy;
+
+    public CompositeStrategy(SpanSuppressionStrategy client, SpanSuppressionStrategy server,
```
mateuszrzeszutek, line 85:
```suggestion
   CompositeStrategy(SpanSuppressionStrategy client, SpanSuppressionStrategy server,
```

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategyTest.java
```diff
@@ -0,0 +1,185 @@
...
+    allClientSpanKeys()
+        .forEach(spanKey -> assertThat(spanKey.fromContextOrNull(context)).isNull());
+
+    assertThat(SpanKey.SERVER.fromContextOrNull(context)).isNull();
+  }
+
+  @ParameterizedTest
+  @MethodSource("allClientSpanKeys")
+  public void clientSpan_differentForAllTypes(SpanKey spanKey) {
+    SpanSuppressionStrategy strategy = SpanSuppressionStrategy.from(Arrays.asList(spanKey));
```
mateuszrzeszutek, line 81: Nit

```suggestion
    SpanSuppressionStrategy strategy = SpanSuppressionStrategy.from(Collections.singletonList(spanKey));
```

mateuszrzeszutek:
Sorry for being a bit late, but between some additional surprise work and GH issues I didn't manage to look at it yesterday 🙇‍♂️ 

Can you rebase your branch with `main`? My spring-integration PR has been merged recently, so rebasing should make your PR a bit easier to review.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
lmolkova: this is the same for HTTP and RPC - they can be both server and client, but those are client only. May be name should reflect that? `HTTP_CLIENT`, `MESSAGING_PRODUCER`?

trask: this makes sense to me 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
lmolkova: I guess we never suppressed them because there as never 2 producer spans so far and if there were, we'd want to suppress nested ones? With suppression-by-type we'd not suppress HTTP client under messaging producer because types are different. You're right that without it we'd start suppressing them.

BTW, it seems messaging producer [can sometimes be client](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md), not sure if we follow it though.

> A producer of a message should set the span kind to PRODUCER unless it synchronously waits for a response: then it should use CLIENT. 

Anyway, I've separated producer strategy from client and we'd suppress nested producers, but never suppress CLIENT under PRODUCER.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
trask: this makes sense, we can remove the comment (it was mine)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
```diff
@@ -0,0 +1,162 @@
...
+import static java.util.Collections.singletonList;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import java.util.List;
+
+abstract class SpanSuppressionStrategy {
+  private static final SpanSuppressionStrategy SERVER_STRATEGY =
+      new SuppressIfSameSpanKey(singletonList(SpanKey.SERVER));
```
trask, line 17: Intellij is giving me some weird warnings here that I haven't seen before:

> Referencing subclass SuppressIfSameSpanKey from superclass SpanKeyStrategy initializer might lead to class loading deadlock

probably worth splitting the nested classes out to top-level classes

trask (approved):
thx @lmolkova!!

I have some vague thoughts around naming changes, but will keep thinking and possibly propose some naming changes in a follow-up PR, e.g.
* calling the options "single"/"layered"/"all" (I'm hoping no one really wants the "all", but we'll see...)
* the SpanSuppressionStrategy class is more than just about span suppression, it's also about populating span keys for later enrichment, but I'm not sure how to convey both of those purposes. one possible answer to the naming problem is to split SpanSuppressionStratety into two classes, one that handles suppression and one that handles span key management.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
mateuszrzeszutek: > I guess we never suppressed them because there as never 2 producer spans so far and if there were, we'd want to suppress nested ones?

I that case we probably would want to suppress them. 

> BTW, it seems messaging producer [can sometimes be client](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md), not sure if we follow it though.

🤯 
I don't think that we follow it; we always use `PRODUCER` for all outgoing messaging spans. (Anyway, that's another reason why not using the span kind when suppressing might be a good idea)

mateuszrzeszutek:
> the SpanSuppressionStrategy class is more than just about span suppression, it's also about populating span keys for later enrichment, but I'm not sure how to convey both of those purposes. one possible answer to the naming problem is to split SpanSuppressionStratety into two classes, one that handles suppression and one that handles span key management.

We pretty much only use `ServerSpan` and it's just for updating the span name. I don't have any strong opinion on this, but I keep wondering whether we should introduce the `LocalRootSpan` after all - not for suppression, just for keeping it in context so that it's easy to retrieve it (and rename it).

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanSuppressionStrategy.java
lmolkova: it seems we'd have to specialize instrumenters (so specific instumenter instance work for one span kind) to remove the association between kind/suppression. Or perhaps, if we move `CompositeStrategy` into the instrumenter and it would have 4 of them - one per kind, it would not remove the association but maybe a bit cleaner.
Either way, a combination of type + kind is important for suppression (i.e. server HTTP span should not suppress client HTTP span) 

Another thought is that all strategies can be the same regardless of kind - suppress same type+kind (or suppress the same kind if type disabled) removes the association. I assume it is rare to have nested SERVERs of different types and we can easily allow that.

instrumentation-api/build.gradle.kts
```diff
@@ -1,6 +1,5 @@
 plugins {
   id("org.xbib.gradle.plugin.jflex")
-
```
anuraaga, line 3: Can you revert this change? We don't have a strict policy but generally group built-in Gradle plugins, third party plugins, and our own plugins

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -26,6 +32,12 @@
  */
 public final class InstrumenterBuilder<REQUEST, RESPONSE> {
 
+  /** Instrumentation type suppression configuration property key. */
+  private static final boolean ENABLE_SPAN_SUPPRESSION_BY_TYPE =
+      Config.get()
+          .getBooleanProperty(
+              "otel.instrumentation.experimental.outgoing-span-suppression-by-type", false);
```
anuraaga, line 39: Since this can be overridden through direct call to `enableInstrumentationTypeSuppression` I think we should use `default-` in the property name. Or maybe can remove that method

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -189,6 +235,34 @@
...
+    List<SpanKey> spanKeys = spanKeysFromAttributeExtractor(this.attributesExtractors);
+    return SpanSuppressionStrategy.from(spanKeys);
+  }
+
+  private static List<SpanKey> spanKeysFromAttributeExtractor(
+      List<? extends AttributesExtractor<?, ?>> attributesExtractors) {
+
+    List<SpanKey> spanKeys = new ArrayList<>();
+    for (AttributesExtractor<?, ?> attributeExtractor : attributesExtractors) {
+      if (attributeExtractor instanceof HttpAttributesExtractor) {
```
anuraaga, line 253: Does this order matter? If so, let's add a comment to make sure that's not forgotten. If not, consider using `Set` instead of `List` to codify that fact - in fact while we don't expect it in practice, we may have multiple extractors of the same type registered so that would also get deduped.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/NeverSuppressAndStoreStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import java.util.List;
+
+final class NeverSuppressAndStoreStrategy extends SpanSuppressionStrategy {
```
anuraaga, line 13: Nit: I read this with `Never` applying to `Store` too. `StoreOnly` would make sense to me

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/NeverSuppressOrStoreStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+
+final class NeverSuppressOrStoreStrategy extends SpanSuppressionStrategy {
```
anuraaga, line 12: With the above naming suggestion, `NoopSpanSuppresionStrategy` seems ok for this one.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
```diff
@@ -0,0 +1,66 @@
...
+  public static final SpanKey ALL_CLIENTS = new SpanKey(CLIENT_KEY);
+  public static final SpanKey ALL_PRODUCERS = new SpanKey(PRODUCER_KEY);
+
+  private final ContextKey<Span> key;
+
+  SpanKey(ContextKey<Span> key) {
+    this.key = key;
+  }
+
+  public Context with(Context context, Span span) {
```
anuraaga, line 58: I know this pattern is the same as what we have so it's ok, but consider a different name like `setIn`.

anuraaga (approved):
Just some small points, thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
trask: removing `enableInstrumentationTypeSuppression` method sounds good 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java
trask: this was me aligning the name with what we have already, @lmolkova originally had `storeInContext`, I'm good with changing it

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
lmolkova: removing this method makes suppression untestable through instrumetner API. The method is package-private, though, so instrumentations won't be able to call it.

trask:
🥳

PR #3692, opened 2021-07-27
Fix NPE in Apache HttpAsyncClient instrumentation

trask:
Resolves #3659

Missing test coverage added (see ApacheClientHostRequest).

PR #3694, opened 2021-07-28
Fix Apache HttpClient host + absolute uri

trask:
Noticed this gap while fixing #3692


anuraaga (commented):
Do we need for 4.3 also?

trask:
> Do we need for 4.3 also?

What is this 4.3? 😉

anuraaga:
> What is this 4.3? 😉

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientRequest.java#L30

I took a slightly different approach using instanceof, but since the tests are the same may easily have the same bug

instrumentation/apache-httpclient/apache-httpclient-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v4_0/ApacheHttpClientRequest.java
```diff
@@ -121,4 +121,29 @@ public Integer getPeerPort() {
         return null;
     }
   }
+
+  @Nullable
+  private static URI getCalculatedUri(HttpHost httpHost, HttpRequest httpRequest) {
+    final URI uri;
+    try {
+      // this can be relative or absolute
+      uri = new URI(httpRequest.getRequestLine().getUri());
```
anuraaga, line 130: What do you think about recognizing a scheme (`contains("://")`) to avoid parsing (and losing fidelity?) an absolute URI.

By the way I had noticed randomly the issue about concatenating two absolute URIs together and the instanceof check I added fixed that, is there another case where we can have two absolute URIs? I guess we can document it in #3697 where I guess we've found ourselves eating some Apache-flavored spaghetti

anuraaga: By the way here's what Brave does, which of course may also be missing some corner cases, but for reference it seems a bit different than either of our code branches actually

https://github.com/openzipkin/brave/blob/master/instrumentation/httpclient/src/main/java/brave/httpclient/TracingProtocolExec.java#L102

trask: spaghetti overload 😋

> What do you think about recognizing a scheme (`contains("://")`) to avoid parsing (and losing fidelity?) an absolute URI.

Yeah, I was just noticing our (unnecessary) URL constructions while profiling and thinking to add some type of "url builder" to our instrumentation api to optimize. Will add tracking issue.

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientRequest.java
```diff
@@ -132,4 +125,34 @@ Integer getPeerPort() {
...
+      logger.debug(e.getMessage(), e);
+      return null;
+    }
+  }
+
+  @Nullable
+  private static URI getCalculatedUri(HttpHost httpHost, URI uri) {
+    try {
+      System.out.println("host: " + httpHost);
+      System.out.println("uri: " + uri);
```
mateuszrzeszutek, line 144: To be removed? 😛 

```suggestion
```

instrumentation/apache-httpclient/apache-httpclient-4.3/testing/src/main/groovy/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTest.groovy
```diff
@@ -96,11 +96,18 @@ abstract class ApacheHttpClientTest<T extends HttpRequest> extends HttpClientTes
     }
     return builder.toString()
   }
+
+  static String absoluteUriWithNonResolvingHost(URI uri) {
```
mateuszrzeszutek, line 100: Is this method used anywhere?

trask:
@open-telemetry/java-instrumentation-approvers apologies, i got mixed up a bit between branches and apache httpclient spaghetti, and a couple of changes from this PR leaked into #3692, which is why a couple diffs in this PR are a bit weird

anuraaga (approved):

mateuszrzeszutek (approved):

Issue #3673, opened 2021-07-26
NullPointerException when running grails tests


trask:
The tests pass, but log this:

```
[otel.javaagent 2021-07-25 22:02:15:132 -0700] [http-nio-11001-exec-2] DEBUG io.opentelemetry.javaagent.bootstrap.ExceptionLogger - Failed to handle exception in instrumentation for org.apache.catalina.connector.CoyoteAdapter on jdk.internal.loader.ClassLoaders$AppClassLoader@4cc77c2e
java.lang.NullPointerException
	at org.apache.catalina.connector.Request.notifyAttributeAssigned(Request.java:1498)
	at org.apache.catalina.connector.Request.setAttribute(Request.java:1488)
	at io.opentelemetry.javaagent.shaded.instrumentation.servlet.javax.JavaxServletAccessor.setRequestAttribute(JavaxServletAccessor.java:51)
	at io.opentelemetry.javaagent.shaded.instrumentation.servlet.javax.JavaxServletAccessor.setRequestAttribute(JavaxServletAccessor.java:13)
	at io.opentelemetry.javaagent.shaded.instrumentation.servlet.ServletHttpServerTracer.setAsyncListenerResponse(ServletHttpServerTracer.java:153)
	at io.opentelemetry.javaagent.instrumentation.tomcat.common.TomcatServerHandlerAdviceHelper.attachResponseToRequest(TomcatServerHandlerAdviceHelper.java:81)
	at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:476)
	at org.apache.coyote.http11.AbstractHttp11Processor.process(AbstractHttp11Processor.java:1091)
	at org.apache.coyote.AbstractProtocol$AbstractConnectionHandler.process(AbstractProtocol.java:668)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1521)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.run(NioEndpoint.java:1478)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
[...]

PR #3705, opened 2021-07-28
Fix NullPointerException on tomcat

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3673
We store response in a request attribute and use it later to add async completion listener. Currently on tomcat we touch request a bit too early when it ins't fully initialized. Either `org.apache.coyote.Request` doesn't have a reference to `HttpServletRequest` in which case we just don't link response to request or `org.apache.coyote.Request` has reference to `HttpServletRequest` which is empty (request was recycled, reusing a previous request) in which case depending on tomcat version we could get a `NullPointerException`.

### DIFF
--- a/instrumentation/servlet/servlet-3.0/library/src/main/java/io/opentelemetry/instrumentation/servlet/v3_0/Servlet3Accessor.java
+++ b/instrumentation/servlet/servlet-3.0/library/src/main/java/io/opentelemetry/instrumentation/servlet/v3_0/Servlet3Accessor.java
@@ -31,8 +31,6 @@ public class Servlet3Accessor extends JavaxServletAccessor<HttpServletResponse> 
       request
           .getAsyncContext()
           .addListener(new Listener(listener), request, (HttpServletResponse) response);
-    } else {
-      request.getAsyncContext().addListener(new Listener(listener));
     }
   }
 
--- a/instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatServerHandlerInstrumentation.java
+++ b/instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatServerHandlerInstrumentation.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.javaagent.instrumentation.tomcat.common;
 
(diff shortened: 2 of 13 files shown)


instrumentation/servlet/servlet-3.0/library/src/main/java/io/opentelemetry/instrumentation/servlet/v3_0/Servlet3Accessor.java
```diff
@@ -31,8 +31,6 @@ public void addRequestAsyncListener(
       request
           .getAsyncContext()
           .addListener(new Listener(listener), request, (HttpServletResponse) response);
-    } else {
```
laurit, line 34: removed this branch to make tests fail when `response` is null

instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatServerHandlerInstrumentation.java
```diff
@@ -36,6 +40,14 @@ public void transform(TypeTransformer transformer) {
             .and(named("service"))
             .and(takesArgument(0, named("org.apache.coyote.Request")))
             .and(takesArgument(1, named("org.apache.coyote.Response"))),
-        adviceClassName);
+        handlerAdviceClassName);
+
+    transformer.applyAdviceToMethod(
+        isMethod()
+            .and(named("postParseRequest"))
```
mateuszrzeszutek, line 47: `Adapter` doesn't have this method. Is there any reason why we're not instrumenting `CoyoteAdapter`?

mateuszrzeszutek (approved):

instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatServerHandlerInstrumentation.java
laurit: Thanks. Didn't even realize that it wasn't using `CoyoteAdapter`.

trask (approved):
thx

Issue #3319, opened 2021-06-15
Create spans on new Netty connection


iNikem:
Netty instrumentation `ChannelFutureListenerInstrumentation` creates a span with special name `CONNECT` when connection failed to open. As creating new connection may be potentially an expensive operation there is a client request to capture all cases of creating new connection as span. Ideally with correct, non-zero, duration. This also will help exposing problems when connection pool does not work as expected.

PR #3707, opened 2021-07-28
Add option to create span on new netty connection

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3319
Currently we create a zero length span when netty fails to connect and not inside a client span. This pr adds an option to capture a span around netty establishing connection.

### DIFF
--- a/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
+++ b/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
@@ -6,7 +6,6 @@
 package io.opentelemetry.javaagent.instrumentation.netty.v3_8.client;
 
 import static io.opentelemetry.api.trace.SpanKind.CLIENT;
-import static io.opentelemetry.api.trace.SpanKind.INTERNAL;
 import static io.opentelemetry.javaagent.instrumentation.netty.v3_8.client.NettyResponseInjectAdapter.SETTER;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NetTransportValues.IP_TCP;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NetTransportValues.IP_UDP;
@@ -53,7 +52,7 @@ public class NettyHttpClientTracer
   }
 
   public void connectionFailure(Context parentContext, Channel channel, Throwable throwable) {
-    SpanBuilder spanBuilder = spanBuilder(parentContext, "CONNECT", INTERNAL);
+    SpanBuilder spanBuilder = spanBuilder(parentContext, "CONNECT", CLIENT);
     spanBuilder.setAttribute(
         SemanticAttributes.NET_TRANSPORT, channel instanceof DatagramChannel ? IP_UDP : IP_TCP);
     NetPeerAttributes.INSTANCE.setNetPeer(
--- a/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
+++ b/instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.javaagent.instrumentation.netty.v3_8.client;
 
 import static io.opentelemetry.api.trace.SpanKind.CLIENT;
+import static io.opentelemetry.api.trace.SpanKind.INTERNAL;
 import static io.opentelemetry.javaagent.instrumentation.netty.v3_8.client.NettyResponseInjectAdapter.SETTER;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NetTransportValues.IP_TCP;
 import static io.opentelemetry.semconv.trace.attributes.SemanticAttributes.NetTransportValues.IP_UDP;
@@ -52,7 +53,7 @@ public class NettyHttpClientTracer
   }
 
   public void connectionFailure(Context parentContext, Channel channel, Throwable throwable) {
-    SpanBuilder spanBuilder = spanBuilder(parentContext, "CONNECT", CLIENT);
+    SpanBuilder spanBuilder = spanBuilder(parentContext, "CONNECT", INTERNAL);
     spanBuilder.setAttribute(
         SemanticAttributes.NET_TRANSPORT, channel instanceof DatagramChannel ? IP_UDP : IP_TCP);
     NetPeerAttributes.INSTANCE.setNetPeer(
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
@@ -107,8 +107,11 @@ public abstract class AbstractNettyHttpClientTracer<REQUEST extends AbstractNett
       Channel channel,
       Throwable throwable) {
     if (context != null) {
+      // if context is present we started span in startConnectionSpan
       endConnectionSpan(context, channel, throwable);
     } else if (throwable != null && shouldStartSpan(parentContext, CLIENT)) {
+      // if we didn't start span in startConnectionSpan create a span only when the request fails
+      // and when not inside a client span
       connectionFailure(parentContext, remoteAddress, channel, throwable);
     }
   }
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
@@ -106,9 +106,11 @@ public abstract class AbstractNettyHttpClientTracer<REQUEST extends AbstractNett
       SocketAddress remoteAddress,
       Channel channel,
       Throwable throwable) {
-    if (context != null) {
-      // if context is present we started span in startConnectionSpan
-      endConnectionSpan(context, channel, throwable);
+    if (alwaysCreateConnectSpan) {
+      if (context != null) {
+        // if context is present we started span in startConnectionSpan
+        endConnectionSpan(context, channel, throwable);
+      }
     } else if (throwable != null && shouldStartSpan(parentContext, CLIENT)) {
       // if we didn't start span in startConnectionSpan create a span only when the request fails
       // and when not inside a client span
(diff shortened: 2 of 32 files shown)


instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
```diff
@@ -84,39 +89,58 @@ protected String responseHeader(HttpResponse httpResponse, String name) {
...
+  public void endConnectionSpan(
+      Context context,
+      Context parentContext,
+      SocketAddress remoteAddress,
+      Channel channel,
+      Throwable throwable) {
+    if (context != null) {
+      // if context is present we started span in startConnectionSpan
+      endConnectionSpan(context, channel, throwable);
+    } else if (throwable != null && shouldStartSpan(parentContext, CLIENT)) {
```
iNikem, line 112: As `connectionFailure` creates INTERNAL span, I don't think we need `shouldStartSpan` here?

instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/NettyChannelPipelineInstrumentation.java
```diff
@@ -103,15 +96,4 @@ public static void addHandler(
...
     }
   }
-
-  @SuppressWarnings("unused")
-  public static class ChannelPipelineConnectAdvice {
-
-    @Advice.OnMethodEnter
-    public static void addParentSpan(@Advice.This ChannelPipeline pipeline) {
-      Context context = Java8BytecodeBridge.currentContext();
-      Attribute<Context> attribute = pipeline.channel().attr(AttributeKeys.CONNECT_CONTEXT);
```
iNikem, line 113: Why exactly this context propagation is not needed anymore?

iNikem (commented):
Do I understand correctly:

- If `always-create-connect-span=true` then we create span both for successful and failed connections and their durations are "true", not 0ms?
- if `always-create-connect-span=false` we still create span for failed connection with 0 duration?

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
laurit: In case of a http client based on netty that has custom instrumentation that captures both connection creation and actual request there is no need need to create the connection span just for failure, the same information will be recorded under parent span.

instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/NettyChannelPipelineInstrumentation.java
laurit: Previously connection failure was detected when `org.jboss.netty.channel.ChannelFutureListener#operationComplete` was called with a throwable (yes this really is rather arbitrary). As this method was called from a netty thread there was a need to remember the initial context used when establishing connection. Now connection span is started in `Bootstrap. doResolveAndConnect` and ended in the same method or in `ConnectionCompleteListener`, context is passed directly to that listener.

laurit:
@iNikem yes that is correct.

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
```diff
@@ -52,7 +53,7 @@ public Context startSpan(Context parentContext, ChannelHandlerContext ctx, HttpR
   }
 
   public void connectionFailure(Context parentContext, Channel channel, Throwable throwable) {
-    SpanBuilder spanBuilder = spanBuilder(parentContext, "CONNECT", CLIENT);
+    SpanBuilder spanBuilder = spanBuilder(parentContext, "CONNECT", INTERNAL);
```
trask, line 56: why INTERNAL over CLIENT?

laurit: When we always create connection span it could be nested under a client span so using `INTERNAL`. When creating connection span only for failure it could be kept as `CLIENT` though as the connection failed it shouldn't have a child `SERVER` span so changed this also to `INTERNAL`.

iNikem (approved):

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
```diff
@@ -84,39 +89,58 @@ protected String responseHeader(HttpResponse httpResponse, String name) {
...
 
-    if (tracer.shouldStartSpan(parentContext, SpanKind.CLIENT)) {
-      tracer.connectionFailure(parentContext, channelFuture.channel(), cause);
+  public void endConnectionSpan(
+      Context context,
+      Context parentContext,
+      SocketAddress remoteAddress,
+      Channel channel,
+      Throwable throwable) {
+    if (context != null) {
```
mateuszrzeszutek, line 109: Nit: using `alwaysCreateConnectSpan` will probably allow JIT to skip the `else` part easily when that property is true

instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
```diff
@@ -25,6 +32,9 @@
   public void transform(TypeTransformer transformer) {
     transformer.applyAdviceToMethod(
         isConstructor(), BootstrapInstrumentation.class.getName() + "$InitAdvice");
+    transformer.applyAdviceToMethod(
+        named("doConnect").and(takesArgument(0, named("java.net.SocketAddress"))),
```
mateuszrzeszutek, line 36: Nit:
```suggestion
        named("doConnect").and(takesArgument(0, SocketAddress.class)),
```

instrumentation/netty/netty-4.0/javaagent/src/test/groovy/Netty40ConnectionSpanTest.groovy
```diff
@@ -0,0 +1,164 @@
...
+  @Shared
+  private Bootstrap bootstrap = buildBootstrap()
+
+  def setupSpec() {
+    server = new HttpClientTestServer(openTelemetry)
+    server.start()
+  }
+
+  def cleanupSpec() {
+    eventLoopGroup?.shutdownGracefully()
```
mateuszrzeszutek, line 51: Is it possible for `eventLoopGroup` to be null? Is that `?` needed?

mateuszrzeszutek (approved):

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
iNikem: As we currently suppress nested CLIENT spans, we cannot have these new spans as CLIENT. Because they will be suppressed almost always. So we should have them INTERNAL for now.

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/AbstractNettyHttpClientTracer.java
laurit: Added `alwaysCreateConnectSpan` check, though I suspect that it won't change much for jit as I believe it speculatively removes branches that have not executed anyway.

instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
laurit: thanks, fixed.

instrumentation/netty/netty-4.0/javaagent/src/test/groovy/Netty40ConnectionSpanTest.groovy
laurit: Removed it, I guess I have developed a habit of adding too many of these to `cleanupSpec` as npe in `cleanupSpec` is fairly common when something fails during initialization

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/client/NettyHttpClientTracer.java
trask: I'd prefer to keep this one here as CLIENT, and the other one can be INTERNAL for now until we support nested CLIENT spans in #3691.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind:

> CLIENT - Indicates that the span describes a request to some remote service
> INTERNAL - Indicates that the span represents an internal operation within an application

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/v1_0/ReactorNettyTracer.java
```diff
@@ -0,0 +1,105 @@
...
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.instrumentation.api.tracer.BaseTracer;
+import io.opentelemetry.instrumentation.api.tracer.net.NetPeerAttributes;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+
+public class ReactorNettyTracer extends BaseTracer {
```
trask, line 26: can you change this to Instrumenter, or add it to #2713?

trask (approved):

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactornetty/v1_0/ReactorNettyTracer.java
laurit: Added to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/2713

PR #3708, opened 2021-07-28
Muzzle classloading hell

iNikem:
Based on and continuation of #3688

Running muzzle check from the correct classloader.


mateuszrzeszutek (approved):

gradle-plugins/src/main/kotlin/io/opentelemetry/javaagent/muzzle/BogusClassLoader.kt
```diff
@@ -1,14 +0,0 @@
...
-
-package io.opentelemetry.javaagent.muzzle
-
-import java.security.SecureClassLoader
-
-internal class BogusClassLoader : SecureClassLoader() {
-  override fun toString(): String {
-    return "bogus"
-  }
-}
```
trask, line 14: 🎉

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -12,18 +12,32 @@
...
 import java.util.Map;
 import java.util.ServiceLoader;
-import net.bytebuddy.dynamic.ClassFileLocator;
 
-/** Entry point for the muzzle gradle plugin. */
+/**
+ * Entry point for the muzzle gradle plugin.
+ * <p>
+ * In order to understand this class and its weirdness, one has to remember that there are three
+ * different independent class loders at play here.
```
trask, line 25:
```suggestion
 * different independent class loaders at play here.
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -12,18 +12,32 @@
...
+ * <p>
+ * First, Gradle classloader that has loaded the muzzle-check plugin that calls this class. This
```
trask, line 27:
```suggestion
 * First, Gradle class loader that has loaded the muzzle-check plugin that calls this class. This
```

gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle/matcher/MuzzleGradlePluginUtil.java
```diff
@@ -12,18 +12,32 @@
...
+ * one has a lot of Gradle specific stuff and we don't want that it to be available during muzzle checks.
+ * <p>
+ * Second, there is agent or instrumentation class loader, which contains all InstrumentationModules
+ * and helper classes. The actual muzzle check process happens "inside" that classloader. This means
```
trask, line 31:
```suggestion
 * and helper classes. The actual muzzle check process happens "inside" that class loader. This means
```

trask (approved):
nice

PR #3714, opened 2021-07-29
Add HttpClientTestOptions

anuraaga:
We currently have many protected methods to tweak test behavior, but it can be hard to understand what options we have or what the defaults are. I think having all options grouped makes it easier to understand how to configure the test - a big reason for this is to simplify the spock bridge.

I just copied the options as is for now, but I think we can iterate to improve more - I find the knobs now to be confusing on what they control, some are semantic and control multiple tests, many just control a single test, some have different names then the test itself, etc. I also really don't like how many subclasses use the URL 192.0.2.3 as a proxy for what test they are running on to reconfigure the test. So I have some ideas about maybe having an enum corresponding to each test, or similar which will follow up with eventually.

Also if this approach looks fine, will follow up with a PR to migrate the remaining tests.


mateuszrzeszutek (approved):
Nice 👍

iNikem (approved):

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
```diff
@@ -42,12 +42,15 @@
...
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
trask, line 53: will this affect subclasses? will that be less obvious compared to making `options` below static?

trask (approved):

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
anuraaga: This is a good question, I'll confirm it but I would expect it to, most jupiter functionality goes through hierarchy. I don't think I can use the subclass `configure` pattern (or any pattern I can think of) if `static`, especially because the `@BeforeAll` method must be static if this isn't here.

anuraaga: Confirmed that it's respected by subclasses. Though I did come with another pattern using an extension, will play with it in a followup (if I can ever merge past the GwtTest failure :P)

PR #3715, opened 2021-07-29
Test latest version of vert.x reactive

laurit:


instrumentation/vertx-reactive-3.5/javaagent/build.gradle.kts
```diff
@@ -10,12 +11,21 @@ muzzle {
   }
 }
 
-//The first Vert.x version that uses rx-java 2
-val vertxVersion = "3.5.0"
```
iNikem, line 14: This still seems useful, considering in how many places below this value is used

instrumentation/vertx-reactive-3.5/javaagent/build.gradle.kts
```diff
@@ -10,12 +11,21 @@ muzzle {
...
+testSets {
+  create("version35Test")
```
iNikem, line 15: Do we need a separate test set for this? Isn't it just "normal" test?

instrumentation/vertx-reactive-3.5/javaagent/src/latestDepTest/java/VertxReactiveWebServer.java
```diff
@@ -0,0 +1,177 @@
...
+
+  private static final Logger logger = LoggerFactory.getLogger(VertxReactiveWebServer.class);
+
+  private static final Tracer tracer = GlobalOpenTelemetry.getTracer("test");
+
+  public static final String TEST_REQUEST_ID_PARAMETER = "test-request-id";
+  public static final String TEST_REQUEST_ID_ATTRIBUTE = "test.request.id";
+
+  private static final String CONFIG_HTTP_SERVER_PORT = "http.server.port";
+  private static JDBCClient client;
```
iNikem, line 44: Why do we need JDBC/DB in http server test?

instrumentation/vertx-reactive-3.5/javaagent/build.gradle.kts
```diff
@@ -24,23 +34,17 @@ dependencies {
...
+  add("version35TestImplementation", "io.vertx:vertx-rx-java2:3.5.0")
+  add("version35TestImplementation", "io.vertx:vertx-web-client:3.5.0")
+  add("version35TestImplementation", "io.vertx:vertx-jdbc-client:3.5.0")
+  add("version35TestImplementation", "io.vertx:vertx-circuit-breaker:3.5.0")
 
-tasks {
-  named<Test>("test") {
-    systemProperty("testLatestDeps", findProperty("testLatestDeps"))
-  }
+  add("latestDepTestImplementation", "io.vertx:vertx-web:4.+")
```
iNikem, line 45: Does this separation mean that we actually have to extract the support for vert.x 4+ into a separate instrumentation module?

instrumentation/vertx-reactive-3.5/javaagent/build.gradle.kts
laurit: added it back

instrumentation/vertx-reactive-3.5/javaagent/build.gradle.kts
laurit: I'm afraid we do. Due to changes in api tests for 3.5 don't build with 4.x libraries so we need to make sure we use tests only with matching library version.

instrumentation/vertx-reactive-3.5/javaagent/build.gradle.kts
laurit: As instrumentation didn't change 3.5 and 4.x can be in the same module, it is only the tests that have changed.

instrumentation/vertx-reactive-3.5/javaagent/src/latestDepTest/java/VertxReactiveWebServer.java
laurit: Presumably because `JDBCClient` is also reactive. Anyway all of this is a slightly modified copy of 3.5 test.

iNikem (approved):

trask (approved):
@laurit what do you think of changing `version35Test` folder to just `test`? (similar to the finatra instrumentation)?

laurit:
@trask As far as I understand this requires something like
```
tasks {
  if (findProperty("testLatestDeps") as Boolean) {
    named("compileTestJava") {
      enabled = false
    }
    named("compileTestGroovy") {
      enabled = false
    }
  }
}
```
because it will try to compile sources in `test` against latest library which doesn't work.

Issue #3306, opened 2021-06-15
ClassNotFoundException: HandlerMappingResourceNameFilter


trask:
Repro at https://github.com/trask/jboss-ear-otel-javaagent-issue

* `mvn clean package`
* copy `hello-ear/target/hello-ear-1.0-SNAPSHOT.ear` to wildfly deployments directory
* start wildfly
* app will fail during startup with exception below

Initial analysis:

* The issue is caused by spring being provided by the ear libs, but used in the war
  https://github.com/trask/jboss-ear-otel-javaagent-issue/blob/a04696c98860255ac7ac4fe2277cb16b429bf16c/hello-web/pom.xml#L19-L20
* JBoss modules keeps an internal record of all the packages that that ear module can load, and since `io.opentelemetry.javaagent.instrumentation.springwebmvc` is not in that list, it doesn't even bother trying to load the helper class
* One potential fix for this is #3101
  We have this fix in our distro for a few months now with no issues
* I believe DataDog fixed this same issue but in a different way DataDog/dd-trace-java#1879
  I would love to understand this comment better, since it points to a potential issue with the above fix (@randomanderson if you happen to remember)
  > This can't be done at BeanDefinition construction time because Spring heavy use of clone() which sometimes only copies the classname
* I wonder if the same issue can occur for other helper classes that are looked up by name, specifically the SPI classes that we inject as helper classes, in which case maybe the better fix is at the class loader level, either generically, or specifically hacking the JBoss modules class loader
* One other thought is to put all the helper classes inside of the same package that we are instrumenting (e.g. `org.springframework.web.context` in this case). This would be nice to avoid, although I wonder if we can have similar issues with JPMS (or do we already account for this somehow)?

```
[...]

PR #3718, opened 2021-07-29
Move helper class to spring package so that loadClass can find it

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3306


trask (commented):
I would love to have a test for this somehow

laurit:
@trask tests added

trask:
> @trask tests added

❤️ the tests look great, I'll review in the next couple days

instrumentation/spring/spring-web-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springweb/WebApplicationContextInstrumentation.java
```diff
@@ -59,10 +61,29 @@ public void transform(TypeTransformer transformer) {
...
+              .loadClass("org.springframework.web.servlet.DispatcherServlet");
 
-        ((BeanDefinitionRegistry) beanFactory)
-            .registerBeanDefinition(
-                "otelAutoDispatcherFilter", new HandlerMappingResourceNameFilter.BeanDefinition());
+          // Now attempt to load our injected instrumentation class.
+          Class<?> clazz =
+              beanFactory
+                  .getBeanClassLoader()
+                  .loadClass("org.springframework.web.servlet.OpenTelemetryHandlerMappingFilter");
```
mateuszrzeszutek, line 76: `OpenTelemetryHandlerMappingFilter` is defined in the webmvc instrumentation, right? This makes spring-web and spring-webmvc instrumentations depend on each other; one can't function without the other. 
I think it'd be more preferable to have one `InstrumentationModule` delivering a single, coherent piece of functionality (webmvc instrumentation) even if we have to depend on two libs.

laurit: My understanding is that doesn't really work when when 2 modules are in different class loaders. Muzzle will fail on spring-web because instrumentation uses classes from spring-webmvc hence they can't be in the same `InstrumentationModule`.

mateuszrzeszutek: Hmm, that's true -- the instrumentation won't be applied if the classloader doesn't contain *all* references.

🤯

instrumentation/spring/spring-webmvc-3.1/wildfly-testing/src/test/resources/web.xml
```diff
@@ -0,0 +1,47 @@
...
+  -->
+  <filter>
+    <filter-name>otelAutoDispatcherFilter</filter-name>
+    <filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
+  </filter>
+
+  <filter>
+    <filter-name>testFilter</filter-name>
+    <filter-class>org.springframework.web.filter.DelegatingFilterProxy</filter-class>
+  </filter>
```
trask, line 27: it's not clear to me why the double registration?

instrumentation/spring/spring-webmvc-3.1/wildfly-testing/src/test/groovy/OpenTelemetryHandlerMappingFilterTest.groovy
```diff
@@ -0,0 +1,165 @@
...
+import io.opentelemetry.testing.internal.armeria.client.WebClient
+import io.opentelemetry.testing.internal.armeria.common.AggregatedHttpResponse
+import org.jboss.arquillian.container.test.api.Deployment
+import org.jboss.arquillian.container.test.api.RunAsClient
+import org.jboss.arquillian.spock.ArquillianSputnik
+import org.jboss.arquillian.test.api.ArquillianResource
+import org.jboss.shrinkwrap.api.Archive
+import org.jboss.shrinkwrap.api.ShrinkWrap
+import org.jboss.shrinkwrap.api.spec.EnterpriseArchive
+import org.jboss.shrinkwrap.api.spec.WebArchive
```
trask, line 23: this is amazing 👍

instrumentation/spring/spring-web-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springweb/WebApplicationContextInstrumentation.java
trask: do we have this problem for all instrumentations which use currently apply a muzzle `extraDependency`? is this change needed for this PR, or can we split out to discuss separately?

instrumentation/spring/spring-webmvc-3.1/wildfly-testing/src/test/resources/web.xml
laurit: Because there are 2 filters `otelAutoDispatcherFilter` is used by the filter added by instrumentation, `testFilter` is used by test. Basically what the integration does it attempts to figure out what would handle the request before the request processing reaches spring so to test it we need something that also runs before spring request processing to handle it so that it never reaches spring. I guess this could be useful when there is a spring security filter that returns not authorized, without this integration there wouldn't be a nice path for such requests.
With spring boot `otelAutoDispatcherFilter` would be added automatically but not with plain spring so this instrumentation is kind of useless when not using spring boot.

instrumentation/spring/spring-web-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springweb/WebApplicationContextInstrumentation.java
laurit: If we assume that all spring jars are in the same class loader then this would not be needed. Nowadays this probably isn't super important, but years ago when spring was pushing their dm server and all spring jars had osgi manifests then when instrumenting spring it was best to make sure that the code you inject with instrumentation doesn't reference anything from bundles that instrumented bundle doesn't depend on. If we just insert reference to `OpenTelemetryHandlerMappingFilter` into something like `org.springframework.web.context.support.AbstractRefreshableWebApplicationContext` we have created a dependency from `spring-web` to `spring-webmvc` which doesn't exist in original code (dependency is the other way around) which would make application fail if it for some reason used `spring-web` but not `spring-webmvc`.
`extraDependency` is usually completely fine, as long as instrumented library has a non optional dependency to something then we can freely use it in instrumentation.

trask: > `extraDependency` is usually completely fine, as long as instrumented library has a non optional dependency to something then we can freely use it in instrumentation.

oh yes this makes perfect sense thx 👍👍

trask: just adding comment in case I come back here later 😄: we can freely _use_ the `extraDependency` in the instrumentation, but we cannot freely _instrument_ the `extraDependency`

instrumentation/spring/spring-webmvc-3.1/wildfly-testing/src/test/resources/web.xml
trask: thx, I finally got it 😅

trask (approved):

PR #3719, opened 2021-07-29
Propagate context into jdk http client callback

laurit:


instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/HttpClientInstrumentation.java
```diff
@@ -134,6 +134,7 @@ public static void methodExit(
         tracer().endExceptionally(context, null, throwable);
       } else {
         future = future.whenComplete(new ResponseConsumer(context));
+        future = CompletableFutureWrapper.wrap(future);
```
mateuszrzeszutek, line 137: Nit: you can pass the parent context in `@Advice.Local` to avoid calling `Context.current()` inside the wrapper

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

instrumentation/java-http-client/javaagent/src/test/groovy/JdkHttpClientTest.groovy
```diff
@@ -57,10 +57,4 @@ class JdkHttpClientTest extends HttpClientTest<HttpRequest> implements AgentTest
   boolean testWithClientParent() {
     false
   }
-
-  // TODO: context not propagated to callback
-  @Override
-  boolean testErrorWithCallback() {
-    return false
-  }
```
trask, line 65: nice

trask (approved):

instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/CompletableFutureWrapper.java
```diff
@@ -0,0 +1,35 @@
...
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+import java.util.concurrent.CompletableFuture;
+
+public final class CompletableFutureWrapper {
+
+  private CompletableFutureWrapper() {}
+
+  public static <T> CompletableFuture<T> wrap(CompletableFuture<T> future, Context context) {
+    if (future == null || context == Context.root()) {
```
anuraaga, line 17: `future` can't be null

anuraaga (approved):

PR #3726, opened 2021-07-30
Add logback.xml to testing-common

anuraaga:
Adapted from the first one I clicked through to, https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-1.11/javaagent/src/test/resources/logback-test.xml


trask (approved):

iNikem (approved):

mateuszrzeszutek (approved):

laurit:
Should we also add `testLogging.showStandardStreams = true`?

trask:
> Should we also add `testLogging.showStandardStreams = true` or whatever the incantation is to get full test output from gradle.

It looks like this may have made the build log large and hard to find failure.

anuraaga:
Good call out. Will leave the xml and add a property to make it easier to activate logging locally

iNikem:
@anuraaga are you still interested in this PR?

iNikem:
@anuraaga ping

anuraaga (commented):
Thanks @trask

Issue #3724, opened 2021-07-30
GwtTest / VaadinTest testcontainers fails on CI


anuraaga:
GwtTest fails on CI with testcontainers startup failure. We can conceive it to be an issue with a newly pushed docker image, but `docker pull` on my local machine to try to get the latest did not reproduce the failure. Temporarily disabling the test.

```
> Task :instrumentation:gwt-2.0:javaagent:test

GwtTest > initializationError FAILED
    org.testcontainers.containers.ContainerLaunchException: Container startup failed
        at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:330)
        at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:311)
        at GwtTest.setupSpec(GwtTest.groovy:58)

        Caused by:
        org.rnorth.ducttape.RetryCountExceededException: Retry limit hit with exception
            at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:88)
            at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:323)
            ... 2 more

            Caused by:
            org.testcontainers.containers.ContainerLaunchException: Could not create/start container
                at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:497)
                at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:325)
                at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:81)
                ... 3 more

                Caused by:
                org.rnorth.ducttape.TimeoutException: org.rnorth.ducttape.TimeoutException: java.util.concurrent.TimeoutException
                    at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:56)
                    at org.testcontainers.containers.BrowserWebDriverContainer.containerIsStarted(BrowserWebDriverContainer.java:276)
                    at org.testcontainers.containers.GenericContainer.containerIsStarted(GenericContainer.java:659)
                    at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:476)
                    ... 5 more

                    Caused by:
                    org.rnorth.ducttape.TimeoutException: java.util.concurrent.TimeoutException
                        at org.rnorth.ducttape.timeouts.Timeouts.callFuture(Timeouts.java:70)
[...]

PR #3732, opened 2021-07-30
Use Firefox instead of Chrome for selenium tests

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3724
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3727


trask (approved):
thx!

mateuszrzeszutek (approved):

iNikem:
@laurit Do we know the actual reason for this problem? And why taking Firefox solved it?

laurit:
@iNikem my best guess is that https://hub.docker.com/r/selenium/standalone-chrome-debug/tags was updated. I wasn't able to figure out why exactly this happens, wasn't even able to reproduce this locally. As test started to pass with firefox I didn't investigate further.

PR #3733, opened 2021-07-30
Move slf4j logger initialization to bootstrap module

mateuszrzeszutek:
And use a slf4j logger in the `ExtensionClassLoader` class


iNikem (approved):

trask:
this may make it more difficult for distros to provide their own logging configuration, which is something that I'm interested in, and I think @pavolloffay is interested in as well (#566)

mateuszrzeszutek:
> this may make it more difficult for distros to provide their own logging configuration, which is something that I'm interested in, and I think @pavolloffay is interested in as well (#566)

Hmm, that might be true, depending on what's exposed in the custom logging feature. Could you open an issue and clarify what are the requirements for that?

mateuszrzeszutek:
I'll close that PR for now - and possibly will reopen it once things are clearer.

PR #3736, opened 2021-07-30
Fix JDK http client propagation of non-sampled traces

trask:
And adds missing test coverage to `HttpClientTest`


testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
```diff
@@ -195,6 +195,20 @@ void successfulRequestWithParent(String method) {
...
+    String method = "GET";
+    URI uri = resolveAddress("/success");
+    int responseCode = testing.runWithNonRecordingSpan(() -> doRequest(method, uri));
+
+    assertThat(responseCode).isEqualTo(200);
+
+    // sleep to ensure no spans are emitted
+    Thread.sleep(1000);
+
+    assertThat(testing.traces()).isEmpty();
```
tylerbenson, line 209: How does this verify that the proper headers are being sent?  Would you expect the server to generate a trace if it wasn't?

instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/HttpHeadersInstrumentation.java
```diff
@@ -40,9 +39,7 @@ public void transform(TypeTransformer transformer) {
 
     @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
     public static void methodExit(@Advice.Return(readOnly = false) HttpHeaders headers) {
-      if (Java8BytecodeBridge.currentSpan().isRecording()) {
-        headers = tracer().inject(headers);
-      }
+      headers = tracer().inject(headers);
```
tylerbenson, line 42: Looks like I just happened to pick an unlucky client. ;-)

tylerbenson (approved):

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
trask: ya 👍

instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/HttpHeadersInstrumentation.java
trask: or a lucky one, to find this bug 💯

trask:
closing and re-opening to re-trigger CLA check

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
```diff
@@ -195,6 +195,20 @@ void successfulRequestWithParent(String method) {
...
+  @Test
+  void successfulRequestWithNotSampledParent() throws InterruptedException {
+    String method = "GET";
+    URI uri = resolveAddress("/success");
+    int responseCode = testing.runWithNonRecordingSpan(() -> doRequest(method, uri));
+
+    assertThat(responseCode).isEqualTo(200);
+
+    // sleep to ensure no spans are emitted
+    Thread.sleep(1000);
```
anuraaga, line 207: 1s seems too slow how about around 200? I think we get traces back in just a few ms most times.

anuraaga (approved):

iNikem (approved):

PR #3738, opened 2021-07-30
Port opentelemetry-annotations-1.0 to Instrumenter API

HaloFour:
Ports the `opentelemetry-annotations-1.0` project to using Instrumenter API in the tracing aspect.

Refactors some of the attribute binding code under the `MethodSpanAttributesExtractor` class.  Removes the no-longer-used `AsyncSpanEndStrategy`/`AsyncSpanEndStrategies` implementations.


HaloFour:
Looks like `CopyOnWriteArrayList`'s iterator doesn't support `remove` so I had to rework the method to remove collected weak references.  Given the strategies are all registered by `Class<?>` it feels like there should be a way to use `ClassValue<T>` instead.

instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/MethodSpanAttributesExtractor.java
```diff
@@ -54,18 +59,81 @@ protected void onStart(AttributesBuilder attributes, REQUEST request) {
...
+   * Creates a binding of the parameters of the traced method to span attributes.
+   *
+   * @param method the traced method
+   * @return the bindings of the parameters
+   */
+  private AttributeBindings bind(Method method) {
+    AttributeBindings bindings = EmptyAttributeBindings.INSTANCE;
+
+    Parameter[] parameters = method.getParameters();
+    if (parameters == null || parameters.length == 0) {
```
anuraaga, line 72:
```suggestion
    if (parameters.length == 0) {
```

Don't think can be null

instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/MethodSpanAttributesExtractor.java
```diff
@@ -54,18 +59,81 @@ protected void onStart(AttributesBuilder attributes, REQUEST request) {
...
+      return bindings;
+    }
+
+    String[] attributeNames = parameterAttributeNamesExtractor.extract(method, parameters);
+    if (attributeNames == null || attributeNames.length != parameters.length) {
```
anuraaga, line 77: Can attributeNames be null? There are a few more null checks, if you can remove unnecessary ones it's good - we avoid them to make the code easier to read for someone that isn't familiar with it.

instrumentation-api-annotation-support/src/test/groovy/io/opentelemetry/instrumentation/api/annotation/support/MethodCacheTest.groovy
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+package io.opentelemetry.instrumentation.api.annotation.support
+
+import spock.lang.Specification
+
+import java.lang.reflect.Method
+import java.util.function.Function
+
+class MethodCacheTest extends Specification {
```
anuraaga, line 13: These tests look like they'd be easier to maintain in Java

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -61,25 +60,32 @@ public static TracingOperatorBuilder newBuilder() {
...
+  private static <T> Function<? super Publisher<T>, ? extends Publisher<T>> tracingLift(
+      ReactorAsyncOperationEndStrategy asyncOperationEndStrategy) {
+    return Operators.lift(new Lifter<>(asyncOperationEndStrategy));
   }
 
   public static class Lifter<T>
       implements BiFunction<Scannable, CoreSubscriber<? super T>, CoreSubscriber<? super T>> {
 
+    /** Holds reference to strategy to prevent it from being collected. */
+    @SuppressWarnings("FieldCanBeLocal")
```
mateuszrzeszutek, line 82: 👍

instrumentation/opentelemetry-annotations-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/otelannotations/WithSpanInstrumenter.java
```diff
@@ -0,0 +1,90 @@
...
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.instrumentation.api.annotation.support.MethodSpanAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.api.tracer.SpanNames;
+import java.lang.reflect.Method;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class WithSpanInstrumenter {
```
mateuszrzeszutek, line 18: We usually name these classes `*Singletons`:
```suggestion
public final class WithSpanSingletons {
```

mateuszrzeszutek (approved):
👍 
Thanks!

trask:
@HaloFour when you have a chance to look at remaining comments, would love to merge this as we discovered this change will help with #3691

HaloFour:
@trask 

> when you have a chance to look at remaining comments,

Will try to address tonight.  Work has been a bit insane this week.  😅

instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/MethodSpanAttributesExtractor.java
HaloFour: The definition of the functional interface allows for it, `null` indicating that the method has no parameters to map to attributes:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/ParameterAttributeNamesExtractor.java#L26

I could change that to specify returning an empty array if preferred.

instrumentation-api-annotation-support/src/test/groovy/io/opentelemetry/instrumentation/api/annotation/support/MethodCacheTest.groovy
HaloFour: I can port them.  To be honest Spock still confuses the heck out of me, especially mocking interactions.  😅

instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/MethodSpanAttributesExtractor.java
trask: confusingly, I think `@Nullable String[]` means a non-null array of nullable strings, while `String @Nullable []` means a nullable array of non-null strings

HaloFour: That's terrifying.  I do defensively check for either the array being null or each element being null, so would that be `@Nullable String @Nullable []`?

HaloFour: I'll review the two implementations to see if it's possible that either will return `null`, if not I can change the annotation.  Is it still worth having the defensive check even if the interface states that it shouldn't be `null`?  I hate the idea of a bug in instrumentation leading to errors at runtime.

trask: I think @anuraaga's point is that ParameterAttributeNamesExtractor.extract() returns `@Nullable String[]`, so you don't need to check if the array is `null`

HaloFour: Got it, so if the array itself can't be null (and that does seem to be the case) then the null check is unnecessary and I can leave the annotations as they are.  Thanks!

trask: we are (slowly) rolling out [NullAway](https://github.com/uber/NullAway) checking, which will hopefully help (or maybe add to?) some of the confusion

trask: (also, I could be wrong about how NullAway interprets `@Nullable String[]`, my assumption is based on how Checker Framework works, it will be good to get confirmation here from @anuraaga)

trask: I chatted with @anuraaga, NullAway (luckily) does not annotate the contents of the array (like the more powerful, but infinitely more confusing Checker Framework), so `@Nullable String[]` means what you thought it meant (Nullable applies to the array, not to the strings). Sorry for the confusion.

trask:
thx ❤️

PR #3743, opened 2021-07-31
Correctly use muzzleTooling configuration to allow for external additions

iNikem:
Should be merged and published before main project can drop its old plugin definition


gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
```diff
@@ -34,11 +34,11 @@ val RANGE_COUNT_LIMIT = 10
 
 val muzzleConfig = extensions.create<MuzzleExtension>("muzzle")
 
-val muzzleTooling by configurations.creating {
+val muzzleTooling: Configuration by configurations.creating {
```
anuraaga, line 37: We generally wouldn't specify the type for this sort of DSL usage, is it needed?

anuraaga (approved):

PR #3744, opened 2021-07-31
Add JFR option to benchmark

trask:


benchmark-overhead-jmh/build.gradle.kts
```diff
@@ -28,20 +28,37 @@ tasks {
   val jmhFork = gradle.startParameter.projectProperties["jmh.fork"]?.toInt()
   val jmhIncludes = gradle.startParameter.projectProperties["jmh.includes"]
 
+  // this option should point to the benchmark that you want to run with JFR
```
anuraaga, line 31: Is there a reason to not keep this boolean? It's a bit mysterious to have this and jmh.includes both adding to the included benchmarks. Or should jmh.includes be ignored when flightRecording is set because there's some requirement only a single benchmark is run?

benchmark-overhead-jmh/build.gradle.kts
```diff
@@ -28,20 +28,37 @@ tasks {
...
+        "-XX:StartFlightRecording=settings=profile.jfc,delay=50s,duration=300s,filename=recording.jfr",
+        "-XX:FlightRecorderOptions=stackdepth=1000",
+        "-XX:+UnlockDiagnosticVMOptions",
+        "-XX:+DebugNonSafepoints"
+      ))
+
+      includes.add(jmhFlightRecording)
+
+      fork.set(1)
+      iterations.set(30)
```
iNikem, line 60: This is generous :) Why so many?

benchmark-overhead-jmh/build.gradle.kts
trask: ya, it was not clear, I generalized the parameters, removed the duplication, and added some comments

benchmark-overhead-jmh/build.gradle.kts
trask: in order to get a meaningful profile of the agent code when benchmarking a whole app (where the agent code is a small subset)

but that's pretty specific reasoning to this module, so generalized the parameterization, with the goal to push it up to `otel.jmh-conventions` at some point

anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

PR #3749, opened 2021-08-02
Add library instrumentation for Ratpack server

anuraaga:
The attributes are not very complete yet, I will try to follow up to do so. One major annoyance is e.g., accessing remoteAddress involves an unstable Guava type that has had its API changed throughout the versions... On the bright side the tests are shareable.

Ideally we can use this in the javaagent instrumentation I guess, but we have our tricky chicken / egg of Netty instrumentation currently winning over any framework-specific server instrumentation if on the same thread.

On the bright side, might be able to replace the execution-related instrumentation with the library instrumentation pattern.

Will add client in the future.


instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackProductionErrorHandler.java
```diff
@@ -0,0 +1,70 @@
...
+import ratpack.handling.Context;
+
+// Copied from
+// https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/core/error/internal/DefaultProductionErrorHandler.java
+// since it is internal and has had breaking changes.
+final class RatpackProductionErrorHandler implements ClientErrorHandler, ServerErrorHandler {
+
+  static final RatpackProductionErrorHandler INSTANCE = new RatpackProductionErrorHandler();
+
+  private static final Logger logger = LoggerFactory.getLogger(RatpackProductionErrorHandler.class);
```
mateuszrzeszutek, line 38: This is using "our" slf4j, not ratpack's slf4j. Is that okay?

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/OpenTelemetryExecInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public void intercept(Execution execution, ExecType type, Block continuation) throws Exception {
+    Context otelCtx = execution.maybeGet(Context.class).orElse(null);
+    if (otelCtx != null) {
+      // Execution already has a context, this is an asynchronous resumption and we need to make
+      // the context current.
+      try (Scope ignored = otelCtx.makeCurrent()) {
+        continuation.execute();
+      }
+      return;
```
mateuszrzeszutek, line 27: Won't returning here cause the scope to never be closed? Doesn't `OpenTelemetryServerHandler` always add both `Context` and `Scope`? If they're both present then the second part of this method won't execute.

anuraaga: The handler is always invoked as part of the execution of a continuation. So initially, there is no context or scope. After continuation.execute here there will be both context and scope. And then we close the scope, leaving the context. If the execution is reused then the context is remounted above.

I was hoping my comment below captured this is there a way I can make it clearer?

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackGetter.java
```diff
@@ -0,0 +1,29 @@
...
+
+package io.opentelemetry.instrumentation.ratpack;
+
+import io.opentelemetry.context.propagation.TextMapGetter;
+import javax.annotation.Nullable;
+import ratpack.http.Request;
+
+final class RatpackGetter implements TextMapGetter<Request> {
+
+  RatpackGetter() {}
```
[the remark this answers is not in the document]
anuraaga: IIUC a public method in a private class is still accessible with reflection, there is no Class.setAccessible a private class is respected by javac but has no effect at runtime I think. So to ensure privateness of the constructor it has to be made private.

This of course assumes the constructor of a private class is public like public classes I assumed it is but am not sure.

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackProductionErrorHandler.java
anuraaga: Hmm good point. In practice I don't think it's possible to use this class as the context should always have a error handler I think. Should I verify that and throw an exception outside of the loop that finds `delegate` instead of this fallback? Maybe slightly dangerous if there is some unexpected ratpack behavior though.

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/OpenTelemetryExecInterceptor.java
anuraaga: Maybe it's enough if I flip the order of the ifs so the order matches execution order?

mateuszrzeszutek: So when there's both context and scope in the execution then this block `if (otelCtx != null) { /* ... */ return; }` will execute and return from the `ExecInterceptor` before closing & removing scope. Maybe I'm missing or misunderstanding something, but wouldn't this class close the scope only in a "no context & scope" situation? While there seem to be only 3 options possible ("no context & no scope" - before handler, "context & scope" - after handler but before execution finishes, "context & no scope" - after execution)

mateuszrzeszutek: > Maybe it's enough if I flip the order of the ifs so the order matches execution order?

I think it'd be enough - that's what confused me here

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackProductionErrorHandler.java
mateuszrzeszutek: Hmm, maybe just logging something would be enough? And a bit safer

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/OpenTelemetryExecInterceptor.java
anuraaga: Ok - don't think it's possible to get any more descriptive than the methods I extracted too :)

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackProductionErrorHandler.java
anuraaga: Added a WarnOnce

mateuszrzeszutek (approved):

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/OpenTelemetryExecInterceptor.java
mateuszrzeszutek: Ahh I finally understood that - I didn't realize that `ExecInterceptor` basically wraps over the `Handler`, now it makes sense. Thanks!

PR #3751, opened 2021-08-02
Support multiple subscriptions for Reactor spans

HaloFour:
Explores refactoring the Reactor 3.x async strategy to use an operator instead of composing over the various `doOnSignal` methods.  Using an operator provides closer access to the Subscriber/Subscription allowing it to be decorated so that I can intercept each subscription and the signals sent to them.

What this would enable is allowing multiple spans to be started/ended for a Reactor publisher, one for the original method call plus one for each subsequent subscription if the publisher is re-used, which can happen accidentally or can be intentional if treated as a Reader/IO monad.

It looks like RxJava 2/3 also support the same approach although it is a bit more involved as there are separate operator/observer interfaces for each of the reactive types.

Discussion: https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/3753

### DIFF
--- a/instrumentation/reactor-3.1/javaagent/build.gradle.kts
+++ b/instrumentation/reactor-3.1/javaagent/build.gradle.kts
@@ -13,7 +13,7 @@ muzzle {
 
 tasks.withType<Test>().configureEach {
   // TODO run tests both with and without experimental span attributes
-  jvmArgs("-Dotel.instrumentation.reactor.experimental-span-attributes=true")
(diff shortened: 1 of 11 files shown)


trask:
hi @HaloFour! any objection to closing this (just trying to clean up the backlog a little), and you can re-open when you start working on it again? I think if I close it you won't be able to re-open it, but if you close it, I think you will be able to re-open yourself. And of course you can always ping on the closed PR and one of us will re-open it if that fails.

trask:
Closing, but please ping back if you start working on it again and would like it to be re-opened!

PR #3756, opened 2021-08-03
Fill HTTP_CLIENT_IP in ServerInstrumenter

anuraaga:
Looks like this doesn't need to be exposed on the normal HttpAttributesExtractor

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ServerInstrumenter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ServerInstrumenter.java
@@ -5,10 +5,15 @@
 
 package io.opentelemetry.instrumentation.api.instrumenter;
 
+import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.context.propagation.TextMapGetter;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.net.NetAttributesExtractor;
 import io.opentelemetry.instrumentation.api.internal.ContextPropagationDebug;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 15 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ServerInstrumenter.java
```diff
@@ -29,4 +34,133 @@ public Context start(Context parentContext, REQUEST request) {
...
+      String clientIp = getForwardedClientIp(request);
+      if (clientIp == null && netAttributesExtractor != null) {
+        clientIp = netAttributesExtractor.peerIp(request, response);
+      }
+      set(attributes, SemanticAttributes.HTTP_CLIENT_IP, clientIp);
+    }
+
+    @Nullable
+    // Visible for testing
+    String getForwardedClientIp(REQUEST request) {
```
anuraaga, line 86: Just copied from HttpServerTracer

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ServerInstrumenter.java
```diff
@@ -29,4 +34,133 @@ public Context start(Context parentContext, REQUEST request) {
...
+
+    @Override
+    protected void onStart(AttributesBuilder attributes, REQUEST request) {}
+
+    @Override
+    protected void onEnd(
+        AttributesBuilder attributes, REQUEST request, @Nullable RESPONSE response) {
+      String clientIp = getForwardedClientIp(request);
+      if (clientIp == null && netAttributesExtractor != null) {
+        clientIp = netAttributesExtractor.peerIp(request, response);
```
trask, line 79: it's not clear to me that `http.client_ip` should be populated in this case since easy for exporter/backend to fallback to `net.peer.ip`

trask (approved):
I like the auto-capture

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/ServerInstrumenter.java
anuraaga: We do it right now, though I agree it's not clear. Figured we'd stick to what we have for now though.

trask: makes sense, I'll open a spec issue to seek clarification and we can revisit later

PR #3762, opened 2021-08-03
Overhead testing to use external db

breedx-splk:
(this builds on #3648)

This migrates the database out of the JVM and into an external postgresql. It seems unfair to have an in-memory database for benchmarks, because it's an uncommon scenario in production. The in-memory database would likely inflate GC pressure and allocation rates. This PR moves it to an external postgres container, which petclinic readily supports.


testing-overhead/src/test/java/io/opentelemetry/containers/PostgresContainer.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.utility.DockerImageName;
+import org.testcontainers.utility.MountableFile;
+
+public class PostgresContainer {
```
jkwatson, line 15: any thoughts to using this thing?  https://www.testcontainers.org/modules/databases/postgres/

breedx-splk: Interesting. I hadn't seen that. I'll try it out and see how it goes...

breedx-splk: Seems ok. It doesn't provide a whole lot, but it does abstract out a few pg internals (like having to know the environment var names etc).  👍🏻 Thanks!

testing-overhead/src/test/java/io/opentelemetry/containers/PetClinicRestContainer.java
```diff
@@ -54,6 +53,14 @@ public PetClinicRestContainer(Network network, Startable collector, Agent agent,
         .withExposedPorts(PETCLINIC_PORT)
         .withFileSystemBind(namingConventions.localResults(), namingConventions.containerResults())
         .waitingFor(Wait.forHttp("/petclinic/actuator/health").forPort(PETCLINIC_PORT))
+        .withEnv("spring_profiles_active", "hsqldb,spring-data-jpa")
```
iNikem, line 56: If you are using postgre, then `hsqldb` profile shouldn't be needed, right?

testing-overhead/src/test/java/io/opentelemetry/containers/PetClinicRestContainer.java
```diff
@@ -54,6 +53,14 @@ public PetClinicRestContainer(Network network, Startable collector, Agent agent,
...
+        .withEnv("spring_datasource_url", "jdbc:postgresql://postgres:5432/" + PostgresContainer.DATABASE_NAME)
+        .withEnv("spring_datasource_username", PostgresContainer.USERNAME)
+        .withEnv("spring_datasource_password", PostgresContainer.PASSWORD)
+        .withEnv("spring_datasource_driver-class-name", "org.postgresql.Driver")
+        .withEnv("spring_jpa_database", "POSTGRESQL")
+        .withEnv("spring_jpa_database-platform", "org.hibernate.dialect.PostgreSQLDialect")
```
iNikem, line 62: This certainly should not be needed.

testing-overhead/src/test/java/io/opentelemetry/containers/PetClinicRestContainer.java
breedx-splk: Good catch, yeah, the `hsqldb` should have been replaced with `postgresql` according to the docs. Fixed it.

testing-overhead/src/test/java/io/opentelemetry/containers/PetClinicRestContainer.java
breedx-splk: Right, it figures out a sane default, and I confirmed it.

breedx-splk: Doh, sorry I didn't recognize the multi-line. I removed the other two and it still works as you predicted.  :) Thanks.

iNikem (approved):

PR #3763, opened 2021-08-03
Publish muzze plugins to Gradle Plugin Portal only when manually requested

iNikem:
Restored publishing to Sonatype OSS repository on every change

### DIFF
--- a/gradle-plugins/build.gradle.kts
+++ b/gradle-plugins/build.gradle.kts
@@ -5,6 +5,7 @@ plugins {
   `maven-publish`
 
   id("com.gradle.plugin-publish")
+  id("io.github.gradle-nexus.publish-plugin")
 }
 
 group = "io.opentelemetry.instrumentation"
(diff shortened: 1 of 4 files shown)


gradle-plugins/build.gradle.kts
```diff
@@ -53,9 +54,31 @@ gradlePlugin {
...
+    }
+  }
+
+  connectTimeout.set(Duration.ofMinutes(5))
+  clientTimeout.set(Duration.ofMinutes(5))
+}
+
+tasks {
+  publishPlugins {
+    enabled = !version.toString().contains("SNAPSHOT")
```
trask, line 82: should `version` above be changed back to SNAPSHOT for now?

iNikem: I really want to publish this to Gradle Portal. The whole idea of the previous PR was to achieve that.

trask: is `enabled` setting here needed, since we only run `publishPlugins` from manual workflow? is it just for safety?

trask (approved):

gradle-plugins/build.gradle.kts
iNikem: Mostly for safety, yes. I tried to disable `publishToSonatype` as well, but it gradle did not find such task :( Probably some evaluation order or rule magic :(

anuraaga:
Just want to confirm we're still aiming to end up in a state where we publish gradle plugins with the same version as the repo, along with the normal release-build?

iNikem:
> Just want to confirm we're still aiming to end up in a state where we publish gradle plugins with the same version as the repo, along with the normal release-build?

I don't have any objections to this goal. I don't have clarity about achieving this goal and will need some help.

anuraaga:
> I don't have any objections to this goal. I don't have clarity about achieving this goal and will need some help.

Cool - once things settle down with the folder I'll give it a stab. I guess after this PR? :)

anuraaga (approved):

PR #3765, opened 2021-08-03
test script cleans up the data it inserts.

breedx-splk:
Because the test script creates a bunch of data, it's helpful to remove that at the end of the test so that we get more consistent performance across a test run (instead of just continuously slower due to more data). Without this, it is also nearly impossible to perform longer test runs with the in-memory database.

### DIFF
--- a/testing-overhead/k6/basic.js
+++ b/testing-overhead/k6/basic.js
@@ -1,5 +1,5 @@
 import http from "k6/http";
-import { check } from "k6";
+import {check} from "k6";
 import names from "./names.js";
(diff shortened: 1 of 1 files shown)


iNikem:
When exactly this removal runs?

mateuszrzeszutek (approved):

mateuszrzeszutek:
> When exactly this removal runs?

I think it runs as a part of the test - it deletes the data just after creating it.

iNikem (approved):

PR #3771, opened 2021-08-04
Simplify shouldStart() check for SERVER & CONSUMER spans

mateuszrzeszutek:
After our last discussion on Tuesday I wanted to check whether this `ServerSpan.exists(parentContext) || ConsumerSpan.exists(parentContext)` check is really needed after all. It turns out the only instrumentation that made use of it was the spring-integration one, and there's an alternate solution for this one: do not create a span if there's ANY span in the current context - please see the comment in the `TracingChannelInterceptor` class for a detailed explanation on that.

If SERVER & CONSUMER spans don't get any special treatment then we'll be able to handle them with `InstrumentationType` proposed in another PR.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
@@ -103,8 +103,10 @@ public class Instrumenter<REQUEST, RESPONSE> {
     SpanKind spanKind = spanKindExtractor.extract(request);
     switch (spanKind) {
       case SERVER:
+        suppressed = ServerSpan.exists(parentContext);
+        break;
       case CONSUMER:
-        suppressed = ServerSpan.exists(parentContext) || ConsumerSpan.exists(parentContext);
+        suppressed = ConsumerSpan.exists(parentContext);
         break;
       case CLIENT:
         suppressed = ClientSpan.exists(parentContext);
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/BaseTracer.java
@@ -107,8 +107,10 @@ public abstract class BaseTracer {
         suppressed = ClientSpan.exists(context);
         break;
       case SERVER:
+        suppressed = ServerSpan.exists(context);
+        break;
       case CONSUMER:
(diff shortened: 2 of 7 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -103,8 +103,10 @@ public boolean shouldStart(Context parentContext, REQUEST request) {
     SpanKind spanKind = spanKindExtractor.extract(request);
     switch (spanKind) {
       case SERVER:
+        suppressed = ServerSpan.exists(parentContext);
```
lmolkova, line 106: I was thinking along the lines of your proposal here #3734 to get rid of this switch alltogether. 

Here's my latest attempt:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/ec92357ebb1edfdce6e011954dfae7348e909425/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java#L102

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3691/files#diff-abeb4b4ae8d648931084077221aa9f1ca4c6599941dfc9891e0d5f9cb8e7cf09


Then instrumentation type defines suppression strategies for different kinds. It also eliminates the need of Client/Server/Consumer spans

trask (commented):
alternatively, what do you think of replacing `ServerSpan` and `ConsumerSpan` keys with a single `LocalRootSpan` key?

lmolkova:
> alternatively, what do you think of replacing `ServerSpan` and `ConsumerSpan` keys with a single `LocalRootSpan` key?

Please check out my proposal to use `InstrumetnationType` here 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f41ff77958bf00960e89385b92ae7c0fb078c1ce/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/ServerSpan.java#L21


- we have to use it to make tests pass (so Tracers and Instrumenters would work together)

trask (approved):

mateuszrzeszutek:
> alternatively, what do you think of replacing `ServerSpan` and `ConsumerSpan` keys with a single `LocalRootSpan` key?

That'd work too. Introducing a `LocalRootSpan` is not necessary though, the spring-integration instrumentation still makes perfect sense when you just change the condition for adding a span - which is a bit simpler than introducing a new `LocalRootSpan` concept.

PR #3785, opened 2021-08-06
Fix vaadin latest dep test

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3782
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3781 (this one has one more failure, but it seems like a one time occurrence)

### DIFF
--- a/instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/NodeUpdaterInstrumentation.java
+++ b/instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/NodeUpdaterInstrumentation.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.vaadin;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeTransformer;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+/**
+ * This advice is used only when running tests. Works around an issue running testLatestDeps.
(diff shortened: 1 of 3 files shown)


instrumentation/vaadin-14.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vaadin/NodeUpdaterInstrumentation.java
```diff
@@ -0,0 +1,49 @@
...
+import io.opentelemetry.javaagent.extension.instrumentation.TypeTransformer;
+import java.util.Map;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+/**
+ * This advice is used only when running tests. Works around an issue running testLatestDeps.
+ * Currently the latest vaadin release is 20, it is likely that this advice can be deleted when
+ * there is a new version.
```
trask, line 20: can you add to this comment explaining this problem only affects our tests, and is not a problem for end users?

laurit: I would assume that it is a problem for at least some users https://github.com/GoogleChrome/workbox/issues/2904 Our test isn't exactly a regular vaadin application. We let vaadin development mode generate a bunch of files each time test is first run after clean that regular users would add to version control. I suspect that the javascript tooling used probably doesn't attempt to find new package versions on each run so if they set up their project before yesterday this is yet to break for them (this is speculation). If it breaks they probably know how to pin package to working version. I also initially tried to make it use the previous working version of `workbox-build` but failed at it. I guess should have use `pnpm-lock.yaml` to pin version. So after failing with javascript tooling I turned to what I know, patching java code. As an alternative we could find the latest working vaadin version (something that uses `workbox-webpack-plugin` before 6.x?) and make latest dep tests use that until vaadin switches to `workbox-webpack-plugin` 6.2.

trask: > As an alternative we could find the latest working vaadin version (something that uses `workbox-webpack-plugin` before 6.x?) and make latest dep tests use that until vaadin switches to `workbox-webpack-plugin` 6.2.

I think this would be my preference

PR #3789, opened 2021-08-06
Fix tomcat 9.0.52

laurit:
In tomcat 9.0.52 async servlet tests don't pass without instrumenting `org.apache.tomcat.util.threads.ThreadPoolExecutor`

### DIFF
--- a/instrumentation/tomcat/tomcat-7.0/javaagent/build.gradle.kts
+++ b/instrumentation/tomcat/tomcat-7.0/javaagent/build.gradle.kts
@@ -13,15 +13,15 @@ muzzle {
 }
 
 dependencies {
-  library("org.apache.tomcat.embed:tomcat-embed-core:7.0.4")
+  compileOnly("org.apache.tomcat.embed:tomcat-embed-core:7.0.4")
   implementation(project(":instrumentation:tomcat:tomcat-common:javaagent"))
   implementation(project(":instrumentation:servlet:servlet-3.0:javaagent"))
   testInstrumentation(project(":instrumentation:servlet:servlet-javax-common:javaagent"))
   // Make sure nothing breaks due to both 7.0 and 10.0 modules being present together
   testInstrumentation(project(":instrumentation:tomcat:tomcat-10.0:javaagent"))
 
-  // Tests need at least version 9 to have necessary classes to configure the embedded tomcat...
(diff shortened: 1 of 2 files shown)


trask (approved):

instrumentation/tomcat/tomcat-7.0/javaagent/build.gradle.kts
```diff
@@ -13,15 +13,15 @@ muzzle {
...
   implementation(project(":instrumentation:servlet:servlet-3.0:javaagent"))
   testInstrumentation(project(":instrumentation:servlet:servlet-javax-common:javaagent"))
   // Make sure nothing breaks due to both 7.0 and 10.0 modules being present together
   testInstrumentation(project(":instrumentation:tomcat:tomcat-10.0:javaagent"))
 
-  // Tests need at least version 9 to have necessary classes to configure the embedded tomcat...
-  // ... but not newer that version 10, because its servlet 5.
-  testLibrary("org.apache.tomcat.embed:tomcat-embed-core:[9.+, 10)")
+  testLibrary("org.apache.tomcat.embed:tomcat-embed-core:8.0.41")
+  testLibrary("org.apache.tomcat.embed:tomcat-embed-jasper:8.0.41")
```
trask, line 24: is this part of the comment is still relevant?
```suggestion
  // Tests need at least version 8.0.41 to have necessary classes to configure the embedded tomcat...
  // ... but not newer that version 10, because its servlet 5.
  testLibrary("org.apache.tomcat.embed:tomcat-embed-core:8.0.41")
  testLibrary("org.apache.tomcat.embed:tomcat-embed-jasper:8.0.41")
```

laurit: In my opinion this comment isn't relevant. It was previously using `9+` for both regular build and latest deps which is bad because it is really confusing when regular build starts failing without any code changes. `8.0.41` is the version used in `servlet-3.0` tests. I didn't try to figure out whether this is really the earliest version that works, only checked that `8.0.1` didn't work. The part of version 10 is kind of obvious, `servlet-3.0` uses the same construct and doesn't have a comment. Jasper dependency was added to get rid of a non fatal stack trace in test output.

trask:
btw thanks for finding/fixing, this was quite confusing failure 👍

PR #3793, opened 2021-08-07
Call shouldStart for Spring handler span

trask:
(partial motivation is that I'd like to introduce a configuration option to suppress handler spans in the future)


mateuszrzeszutek (approved):

mateuszrzeszutek:
> (partial motivation is that I'd like to introduce a configuration option to suppress handler spans in the future)

It could be an `InstrumentatyionType` maybe?

PR #3796, opened 2021-08-07
Update jaxrs-common to Instrumenter API

trask:

PR #3797, opened 2021-08-08
Update jaxrs-1.0 to Instrumenter API

trask:


instrumentation/jaxrs/jaxrs-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/HandlerData.java
```diff
@@ -1,74 +1,42 @@
...
-    Span serverSpan = ServerSpan.fromContextOrNull(parentContext);
-
-    // When jax-rs is the root, we want to name using the path, otherwise use the class/method.
-    String spanName;
-    if (serverSpan == null) {
-      spanName = pathBasedSpanName;
-    } else {
-      spanName = SpanNames.fromMethod(target, method);
-      updateServerSpanName(parentContext, serverSpan, pathBasedSpanName);
-    }
```
trask, line 53: Similar to #3796 

> I removed this special case of having a different name for the handler span name when there is no server span (now the handler span has the same name whether or not there is a server span, which is consistent with other instrumentations).

instrumentation/jaxrs/jaxrs-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrs/v1_0/HandlerData.java
```diff
@@ -5,70 +5,43 @@
...
+    this.method = method;
   }
 
-  private static void setCodeAttributes(SpanBuilder spanBuilder, Class<?> target, Method method) {
-    spanBuilder.setAttribute(SemanticAttributes.CODE_NAMESPACE, target.getName());
-    if (method != null) {
-      spanBuilder.setAttribute(SemanticAttributes.CODE_FUNCTION, method.getName());
-    }
+  String spanName() {
+    return SpanNames.fromMethod(target, method);
```
mateuszrzeszutek, line 36: Why not use `CodeSpanNameExtractor`?

mateuszrzeszutek (approved):

anuraaga (approved):

Issue #3144, opened 2021-05-31
What is a "handler span"?


trask:
Related to #3140.

Here's one proposal:

-----------

Most web frameworks have a concept of routing and handlers.

Handler spans have a couple of purposes:
* They represent the portion of a request that is spent in an application handler (as opposed to time spent in the web framework)
* They help users map requests easily to application code (while server span names are based on url paths, the handler span names are based on application class name / method name)

Note: If user code returns a "promise" of some kind (e.g. Future, Mono) the handler span should end when the promise completes.

-----------

I think this is mostly consistent with how we implement handler spans, with a couple of questions so far:

* If a route is not found (and no application handler is invoked), should we capture a handler span? (currently in some cases we do and in some cases we don't)

* The current webflux handler span maps to `ServerWebExchange.handle()`, which is pretty much the entire request-response lifecycle (including routing), it would be nice if this can be scoped down a bit

anuraaga:
Another question is whether we should have these without adding something to the spec.

iNikem:
I don't think we are ready and should introduce one more concept to our instrumentations. I mean, we (as Otel community as a whole) have not yet solved the problem with (nested) CLIENT spans. Should we open one more can of worms before we close that one?

trask:
Closing, these spans are now opt-in under an experimental flag, can revisit later if anyone has renewed interest.

PR #3798, opened 2021-08-08
Update webflux to Instrumenter API (and improvements/simplifications)

trask:
In addition to updating to Instrumenter API:

* Handler span now maps to `org.springframework.web.reactive.HandlerAdapter.handle()` instead of `org.springframework.web.reactive.DispatcherHandler.handle()`, which more closely maps to the user handler. This aligns better with other handler spans, and the previous (strange to me) handler instrumentation placement was what motivated #3144. Also, I think this change addresses the flaky problems that required #3150 previously (without needing the fix from #3150). @laurit I was not able to get any flakes following your notes to repro the flaky problem (increased request count to 300 for "Multiple GETs to delaying route" test and used spring boot 2.5.0), but would be great if you have time to give it a look/try also.

* Removed `spring-webflux.request.predicate` experimental attribute because the handler span is not available yet when this is calculated. If anyone wants to keep this, I can either put it on the SERVER span, or propagate it down to where the handler span is started.


instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/HandlerAdapterInstrumentation.java
```diff
@@ -64,55 +65,70 @@ public void transform(TypeTransformer transformer) {
...
+        handlerType = handlerMethod.getMethod().getDeclaringClass().getName();
+      } else {
+        spanName = AdviceUtils.spanNameForHandler(handler);
+        handlerType = handler.getClass().getName();
       }
+
+      span.updateName(spanName);
+      if (SpringWebfluxConfig.captureExperimentalSpanAttributes()) {
+        span.setAttribute("spring-webflux.handler.type", handlerType);
+      }
```
mateuszrzeszutek, line 109: Could we move that code inside `instrumenter.start()` call?

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/DispatcherHandlerInstrumentation.java
```diff
@@ -41,41 +37,20 @@ public void transform(TypeTransformer transformer) {
...
-        @Advice.Return(readOnly = false) Mono<Void> mono,
-        @Advice.Local("otelScope") Scope otelScope,
-        @Advice.Local("otelContext") Context otelContext) {
-      if (throwable == null && mono != null) {
-        mono = AdviceUtils.setPublisherSpan(mono, otelContext);
-      } else if (throwable != null) {
-        AdviceUtils.finishSpanIfPresent(exchange, throwable);
+        @Advice.Return(readOnly = false) Mono<Void> mono) {
+      if (mono != null) {
+        // note: it seems like this code should go in HandleAdvice @OnMethodExit
```
anuraaga, line 49: Do you mean Enter? This code seems to be in HandleAdvice.OnMethodExit right now

anuraaga (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/DispatcherHandlerInstrumentation.java
trask: oh no, two different `HandleAdvice` classes, I fixed the comment 👍

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
```diff
@@ -35,114 +27,29 @@ public static String spanNameForHandler(Object handler) {
...
-    }
-
-    @Override
-    public void cancel() {
-      if (completed.compareAndSet(false, true)) {
-        finishSpanIfPresent(otelContext, null);
-      }
-      subscription.cancel();
-    }
+  interface OnSpanEnd {
```
mateuszrzeszutek, line 52: Nit
```suggestion
  @FunctionalInterface
  interface OnSpanEnd {
```

mateuszrzeszutek (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
```diff
@@ -5,25 +5,17 @@
...
-import reactor.core.publisher.Operators;
-import reactor.util.context.Context;
 
 public class AdviceUtils {
 
-  public static final String CONTEXT_ATTRIBUTE = AdviceUtils.class.getName() + ".Context";
+  public static final String ON_SPAN_END = AdviceUtils.class.getName() + ".Context";
 
   public static String spanNameForHandler(Object handler) {
     String className = ClassNames.simpleName(handler.getClass());
```
laurit, line 21: Although this part wasn't changed in this pr getting a span name from an object is also used in other instrumentations so it would be nice if someone figured out how to do this properly. The problem here is that handler could be anything. For example `SpringWebFluxTestApplication` when you add
```
  @Component("greetingHandlerFunction")
  @Scope(value = "prototype", proxyMode = ScopedProxyMode.TARGET_CLASS)
  static class GreetingHandlerFunction implements HandlerFunction<ServerResponse> {
    private final GreetingHandler greetingHandler

    GreetingHandlerFunction(GreetingHandler greetingHandler) {
      this.greetingHandler = greetingHandler
    }

    @Override
    Mono<ServerResponse> handle(ServerRequest request) {
      return greetingHandler.defaultGreet()
    }
  }
```
and change the start of `greetRouterFunction` to
```
  @Bean
  RouterFunction<ServerResponse> greetRouterFunction(GreetingHandler greetingHandler, HandlerFunction<ServerResponse> greetingHandlerFunction) {
    return route(GET("/greet"), greetingHandlerFunction
```
Then span name is
`SpringWebFluxTestApplication$GreetingHandlerFunction$$EnhancerBySpringCGLIB$$d662c0c3.handle`
When you change scope annotation to 
`@Scope(value = "prototype", proxyMode = ScopedProxyMode.INTERFACES)`
then span name is
`$Proxy119.handle`
If we'd assume that handler is a spring bean we could try to unwrap it with something like https://github.com/spring-projects/spring-framework/commit/efe3a35da871a7ef34148dfb65d6a96cac1fccb9

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/ExperimentalAttributesExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+  @Override
+  protected void onEnd(AttributesBuilder attributes, Object handler, @Nullable Void unused) {}
+
+  private static String getHandlerType(Object handler) {
+    if (handler instanceof HandlerMethod) {
+      // Special case for requests mapped with annotations
+      HandlerMethod handlerMethod = (HandlerMethod) handler;
+      return handlerMethod.getMethod().getDeclaringClass().getName();
+    } else {
+      return handler.getClass().getName();
```
laurit, line 33: Handler could be a proxy so name could be `SpringWebFluxTestApplication$GreetingHandlerFunction$$EnhancerBySpringCGLIB$$d662c0c3` or `$Proxy119`

laurit (approved):
@trask I couldn't reproduce it either, perhaps something else has changed that makes triggering this harder.

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
trask: 👍 I created issue for this #3813

trask:
@HaloFour have you seen the cancellation issue that was fixed in #3150? I'm wondering if you have any thoughts whether that fix should be applied to the `@WithSpan` Mono handling? I'm trying to get a sense if it's a general issue that should be handled everywhere Monos are returned (regardless of whether we can reproduce the issue anymore that led to #3150).

HaloFour:
@trask 

The Reactor instrumentation for `Mono<T>` and `Flux<T>` do handle cancellation, although using a different approach:  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorAsyncOperationEndStrategy.java#L64

I also have another [PR](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3751) up which refactors the Reactor instrumentation to allow for multiple subscriptions and in that implementation I follow a similar approach to the PR you referenced: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3751/files#diff-0865720f5bd8562039a4e795c46fd7f745a94a5ba6852ff463ffa0d0424fa236R65

trask:
putting this PR on hold until #3816

PR #3802, opened 2021-08-09
Refactor couchbase and hystrix to Instrumenter API

mateuszrzeszutek:
Part of #2713

Both Hystrix and Couchbase instrumentations make use of the rxjava 1.0 library instrumentation, so I had to refactor all three at once.


instrumentation/rxjava/rxjava-1.0/library/src/main/java/io/opentelemetry/instrumentation/rxjava/TracedOnSubscribe.java
```diff
@@ -5,42 +5,44 @@
...
-      delegate.call(new TracedSubscriber<>(context, subscriber, tracer));
+    /*
+    TODO: can't really call shouldStart() - couchbase async instrumentation nests CLIENT calls
+    which normally should happen in a sequence
+    InstrumentationTypes to the rescue?
+
+    if (!instrumenter.shouldStart(parentContext, request)) {
+      delegate.call(subscriber);
+      return;
     }
```
trask, line 40: cc: @lmolkova

trask (approved):
nice

instrumentation/couchbase/couchbase-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_0/CouchbaseRequest.java
```diff
@@ -0,0 +1,42 @@
...
+import javax.annotation.Nullable;
+
+@AutoValue
+public abstract class CouchbaseRequest {
+
+  public static CouchbaseRequest create(@Nullable String bucket, Method method) {
+    Class<?> declaringClass = method.getDeclaringClass();
+    String className =
+        declaringClass.getSimpleName().replace("CouchbaseAsync", "").replace("DefaultAsync", "");
+    String operation = className + "." + method.getName();
```
anuraaga, line 20: We should cache this, maybe in another PR

anuraaga (approved):

PR #3806, opened 2021-08-09
Add a few missing @Nullables in instrumentation-api

mateuszrzeszutek:
A little bit less warnings from IntelliJ

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
@@ -12,6 +12,7 @@ import io.opentelemetry.instrumentation.api.config.Config;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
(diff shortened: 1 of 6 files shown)


trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
```diff
@@ -68,6 +69,7 @@ public static void debugContextLeakIfEnabled() {
     }
   }
 
+  @Nullable
```
anuraaga, line 72: Is it possible to return an empty list instead of null?

anuraaga (approved):

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/ContextPropagationDebug.java
mateuszrzeszutek: It is, but it'd result in a bit more code being added - and since it's a private, internal utility method I think it's fine to keep `@Nullable` here.

PR #3809, opened 2021-08-10
Limit netty exception capture to netty spans

trask:
The existing code recorded exception on `Span.current()` sometimes, which could be from a higher level CLIENT span (e.g. a database client span) which handles its own (higher level) exception/status tracking.


iNikem (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AbstractChannelHandlerContextInstrumentation.java
```diff
@@ -46,15 +44,14 @@ public void transform(TypeTransformer transformer) {
...
-              .onException(Java8BytecodeBridge.currentContext(), throwable);
-        }
+      if (channelContext.channel().hasAttr(AttributeKeys.CLIENT_CONTEXT)) {
+        Attribute<Context> clientContextAttr =
+            channelContext.channel().attr(AttributeKeys.CLIENT_CONTEXT);
+        NettyHttpClientTracer.tracer().endExceptionally(clientContextAttr.get(), throwable);
+      } else if (channelContext.channel().hasAttr(AttributeKeys.SERVER_CONTEXT)) {
+        Attribute<Context> clientContextAttr =
+            channelContext.channel().attr(AttributeKeys.SERVER_CONTEXT);
+        NettyHttpClientTracer.tracer().onException(clientContextAttr.get(), throwable);
```
mateuszrzeszutek, line 54:
```suggestion
        Attribute<Context> serverContextAttr =
            channelContext.channel().attr(AttributeKeys.SERVER_CONTEXT);
        NettyHttpClientTracer.tracer().onException(serverContextAttr.get(), throwable);
```

mateuszrzeszutek (approved):

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/AbstractChannelHandlerContextInstrumentation.java
```diff
@@ -46,15 +44,14 @@ public void transform(TypeTransformer transformer) {
...
-          NettyHttpClientTracer.tracer().endExceptionally(clientContextAttr.get(), throwable);
-        } else {
-          NettyHttpServerTracer.tracer()
-              .onException(Java8BytecodeBridge.currentContext(), throwable);
-        }
+      if (channelContext.channel().hasAttr(AttributeKeys.CLIENT_CONTEXT)) {
+        Attribute<Context> clientContextAttr =
+            channelContext.channel().attr(AttributeKeys.CLIENT_CONTEXT);
+        NettyHttpClientTracer.tracer().endExceptionally(clientContextAttr.get(), throwable);
+      } else if (channelContext.channel().hasAttr(AttributeKeys.SERVER_CONTEXT)) {
```
laurit, line 51: Netty40 has exactly the same logic, could you change that also.

trask: good call

PR #3811, opened 2021-08-10
Make spring webflux tests pass with strict context check

laurit:


instrumentation/spring/spring-webflux-5.0/javaagent/src/test/groovy/util/SpringWebfluxTestUtil.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+package util
+
+import static org.awaitility.Awaitility.await
+
+import java.util.concurrent.TimeUnit
+
+class SpringWebfluxTestUtil {
+
+  static waitForRequestsToComplete() {
```
trask, line 14: can you add a comment here explaining why we are doing this?

trask (approved):

mateuszrzeszutek (approved):

iNikem:
Do we actually run with `enableStrictContext=true`? I was unable to find where we set it to true. Default value is false, right?
```
rootProject.findProperty("enableStrictContext") ?: false
```

iNikem:
So after this PR it is still disabled? Unless somebody manually runs it with strict enabled?

laurit:
@iNikem that is also correct. Besides this one there are 2 more usages of `enableStrictContext=false` when those get cleared up then perhaps we can attempt enabling strict context check again.

PR #3812, opened 2021-08-10
Implement a Call.Factory for okhttp 3.x+ library instrumentation

jkwatson:
This now actually will handle async context connections correctly!


instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @Nullable
+  static Context getCallingContextForRequest(Request request) {
+    return contextsByRequest.get(request);
+  }
+
+  @Override
+  public Call newCall(Request request) {
+    Context callingContext = Context.current();
+    contextsByRequest.put(request, callingContext);
```
jkwatson, line 32: rather than trying to detect whether there is the right kind of `tag` method, it's definitely simpler to just use the cache here always.

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    public Call clone() throws CloneNotSupportedException {
+      try {
+        Method clone = delegate.getClass().getDeclaredMethod("clone");
+        return new TracingCall((Call) clone.invoke(delegate), Context.current());
+      } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+        return (Call) super.clone();
+      }
```
jkwatson, line 57: Does this seem right? It makes me sad that they added a `clone` to the http call. :(

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingInterceptor.java
```diff
@@ -26,8 +26,11 @@
 
   @Override
   public Response intercept(Chain chain) throws IOException {
-    Context parentContext = Context.current();
     Request request = chain.request();
+    Context parentContext = TracingCallFactory.getCallingContextForRequest(request);
+    if (parentContext == null) {
+      parentContext = Context.current();
```
jkwatson, line 32: does this seem like the right default, or should we not try to propagate at all if we don't have one in the cache?

instrumentation/okhttp/okhttp-3.0/testing/src/main/groovy/io/opentelemetry/instrumentation/okhttp/v3_0/AbstractOkHttp3Test.groovy
```diff
@@ -45,7 +45,10 @@ abstract class AbstractOkHttp3Test extends HttpClientTest<Request> {
 
   @Override
   int sendRequest(Request request, String method, URI uri, Map<String, String> headers) {
-    return client.newCall(request).execute().code()
+    def response = client.newCall(request).execute()
+    response.body().withCloseable {
+      return response.code()
+    }
```
jkwatson, line 51: it's really important to close the response body *always* when using okhttp, otherwise we're leaking responses in our test code.

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
```diff
@@ -0,0 +1,121 @@
...
+    @Override
+    public void cancel() {
+      delegate.cancel();
+    }
+
+    @Override
+    public Call clone() throws CloneNotSupportedException {
+      try {
+        Method clone = delegate.getClass().getDeclaredMethod("clone");
+        return new TracingCall((Call) clone.invoke(delegate), Context.current());
```
anuraaga, line 54: We should probably have a comment for the use of `Context.current()`. Brave seems to pick the call's context, but I agree with you that current makes more sense (clone is effectively newCall)

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingInterceptor.java
anuraaga: Let's not propagate at all

instrumentation/okhttp/okhttp-3.0/library/src/test/groovy/io/opentelemetry/instrumentation/okhttp/v3_0/OkHttp3Test.groovy
```diff
@@ -5,18 +5,16 @@
...
-import okhttp3.Dispatcher
+import okhttp3.Call
 import okhttp3.OkHttpClient
 
 class OkHttp3Test extends AbstractOkHttp3Test implements LibraryTestTrait {
+
   @Override
-  OkHttpClient.Builder configureClient(OkHttpClient.Builder clientBuilder) {
-    return clientBuilder
-      // The double "new Dispatcher" style is the simplest way to decorate the default executor.
```
anuraaga, line 17: Well it's also nice for this to be gone :)

instrumentation/okhttp/okhttp-3.0/testing/src/main/groovy/io/opentelemetry/instrumentation/okhttp/v3_0/AbstractOkHttp3Test.groovy
anuraaga: 😱

instrumentation/okhttp/okhttp-3.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/okhttp/v3_0/OkHttp3Test.groovy
```diff
@@ -6,12 +6,37 @@
...
 class OkHttp3Test extends AbstractOkHttp3Test implements AgentTestTrait {
+
   @Override
-  OkHttpClient.Builder configureClient(OkHttpClient.Builder clientBuilder) {
-    return clientBuilder
+  Call.Factory createCallFactory(OkHttpClient.Builder clientBuilder) {
+    return clientBuilder.build()
+  }
+
+  def "reused builder has one interceptor"() {
```
anuraaga, line 23: Good move

anuraaga:
For the record, we see a similar issue as aws sdk instrumentation, where muzzle is forcing reflection where we otherwise wouldn't need it. An annotation approach to solving this seems like a nice Splunk intern project :-) (guess it's too late this year...)

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
anuraaga: Nice catch I guess we need to build a copy of the Request even when not using the `tag` method.

jkwatson:
One downside to this approach (not that there's a better one) is that libraries that only accept an actual OkHttpClient as a dependency won't be able to consume the Call.Factory and use it. I guess if users run into that, we can create issues in the relevant library and submit fixes. :)

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
jkwatson: fixed

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingInterceptor.java
anuraaga: Ah ok

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
```diff
@@ -0,0 +1,154 @@
...
+import okhttp3.Response;
+import okio.Timeout;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+class TracingCallFactory implements Call.Factory {
+  private static final Cache<Request, Context> contextsByRequest =
+      Cache.newBuilder().setWeakKeys().build();
+
+  @Nullable private static Method timeoutMethod;
+  @Nullable private static Method cloneMethod;
```
mateuszrzeszutek, line 27: Can we use `MethodHandle`s instead of reflection?

mateuszrzeszutek (approved):

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
jkwatson: I'll take a look. I've been meaning to learn how to use them, so this is a good chance!

jkwatson: done. please take a look because I'm not sure how to test that it works without using reflection.

instrumentation/okhttp/okhttp-3.0/library/README.md
```diff
@@ -0,0 +1,57 @@
+# Manual Instrumentation for OkHttp3 version 3.0.0+
+
+Provides OpenTelemetry instrumentation for [okhttp3](https://square.github.io/okhttp/).
+
+## Quickstart
+
+### Add these dependencies to your project:
+
+Replace `OPENTELEMETRY_VERSION` with the latest stable
+[release](https://mvnrepository.com/artifact/io.opentelemetry). `Minimum version: 1.4.0`
```
trask, line 10:
```suggestion
[release](https://mvnrepository.com/artifact/io.opentelemetry). `Minimum version: 1.5.0`
```

instrumentation/okhttp/okhttp-3.0/library/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+public class OkHttpConfiguration {
+
+  //Use this Call.Factory implementation for making standard http client calls.
+  public Call.Factory createTracedClient(OpenTelemetry openTelemetry) {
+    return OkHttpTracing.newBuilder(openTelemetry).build().newCallFactory(createClient());
+  }
+
+  //your configuration of the OkHttpClient goes here:
+  public OkHttpClient createClient() {
```
trask, line 53: maybe?
```suggestion
  private OkHttpClient createClient() {
```

instrumentation/okhttp/okhttp-3.0/library/src/test/groovy/io/opentelemetry/instrumentation/okhttp/v3_0/OkHttp3Test.groovy
```diff
@@ -25,8 +23,4 @@ class OkHttp3Test extends AbstractOkHttp3Test implements LibraryTestTrait {
     false
   }
 
-  @Override
-  boolean testCausalityWithCallback() {
-    false
-  }
```
trask, line 31: 🥳

trask (approved):

instrumentation/okhttp/okhttp-3.0/library/README.md
jkwatson: sure. doesn't matter too much for some sample code, but I'm ok with it. :)

jkwatson:
any idea what we need to do to get this thing to finish the checks?

mateuszrzeszutek:
Closed and reopened -- hope they'll get triggered now.

PR #3815, opened 2021-08-10
Overhead benchmarks: Perform a warmup phase.

### LINKED ISSUE #3687: Allow overhead tests to specify a warm-up period
In #3627, it was noted that we may wish to have some tests run after a warm-up phase. That is, we don't want to actually start the measurements (with JFR) until after some warm-up cycle so that the jit compilation has had a chance to do its thing and the app has reached more of a steady-state.

As noted in #3627, there are pros and cons of doing this. On one hand, will miss the overhead introduced by the agent during startup, but for many users (especially those with long-running enterprise JVM applications) they may only care about steady-state overhead.

We should modify the overhead test configuration to allow for an optional warm-up cycle, specified in seconds, during which a moderate amount of traffic should be generated. Only after the startup has completed should the relevant measurements begin.  With JFR, this can be accomplished with the `delay` parameter (dynamically configured on the commandline via system props).

breedx-splk:
This resolves #3687.

After the petclinic app is healthy, we do a configurable warmup phase where we send some meager load while the app gets warm. After the warmup is expired, THEN we start recording JFR.

Previously, JFR was just enabled on startup and the recording started at JVM launch. There were some challenges around changing this to be started later. There were several options:

1. use the `delay=n` JFR parameter to delay recording start. This approach is very hard to synchronize with because we don't know how long it takes for the app to become healthy prior to starting our warmup period. It's just not really compatible with an inconsistent start-to-healthy time.
2. use `jcmd` in the container to start JFR. Unfortunately, to reduce container size, we chose to use the JRE instead of JDK....and the JRE does not include `jcmd`. I tried copying the command in, but it relies on a modular JVM and that module is also missing from JRE.  We could have switched to the JDK and bloated the container.
3. Start the recording via JMX. This is pretty common in the wild, so I went with this approach. HOWEVER, because we talk JMX over RMI, and the fact that RMI has a goofy handshake mechanism that relies on knowable hostnames, it is challenging to to JMX into the container from outside the network. The workaround was to build a tiny little app that all it does is turn on a recording...and we run that app from inside the docker composed network environment.

Curious what people feel about the approach. Feels hacky, but seemed like a reasonable workaround given the other constraints.


trask:
+1 for option 2, I don't think smaller test container size is worth additional complexity

iNikem:
Regarding image sizes (on my machine):
```
adoptopenjdk                      11-jre     244MB
adoptopenjdk                      11-jdk     437MB
bellsoft/liberica-openjdk-alpine  11         131MB
```

```
docker run bellsoft/liberica-openjdk-alpine:11 jcmd 
1 jdk.jcmd/sun.tools.jcmd.JCmd
```

breedx-splk:
Thanks for the input. I switched to the jdk and it simplifies things. We will need to build a new base image, but there's an action for it.

benchmark-overhead/src/test/java/io/opentelemetry/containers/PetClinicRestContainer.java
```diff
@@ -27,6 +27,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(PetClinicRestContainer.class);
   private static final int PETCLINIC_PORT = 9966;
+  public static final int PETCLINIC_JMX_PORT = 9000;
```
jkwatson, line 30: do we need JMX with jmd available now?

iNikem:
> Thanks for the input. I switched to the jdk and it simplifies things. We will need to build a new base image, but there's an action for it.

Do you care to switch to `bellsoft/liberica-openjdk-alpine` for almost 3x decrease in base image size?

benchmark-overhead/src/test/java/io/opentelemetry/containers/PetClinicRestContainer.java
breedx-splk: nope, I can take that back out.

breedx-splk:
> > Thanks for the input. I switched to the jdk and it simplifies things. We will need to build a new base image, but there's an action for it.
> 
> Do you care to switch to `bellsoft/liberica-openjdk-alpine` for almost 3x decrease in base image size?

If other people like this then I'm happy to make the change. I tried it locally and it works at least...but I'm not sure about running the benchmarks on a [much less popular distro](https://snyk.io/jvm-ecosystem-report-2021/). What does the crowd here think?

iNikem:
I will be surprised a lot, if different OpenJDK Hotspot distros will demonstrate different performance in our tests.

trask (approved):
👍

benchmark-overhead/Dockerfile-petclinic-base
```diff
@@ -10,6 +10,9 @@ RUN git clone http://github.com/spring-petclinic/spring-petclinic-rest.git
 WORKDIR /app/spring-petclinic-rest
 RUN ./mvnw package -Dmaven.test.skip=true
 RUN cp target/spring-petclinic-rest*.jar /app/spring-petclinic-rest.jar
+
+FROM bellsoft/liberica-openjdk-alpine:11
```
iNikem, line 14: I think you can build with Liberica as well. I do :)

iNikem (approved):

benchmark-overhead/Dockerfile-petclinic-base
breedx-splk: Maybe, but it doesn't have `git` and I have no idea how to reasonably install packages in there without making it a bunch more complicated.

PR #3816, opened 2021-08-11
Simplify webflux mono handling

trask:
Splitting the more problematic (and independent) part of #3798 out into a separate PR for better scrutiny/history.

Based on @HaloFour's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3798#issuecomment-896432083, I realized that the change in #3789 that seems to fix the flakiness that we previously fixed via #3150, is not the change in instrumentation placement, but the registeration of a cancellation handler using Mono's `doOnCancel`.

### DIFF
--- a/instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
+++ b/instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
@@ -30,7 +30,8 @@ public class AdviceUtils {
 
   public static <T> Mono<T> setPublisherSpan(Mono<T> mono, Context context) {
     return mono.doOnError(t -> finishSpanIfPresent(context, t))
-        .doOnTerminate(() -> finishSpanIfPresent(context, null));
+        .doOnSuccess(x -> finishSpanIfPresent(context, null))
+        .doOnCancel(() -> finishSpanIfPresent(context, null));
   }
 
   public static void finishSpanIfPresent(ServerWebExchange exchange, Throwable throwable) {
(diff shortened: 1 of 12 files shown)


mateuszrzeszutek (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
```diff
@@ -35,20 +28,10 @@ public static String spanNameForHandler(Object handler) {
...
-   * Idea for this has been lifted from https://github.com/reactor/reactor-core/issues/947. Newer
-   * versions of reactor-core have easier way to access context but we want to support older
-   * versions.
-   */
-  public static <T> Function<? super Publisher<T>, ? extends Publisher<T>> finishSpanNextOrError(
-      io.opentelemetry.context.Context context) {
-    return Operators.lift(
-        (scannable, subscriber) -> new SpanFinishingSubscriber<>(subscriber, context));
+  public static <T> Mono<T> setPublisherSpan(Mono<T> mono, Context context) {
+    return mono.doOnError(t -> finishSpanIfPresent(context, t))
```
anuraaga, line 32: Any thoughts on `doOnError` only setting the exception and `doOnTerminate` finishing the span?

anuraaga (approved):

instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/server/AdviceUtils.java
trask: that doesn't really fit the Instrumenter model though, of calling end() and passing the exception?

trask: (Instrumenter coming in #3798)

anuraaga: Ah that makes sense too

laurit (approved):

PR #3818, opened 2021-08-11
Bump gradle-plugins to 0.4 (and fix one more thing in ReferencesPrinter)

mateuszrzeszutek:


muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferencesPrinter.java
```diff
@@ -26,7 +26,7 @@
    */
   public static void printMuzzleReferences() {
     for (InstrumentationModule instrumentationModule :
-        ServiceLoader.load(InstrumentationModule.class)) {
+        ServiceLoader.load(InstrumentationModule.class, ReferencesPrinter.class.getClassLoader())) {
```
iNikem, line 29: Why does this differ from the current context classloader?

mateuszrzeszutek: That's what surprised me too a bit - apparently the current context CL is still the gradle CL; while we want to use the instrumentation CL for everything that happens in this class.

iNikem: Should we do similar to https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/92a69c33097c7fc0d3d2453205c2f44c1289fbda/gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts#L219 ?

mateuszrzeszutek: Done.

gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
```diff
@@ -336,3 +334,13 @@ fun muzzleDirectiveToArtifacts(muzzleDirective: MuzzleDirective, system: Reposit
 
   yieldAll(allVersionArtifacts)
 }
+
+fun withContextClassLoader(classLoader: ClassLoader, func: () -> Unit) {
```
iNikem, line 338: Rename to `action`?

iNikem (approved):

trask (approved):

PR #3822, opened 2021-08-11
Enable strict context check in tests by default

trask:
Based on @iNikem's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3811#issuecomment-896628035

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -56,13 +56,6 @@ jobs:
         vm:
           - hotspot
           - openj9
-        # THIS IS TEMPORARY TO FLUSH OUT SPORADIC FAILURES
-        count:
-          - 1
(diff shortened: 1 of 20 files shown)


trask:
will wait until after 1.5.0 to mark this ready for review, since it will likely introduce some sporadic failures until we apply `enableStrictContext=false` to them

trask:
Got a green build.. planning to re-run several times to try to iron out other sporadic failures this may introduce..

mateuszrzeszutek (approved):
WDYT about creating a tracking issue for all those modules that have this check disabled?

laurit (approved):

trask:
> WDYT about creating a tracking issue for all those modules that have this check disabled?

👍 created #3916

Issue #766, opened 2020-07-23
Support looking up a `ContextStore` from outside of Advice


trask:
I haven't thought through this, but I think it should be possible to add support for this.

It would be one less confusing thing about writing instrumentation.

And it would make the OpenTelemetry API bridge slightly more efficient instead of having to pass around the ContextStore from the Advice to the Span.Builder where it is needed.

See https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/746#discussion_r458511013 for some background.

PR #3827, opened 2021-08-12
Support looking up a ContextStore from outside of Advice

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/766
Calling agent from instrumentation is implemented the same way as in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/OpenTelemetrySdkAccess.java

### DIFF
--- a/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
+++ b/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
@@ -34,7 +34,7 @@ public class InstrumentationContext {
   public static <Q extends K, K, C> ContextStore<Q, C> get(
       Class<K> keyClass, Class<C> contextClass) {
     if (contextStoreSupplier == null) {
-      throw new IllegalStateException("");
+      throw new IllegalStateException("Context store supplier not set");
     }
     return contextStoreSupplier.get(keyClass, contextClass);
   }
--- a/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
+++ b/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
@@ -18,7 +18,12 @@ public class InstrumentationContext {
    * <p>In reality, the <em>calls</em> to this method are re-written to something more performant
    * while injecting advice into a method.
    *
-   * <p>This method must only be called within an Advice class.
+   * <p>When this method is called from outside of an Advice class it can only access {@link
+   * ContextStore} when it is already created. For this {@link ContextStore} either needs to be
+   * added to {@code
+   * io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule#getMuzzleContextStoreClasses()}
+   * or be used in an Advice class which automatically adds it to {@code
+   * InstrumentationModule#getMuzzleContextStoreClasses()}
    *
    * @param keyClass The key class context is attached to.
    * @param contextClass The context class attached to the user class.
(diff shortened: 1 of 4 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -65,6 +67,8 @@ public static synchronized void installAgentTracer(Config config) {
             });
       }
 
+      InstrumentationContext.internalSetContextStoreSupplier(
+          (keyClass, contextClass) -> FieldBackedProvider.getContextStore(keyClass, contextClass));
```
mateuszrzeszutek, line 71: I'd probably put that somewhere in `InstrumentationLoader` where `InstrumentationContextProvider` implementation gets chosen. Maybe that `getContextStore()` method should be a part of that provider interface?

javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
```diff
@@ -18,7 +18,12 @@ private InstrumentationContext() {}
    * <p>In reality, the <em>calls</em> to this method are re-written to something more performant
    * while injecting advice into a method.
    *
-   * <p>This method must only be called within an Advice class.
+   * <p>When this method is called from outside of an Advice class it can only access {@link
+   * ContextStore} when it is already created. For this {@link ContextStore} either needs to be
+   * added to {@code
+   * io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule#getMuzzleContextStoreClasses()}
+   * or be used in an Advice class which automatically adds it to {@code
+   * InstrumentationModule#getMuzzleContextStoreClasses()}
```
mateuszrzeszutek, line 26: The muzzle codegen plugin collects invocations from all helper classes, not only advices -- at least it should work that way.

javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
```diff
@@ -28,7 +33,27 @@ private InstrumentationContext() {}
    */
   public static <Q extends K, K, C> ContextStore<Q, C> get(
       Class<K> keyClass, Class<C> contextClass) {
-    throw new IllegalStateException(
-        "Calls to this method will be rewritten by Instrumentation Context Provider (e.g. FieldBackedProvider)");
+    if (contextStoreSupplier == null) {
+      throw new IllegalStateException("");
+    }
+    return contextStoreSupplier.get(keyClass, contextClass);
```
mateuszrzeszutek, line 39: I wonder if we still need to replace `InstrumentationContext.get()` calls in `FieldBackedProvider` - if the provider idea works for helper classes, wouldn't it work everywhere? Removing that part would certainly make `FieldBackedProvider` a bit simpler.

mateuszrzeszutek (commented):
That's a really cool idea 🚀 
Left a few comments, maybe we can simplify the `InstrumentationContextProvider` a bit more.

javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
laurit: I believe that it is better to keep replacing `InstrumentationContext.get()` calls in `FieldBackedProvider` because this `InstrumentationContext.get()` implementation isn't really efficient. It involves `Class.forName` and `Class.getMethod` which is fine in helper classes where you can do the lookup once and keep `ContextStore` in a field.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
laurit: Moved to `InstrumentationModuleInstaller`

mateuszrzeszutek (approved):

Issue #3826, opened 2021-08-11
jetty-httpclient erroneously returns an empty response when instrumented


seshness:
**Describe the bug**
When I make a request with `jetty-httpclient`, with instrumentation applied, the response is empty.
When I make the same request without instrumentation, the response is non-empty as expected.

**Steps to reproduce**
1. Create a project with `jetty-httpclient`
2. Run this test with the jvm argument `-javaagent:<path to opentelemetry-javaagent-all.jar>`
```
import org.eclipse.jetty.client.HttpClient;
import org.eclipse.jetty.client.api.ContentResponse;
import org.junit.jupiter.api.Test;

public class JettyClientTest {
    @Test
    public void testJettyResponse() throws Exception {
        HttpClient httpClient = new HttpClient();
        httpClient.start();

        ContentResponse response = httpClient.GET("http://google.com");
        assert response.getContentAsString().length() != 0;
    }
}
```

**What did you expect to see?**
When I make this request with `curl` or without instrumentation I see a non-empty response:
```
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
```

**What did you see instead?**
With instrumentation enabled the response is empty. `response.getContentAsString()` produces an empty string.

**What version are you using?**
`org.eclipse.jetty:jetty-client:9.4.40.v20210413`
`opentelemetry-javaagent-all.jar` at v1.4.1

**Environment**
Compiler: AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OS: macOS 10.15.7
Runtime (if different from JDK above): AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OS (if different from OS compiled on): macOS 10.15.7

**Additional context**
I tried reproducing with a few other `jetty-httpclient` versions too:
| jetty-httpclient version | works as expected? |
| ----------------------- | -------------------- |
| 9.4.40.v20210413          | ❌                               |
| 9.4.43.v20210629         | ❌                               |
| 10.0.6                              | ✅                               |
| 11.0.6                              | ✅                               |

trask:
cc: @robododge

robododge:
Hi @seshness, thanks for pointing this out.

[...]

PR #3831, opened 2021-08-12
Improve wrapping jetty response listener

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3826
Jetty response listener can implement multiple interfaces, we need to ensure that after wrapping it implements the same interfaces as it originally did.


iNikem (approved):

mateuszrzeszutek (approved):

anuraaga (approved):
Thanks!

robododge:
This is great, thanks for correcting the Response listener wrapping.  Now the Request listener wrapping needs the same treatment.  Follow up to fix the Request-side listeners in #3833

PR #3832, opened 2021-08-12
Benchmark overhead: Accumulate and persist results

breedx-splk:
This will run nightly and persist the results back into the `/benchmark-overhead/results` directory.

One subdirectory per test configuration. Each directory gets a summary file for the last run and a csv with all the details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
+      - 'benchmark-overhead/**'
   workflow_dispatch:
 
 jobs:
--- a/benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
+++ b/benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package io.opentelemetry.results;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Comparator;
+import java.util.List;
+import java.util.function.Consumer;
+
+class CsvPersister implements ResultsPersister {
+
+  private final Path resultsFile;
+
+  public CsvPersister(Path resultsFile) {this.resultsFile = resultsFile;}
+
+  @Override
+  public void write(List<AppPerfResults> results) {
+
+    ensureFileCreated(results);
+
+    StringBuffer sb = new StringBuffer().append(System.currentTimeMillis() / 1000);
+    doSorted(results, result -> {
+      sb.append(",").append(result.startupDurationMs);
+      sb.append(",").append(result.heapUsed.min);
+      sb.append(",").append(result.heapUsed.max);
+      sb.append(",").append(result.getTotalAllocatedMB());
+      sb.append(",").append(result.totalGCTime);
+      sb.append(",").append(result.maxThreadContextSwitchRate);
+      sb.append(",").append(result.iterationAvg);
+      sb.append(",").append(result.iterationP95);
+      sb.append(",").append(result.requestAvg);
+      sb.append(",").append(result.requestP95);
+      sb.append(",").append(result.peakThreadCount);
+    });
+    sb.append("\n");
+    try {
+      Files.writeString(resultsFile, sb.toString(), StandardOpenOption.APPEND);
+    } catch (IOException e) {
+      throw new RuntimeException("Error writing csv content", e);
+    }
+  }
+
+  private void ensureFileCreated(List<AppPerfResults> results) {
+    if (Files.exists(resultsFile)) {
+      return;
+    }
+    StringBuffer sb = new StringBuffer("timestamp");
+    doSorted(results, result -> {
+      String agent = result.getAgentName();
+      sb.append(",").append(agent).append(":startupTimeMs");
+      sb.append(",").append(agent).append(":minHeapUsed");
+      sb.append(",").append(agent).append(":maxHeapUsed");
+      sb.append(",").append(agent).append(":totalAllocatedMB");
+      sb.append(",").append(agent).append(":totalGCTime");
+      sb.append(",").append(agent).append(":maxThreadContextSwitchRate");
+      sb.append(",").append(agent).append(":iterationAvg");
+      sb.append(",").append(agent).append(":iterationP95");
+      sb.append(",").append(agent).append(":requestAvg");
+      sb.append(",").append(agent).append(":requestP95");
+      sb.append(",").append(agent).append(":peakThreadCount");
+    });
+    sb.append("\n");
+    try {
+      Files.writeString(resultsFile, sb.toString());
+    } catch (IOException e) {
+      throw new RuntimeException("Error creating csv output stub", e);
+    }
+  }
+
+  private void doSorted(List<AppPerfResults> results, Consumer<AppPerfResults> consumer) {
+    results.stream()
+        .sorted(Comparator.comparing(AppPerfResults::getAgentName))
+        .forEach(consumer);
+  }
+}
--- a/benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
+++ b/benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
@@ -49,6 +49,15 @@ class CsvPersister implements ResultsPersister {
     if (Files.exists(resultsFile)) {
       return;
     }
+    try {
+      String headerLine = createHeaderLine(results);
+      Files.writeString(resultsFile, headerLine);
+    } catch (IOException e) {
+      throw new RuntimeException("Error creating csv output stub", e);
+    }
+  }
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    paths-ignore:
```
jkwatson, line 7: 👍🏽

jkwatson (approved):

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
```diff
@@ -0,0 +1,80 @@
...
+      sb.append(",").append(result.getTotalAllocatedMB());
+      sb.append(",").append(result.totalGCTime);
+      sb.append(",").append(result.maxThreadContextSwitchRate);
+      sb.append(",").append(result.iterationAvg);
+      sb.append(",").append(result.iterationP95);
+      sb.append(",").append(result.requestAvg);
+      sb.append(",").append(result.requestP95);
+      sb.append(",").append(result.peakThreadCount);
+    });
+    sb.append("\n");
```
iNikem, line 40: Only one line in the result file?

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
```diff
@@ -0,0 +1,80 @@
...
+      sb.append(",").append(agent).append(":requestAvg");
+      sb.append(",").append(agent).append(":requestP95");
+      sb.append(",").append(agent).append(":peakThreadCount");
+    });
+    sb.append("\n");
+    try {
+      Files.writeString(resultsFile, sb.toString());
+    } catch (IOException e) {
+      throw new RuntimeException("Error creating csv output stub", e);
+    }
```
iNikem, line 72: Please extract this to something a la `writeHeader` or something.

And it now seems that it is indeed expected to write one long line. Please add some comment about that. I bet everybody will be confused that a list of items is all written as one single line.

benchmark-overhead/src/test/java/io/opentelemetry/results/FileSummaryPersister.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+class FileSummaryPersister implements ResultsPersister {
+
+  private final Path file;
+  public FileSummaryPersister(Path file) {this.file = file;}
+
+  @Override
+  public void write(List<AppPerfResults> results) {
+    try {
+      PrintStream out = new PrintStream(file.toFile());
```
iNikem, line 23: Shouldn't you close `PrintStream`?

benchmark-overhead/src/test/java/io/opentelemetry/results/PrintStreamPersister.java
```diff
@@ -0,0 +1,59 @@
...
+    display(results, "Iter. mean", res -> format(res.iterationAvg));
+    display(results, "Iter. p95", res -> format(res.iterationP95));
+    display(results, "Peak threads", res -> String.valueOf(res.peakThreadCount));
+  }
+
+  private void display(List<AppPerfResults> results, String pref,
+      Function<AppPerfResults, String> vs) {
+    out.printf("%-20s: ", pref);
+    results.stream()
+        .sorted(Comparator.comparing(AppPerfResults::getAgentName))
```
iNikem, line 48: Instead of sorting it for each metric, how about sorting it once at the beginning of `write`?

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
```diff
@@ -0,0 +1,80 @@
...
+  private final Path resultsFile;
+
+  public CsvPersister(Path resultsFile) {this.resultsFile = resultsFile;}
+
+  @Override
+  public void write(List<AppPerfResults> results) {
+
+    ensureFileCreated(results);
+
+    StringBuffer sb = new StringBuffer().append(System.currentTimeMillis() / 1000);
```
laurit, line 26: you probably meant to use `StringBuilder`

benchmark-overhead/src/test/java/io/opentelemetry/results/FileSummaryPersister.java
breedx-splk: Yeah, will use autocloseable. 👍🏻

benchmark-overhead/src/test/java/io/opentelemetry/results/PrintStreamPersister.java
breedx-splk: Sure, can do.

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
breedx-splk: I did.  :) Thank you.

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
breedx-splk: One header line, and then one line per run. It appends over time.

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
breedx-splk: Factoring out the named method makes sense.
The long line shouldn't really be surprising for a csv file tho should it?

iNikem: when I see a loop inside "write to CSV" method, my first expectation is that each loop iteration writes one line. I bet I am not the only one.

breedx-splk: 🤷🏻 Ok I'll add a comment.

trask:
@anuraaga asked yesterday why not commit results to `gh-pages` branch, and I couldn't remember if there was reason?

iNikem:
Why committing results to a separate branch is good?

benchmark-overhead/src/test/java/io/opentelemetry/results/CsvPersister.java
```diff
@@ -0,0 +1,93 @@
...
+      sb.append(",").append(result.maxThreadContextSwitchRate);
+      sb.append(",").append(result.iterationAvg);
+      sb.append(",").append(result.iterationP95);
+      sb.append(",").append(result.requestAvg);
+      sb.append(",").append(result.requestP95);
+      sb.append(",").append(result.peakThreadCount);
+    });
+    sb.append("\n");
+    try {
+      Files.writeString(resultsFile, sb.toString(), StandardOpenOption.APPEND);
```
mateuszrzeszutek, line 46: Instead of opening the result file twice and aggregating the results in string builders, why not use the `Writer` for that?

breedx-splk: I like that it consolidates the mandatory exception handling to a single line instead of spreading it across multiple...but there is more than one way to skin a 🐱 .

breedx-splk:
> @anuraaga asked yesterday why not commit results to `gh-pages` branch, and I couldn't remember if there was reason?

We can easily _also_ push the results to `gh-pages`, but since we're appending and growing the file, we would need to also check out the `gh-pages` branch in order to append. There has to be a single origin "main" branch, and since we're building from `main` it made sense and I think is less work than trying to coordinate.

If folks prefer that the results originate and push to `gh-pages` branch, I can probably do that but it'll take some fiddling.

iNikem:
I probably miss something, but I don't understand at all where this separate branch came from and why

breedx-splk:
> I probably miss something, but I don't understand at all where this separate branch came from and why

When we want to build a GitHub Pages hosted GUI on it (similar to iguanodon), then having the results in the `gh-pages` is nice and makes the UI play nicer with the data. That's the only reason I can see for doing that...but we still have to have one true source.

iNikem:
> > I probably miss something, but I don't understand at all where this separate branch came from and why
> 
> When we want to build a GitHub Pages hosted GUI on it (similar to iguanodon), then having the results in the `gh-pages` is nice and makes the UI play nicer with the data. That's the only reason I can see for doing that...but we still have to have one true source.

In my usual manner, I will say that this is a feature creep and should be addressed separately when and if we start doing GH Pages. Not in this PR :)

iNikem (approved):

anuraaga:
I thought we were aiming to publish to gh-pages since I saw iguanadon :P But if not for now that's cool. I think pushing to main might not work anyways though since it's a protected branch - we never found a way to push docs updates to main from the workflow in the SDK repo. This was a few months ago but if it's still an issue, then gh-pages might still be the easier approach strangely enough.

mateuszrzeszutek (approved):

trask:
> I thought we were aiming to publish to gh-pages since I saw iguanadon :P But if not for now that's cool. I think pushing to main might not work anyways though since it's a protected branch - we never found a way to push docs updates to main from the workflow in the SDK repo. This was a few months ago but if it's still an issue, then gh-pages might still be the easier approach strangely enough.

Discussed on SIG call, this makes sense, will address in follow-up PR.

PR #3837, opened 2021-08-13
Convert tapestry-5.4 to instrumenter api

laurit:
Part of #2713


instrumentation/tapestry-5.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tapestry/TapestryServerSpanNaming.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public static void updateServerSpanName(String pageName) {
+    Context parentContext = Context.current();
+    ServerSpanNaming.updateServerSpanName(
+        parentContext, CONTROLLER, getServerSpanNameSupplier(parentContext, pageName));
+  }
+
+  private static Supplier<String> getServerSpanNameSupplier(Context context, String pageName) {
+    return () -> getServerSpanName(context, pageName);
+  }
```
trask, line 25: inline this method into caller?

trask (approved):

mateuszrzeszutek (approved):

PR #3838, opened 2021-08-13
Convert jsf to instrumenter api

laurit:
Part of #2713


instrumentation/jsf/jsf-common/library/src/main/java/io/opentelemetry/instrumentation/jsf/JsfServerSpanNaming.java
```diff
@@ -0,0 +1,42 @@
...
+  public static void updateServerSpanName(FacesContext facesContext) {
+    Context parentContext = Context.current();
+    ServerSpanNaming.updateServerSpanName(
+        parentContext, CONTROLLER, getServerSpanNameSupplier(parentContext, facesContext));
+  }
+
+  private static Supplier<String> getServerSpanNameSupplier(
+      Context context, FacesContext facesContext) {
+    return () -> getServerSpanName(context, facesContext);
+  }
```
trask, line 28: inline into caller?

instrumentation/jsf/jsf-common/library/src/main/java/io/opentelemetry/instrumentation/jsf/JsfRequest.java
```diff
@@ -0,0 +1,44 @@
...
+
+public class JsfRequest {
+  private final String spanName;
+
+  public JsfRequest(ActionEvent event) {
+    this.spanName = findSpanName(event);
+  }
+
+  public String spanName() {
+    return spanName;
```
trask, line 19: maybe
```suggestion
    return Objects.requireNonNull(spanName);
```

trask (approved):

mateuszrzeszutek (approved):

PR #3847, opened 2021-08-16
Update to SDK 1.5.0

anuraaga:
I have a few more tests to update but sending for initial sanity checking of metrics


instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/MeterTest.java
```diff
@@ -43,16 +43,15 @@
...
     instrumentationName = "test-" + test.getDisplayName();
-    meter = GlobalMeterProvider.getMeter(instrumentationName, "1.2.3");
+    meter = GlobalMeterProvider.get().get(instrumentationName, "1.2.3", "http://schema.org");
   }
 
   @Test
   void longCounter() {
-    LongCounter instrument =
-        meter.longCounterBuilder("test").setDescription("d").setUnit("u").build();
+    LongCounter instrument = meter.counterBuilder("test").setDescription("d").setUnit("u").build();
```
anuraaga, line 51: @jsuereth Can you just go through MeterTest to verify I migrated from old to new correctly?

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/MeterTest.java
```diff
@@ -333,8 +334,10 @@ void longValueRecorder() {
                       .hasDescription("d")
                       .hasUnit("u")
                       .hasInstrumentationLibrary(
-                          InstrumentationLibraryInfo.create(instrumentationName, "1.2.3"));
-                  assertThat(metric.getDoubleSummaryData().getPoints())
+                          InstrumentationLibraryInfo.create(instrumentationName, "1.2.3"))
+                      // TODO(anuraaga): Switch to histogram in future version?
```
anuraaga, line 338: Actually I think I need to fix the reverse OTLP parse to use histogram, not summary. Will check it

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java
```diff
@@ -51,19 +48,19 @@ public static RequestMetrics get() {
   }
 
   private final LongUpDownCounter activeRequests;
-  private final DoubleValueRecorder duration;
+  private final DoubleHistogram duration;
```
anuraaga, line 51: Is histogram always the one to pick instead of value recorder?

jkwatson: yes, ValueRecorder has been renamed to Histogram. Semantics should be identical (although maybe no negative recordings allowed on the Histogram?).

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/MeterTest.java
jkwatson: I don't think we have histograms as the default aggregation in 1.5.0. I think that comes in 1.6.0: https://github.com/open-telemetry/opentelemetry-java/pull/3491

jkwatson:
This looks very good to me! Not as big as I was fearing, actually.

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/MeterTest.java
anuraaga: Ah - let me know if I should revert the change to OpenTelemetryInstaller that I have in this PR.

jkwatson: yeah, we should probably keep that part in place until 1.6.0 hits, rather than thrash back to summary, then back to histogram again.

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/bridge/ApplicationDoubleHistogram.java
```diff
@@ -0,0 +1,46 @@
...
+  public void record(double value, Attributes applicationAttributes) {
+    agentHistogram.record(value, Bridging.toAgent(applicationAttributes));
+  }
+
+  @Override
+  public void record(double value, Attributes applicationAttributes, Context applicationContext) {
+    agentHistogram.record(
+        value,
+        Bridging.toAgent(applicationAttributes),
+        AgentContextStorage.getAgentContext(applicationContext));
```
anuraaga, line 38: In a future PR I will clean up this bridging pattern

mateuszrzeszutek (approved):

jkwatson (approved):

trask:
Thanks @anuraaga!

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/MeterTest.java
jsuereth: Hey, sorry, I was out sick two days ago, just getting to this now.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java
jsuereth: Currently histogram shouldn't have negative recordings.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetricsTest.java
```diff
@@ -61,33 +57,36 @@ void collectsMetrics() {
...
-                      entry(stringKey("http.method"), "GET"),
-                      entry(stringKey("http.scheme"), "https"),
-                      entry(stringKey("net.host.name"), "localhost"),
-                      entry(stringKey("net.host.port"), "1234"));
-              assertThat(data.getPercentileValues()).isNotEmpty();
-            });
+            metric ->
+                assertThat(metric)
+                    .hasName("http.client.duration")
+                    .hasDoubleSummary()
```
jsuereth, line 63: Glad this was useful!

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/bridge/ApplicationBoundDoubleCounter.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  public void add(double value) {
+    agentCounter.add(value);
+  }
+
+  @Override
+  public void add(double value, Context context) {
+    agentCounter.add(value, AgentContextStorage.getAgentContext(context));
```
jsuereth, line 27: Interesting.  AgentContext needs to be wrapped in some fashion?  Does this not need to apply when looking up the implicit context in the normal "add" method?

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/bridge/ApplicationDoubleHistogram.java
jsuereth: Happy to talk ideas around this.  I'm a bit stale on java modules, but have done my fair share of classloader sharing magic in the past.

instrumentation/oshi/library/src/main/java/io/opentelemetry/instrumentation/oshi/SystemMetrics.java
```diff
@@ -30,118 +34,111 @@ public static void registerObservers() {
...
+                r.observe(recv, Attributes.of(DEVICE_KEY, device, DIRECTION_KEY, "receive"));
+                r.observe(sent, Attributes.of(DEVICE_KEY, device, DIRECTION_KEY, "transmit"));
               }
-            })
-        .build();
+            });
 
     meter
-        .longSumObserverBuilder("system.disk.io")
+        .gaugeBuilder("system.disk.io")
```
jsuereth, line 113: If you wanted to preserve this as a sum, you should (likely) be using `upDownCounter().ofLongs()...buildWithCallback`.

However, looking at the attributes you have here, I'm not convinced up-down-sum is correct.

instrumentation/opentelemetry-api-metrics-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/metrics/bridge/ApplicationBoundDoubleCounter.java
anuraaga: We bridge the actual context API too, so if a user made a context current in threadloc, it's actually an agent context that is made current, app contexts just wrap and don't have their own storage. Here we just unwrap if the user passed such a Context in explicitly.

PR #3851, opened 2021-08-16
record exception in dubbo inst

tydhot:


instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboTracer.java
```diff
@@ -46,6 +46,9 @@ public Context startClientSpan(String interfaceName, String methodName) {
   }
 
   public void end(Context context, Result result) {
+    if (result.hasException()) {
```
iNikem, line 49: If you do this, then you don't need to handle span status below.

tydhot: yes i got it:)

iNikem (approved):

mateuszrzeszutek (approved):

tydhot:
@iNikem @mateuszrzeszutek hi, it seemed that there is a windows smoke test failed. I have no idea about this, what should I do to fix it?

mateuszrzeszutek:
Looks like a random failure - I reran all checks, hope they pass this time.

tydhot:
@mateuszrzeszutek thanks!

PR #3852, opened 2021-08-16
Convert okhttp-2.2 to instrumenter api

laurit:
Part of #2713


instrumentation/okhttp/okhttp-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/okhttp/v2_2/TracingInterceptor.java
```diff
@@ -5,34 +5,56 @@
...
     }
-    tracer().end(context, response);
+    instrumenter.end(context, request, response, null);
     return response;
   }
+
+  // Context injection is being handled manually for a reason: we want to use the OkHttp Request
+  // type for additional AttributeExtractors provided by the user of this library
+  // thus we must use Instrumenter<Request, Response>, and Request is immutable
+  private Request injectContextToRequest(Request request, Context context) {
```
laurit, line 53: copied from okhttp3 instrumentation

mateuszrzeszutek: Okhttp2 doesn't have library instrumentation, only javaagent - can you remove the part that mentions library instrumentation from the comment?

mateuszrzeszutek (approved):

anuraaga (approved):

PR #3853, opened 2021-08-16
Jetty8 instrumenter api

breedx-splk:
This is part of #2713.

This is the first one that @jkwatson and and I have done, so please give feedback!

It is acknowledged that some of the patterns built here (extractors) will eventually be re/used by other servlet-based servlet instrumentations. We intentionally did not factor that out (in the interest of doing the simplest thing first). It is expected that future instrumentations that need to do similar things can then factor this stuff out for reuse (servlet-common instrumentation most likely).


instrumentation/jetty/jetty-8.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jetty/v8_0/Jetty8AttributesExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+    extends HttpAttributesExtractor<HttpServletRequest, HttpServletResponse> {
+
+  @Override
+  protected @Nullable String method(HttpServletRequest httpServletRequest) {
+    return httpServletRequest.getMethod();
+  }
+
+  @Override
+  protected @Nullable String url(HttpServletRequest httpServletRequest) {
+    URI uri = URI.create(httpServletRequest.getRequestURL().toString());
```
laurit, line 26: I believe you don't really need to construct this `URI`, just get everything you need from `HttpServletRequest` like in `ServletHttpServerTracer.url()`

instrumentation/jetty/jetty-8.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jetty/v8_0/Jetty8HandlerAdvice.java
```diff
@@ -46,6 +55,17 @@ public static void stopSpan(
...
 
-    JettyHandlerAdviceHelper.stopSpan(tracer(), request, response, throwable, context, scope);
+    if (scope == null) {
+      return;
+    }
+    scope.close();
+
+    if (throwable != null) {
+      if (response.getStatus() < 400) {
+        response.setStatus(500);
```
laurit, line 65: This feels a bit scary. It would be best if we didn't need to change response status, kind of hard to tell what behavioural changes it might cause.

instrumentation/jetty/jetty-8.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jetty/v8_0/Jetty8HandlerAdvice.java
```diff
@@ -46,6 +55,17 @@ public static void stopSpan(
...
+      }
+    }
+
+    instrumenter().end(context, request, response, throwable);
```
laurit, line 69: `JettyHandlerAdviceHelper.stopSpan` doesn't always end span (in case of async requests span will be ended from a listener). I suspect you really can't change that without breaking something.

breedx-splk:
Thanks @laurit. Agree that this needs a lot more work. Closing for now.

PR #3862, opened 2021-08-17
Deprecate MetricExporterFactory

mateuszrzeszutek:
...and remove `MetricServer`! When I opened this class in IntelliJ today I was suprised that the IDE has greyed out its name - turns out it hasn't been in use since January 🤣  We've accidentally removed all usages while migrating to the SDK autoconfigure module and nobody even noticed that - I guess we can skip deprecating it and just delete it right away 😂


trask (approved):
oops(?)

PR #3863, opened 2021-08-17
Use valid Java identifiers for message keys

trask:
This is causing error on IBM MQ:

```
Error Setting JMS Property io.opentelemetry.javaagent.shaded.instrumentation.spring.integration.ContextAndScope with value ContextAndScope{context=null, scope=INSTANCE}
com.ibm.msg.client.jms.DetailedMessageFormatException: JMSCC0049: The property name 'io.opentelemetry.javaagent.shaded.instrumentation.spring.integration.ContextAndScope' is not a valid Java(tm) identifier.
```

Probably could rename to "valid" header name, but seems better not to use message header for local propagation.

### DIFF
--- a/instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
+++ b/instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
@@ -22,8 +22,11 @@ import org.springframework.messaging.support.NativeMessageHeaderAccessor;
 import org.springframework.util.LinkedMultiValueMap;
 
 final class TracingChannelInterceptor implements ExecutorChannelInterceptor {
-  private static final String CONTEXT_AND_SCOPE_KEY = ContextAndScope.class.getName();
-  private static final String SCOPE_KEY = TracingChannelInterceptor.class.getName() + ".scope";
+  // some messaging products (e.g. IBM MQ) require message keys to be valid Java identifiers
+  private static final String CONTEXT_AND_SCOPE_KEY =
+      ContextAndScope.class.getName().replace('.', '_');
+  private static final String SCOPE_KEY =
+      TracingChannelInterceptor.class.getName().replace('.', '_') + "_scope";
 
   private final ContextPropagators propagators;
   private final Instrumenter<MessageWithChannel, Void> instrumenter;
--- a/instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
+++ b/instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
@@ -9,7 +9,6 @@ import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.context.propagation.ContextPropagators;
-import io.opentelemetry.instrumentation.api.caching.Cache;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -22,12 +23,16 @@
...
 
 final class TracingChannelInterceptor implements ExecutorChannelInterceptor {
-  private static final String CONTEXT_AND_SCOPE_KEY = ContextAndScope.class.getName();
-  private static final String SCOPE_KEY = TracingChannelInterceptor.class.getName() + ".scope";
 
   private final ContextPropagators propagators;
   private final Instrumenter<MessageWithChannel, Void> instrumenter;
 
+  // TODO (trask): optimize for javaagent by using field-backed context store
+  private final Cache<MessageChannel, ContextAndScope> sendContextAndScopeHolder =
```
laurit, line 31: Isn't `MessageChannel` a shared object that is used to send many possibly concurrent messages?

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -22,12 +23,16 @@
 import org.springframework.util.LinkedMultiValueMap;
 
 final class TracingChannelInterceptor implements ExecutorChannelInterceptor {
-  private static final String CONTEXT_AND_SCOPE_KEY = ContextAndScope.class.getName();
-  private static final String SCOPE_KEY = TracingChannelInterceptor.class.getName() + ".scope";
```
laurit, line 26: From what I understand dot is invalid, underscore would be fine

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
trask: oh no 🙁

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
trask: 👍 I'll try that. Any idea what a non-serializable message header even means tho?

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
mateuszrzeszutek: If the underscore idea doesn't work out, here's another take on this problem that I considered when developing this instrumentation: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3120/commits/7c8d614f4c9734580ab3b54fcd916383ffeffddd

mateuszrzeszutek (approved):

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
laurit: This is a very good question. In out test with rabbitmq https://github.com/spring-projects/spring-amqp/blob/bd088731b08ceb7ca90543e18b19cbf70a21c6c8/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/support/DefaultMessagePropertiesConverter.java#L184 converts it to string by calling `toString`. For actual jms I guess the relevant part is https://github.com/spring-projects/spring-integration/blob/590fb01743d537f3436bccf3a8c445c51b7640d1/spring-integration-jms/src/main/java/org/springframework/integration/jms/DefaultJmsHeaderMapper.java#L133 which seems to ignore properties of unknown type. Though this doesn't explain how there is an exception with IBM MQ, there must be some other place where property is set on jms message. Even if we correct the property name it is likely to still fail as https://docs.oracle.com/javaee/5/api/javax/jms/Message.html#setObjectProperty(java.lang.String,%20java.lang.Object) does not accept arbitrary objects (assuming that this is the method that caused original exception).

trask:
let's try this simple fix, and revisit if needed

PR #3864, opened 2021-08-18
Support versioned instrumentation helper resources

trask:
This problem hasn't impacted our instrumentation yet, but it will (and it has impacted the azure-sdk instrumentation that I still haven't upstreamed).

For example:

azure-core-1.14 needs to inject an SPI file (`META-INF/services/com.azure.core.util.tracing.Tracer`) that points to azure-core-1.14 implementation class, while azure-core-1.19 needs to inject the same SPI file that points to azure-core-1.19 implementation class.

The problem is that these two files (`META-INF/services/com.azure.core.util.tracing.Tracer`) get clobbered together when they are placed in the agent jar at `inst/META-INF/services/com.azure.core.util.tracing.Tracer`.

The proposal in this PR is for instrumentation to be able to store their helper resources anywhere, e.g. `azure-core-1.14/META-INF/services/com.azure.core.util.tracing.Tracer`, and when the instrumentation registers its resources, it can provide both a "from" (`azure-core-1.14/META-INF/services/com.azure.core.util.tracing.Tracer`) and a "to" (`META-INF/services/com.azure.core.util.tracing.Tracer`).

Also, making changes to muzzle appears to be more difficult now, I think it requires minimum 3 PRs:

* This PR: updates the artifacts that are needed to build the gradle plugin
* Next PR: update the gradle plugin itself (this change requires updating `MuzzleCodeGenerator`)
* The PR after that: use the changes in instrumentation (e.g. use the new API and remove calls to the deprecated method)


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModule.java
```diff
@@ -117,11 +117,19 @@ public boolean isHelperClass(String className) {
...
+   *
+   * @deprecated use {@link #registerHelperResources(HelperResources)} instead.
+   */
+  @Deprecated
   public List<String> helperResourceNames() {
     return Collections.emptyList();
   }
 
+  /** Register resource names to inject into the user's class loader. */
+  public void registerHelperResources(HelperResources helperResources) {}
```
trask, line 131: better name for `HelperResources`?

anuraaga (commented):
Even if it's not required, should we apply the pattern to our libraries to demonstrate the pattern and possible future proof?
x

trask:
> Even if it's not required, should we apply the pattern to our libraries to demonstrate the pattern and possible future proof?

👍 tho I can't do this until the 3rd PR though b/c need to publish new artifact snapshots from 1st PR, then need to update and publish new gradle plugin snapshot from 2nd PR, then can do this in the 3rd PR 😭

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/HelperResources.java
```diff
@@ -0,0 +1,32 @@
...
+   * <p>{@code resourceName} and {@code resourceNameForContent} can be the same, but it is often
+   * desirable to use a slightly different name or path for the resource file provided by the
+   * instrumentation so that multiple instrumentation modules (e.g. applying to different versions
+   * of a particular library) can co-exist inside the agent jar file.
+   *
+   * @param resourceName the name of the resource to inject in to the user's class loader
+   * @param resourceNameForContent the name of the instrumentation resource which provides the
+   *     content for injection
+   */
+  void register(String resourceName, String resourceNameForContent);
```
mateuszrzeszutek, line 31: Nit: WDYT about naming those `agentResourceName`, `applicationResourceName`? (or `...ResourcePath`?)

mateuszrzeszutek (approved):

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModule.java
mateuszrzeszutek: `HelperResourcesBuilder`? not sure if it's better though 😄

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/HelperInjector.java
```diff
@@ -149,16 +149,19 @@ public static HelperInjector forDynamicTypes(
       classLoader = injectHelperClasses(typeDescription, classLoader, module);
     }
 
-    if (helpersSource != null && !helperResourceNames.isEmpty()) {
-      for (String resourceName : helperResourceNames) {
-        URL resource = helpersSource.getResource(resourceName);
+    if (helpersSource != null && !helperResourcePathMappings.isEmpty()) {
+      for (Map.Entry<String, String> entry : helperResourcePathMappings.entrySet()) {
+        String agentResourcePath = entry.getValue();
+        String applicationResourcePath = entry.getKey();
```
mateuszrzeszutek, line 155: Thanks for those variables 👍 
This part is now much easier to understand

mateuszrzeszutek (approved):

PR #3865, opened 2021-08-18
Option to suppress spring-webmvc internal spans

trask:
Introduces a new option `otel.instrumentation.spring-webmvc.experimental.suppress-internal-spans` to suppress the internal spans produced by the spring-webmvc instrumentation, while still retaining the nice SERVER span names that it produces.

If this is agreeable, I will add similar options to other web routing/handling instrumentation.


trask:
I explored using the new "instrumentation type" for this suppression, but I think the instrumentation type is nicely scoped to semantic conventions, and so this doesn't quite fit. And since there's not semantic conventions, or semantic convention attribute extractors for controllers and views, there's no way to auto-detect the instrumentation type in this case.

An open question I have is that it might be nicer to have a "noop" Instrumenter that can be used, instead of having `setDisabled()` on InstrumenterBuilder, but since Instrumenter is not an interface creating a "noop" Instrumenter is a bit awkward.

trask:
I went ahead and applied to all existing Instrumenters to get a fuller picture

instrumentation/dropwizard-views-0.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/dropwizardviews/DropwizardSingletons.java
```diff
@@ -7,15 +7,21 @@
...
+import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 
 public final class DropwizardSingletons {
 
   private static final String INSTRUMENTATION_NAME = "io.opentelemetry.dropwizard-views-0.7";
 
+  private static final boolean SUPPRESS_CONTROLLER_SPANS =
+      Config.get()
+          .getBoolean("otel.instrumentation.common.experimental.suppress-controller-spans", false);
```
anuraaga, line 19: We should somehow have this in the javaagent-api artifact rather than copying into every module I guess. Either a method on `Config` or something like `JavaAgentConfig` with all of them.

anuraaga (approved):

anuraaga (approved):

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/ExperimentalConfig.java
```diff
@@ -0,0 +1,41 @@
...
+
+package io.opentelemetry.instrumentation.api.config;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public final class ExperimentalConfig {
+
+  // lazy initialized, so that javaagent can set it, and library instrumentation can fall back and
+  // read system properties
+  @Nullable private static ExperimentalConfig instance = null;
```
mateuszrzeszutek, line 14: That's not true for this class - I think we can safely use a `static final` instance. It is `Config` that needs to have lazy initialization, because it's "real" instance is being set by the agent.

mateuszrzeszutek (approved):

PR #3866, opened 2021-08-18
Don't pass configuration to SDK autoconfigure through system props

mateuszrzeszutek:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -61,98 +77,186 @@ public static Config get() {
...
+  @Nullable
+  public Boolean getBoolean(String name) {
+    return getTypedProperty(name, Boolean::parseBoolean, null);
+  }
+
+  /**
+   * Returns a boolean-valued configuration property or {@code defaultValue} if a property with name
+   * {@code name} has not been configured.
    */
   public boolean getBooleanProperty(String name, boolean defaultValue) {
```
mateuszrzeszutek, line 109: I'm planning to rename all `get*Property()` methods to `get*()` in the next PR - those shorter names seem a bit nicer, and they're consistent with the SDK `ConfigProperties` interface too.

jkwatson: is the plan to have this implement `ConfigProperties` then?

mateuszrzeszutek: Unfortunately that won't be possible with the current shape of this repo: `instrumentation-api` only uses OTel API, not the SDK, so we can't directly implement the `ConfigProperties` interface here.

jkwatson: ok then. makes sense!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -70,28 +69,6 @@ public static synchronized void installAgentTracer(Config config) {
...
-            .getAllProperties();
-
-    allProperties.forEach(
-        (key, value) -> {
-          if (!environmentProperties.containsKey(key)
-              && key.startsWith("otel.")
-              && !key.startsWith("otel.instrumentation")) {
-            System.setProperty(key, value);
-          }
-        });
```
trask, line 92: 🎉

trask (approved):

PR #3868, opened 2021-08-18
Support extraResources in WindowsTestContainerManager

mateuszrzeszutek:


anuraaga (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LibertyServletOnlySmokeTest.groovy
```diff
@@ -5,24 +5,11 @@
...
-@IgnoreIf({ os.windows }) //WindowsTestContainerManager does not support extra resources
 class LibertyServletOnlySmokeTest extends LibertySmokeTest {
 
   @Override
   protected Map<String, String> getExtraResources() {
     return ["liberty-servlet.xml": "/config/server.xml"]
   }
-
-  @Override
-  protected String getSpanName(String path) {
```
laurit, line 20: I believe this is wrong. On liberty static files are not served by a servlet. On default conf server adds a servelt filter due to which these get named `/app/*` but on servlet only conf these really are named `HTTP GET`.
The real problem is probably that windows and linux images have different paths for files `/config/server.xml` works only for linux image on windows image configuration file should be copied to `/server/usr/servers/defaultServer/server.xml` (I hope this is correct).

mateuszrzeszutek: Thanks, I'll try it out.

laurit (approved):

PR #3869, opened 2021-08-18
Make empty agent bridged context equal root context

laurit:
Inside agent `Context.current() == Context.root()` when context is empty. The same holds when using sdk without agent. It seems reasonable to assume that this should also hold when using sdk with agent.


mateuszrzeszutek (approved):

instrumentation/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/context/AgentContextStorage.java
```diff
@@ -40,6 +40,11 @@
   private static final Logger logger = LoggerFactory.getLogger(AgentContextStorage.class);
 
   public static final AgentContextStorage INSTANCE = new AgentContextStorage();
```
anuraaga, line 42: Is it possible to use the extension?

https://github.com/open-telemetry/opentelemetry-java/blob/main/context/src/main/java/io/opentelemetry/context/ContextStorage.java#L112

Does it break muzzle?

mateuszrzeszutek (approved):

instrumentation/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/context/AgentContextStorage.java
laurit: Yes, it fails muzzle when older api is used because I call `ContextStorage.root()` which was just added. Seems like wrapper based approach won't work on older api because `Context.root()` just returns `ArrayBasedContext.root()` which we would still need to wrap. Similarly old approach would fail when `ArrayBasedContext` is not used, should also wrap what is returned from `ContextStorage.root()`.

anuraaga (approved):

instrumentation/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/context/AgentContextStorage.java
laurit: Replaced call to `ContextStorage.root()` with method handle, @anuraaga could you review.

trask (approved):
👍 created #3884 to add tests for the new AgentContext wrapper bridging

PR #3872, opened 2021-08-18
[overhead] - Add network average read/write to output

breedx-splk:
This builds on #3832 and will require a rebase after that gets merged.


jkwatson:
I'm curious what you think this particular information will be useful for. Especially since it will be confounded by the actual network traffic from running a regular webapp and sending traffic at it.

jkwatson (approved):

breedx-splk:
> I'm curious what you think this particular information will be useful for. Especially since it will be confounded by the actual network traffic from running a regular webapp and sending traffic at it.

My hope is that the webapp traffic is more or less the same with/without agent or and between different agent versions. Many users pay for bandwidth and would like to have an understanding of how their costs might change when they add the agent. This also allows us to see how changes like turning off certain instrumentations or (down the road) metrics etc. might impact bandwidth usage which translates to cost.

trask (approved):

PR #3881, opened 2021-08-20
Rename Config get*Property() methods to get*()

mateuszrzeszutek:
... and rewrite the rest of the Spock test to JUnit.

See https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3866#discussion_r690978170


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -207,16 +207,16 @@ public Duration getDuration(String name, Duration defaultValue) {
...
+    return getList(name, Collections.emptyList());
   }
 
   /**
    * Returns a list-valued configuration property or {@code defaultValue} if a property with name
    * {@code name} has not been configured. The format of the original value must be comma-separated,
    * e.g. {@code one,two,three}.
    */
-  public List<String> getListProperty(String name, List<String> defaultValue) {
+  public List<String> getList(String name, List<String> defaultValue) {
```
mateuszrzeszutek, line 219: I used simpler `getList`/`getMap` names instead of `getCommaSeparatedValues`/`getCommaSeparatedMap` that SDK uses -- all other method names are consistent with the `ConfigProperties` interface.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/config/ConfigTest.java
```diff
@@ -144,17 +150,65 @@ void shouldGetMap() {
...
+  }
+
+  @SuppressWarnings("unused")
+  private static Stream<Arguments> agentDebugParams() {
+    return Stream.of(
+        Arguments.of("true", true), Arguments.of("blather", false), Arguments.of(null, false));
+  }
+
+  @ParameterizedTest
+  @MethodSource("instrumentationEnabledParams")
```
anuraaga, line 179: I tend to prefer `ArgumentsSource` for the compile safety at the expense of some lines of class definition IntelliJ quickly completes. Looks like it would remove the `unused` suppression

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
anuraaga: Seems like a nice change we can make to the SDK too (since it's alpha)

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/Config.java
```diff
@@ -88,11 +91,31 @@ public String getProperty(String name) {
   /**
    * Returns a string-valued configuration property or {@code defaultValue} if a property with name
    * {@code name} has not been configured.
+   *
+   * @deprecated Use {@link #getString(String, String)} instead.
    */
+  @Deprecated
```
mateuszrzeszutek, line 97: --staged

mateuszrzeszutek, line 97: I had to leave all those old methods and make them deprecated -- muzzle wouldn't let me remove them in this PR, I'll have to do that in the next one.

trask (approved):

PR #3882, opened 2021-08-20
AsyncHttpClient: extract net attributes from InetSocketAddress

laurit:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/ca8a119e01a2e56c9094d23577838b6e9fccda00/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java#L888 suggests that filling net attributes from request url isn't ideal.


anuraaga (approved):
Great!

mateuszrzeszutek (approved):

trask (approved):

PR #3885, opened 2021-08-20
[overhead] add in some additional verbosity for debugging.

breedx-splk:
The [build failed](https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/3378798745?check_suite_focus=true) with a file not found, but it's unclear which file was missing. Probably one of the result output files (jfr or k6 or startup time file) but without a filename it is hard to continue troubleshooting.


jkwatson (approved):

breedx-splk:
Derp. Didn't realize I can run the actions in my fork.

PR #3888, opened 2021-08-20
[overhead] fix permissions issue with test results

breedx-splk:
Well, after much troubleshooting I was able to determine that the test results couldn't write back to the parent mounted directory due to the user being restricted. This PR just runs the testrunner as root inside the container and also expands the max run length of the testrunner.


jkwatson (approved):

anuraaga (approved):

laurit:
closing and reopening to trigger checks again

mateuszrzeszutek (approved):

Issue #3516, opened 2021-07-07
Inconsistencies with version tags


iNikem:
If you open [`CHANGELOG` file tagged with version 1.3.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.3.0/CHANGELOG.md) then you see that it does NOT contain a section for 1.3.0 release. Also examples [distro example](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/v1.3.0/examples/distro/build.gradle) uses javaagent version 1.2.0

How about having repository tagged by a version tag in a state which is completely compatible with that version? Meaning having up-to-date documentation and changelog for that version and having all examples using that version. Yes, it means that releases require a separate PR, but we already have one. Only now we do it post-release, instead of pre-release.

anuraaga:
Currently we tag the exact commit that built the code as forced by the tooling. With the proposal, we would tag a commit after that which shouldn't have code changes so in practice is very similar. One corner case would be if a PR is merged between running release build and updating the examples, CHANGELOG. The only way I can think of to solve that is to update examples on main and CHANGELOG, change release flow to always create a release branch, always cherrypick the example / CHANGELOG update into the release branch, and tag the release branch, not main branch. I guess it could be considered ideal, but a lot of machinery

iNikem:
Wait, what do I miss?

- I want to release next version, 1.4.0
- I make a PR, which updates changelog and examples to reflect the state of (yet unreleased) 1.4.0
- PR is merged
- Now I manually tag the resulting commit on the main branch with `v1.4.0` tag
- GHA kicks in, builds and publishes 1.4.0, then build examples against just published version (because examples on that commit are using exact version 1.4.0) as a smoke test.

anuraaga:
> PR is merged

How can we merge this PR? Since 1.4.0 is unreleased, there's no way it could pass with a green build right?

iNikem:
> > PR is merged
> 
> How can we merge this PR? Since 1.4.0 is unreleased, there's no way it could pass with a green build right?

We don't actually need to build examples in `ci` workflow. If they usually use snapshot versions, then we can build them in PR, forcing to always update examples alongside the actual changes.

anuraaga:
> If they usually use snapshot versions, then we can build them in PR, forcing to always update examples alongside the actual changes.

[...]

PR #3899, opened 2021-08-22
Fix examples build

### LINKED ISSUE #3889: Nightly build
<a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/1152786361">
Nightly build #71</a> failed. Please take a look and fix it ASAP.

trask:
Closes #3889 (long-term solution is #3516)

### DIFF
--- a/testing/agent-for-testing/build.gradle.kts
+++ b/testing/agent-for-testing/build.gradle.kts
@@ -70,10 +70,6 @@ project(":instrumentation").subprojects {
 }
 
 tasks {
-  jar {
-    enabled = false
-  }
-
   val relocateJavaagentLibs by registering(ShadowJar::class) {
     configurations = listOf(javaagentLibs)
 
@@ -114,15 +110,6 @@ tasks {
 
       jvmArgs("-Dotel.javaagent.debug=true")
       jvmArgs("-javaagent:${shadowJar.get().archiveFile.get().asFile.absolutePath}")
-    }
-  }
-
-  // Because shadow does not use default configurations
-  publishing {
-    publications {
-      named<MavenPublication>("maven") {
-        project.shadow.component(this)
-      }
     }
   }
 }
--- a/.github/scripts/local.init.gradle.kts
+++ b/.github/scripts/local.init.gradle.kts
@@ -1,10 +1,6 @@
-addListener(object : BuildAdapter() {
-  override fun projectsEvaluated(gradle: Gradle) {
-    gradle.rootProject {
-      repositories {
-        mavenLocal()
-        removeIf { it.name == "sonatype" }
-      }
-    }
+allprojects {
+  repositories {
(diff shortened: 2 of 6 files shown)


laurit (approved):

trask:
oh noes, the examples build needs a successful 1.6.0-SNAPSHOT. I thought I could kick off CI build to trigger 1.6.0-SNAPSHOT, but of course that fails on the examples build. trying to decide what is the least bad way out of this 😭

trask:
`local.init.gradle.kts` didn't seem to be working. I updated it based on the example at https://docs.gradle.org/current/userguide/init_scripts.html#sec:configuring_projects_from_an_init_script, and at least seems to be working locally for me now, we'll see how build goes...

trask:
Fixing `local.init.gradle.kts` revealed that #3892 caused `agent-for-testing` to be published with packaging `pom` 😭 will go work on that now.. (and looks like we'll be needing a 1.5.1 release soon)

trask:
I went to leverage existing shadowJar solution from `opentelemetry-javaagent`, but found that was also published with packaging "pom"..

trask:
never been so lucky with gradle shadow, deleted some LOC and somehow it's working 😂🎉

trask:
went ahead and merged because this is blocking other PRs, but happy to revisit any of it

testing/agent-for-testing/build.gradle.kts
```diff
@@ -70,10 +70,6 @@ project(":instrumentation").subprojects {
 }
 
 tasks {
-  jar {
-    enabled = false
```
mateuszrzeszutek, line 74: Hmm, `jar` and `shadowJar` produce jars with the exact same name (since shadow jar has `archiveClassifier.set("")`). Is it safe to assume that the shadowed one will be published?

ryandens:
If anyone is looking for a workaround trying to consume the `agent-for-testing` module in the `1.4.0 - 1.5.0` release range, a workaround I found was to specify the file extension of the maven dependency, e.g. 

```groovy
  testAgent group: "io.opentelemetry.javaagent", name: "opentelemetry-agent-for-testing", version: "${versions.opentelemetryJavaagentAlpha}", ext: "jar"
```

The difference between the `agent-for-testing` maven publication and the `opentelemetry-javaagent` publication is that, although they both have `packaging` type `pom`, `opentelemetry-javaagent`'s agent jar artifact is published with classifier `all`. 

Thanks for this fix! It helped me figure out a workaround that works for me 😄

iNikem:
@trask Something is fishy here...

First, I definitely tested local init and my version worked, but your version did not. Very strange that now it is reversed...

Second, do I read it correctly, that we now always test extensions with locally published dependencies? I think we have to test them with globally published as well.

PR #3910, opened 2021-08-23
Revert okhttp library instrumentation back to using standard reflection to support Android usage

jkwatson:
It turns out that MethodHandles are not supported on Android API levels 21-25.

### DIFF
--- a/instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
+++ b/instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
@@ -22,7 +22,8 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 class TracingCallFactory implements Call.Factory {
   private static final Cache<Request, Context> contextsByRequest =
       Cache.newBuilder().setWeakKeys().build();
-
+  // We use old-school reflection here, rather than MethodHandles because Android doesn't support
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
```diff
@@ -24,22 +23,19 @@
   private static final Cache<Request, Context> contextsByRequest =
       Cache.newBuilder().setWeakKeys().build();
 
-  @Nullable private static MethodHandle timeoutMethodHandle;
-  @Nullable private static MethodHandle cloneMethodHandle;
+  @Nullable private static Method timeoutMethod;
```
trask, line 26: can you add a comment here about not being able to use method handles (so no one changes it back), at least until we get some build tooling in place to prevent similar to otel-java repo?

trask (approved):
> Since this is a regression for Android, could we get a patch release with this commit, possibly?

yes, definitely

anuraaga (approved):

instrumentation/okhttp/okhttp-3.0/library/src/main/java/io/opentelemetry/instrumentation/okhttp/v3_0/TracingCallFactory.java
jkwatson: done

anuraaga (approved):

Issue #3916, opened 2021-08-24
Fix context leaks in existing tests


trask:
This is a tracking issue to fix context leaks in existing tests for all instrumentations that currently have to disable the strict context check.

- [ ] akka-http-10.0/javaagent/build.gradle.kts
- [ ] aws-sdk/aws-sdk-1.11/javaagent-unit-tests/build.gradle.kts 
- [ ] finatra-2.9/javaagent/build.gradle.kts 
- [ ] grpc-1.6/javaagent/build.gradle.kts 
- [ ] kotlinx-coroutines/javaagent/build.gradle.kts 
- [ ] kubernetes-client-7.0/javaagent/build.gradle.kts 
- [ ] okhttp/okhttp-3.0/javaagent/build.gradle.kts

trask:
Starting to track associated errors in case they are useful:
```
Elasticsearch6NodeClientTest > test elasticsearch get[0] FAILED	
[...]

PR #3923, opened 2021-08-24
Strict context check on tomcat and jetty

laurit:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3916

### DIFF
--- a/instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/AbstractServlet3Test.groovy
+++ b/instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/AbstractServlet3Test.groovy
@@ -10,11 +10,13 @@ import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEn
 import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.QUERY_PARAM
 import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.REDIRECT
 import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.SUCCESS
+import static org.awaitility.Awaitility.await
 
 import io.opentelemetry.instrumentation.test.AgentTestTrait
 import io.opentelemetry.instrumentation.test.asserts.TraceAssert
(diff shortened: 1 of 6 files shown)


instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/AbstractServlet3Test.groovy
```diff
@@ -45,6 +47,25 @@ abstract class AbstractServlet3Test<SERVER, CONTEXT> extends HttpServerTest<SERV
...
+    // wait for async request threads to complete
+    await()
+      .atMost(15, TimeUnit.SECONDS)
+      .until({ !isRequestRunning() })
+  }
+
+  static boolean isRequestRunning() {
+    def result = Thread.getAllStackTraces().values().find {stackTrace ->
+      def element = stackTrace.find {
+        return ((it.className == "org.apache.catalina.core.AsyncContextImpl\$RunnableWrapper" && it.methodName == "run")
```
anuraaga, line 60: I'm a little scared of adding too many of these sort of checks. Have we ruled out

- Test, client call doesn't actually consume entire response or otherwise ends test before request completes properly

- Instrumentation is not ordering context close, span finish, response return, correctly

laurit: My understanding is that async request is completed when response is sent to client as this happens inside the async task it hasn't exited it's scope yet. Usually async task will manage to exit before test completes asserting spans and does context leak check but as our tests generate high load it can happen that everything else completes before async context runnable exits.
Build with failures on tomcat https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/3407877673
I hope we can find some other way of solving this as the same issue will affect a lot of tests.

trask (approved):
👍😅 this is very tricky issue, but I think a good solution for now at least

since it looks like we'll need this in lots of places, can we pull this behavior up into `HttpServerTest` and configure the list of class names in each test?

iNikem (approved):

trask:
> since it looks like we'll need this in lots of places, can we pull this behavior up into `HttpServerTest` and configure the list of class names in each test?

this can be done as a follow up if it makes sense

PR #3928, opened 2021-08-24
Mark our generated classes and members as synthetic

laurit:


testing-common/integration-tests/build.gradle.kts
```diff
@@ -27,14 +27,17 @@ tasks {
   val testFieldInjectionDisabled by registering(Test::class) {
     filter {
       includeTestsMatching("context.FieldInjectionDisabledTest")
+      isFailOnNoMatchingTests = false
     }
+    include("**/FieldInjectionDisabledTest.*")
```
trask, line 32: do you know if this line is needed? I noticed some other tests also have it (in addition to `includeTestsMatching`), but it doesn't seem to be needed for me to run the single test from IDE

trask (approved):

mateuszrzeszutek (approved):

testing-common/integration-tests/build.gradle.kts
laurit: I added these. Maybe I'm doing something wrong but for me intellij gets confused when there are 2 test tasks. For example when I run `FieldBackedProviderTest` intellij asks whether to run `test` or `testFieldInjectionDisabled`, choose `test` and then it still runs `testFieldInjectionDisabled` because `test` depends on it, for some reason `includeTestsMatching` gets ignored and test fails. This extra include makes it work for me.

trask: got it, thx for the explanation! yup I am seeing that behavior too

Issue #3900, opened 2021-08-22
opentelemetry-javaagent 1.5.0 published with packaging "pom" instead of "jar"


trask:
~And `opentelemetry-agent-for-testing` too.~ `opentelemetry-agent-for-testing` was fixed in #3899

iNikem:
@trask @mateuszrzeszutek we are supposed to have an extension test which verifies this. Did not it work?

mateuszrzeszutek:
I'm guessing it didn't - the way the javaagent was built has changed some time ago (and the publishing issue was introduced at that time) and extension tests didn't fail on that at all.

PR #3929, opened 2021-08-24
Fix javaagent publishing

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3900

I have no idea why, but when you have the default `shadowJar` with empty classifier it'll overwrite the `jar` task output and get attached to the maven publication properly. So I switched those two shadowJar tasks that we had: now the default `shadowJar` builds the javaagent without exporters, no classifier (previously `lightShadow`); and a new custom shadow task builds the full javaagent with exporters and `-all` classifier (previously `shadowJar`).
It's similar to what was done in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3899, but we're publishing one more jar in this case.

### DIFF
--- a/javaagent/build.gradle.kts
+++ b/javaagent/build.gradle.kts
@@ -100,11 +100,6 @@ tasks {
     }
   }
 
-  // lightShadow is the default classifier we publish so disable the default jar.
-  jar {
-    enabled = false
-  }
-
   val relocateJavaagentLibs by registering(ShadowJar::class) {
     configurations = listOf(javaagentLibs)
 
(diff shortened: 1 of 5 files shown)


trask (approved):
❤️

trask:
@mateuszrzeszutek it looks like `telemetry.auto.version` / `Implementation-Version` got lost somehow

javaagent/build.gradle.kts
```diff
@@ -137,6 +132,7 @@ tasks {
     archiveClassifier.set("")
 
     manifest {
+      attributes(jar.get().manifest.attributes)
```
mateuszrzeszutek, line 135: I have absolutely no idea how this worked without this line before 🙈

PR #3944, opened 2021-08-25
Convert play-ws to instrumenter api

laurit:
Part of #2713


instrumentation/play-ws/play-ws-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/playws/PlayWsClientHttpAttributesExtractor.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  @Nullable
+  protected String scheme(Request request) {
+    return request.getUri().getScheme();
+  }
+
+  @Override
+  @Nullable
+  protected String userAgent(Request request) {
+    return null;
```
mateuszrzeszutek, line 53: It can be extracted from the headers, can't it?
```suggestion
    return request.getHeaders().get("User-Agent");
```

mateuszrzeszutek (approved):

instrumentation/play-ws/play-ws-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/playws/PlayWsClientHttpAttributesExtractor.java
laurit: Turns out it isn't that easy. PlayWs uses AHC and AHC only sets `User-Agent` on underlying netty request so getting access to this header requires more effort, see #3930. To me this seems like an unsolved problem with our current instrumentation strategy, if higher level library uses lower level library then higher level library instrumentation might not have access to all the details that lower level library instrumentation has.

trask: this might be a good use case for trying out the CLIENT span enrichment idea, of having nested CLIENT span update some attributes on the parent CLIENT span

trask (approved):

Issue #3661, opened 2021-07-23
Restlet instrumentation


kubawach:
**Is your feature request related to a problem? Please describe.**
Restlet (https://restlet.talend.com/) is currently not instrumented, although a very popular library.

**Describe the solution you'd like**
Instrumentation of the core Restlet library.
Supported versions:
- 1.1 (legacy) with com.noelios.restlet reference implementation of the API (org.restlet:org.restlet is API and org.restlet: org.restlet.ext.spring is the implementation
- if feasible - new 2.x stream

Issue #3662, opened 2021-07-23
Restlet spring extension instrumentation


kubawach:
**Is your feature request related to a problem? Please describe.**
After core Restlet instrumentation (#3661) is implemented, popular spring extension should be done as well.

**Describe the solution you'd like**
Spring extension (org.restlet: org.restlet.ext.spring) can yield some additional data for instrumentation. 
Versions:
- 1.1.x legacy stream
- new 2.x stream if feasible

PR #3946, opened 2021-08-25
Restlet instrumentation

Enkelian:
Beginning of Restlet instrumentation - issue #3661. 
* server instrumentation - creation of server span
* works with versions 1.0.x and 1.1.x, 1.2 will be checked in future PRs


instrumentation/restlet/restlet-core-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_1/RestletInstrumentationModule.java
```diff
@@ -0,0 +1,25 @@
...
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.Arrays;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class RestletInstrumentationModule extends InstrumentationModule {
+
+  public RestletInstrumentationModule() {
+    super("restlet-core", "restlet-core-1.1");
```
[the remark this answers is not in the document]
Enkelian: I named it that way because there are extensions to the Restlet library that we want to support  - e.g. Spring Extension #3662. But after giving it some thought, I guess extension instrumentation will have names like "restlet-spring-extension", so indeed naming core instrumentation just "restlet" should suffice. WDYT?

instrumentation/restlet/restlet-core-1.1/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_1/RestletHttpServerTracer.java
```diff
@@ -0,0 +1,78 @@
...
+package io.opentelemetry.instrumentation.restlet.v1_1;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapGetter;
+import io.opentelemetry.instrumentation.api.tracer.HttpServerTracer;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+
+public class RestletHttpServerTracer extends HttpServerTracer<Request, Response, Request, Request> {
```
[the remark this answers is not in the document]
Enkelian: Looked into the new API, I think I can migrate this in this PR.

instrumentation/restlet/restlet-core-1.1/testing/src/main/groovy/io/opentelemetry/instrumentation/restlet/v1_1/AbstractRestletServerTest.groovy
```diff
@@ -0,0 +1,144 @@
...
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.NOT_FOUND
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.PATH_PARAM
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.QUERY_PARAM
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.REDIRECT
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.SUCCESS
+
+abstract class AbstractRestletServerTest extends HttpServerTest<Server> {
+
+  def component
+  def host
```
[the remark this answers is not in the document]
Enkelian: Sure, thanks for the explanation!

instrumentation/restlet/restlet-core-1.1/library/src/test/groovy/RestletServerTest.groovy
```diff
@@ -0,0 +1,81 @@
...
+class RestletServerTest extends AbstractRestletServerTest implements LibraryTestTrait {
+
+  @Override
+  Restlet wrapRestlet(Restlet restlet, String path){
+    def filter = new TracingFilter(path)
+    filter.setNext(restlet)
+    return filter
+  }
+
+  static class TracingFilter extends Filter {
```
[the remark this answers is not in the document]
Enkelian: Seems like a good idea, I'll try to move this class

instrumentation/restlet/restlet-core-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_1/RestletInstrumentationModule.java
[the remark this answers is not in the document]
mateuszrzeszutek: Yeah that sounds good. In general, we aim to keep the instrumentation name as close to the library/artifact name as possible.

instrumentation/restlet/restlet-core-1.1/library/src/test/groovy/RestletServerTest.groovy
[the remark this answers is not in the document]
Enkelian: Successfully moved this class, decided to use thread locals. I also enabled concurrency tests to ensure it's working as intended :)

instrumentation/restlet/restlet-core-1.1/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
+  id("otel.javaagent-instrumentation")
+}
+
+muzzle {
+  pass {
+    group.set("org.restlet")
+    module.set("org.restlet")
+    versions.set("[1.0.0, 1.2.0)")
```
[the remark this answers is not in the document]
Enkelian: After discussion on Slack changed name to 1.0

instrumentation/restlet/restlet-1.0/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
+  id("otel.javaagent-instrumentation")
+}
+
+muzzle {
+  pass {
+    group.set("org.restlet")
+    module.set("org.restlet")
+    versions.set("[1.0.0, 1.2-M1)")
```
mateuszrzeszutek, line 9: How about adding
```suggestion
    versions.set("[1.0.0, 1.2-M1)")
    assertInverse.set(true)
```
to verify that other versions do not match?

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.instrumentation.api.tracer.ServerSpan;
+import org.restlet.Filter;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+
+public final class TracingFilter extends Filter {
+
+  private static final ThreadLocal<Scope> otelScope = new ThreadLocal<>();
+  private static final ThreadLocal<Context> otelContext = new ThreadLocal<>();
```
mateuszrzeszutek, line 21: Consider adding a new `ContextAndScope`/`State` class (use `@AutoValue`) and keeping it in a single thread local - you'll have one thread local lookup less for each request.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+    this.path = path;
+  }
+
+  @Override
+  protected int beforeHandle(Request request, Response response) {
+
+    Context parentContext = Context.current();
+
+    if (!instrumenter().shouldStart(parentContext, request)) {
+      return CONTINUE;
```
mateuszrzeszutek, line 34: So if there already is a server span in the context the `span.updateName(path)` call won't execute. How about structuring the code a bit differently, e.g.
```java
if (instrumenter().shouldStart(parentContext, request)) {
  // start span
  // start scope
}

Span serverSpan = ServerSpan.fromContextOrNull(currentContext);
if (serverSpan != null) {
  serverSpan.updateName(path);
}
```
This way you'll only create the span when it should be created *and* set the server span name in all cases.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletNetAttributesExtractor.java
```diff
@@ -0,0 +1,34 @@
...
+
+final class RestletNetAttributesExtractor extends NetAttributesExtractor<Request, Response> {
+  @Override
+  public String transport(Request request) {
+    return SemanticAttributes.NetTransportValues.IP_TCP;
+  }
+
+  @Override
+  public @Nullable String peerName(Request request, @Nullable Response response) {
+    return request.getHostRef().getUserInfo();
```
mateuszrzeszutek, line 22: Is this the remote (client) host name?

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpAttributesExtractor;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.restlet.data.Reference;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+
+final class RestletHttpAttributesExtractor extends HttpAttributesExtractor<Request, Response> {
+  @Override
+  protected @Nullable String method(Request request) {
```
mateuszrzeszutek, line 17: It probably isn't `@Nullable`, is it?
```suggestion
  protected String method(Request request) {
```

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+    return request.getMethod().toString();
+  }
+
+  @Override
+  protected @Nullable String url(Request request) {
+    return request.getOriginalRef().toString();
+  }
+
+  @Override
+  protected @Nullable String target(Request request) {
```
mateuszrzeszutek, line 27:
```suggestion
  protected String target(Request request) {
```

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+import org.restlet.data.Response;
+
+final class RestletHttpAttributesExtractor extends HttpAttributesExtractor<Request, Response> {
+  @Override
+  protected @Nullable String method(Request request) {
+    return request.getMethod().toString();
+  }
+
+  @Override
+  protected @Nullable String url(Request request) {
```
mateuszrzeszutek, line 22:
```suggestion
  protected String url(Request request) {
```

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+  @Override
+  protected @Nullable Long requestContentLengthUncompressed(
+      Request request, @Nullable Response response) {
+    return null;
+  }
+
+  @Override
+  protected @Nullable String flavor(Request request, @Nullable Response response) {
+    String version = (String) request.getAttributes().get("org.restlet.http.version");
+    switch (version) {
```
mateuszrzeszutek, line 67: Can `version` be null? If so, please add a nullcheck before the `switch` statement.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+    return null;
+  }
+
+  @Override
+  protected @Nullable String serverName(Request request, @Nullable Response response) {
+    return null;
+  }
+
+  @Override
+  protected @Nullable Integer statusCode(Request request, Response response) {
```
mateuszrzeszutek, line 86:
```suggestion
  protected Integer statusCode(Request request, Response response) {
```

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletSingletons.java
```diff
@@ -0,0 +1,57 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.SpanNameExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.SpanStatusExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpSpanNameExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpSpanStatusExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.net.NetAttributesExtractor;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+
+public final class RestletSingletons {
```
mateuszrzeszutek, line 21: For library instrumentations we don't really have `*Singleton` classes - instead we create `*Tracing`/`*TracingBuilder` classes that serve as the public API of the instrumentation and allow it to be customized (e.g. passing `OpenTelemetry`, custom `AttributesExtractor`s).
Please take a look at `OkHttpTracing` and its builder for an example of this.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletSingletons.java
```diff
@@ -0,0 +1,57 @@
...
+            .addAttributesExtractor(httpAttributesExtractor)
+            .addAttributesExtractor(netAttributesExtractor)
+            .newServerInstrumenter(RestletHeadersGetter.GETTER);
+  }
+
+  public static Instrumenter<Request, Response> instrumenter() {
+    return INSTRUMENTER;
+  }
+
+  public static TextMapGetter<Request> getter() {
```
mateuszrzeszutek, line 52: Is this used somewhere?

instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/ServerInstrumentation.java
```diff
@@ -0,0 +1,89 @@
...
+
+      if (scope == null) {
+        return;
+      }
+
+      scope.close();
+
+      if (context == null) {
+        return;
+      }
```
mateuszrzeszutek, line 76: No need to check that - context can't be null if `scope != null`

instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/RouteInstrumentation.java
```diff
@@ -0,0 +1,61 @@
...
+
+      Context context = currentContext();
+
+      Span serverSpan = ServerSpan.fromContextOrNull(context);
+
+      if (serverSpan == null) {
+        return;
+      }
+
+      serverSpan.updateName(pattern);
```
mateuszrzeszutek, line 58: Now that I think of it, since Restlet can be run inside a Servlet container, we should probably use `ServerSpanNaming` (and `ServletContextPath`) to correctly update the server span name (other container/servlet instrumentations might want to do that too, and using that class ensures they all cooperate correctly).

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
Enkelian: Sure, something like `ContextAndScope` from `rocketmq` instrumentation?

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
Enkelian: Thanks, my approach was indeed incorrect if some server span already existed.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletSingletons.java
Enkelian: So this class should be moved to the `javaagent` module and `library` should have `*Tracing` and `*TracingBuilder`?

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletSingletons.java
Enkelian: nope, I can delete that

instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/RouteInstrumentation.java
Enkelian: Sure, I'll try to change that

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
mateuszrzeszutek: Yeah, exactly that

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletSingletons.java
mateuszrzeszutek: Yes please 👍

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletNetAttributesExtractor.java
Enkelian: Looked into that and it wasn't the client host name. Seems like these versions of restlet allow only to retrieve the IP address, so I'm changing this to return null

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
Enkelian: `ref.getPath()` can return null, so I think it should be `@Nullable`

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHttpAttributesExtractor.java
Enkelian: Version is always set in the default implementation of Restlet's `HttpClientCall`

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletServerSpanNaming.java
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.restlet.v1_0;
+
+import io.opentelemetry.instrumentation.api.servlet.ServerSpanNameSupplier;
+import io.opentelemetry.instrumentation.api.servlet.ServletContextPath;
+
+public class RestletServerSpanNaming {
```
mateuszrzeszutek, line 11: Nit:
```suggestion
public final class RestletServerSpanNaming {
```

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletTracing.java
```diff
@@ -0,0 +1,36 @@
...
+package io.opentelemetry.instrumentation.restlet.v1_0;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+
+public final class RestletTracing {
+
+  public static RestletTracing create(OpenTelemetry openTelemetry) {
```
mateuszrzeszutek, line 15: Can you add Javadocs to public classes and methods? (Just copy them from some other `*Tracing`/`*TracingBuilder` classes)

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletServerSpanNaming.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.restlet.v1_0;
```
mateuszrzeszutek, line 6: How about moving it to a `io.opentelemetry.instrumentation.restlet.v1_0.internal` package? It needs to be public because javaagent uses it, but it shouldn't really belong to the public API of the library instrumentation (which should be minimal)

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletTracing.java
```diff
@@ -0,0 +1,36 @@
...
+    return new RestletTracingBuilder(openTelemetry);
+  }
+
+  private final Instrumenter<Request, Response> serverInstrumenter;
+
+  RestletTracing(Instrumenter<Request, Response> serverInstrumenter) {
+    this.serverInstrumenter = serverInstrumenter;
+  }
+
+  public TracingFilter newFilter(String path) {
```
mateuszrzeszutek, line 29: Use the interface, not the implementation:
```suggestion
  public Filter newFilter(String path) {
```

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletTracingBuilder.java
```diff
@@ -0,0 +1,61 @@
...
+        new RestletNetAttributesExtractor();
+
+    Instrumenter<Request, Response> instrumenter =
+        Instrumenter.<Request, Response>newBuilder(
+                openTelemetry, INSTRUMENTATION_NAME, spanNameExtractor)
+            .setSpanStatusExtractor(spanStatusExtractor)
+            .addAttributesExtractor(httpAttributesExtractor)
+            .addAttributesExtractor(netAttributesExtractor)
+            .addAttributesExtractors(additionalExtractors)
+            .newServerInstrumenter(RestletHeadersGetter.GETTER);
```
mateuszrzeszutek, line 57: You don't need the `GETTER` singleton, you can just create a new instance:
```suggestion
            .newServerInstrumenter(new RestletHeadersGetter());
```
Those setter/getter singletons are a mostly remnant from the BaseTracer API

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,65 @@
...
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.api.servlet.ServerSpanNaming;
+import org.restlet.Filter;
+import org.restlet.data.Request;
+import org.restlet.data.Response;
+
+public final class TracingFilter extends Filter {
```
mateuszrzeszutek, line 18: This class could be package-private - there's no need to expose it in the public API of the library instrumentation.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    Context parentContext = Context.current();
+
+    if (instrumenter.shouldStart(parentContext, request)) {
+      Context context = instrumenter.start(parentContext, request);
+      Scope scope = context.makeCurrent();
+      otelContextAndScope.set(ContextAndScope.create(context, scope));
+    }
+
+    ServerSpanNaming.updateServerSpanName(
+        otelContextAndScope.get().getContext(),
```
mateuszrzeszutek, line 41: A little bit of performance-related nitpicking: to avoid that one thread local lookup you could create a `Context currentContext` variable with initial value `parentContext`, updated to `context` inside the if.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,65 @@
...
+
+    return CONTINUE;
+  }
+
+  @Override
+  protected void afterHandle(Request request, Response response) {
+
+    ContextAndScope contextAndScope = otelContextAndScope.get();
+
+    if (contextAndScope.getScope() == null) {
```
mateuszrzeszutek, line 54: `contextAndScope.getScope()` will always be non null - you should probably check if `contextAndScope` is non null instead.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,65 @@
...
+      return;
+    }
+
```
mateuszrzeszutek, line 57: Remember to clear the thread local once you retrieve the value:
```suggestion
    otelContextAndScope.remove();

```

mateuszrzeszutek (commented):
Looks nice! I left just a few (mostly minor) comments.

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletTracing.java
Enkelian: Sure!

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
Enkelian: Moved `context` declaration before if, with `parentContext` assigned to it

mateuszrzeszutek (approved):
LGTM 👍 
Thanks!

instrumentation/restlet/restlet-1.0/library/src/test/groovy/io/opententelemetry/instrumentation/restlet/v1_0/RestletServerTest.groovy
```diff
@@ -0,0 +1,32 @@
...
+  Restlet wrapRestlet(Restlet restlet, String path){
+    RestletTracing tracing = RestletTracing.create(openTelemetry)
+    def filter = tracing.newFilter(path)
+    filter.setNext(restlet)
+    return filter
+  }
+
+  @Override
+  boolean testException() {
+    false //Filter's afterHandle does not execute if exception was thrown in the next restlet
```
trask, line 30: if `afterHandle` is not executed, will the Scope that was started in `beforeHandle` be left open (and the Context left bound to the thread)?

instrumentation/restlet/restlet-1.0/testing/src/main/groovy/io/opentelemetry/instrumentation/restlet/v1_0/AbstractRestletServerTest.groovy
```diff
@@ -0,0 +1,172 @@
...
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.ERROR
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.EXCEPTION
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.INDEXED_CHILD
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.NOT_FOUND
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.PATH_PARAM
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.QUERY_PARAM
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.REDIRECT
+import static io.opentelemetry.instrumentation.test.base.HttpServerTest.ServerEndpoint.SUCCESS
+
+abstract class AbstractRestletServerTest extends HttpServerTest<Server> {
```
trask, line 32: 👍

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletHeadersGetter.java
```diff
@@ -0,0 +1,22 @@
...
+
+  @Override
+  public Iterable<String> keys(Request carrier) {
+    return HeadersAdapter.getHeaders(carrier).getNames();
+  }
+
+  @Override
+  public String get(Request carrier, String key) {
+    return HeadersAdapter.getValue(carrier, key);
+  }
```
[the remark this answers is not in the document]
Enkelian: sure, I think it's not necessary to keep them separate

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletTracing.java
```diff
@@ -0,0 +1,47 @@
...
+   * implementations.
+   */
+  public Filter newFilter(String path) {
+    return new TracingFilter(serverInstrumenter, path);
+  }
+
+  /** Returns a server {@link Instrumenter}. */
+  public Instrumenter<Request, Response> getServerInstrumenter() {
+    return serverInstrumenter;
+  }
```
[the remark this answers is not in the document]
Enkelian: I'm planning on adding client instrumentation in the future, so I'd prefer to leave the current naming

instrumentation/restlet/restlet-1.0/library/src/test/groovy/io/opententelemetry/instrumentation/restlet/v1_0/RestletServerTest.groovy
Enkelian: after some investigation and looking into Restlet's `StatusFilter` I found a way to make `afterHandle` execute even if an exception was thrown, so the Scope will be safely closed in every case now

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
```diff
@@ -0,0 +1,80 @@
...
+    this.instrumenter = instrumenter;
+    this.path = path;
+  }
+
+  @Override
+  public int doHandle(Request request, Response response) {
+    try {
+      super.doHandle(request, response);
+    } catch (Throwable t) {
+      response.setStatus(new Status(Status.SERVER_ERROR_INTERNAL, t));
```
mateuszrzeszutek, line 36: You could implement the whole instrumentation inside `doHandle` to avoid modifying the response object

mateuszrzeszutek (approved):

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/TracingFilter.java
Enkelian: Without setting status this way, the throwable is not propagated and we get a response with status of 200, not 500. My approach based on [StatusFilter](http://repository.transtep.com/repository/thirdparty/restlet/1.1.6/docs/nre/com/noelios/restlet/StatusFilter.html#doHandle(org.restlet.data.Request,%20org.restlet.data.Response)). 
I can move instrumentation code to `doHandle`, leaving this line as it is.

mateuszrzeszutek: I see. How about you move the instrumentation code to `doHandle` and use the `StatusFilter` in library instrumentation tests? We generally want an instrumentation to be as transparent as possible - I agree that returning 200 for unhandled exceptions is a bit weird and unintuitive, but as instrumentation authors we should avoid changing this default behavior on our whim. If a user wants to return 500s instead then they're going to use `StatusFilter` anyway.

Enkelian: Sure, thanks for the explanation!

mateuszrzeszutek (approved):

instrumentation/restlet/restlet-1.0/library/src/main/java/io/opentelemetry/instrumentation/restlet/v1_0/RestletTracing.java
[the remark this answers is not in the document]
trask: make sense 👍

trask (approved):
thx @Enkelian!

Issue #3926, opened 2021-08-24
app using codehaus.plexus  failed to generate bean with java instrumentation


tydhot:
<img width="719" alt="image" src="https://user-images.githubusercontent.com/27889201/130615343-7031a642-bc17-49c5-88e5-bdbe09af0a1f.png">
And here is the class file:  https://github.com/dianping/cat/blob/d8033c988b825f35fc06c20660f4aed927e43465/cat-client/src/main/java/com/dianping/cat/message/io/TcpSocketSender.java

It is the config file: https://github.com/dianping/cat/blob/d8033c988b825f35fc06c20660f4aed927e43465/cat-client/src/main/resources/META-INF/plexus/components.xml.
I am little confused why there will be a field named"__opentelemetryContext$java$lang$Runnable" in the log, maybe I should add "-Dotel.javaagent.experimental.field-injection.enabled=false"?
btw, without java instrumentation the app is ok.

mateuszrzeszutek:
Yep, those fields are added by the `FieldBackedProvider`. Setting `"otel.javaagent.experimental.field-injection.enabled"` to false should help, it's worth testing that.
I'm not familiar at all with the codehaus-plexus plugin/software - but is it okay that it tries to map transient fields? All those fields added by the javaagent are marked as transient.

trask:
@tydhot it would be great if you could test once #3928 is merged and the SNAPSHOT build is published, to see if that helps

laurit:
@trask  as far as I can tell #3928 won't help for this :( Though I have an idea how to make this work.

PR #3948, opened 2021-08-25
Hide our generated fields and methods from reflection

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3926
Hide fields and methods added by `FieldBackedProvider` from reflection. Note this solution will not work on early ibm 8 vms that do not call `Reflection.filterFields` from `Class.getDeclaredFields`.


trask (approved):
pure magic 🧙‍♂️✨

mateuszrzeszutek (approved):

PR #3974, opened 2021-08-26
Simpler test agent

iNikem:
I wanted to try this idea for a long time and at last here it is. Would like to have your feedback if we want to actually merge it.

The idea is to have a really bare-bone agent, which does not actually contain anything but is still able to load extensions.

Having that, we can greatly simplify `agent-for-testing` by converting it into extension and embedding into that bare bone agent. This removes the need to do any shadowing and relocation and isolation in `agent-for-testing` build script, which is a duplicate from the main javaagent build script.


iNikem:
@trask @anuraaga any feedback from you?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/RemappingUrlConnection.java
```diff
@@ -31,7 +31,7 @@
               "#io.opentelemetry.javaagent.shaded.io.opentelemetry.context"),
           rule(
               "#io.opentelemetry.instrumentation",
-              "#io.opentelemetry.javaagent.shaded.io.opentelemetry.instrumentation"),
+              "#io.opentelemetry.javaagent.shaded.instrumentation"),
```
trask, line 34: oh you found this! #4064

testing/agent-for-testing/build.gradle.kts
```diff
@@ -1,128 +1,54 @@
...
-}
-
-fun CopySpec.isolateClasses(jars: Iterable<File>) {
-  jars.forEach {
-    from(zipTree(it)) {
-      // important to keep prefix "inst" short, as it is prefixed to lots of strings in runtime mem
-      into("inst")
-      rename("(^.*)\\.class\$", "\$1.classdata")
-      // Rename LICENSE file since it clashes with license dir on non-case sensitive FSs (i.e. Mac)
-      rename("""^LICENSE$""", "LICENSE.renamed")
```
trask, line 125: nice

trask (approved):
I get nervous 😬 any time we touch shadow...

but this looks great 👍

anuraaga (approved):

Issue #3730, opened 2021-07-30
Can't listen error info by RequestListener in Instrumenter lifecycle end


zmapleshine:
**Is your feature request related to a problem? Please describe.**
I am trying to add "hasError” tag to my custom component metrics by RequestListener, but this interface has no way to transfer ``Throwable``  info to its implementation class. 

At first I wanted to pass exceptions by RESPONSE, and I needed to modify the generics of the instrument class but I thought it was important to make a distinction.

```java
/** Listener method that is called at the end of a request. */
  void end(Context context, Attributes responseAttributes);
```

**Describe the solution you'd like**

add throwable param to "end" method in interface ``RequestListener``.

**Describe alternatives you've considered**

None

**Additional context**

None

zmapleshine:
> **Describe alternatives you've considered**
> 
> None

I think, adding a default method in RequestListener to pass throwable is the best option?

zmapleshine:
Maybe ``RequestListener``  interface is  better called "InstrumentListener"?

PR #3988, opened 2021-08-27
Add error parameter to EndTimeExtractor and AttributesExtractor#onEnd()

mateuszrzeszutek:
I've added a `@Nullable Throwable error` parameter to `EndTimeExtractor` and `AttributesExtractor#onEnd()`. Now, along with the `SpanStatusExtractor`, every method in the Instrumenter API that accepts a `RESPONSE` also accepts a `Throwable`. The operation can end with either a `RESPONSE` or an `error` being passed to the instrumenter so I thought it makes sense to always pass both of these to have a complete view of how it ended.

This PR resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3730. It is now possible to extract custom attributes in an `AttributesExtractor` and they'll be passed along to the `RequestListener`, thus making it possible to use those attributes in custom metrics.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/EndTimeExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/EndTimeExtractor.java
@@ -17,5 +17,5 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 public interface EndTimeExtractor<REQUEST, RESPONSE> {
 
   /** Returns the timestamp marking the end of the response processing. */
-  Instant extract(REQUEST request, @Nullable RESPONSE response);
+  Instant extract(REQUEST request, @Nullable RESPONSE response, @Nullable Throwable error);
 }
--- a/instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/MethodSpanAttributesExtractor.java
+++ b/instrumentation-api-annotation-support/src/main/java/io/opentelemetry/instrumentation/api/annotation/support/MethodSpanAttributesExtractor.java
@@ -57,7 +57,10 @@ public final class MethodSpanAttributesExtractor<REQUEST, RESPONSE>
 
   @Override
   protected void onEnd(
-      AttributesBuilder attributes, REQUEST request, @Nullable RESPONSE response) {}
+      AttributesBuilder attributes,
+      REQUEST request,
+      @Nullable RESPONSE response,
+      @Nullable Throwable error) {}
 
   /**
    * Creates a binding of the parameters of the traced method to span attributes.
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
@@ -15,10 +15,10 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 /**
  * Extractor of {@link io.opentelemetry.api.common.Attributes} for a given request and response.
  * Will be called {@linkplain #onStart(AttributesBuilder, Object) on start} with just the {@link
- * REQUEST} and again {@linkplain #onEnd(AttributesBuilder, Object, Object) on end} with both {@link
- * REQUEST} and {@link RESPONSE} to allow populating attributes at each stage of a request's
- * lifecycle. It is best to populate as much as possible in {@link #onStart(AttributesBuilder,
- * Object)} to have it available during sampling.
+ * REQUEST} and again {@linkplain #onEnd(AttributesBuilder, Object, Object, Throwable) on end} with
+ * both {@link REQUEST} and {@link RESPONSE} to allow populating attributes at each stage of a
+ * request's lifecycle. It is best to populate as much as possible in {@link
+ * #onStart(AttributesBuilder, Object)} to have it available during sampling.
  *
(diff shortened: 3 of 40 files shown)


zmapleshine:
Hey @mateuszrzeszutek ,thank you for resolve the problem in #3730 .But...I do not understand the question about the label value of a metrics.
Because from the design of Instrumentation API module, it is almost certain that the label value of a metrics comes from attributes, but it is a "processed" value, and sometimes such a value from attributes is not suitable as label of a metrics . Similarly, labels that were put into attributes for use in metrics do not have to be a span tag .
I think if it is possible to pass the original object to the hook function as well? Or is there a clever way to combine trace and metrics organically without coupling?

zmapleshine:
> but it is a "processed" value, and sometimes such a value from attributes is not suitable as label of a metrics . Similarly, labels that were put into attributes for use in metrics do not have to be a span tag .
>

In a practical scenario, in the monitoring process of the Lettuce client, commands and parameters are usually concatenated together as span labels:

```
sql.statement: mget key1,key2...
```

Then, attributes must have a key named "sql.statement", but this value is not suitable for a label of redis metric, because the cardinality is too large and "mget" should be used instead.

trask:
@mateuszrzeszutek this one has merge conflict

mateuszrzeszutek:
> Because from the design of Instrumentation API module, it is almost certain that the label value of a metrics comes from attributes, but it is a "processed" value, and sometimes such a value from attributes is not suitable as label of a metrics . Similarly, labels that were put into attributes for use in metrics do not have to be a span tag .

Yeah that's true. In case of metrics, the [View API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view) should solve that problem, but we don't really have anything like that for spans (well, you can remove attributes in the collector I guess).

> I think if it is possible to pass the original object to the hook function as well?

I wanted to avoid passing the response & exception everywhere. I think that those `endAttributes` passed to the `RequestListener` are basically a derivative of the response, they should contain all useful information. Passing a response object to the listener would duplicate some of the `AttributesExtractor` logic and responsibility. I'd like to get @anuraaga's opinion on that though - whether what I'm describing here makes sense or not.

> In a practical scenario, in the monitoring process of the Lettuce client, commands and parameters are usually concatenated together as span labels:
> 
> ```
> sql.statement: mget key1,key2...
> ```
> 
> Then, attributes must have a key named "sql.statement", but this value is not suitable for a label of redis metric, because the cardinality is too large and "mget" should be used instead.

That's why almost all (if not all) database clients' instrumentations emit `db.operation` attribute - in case of your example it'd contain just `"mget"` and would be a perfect choice for a metric label/attribute.

zmapleshine:
> That's why almost all (if not all) database clients' instrumentations emit `db.operation` attribute - in case of your example it'd contain just `"mget"` and would be a perfect choice for a metric label/attribute.

@mateuszrzeszutek 
So how do you extract the URI from the attributes of HTTP client instrumentation?  : ) :) :) 

 It might be given a full URL like "http://10.x.x.x/user/123456?sign=xxxxx&...", but it's not appropriate for metrics, which typically require a tag like "/user/{id}" (perhaps splicing together the service name to buid a new tag called "router", perhaps a slightly inappropriate example, but I mean the loss of the original request object may also lose the accuracy of the metric tag and even increase the difficulty of obtaining the tag value.)

anuraaga:
The goal of instrumentation is to collect data - how the data gets transformed aside from that is signal specific. Traces generally follow a pattern of full data on a sample of traces. So spans just accept the attributes as is.

Metrics instead don't sample but aggregate - the metric view configuration configures the aggregation. Just like they aggregate points to a sum, the views will also be used to pick correct attributes - attribute selection is a sort of aggregation after all. The intention is definitely not to record high cardinality attributes. But the metrics SDK is currently in its infancy so we are still working on that piece of machinery. Hopefully that makes sense.

anuraaga (approved):
Mapping error to semantic attributes makes a lot of sense.

Less sure about end time ;) I guess it's symmetry but don't expect that to ever be useful.

mateuszrzeszutek:
> Less sure about end time ;) I guess it's symmetry but don't expect that to ever be useful.

Yeah, I added it there mostly for symmetry - honestly I have no idea how you can extract a useful end time from an exception.

trask:
> > Less sure about end time ;) I guess it's symmetry but don't expect that to ever be useful.
> 
> Yeah, I added it there mostly for symmetry - honestly I have no idea how you can extract a useful end time from an exception.

is the symmetry "everywhere we pass RESPONSE, we also pass error"?

mateuszrzeszutek:
> is the symmetry "everywhere we pass RESPONSE, we also pass error"?

Yep, exactly that.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/EndTimeExtractor.java
```diff
@@ -17,5 +17,5 @@
 public interface EndTimeExtractor<REQUEST, RESPONSE> {
 
   /** Returns the timestamp marking the end of the response processing. */
-  Instant extract(REQUEST request, @Nullable RESPONSE response);
+  Instant extract(REQUEST request, @Nullable RESPONSE response, @Nullable Throwable error);
```
trask, line 20:
```suggestion
  // passing error for consistency with other extractors of always passing response and error together
  // (we don't know of a use case yet for extracting end time from error)
  Instant extract(REQUEST request, @Nullable RESPONSE response, @Nullable Throwable error);
```

PR #3990, opened 2021-08-27
Add documentation for suppression by type flag

### LINKED ISSUE #3964: Add documentation for otel.instrumentation.experimental.outgoing-span-suppression-by-type
Although this is an experimental property, I assume we still want people to try it out and provide as feedback. For that they need to know about this option :)

/cc @lmolkova

lmolkova:
Fixes #3964

### DIFF
--- a/docs/suppressing-instrumentation.md
+++ b/docs/suppressing-instrumentation.md
@@ -122,5 +122,5 @@ You can suppress all auto instrumentations but have support for manual instrumen
 
 ## Enable instrumentation suppression by type
 
-You can suppress duplicate instrumentation (e.g. multiple layers of HTTP clients) with `-Dotel.instrumentation.experimental.outgoing-span-suppression-by-type=true`.
-With this flag set to `false` (default) all nested `CLIENT` spans are suppressed. With this flag set to `true`, you can have nested `CLIENT` spans of different types, e.g. HTTP `CLIENT` span nested under DB `CLIENT` span.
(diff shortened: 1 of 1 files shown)


docs/suppressing-instrumentation.md
```diff
@@ -119,3 +119,8 @@ so that we can try to come up with a better solution to address your need.
 
 You can suppress all auto instrumentations but have support for manual instrumentation with `@WithSpan` and normal API interactions by using
 `-Dotel.instrumentation.common.default-enabled=false -Dotel.instrumentation.opentelemetry-annotations.enabled=true`
+
+## Enable instrumentation suppression by type
+
+You can suppress duplicate instrumentation (e.g. multiple layers of HTTP clients) with `-Dotel.instrumentation.experimental.outgoing-span-suppression-by-type=true`.
```
mateuszrzeszutek, line 125: Saying that `You can suppress duplicate instrumentation` with setting that flag to `true` is a bit misleading, cause the default behavior is actually stricter since it suppresses all but one CLIENT spans, no matter their actual type. How about sth like

```markdown
By default the agent suppresses all nested CLIENT instrumentations, without discerning their actual type (e.g. a database `CLIENT` using an HTTP client that generates `CLIENT` spans too).
By setting `-Dotel.instrumentation.experimental.outgoing-span-suppression-by-type=true` you can enable a more sophisticated suppression strategy: only `CLIENT` spans of the same types (e.g. DB, HTTP, RPC) will be suppressed.
```

trask (approved):
thx!

Issue #3998, opened 2021-08-27
Windows liberty smoke test sporadically failing


trask:
It's running out of memory on the whole Github Runner (not just memory inside of the smoke test container that we spin up):

```
Daemon pid: 1752
  log file: C:\Users\runneradmin\.gradle\daemon\7.1\daemon-1752.out.log
----- Last  20 lines from daemon log file - daemon-1752.out.log -----
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 883776 bytes for Chunk::new
```

EDIT: and here's one that ran out of memory inside the smoke test container:

```
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (malloc) failed to allocate 32744 bytes for ChunkPool::allocate
# An error report file with more information is saved as:
# C:\server\usr\servers\defaultServer\hs_err_pid11468.log

[error occurred during error reporting , id 0xc0000005]

Picked up JAVA_TOOL_OPTIONS: -Xmx1g -javaagent:/opentelemetry-javaagent.jar
```

trask:
Hopefully fixed by #4000, will re-open if continue seeing it

PR #4000, opened 2021-08-27
Use --no-daemon on windows builds

trask:
Trying something for #3998, as those sporadic failures are becoming more of a nuisance.


trask:
I haven't been able to get windows smoke tests to fail in this PR (temporarily exploded the matrix to make it run lots of duplicates in parallel), so marking this ready for review

iNikem (approved):

PR #4001, opened 2021-08-28
Add strict context stressor testing option

trask:
this is @laurit's code from slack, but it's so useful I think it deserves to be a permanent testing option


trask:
pinging for a review 😄

anuraaga (approved):

mateuszrzeszutek (approved):

PR #4002, opened 2021-08-28
Fix netty strict context checks

trask:
most of the netty-4.0 changes were just syncing it with netty-4.1 in order to resolve one of its strict context checks (it was ending the span too early on the first http response chunk)

### DIFF
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/NettyFutureInstrumentation.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/NettyFutureInstrumentation.java
@@ -61,7 +61,15 @@ public class NettyFutureInstrumentation implements TypeInstrumentation {
     public static void wrapListener(
         @Advice.Argument(value = 0, readOnly = false)
             GenericFutureListener<? extends Future<?>> listener) {
-      listener = FutureListenerWrappers.wrap(Java8BytecodeBridge.currentContext(), listener);
+      // wrapping our "end" listener leads to strict context leak failures since there will be an
+      // active scope when we call "end"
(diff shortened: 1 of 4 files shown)


instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/NettyFutureInstrumentation.java
```diff
@@ -61,7 +61,15 @@ public void transform(TypeTransformer transformer) {
     public static void wrapListener(
         @Advice.Argument(value = 0, readOnly = false)
             GenericFutureListener<? extends Future<?>> listener) {
-      listener = FutureListenerWrappers.wrap(Java8BytecodeBridge.currentContext(), listener);
+      // wrapping our "end" listener leads to strict context leak failures since there will be an
```
anuraaga, line 64: I think it's good to add the phrasing "We only want to wrap user callbacks"

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/NettyFutureInstrumentation.java
```diff
@@ -61,7 +61,15 @@ public void transform(TypeTransformer transformer) {
...
+      // active scope when we call "end"
+      // wrapping internal netty listeners also leads to strict context leak failures since some of
+      // those are called after our "end" listener
+      String listenerClassName = listener.getClass().getName();
+      if (!listenerClassName.startsWith("io.opentelemetry.javaagent.")
```
anuraaga, line 69: ClassValue? :)

anuraaga (approved):

mateuszrzeszutek (approved):

PR #4004, opened 2021-08-28
Add span names cache

trask:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
```diff
@@ -5,10 +5,21 @@
 
 package io.opentelemetry.instrumentation.api.tracer;
 
+import io.opentelemetry.instrumentation.api.caching.Cache;
 import java.lang.reflect.Method;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 public final class SpanNames {
+
+  private static final ClassValue<Cache<String, String>> spanNameCaches =
```
anuraaga, line 14: Can you check whether instrumentation have their own cache to not double cache?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
```diff
@@ -5,10 +5,21 @@
...
+      new ClassValue<Cache<String, String>>() {
+        @Override
+        protected Cache<String, String> computeValue(Class<?> clazz) {
+          // should be naturally bounded, but setting a limit just in case
```
anuraaga, line 18: This seems pretty safe to me I'd probably just use ConcurrentHashMap. I think I've seen us using it in other similar caches

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/SpanNames.java
trask: sure, just looked through all callers, didn't find any secondary caches

anuraaga (approved):

mateuszrzeszutek (approved):

PR #4005, opened 2021-08-28
CallDepthThreadLocalMap simplification

trask:


trask:
nm, don't know what i was thinking 🤔

Issue #3699, opened 2021-07-28
Remove unnecessary URL construction


trask:
Constructing URL objects is a bit costly (in addition to memory, it does a lot of validation), and we construct them in various http server and http client instrumentations, generally in order to populate `http.url` attribute. This could probably be optimized by adding some kind of "uri builder" to our api that can be used instead.

anuraaga:
I think the intent of the semantic conventions is to only return a URL if we have a URL, or otherwise return the parts, target, host, etc. Though not sure if this just makes it too confusing when writing instrumentation.

PR #4008, opened 2021-08-28
Optimize tomcat uri construction

### LINKED ISSUE #3700: Don't capture http.scheme, http.host and http.target if already capturing http.url
`http.scheme`, `http.host` and `http.target` are redundant if already capturing `http.url`.

While more telemetry is good, this redundancy doesn't seem worth the (small) additional performance and storage costs.

trask:
Part of #3699

At least until we resolve #3700

(this particular one is showing up in my benchmarks)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/UriBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+    if (!isDefaultPort && serverPort != -1) {
+      // +6 is space for ":" and max port number (65535)
+      length += 6;
+    }
+    if (query != null) {
+      // the +1 is space for "?"
+      length += 1 + query.length();
+    }
+
+    StringBuilder url = new StringBuilder(length);
```
laurit, line 33: I initially thought that as this does not encode any characters it could result in an invalid url (would that actually matter?), but javadoc for `HttpServletRequest.getQueryString` states that "The value is not decoded by the container" and the same for `getRequestURI` so as far as I understand this looks good.

laurit (approved):

anuraaga (approved):

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/UriBuilder.java
trask: good point, I added a TODO to this method to investigate and document encoding requirement

PR #4017, opened 2021-08-30
Add instrumentation for Quartz 2.0

### LINKED ISSUE #3967: Quartz Instrumentation Support
I'd like to open a feature request to add instrumentation support for the [Quartz scheduler framework](http://www.quartz-scheduler.org/). Quartz is an open-source, highly scalable, asynchronous job scheduling framework with very high adoption (as measured by its [GitHub usage](https://github.com/quartz-scheduler/quartz/network/dependents?package_id=UGFja2FnZS0xNzk3NzkxMDM%3D)). I believe the instrumentation for Quartz would be something like wrapping each of the jobs with `Internal` spans.

I unfortunately do not have bandwidth to contribute this myself, but would like to see the level of community support for such an instrumentation and if anyone else would be able to.

anuraaga:
Quartz 2.0 was released 11 years ago so should be a reasonable target - it has significant API changes from 1.x that affect instrumentation, notably getting the job name from a `JobDetail`.

Fixes #3967


instrumentation/quartz-2.0/library/src/main/java/io/opentelemetry/instrumentation/quartz/v2_0/QuartzTracing.java
```diff
@@ -0,0 +1,59 @@
...
+  QuartzTracing(JobListener jobListener) {
+    this.jobListener = jobListener;
+  }
+
+  private final JobListener jobListener;
+
+  /**
+   * Configures the {@link Scheduler} to enable tracing of jobs.
+   *
+   * <p><strong>NOTE:</strong> If there are job listeners already registered on the Scheduler that
```
anuraaga, line 36: Note that unlike something like AWS SDK instrumentation, where a bad executioninterceptor doesn't kill the entire SDK experience, a job listener that throws in the begin listener will cause the job itself to fail. Presumably quartz users or implementations of quartz listeners do know that they cannot throw exceptions or have the scheduler completely broken, not just tracing.

anuraaga:
Build failure is OpenJ9 flake, will wait for comments / potential commits instead of rerunning

instrumentation/quartz-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/quartz/v1_7/QuartzIgnoredTypesConfigurer.java
```diff
@@ -0,0 +1,22 @@
...
+import io.opentelemetry.instrumentation.api.config.Config;
+import io.opentelemetry.javaagent.extension.ignore.IgnoredTypesBuilder;
+import io.opentelemetry.javaagent.extension.ignore.IgnoredTypesConfigurer;
+
+@AutoService(IgnoredTypesConfigurer.class)
+public class QuartzIgnoredTypesConfigurer implements IgnoredTypesConfigurer {
+
+  @Override
+  public void configure(Config config, IgnoredTypesBuilder builder) {
+    // Quartz executes jobs themselves in a synchronous, there's no reason to propagate context
```
mateuszrzeszutek, line 18: "In a synchronous"? Do you mean that jobs themselves are executed synchronously?

instrumentation/quartz-2.0/library/src/main/java/io/opentelemetry/instrumentation/quartz/v2_0/TracingJobListener.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public String getName() {
+    return TracingJobListener.class.getName();
+  }
+
+  @Override
+  public void jobExecutionVetoed(JobExecutionContext jobExecutionContext) {
+    // TODO(anuraaga): Consider adding an attribute for vetoed jobs.
```
mateuszrzeszutek, line 31: Is a veto similar to cancelling? We have some `<someting>.cancelled` attributes being added in various instrumentations.

instrumentation/quartz-2.0/library/src/main/java/io/opentelemetry/instrumentation/quartz/v2_0/TracingJobListener.java
```diff
@@ -0,0 +1,73 @@
...
+    if (context == null) {
+      // Would only happen if we didn't start a span (maybe a previous joblistener threw an
+      // exception before ours could process the start event).
+      return;
+    }
+
+    Throwable userError = error;
+    while (userError instanceof SchedulerException) {
+      userError = ((SchedulerException) userError).getUnderlyingException();
+    }
```
mateuszrzeszutek, line 70: Should we use an `ErrorCauseExtractor` for that?

instrumentation/quartz-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/quartz/v1_7/QuartzInstrumentationModule.java
```diff
@@ -0,0 +1,25 @@
...
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.Collections;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class QuartzInstrumentationModule extends InstrumentationModule {
+
+  public QuartzInstrumentationModule() {
+    super("quartz", "quartz-1.7");
```
mateuszrzeszutek, line 18: Shouldn't it be 2.0? Same in the package name

```suggestion
    super("quartz", "quartz-2.0");
```

mateuszrzeszutek (approved):

instrumentation/quartz-2.0/library/src/main/java/io/opentelemetry/instrumentation/quartz/v2_0/TracingJobListener.java
anuraaga: Let me leave this as a todo - I think cancelled is about right but lacking a semantic convention and significant knowledge of quartz will stick with simple for now.

PR #4027, opened 2021-08-30
[overhead] ignore temp results files

breedx-splk:
So in the latest [failure mode](https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/3458373282?check_suite_focus=true), we couldn't switch to the `gh-pages` branch to commit our results because they would have been overwritten.

So this sets up a `.gitignore` that will allow the results to be safely ignored when the action switches to the target commit branch.  🤞🏻

### DIFF
--- a/benchmark-overhead/.gitignore
+++ b/benchmark-overhead/.gitignore
@@ -1,0 +1,2 @@
+results
+!results/README.md


mateuszrzeszutek (approved):

laurit (approved):

trask (commented):
I don't quite understand, won't this cause the fresh results to be overwritten by the gh-pages results when switching to gh-pages?

breedx-splk:
Ugh, I think you're right @trask .  Thing were so much simpler when branch protection was lax. Lemme try a different approach with a temp dir instead.

PR #4029, opened 2021-08-30
Transition spring batch 3.0 module to Instrumenter API

jack-berg:
Contributing to Issue #2713.


instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionInstrumenter.java
```diff
@@ -0,0 +1,59 @@
...
+          .newInstrumenter();
+
+  public static Instrumenter<ChunkContextAndBuilder, Void> chunkExecutionInstrumenter() {
+    return INSTRUMENTER;
+  }
+
+  private static void extractSpanLinks(
+      SpanLinksBuilder spanLinks, Context unused, ChunkContextAndBuilder request) {
+    // The context passed will be Context.root() if shouldCreateRootSpanForChunk()
+    Context parentContext = currentContext();
```
jack-berg, line 38: This is a bit strange. Rather than using the context passed as an argument, I need to get the current context. This is necessary because `TracingChunkExecutionListener#beforeChunk` needs to conditionally set the span as a root span based off `shouldCreateRootSpanForChunk`. The only way (as far as I know) for `TracingChunkExecutionListener#beforeChunk` to start a span as a root span is by passing `Context.root()` as the parent context. 

But if `Context.root()` is passed as the parent context, then this span link extractor code will not be able to access the original parent context necessary to add a span link.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemInstrumenter.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.batch.core.scope.context.ChunkContext;
+
+public class ItemInstrumenter {
+
+  static final String ITEM_OPERATION_READ = "ItemRead";
+  static final String ITEM_OPERATION_WRITE = "ItemWrite";
+  static final String ITEM_OPERATION_PROCESS = "ItemProcess";
+
+  private static final Instrumenter<String, Void> INSTRUMENTER =
+      Instrumenter.<String, Void>newBuilder(
```
jack-berg, line 23: This is a bit strange. There's no obvious `REQUEST` type for the instrumenter. The span name consists of data from the `ChunkContext` and a `String itemOperationName`. I could make a value class that combines these, but I figured since there is no attribute extraction I might as well use a `String` as the `REQUEST` type, and use `Function.identity()` (i.e. `str -> str`) as the span name extractor.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/step/StepExecutionInstrumenter.java
```diff
@@ -0,0 +1,32 @@
...
+  private static final Instrumenter<StepExecution, Void> INSTRUMENTER =
+      Instrumenter.<StepExecution, Void>newBuilder(
+              GlobalOpenTelemetry.get(), instrumentationName(), StepExecutionInstrumenter::spanName)
+          .newInstrumenter();
+
+  public static Instrumenter<StepExecution, Void> stepExecutionInstrumenter() {
+    return INSTRUMENTER;
+  }
+
+  public static String spanName(StepExecution stepExecution) {
```
mateuszrzeszutek, line 25:
```suggestion
  private static String spanName(StepExecution stepExecution) {
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/TracingChunkExecutionListener.java
```diff
@@ -27,32 +31,40 @@ public TracingChunkExecutionListener(
 
   @Override
   public void beforeChunk(ChunkContext chunkContext) {
-    Context context = tracer().startSpan(chunkContext, builderClass);
+    Context parentContext = shouldCreateRootSpanForChunk() ? Context.root() : currentContext();
```
mateuszrzeszutek, line 34: Please use the Java 8 bridge for `root()` too:
```suggestion
    Context parentContext = shouldCreateRootSpanForChunk() ? rootContext() : currentContext();
```

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkExecutionInstrumenter.java
mateuszrzeszutek: I don't think there's a better way to solve that - your version is perfectly fine.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemInstrumenter.java
mateuszrzeszutek: Yeah that's fine - if we ever need to extract some attributes (I doubt we'll add any to the item-level spans - they're only here for the trace tree structure) we can always change that.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/ItemInstrumenter.java
```diff
@@ -0,0 +1,55 @@
...
+   * Item-level listeners do not receive chunk/step context as parameters. Fortunately the whole
+   * chunk always executes on one thread - in Spring Batch chunk is almost synonymous with a DB
+   * transaction; this makes {@link ChunkContext} a good candidate to be stored in {@link Context}.
+   */
+  public static Context startChunk(ChunkContext chunkContext) {
+    return Context.current().with(CHUNK_CONTEXT_KEY, chunkContext);
+  }
+
+  public static ChunkContext getChunkContext() {
+    return Context.current().get(CHUNK_CONTEXT_KEY);
```
mateuszrzeszutek, line 51: You could pass the current `Context` as method param to avoid an additional lookup.

jack-berg:
Thanks for the feedback!

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/chunk/ChunkSingletons.java
```diff
@@ -0,0 +1,57 @@
...
+          .newInstrumenter();
+
+  public static Instrumenter<ChunkContextAndBuilder, Void> chunkInstrumenter() {
+    return INSTRUMENTER;
+  }
+
+  private static void extractSpanLinks(
+      SpanLinksBuilder spanLinks, Context unused, ChunkContextAndBuilder request) {
+    // The context passed will be Context.root() if shouldCreateRootSpanForChunk()
+    Context parentContext = currentContext();
```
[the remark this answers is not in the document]
jack-berg: Better yet, I can remove the span link extractor altogether if `!shouldCreateRootSpanForChunk()`.

instrumentation/spring/spring-batch-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/batch/item/SimpleChunkProcessorInstrumentation.java
```diff
@@ -44,32 +49,35 @@ public void transform(TypeTransformer transformer) {
...
-      if (!shouldTraceItems()) {
+        @Advice.Local("otelScope") Scope scope,
+        @Advice.Local("otelItem") String item) {
+      ChunkContext chunkContext = getChunkContext(currentContext());
+      if (chunkContext == null || !shouldTraceItems()) {
         return;
       }
-      if (itemProcessor == null) {
+
+      Context parentContext = currentContext();
```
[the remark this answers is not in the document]
jack-berg: Should have caught that. This advice applies in 5 other places as well.

mateuszrzeszutek (approved):
Thanks! 👍

PR #4032, opened 2021-08-31
Fix parsing of unclean map values in Config.

anuraaga:
We currently reject map values where elements within them are e.g., empty, but we should ignore those (it's easy to have empty values, trailing commas, etc when creating a variable in a shell script).

I just copied the implementation as is from SDK.

Also added the ConfigProperties "TCK" from the SDK :)

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/ConfigValueParsers.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/ConfigValueParsers.java
@@ -6,36 +6,45 @@
 package io.opentelemetry.instrumentation.api.config;
 
 import java.time.Duration;
+import java.util.AbstractMap;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 final class ConfigValueParsers {
 
   static List<String> parseList(String value) {
-    String[] tokens = value.split(",", -1);
-    // Remove whitespace from each item.
-    for (int i = 0; i < tokens.length; i++) {
-      tokens[i] = tokens[i].trim();
-    }
-    return Collections.unmodifiableList(Arrays.asList(tokens));
+    return Collections.unmodifiableList(filterBlanksAndNulls(value.split(",")));
   }
 
   static Map<String, String> parseMap(String value) {
-    Map<String, String> result = new LinkedHashMap<>();
-    for (String token : value.split(",", -1)) {
-      token = token.trim();
-      String[] parts = token.split("=", -1);
-      if (parts.length != 2) {
-        throw new IllegalArgumentException(
-            "Invalid map config part, should be formatted key1=value1,key2=value2: " + value);
-      }
-      result.put(parts[0], parts[1]);
-    }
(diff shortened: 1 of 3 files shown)


javaagent-tooling/src/test/java/io/opentelemetry/javaagent/tooling/config/ConfigPropertiesAdapterTest.java
```diff
@@ -0,0 +1,193 @@
...
+    assertThat(config.getInt("int")).isNull();
+    assertThat(config.getLong("long")).isNull();
+    assertThat(config.getDouble("double")).isNull();
+    assertThat(config.getCommaSeparatedValues("list")).isEmpty();
+    assertThat(config.getCommaSeparatedMap("map")).isEmpty();
+    assertThat(config.getDuration("duration")).isNull();
+  }
+
+  @Test
+  @Disabled("Currently invalid values are not handled the same as the SDK")
```
anuraaga, line 81: @mateuszrzeszutek I think we should make sure the behavior is the same as the SDK for these too

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/ConfigValueParsers.java
```diff
@@ -6,36 +6,45 @@
...
-    }
-    return Collections.unmodifiableMap(result);
+    return parseList(value).stream()
+        .map(keyValuePair -> filterBlanksAndNulls(keyValuePair.split("=", 2)))
+        .map(
+            splitKeyValuePairs -> {
+              if (splitKeyValuePairs.size() != 2) {
+                throw new IllegalArgumentException(
+                    "Invalid map property, should be formatted key1=value1,key2=value2: " + value);
+              }
```
trask, line 32: should we allow empty values? e.g. `key1=val1,key2=,key3=val3`

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/ConfigValueParsers.java
anuraaga: We do for `getString` (at least in SDK) though seem to be missing tests for it, and the map with empty values. But I think returning empty when it's empty is reasonable behavior, it's similar to how query params work

mateuszrzeszutek (approved):

javaagent-tooling/src/test/java/io/opentelemetry/javaagent/tooling/config/ConfigPropertiesAdapterTest.java
mateuszrzeszutek: Yeah, that'd certainly make sense to do so. On the other hand, in the agent code I think we'd probably prefer to fall back on the default value instead of throwing an exception in the middle of an advice class. So, how about the following:
* in the `@Nullable X getX(String name)` methods (these are used only in the SDK bridge, I think) we'll throw an exception if the value is invalid;
* but in the `X getX(String name, X default)` (which are used pretty much everywhere in the agent) we'll return the default value in case of any errors

anuraaga: Yeah I think that's fine

PR #4035, opened 2021-08-31
Make Config behave exactly as SDK DefaultConfigProperties

mateuszrzeszutek:
Implements https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4032#discussion_r699006611


javaagent-tooling/src/test/java/io/opentelemetry/javaagent/tooling/config/ConfigPropertiesAdapterTest.java
```diff
@@ -79,7 +78,6 @@ void allEmpty() {
   }
 
   @Test
-  @Disabled("Currently invalid values are not handled the same as the SDK")
```
anuraaga, line 82: :tada:

anuraaga (approved):

trask (approved):

PR #4038, opened 2021-08-31
[overhead] Append and update results (redux)

breedx-splk:
Switch from `stefanzweifel/git-auto-commit-action` to `EndBug/add-and-commit` (it's more of what we need and is less targeted at static site generation).  We leverage the gh-branch as a cloned subdir, copy the results into a local dir that is appended to by the test run. After the tests finish, we rsync the dir back into the subdir and remove the local copy. 

Tested multiple times in another repo and seems to work fine -- we can append the gh-pages this way without needing to commit back to main.  🤞🏻 🍀 🐰 🦶🏻

### DIFF
--- a/.github/workflows/nightly-benchmark-overhead.yml
+++ b/.github/workflows/nightly-benchmark-overhead.yml
@@ -18,13 +18,19 @@ jobs:
           path: gh-pages
       - name: copy results from gh-pages branch
         run: |
-          rsync -avv gh-pages/benchmark-overhead/results/ benchmark-overhead/results/ && rm -rf gh-pages
+          rsync -avv gh-pages/benchmark-overhead/results/ benchmark-overhead/results/
       - name: run tests
-        run: ./gradlew test
         working-directory: benchmark-overhead
+        run: ./gradlew test
+      - name: inspect the results dir
+        working-directory: benchmark-overhead
+        run: ls -lR results
+      - name: copy results back to gh-pages branch
+        run: rsync -avv benchmark-overhead/results/ gh-pages/benchmark-overhead/results/ && rm -rf benchmark-overhead/results
       - name: commit updated results
-        uses: stefanzweifel/git-auto-commit-action@v4
+        uses: EndBug/add-and-commit@v7
         with:
-          branch: gh-pages
-          commit_message: update test result data
-          file_pattern: benchmark-overhead/results
+          add: 'benchmark-overhead/results'
+          cwd: './gh-pages'
+          branch: 'gh-pages'
+          message: 'update test result data'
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly-benchmark-overhead.yml
```diff
@@ -18,13 +18,19 @@ jobs:
...
-          rsync -avv gh-pages/benchmark-overhead/results/ benchmark-overhead/results/ && rm -rf gh-pages
+          rsync -avv gh-pages/benchmark-overhead/results/ benchmark-overhead/results/
       - name: run tests
+        working-directory: benchmark-overhead
         run: ./gradlew test
+      - name: inspect the results dir
         working-directory: benchmark-overhead
+        run: ls -lR results
+      - name: copy results back to gh-pages branch
+        run: rsync -avv benchmark-overhead/results/ gh-pages/benchmark-overhead/results/ && rm -rf benchmark-overhead/results
```
trask, line 29: is the `rm -rf` part needed? (not that it's hurting..)
```suggestion
        run: rsync -avv benchmark-overhead/results/ gh-pages/benchmark-overhead/results/
```

trask (approved):

.github/workflows/nightly-benchmark-overhead.yml
mateuszrzeszutek: Wasn't that because of this issue: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3973 ?

mateuszrzeszutek (approved):

.github/workflows/nightly-benchmark-overhead.yml
```diff
@@ -18,13 +18,19 @@ jobs:
...
       - name: copy results from gh-pages branch
         run: |
-          rsync -avv gh-pages/benchmark-overhead/results/ benchmark-overhead/results/ && rm -rf gh-pages
+          rsync -avv gh-pages/benchmark-overhead/results/ benchmark-overhead/results/
       - name: run tests
+        working-directory: benchmark-overhead
         run: ./gradlew test
+      - name: inspect the results dir
         working-directory: benchmark-overhead
+        run: ls -lR results
```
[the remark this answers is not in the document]
breedx-splk: Why not both?  :). I can remove it if you think it's too much clutter.

.github/workflows/nightly-benchmark-overhead.yml
breedx-splk: I have absolutely encountered problems with some of the GH actions when you have local changes. Some of them try and change branches and can't because of untracked changes. Removing the results like this helps to avoid those conditions. I'm hesitant to remove it...all the stars have to be aligned for it to work. ⭐ 😁

trask:
👍

PR #4041, opened 2021-09-01
Start a CONSUMER span for Kafka poll(); and refactor spring-kafka...

mateuszrzeszutek:
... so that it has 1 batch receive span + 1 batch process span. This basically implements this idea: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3922#discussion_r696407770 (well, a part of it).

The process CONSUMER span is not a child of the receive CONSUMER span yet (spec [says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#batch-receiving) it should be) because successive CONSUMERs are suppressed; but I'm planning to introduce a new suppression strategy for consumer spans in the next PR (separate span keys for receive an process).

This PR is split into 2 commits: the first contains just `kafka-clients` changes, the second just `spring-kafka` changes. Maybe that'll help reviewing it.


instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/ReceivedRecords.java
```diff
@@ -0,0 +1,30 @@
...
+  public abstract ConsumerRecords<?, ?> records();
+
+  abstract Timer timer();
+
+  public Instant startTime() {
+    return timer().startTime();
+  }
+
+  public Instant endTime() {
+    return timer().endTime();
```
trask, line 28: maybe
```suggestion
  public Instant now() {
    return timer().now();
```

trask (approved):
I think it would be nice if we can avoid releasing in the intermediate state where we are creating disjointed consumer spans

mateuszrzeszutek:
> I think it would be nice if we can avoid releasing in the intermediate state where we are creating disjointed consumer spans

Yeah, definitely - I'll start working on the next part soon, we should definitely get it merged before the next release.

arm-chair:
Thanks @mateuszrzeszutek, were you able to make any progress on the next part? I think I am maybe seeing the disjointed consumer in an downstream Lightstep alpha release.

trask:
@arm-chair does lightstep backend support links? if so you should be seeing the consumer linked to the producer.
if not, check out #4187

mateuszrzeszutek:
Span links were added here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4122 and https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4151

PR #4044, opened 2021-09-01
Transition http-url-connection module to instrumenter API

jack-berg:
Contributing to Issue #2713.


instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlHttpAttributesExtractor.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Override
+  protected @Nullable String target(HttpURLConnection connection) {
+    return null;
+  }
+
+  @Override
+  protected @Nullable String host(HttpURLConnection connection) {
+    return null;
```
mateuszrzeszutek, line 32: Nit: could be
```suggestion
    return connection.getRequestProperty("Host");
```

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -122,16 +122,26 @@ public static void methodExit(
...
       } else if (methodName.equals("getInputStream") && responseCode > 0) {
         // responseCode field is sometimes not populated.
         // We can't call getResponseCode() due to some unwanted side-effects
         // (e.g. breaks getOutputStream).
-        tracer().end(httpUrlState.context, new HttpUrlResponse(connection, responseCode));
+        instrumenter()
+            .end(
+                httpUrlState.context,
+                connection,
+                new HttpUrlResponse(connection, responseCode),
```
mateuszrzeszutek, line 143: Is the `connection` field of `HttpUrlResponse` used anymore? Could we just pass `responseCode` as the response instead?

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlHttpAttributesExtractor.java
jack-berg: This relates to a general question I've had about instrumeter refactoring: often times the `Instrumenter` is replacing `HttpClientTracer`. `HttpClientTracer` collects a small subset of the attributes available in `HttpAttributesExtractor`. So far, I've been trying to replace `HttpClientTracer` with an implementation of `HttpAttributesExtractor` that collects the same set of attributes, even when it would be straight forward to capture additional ones. I figure better to keep this a strict refactoring, and enhance the behavior by collecting additional attributes in subsequent commits. 

If you agree with keeping this a strict refactoring, then we should collect the host attribute in a followup PR, since its not currently set by `HttpClientTracer`.

mateuszrzeszutek: Sure, that's fine by me 👍

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentation.java
jack-berg: Nice catch!

mateuszrzeszutek (approved):

trask (approved):
👍

PR #4045, opened 2021-09-01
Transition twilio-6.6 module to instrumenter API

jack-berg:
Contributing to Issue #2713.


instrumentation/twilio-6.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/twilio/TwilioTracer.java
```diff
@@ -5,128 +5,39 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder;
 import io.opentelemetry.instrumentation.api.tracer.SpanNames;
-import java.lang.reflect.Method;
-import java.util.concurrent.TimeUnit;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-public class TwilioTracer extends BaseTracer {
+public class TwilioTracer {
```
mateuszrzeszutek, line 16: How about 
```suggestion
public class TwilioSingletons {
```
?

mateuszrzeszutek (approved):

trask (commented):
it feels a little odd using "span name" as the request, and passing span name around, would it work to pass around the service executor `Object` as the request?

jack-berg:
@trask you'd need both the `Object serviceExecutor` and the `String methodName`, since both are involved in generating the span name. It's easy enough to create an object composed of these fields and use that instead. However, there is precedent for using a span name as the Instrumenter `REQUEST` type in scenarios where there's no other attribute extraction. 

I don't have a strong preference.

trask:
> @trask you'd need both the `Object serviceExecutor` and the `String methodName`, since both are involved in generating the span name.

makes sense, ya, not worth having a wrapper object if we don't need to 👍

Issue #1708, opened 2020-11-20
Update groovy import order to match default Intellij


trask:
Since we can't enforce the import order for groovy files, it makes sense to use the default Intellij ordering. Hopefully this should reduce import order drift.

Idea from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1690#issuecomment-730792005

mcmho:
@trask  can I work on this one as a starter?  :-)

mcmho:
On https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/intellij-setup.md, 
do we recommend some particular versions of IntelliJ (2020.3,  2021.2.1)?

trask:
> @trask can I work on this one as a starter? :-)

👍 assigned to you

> @trask, just to make sure, this import update is for all .groovy files, correct?

👍

mcmho:
@trask ,  also what I found is the following is the default IntelliJ 2019.3 and 2021.2.1 setting for .groovy files.   
I just want to make sure this is what we are following correct?   
![image](https://user-images.githubusercontent.com/5099946/131568336-e58b8174-5a80-494d-8e33-e7558853edc5.png)

if so, then this needs to be updated as well - https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/intellij-setup.md

trask:
> I just want to make sure this is what we are following correct?

👍

> if so, then this needs to be updated as well - https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/intellij-setup.md

👍

iNikem:
Closed by #4047

PR #4047, opened 2021-09-02
Update groovy import order to match default IntelliJ (#1708)

mcmho:
* organized imports for .groovy files to follow default IntelliJ import settings

* updated intellij-setup.md

* reformat lines with missing empty space


trask:
hey @mcmho! it looks like there may be a couple of missing imports which is causing the CI failure

mcmho:
> hey @mcmho! it looks like there may be a couple of missing imports which is causing the CI failure

@trask ,  got it!   I'm on it!

trask:
@mcmho it looks like the screenshot needs to be updated, and then i think we'll be good to merge

trask:
> Thanks @trask! sorry had issue with my dev env. but will get better.

no worries, thanks for diving in!

trask (approved):
thx!

PR #4051, opened 2021-09-06
Fix spring context reload issue (open-telemetry#3841)

frommeyerc:
I've created a small PR which will cause spring boot autoconfiguration not not register the OpenTelemetry Bean any more.  If for a project some tracing would rely on the global registration of the bean this could be done in a second step manually, so this would still be supported.


mateuszrzeszutek:
Hey @frommeyerc ,
Thanks for your contribution! Can you sign the CLA?

trask (approved):
makes sense to me, not registering global seems like a good default for the spring starter

iNikem (approved):

Issue #4040, opened 2021-09-01
It seemed that there is a memory leak in spring cloud gateway.(instrumentation version release.1.4.0)


tydhot:
I run spring cloud gateway with java instrumentation and the heap size increases overtime and  oom finally.
<img width="508" alt="image" src="https://user-images.githubusercontent.com/27889201/131657957-706014b9-5f43-453a-9b16-0d3394593f2c.png">
I dump it and found there is 402k TracingSubscriber and there is 30k RecordEventsReadableSpan. Maybe the info can help?
<img width="1163" alt="image" src="https://user-images.githubusercontent.com/27889201/131658461-413f3bc7-3cba-4d84-9158-c12bb535ebfb.png">

tydhot:
<img width="901" alt="image" src="https://user-images.githubusercontent.com/27889201/131703229-97baa3d2-077c-43c4-8899-70d53c0e0f6d.png">
<img width="1013" alt="image" src="https://user-images.githubusercontent.com/27889201/131703280-e66cda15-6d60-4213-a0b6-365482a017b3.png">
According to gc root, I think the num of the channelHandler from FieldBackedProvider seemed a little big.

mateuszrzeszutek:
It's just my suspicion, but in `AbstractNettyChannelPipelineInstrumentation` we're not instrumenting `ChannelPipeline`'s `removeFirst()`, `removeLast()` and `replace()` methods. If somebody used those methods for removing/replacing handlers that were added with instrumented `add...()` our own handlers would probably not get removed.

mateuszrzeszutek:
@tydhot If you're able to reproduce that issue, please try turning off the `netty` instrumentation and check if it helps.

tydhot:
@mateuszrzeszutek hi, I have turned off the Netty instrumentation like below:
<img width="602" alt="image" src="https://user-images.githubusercontent.com/27889201/131794139-c0031362-f62f-4dae-b4e6-18b762a5039d.png">
This is the heap size turn off nett:
<img width="788" alt="image" src="https://user-images.githubusercontent.com/27889201/131794265-1b1f6e01-b279-4986-a62f-ca9e304b3a8d.png">
This is the original control group corresponding to turning off netty：
<img width="770" alt="image" src="https://user-images.githubusercontent.com/27889201/131794379-385347d2-6d35-4491-8034-5f1e6ff5efd0.png">
I think it is worked.

[...]

PR #4053, opened 2021-09-06
Fix memleak in the Netty ChannelPipeline instrumentation

mateuszrzeszutek:
Hopefully that should resolve https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4040


anuraaga (approved):

trask (approved):

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/AbstractNettyChannelPipelineInstrumentation.java
```diff
@@ -109,4 +112,20 @@ public static void removeHandler(
       }
     }
   }
+
+  @SuppressWarnings("unused")
+  public static class RemoveFirstLastAdvice {
```
tydhot, line 117: I have some questions about the handling of `removelast`, if user call  `removelast` here, maybe the handler be removed will be ourHander? And the origin handler will be remained if user only call `removelast` once.

mateuszrzeszutek: Yeah, you're exactly right - I'll create a separate issue for this, since it has never worked correctly.
Thanks!

Issue #4057, opened 2021-09-07
DubboTracer startServerSpan transform baggage info fail


zhq734:
**Describe the bug**
In DubboTracer startServerSpan transform baggage info fail.

The current official implementation:
`return withServerSpan(Context.current(), spanBuilder.startSpan());`
But in this case, the data passed by Propogator will be lost

**My advice**
I suggest that parentContext should be used here instead of Context.current();

The final change is like this:
`return withServerSpan(parentContext, spanBuilder.startSpan());`

**What version are you using?**
(e.g., `v1.4.1`, `393e4a2`, etc)

**Environment**
Compiler: (e.g., "AdoptOpenJDK 11.0.6")
OS: (e.g., "Ubuntu 20.04")
Runtime (if different from JDK above): (e.g., "Oracle JRE 8u251")
OS (if different from OS compiled on): (e.g., "Windows Server 2019")

**Additional context**
`public Context startServerSpan(
      String interfaceName, String methodName, RpcInvocation rpcInvocation) {
    Context parentContext = extract(rpcInvocation, getGetter());
    SpanBuilder spanBuilder =
        spanBuilder(parentContext, DubboHelper.getSpanName(interfaceName, methodName), SERVER)
            .setAttribute(SemanticAttributes.RPC_SYSTEM, "dubbo");
    DubboHelper.prepareSpan(spanBuilder, interfaceName, methodName);
    NetPeerAttributes.INSTANCE.setNetPeer(spanBuilder, RpcContext.getContext().getRemoteAddress());
   return withServerSpan(Context.current(), spanBuilder.startSpan());
  }`

PR #4059, opened 2021-09-07
Migrate Dubbo to Instrumenter API

mateuszrzeszutek:
Part of #2713, also resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4057.


instrumentation/apache-dubbo-2.7/library/build.gradle.kts
```diff
@@ -3,6 +3,12 @@ plugins {
 }
 
 dependencies {
+  compileOnly("com.google.auto.value:auto-value-annotations")
+  annotationProcessor("com.google.auto.value:auto-value")
+
+  // required for PeerServiceAttributesExtractor
+  implementation(project(":javaagent-instrumentation-api"))
```
mateuszrzeszutek, line 10: The library instrumentation here is a bit unusual, because the only way to inject a `Filter` is through SPI - so we have to use `GlobalOpenTelemetry`, `PeerServiceExtractor` et cetera in the library instrumentation; customization is just not possible.

trask: could we have a separate SPI class that we inject via the javaagent (which does some init and delegates to the library SPI class)?

trask (approved):

instrumentation/apache-dubbo-2.7/library/build.gradle.kts
mateuszrzeszutek: That was my first idea (since AWS SDK does this) - but unfortunately it's not possible to inject a `Filter` manually, there is no way to add it to Dubbo except for the SPI. So the users of the library instrumentation would have to create a delegate `Filter` implementation with correct `@Activate` annotation and add SPI file to make it work - that's a lot of work required just to make it work.

trask: sent a PR to your branch with a possible idea https://github.com/mateuszrzeszutek/opentelemetry-java-instrumentation/pull/29

anuraaga (approved):

instrumentation/apache-dubbo-2.7/library/build.gradle.kts
mateuszrzeszutek: Thanks! I'll try that out

mateuszrzeszutek: Okay, done - I changed your version a bit and added those "standard" library instrumentation `DubboTracing`/`DubboTracingBuilder` classes.

trask (approved):

PR #4068, opened 2021-09-08
Add otlp http exporters

trask:
So that `OTEL_EXPERIMENTAL_EXPORTER_OTLP_PROTOCOL=http/protobuf` will work once we pull in Java SDK 1.6.0 which introduces autoconfigure support for this https://github.com/open-telemetry/opentelemetry-java/pull/3522

### DIFF
--- a/javaagent-exporters/build.gradle.kts
+++ b/javaagent-exporters/build.gradle.kts
@@ -8,6 +8,8 @@ dependencies {
   implementation("io.opentelemetry:opentelemetry-exporter-jaeger")
   implementation("io.opentelemetry:opentelemetry-exporter-otlp")
   implementation("io.opentelemetry:opentelemetry-exporter-otlp-metrics")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-http-trace")
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp-http-metrics")
   implementation("io.opentelemetry:opentelemetry-exporter-logging-otlp")
 
   implementation("io.opentelemetry:opentelemetry-exporter-prometheus")


anuraaga (approved):

anuraaga:
Oh btw I think we'll need to force the okhttp dependency to avoid pulling in the kotlin stdlib (assuming we didn't accidentally pull it in elsewhere again :P). Though with it EoL, it's too bad really, I think we may need to reconsider using okhttp @jkwatson. Kotlin stdlib is just too heavyweight for a telemetry exporter I think

mateuszrzeszutek (approved):

jkwatson:
> Oh btw I think we'll need to force the okhttp dependency to avoid pulling in the kotlin stdlib (assuming we didn't accidentally pull it in elsewhere again :P). Though with it EoL, it's too bad really, I think we may need to reconsider using okhttp @jkwatson @jack-berg . Kotlin stdlib is just too heavyweight for a telemetry exporter I think

Which EOL are you referring to, exactly here?

jkwatson:
> > Oh btw I think we'll need to force the okhttp dependency to avoid pulling in the kotlin stdlib (assuming we didn't accidentally pull it in elsewhere again :P). Though with it EoL, it's too bad really, I think we may need to reconsider using okhttp @jkwatson @jack-berg . Kotlin stdlib is just too heavyweight for a telemetry exporter I think
> 
> Which EOL are you referring to, exactly here?

oh, I get it. okhttp 3.x didn't depend on the kotlin stdlib, but has been eol'd. 4.x does depend on the kotlin stdlib, which is definitely some significant bloat.

jkwatson:
> Oh btw I think we'll need to force the okhttp dependency to avoid pulling in the kotlin stdlib (assuming we didn't accidentally pull it in elsewhere again :P). Though with it EoL, it's too bad really, I think we may need to reconsider using okhttp @jkwatson @jack-berg . Kotlin stdlib is just too heavyweight for a telemetry exporter I think

We should probably build an multi-version jar, and use the java11 httpclient if it's available. Maybe using HttpUrlConnection (ugh!) for java 8.

trask:
> I think we'll need to force the okhttp dependency to avoid pulling in the kotlin stdlib

thx for this 👍

going to merge this now that https://github.com/open-telemetry/opentelemetry-java/pull/3590 is merged

PR #4075, opened 2021-09-09
JaxRS-client: remove our context when request is done

laurit:
Remove our context from request properties when request is done.
I saw a heap dump where `ClientRequestContext` implementation is finalizable. There was a bunch of these queued for finalization and each had reference to our context due to which one could see `ArrayBasedContext` and `RecordEventsReadableSpan` instances consuming a bit of memory. This change should ensure that our context instances don't show up in such heap dumps.

### DIFF
--- a/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ClientTracingFilter.java
+++ b/instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ClientTracingFilter.java
@@ -31,10 +31,12 @@ public class ClientTracingFilter implements ClientRequestFilter, ClientResponseF
   @Override
   public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) {
     Object contextObj = requestContext.getProperty(CONTEXT_PROPERTY_NAME);
-    requestContext.removeProperty(CONTEXT_PROPERTY_NAME);
     if (contextObj instanceof Context) {
       Context context = (Context) contextObj;
       instrumenter().end(context, requestContext, responseContext, null);
     }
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ClientTracingFilter.java
```diff
@@ -31,6 +31,7 @@ public void filter(ClientRequestContext requestContext) {
   @Override
   public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) {
     Object contextObj = requestContext.getProperty(CONTEXT_PROPERTY_NAME);
+    requestContext.removeProperty(CONTEXT_PROPERTY_NAME);
```
trask, line 34: can you add a comment why we are removing the property? (since it's not standard practice across instrumentation to remove properties/attributes that we add - or should it be?)

trask (approved):

instrumentation/jaxrs-client/jaxrs-client-2.0/jaxrs-client-2.0-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jaxrsclient/v2_0/ClientTracingFilter.java
laurit: Comment added. If we can easily remove stuff added by us then it would be nice, for example in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jaxws/jaxws-2.0-cxf-3.0/library/src/main/java/io/opentelemetry/javaagent/instrumentation/cxf/CxfHelper.java there is a `remove` method that returns the current value so using it instead of `get` doesn't complicate things.

PR #4076, opened 2021-09-09
Separate suppression strategies for CONSUMER receive/process spans

mateuszrzeszutek:
I didn't apply this yet to Kafka instrumentations (because of the `ContextStore` issue, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4067), that'll be done in one of the next PRs. This one contains only pretty much only suppression strategy changes.


lmolkova (approved):
SpanKey changes look fine to me.  There are some controversial things baked into the spec though: e.g. CONSUMER receive span can have nested HTTP/other transport CLIENT span - we won't suppress them.
Also, the messaging spec is under discussion now and is likely to change.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKeyExtractor.java
```diff
@@ -0,0 +1,54 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.rpc.RpcAttributesExtractor;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+final class SpanKeyExtractor {
+
+  /**
+   * Automatically determines {@link SpanKey}s that should be ashould be should be applied to the
```
trask, line 19:
```suggestion
   * Automatically determines {@link SpanKey}s that should be applied to the
```

instrumentation/jms-1.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jms/MessageWithDestination.java
```diff
@@ -24,8 +23,6 @@
 
   public abstract Message message();
 
-  public abstract MessageOperation messageOperation();
```
trask, line 27: 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/messaging/MessagingAttributesExtractor.java
```diff
@@ -61,6 +61,8 @@ protected final void onEnd(
     set(attributes, SemanticAttributes.MESSAGING_MESSAGE_ID, messageId(request, response));
   }
 
+  public abstract MessageOperation operation();
```
trask, line 64: 👍 removing the arg

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/SpringMessagingAttributesExtractor.java
```diff
@@ -0,0 +1,76 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.spring.integration;
+
+import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessageOperation;
+import io.opentelemetry.instrumentation.api.instrumenter.messaging.MessagingAttributesExtractor;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+// this class is needed mostly for correct CONSUMER span suppression
```
trask, line 12: 👍

trask (approved):

mateuszrzeszutek:
> There are some controversial things baked into the spec though: e.g. CONSUMER receive span can have nested HTTP/other transport CLIENT span - we won't suppress them.

I'd say that it's probably how it should work - the actual http/socket call made when polling/receiving has completely different semantics than a messaging consumer span, it sort of maps to the "DB/RPC over HTTP" idea that motivated the client span suppression changes.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -146,25 +141,26 @@
...
    * </ul>
    *
    * <p><strong>In both cases:</strong>
    *
    * <ul>
-   *   <li>SERVER and CONSUMER nested spans are always suppressed
+   *   <li>SERVER and PRODUCER nested spans are always suppressed
+   *   <li>CONSUMER spans are suppressed depending on their operation: receive or process
    *   <li>INTERNAL spans are never suppressed
    * </ul>
```
iNikem, line 165: I haven't read the full PR, but I sense some inconsistencies in this documentation.

First, `PRODUCER` is equivalent to the `CLIENT`, why are they handled differently? E.g. if there is a messaging client that uses http protocol underneath, shouldn't we suppress http client by default and not suppress when using suppression by type?

Second, this javadoc talks about `SpanKey` in case of `CLIENT` spans but uses different language for `CONSUMER` spans, although they are both implemented using `SpanKey`.

Third, do we have to reflect this somehow in some user-facing documentation?

mateuszrzeszutek: > First, `PRODUCER` is equivalent to the `CLIENT`, why are they handled differently?

Is it really? A `PRODUCER` puts a message with id=123 into queue with name=abc, while a `CLIENT` sends a HTTP message to URL http://... - they have different semantics, so should `PRODUCER` spans really suppress the `CLIENT` s underneath?
(Also, we've never suppressed anything related to `PRODUCER` spans in the agent)

> Second, this javadoc talks about `SpanKey` in case of `CLIENT` spans but uses different language for `CONSUMER` spans, although they are both implemented using `SpanKey`.

Good point, will fix that 👍 

> Third, do we have to reflect this somehow in some user-facing documentation?

We probably should - the various suppression behaviors exhibited by the agent definitely deserve their own documentation page.

iNikem: `PRODUCER` is equivalent to `CLIENT` in the same sense as `SERVER` is equivalent to `CONSUMER`, the only difference between them in async vs sync.

mateuszrzeszutek: TIL https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#span-kind 😅 

Ideally we should avoid suppressing based on the span kind and use the instrumentation/semantic type instead - it'd probably be a lot easier if we changed the default behavior from "suppress all nested clients" to the "suppress per type" one.

PR #4077, opened 2021-09-09
Update supported libs list

mateuszrzeszutek:
I've noticed that this hasn't been updated for a while


laurit (approved):

trask (approved):
👍

iNikem:
Just an idea: can we generate this list from out muzzle directives?

PR #4080, opened 2021-09-10
Update to InstrumentationContext javadoc

trask:
Context store class pairs appear to be collected when traversing helper classes (otherwise #3987 wouldn't work), so no need to call `registerMuzzleContextStoreClasses` explicitly when using `InstrumentationContext.get` outside of Advice methods.

### DIFF
--- a/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
+++ b/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/InstrumentationContext.java
@@ -18,12 +18,8 @@ public class InstrumentationContext {
    * <p>In reality, the <em>calls</em> to this method are re-written to something more performant
    * while injecting advice into a method.
    *
-   * <p>When this method is called from outside of an Advice class it can only access {@link
-   * ContextStore} when it is already created. For this {@link ContextStore} either needs to be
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

anuraaga (approved):

laurit:
#3987 only works because of https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/AgentCacheInstrumentation.java Helper class that uses `InstrumentationContext.get` is used from an advice class that doesn't match any methods.

trask:
@laurit [AgentCacheInstrumentation.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jdbc/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jdbc/AgentCacheInstrumentation.java) seems more about registering the helper classes (with the side-effect of registering the InstrumentationContext).

tho re-reading the existing javadoc, I missed "or Helper" in

> used in an Advice or Helper class which automatically adds it

so the existing javadoc makes sense to me now.

tho isn't all instrumentation code either an advice or helper class?

oh, we do support instrumentation "bootstrap" classes now..

laurit (approved):

Issue #4015, opened 2021-08-30
opentelemetry-api bridge doesn't work with GlobalOpenTelemetry.get() pattern


anuraaga:
`GlobalOpenTelemetry.getTracer("foo")` returns a real tracer. `GlobalOpenTelemetry.get().tracerProvider("foo").build()` returns a no-op tracer. Presumably are not instrumenting a new method that was added and it's calling the default implementation.

PR #4085, opened 2021-09-10
Bridge tracer builder

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4015


mateuszrzeszutek (approved):

laurit:
Turns out it is not that easy, as `TraceBuilder` was added in 1.4 this will fail when application includes 1.0 api.

PR #4087, opened 2021-09-10
Remove generated InstrumentationModule#getMuzzleReferences() method from the source code

iNikem:
It bothered me for a long time that we have a bunch of public classes in `io.opentelemetry.javaagent.extension.muzzle` package that in fact should never ever be used directly by the users of `javaagent-extension-api` module. They are there and public only because `InstrumentationModule#getMuzzleReferences()` exposes them. But that method is also should never be used by end-users. It is auto-generated and is used by `muzzle` and `tooling` modules.

This PR does the following:

- removes `InstrumentationModule#getMuzzleReferences()` method from the source code
- introduces `MuzzleReferencesAccessor` class into `muzzle` module to access the generated version of `getMuzzleReferences` method via `MethodHandle`
- Moves all usages of `getMuzzleReferences` method inside `ReferenceMatcher` class (and `ReferencesPrinter`)

All of the above would allow us to move `io.opentelemetry.javaagent.extension.muzzle.ClassRef` class (and all related classes from that package) from `javaagent-extension-api` module into `muzzle`, but unfortunately they are used by `gradle-plugins`. So as preparation for the future, I created a verbatim copies of all classes from `io.opentelemetry.javaagent.extension.muzzle` package in `io.opentelemetry.javaagent.tooling.muzzle.references`. Next PR can change gradle plugins to use these new classes. Old copies of them can be removed then from `javaagent-extension-api`.

### DIFF
--- a/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
+++ b/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
@@ -43,12 +43,12 @@ class MuzzleReferencesAccessor {
    * Returns references to helper and library classes used in the given module's type
    * instrumentation advices, grouped by {@link ClassRef#getClassName()}.
    */
+  @SuppressWarnings("unchecked")
   static Map<String, ClassRef> getFor(InstrumentationModule instrumentationModule) {
     Map<String, ClassRef> muzzleReferences = emptyMap();
     MethodHandle methodHandle = getMuzzleReferences.get(instrumentationModule.getClass());
     if (methodHandle != null) {
       try {
-        //noinspection unchecked
         muzzleReferences = (Map<String, ClassRef>) methodHandle.invoke(instrumentationModule);
       } catch (Throwable ignored) {
         // silence error prone
--- a/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
+++ b/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
@@ -51,6 +51,7 @@ class MuzzleReferencesAccessor {
         muzzleReferences =
             (Map<String, ClassRef>) getMuzzleReferences.invoke(instrumentationModule);
       } catch (Throwable ignored) {
+        //silence error prone
       }
     }
     return muzzleReferences;
--- a/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
+++ b/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
@@ -21,23 +21,23 @@ import java.util.Map;
  * io.opentelemetry.instrumentation.javaagent-codegen} Gradle plugin.
  */
 class MuzzleReferencesAccessor {
-  private static final MethodHandle getMuzzleReferences;
-
-  static {
-    MethodHandles.Lookup lookup = MethodHandles.publicLookup();
-    MethodHandle handle;
-    try {
-      // This method is generated automatically during compilation by
-      // the io.opentelemetry.instrumentation.javaagent-codegen Gradle plugin.
-      //noinspection JavaLangInvokeHandleSignature
-      handle =
-          lookup.findVirtual(
-              InstrumentationModule.class, "getMuzzleReferences", MethodType.methodType(Map.class));
-    } catch (NoSuchMethodException | IllegalAccessException e) {
-      handle = null;
-    }
-    getMuzzleReferences = handle;
-  }
+  private static final ClassValue<MethodHandle> getMuzzleReferences =
+      new ClassValue<MethodHandle>() {
+        @Override
+        protected MethodHandle computeValue(Class<?> type) {
+          MethodHandles.Lookup lookup = MethodHandles.publicLookup();
+          MethodHandle handle;
+          try {
+            // This method is generated automatically during compilation by
+            // the io.opentelemetry.instrumentation.javaagent-codegen Gradle plugin.
+            handle =
+                lookup.findVirtual(type, "getMuzzleReferences", MethodType.methodType(Map.class));
+          } catch (NoSuchMethodException | IllegalAccessException e) {
+            handle = null;
+          }
+          return handle;
+        }
+      };
 
   /**
    * Returns references to helper and library classes used in the given module's type
@@ -45,13 +45,13 @@ class MuzzleReferencesAccessor {
    */
   static Map<String, ClassRef> getFor(InstrumentationModule instrumentationModule) {
     Map<String, ClassRef> muzzleReferences = emptyMap();
-    if (getMuzzleReferences != null) {
+    MethodHandle methodHandle = getMuzzleReferences.get(instrumentationModule.getClass());
+    if (methodHandle != null) {
       try {
         //noinspection unchecked
-        muzzleReferences =
-            (Map<String, ClassRef>) getMuzzleReferences.invoke(instrumentationModule);
+        muzzleReferences = (Map<String, ClassRef>) methodHandle.invoke(instrumentationModule);
       } catch (Throwable ignored) {
-        //silence error prone
+        // silence error prone
       }
     }
     return muzzleReferences;
--- a/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
+++ b/muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.tooling.muzzle;
+
+import static java.util.Collections.emptyMap;
+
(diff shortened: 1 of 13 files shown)


muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
```diff
@@ -0,0 +1,58 @@
...
+  static {
+    MethodHandles.Lookup lookup = MethodHandles.publicLookup();
+    MethodHandle handle;
+    try {
+      // This method is generated automatically during compilation by
+      // the io.opentelemetry.instrumentation.javaagent-codegen Gradle plugin.
+      //noinspection JavaLangInvokeHandleSignature
+      handle =
+          lookup.findVirtual(
+              InstrumentationModule.class, "getMuzzleReferences", MethodType.methodType(Map.class));
```
mateuszrzeszutek, line 35: The `InstrumentationModule` base class does not contain this method though, each of its implementations does - and they're all different methods (with same descriptor though) now that they don't override a common base method. Will this work properly?

iNikem: I will let our tests answer that...

iNikem: Seems like you were right :) Checking the new version...

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcher.java
```diff
@@ -36,7 +37,14 @@
   private final Set<String> helperClassNames;
   private final InstrumentationClassPredicate instrumentationClassPredicate;
 
-  public ReferenceMatcher(
+  public static ReferenceMatcher of(InstrumentationModule instrumentationModule) {
```
trask, line 40: 👍

trask (approved):
nice

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Returns references to helper and library classes used in the given module's type
+   * instrumentation advices, grouped by {@link ClassRef#getClassName()}.
+   */
+  static Map<String, ClassRef> getFor(InstrumentationModule instrumentationModule) {
+    Map<String, ClassRef> muzzleReferences = emptyMap();
+    MethodHandle methodHandle = getMuzzleReferences.get(instrumentationModule.getClass());
+    if (methodHandle != null) {
+      try {
+        //noinspection unchecked
```
anuraaga, line 51: We'd usually use `@SuppressWarnings("unchecked")` I think

anuraaga (approved):

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * This class is used to access {@code getMuzzleReferences} method from the {@link
+ * InstrumentationModule} class. That method is not visible in the source code and instead is
+ * generated automatically during compilation by the {@code
+ * io.opentelemetry.instrumentation.javaagent-codegen} Gradle plugin.
+ */
+class MuzzleReferencesAccessor {
+  private static final ClassValue<MethodHandle> getMuzzleReferences =
+      new ClassValue<MethodHandle>() {
```
mateuszrzeszutek, line 25: Now that I think of it we don't really need to use `ClassValue`s and `MethodHandle`s - the `getMuzzleReferences()` will be called exactly once per `InstrumentationModule` implementation, so there's no need to cache the result here (it's already cached in the `MuzzleMatcher`).

mateuszrzeszutek (commented):
Nice 👍 
I think we can do the same thing for the new  `registerMuzzleContextStoreClasses()` method that I introduced recently.

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
```diff
@@ -0,0 +1,59 @@
...
+import java.lang.invoke.MethodType;
+import java.util.Map;
+
+/**
+ * This class is used to access {@code getMuzzleReferences} method from the {@link
+ * InstrumentationModule} class. That method is not visible in the source code and instead is
+ * generated automatically during compilation by the {@code
+ * io.opentelemetry.instrumentation.javaagent-codegen} Gradle plugin.
+ */
+class MuzzleReferencesAccessor {
```
laurit, line 23: Perhaps `javaagent-codegen` should add an interface that has `getMuzzleReferences` to instrumentation module so you wouldn't need to use reflection.

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
iNikem: I prefer statement-level suppression to the method-level one. But I will change this :)

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
iNikem: Are you sure? Will be happy to remove this cache!

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
iNikem: I assume that MethodHandle has no performance overhead on invocation. But I will think about this idea, it may be more elegant indeed.

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
mateuszrzeszutek: Yep -- the only place (`MuzzleMatcher`) that constructs `ReferenceMatcher` in the agent uses an `AtomicBoolean initialized` field to make sure it only does it once (building a giant map of references can get expensive).

iNikem: What about `io.opentelemetry.javaagent.tooling.muzzle.ClassLoaderMatcher#matchesAll`?

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
iNikem: I will try this idea as a follow-up optimisation. I like it, but prefer to finish current iteration of removing ClassRef and friends from public API.

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
iNikem: Also, I will try Lauri idea from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4087#discussion_r707110186 anyway later on. That will probably remove this whole class.

mateuszrzeszutek: That one is only used in the gradle muzzle-check plugin, it doesn't affect the agent at all.

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
mateuszrzeszutek: This will also be easier to extend with other muzzle-generated methods.

laurit: If it is possible to build extension jars that contain instrumentation modules then all of the muzzle and codegen stuff has become part of our api. Have you thought about how to version it and how to handle cases where agent is incompatible with codegen version used?

iNikem: Ohhh, good question... The only vague though I have atm is "javagent must support all previous stable versions of codegen within any given major version." Which assumes "in sync" versioning of the agent and gradle plugins.

trask: > If it is possible to build extension jars that contain instrumentation modules then all of the muzzle and codegen stuff has become part of our api

oh no, this is a great point, so maybe we shouldn't get rid of these classes from our public API? 😭

iNikem: These classes should be part of the contract between muzzle generation and muzzle verification. But no end-user, including extension authors, should see them.

PR #4090, opened 2021-09-10
Bridge trace builder

### LINKED ISSUE #4015: opentelemetry-api bridge doesn't work with GlobalOpenTelemetry.get() pattern

laurit:
Resolves #4015
Introduce a separate module for bridging 1.4 api because `TracerBuilder` was added in 1.4.


instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/build.gradle.kts
```diff
@@ -34,6 +34,6 @@ dependencies {
   testImplementation("io.opentelemetry:opentelemetry-extension-annotations")
   testInstrumentation(project(":instrumentation:opentelemetry-annotations-1.0:javaagent"))
 
-  testImplementation(project(":instrumentation:opentelemetry-api-1.0:testing"))
-  testInstrumentation(project(":instrumentation:opentelemetry-api-1.0:testing"))
+  testImplementation(project(":instrumentation:opentelemetry-api:opentelemetry-api-1.0:testing"))
+  testInstrumentation(project(":instrumentation:opentelemetry-api:opentelemetry-api-1.0:testing"))
```
trask, line 38: unrelated to this PR but
```suggestion
```

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/trace/ApplicationTracerProvider.java
```diff
@@ -0,0 +1,61 @@
...
+      Class clazz =
+          Class.forName(
+              "io.opentelemetry.javaagent.instrumentation.opentelemetryapi.v1_4.trace.ApplicationTracerProvider14");
+      return clazz.getDeclaredConstructor(io.opentelemetry.api.trace.TracerProvider.class);
+    } catch (ClassNotFoundException | NoSuchMethodException exception) {
+      return null;
+    }
+  }
+
+  public static TracerProvider make(io.opentelemetry.api.trace.TracerProvider agentTracerProvider) {
```
trask, line 37: nothing against `make` 😄, just to conform with existing namings
```suggestion
  public static TracerProvider create(io.opentelemetry.api.trace.TracerProvider agentTracerProvider) {
```

trask (approved):
👍

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/build.gradle.kts
trask: ignore me, meant to delete this comment (because it's wrong 😄)

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/trace/ApplicationTracerProvider.java
laurit: make did feel wrong when I wrote it :)

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/trace/ApplicationTracerProvider.java
```diff
@@ -0,0 +1,62 @@
...
+    } catch (ClassNotFoundException | NoSuchMethodException exception) {
+      return null;
+    }
+  }
+
+  public static TracerProvider create(
+      io.opentelemetry.api.trace.TracerProvider agentTracerProvider) {
+    if (TRACE_PROVIDER_14 != null) {
+      try {
+        return (TracerProvider) TRACE_PROVIDER_14.newInstance(agentTracerProvider);
```
mateuszrzeszutek, line 41: Can we use a `MethodHandle` instead of reflection?

mateuszrzeszutek (approved):

mateuszrzeszutek (approved):

Issue #4092, opened 2021-09-11
LocalCacheFactory fails if `jdk.unsupported` module is not available


brunojcm:
**Is your feature request related to a problem? Please describe.**
The current version of opentelemetry-java (1.5.3 as of this bug report) requires the `jdk.unsupported` JDK module to be included if one is using `jlink` to build a custom JDK. This is not a sustainable approach.

The following error occurs:
```
[otel.javaagent 2021-09-11 08:20:50:638 +0000] [main] INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent - version: 1.5.3
[otel.javaagent 2021-09-11 08:20:50:911 +0000] [main] INFO io.grpc.netty.shaded.io.netty.util.internal.PlatformDependent - Your platform does not provide complete low-level API for accessing direct buffers reliably. Unless explicitly requested, heap buffer will always be preferred to avoid potential system instability.
ERROR io.opentelemetry.javaagent.OpenTelemetryAgent
java.lang.reflect.InvocationTargetException
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
    at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
    at java.base/java.lang.reflect.Method.invoke(Unknown Source)
    at io.opentelemetry.javaagent.bootstrap.AgentInitializer.initialize(AgentInitializer.java:40)
    at io.opentelemetry.javaagent.OpenTelemetryAgent.agentmain(OpenTelemetryAgent.java:51)
    at io.opentelemetry.javaagent.OpenTelemetryAgent.premain(OpenTelemetryAgent.java:44)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
    at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
    at java.base/java.lang.reflect.Method.invoke(Unknown Source)
    at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(Unknown Source)
    at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(Unknown Source)
[...]

PR #4093, opened 2021-09-11
Upgrade caffeine to 3.0.2, no more sun.misc.Unsafe

brunojcm:
Possible solution to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4092

### DIFF
--- a/dependencyManagement/build.gradle.kts
+++ b/dependencyManagement/build.gradle.kts
@@ -85,7 +85,7 @@ val DEPENDENCY_SETS = listOf(
 val DEPENDENCIES = listOf(
   "ch.qos.logback:logback-classic:1.2.3",
   "com.blogspot.mydailyjava:weak-lock-free:0.18",
-  "com.github.ben-manes.caffeine:caffeine:2.9.0",
+  "com.github.ben-manes.caffeine:caffeine:3.0.2",
   "com.github.stefanbirkner:system-lambda:1.2.0",
   "com.github.stefanbirkner:system-rules:1.19.0",
   "com.google.auto.service:auto-service:1.0",


laurit:
Caffeine 3 requires java 11 our agent aims to support java 8. Could you verify that the agent still works with jdk8. You can run test with jdk8 by adding `-PtestJavaVersion=8` to gradle arguments, gradle will download jdk8 if needed.

brunojcm:
> Caffeine 3 requires java 11 our agent aims to support java 8. Could you verify that the agent still works with jdk8. You can run test with jdk8 by adding `-PtestJavaVersion=8` to gradle arguments, gradle will download jdk8 if needed.

It obviously doesn't, completely missed that.

The [release notes](https://github.com/ben-manes/caffeine/releases/tag/v3.0.0) say that:
> Highlights
>  Java 11 or above is required
>  Java 8 users can continue to use version 2.x, which will be supported

Any plans to support multiple releases? This is going to be very hard to fix if we need a single distribution targetting Java 8/11/17.

laurit:
> Any plans to support multiple releases? This is going to be very hard to fix if we need a single distribution targetting Java 8/11/17.

Although I haven't discussed with others I'm sure that everybody agrees that we want to continue providing a single distribution that works on all supported jvms. Having separate agents for different jvms is a hassle.
Luckily the way I see this issue is neither complex not serious enough to force us away from single agent.
We could just state that the agent depends on `jdk.unsupported` and we expect it to be available. Or we could shade caffeine 2&3 under different package names and use both. We only use a very small set of features from caffeine and all of our usages of caffeine go through a single builder class that could easily switch between caffeine 2&3 depending on jvm version. Though caffeine is the least of our problems in this `Unsafe` mess. Other usages of `Unsafe` in bytebuddy etc. seem more troublesome. We could try working with the authors of these frameworks or find replacements that we could use. Alternatively we could provide our own implementation of `sun.misc.Unsafe` that uses `jdk.internal.misc.Unsafe`, jdk11 also implements `sun.misc.Unsafe` this way. The last options would let the libraries that we use continue using `sun.misc.Unsafe` when running with module system without requiring adding dependency to `jdk.unsupported`.

trask:
thanks @brunojcm! closing with alternative solution #4154

PR #4097, opened 2021-09-12
Fix grpc instrumentation of callbacks

### LINKED ISSUE #4088: gRPC client interceptor: incorrect parent span attribution for spans following asynchronous calls
**Describe the bug**

Spans created after the completion of a gRPC client call are parented to the gRPC call span rather than the span that was active at the beginning of the client call.

**Steps to reproduce**
Create a span (P).  Make an asynchronous unary gRPC call (A) using `ClientCalls.futureUnaryCall`. Set up a callback on completion to invoke another gRPC client call (B).

Spans A and B should have the same parent (P)
Span A should end before Span B begins.

**What did you see instead?**

Span B is a child of Span A.  
Span A continues until the completion of span B.

v1.5.0

**Environment**
JDK: AdoptOpenJDK8
OS: RHEL7

**Additional context**
Discussed at Java instrumentation office-hours with @anuraaga, @trask and @jkwatson

trask:
Resolves #4088

### DIFF
--- a/instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcTest.groovy
+++ b/instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcTest.groovy
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.instrumentation.grpc.v1_6
 
+import com.google.common.util.concurrent.Futures
 import com.google.common.util.concurrent.MoreExecutors
 import example.GreeterGrpc
 import example.Helloworld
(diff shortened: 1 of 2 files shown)


laurit (approved):

instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcTest.groovy
```diff
@@ -143,6 +143,124 @@ abstract class AbstractGrpcTest extends InstrumentationSpecification {
     paramName << ["some name", "some other name"]
   }
 
+  def "test callback"() {
```
anuraaga, line 146: Can we add a test using GreeterFutureStub too? It should work but will test the more common usage.

instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcTest.groovy
```diff
@@ -143,7 +144,117 @@ abstract class AbstractGrpcTest extends InstrumentationSpecification {
...
+    }
+    ManagedChannel channel = channelBuilder.build()
+    GreeterGrpc.GreeterFutureStub client = GreeterGrpc.newFutureStub(channel)
+
+    when:
+    AtomicReference<Helloworld.Response> response = new AtomicReference<>()
+    AtomicReference<Throwable> error = new AtomicReference<>()
+    CountDownLatch latch = new CountDownLatch(1)
+    runWithSpan("parent") {
+      def future = client.sayHello(Helloworld.Request.newBuilder().setName("test").build())
```
anuraaga, line 176: It's probably worth using the idiomatic API

```
def future = Futures.transform(
  client.sayHelo(...)
  response -> {
    runWithSpan("child");
    return response;
  },
  MoreExecutors.directExecutor());

try {
  response.set(future.get());
 } catch (Throwable t) {
  error.set(t);
}
```

anuraaga (approved):

mateuszrzeszutek (approved):

PR #4098, opened 2021-09-12
More grpc events

### LINKED ISSUE #4089: gRPC client interceptor: creates SENT event in response to message being received from the server
**Describe the bug**
The gRPC client interceptor creates a `SENT` event on the current span in response to a [`onMessage(RESPONSE message)`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/4c64f70f94e5e5f4e80e3ff263ef7a477962d460/instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java#L118-L134) callback.

From the client perspective, this callback is invoked when a message is *received* from the server.

**Steps to reproduce**
Make any gRPC client call.

A `RECEIVED` event when responses from the server arrive.

**What did you see instead?**
A `SENT` event when a response from the server arrives.

`v1.5.0`

**Environment**
Compiler: AdoptOpenJDK8
OS: RHEL7

**Additional context**

Discussed with @anuraaga at Instrumentation office hours. Should consider whether it makes sense to create these events  for unary calls, when they really only make sense for streaming responses with multiple messages.

trask:
Resolves #4089 (I will split out the good question about unary calls to a separate issue).

This PR does 3 things:

* Fixes the "onMessage" event on the grpc client span from `SENT` to `RECEIVED`
* Adds (a correct) `SENT` event on the grpc client span
* Adds a `SENT` event on the grpc server span

(the diff looks worse than it is, because I moved a couple of nested classes inside of other nested classes due to needing to use the AtomicLongFieldUpdater from multiple places now)


instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
```diff
@@ -125,32 +107,57 @@ public void onMessage(RESPONSE message) {
               GrpcHelper.MESSAGE_ID,
```
anuraaga, line 107: I was looking for the SENT and found it in this "unchanged code" :P

anuraaga (approved):

mateuszrzeszutek (approved):

PR #4106, opened 2021-09-13
Change the default javaagent artifact to have exporters

### LINKED ISSUE #3959: Make opentelemetry-javaagent "all" artifact the default artifact
And introduce a "slim" artifact that has no exporters (same as what is now the default artifact).

Some notes from Aug 12 SIG meeting:

* The “-all” jar will be confusing once we move some of the instrumentation to contrib
  * Then it will not really be “all” (will likely have a separate javaagent released from contrib that has everything, similar to otel collector)
* The “-all” jar has already been a source of confusion
* For a while the default had exporters anyways

and see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3900#issuecomment-906096077

trask:
And introduce a new artifact with classifier "slim" ~which does not have any exporters (same as what used to be the default artifact).~ EDIT: which has only otlp/grpc (and using okhttp instead of netty for size)

Resolves #3959

Will send separate PR to be merged after the release updating:
* [README.md](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/README.md)
* [examples/extension/README.md](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/examples/extension/README.md)
* [benchmark-overhead/src/test/java/io/opentelemetry/agents/Agent.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/benchmark-overhead/src/test/java/io/opentelemetry/agents/Agent.java)


.github/workflows/patch-release-build.yml
```diff
@@ -265,6 +265,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: javaagent/build/libs/opentelemetry-javaagent-${{ github.event.inputs.version }}-all.jar
-          asset_name: opentelemetry-javaagent-all.jar
+          asset_path: javaagent/build/libs/opentelemetry-javaagent-${{ github.event.inputs.version }}.jar
+          asset_name: opentelemetry-javaagent.jar
```
anuraaga, line 269: I agree with the rename but it's sort of a breaking change for e.g., Dockerfiles that may be curling the "latest release" URL. Should we upload both at least for one release?

We may also consider whether it is bad practice in general to point at an unversioned URL - if so we can probably update our README to not point at the latest version URL to discourage usage.

docs/manual-instrumentation.md
```diff
@@ -18,7 +18,7 @@ or they might want to manually create spans for their own custom code.
 
 # Dependencies
 
-> :warning: prior to version 1.0.0, `opentelemetry-javaagent-all.jar`
+> :warning: prior to version 1.0.0, the Java agent
```
anuraaga, line 21: Delete this doc instead?

.github/workflows/patch-release-build.yml
trask: > Should we upload both at least for one release?

makes sense to me, I added attaching backwards compatible "all" artifact step to both release and patch release builds

> update our README to not point at the latest version URL to discourage usage

I like explicit versions in docs (and keeping it updated), IIRC we were stuck on how to automate this? maybe we can tack this problem on to #3516 @iNikem?

benchmark-overhead/src/test/java/io/opentelemetry/agents/LatestAgentSnapshotResolver.java
```diff
@@ -49,11 +49,12 @@ private String fetchLatestFilename(String version) throws IOException {
         .filter(elem -> {
           String classifier = $(elem).child("classifier").content();
           String extension = $(elem).child("extension").content();
+          // TODO (trask) this needs to be updated now that no more "all" artifact
           return "all".equals(classifier) && "jar".equals(extension);
```
trask, line 53: @breedx-splk any idea with this change if the classifier check should be changed to `""` or `null`?

anuraaga (approved):

mateuszrzeszutek (approved):

iNikem:
I vaguely remember that we wanted to make this change after extracting contrib portion. To minimize the number of times we change the meaning of our artifacts.

benchmark-overhead/src/test/java/io/opentelemetry/agents/LatestAgentSnapshotResolver.java
breedx-splk: Is the idea here that we'll be publishing the full agent with exporters now without the `-all` classifier? If so, then I think we should just remove the classifier check entirely.

It looks like you're also publishing `-all` for backwards compat? Assuming that's temporary, yeah, I think we'll want to remove this classifier check...but it might be nice to see what the xml looks like after the first publish to confirm.

trask: the `-all` artifact will be attached to the github release for now for backwards compatibility, but won't be published.

there will be a new artifact `-slim` published after this tho

> it might be nice to see what the xml looks like after the first publish to confirm

this works for me 👍

trask:
> I vaguely remember that we wanted to make this change after extracting contrib portion. To minimize the number of times we change the meaning of our artifacts.

I added to tomorrow's meeting agenda

trask:
> > I vaguely remember that we wanted to make this change after extracting contrib portion. To minimize the number of times we change the meaning of our artifacts.
> 
> I added to tomorrow's meeting agenda

moving back to draft for now and deferring until after the 1.6.0 release

trask:
> moving back to draft for now and deferring until after the 1.6.0 release

we discussed in SIG meetings this week, and I think there's general agreement to go forward with this PR now (see discussion in #3959)

I'll raise once more in next Tue SIG meeting and will merge after that if there's still agreement.

trask:
and will add otlp/http exporter to "-slim" as a follow-up PR (for now in this PR "-slim" has no exporters, same as the previous default artifact)

javaagent/build.gradle.kts
```diff
@@ -73,6 +78,10 @@ dependencies {
 
   exporterLibs(project(":javaagent-exporters"))
 
+  exporterSlimLibs("io.opentelemetry:opentelemetry-exporter-otlp")
+  exporterSlimLibs("io.opentelemetry:opentelemetry-exporter-otlp-metrics")
+  exporterSlimLibs("io.grpc:grpc-okhttp:1.41.0")
```
anuraaga, line 83: Just realized that because this is using `okhttp`, adding the HTTP exporters as well should be almost no additional bytecode, so maybe worth it?

anuraaga: Well assuming it doesn't cause dependency hell, wow that's old

https://github.com/grpc/grpc-java/blob/master/build.gradle#L158

anuraaga: Err - nevermind, yeah it wouldn't cause dependency hell since it's a different artifact

https://github.com/open-telemetry/opentelemetry-java/blob/main/dependencyManagement/build.gradle.kts#L78

So scratch the almost-no-additional-bytecode idea.

anuraaga: By the way, just some random trivia https://github.com/grpc/grpc-java/issues/6119#issuecomment-529990399

trask: > "OkHttp" name for the transport is a bit of a misnomer

this thread helped my mental picture 👍

anuraaga: It appears that we can exclude okhttp 2 from the transitive dependencies

https://github.com/open-telemetry/opentelemetry-java/pull/3678

though I added a comment on the grpc-java issue to see if there is anything obviously bad with that.

Currently I guess we effectively have two copies of okhttp2, the forked-into-grpc one and the transitive dependency which we don't seem to need.

docs/manual-instrumentation.md
```diff
@@ -18,11 +18,6 @@ or they might want to manually create spans for their own custom code.
 
 # Dependencies
 
-> :warning: prior to version 1.0.0, `opentelemetry-javaagent-all.jar`
-only supports manual instrumentation using the `opentelemetry-api` version with the same version
-number as the Java agent you are using. Starting with 1.0.0, the Java agent will start supporting
-multiple (1.0.0+) versions of `opentelemetry-api`.
-
```
iNikem, line 25: Are we sure about that? Have we tested that?

javaagent/build.gradle.kts
```diff
@@ -164,17 +184,15 @@ tasks {
     }
   }
 
-  // Includes everything needed for OOTB experience
-  val fullJavaagentJar by registering(ShadowJar::class) {
+  // Includes instrumentations, but not exporters
```
iNikem, line 187: Slim does include OTLP exporter, right?

docs/manual-instrumentation.md
trask: we have a smoke test that verifies interop with opentelemetry-api 1.0.0, and #3044 to track doing better than that with muzzle

(also, this is old doc, I'm removing it, see https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4106#discussion_r706967424)

javaagent/build.gradle.kts
trask: fixed 👍

javaagent/build.gradle.kts
trask: I added #4240 to discuss/track

iNikem (approved):

mateuszrzeszutek (approved):

PR #4111, opened 2021-09-13
Update ci.yml

iNikem:
Require successful `muzzle` step before publishing snapshots

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,7 +284,7 @@ jobs:
 
   snapshot:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


laurit (approved):

mateuszrzeszutek (approved):

trask:
it _might_ have been intentional for similar reason that it's excluded from release build:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/436184002c26f5d01594460afb122bab8603d7f7/.github/workflows/release-build.yml#L102-L104

but there should have been a comment in the file if it was intentional, and I'm good with this PR anyways

PR #4112, opened 2021-09-13
[WIP DO NOT MERGE] Convert gradle-plugins to a module instead of a separate project

iNikem:
I am definitely missing something here. It cannot be that easy.


mateuszrzeszutek:
We still hard-code gradle plugins' versions in `buildSrc/build.gradle.kts`:
```kotlin
  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.7.0")
  implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.7.0")
```
So even if you pull the `gradle-plugins` module into the main project the code generator will still run the old version; and you'll probably need to release a new snapshot plugins version before you can use it in instrumentations.

iNikem:
> We still hard-code gradle plugins' versions in `buildSrc/build.gradle.kts`:
> 
> ```kotlin
>   implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.7.0")
>   implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.7.0")
> ```
> 
> So even if you pull the `gradle-plugins` module into the main project the code generator will still run the old version; and you'll probably need to release a new snapshot plugins version before you can use it in instrumentations.

Yeah, now I remember, that was the problem. Not building the plugin together with other code, but actually using it...

iNikem:
I mean, however we package and version and whatnot our muzzle infrastructure, we have to make sure that newer agents can read and work with muzzle code generated by older versions of plugin. There is no war around that.

iNikem:
Writing down this idea before I forget it in my sleep...

So we have a problem, that our agent uses muzzle that uses our project. Thus we have a cyclic dependency between agent - muzzle plugin - agent. But actually this is wrong view. What we have is a "core agent machinery", which includes/publishes (but not depends on) muzzle and gradle plugin. And then we have instrumentations, which use that core machinery, including muzzle classes and muzzle gradle plugins. So no cycle: core -> muzzle/plugins -> instrumentations. And javaagent just combines all that together.

This way, if we can guarantee that newer agent/core will continue working with older instrumentations (built against older core), then we solve this problem of having muzzle gradle plugins as part of the main project.

Yes, that still means that `ClassRef` and friends are a part of stability contract between core and instrumentations.

anuraaga:
I can try this idea out but just want to run it by first - can't we put all the code in `gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle` into an artifact like `opentelemetry-muzzle-build`? And add it to the `codegen` configuration

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts#L20

and substitute the project dependency in for our own build

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts#L307

There's no compile dependency from the `gradle-plugins` kotlin code on the MuzzleCodeGenerationPlugin, we give a string which is used to invoke via reflection. So as long as it's on the classpath, which we use the `codegen` configuration to define, I think it should function the same

iNikem:
> I can try this idea out but just want to run it by first - can't we put all the code in `gradle-plugins/src/main/java/io/opentelemetry/javaagent/muzzle` into an artifact like `opentelemetry-muzzle-build`? And add it to the `codegen` configuration
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/buildSrc/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts#L20
> 
> and substitute the project dependency in for our own build
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts#L307
> 
> There's no compile dependency from the `gradle-plugins` kotlin code on the MuzzleCodeGenerationPlugin, we give a string which is used to invoke via reflection. So as long as it's on the classpath, which we use the `codegen` configuration to define, I think it should function the same

`codegent` says: 

> They are not part of the observer, they are part of the observation

`MuzzleCodeGenerationPlugin` must be by of the observer

iNikem:
> We still hard-code gradle plugins' versions in `buildSrc/build.gradle.kts`:
> 
> ```kotlin
>   implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.7.0")
>   implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.7.0")
> ```
> 
> So even if you pull the `gradle-plugins` module into the main project the code generator will still run the old version; and you'll probably need to release a new snapshot plugins version before you can use it in instrumentations.

Right now this will break our repo between two PRs. After muzzle declared stable, all (compatible) changes in muzzle (core) will result in a repo with newer core and older instrumentations (as they will still use older muzzle version). By stability guarantee this should work.

anuraaga:
Ah got it - yeah it's because the byte buddy plugin just calls into the bytebuddy java code directly. This is similar to the problems with the spotless plugin now - because it calls googlejavaformat directly, it can't be used on Java 16+. In principle, all similar plugins that call into Java code should be migrated to use the Gradle worker API, which can allow customizing the JVM environment, including classloader isolation (or even the JVM toolchain itself). I guess our ideal is to add `opentelemetry-muzzle-build` into the classpath of an isolated worker - then it's not tied to what's in buildSrc and can be substituted to anything, such as a project dependency

https://docs.gradle.org/current/javadoc/org/gradle/workers/WorkerExecutor.html#classLoaderIsolation-org.gradle.api.Action-

The byte buddy plugin isn't _that_ much code really

https://github.com/raphw/byte-buddy/blob/master/byte-buddy-gradle-plugin/src/main/java/net/bytebuddy/build/gradle/AbstractByteBuddyTask.java#L298

And if stripped away of configuration we don't use would be fairly small I think. I think we may consider removing the dependency on byte-buddy-plugin, fork in a simple byte buddy task to use from our codegen, which uses the worker API to fix the muzzle dance we have right now. We may want to consider this approach.

iNikem:
@anuraaga I still completely miss how will this help with the task of:

1. Change muzzle classes
2. Release new muzzle plugin version
3. Apply new muzzle plugin to our instrumentations

Ok, I understand that if we can substitute https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/74c1f55e35fdeccf88b58cd0f160dc93f242401c/buildSrc/build.gradle.kts#L32 with project dependency, then our instrumentations will use the current classes in the repo, not the published ones. But we still have to have a stable contract for them, for all external instrumentations, including contrib. So I am not sure how this isolation/substitution/vendoring in will be worth the effort? :)

anuraaga:
IIUC @trask and @mateuszrzeszutek  have been troubled that to use new codegen features, even if they follow a stable contract, can't actually be used in the agent without releasing plugins. I think it is probably worth some effort to try substitution if they we can avoid that.

iNikem:
How about a wild idea of better separating core from individual instrumentations/extensions? ;)

I mean, what is the use case of "I want to change muzzle codegen and use that in my instrumentation before publishing new Gradle plugin that uses new muzzle codegen"?

anuraaga:
It'd allow us to release the plugins along with the repo, and version them together. I think it provides the best user experience for users of the plugin that way, just pop in the plugin which has the version of the agent being targeted. It also reduces our release load a little, no need to "publish the plugins", they're just part of the normal release process.

The downside is some work in implementing a Gradle task using the Worker API but it seems OK, are there other issues with trying this approach?

iNikem:
> It'd allow us to release the plugins along with the repo, and version them together. I think it provides the best user experience for users of the plugin that way, just pop in the plugin which has the version of the agent being targeted. It also reduces our release load a little, no need to "publish the plugins", they're just part of the normal release process.
> 

My idea of separating core from instrumentations will achieve the same. Whole core has single version and our instrumentations can have the same version as the core they are using.

iNikem:
> The downside is some work in implementing a Gradle task using the Worker API but it seems OK, are there other issues with trying this approach?

Certainly no issue with trying it. I honestly cannot predict which approach will be easier in the long run. Our project is too complicated.

mateuszrzeszutek:
What about those `internal-...` instrumentations? They should be part of the core agent too, as it can't function without them.

trask:
> My idea of separating core from instrumentations will achieve the same. Whole core has single version and our instrumentations can have the same version as the core they are using.

@iNikem when you have time, can you open an issue to document/discuss this idea? I don't think I'm following it yet.

buildSrc/src/main/kotlin/otel.spotless-conventions.gradle.kts
```diff
@@ -24,6 +24,10 @@ spotless {
     kotlin {
       ktlint().userData(mapOf("indent_size" to "2", "continuation_indent_size" to "2", "disabled_rules" to "no-wildcard-imports"))
       licenseHeaderFile(rootProject.file("buildscripts/spotless.license.java"), "(package|import|class|// Includes work from:)")
+      targetExclude(
+        "src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts",
```
anuraaga, line 28: I guess `kotlin` looks at all files in the folder and `kotlinGradle` only looks at top level causing problem. This can probably be `src/main/kotlin/**/*.gradle.kts`

anuraaga (approved):
My understanding is regardless of whether we try removing the muzzle build plugin from the gradle classpath or not, this is a good change. It LGTM

trask (approved):
does this mean we can delete the `build-gradle-plugins` jobs in `pr.yml`, `ci.yml`, etc?

mateuszrzeszutek (approved):

iNikem:
> does this mean we can delete the `build-gradle-plugins` jobs in `pr.yml`, `ci.yml`, etc?

Yes, after making sure that other publishing jobs work correctly with plugins.

trask:
@iNikem is this replaced by #4183?

iNikem:
Yes.

PR #4113, opened 2021-09-13
Make InstrumenterBuilder javadoc about span suppression a bit clearer

mateuszrzeszutek:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -139,29 +143,33 @@
...
    * </ul>
    *
-   * <p><strong>In both cases:</strong>
+   * <p>For all other {@linkplain SpanKind span kinds} the suppression rules are as follows:
    *
    * <ul>
-   *   <li>SERVER and PRODUCER nested spans are always suppressed
-   *   <li>CONSUMER spans are suppressed depending on their operation: receive or process
-   *   <li>INTERNAL spans are never suppressed
+   *   <li>SERVER nested spans are always suppressed. If a SERVER span with is present in the parent
```
trask, line 166:
```suggestion
   *   <li>SERVER nested spans are always suppressed. If a SERVER span is present in the parent
```

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
mateuszrzeszutek: Typo 🙈

PR #4118, opened 2021-09-14
Use local gradle wrappers

trask:


trask:
nope, this is wrong, there is no local gradle wrapper in `gradle-plugins`, will update the docs instead: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/upgrading-gradle-wrappers.md#upgrading-the-gradle-wrappers

PR #4120, opened 2021-09-14
Update to OTel 1.6

anuraaga:


trask (approved):

mateuszrzeszutek (approved):

laurit (approved):

examples/distro/build.gradle
```diff
@@ -30,6 +30,7 @@ subprojects {
 
   repositories {
     mavenCentral()
+    mavenLocal()
```
anuraaga, line 33: We publishToMavenLocal in the workflow but pretty sure I needed to add this here to get that to kick in

PR #4122, opened 2021-09-14
Make CONSUMER receive span a parent of CONSUMER process spans in Kafka…

mateuszrzeszutek:
… instrumentations. `kafka-clients` and `spring-kafka` to be more precise -- `kafka-streams` needs to be converted to Instrumenter API first (another PR). Now `kafka-clients` and `spring-kafka` implement the batch receive (and process) conventions correctly.

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -29,8 +29,8 @@ dependencies {
   implementation(gradleApi())
   implementation(localGroovy())
 
-  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.7.0")
-  implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.7.0")
+  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.8.0-SNAPSHOT")
+  implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.8.0-SNAPSHOT")
 
   implementation("org.eclipse.aether:aether-connector-basic:1.1.0")
   implementation("org.eclipse.aether:aether-transport-http:1.1.0")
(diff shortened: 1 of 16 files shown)


instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/test/groovy/KafkaClientPropagationDisabledTest.groovy
```diff
@@ -3,7 +3,12 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import static io.opentelemetry.api.trace.SpanKind.CONSUMER
+import static io.opentelemetry.api.trace.SpanKind.PRODUCER
+
```
trask, line 8: you may need to update your Intellij groovy import configuration

buildSrc/build.gradle.kts
```diff
@@ -29,8 +29,8 @@ dependencies {
   implementation(gradleApi())
   implementation(localGroovy())
 
-  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.7.0")
-  implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.7.0")
+  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.8.0-SNAPSHOT")
+  implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.8.0-SNAPSHOT")
```
trask, line 33: is this intentional?

trask (approved):
thx 👍

buildSrc/build.gradle.kts
mateuszrzeszutek: Somewhat - I wondered whether I should release a 0.8.0 plugins version that depends on 1.6.0-SNAPSHOT or depend on 0.8.0-SNAPSHOT here, I thought the latter might be a bit simpler. I'm not sure what's the correct choice here though.

instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/test/groovy/KafkaClientPropagationDisabledTest.groovy
mateuszrzeszutek: Hmm, I can't seem to make this work...

mateuszrzeszutek: Okay, I've managed to fix that

buildSrc/build.gradle.kts
trask: I think there are no good choices, so let's go with this

Issue #4100, opened 2021-09-12
Kotlin Coroutines instrumentation fails if `jdk.unsupported` module is not available


brunojcm:
**Is your feature request related to a problem? Please describe.**
The current version of `opentelemetry-java-instrumentation` (1.5.3 as of this report) requires the `jdk.unsupported` JDK module to be available at runtime.

If `jdk.unsupported` is not present, the following error occurs:

```
[otel.javaagent 2021-09-12 04:59:34:994 +0000] [eventLoopGroupProxy-3-1] ERROR io.opentelemetry.javaagent.tooling.HelperInjector - Error preparing helpers while processing class kotlinx.coroutines.BuildersKt for kotlinx-coroutines. Failed to inject helper classes into instance jdk.internal.loader.ClassLoaders$AppClassLoader@46678e49
java.lang.UnsupportedOperationException: Cannot get defined package using reflection: sun.misc.Unsafe
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection$Dispatcher$Initializable$Unavailable.getDefinedPackage(ClassInjector.java:445)
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection.injectRaw(ClassInjector.java:221)
	at io.opentelemetry.javaagent.tooling.HelperInjector.injectClassLoader(HelperInjector.java:247)
	at io.opentelemetry.javaagent.tooling.HelperInjector.lambda$injectHelperClasses$1(HelperInjector.java:195)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.caching.WeakLockFreeCache.computeIfAbsent(WeakLockFreeCache.java:35)
	at io.opentelemetry.javaagent.tooling.HelperInjector.injectHelperClasses(HelperInjector.java:184)
	at io.opentelemetry.javaagent.tooling.HelperInjector.transform(HelperInjector.java:150)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.doTransform(AgentBuilder.java:10889)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.transform(AgentBuilder.java:10827)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.access$1600(AgentBuilder.java:10569)
[...]

Issue #4096, opened 2021-09-12
Jetty instrumentation fails if `jdk.unsupported` module is not available


brunojcm:
**Is your feature request related to a problem? Please describe.**
The current version of `opentelemetry-java-instrumentation` (1.5.3 as of this report) requires the `jdk.unsupported` JDK module to be available at runtime. This is not a sustainable approach.

If `jdk.unsupported` is not present, the following error occurs:

```
2021-09-11 13:21:15.629000 +0000 [main] INFO org.eclipse.jetty.server.Server - Started @9476ms
[otel.javaagent 2021-09-11 13:21:20:451 +0000] [qtp413177525-18] ERROR io.opentelemetry.javaagent.tooling.HelperInjector - Error preparing helpers while processing class org.eclipse.jetty.server.Request for servlet. Failed to inject helper classes into instance jdk.internal.loader.ClassLoaders$AppClassLoader@ea9e141
java.lang.UnsupportedOperationException: Cannot get defined package using reflection: sun.misc.Unsafe
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection$Dispatcher$Initializable$Unavailable.getDefinedPackage(ClassInjector.java:445)
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection.injectRaw(ClassInjector.java:221)
	at io.opentelemetry.javaagent.tooling.HelperInjector.injectClassLoader(HelperInjector.java:247)
	at io.opentelemetry.javaagent.tooling.HelperInjector.lambda$injectHelperClasses$1(HelperInjector.java:195)
	at io.opentelemetry.javaagent.shaded.instrumentation.api.caching.WeakLockFreeCache.computeIfAbsent(WeakLockFreeCache.java:35)
	at io.opentelemetry.javaagent.tooling.HelperInjector.injectHelperClasses(HelperInjector.java:184)
	at io.opentelemetry.javaagent.tooling.HelperInjector.transform(HelperInjector.java:150)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.doTransform(AgentBuilder.java:10889)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.transform(AgentBuilder.java:10827)
	at net.bytebuddy.agent.builder.AgentBuilder$Default$ExecutingTransformer.access$1600(AgentBuilder.java:10569)
[...]

PR #4124, opened 2021-09-14
Generate our own sun.misc.Unsafe if it is not available

laurit:
Couldn't resist doing this.
When running a modular application on jdk11 users will have to add `jdk.unsupported` module to make `sun.misc.Unsafe` available. If `sun.misc.Unsafe` is not available generate our own version of it that delegates to `jdk.internal.misc.Unsafe` just like `sun.misc.Unsafe` in jdk does.
~~This pr also helps grpc/netty that is bundled in the agent access some jdk internals that it currently can't access.~~
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4100
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4096

### DIFF
--- a/javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
+++ b/javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
@@ -26,7 +26,6 @@ import io.opentelemetry.javaagent.bootstrap.AgentClassLoader;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -197,7 +196,6 @@ class SunMiscUnsafeGenerator {
     addMethod("pageSize", int.class);
     addMethod("defineAnonymousClass", Class.class, Class.class, byte[].class, Object[].class);
     addMethod("getLoadAverage", int.class, double[].class, int.class);
-    addMethod("invokeCleaner", void.class, ByteBuffer.class);
   }
 
   private static List<String> getNameCandidates(String name) {
--- a/javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
+++ b/javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
@@ -1,0 +1,401 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.tooling;
+
+import static net.bytebuddy.jar.asm.Opcodes.ACC_FINAL;
+import static net.bytebuddy.jar.asm.Opcodes.ACC_PRIVATE;
+import static net.bytebuddy.jar.asm.Opcodes.ACC_PUBLIC;
+import static net.bytebuddy.jar.asm.Opcodes.ACC_STATIC;
+import static net.bytebuddy.jar.asm.Opcodes.ACC_SUPER;
+import static net.bytebuddy.jar.asm.Opcodes.ALOAD;
+import static net.bytebuddy.jar.asm.Opcodes.DUP;
+import static net.bytebuddy.jar.asm.Opcodes.GETSTATIC;
+import static net.bytebuddy.jar.asm.Opcodes.ILOAD;
+import static net.bytebuddy.jar.asm.Opcodes.INVOKESPECIAL;
+import static net.bytebuddy.jar.asm.Opcodes.INVOKESTATIC;
+import static net.bytebuddy.jar.asm.Opcodes.INVOKEVIRTUAL;
+import static net.bytebuddy.jar.asm.Opcodes.IRETURN;
+import static net.bytebuddy.jar.asm.Opcodes.NEW;
+import static net.bytebuddy.jar.asm.Opcodes.PUTSTATIC;
+import static net.bytebuddy.jar.asm.Opcodes.RETURN;
+
+import io.opentelemetry.javaagent.bootstrap.AgentClassLoader;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
```diff
@@ -0,0 +1,399 @@
...
+import java.util.List;
+import net.bytebuddy.jar.asm.ClassVisitor;
+import net.bytebuddy.jar.asm.ClassWriter;
+import net.bytebuddy.jar.asm.FieldVisitor;
+import net.bytebuddy.jar.asm.MethodVisitor;
+import net.bytebuddy.jar.asm.Opcodes;
+import net.bytebuddy.jar.asm.Type;
+
+/**
+ * Helper class for generating our own copy of sun.misc.Unsafe that delegates to
```
anuraaga, line 41: Just curious if you considered an approach that just compiles Java, I guess we'd be able to expose `jdk.internal` via `--add-exports`?

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/UnsafeInitializer.java
```diff
@@ -0,0 +1,77 @@
...
+      try {
+        SunMiscUnsafeGenerator.generateUnsafe(unsafeClass, (AgentClassLoader) classLoader);
+      } catch (Throwable throwable) {
+        logger.warn("Unsafe generation failed", throwable);
+        return;
+      }
+    }
+
+    // tell grpc/netty that it is ok to use setAccessible(true) so it can access DirectByteBuffer
+    System.setProperty("io.grpc.netty.shaded.io.netty.tryReflectionSetAccessible", "true");
```
trask, line 66: since we don't shade `io.grpc`, this could affect user app?

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
trask: this sounds nice if it can work

trask (approved):
very interesting... 😁

I opened #4128 to track creating a smoke test at some point to exercise this.

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/SunMiscUnsafeGenerator.java
laurit: My initial attempt was to generate a java version of sun.misc.Unsafe that uses method handles (I went for method handles from the start because I had a vague memory that some methods in internal unsafe have been renamed) to forward to internal unsafe but this failed with `error: package exists in another module: jdk.unsupported`. I did try a bit to coax it to work but failed. In retrospect I think I should have just moved it into a module compiled with java8, though then if I'd want to replace some method handles with direct calls I'd probably need to create a stub of internal unsafe. Anyway while fighting the compiler I realized that I could just generate bytecode and asm is my favourite framework and I haven't had a chance to use it for almost a year so it was pretty much drop everything else and start generating bytecode.

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/UnsafeInitializer.java
laurit: Oh, that is a good catch. I'll remove this.

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/UnsafeInitializer.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    Map<String, Set<Module>> exports = new HashMap<>();
+    // expose jdk.internal.misc.Unsafe to our agent
+    // this is used to generate our replacement sun.misc.Unsafe and also by grpc/netty to call
+    // jdk.internal.misc.Unsafe.allocateUninitializedArray
+    exports.put(
+        unsafeClass.getPackage().getName(),
+        Collections.singleton(UnsafeInitializer.class.getModule()));
+    instrumentation.redefineModule(
```
mateuszrzeszutek, line 40: Should we do that before the `if (testUnsafePresent...` check? If there's a `sun.misc.Unsafe` present in the classpath do we need to expose the `jdk.internal` one to the agent explicitly?

laurit: I noticed that grpc/netty that we use also desires access to some things that would require extra jvm arguments. Thought that might as well use the opportunity and let netty also use jvm internals. For example granting access to internal unsafe allows netty to allocate uninitialized arrays
```
17:58:55.427 INFO  i.o.s.LinuxTestContainerManager - STDERR: [otel.javaagent 2021-09-15 14:58:55:441 +0000] [main] DEBUG io.grpc.netty.shaded.io.netty.util.internal.PlatformDependent0 - jdk.internal.misc.Unsafe.allocateUninitializedArray(int): available
```
To be honest I have no idea whether this really is beneficial.

mateuszrzeszutek: Hmm, if exporter netty's able to use some of `jdk.internal` methods to make itself more efficient (I've no idea though, I just assume that 😅 ) that's a pretty good reason to expose `Unsafe` to the agent even if `sun.misc` is available.

mateuszrzeszutek (approved):

trask:
@laurit how did you test this without jlink? I'd like to try that out.

@open-telemetry/java-instrumentation-approvers planning to merge this before the 1.6.0 release tomorrow, unless someone prefers to delay it until after the release.

laurit:
@trask Turns out I didn't test this properly :(. The problem is that we add caffeine in boot loader not agent loader so it can't see the generated class.
To reproduce create a new project, add a main class and module-info. Leave the module info empty (later you can try adding `requires jdk.unsupported;`). Run main method with agent, agent will fail because unsafe is not found.

trask:
🥳

brunojcm-vgw:
Good job, guys, thanks for the fix :)

brunojcm:
Good job, guys, thanks for the fix :)

PR #4078, opened 2021-09-09
Convert servlet instrumentation to instrumenter api

laurit:
Part of #2713
Converts servlet, jetty and tomcat instrumentation to instrumenter api. Jetty is converted along with servlet to give an idea how server instrumenation that relies on servlet would look. Similarly tomcat is converted along with servlet to give an idea how server instrumenation that does not rely on servlet would look.
This pr deliberately avoids filling more attributes than what tracer api filled to avoid changing test and making this pr event larger. Code to fill more attributes is left commented out.

PR #4125, opened 2021-09-14
Make servlet context path key private

laurit:
Addresses a review comment from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4078


instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/BaseServletHelper.java
```diff
@@ -67,12 +67,8 @@ protected Context customizeContext(Context context, REQUEST request) {
...
   }
 
-  private Context addServletContextPath(Context context, REQUEST request) {
-    String contextPath = accessor.getRequestContextPath(request);
-    if (contextPath != null && !contextPath.isEmpty() && !contextPath.equals("/")) {
-      return context.with(ServletContextPath.CONTEXT_KEY, contextPath);
-    }
-    return context;
+  protected Context addServletContextPath(Context context, REQUEST request) {
+    return ServletContextPath.init(context, () -> accessor.getRequestContextPath(request));
```
trask, line 71: can we do non-capturing lambda here similar to #4014?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServletContextPath.java
```diff
@@ -22,18 +23,41 @@
 
   // Keeps track of the servlet context path that needs to be prepended to the route when updating
   // the span name
-  public static final ContextKey<String> CONTEXT_KEY =
+  private static final ContextKey<ServletContextPath> CONTEXT_KEY =
```
trask, line 26: do we need the `ServletContextPath` wrapper object?

instrumentation/servlet/servlet-common/library/src/main/java/io/opentelemetry/instrumentation/servlet/ServletHttpServerTracer.java
```diff
@@ -42,9 +43,11 @@
       Config.get().getBoolean("otel.instrumentation.servlet.experimental-span-attributes", false);
 
   private final ServletAccessor<REQUEST, RESPONSE> accessor;
+  private final Function<REQUEST, String> contextPathExtractor;
 
   protected ServletHttpServerTracer(ServletAccessor<REQUEST, RESPONSE> accessor) {
     this.accessor = accessor;
+    this.contextPathExtractor = (request) -> accessor.getRequestContextPath(request);
```
mateuszrzeszutek, line 50: Nit
```suggestion
    this.contextPathExtractor = accessor::getRequestContextPath;
```

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServletContextPath.java
laurit: Using wrapper helps to distinguish context path not yet set from context path empty. I guess I could use `""` to mark empty context path.

trask (approved):

PR #4130, opened 2021-09-15
Remove unneeded catch blocks

trask:
I tested this by throwing an exception inside of the try block, and it still called `onClose` (and so ended the span twice).


trask:
and a test changed for the better(!)

mateuszrzeszutek (approved):

laurit (approved):

PR #4134, opened 2021-09-15
Fix smoke test play build

trask:
This failed on the CI build after #4107 but we didn't notice.

I'll open an issue to track adding (a non-publishing) PR workflow to verify this.

And maybe auto-open a github issue when the smoke test image publishing builds fail, since we don't (and probably don't want to) run them nightly.


mateuszrzeszutek (approved):

mateuszrzeszutek:
> And maybe auto-open a github issue when the smoke test image publishing builds fail, since we don't (and probably don't want to) run them nightly.

💯

iNikem (approved):

PR #4151, opened 2021-09-17
Make CONSUMER receive span a parent of CONSUMER process spans in kafk…

mateuszrzeszutek:
…a-streams

This PR establishes in kafka-streams the same CONSUMER receive->process span relationship as in other instrumentations; now they're all consistent.


trask (approved):
thx!

PR #4153, opened 2021-09-17
Convert elasticsearch-rest to Instrumenter API

mateuszrzeszutek:
Part of #2713


instrumentation/elasticsearch/elasticsearch-rest-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/rest/ElasticsearchRestNetAttributesExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+
+import io.opentelemetry.instrumentation.api.instrumenter.net.NetAttributesExtractor;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.elasticsearch.client.Response;
+
+final class ElasticsearchRestNetAttributesExtractor
+    extends NetAttributesExtractor<String, Response> {
+  @Override
+  public @Nullable String transport(String s) {
+    return null;
```
laurit, line 16: Any reason for not returning `SemanticAttributes.NetTransportValues.IP_TCP`?

laurit (approved):

instrumentation/elasticsearch/elasticsearch-rest-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/elasticsearch/rest/ElasticsearchRestNetAttributesExtractor.java
mateuszrzeszutek: I initially didn't include it because the original tracer instrumentation didn't set this. You're tight though: the rest client is always using HTTP, so we can safely set that one. Added, thanks!

instrumentation/elasticsearch/elasticsearch-rest-common/javaagent/build.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-  id("otel.library-instrumentation")
+  id("otel.javaagent-instrumentation")
```
trask, line 2: 👍

trask (approved):

PR #4154, opened 2021-09-17
Use caffeine3 on jdk11 and caffein2 on older jdks

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4092


dependencyManagement/build.gradle.kts
```diff
@@ -17,6 +17,9 @@ rootProject.extra["otelVersion"] = otelVersion
 // Need both BOM and -all
 val groovyVersion = "2.5.11"
 
+rootProject.extra["caffeine2Version"] = "2.9.2"
+rootProject.extra["caffeine3Version"] = "3.0.3"
```
mateuszrzeszutek, line 21: 3.0.4 was released just a few days ago 😄 
```suggestion
rootProject.extra["caffeine3Version"] = "3.0.4"
```

mateuszrzeszutek (approved):
Nice 👍

instrumentation-api-caching/src/main/java/io/opentelemetry/instrumentation/api/caching/Caffeine2Cache.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.caching;
+
+import io.opentelemetry.instrumentation.api.internal.shaded.caffeine2.cache.Caffeine;
```
anuraaga, line 8: I was hoping it's possible to use the same package but put caffeine3 into the mrjar folder. And with just one implementation jvm would automatically pick the correct one. Could it work?

laurit: Unfortunately afaik boot loader does not support mrjars.

anuraaga (approved):

PR #4155, opened 2021-09-17
Add start/end time to RequestListener

mateuszrzeszutek:
Instrumentations that use custom `StartTimeExtractor` & `EndTimeExtractor` will have completely wrong duration metrics, since `RequestListener` implementations tended to get the time using `System.nanoTime()` without taking the custom times into account.
Currently it's just JMS & Kafka instrumentations that use custom time extractors and they do not have any metrics at all - but once we introduce messaging metrics it'll become a problem. (Also, the Armeria instrumentation might make use of those custom extractors; the tracer-using instrumentation certainly did)

CC @anuraaga


trask (approved):
nice find 👍

PR #4159, opened 2021-09-17
Allow reactor instrumentation to pick up spans from reactor context

lmolkova:
I'm trying to instrument my async library APIs that return e.g. `Mono` and never subscribe/block. Internally they do HTTP requests and logs.  I'm looking for a way for HTTP auto-instrumentation and MDC to pick up context from my library instrumentation.

Let's say that's the library call I'm trying to instrument:

```java
public Mono<String> call() {
    return Mono.defer( () -> {
        checkCurrenSpan("internal", true);

        Request request = new Request.Builder()
                .url("http://example.com")
                .build();

        httpClient.newCall(request).execute();
        return Mono.just("done");
    });
}
```

And let's say here's how I trace it

```java
private Mono<?> traceMono(Mono<?> publisher) {
    final AtomicReference<Span> spanRef = new AtomicReference<>(tracer.spanBuilder("traceMono").startSpan());

    var res =  publisher
        .doOnError(t -> spanRef.get().recordException(t))
        .doFinally(t -> spanRef.get().end());

   return res;
}
```

If I do this with agent enabled, I will have my HTTP spans unrelated.

```
LoggingSpanExporter - 'HTTP GET' : 73202ef91912cd88593276ccaa38b806 5aeffa52ef554cea CLIENT 
LoggingSpanExporter - 'traceMono' : 34b491175e64ec59f19c64e81be82283 b341875781ae7a78 INTERNAL
```

**There is no good way to make `traceMono` span current without leaking context.**

## Proposal

Let reactor instrumentation  capture trace context from `current` **or** from explicit reactor context. 
This way I don't ever need to make `traceMono` span current myself and instrumentation will make it current in all callbacks.

```java
private Mono<?> traceMono(Mono<?> publisher) {
    final AtomicReference<Span> spanRef = new AtomicReference<>(tracer.spanBuilder("traceMono").startSpan());

    var res =  publisher
        .doOnError(t -> spanRef.get().recordException(t))
        .doFinally(t -> spanRef.get().end())
        .contextWrite(Context.of(TracingSubscriber.EXPLICIT_SPAN_KEY, spanRef.get())); // !!!

   return res;
}
````

Result:

```
LoggingSpanExporter - 'HTTP GET' : f15dee3bda6ac670e720990c7b5aa049 436355277322d632 CLIENT 
LoggingSpanExporter - 'traceMono' : f15dee3bda6ac670e720990c7b5aa049 6019d03b614ca7b2 INTERNAL
```

Happy to add tests and address any feedback if we're in general ok with this approach.

### CODE REVIEW

**Comment by iNikem** [comment]
Our current reactive framework instrumentations (usually) try to execute all callbacks in the context that was active during subscribe. Do you propose to replace that or augment that with the ability to specify span context explicitly?

**Comment by lmolkova** [comment]
> Our current reactive framework instrumentations (usually) try to execute all callbacks in the context that was active during subscribe. Do you propose to replace that or augment that with the ability to specify span context explicitly?

I'm proposing to support both - full change is in this pr and adds explicit context support.

**Comment by trask** [comment]
I see this problem not as changing when context is captured (I think it should still be on subscribe), but rather allowing downstream publishers to influence the otel context of upstream publishers.

E.g. here's a more complex example that I would love to see create the trace `parent` -> `readItem` -> `httpGet` somehow:

```
public class Example {

    private static final Tracer tracer = GlobalOpenTelemetry.getTracer("test");

    private static final Mono<String> READ_ITEM = readItem();

    public static void main(String[] args) {
        Span span = tracer.spanBuilder("parent").startSpan();
        try (Scope ignored = span.makeCurrent()) {
            String item = READ_ITEM.block();
            System.out.println(item);
        }
    }

    static Mono<String> readItem() {
        Mono<String> readItem = httpGet("https://mydatabase").map(v -> "item: " + v);
        return traceMono(readItem, "readItem");
    }

    static Mono<String> httpGet(String url) {
        Mono<String> httpGet = Mono.just("http response from " + url);
        return traceMono(httpGet, "httpGet");
    }

    static <T> Mono<T> traceMono(Mono<T> publisher, String spanName) {
        ???
    }
}
```

**Comment by lmolkova** [comment]
@trask 

> I see this problem not as changing when context is captured

I'm not changing when context is captured, I'm only adding a new way how. It still happens in subscribe and at the same time as current context (just in the  common constructor overload)

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/3c4c46967987b30d628851934af25e0a9d7a2c1d/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/TracingSubscriber.java#L40-L57


Back to the example: in the way it's written, it's not parent -> readItem -> httpGet, since you don;t defer, resul is completed immediately, it's httpGet -> readItem -> main (in the order of execution and 3 unrelated spans :(
- `READ_ITEM = readItem()` completes before starting new span
- `Mono<String> readItem = httpGet("https://mydatabase").map(v -> "item: " + v);` completes before calling `traceMono`

So, let's change it to something with order of execution that we want

```java
private static final Mono<String> READ_ITEM = Mono.defer(App::readItem);

    public static void main(String[] args) {
        System.out.println("Starting span 'parent', Parent span id: -");
        Span span = tracer.spanBuilder("parent").startSpan();
        try (Scope ignored = span.makeCurrent()) {
            String item = READ_ITEM.block();
            System.out.println(item);
        }

        span.end();
    }

    static Mono<String> readItem() {
        Mono<String> readItem = Mono.defer(() ->
                httpGet("https://mydatabase").map(v -> "item: " + v));
        return traceMono(readItem, "readItem");
    }

    static Mono<String> httpGet(String url) {
        Mono<String> httpGet = Mono.defer(() -> Mono.just("http response from " + url));
        return traceMono(httpGet, "httpGet");
    }

    static <T> Mono<T> traceMono(Mono<T> publisher, String spanName) {
        System.out.printf("Starting span '%s', Parent span id: %s\n", spanName, Span.current().getSpanContext().getSpanId());
        Span span = tracer.spanBuilder(spanName).startSpan();
        return publisher
                .doOnError(t -> span.recordException(t))
                .doFinally(t -> span.end());
    }
```

The outcome is not great (readItem and  httpGet) are siblings

```
Starting span 'parent', Parent span id: -
Starting span 'readItem', Parent span id: 1e71b0e9fa1e371a
Starting span 'httpGet', Parent span id: 1e71b0e9fa1e371a
item: http response from https://mydatabase
[otel.javaagent 2021-09-20 09:14:13:616 -0700] [main] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'readItem' : 050b157bcd2a5d40e37f8af4aed0bd33 acf6e15336b0b616 INTERNAL [tracer: test:] AttributesMap{data={thread.name=main, thread.id=1}, capacity=128, totalAddedValues=2}
[otel.javaagent 2021-09-20 09:14:13:617 -0700] [main] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'httpGet' : 050b157bcd2a5d40e37f8af4aed0bd33 bb48e31f835c65f0 INTERNAL [tracer: test:] AttributesMap{data={thread.name=main, thread.id=1}, capacity=128, totalAddedValues=2}
[otel.javaagent 2021-09-20 09:14:13:624 -0700] [main] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'parent' : 050b157bcd2a5d40e37f8af4aed0bd33 1e71b0e9fa1e371a INTERNAL [tracer: test:] AttributesMap{data={thread.name=main, thread.id=1}, capacity=128, totalAddedValues=2}
```

Even if I add reactor context, this change won't help. 
Why? Because current() context is set too late - on `onNext` callback, but call to httpGet happens before any of the `TracingSubscriber` callbacks, where current is set:
- If you look into the screenshot below, first supplier is got
- then subscribe is called (which causes onSubscribe callback in `TracingSubscriber`

image


**So this is a problem**, but I'd argue it's a different one though. I can still write the example   in the way that it works:

## Version 1: explicit:

```java
```


## Version 2 implicit:

lmolkova:
So as a summary:
1.  there is a problem of not being able to make context current in  reactor - this PR is a solution
2.  another problem of current not being available when the supplier is called - this PR does not address it

While I'd like to solve p2 as well, and will look for the solution - p1 is still very valuable for the reactor instrumentation. I hope we can start with it and see how p2 can be fixed for the implicit case.

In our libraries, I'll still use explicit context with reactor context whenever possible.

lmolkova:
Chatted with @trask (thank you!) and the problem is that we don't want to start span until block/subscribe happens

```java
private static final Mono<String> READ_ITEM = readItem();

public static void main(String[] args) {

    var reactorInstr = TracingOperator.newBuilder().build();
    reactorInstr.registerOnEachOperator();

    Span span = tracer.spanBuilder("parent").startSpan();
    try (Scope ignored = span.makeCurrent()) {
        System.out.println("Starting span 'parent', Parent span id: -, started id: " + span.getSpanContext().getSpanId());
        String item = READ_ITEM
                .block();
        System.out.println(item);
    }

    span.end();
}

static Mono<String> readItem() {
    Mono<String> readItem = httpGet("https://mydatabase").map(v -> "item: " + v);
    return traceMono(readItem, "readItem");
}

static Mono<String> httpGet(String url) {
    Mono<String> httpGet = Mono.defer(() -> Mono.just("http response from " + url));
    return traceMono(httpGet, "httpGet");
}

static  Mono<String> traceMono(Mono<String> publisher, String spanName) {
    AtomicReference<Span> s = new AtomicReference<>(null);
    return publisher
            .doOnError(t -> s.get().recordException(t))
            .doFinally(t -> s.get().end())
            .contextWrite(ctxUpstream -> {
                    SpanBuilder builder = tracer.spanBuilder(spanName);
                    Span explicitParent = ctxUpstream.getOrDefault(EXPLICIT_SPAN_KEY, Span.getInvalid());

                    if (explicitParent.getSpanContext().isValid()) {
                        builder.setParent(io.opentelemetry.context.Context.current().with(explicitParent));
                    }

                    Span span =  builder.startSpan();
                    s.set(span);
                    return ctxUpstream.put(EXPLICIT_SPAN_KEY, span);
            });
}
```

So, the pattern here:
- we start span in `contextWrite` callback (!) 
- we allow `current` and explicit parent 

The example above works perfectly without this PR because explicit context is always used
image

## Implicit context for auto-instrumentation

Let's say we do this:
```java
static Mono<String> httpGet(String url) {
    Mono<String> httpGet = Mono.defer(() -> {
        Request request = new Request.Builder()
                .url("http://example.com")
                .build();

        try {
            httpClient.newCall(request).execute();
        } catch (IOException e) {
            e.printStackTrace();
        }

        return Mono.just("item");
    });
    return traceMono(httpGet, "httpGet");
}
```

Results in:
image

Now, even if I add this PR, we have problem2 - `httpGet` mono execution starts outside of `TracingSubscriber` (reactor instrumentation) callback:
image

## Hacky solution

I can still write my code in the way that `publisher` in `traceMono` is called in the `TracingSubscriber` callback rather than outside, but it smells and gets fragile.  

This is the full example
```java

  static Mono<String> readItem() {
      // !!! have to defer here
      Mono<String> readItem = Mono.defer(() -> httpGet("http://example.com").map(v -> "item: " + v)); 
      return traceMono(readItem, "readItem");
  }

  static Mono<String> httpGet(String url) {
      Mono<String> httpGet = Mono.defer(() -> {
          System.out.println("Making http request, current span: " + Span.current().getSpanContext().getSpanId() );
          Request request = new Request.Builder()
                  .url(url)
                  .build();

          try {
              httpClient.newCall(request).execute();
          } catch (IOException e) {
              e.printStackTrace();
          }

          return Mono.just("item");
      });
      return traceMono(httpGet, "httpGet");
  }

  static  Mono<String> traceMono(Mono<String> publisher, String spanName) {
      AtomicReference<Span> s = new AtomicReference<>(null);
      // have to hack here, so publisher is called in TracingSubscriber.onNext callback, not before corresponding  TracingSubscriber is created
      return Mono.just("")
              .flatMap(t -> publisher)
              .doOnError(t -> s.get().recordException(t))
              .doFinally(t -> s.get().end())
              .contextWrite(ctxUpstream -> {
                  SpanBuilder builder = tracer.spanBuilder(spanName);
                  Span explicitParent = ctxUpstream.getOrDefault(EXPLICIT_SPAN_KEY, Span.getInvalid());

                  if (explicitParent.getSpanContext().isValid()) {
                      builder.setParent(io.opentelemetry.context.Context.current().with(explicitParent));
                  }

                  System.out.printf("Starting span '%s', Parent span id (current): %s,  Parent span id (ctx): %s\n",
                          spanName,
                          Span.current().getSpanContext().getSpanId(),
                          explicitParent.getSpanContext().getSpanId());

                  Span span =  builder.startSpan();
                  s.set(span);
                  return ctxUpstream.put(EXPLICIT_SPAN_KEY, span);
              });
  }
```

lmolkova:
So, can we discuss it at one of the instrumentation meetings? Can someone more familiar with the reactor help to find a more elegant approach?

And just to clarify: this PR, even though it does not solve `current` context issue for the most important case, still feels like an improvement (we can propagate spans through reactor without making them current to at least onNext callback).

trask:
> So, can we discuss it at one of the instrumentation meetings?

definitely, I just added it to the agenda for Thu, and will try to understand things a bit better before then

> Can someone more familiar with the reactor help to find a more elegant approach?

cc: @HaloFour 👀

HaloFour:
I've been thinking about a general API for tracing Reactor that didn't entirely rely on the instrumentation/operator being installed.  It's something I'd be interested in adding to the instrumentation library, but it would work a bit differently than what is proposed here.

Beyond that I'm not sure I understand the purpose of this PR.  The TracingOperator is supposed to capture the current Context at the point of assembly, so the following should work fine and not leak:

```java
private Mono<?> traceMono(Mono<?> publisher) {
    Span span =  tracer.spanBuilder("traceMono").startSpan();

    try (Scope scope = Context.current().with(span)) {
        return publisher
            .doOnError(t -> span.recordException(t))
            .doFinally(t -> span.end());
    }
}
```

Apart from that I'm concerned that the proposed PR captures the Span separately from the Context and then tries to tie them back together.  If we were to put anything into the Reactor context it feels like it should be the full OTel Context.

As for creating the span on subscribe, which is certainly necessary for cold and reusable publishers, that is something probably better served with a separate tracing operator that can be applied with `transform`.  Apart from working better with the fluent Reactor APIs it would also make it very easy to trace each subscription.

anuraaga:
> As for creating the span on subscribe, which is certainly necessary for cold

Unfortunately I don't think there is a way for code to know if a Mono is going to be subscribed in the future or not, so we couldn't ever use an approach in auto instrumentation that doesn't trace on subscription. While manual code could probably do it, it feels hairy to have both approaches interacting with each other.

One issue we often have is when a framework does not provide a filter-like API for interception. For example, tracing seems best implemented in `doOnSubscribe`

https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#doOnSubscribe-java.util.function.Consumer-

The javadoc describes perfectly what we want to do, add behavior when the Mono is surprised. But without something like `delegate.next()`, we cannot make the context current, which would otherwise be picked up by our instrumentation to then allow propagation throughout the reactive stream.

Reactive frameworks could be nice to us and provide such a callback which I think could help here. But I suspect their general stance would be, use the reactive context. So in this sense allowing reading from either `Context.current()` or the reactor context in our instrumentation does seem like a reasonable feature. I can't think of an alternative for @lmolkova's use case, which is a valid one. My main concern is whether it's always correct to prioritize the reactor context, or is there a case where the threadlocal is still the correct parent despite there being something in the reactor context.

Also, agree with @halofour that we want to expect the OTel Context object in the reactor context, not just a Span. And rather than expose a String, we should expose static methods for 1) create new reactor context with an OTel context and 2) read OTel context from reactor context.

HaloFour:
> One issue we often have is when a framework does not provide a filter-like API for interception. For example, tracing seems best implemented in `doOnSubscribe`.

I'd propose an API utilize `Mono#transform`/`Flux#transform` + `Operators#lift` which allows you to basically do what `TracingOperator` does and decorate the subscriber.  That gives easy access to all of the callbacks necessary to start the span on subscription, end that span on complete/error/cancel, propagate context onto the subscriber/publisher thread, and access the Reactor context which is otherwise tricky.  For the consumer it could look as simple as:

```java
Mono<T> mono = ...;
Mono<T> traced = mono.transform(withSpan("MySpan"));
```

I started to work on a PoC of this when I was toying with multiple subscriptions via `@WithSpan`.

> My main concern is whether it's always correct to prioritize the reactor context, or is there a case where the threadlocal is still the correct parent despite there being something in the reactor context.

This is definitely my biggest concern here as well.  It's quite likely that the Context captured on assembly is the better option but that could be difficult to discern.  I've not fully thought it through but perhaps the Reactor operator could put a cookie into the OTel context that is captured in the Reactor context.  The operator could compare the cookies within the current/captured OTel context with that in the Reactor context to determine if the former is based on the latter and would be a better context to use.  But that probably doesn't solve every case.

HaloFour:
@trask 

> here's a more complex example

Here's a real quick&dirty implementation of the TracingOperator API I suggested that can do this:  https://gist.github.com/HaloFour/6cac6653120da96fc1e8e1effca7b5b5

The method `TracingOperator#figureOutParentContext` is where the operator needs to figure out the correct context to use and I don't know that there is a cut&dry way to do that.  The precedence of which Context to use might change depending on the style, especially OnAssembly vs. Current.  And it might be challenging to discern when to favor one of those Contexts vs. one stored in the Reactor subscriber context.  Perhaps an API could offer a way to hint which Context to use.

lmolkova:
Thanks everyone for the input! A few comments

1. Span -> Context  in reactor context - sure
2. Reactor context vs current: If we choose current, it would mean it will never be picked. I.e. it sounds like the rule of thumb would be: if you're using reactor, always propagate span through reactor context (which @HaloFour instrumentation is perfect for).
    - HTTP SERVER span (current)
       -  new span, write it to reactor context
          - new span: HTTP SERVER would win as its current
3. Instrumentation through `transform` looks great!!  It doesn't work for my example though as cold start happens out of any subscriber callback, i.e. before span is made current

```java
 static Mono<String> readItem() {
      return httpGet("http://example.com").map(v -> "item: " + v)
              .transform(TracingOperator.withSpan("readItem"));
  }

  private static OkHttpClient httpClient = new OkHttpClient();
  static Mono<String> httpGet(String url) {
      return Mono.defer(() -> {
          System.out.println("Making http request, current span: " + Span.current().getSpanContext().getSpanId());
          Request request = new Request.Builder()
                  .url(url)
                  .build();

          try {
              httpClient.newCall(request).execute();
          } catch (IOException e) {
              e.printStackTrace();
          }

          return Mono.just("item");
      }).transform(TracingOperator.withSpan("httpGet", builder -> builder
                      .setAttribute("url", url)));
  }
```

```
Started span 'parent',  Parent span id, started id - -: bcc2134a786970bb
Started span 'readItem',  Parent span id - bcc2134a786970bb, started id - cd9d2474f0596f96
Started span 'httpGet',  Parent span id - cd9d2474f0596f96, started id - b9a9159cdab07ab7
Making http request, current span: bcc2134a786970bb
Making span current 'onSubscribe', current id - b9a9159cdab07ab7
```

I'll keep playing with it and let's talk about it tomorrow's meeting.

lmolkova:
Another discussion point on exposing APIs on `TracingOperator`:

>  A little copying is better than a little dependency.

1. It's useful and looks great, but using it requires dependency on reactor instrumentation library. We, in Azure SDKs, would never depend on it, however beautiful it is (because of dependency hell), and would rather create spans ourselves and set them in the context. I guess other libraries would have the same concern
2. End users would probably rarely discover it unless it's in an annotations package (that'd have to depend on reactor - bummer)

I think the main (and huge!) value comes as an example and guidance on how to instrument reactor - may be a blog post or md file in this repo?  I promise to steal from it a lot.

HaloFour:
@lmolkova 

> A little copying is better than a little dependency.

I understand the sentiment but I do think that copied code will not be trivial.  It also seemingly doesn't make sense to have a proposal here for behavior in the Reactor instrumentation if it is a goal to not use the Reactor instrumentation.  Something needs to extract that explicit Reactor context into the current Context for the rest of the SDK and instrumentation to work with it.

lmolkova:
> @lmolkova
> 
> > A little copying is better than a little dependency.
> 
> I understand the sentiment but I do think that copied code will not be trivial. It also seemingly doesn't make sense to have a proposal here for behavior in the Reactor instrumentation if it is a goal to not use the Reactor instrumentation. Something needs to extract that explicit Reactor context into the current Context for the rest of the SDK and instrumentation to work with it.

`contextWrite` seems to be this minimal sufficient API that allows me to pass context, so as long as I can rely on key not changing, I'm happy.
I'm saying there is no strong need to expose any API (except enabling instrumentation) as long as one in the agent works. And guidance is to how to write my code so it works properly with this auto-instrumentation

HaloFour:
@lmolkova 

> `I'm saying there is no strong need to expose any API (except enabling instrumentation) as long as one in the agent works. And guidance is to how to write my code so it works properly with this auto-instrumentation

Ah, I get you now.  It's not that you're avoiding the instrumentation, it's that you don't want a dependency on the library APIs.  That does make more sense.  I assume that would also preclude any helper methods that would avoid leaking a String key to the consumer.

I am still quite concerned about how this feature would interact with other instrumentation around Reactor or libraries that utilize Reactor.  It seems like it would be really easy for me for the OTel Context stored in the Reactor Context to not be the correct context.

lmolkova:
>  It seems like it would be really easy for the OTel Context stored in the Reactor Context to not be the correct context.

Do you have any examples of it? I'd like to play and find some reasonable solutions. My point here is that we know for sure current trace context does not work (see my example above). Or try running the same example with `@WithSpan` annotation - it does not work well.

It sounds like reactor context works at least better than `current()` and all confusion comes from the fact that we have two different contexts. 

So I hope to find some reasonable solution here.
The one I have in mind is to mostly rely on reactor context: anything that instruments reactor calls has to follow some rules: e.g.
  - when starting spans, give preference to reactor context over current()
  - start span and store it in the reactor context behind standard known key

It involves some manual steps, but people who create spans are probably willing to put a bit extra effort to make their instrumentation work (and current() doesn't).

I'm new to the reactor world (which I'm working on), so I'm looking for some examples here, based on the community expertise, where this solution does not work (or works worse than current()) and am happy to adjust it.

HaloFour:
> Do you have any examples of it?

I've not tried to put together a proof of concept, but if your traced Reactor chain used `flatMap` to invoke other instrumented code that would result in child spans which would not be reflected in that Reactor context.  Any subsequent instrumentation in that reactive flow would pick up that Reactor context and not the correct context containing those child spans.

> It sounds like reactor context works at least better than `current()` and all confusion comes from the fact that we have two different contexts.

I disagree that the Reactor context would be better.

> Or try running the same example with `@WithSpan` annotation - it does not work well.

Ultimately I think the problem is not one of explicitness, it's that the `TracingOperator`/`TracingSubscriber` provided by auto-instrumentation is not handling certain use cases.  I think it'd be more ideal to identify these cases and see how the operator can be modified so that they work "out of the box", that way you wouldn't have to write any code to wrangle the context.

lmolkova:
Thanks, everyone for your help and discussions!

Quick summary:
- examples of Mono tracing and how mixed context propagation may work https://gist.github.com/lmolkova/156a9e6453f5d3f8242689cfe6a6c07a
- it seems reasonable, doesn't break, and allows tracing nested non-subscribing reactor calls

Next steps:
- I'll update this PR with API that allows reading/writing tracing context to reactor context.
  - Context keys should be hidden
  - Users (apps/libraries) are expected to call into this API  manually 
- Convenience tracing API that traces publishers (not in the scope of this PR),  @HaloFour's proposal looks great https://gist.github.com/HaloFour/6cac6653120da96fc1e8e1effca7b5b5

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -53,6 +63,52 @@ public static TracingOperatorBuilder newBuilder() {
...
+   *
+   * @param context Reactor's context to get trace context from.
+   * @param defaultTraceContext Default value to be returned if no trace context is found on Reactor
+   *     context.
+   * @return Trace context or default value.
+   */
+  public static Context fromContextOrDefault(
+      reactor.util.context.Context context, Context defaultTraceContext) {
+    return context.getOrDefault(TRACE_CONTEXT_KEY, defaultTraceContext);
+  }
```
trask, line 110: what do you think of:

* `storeInContext()` -> `storeOpenTelemetryContext()`
* `fromContextOrDefault()` -> `getOpenTelemetryContext()`

also, I think (future) renaming of the class might help:

* `TracingOperator` -> `ReactorTracing`
* `TracingOperatorBuilder` -> `ReactorTracingBuilder`

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java
```diff
@@ -44,6 +47,13 @@ public static TracingOperatorBuilder newBuilder() {
     return new TracingOperatorBuilder();
   }
 
+  private static final ContextKey<Context> TRACE_CONTEXT_KEY =
+      ContextKey.named("otel-trace-context");
+
+  // have to be re-initialized after hooks are set/reset
+  private static Mono<String> dummyMono = Mono.just("");
+  private static Flux<String> dummyFlux = Flux.just("");
```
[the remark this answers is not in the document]
lmolkova: Unfortunately, they have to be updated when hook is registered/reset, otherwise, they will always make `runInScope` 
 populate current from the context

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
```diff
@@ -0,0 +1,82 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.reactor;
+
+import io.opentelemetry.context.Context;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class ReactorTracing {
```
anuraaga, line 12: I think we'll eventually need to provide help for actually adding spans for `Mono`s since it's pretty tedious. Since different `Mono`s may have very different attributes / semantics, I guess it will be common to have multiple `ReactorTracing` in an application.

Mix-and-match of that and a static initializer like this one does seem like it can get confusing. How about we rename this to `ReactorContextPropagation` to make clear this is all about propagation?

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static synchronized void disable() {
+    // have to be reset as they capture async strategy and lift
+    dummyMono = Mono.just("");
+    dummyFlux = Flux.just("");
+  }
+
+  /** Forces Mono to run in traceContext scope. */
+  static <T> Mono<T> runInScope(Mono<T> publisher, Context tracingContext) {
```
anuraaga, line 38: I think we'd usually use naming `runWithContext` though not a big deal

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
```diff
@@ -0,0 +1,82 @@
...
+    // this hack forces 'publisher' to run in the onNext callback of `TracingSubscriber`
+    // created for this publisher and with current() span that refer to span created here
+    // without the hack, publisher runs in the onAssembly stage, before traceContext is made current
+    return dummyFlux
+        .flatMap(i -> publisher)
+        .subscriberContext(ctx -> storeOpenTelemetryContext(ctx, tracingContext));
+  }
+
+  /**
+   * Stores Trace {@link io.opentelemetry.context.Context} in Reactor {@link
```
anuraaga, line 58: Especially with renaming to `ReactorContextPropagation` it becomes less awkward to remove these references to tracing which we don't need

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
```diff
@@ -0,0 +1,82 @@
...
+        @Override
+        public String toString() {
+          return "otel-trace-context";
+        }
+      };
+
+  static volatile Mono<String> dummyMono;
+  static volatile Flux<String> dummyFlux;
+
+  static synchronized void enable() {
```
anuraaga, line 25: Also it's probably good to merge the public API of `TracingOperator` into this class if I'm not mistaking anything. We just want one public entry point for reactor context propagation. I guess then a user would do `ReactorContextPropagation.enable()` or `ReactorContextPropagation.disable()`, those methods would call into TracingOperator, instead of other way around.

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  /**
+   * Stores Trace {@link io.opentelemetry.context.Context} in Reactor {@link
+   * reactor.util.context.Context}.
+   *
+   * @param context Reactor's context to store trace context in.
+   * @param traceContext Trace context to be stored.
+   */
+  public static reactor.util.context.Context storeOpenTelemetryContext(
```
anuraaga, line 64: Instead of exposing `store/get`, can we expose `runInScope` (`runWithContext`)? It seems to be the better API

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
anuraaga: Oh I had missed trask's comment, it seems to overlap. 

> Yeah, I can do it. I thought that as we'll add more tracing stuff (start/end span convenience) having configuration (register/reset hooks) on the same class may get noisy, so I thought about separating tracing.

I agree with this - but my understanding is the operator is about context propagation and not creating spans, so it would still be appropriate to merge in

instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ReactorTracing.java
lmolkova: We can, but it implies using a hack with dummy mono/flux even when it's not needed (not a big deal). But I think we ultimately don't need `runWithContext` - having something like `tracePublisher(Publisher, SpanBuilder, AttributeExtractors?, whatnot)`  will eliminate the need for it and provide convenience.

For those who want low-level control (e.g. if I want to start/end a span conditionally or in some funky way), get/set context is the minimal needed functionality

lmolkova: also, if I need to inject headers or enrich span, I'd prefer explicit context from reactor (i.e. `get` method) over `Context.current()`. Since current can change in any user callback, I never know what it refers to. There are no guarantees with explicit reactor context either, so it's rather a preference.

iNikem:
For existing customers of the javaagent who monitors their Reactor-Netty call, will anything change after this PR? I mean, will the produced telemetry change?

anuraaga:
@iNikem As far as I know, the only existing instrumentation affected by this change is `@WithSpan`.

@lmolkova Do you think it's worth having a test including `@WithSpan` and a reactor-netty call to confirm the propagation works between them? Seems like it should, but is a good "real world" test case probably

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/ReactorNettyWithSpanTest.groovy
```diff
@@ -0,0 +1,80 @@
...
+import reactor.test.StepVerifier
+import spock.lang.Shared
+
+import java.time.Duration
+
+import static io.opentelemetry.api.trace.SpanKind.CLIENT
+import static io.opentelemetry.api.trace.SpanKind.INTERNAL
+import static io.opentelemetry.api.trace.SpanKind.SERVER
+
+class ReactorNettyWithSpanTest extends InstrumentationSpecification implements AgentTestTrait {
```
anuraaga, line 21: Thanks!

HaloFour:
@trask 

Looks good to me!  The approach with the "dummy" publisher and `flatMap` is interesting and a clever way to force the OTel context.

PR #4160, opened 2021-09-17
Remove incorrect lettuce-4.0 db.statement attribute

trask:
following #4152, and removing `db.statement` since it's not really capturing `db.statement`

(note: lettuce-5.0 and lettuce-5.1 do capture `db.statement` correctly, I looked and it wasn't straight forward to port that back to lettuce-4.0)


trask:
Hm, but `db.statement` is marked, so maybe existing behavior is better?

> Required if applicable and not explicitly disabled via instrumentation configuration.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#call-level-attributes

mateuszrzeszutek (approved):
In case of lettuce 4.0 we really *can't* extract the statement, the library doesn't allow doing that. I think it's probably better not to set the statement if we can't get the real value.

trask:
> I think it's probably better not to set the statement if we can't get the real value.

👍

also, I just noticed this part of the spec that I missed previously:

> `db.operation` | Required, if `db.statement` is not applicable.

Issue #4136, opened 2021-09-15
Use latest collector release for smoke test fake backend


trask:
See TODO:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8c23274d7a293f48c244a8a77e451ba8d1bc366b/smoke-tests/fake-backend/build.gradle#L72-L77

trask:
Closing in favor of #4186

PR #4170, opened 2021-09-21
Use the latest collector (from new repo) in smoke test images.

breedx-splk:
Maybe a pretty inelegant way of tackling #4136.

It's leveraging the github `latest` redirect URLs for finding the latest version number, and then using that in url building (which breaks my RESTful sensibilities...but might be ok).

I'm not a gradle expert, but this seems to get the right files in the right places for the rest to continue.

### DIFF
--- a/smoke-tests/fake-backend/build.gradle
+++ b/smoke-tests/fake-backend/build.gradle
@@ -64,18 +64,33 @@ task windowsBackendImageBuild(type: DockerBuildImage) {
 
 def collectorDockerBuildDir = new File(project.buildDir, "docker-collector")
 
-task windowsCollectorBinaryDownload(type: Download) {
-  doFirst {
-    collectorDockerBuildDir.mkdirs()
-  }
+task windowsCollectorBinaryDownload(){
 
-  // TODO (trask) update this to point to new opentelemetry-collector-releases repo
-  //   this involves making two further changes:
-  //     * all the artifacts in the new repo now have version in their name (so no single url for "latest")
-  //     * the artifacts are now tar.gz format, so need to extract the exe
-  //   note: version is hardcoded below because that is last version in the collector repo that has the releases attached
-  src("https://github.com/open-telemetry/opentelemetry-collector/releases/download/v0.33.0/otelcol_windows_amd64.exe")
-  dest(collectorDockerBuildDir)
+  def latestDownloadUrl = "https://github.com/open-telemetry/opentelemetry-collector-releases/releases/latest/download"
+  def conn = new URL(latestDownloadUrl).openConnection()
+  conn.followRedirects = false
+  def latestVersion = conn.headerFields["Location"][0].replaceFirst(/^.*\/v/, '')
+  def versionedTarFile = new File("${collectorDockerBuildDir}/otelcol_${latestVersion}_windows_amd64.tar.gz")
+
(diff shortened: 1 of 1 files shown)


anuraaga:
Thanks @breedx-splk - this has brought to light several issues to point out and maybe take a different approach.

Currently, our linux and windows smoke tests are very different. Linux smoke tests run against latest HEAD of the collector by nature of using the `opentelemetry-collector-contrib-dev` image.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/smoke-tests/src/test/java/io/opentelemetry/smoketest/LinuxTestContainerManager.java#L48

I think our Windows tests have always been pointing at the latest released version. I think we should have consistent behavior for both. Initially, pointing to dev was important as protocol changed a lot and keeping instrumentation / SDK up to date was more effective when targeting -dev image, and SDK snapshots. Now we don't target SDK snapshots and protocol doesn't change that much (or at least we are generally being able to keep up on releases) so it seems to make sense to have the Linux tests target a release as well.

So then we can decide whether we want to automatically target the latest release, or pin to a release version - @jkwatson brought up the idea of pinning in the SDK repo, for reference. This prevents the build breaking with unrelated changes like the health check extension being removed. I don't see a big advantage in using the latest release version automatically, vs pinning a version, and the later is considered sort of good practice to ensure build reproducibility.

Though we could also just remove the collector from the smoke tests anyways - the SDK repo is suprisingly late to it but does have collector integration tests now.

One of the main reasons for the collector in our smoke tests seems to be testing zipkin / jaeger exporter - I don't think we need such smoke tests in our repo, data export is implemented in SDK and completely independent from instrumentation. It seems like overkill here and we could simplify our tests by removing the collector IIUC.

iNikem:
If we remove Collector from our smoke tests, how do we verify that telemetry leaves the JVM under test in a good shape?

anuraaga:
@iNikem They'd still be hitting the fake backend which we would still use to check the exported data.

iNikem:
> @iNikem They'd still be hitting the fake backend which we would still use to check the exported data.

If we are confident that SDK correctly implements OTLP exporter, then yes, we can skip Collector from our tests

anuraaga:
Yeah my thinking is it's the SDK repo's job to make sure it does that, so having OTLP collector integration tests there is very important. But then having essentially the same integration testing coverage in every other repo just seems like overkill. If it helps with our windows smoke testing, then simplifying seems like a reasonable win for us here (I also noticed that integration tests being annoying to maintain in the java-contrib repo for jmx-metrics and had the same thought of not really needing to verify SDKData -> OTLP conversion there)

breedx-splk:
I think I'm onboard with this, and agree that the collector is really just an additional piece of unnecessary complexity here for these tests. I hadn't ever looked at how the collector was sending data to the fake backend -- since it is indeed otlp I think it makes perfect sense to just leave it out and ingest right to the fake backend.

I will close this and I think we can also close #4136 then and create a new issue to omit the collector from the smoke tests entirely.  👍🏻

trask:
> Does @trask (author of #4136) have input?

this sounds great. and it's breaking the build again today with new release of collector a few hours ago, so I will work on this right away (I _think_ it should be pretty straight forward... famous last words...)

trask:
Closing in favor of #4186

PR #4177, opened 2021-09-21
Limit latest dep version for kafka clients

laurit:


iNikem (approved):

mateuszrzeszutek (changes requested):
I just pushed https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4178 that removes `EmbeddedKafka` usage and all those latest dep limits

PR #4179, opened 2021-09-21
Include gradle-plugins as a composite build

iNikem:
Two major changes in PR.

First, `buildSrc` was renamed to `conventions` and included into the main build as a composite.

This allows for the second change: include `gradle-plugins` into `conventions` as a composite build. This way we can substitute maven dependencies on gradle plugins artifacts with local project dependency and always use the latest version of it.


iNikem:
Superseded by #4183

PR #4183, opened 2021-09-21
Remove muzzle from gradleplugins

iNikem:
Several changes in this PR.

First, `buildSrc` was renamed to `conventions` and included into the main build as a composite.

This allows for the second change: include `gradle-plugins` into `conventions` as a composite build. This way we can substitute maven dependencies on gradle plugins artifacts with local project dependency and always use the latest version of it.

Third, all classes that actually depend on our main project were removed from `gradle-plugins` module and moved to `muzzle`. This completely breaks the cycle between main project and plugins and allows us always use the latest version of our code without intermediate publishing. The only interface between them is now a _name_ of the class `io.opentelemetry.javaagent.tooling.muzzle.generation.MuzzleCodeGenerationPlugin`


conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -10,6 +10,16 @@ plugins {
   id("io.opentelemetry.instrumentation.javaagent-shadowing")
 }
 
+dependencies {
+  /*
+    Dependencies added to this configuration will be found by the muzzle gradle plugin during code
+    generation phase. These classes become part of the code that plugin inspects and traverses during
+    references collection phase.
+   */
+  add("codegen", "io.opentelemetry.javaagent:opentelemetry-javaagent-tooling")
```
mateuszrzeszutek, line 19: Just wondering: we already have this in the javaagent-instrumentation plugin, why do we need to duplicate it here?

conventions/settings.gradle.kts
```diff
@@ -0,0 +1,6 @@
+includeBuild("../gradle-plugins") {
+  dependencySubstitution {
+    substitute(module("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin")).using(project(":"))
+    substitute(module("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin")).using(project(":"))
```
mateuszrzeszutek, line 4: That's neat 💯

mateuszrzeszutek (approved):

conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
anuraaga: IIRC, which I might not :P, I think we said that we don't want this artifact on the muzzle check's transformation classpath. Is that not an issue anymore, or is that not what happens with this change?

anuraaga (approved):

conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
iNikem: "This" - you mean "tooling"? `io.opentelemetry.instrumentation.javaagent-instrumentation.gradle.kts` always included tooling to `codegen` conf. Or do I miss something?

anuraaga: Oh ok - previously for javaagent-testing.gradle.kts it was picking it up through the build.gradle dependencies but now both of them use the configuratio? Cool

iNikem: What I don't fully understand, is if we need muzzle-generation plugin at all in `io.opentelemetry.instrumentation.javaagent-testing.gradle`.

anuraaga: At least one of our test projects includes a test instrumentation so it doesn't work without the codegen.

trask (approved):
thx ❤️ I will cherish the memory of the one and only time time I did the gradle-plugins publishing dance 😂

settings.gradle.kts
```diff
@@ -1,4 +1,12 @@
 pluginManagement {
+  repositories {
+    mavenCentral()
+    gradlePluginPortal()
+    maven {
+      url = uri("https://oss.sonatype.org/content/repositories/snapshots")
+    }
```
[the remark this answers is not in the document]
iNikem: Hm, probably it is not anymore... I will try to remove

PR #4187, opened 2021-09-22
Experimental option to suppress messaging receive spans

trask:
Introducing this option was discussed in last week's Messaging SIG. Experience from Node is that users think messaging instrumentation is broken when the producer and consumer spans don't have the same traceId.

EDIT: the current modeling is correct per the current messaging semantic conventions, and you should be seeing links from consumers to producers, but links are not modeled great in all backends yet, which is where this option comes in handy (as well as if you just want a simpler trace structure)

### DIFF
--- a/instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/KafkaSingletons.java
+++ b/instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/KafkaSingletons.java
@@ -72,16 +72,14 @@ public final class KafkaSingletons {
       builder.addAttributesExtractor(new KafkaConsumerExperimentalAttributesExtractor());
     }
 
-    if (KafkaPropagation.isPropagationEnabled()) {
-      if (ExperimentalConfig.get().suppressMessagingReceiveSpans()) {
-        return builder.newConsumerInstrumenter(new KafkaHeadersGetter());
-      } else {
-        builder.addSpanLinksExtractor(
-            SpanLinksExtractor.fromUpstreamRequest(
-                GlobalOpenTelemetry.getPropagators(), new KafkaHeadersGetter()));
-        return builder.newInstrumenter(SpanKindExtractor.alwaysConsumer());
-      }
+    if (!KafkaPropagation.isPropagationEnabled()) {
+      return builder.newInstrumenter(SpanKindExtractor.alwaysConsumer());
+    } else if (ExperimentalConfig.get().suppressMessagingReceiveSpans()) {
+      return builder.newConsumerInstrumenter(new KafkaHeadersGetter());
     } else {
+      builder.addSpanLinksExtractor(
+          SpanLinksExtractor.fromUpstreamRequest(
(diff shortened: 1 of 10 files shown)


instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/KafkaSingletons.java
```diff
@@ -52,6 +53,7 @@
             GlobalOpenTelemetry.get(), INSTRUMENTATION_NAME, spanNameExtractor)
         .addAttributesExtractor(attributesExtractor)
         .setTimeExtractors(ReceivedRecords::startTime, (request, response, error) -> request.now())
+        .setDisabled(ExperimentalConfig.get().suppressMessagingReceiveSpans())
```
mateuszrzeszutek, line 56: Nice 👍

instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/KafkaSingletons.java
```diff
@@ -69,12 +71,19 @@
     if (KafkaConsumerExperimentalAttributesExtractor.isEnabled()) {
       builder.addAttributesExtractor(new KafkaConsumerExperimentalAttributesExtractor());
     }
+
     if (KafkaPropagation.isPropagationEnabled()) {
```
mateuszrzeszutek, line 75: WDYT about inverting this condition and removing one level of nesting?
```java
if (!KafkaPropagation.isPropagationEnabled()) {
  // ...
} else if (ExperimentalConfig.get().suppressMessagingReceiveSpans()) {
  // ...
} else {
  // ...
}
```

instrumentation/kafka-streams-0.11/javaagent/src/test/groovy/KafkaStreamsSuppressReceiveSpansTest.groovy
```diff
@@ -0,0 +1,248 @@
...
+
+  @Shared
+  static KafkaContainer kafka
+  @Shared
+  static Producer<Integer, String> producer
+  @Shared
+  static Consumer<Integer, String> consumer
+
+
+  def setupSpec() {
```
mateuszrzeszutek, line 53: How about extracting a base class that'll contain the whole kafka setup?

instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/test/groovy/KafkaClientSuppressReceiveSpansTest.groovy
```diff
@@ -0,0 +1,193 @@
...
+        spans.sort({
+          // job span is first
+          if (it.name == "BatchJob parallelItemsJob") {
+            return 0
+          }
+          // step span is second
+          if (it.name == "BatchJob parallelItemsJob.parallelItemsStep") {
+            return 1
+          }
+        })
```
mateuszrzeszutek, line 53: To be removed? 😄

mateuszrzeszutek (approved):
👍

PR #4189, opened 2021-09-22
Fix kafka tests failing when consumer is not yet subscribed to topic

mateuszrzeszutek:
Thanks @alesj for finding this issue


trask (approved):
👍

PR #4195, opened 2021-09-23
Separate HTTP client/server AttributesExtractors

mateuszrzeszutek:
This PR splits `HttpAttributesExtractor` into `HttpClientAttributesExtractor` and `HttpServerAttributesExtractor` - I need to do that before I implement the [http headers semantic attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers) that were recently added to the spec (because we'll want separate configurations for clients/servers).

And maybe, if we decide on a solution here we'll be able to get https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3700 done

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
@@ -8,7 +8,7 @@ package io.opentelemetry.instrumentation.api.instrumenter;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.instrumentation.api.instrumenter.db.DbAttributesExtractor;
-import io.opentelemetry.instrumentation.api.instrumenter.http.HttpAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpClientAttributesExtractor;
 import io.opentelemetry.instrumentation.api.instrumenter.net.NetAttributesExtractor;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
@@ -21,7 +21,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  * #onStart(AttributesBuilder, Object)} to have it available during sampling.
  *
  * @see DbAttributesExtractor
- * @see HttpAttributesExtractor
+ * @see HttpClientAttributesExtractor
  * @see NetAttributesExtractor
  */
 public abstract class AttributesExtractor<REQUEST, RESPONSE> {
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractor.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.http;
+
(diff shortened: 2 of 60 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
```diff
@@ -21,7 +21,7 @@
  * #onStart(AttributesBuilder, Object)} to have it available during sampling.
  *
  * @see DbAttributesExtractor
- * @see HttpAttributesExtractor
+ * @see HttpClientAttributesExtractor
```
trask, line 24:
```suggestion
 * @see HttpClientAttributesExtractor
 * @see HttpServerAttributesExtractor
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+  protected final void onStart(AttributesBuilder attributes, REQUEST request) {
+    super.onStart(attributes, request);
+
+    set(attributes, SemanticAttributes.HTTP_SCHEME, scheme(request));
+    set(attributes, SemanticAttributes.HTTP_HOST, host(request));
+    set(attributes, SemanticAttributes.HTTP_TARGET, target(request));
+    set(attributes, SemanticAttributes.HTTP_ROUTE, route(request));
+
+    // TODO: this is specific to clients, should we remove this?
+    set(attributes, SemanticAttributes.HTTP_URL, url(request));
```
trask, line 36: same, +1 for removing this and we can always add it later if there's really a compelling reason.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanNameExtractor.java
```diff
@@ -42,4 +43,12 @@ public String extract(REQUEST request) {
...
     return "HTTP request";
   }
+
+  @Nullable
+  private String extractRoute(REQUEST request) {
+    if (attributesExtractor instanceof HttpServerAttributesExtractor) {
+      return ((HttpServerAttributesExtractor<REQUEST, ?>) attributesExtractor).route(request);
+    }
+    return null;
+  }
```
trask, line 53: is it worth `HttpClientSpanNameExtractor` and `HttpServerSpanNameExtractor`?

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanNameExtractor.java
mateuszrzeszutek: Hmm, maybe we could have different factory methods in `HttpSpanNameExtractor`? One accepting the client extractor, the other the server one

anuraaga: `instanceof` seems OK too from what I can tell

anuraaga (approved):
Thanks for doing this!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractor.java
iNikem: I am not sure that we can remove this. I arbitrary took a couple of server instrumentation's extractor, and they all returned `null` for attributes above

trask:
👍 this is great, going to merge, TODOs make sense (to try) as follow-up(s)

PR #4197, opened 2021-09-23
Build servlet mapping resolver only when needed

laurit:
While testing on weblogic 12.1.3 I noticed that building mapping resolver for some weblogic internal servlets and filters fails with `UnsupportedOperationException` that appears in our debug log and
`<Sep 23, 2021 11:55:28 AM GMT> <Error> <HTTP> <BEA-101388> <The ServletContext was passed to the ServletContextListener.contextInitialized method of a ServletContextListener that was neither declared in web.xml or web-fragment.xml, nor annotated with javax.servlet.annotation.WebListener.>`
gets logged by the server. The same did not happen on weblogic 14.1.1.0.
It seems like weblogic 12 does not like it when we start looking for servlet and filter registrations when it is starting the webapp, it should work when we do it a bit later when the first request comes in.


instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3InitAdvice.java
```diff
@@ -20,7 +20,7 @@ public static void servletInit(
     if (servletConfig == null) {
       return;
     }
-    InstrumentationContext.get(Servlet.class, MappingResolver.class)
+    InstrumentationContext.get(Servlet.class, MappingResolverFactory.class)
         .putIfAbsent(servlet, new Servlet3MappingResolverFactory(servletConfig));
```
trask, line 24: 😅 it took me a long time to understand how this PR changes anything in the init sequence, until I realized this line (which has no code changes) used to call `putIfAbsent(key, Supplier)` which immediately triggered `Servlet3MappingResolverFactory.getMappings()` via the super classes implementation of `Supplier.get()`, but now calls `putIfAbsent(key, non-supplier-value)` which does not

trask (approved):
looks like there's a shading issue:

```
java.lang.ClassCastException: io.opentelemetry.javaagent.instrumentation.servlet.v3_0.Servlet3MappingResolverFactory incompatible with io.opentelemetry.javaagent.shaded.instrumentation.servlet.naming.MappingResolverFactory
```

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Singletons.java
```diff
@@ -39,6 +40,11 @@
   private static final ServletHelper<HttpServletRequest, HttpServletResponse> HELPER =
       new ServletHelper<>(INSTRUMENTER, Servlet3Accessor.INSTANCE);
 
+  private static final ContextStore<Servlet, MappingResolver.Factory> SERVLET_CONTEXT_STORE =
+      InstrumentationContext.get(Servlet.class, MappingResolver.Factory.class);
```
trask, line 44: 👍

PR #4203, opened 2021-09-26
Remove problem files from META-INF

trask:
These are only creating a problem because while building native image I'm writing out the `inst` directory to a real jar file.

* `INDEX.LIST` is from shadow'd dependency and subverts normal class lookup (I wonder if generating this correctly could speed up class loading?)
* `*.DSA` and `*.SF` are signature files from shadow'd dependencies


anuraaga (approved):

iNikem:
https://imperceptiblethoughts.com/shadow/getting-started/#default-java-groovy-tasks claims that they are already excluded by default

anuraaga:
That seems to talk about a task named `shadowJar`, but we have many shadow tasks with a different name.

iNikem (approved):

PR #4205, opened 2021-09-27
Convert liberty instrumentation to instrumenter api

laurit:
Part of #2713


instrumentation/liberty/liberty/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/ThreadLocalContext.java
```diff
@@ -45,6 +48,10 @@ public HttpServletRequest getRequest() {
     return request;
```
trask, line 48: and then can get rid of the `request` field
```suggestion
    return requestContext.request();
```

trask (approved):

mateuszrzeszutek (approved):

PR #4206, opened 2021-09-27
Refine 404 handling in Restlet instrumentation

Enkelian:
A small fix to the Restlet server instrumentation - if response code is 404, span name will be `/*`

### DIFF
--- a/instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/ServerInstrumentation.java
+++ b/instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/ServerInstrumentation.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.javaagent.instrumentation.restlet.v1_0;
 
+import static io.opentelemetry.instrumentation.api.servlet.ServerSpanNaming.Source.CONTROLLER;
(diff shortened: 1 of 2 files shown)


instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/ServerInstrumentation.java
```diff
@@ -71,6 +75,12 @@ public static void finishRequest(
 
       scope.close();
 
+      if (response.getStatus() != null
+          && response.getStatus().equals(Status.CLIENT_ERROR_NOT_FOUND)) {
```
kubawach, line 79: NIT: this can be easily compacted to Status.CLIENT_ERROR_NOT_FOUND.equals(response.getStatus()) 
In general comparing constant - to - value allows to skip a nullcheck :)

kubawach (approved):
LGTM

instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/ServerInstrumentation.java
Enkelian: Certainly more elegant, thanks!

trask (approved):
👍

PR #4207, opened 2021-09-27
Introduce muzzle-specific interface to InstrumentationModule

iNikem:
It will contain all methods that muzzle automatically generates.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/instrumentation/InstrumentationModuleInstaller.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/instrumentation/InstrumentationModuleInstaller.java
@@ -23,6 +23,7 @@ import io.opentelemetry.javaagent.tooling.context.NoopContextProvider;
 import io.opentelemetry.javaagent.tooling.muzzle.ContextStoreMappings;
 import io.opentelemetry.javaagent.tooling.muzzle.HelperResourceBuilderImpl;
 import io.opentelemetry.javaagent.tooling.muzzle.InstrumentationContextBuilderImpl;
+import io.opentelemetry.javaagent.tooling.muzzle.InstrumentationModuleMuzzle;
 import io.opentelemetry.javaagent.tooling.muzzle.Mismatch;
 import io.opentelemetry.javaagent.tooling.muzzle.ReferenceMatcher;
 import java.lang.instrument.Instrumentation;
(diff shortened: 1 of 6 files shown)


muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
```diff
@@ -45,6 +45,11 @@ protected MethodHandle computeValue(Class<?> type) {
    */
   @SuppressWarnings("unchecked")
   static Map<String, ClassRef> getFor(InstrumentationModule instrumentationModule) {
+    if (instrumentationModule instanceof InstrumentationModuleMuzzle) {
+      return ((InstrumentationModuleMuzzle) instrumentationModule).getMuzzleReferences();
+    }
+
+    //Older classes created and compiled outside of this repo may not yet have the interface above.
```
iNikem, line 52: Not sure if we should be backward-compatible here...

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/instrumentation/InstrumentationModuleInstaller.java
```diff
@@ -120,20 +121,23 @@ AgentBuilder install(
...
       InstrumentationModule instrumentationModule) {
-    InstrumentationContextBuilderImpl builder = new InstrumentationContextBuilderImpl();
-    instrumentationModule.registerMuzzleContextStoreClasses(builder);
-    ContextStoreMappings mappings = builder.build();
-    if (!mappings.isEmpty()) {
-      return FieldBackedProviderFactory.get(instrumentationModule.getClass(), mappings);
-    } else {
-      return NoopContextProvider.INSTANCE;
+
+    if (instrumentationModule instanceof InstrumentationModuleMuzzle) {
```
trask, line 125: may be worth a `debug` log in an `else`?

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/MuzzleReferencesAccessor.java
```diff
@@ -45,6 +45,11 @@ protected MethodHandle computeValue(Class<?> type) {
    */
   @SuppressWarnings("unchecked")
   static Map<String, ClassRef> getFor(InstrumentationModule instrumentationModule) {
+    if (instrumentationModule instanceof InstrumentationModuleMuzzle) {
+      return ((InstrumentationModuleMuzzle) instrumentationModule).getMuzzleReferences();
+    }
+
+    // Older classes created and compiled outside of this repo may not yet have the interface above.
```
trask, line 52: how about log at `warn` level below in `methodHandle != null`, saying that this path is deprecated and will be removed in the near future?

trask (approved):

mateuszrzeszutek (approved):
Nice 💯

PR #4209, opened 2021-09-27
Remove url from HttpServerAttributesExtractor

### LINKED ISSUE #3699: Remove unnecessary URL construction

### LINKED ISSUE #3700: Don't capture http.scheme, http.host and http.target if already capturing http.url

trask:
follow-up to #4195

resolves #3700
resolves #3699


instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackHttpAttributesExtractor.java
```diff
@@ -63,7 +52,15 @@ protected String route(Request request) {
...
-    return null;
+    Context ratpackContext = request.get(Context.class);
+    if (ratpackContext == null) {
+      return null;
+    }
+    PublicAddress publicAddress = ratpackContext.get(PublicAddress.class);
+    if (publicAddress == null) {
+      return null;
+    }
+    return publicAddress.builder().build().getScheme();
```
mateuszrzeszutek, line 63: Can we use 
```suggestion
    return publicAddress.get().getScheme();
```
instead of building another copy?

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractorTest.java
```diff
@@ -120,7 +115,6 @@ void normal() {
     assertThat(attributes.build())
         .containsOnly(
             entry(SemanticAttributes.HTTP_METHOD, "POST"),
-            entry(SemanticAttributes.HTTP_URL, "http://github.com"),
             entry(SemanticAttributes.HTTP_TARGET, "github.com"),
```
anuraaga, line 118: I suspect I fudged - this test case should probably be changed to something that looks more normal, e.g. `/repositories/1`

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackHttpAttributesExtractor.java
```diff
@@ -33,24 +38,8 @@ protected String url(Request request) {
...
-  protected String target(Request request) {
-    // Uri is the path + query string, not a full URL
-    return request.getUri();
-  }
-
-  @Override
-  @Nullable
-  protected String host(Request request) {
-    return null;
+    URI uri = publicAddress.get();
```
anuraaga, line 41: Private method to get the public address (the simplification I was hoping for with my TODO didn't materialize...

anuraaga (approved):

docs/semantic-conventions.md
```diff
@@ -23,12 +23,17 @@ are implemented by Java autoinstrumentation and which ones are not.
 | `http.route` | N | - |
 | `http.client_ip` | N | + |
 
-**[1]:** As the majority of Java frameworks don't provide a standard way to obtain "The full request
+**[1]:** Most server instrumentation captures `http.scheme`, `http.host`, and `http.target`
+(and does not capture `http.url`).
+Netty instrumentation is currently the only exception to this rule. Netty instrumentation
+captures `http.url` (and does not capture `http.scheme`, `http.host`, or `http.target`).
+
+As the majority of Java frameworks don't provide a standard way to obtain "The full request
```
iNikem, line 31: This paragraph and the previous one are in direct contradiction to one another

docs/semantic-conventions.md
```diff
@@ -23,12 +23,17 @@ are implemented by Java autoinstrumentation and which ones are not.
 | `http.route` | N | - |
 | `http.client_ip` | N | + |
 
-**[1]:** As the majority of Java frameworks don't provide a standard way to obtain "The full request
+**[1]:** Most server instrumentation captures `http.scheme`, `http.host`, and `http.target`
+(and does not capture `http.url`).
```
mateuszrzeszutek, line 27:
```suggestion
**[1]:** Most server instrumentations capture `http.scheme`, `http.host`, and `http.target`
(and do not capture `http.url`).
```

docs/semantic-conventions.md
mateuszrzeszutek: WDYT about rewriting that paragraph (splitting in two?) so that it describes servers and clients separately?

trask: > This paragraph and the previous one are in direct contradiction to one another

oops I should have deleted this paragraph, it's gone now

> WDYT about rewriting that paragraph (splitting in two?) so that it describes servers and clients separately?

http client is split out below

docs/semantic-conventions.md
trask: I went back and forth on this 😂

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractorTest.java
trask: fixed 😄

mateuszrzeszutek (approved):

iNikem (approved):

PR #4214, opened 2021-09-28
Sync linux and windows smoke test images

trask:
I hadn't realized this before, but the linux and windows smoke test images need to be in sync, since the tests apply equally to both.

### DIFF
--- a/smoke-tests/matrix/build.gradle
+++ b/smoke-tests/matrix/build.gradle
@@ -80,8 +80,7 @@ def windowsTargets = [
     [version: ["7.0.107"], vm: ["hotspot", "openj9"], jdk: ["8"], args: [majorVersion: "7"]],
     [version: ["8.5.60"], vm: ["hotspot", "openj9"], jdk: ["8", "11"], args: [majorVersion: "8"]],
     [version: ["9.0.40"], vm: ["hotspot", "openj9"], jdk: ["8", "11"], args: [majorVersion: "9"]],
-    [version: ["10.0.4"], vm: ["hotspot"], jdk: ["11", "16", "17"], args: [majorVersion: "10"], war: "servlet-5.0"],
-    [version: ["10.0.4"], vm: ["openj9"], jdk: ["11", "16"], args: [majorVersion: "10"], war: "servlet-5.0"]
+    [version: ["10.0.4"], vm: ["hotspot", "openj9"], jdk: ["11", "15"], args: [majorVersion: "10"], war: "servlet-5.0"],
   ],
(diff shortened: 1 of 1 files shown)


iNikem:
What do you mean, "in sync"? Do they need to have the same JVM version? Why?

trask:
> What do you mean, "in sync"? Do they need to have the same JVM version? Why?

there's only a single list of app server versions / jdk versions that the tests use when running against both linux and windows, e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6aaf5ef9e694f1466714ebcddd37408a1559e663/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/JettySmokeTest.groovy#L8-L22

mateuszrzeszutek (approved):

smoke-tests/matrix/build.gradle
```diff
@@ -80,8 +80,7 @@ def windowsTargets = [
     [version: ["7.0.107"], vm: ["hotspot", "openj9"], jdk: ["8"], args: [majorVersion: "7"]],
     [version: ["8.5.60"], vm: ["hotspot", "openj9"], jdk: ["8", "11"], args: [majorVersion: "8"]],
     [version: ["9.0.40"], vm: ["hotspot", "openj9"], jdk: ["8", "11"], args: [majorVersion: "9"]],
-    [version: ["10.0.4"], vm: ["hotspot"], jdk: ["11", "16", "17"], args: [majorVersion: "10"], war: "servlet-5.0"],
-    [version: ["10.0.4"], vm: ["openj9"], jdk: ["11", "16"], args: [majorVersion: "10"], war: "servlet-5.0"]
+    [version: ["10.0.4"], vm: ["hotspot", "openj9"], jdk: ["11", "15"], args: [majorVersion: "10"], war: "servlet-5.0"],
```
iNikem, line 83: Is it correct that we downgrade 16/17 to 15 here?

mateuszrzeszutek: The `TomcatSmokeTest` class doesn't run 16/17 tests anyway: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/TomcatSmokeTest.groovy

(We should upgrade to 17 though; preferably both for linux and windows)

trask: > We should upgrade to 17 though; preferably both for linux and windows

👍 I'll do this in follow-up PR, this PR just gets us back to steady state.

Issue #3156, opened 2021-06-01
Rename FooTracing to FooInstrumentation or FooInstrumenter


anuraaga:
Now that it's not just tracing we need to rename our library instrumentations. Which name is better? I lean towards `FooInstrumentation` since users don't interact with `Instrumenter`.

anuraaga:
It was sort of nice how we didn't have overlap with naming in the javaagent modules though, where we use `FooInstrumentation` as a `TypeInstrumentation`

mateuszrzeszutek:
> Will be effectively implemented by #2713

Hm, I don't think it will - this issue is about deciding whether to rename the library instrumentation classes e.g. `OkHttpTracing` to `OkHttpInstrumentation`/`OkHttpTelemetry`. It's not directly connected to #2713 (we'd have to rename already existing library classes too)

PR #4215, opened 2021-09-28
Add a TracingBuilder for library instrumentation to extend from for c…

anuraaga:
…ommon configuration.

We'd like library instrumentation to expose a consistent set of configuration to users, and this makes sure that is automatic.

Sending for an initial look to make sure it generally looks like a sane approach.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
@@ -42,7 +42,7 @@ public final class InstrumenterBuilder<REQUEST, RESPONSE> {
   final OpenTelemetry openTelemetry;
   final Meter meter;
   final String instrumentationName;
-  final SpanNameExtractor<? super REQUEST> spanNameExtractor;
+  SpanNameExtractor<? super REQUEST> spanNameExtractor;
 
   final List<SpanLinksExtractor<? super REQUEST>> spanLinksExtractors = new ArrayList<>();
   final List<AttributesExtractor<? super REQUEST, ? super RESPONSE>> attributesExtractors =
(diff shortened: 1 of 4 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -199,6 +199,17 @@
         InstrumenterConstructor.propagatingToDownstream(setter), SpanKindExtractor.alwaysClient());
   }
 
+  /**
+   * Returns a new {@link Instrumenter} which will create client spans and inject context into
+   * requests.
+   */
+  public Instrumenter<REQUEST, RESPONSE> newClientInstrumenter(
+      TracingBuilder<REQUEST, RESPONSE, ?> tracingBuilder, TextMapSetter<REQUEST> setter) {
```
anuraaga, line 207: The idea is to remove the methods that don't accept a `TracingBuilder`

mateuszrzeszutek: Should we really do that? Requiring a `TracingBuilder` will make it harder to use `Instrumenter` ad-hoc - for example in our own javaagent instrumentations (or in libs like spring sleuth, probably).

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/TracingBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+
+package io.opentelemetry.instrumentation.api.instrumenter;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+
+public abstract class TracingBuilder<
```
mateuszrzeszutek, line 14: This is probably a good opportunity to revisit https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3156 - should it really be `TracingBuilder`? Maybe `TelemetryBuilder`? (and `FooTelemetryBuilder`, `FooTelemetry`)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
anuraaga: Good point. I struggled to come up with a pattern for applying the settings to the Instrumenter that isn't easy to forget to do, e.g. a protected `apply` method.

Any idea come to mind?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/TracingBuilder.java
anuraaga: Yeah kept this consistent with what we have for now but realized we should rename to something like FooInstrumentation

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
mateuszrzeszutek: I mean, adding new `newInstrumenter()` methods that accept a `TracingBuilder` seems fine, it's the removing idea that might be problematic; I don't really have better ideas that this. `copySettingsFrom(TracingBuilder)`?

trask:
@mateuszrzeszutek @anuraaga is there anything from this PR that we should consider as part of stabilizing the instrumenter api?

breedx-splk:
Closing a few old/stale draft PRs. Reopen if interested.

Issue #4081, opened 2021-09-10
Simpler solution for library instrumentation to benefit from InstrumentationContext when used in agent


trask:
If we had an instrumentation-api something like:

```
public AssociatingCache {
  public static <Q extends K, K, C> Cache<Q, C> get(Class<K> keyClass, Class<C> contextClass) {
    return Cache.newBuilder().setWeakKeys().build();
  }
}
```

then the agent could replace the implementation of that method at runtime with

```
InstrumentationContext.get(keyClass, contextClass).asCache()
```

and I think that would make it easier to migrate things to library instrumentation while not losing the benefit of InstrumentationContext when used in agent.

(see [JdbcData.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/4c64f70f94e5e5f4e80e3ff263ef7a477962d460/instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/JdbcData.java#L26-L39) for the current solution to using InstrumentationContext from library instrumentation when it is used in agent)

mateuszrzeszutek:
We'd probably have to do sth like
```
if (contextStoreSupplier == null) {
  return Cache.newBuilder().setWeakKeys().build();
}
return contextStoreSupplier.get(keyClass, contextClass);
```
With the `contextStoreSupplier` being set by the agent.

Issue #441, opened 2020-05-26
Memory leak


romil-punetha:
There's probably a memory leak in the instrumentation jar. Two of the services required a revert, one of them as soon as 10 mins. Im attaching screenshots to show the performance after the code was deployed with instrumentation. Both these services were deployed at around 10. 

<img width="1277" alt="Screenshot 2020-05-26 at 7 16 41 PM" src="https://user-images.githubusercontent.com/47735999/82910068-c4d64700-9f87-11ea-8ca3-417eec7d8873.png">
 

<img width="1279" alt="Screenshot 2020-05-26 at 7 16 54 PM" src="https://user-images.githubusercontent.com/47735999/82909848-71fc8f80-9f87-11ea-839e-b52af7c97c61.png">

<img width="747" alt="Screenshot 2020-05-26 at 7 51 40 PM" src="https://user-images.githubusercontent.com/47735999/82912163-5fd02080-9f8a-11ea-88c8-35095e962489.png">

Some of the common libraries are 
```
org.jetbrains.kotlin:kotlin-stdlib-jdk8:jar:1.3.61
com.fasterxml.jackson.module:jackson-module-kotlin:jar:2.9.10
com.amazonaws:aws-java-sdk-s3:jar:1.11.676 and other aws ki lib
org.springframework.boot:spring-boot-starter-actuator:jar:2.1.12
de.codecentric:spring-boot-admin-starter-client
de.codecentric:spring-boot-admin-starter-server:jar:2.1.1
org.springframework.boot:spring-boot-starter-undertow:jar:2.1.12.RELEASE
org.springframework.boot:spring-boot-starter-quartz:jar:2.1.12.RELEASE
org.springframework.boot:spring-boot-starter-data-redis:jar:2.1.12.RELEASE
org.springframework.boot:spring-boot-starter-data-jpa:jar:2.1.12.RELEASE
org.springframework.cloud:spring-cloud-starter-netflix-hystrix:jar:2.1.0.M3
org.springframework.boot:spring-boot-starter-web:jar:2.1.12.RELEASE
```

trask:
Hi @romil-punetha!

The image you shared in gitter caught my interest:

![image](https://user-images.githubusercontent.com/218610/82977215-0cfc6600-9f96-11ea-9da8-c303a60ae495.png)

[...]

PR #4218, opened 2021-09-28
Make it possible to use InstrumentationContext (now VirtualField) fro…

mateuszrzeszutek:
…m library instrumentation

Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4081

* Renamed `ContextStore` to `VirtualField` in public APIs
* Moved & renamed `InstrumentationContext#get()` to `VirtualField#find()`
* Moved `VirtualField` to `intstrumentation-api`

There's still several mentions of `ContextStore`/`InstrumentationContext` remaining, but they're all located somewhere in the agent internals (or muzzle). I'll rename them in a follow-up PR (once muzzle `InstrumenterConfigBuilder` PR gets merged)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /** Sets the new value of this virtual field. */
+  public abstract void set(T object, @Nullable F fieldValue);
+
+  /**
+   * Sets the new value of this virtual field if the current value is {@code null} or absent.
+   *
+   * @return The old field value if it was present, or the passed {@code fieldValue}.
+   */
+  public abstract F setIfAbsentAndGet(T object, F fieldValue);
```
laurit, line 55: Perhaps call it just `setIfAbsent`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /** Gets the value of this virtual field. */
+  @Nullable
+  public abstract F get(T object);
+
+  /** Sets the new value of this virtual field. */
+  public abstract void set(T object, @Nullable F fieldValue);
+
+  /**
+   * Sets the new value of this virtual field if the current value is {@code null} or absent.
```
laurit, line 51: I would remove the or absent part as we really only check whether the field is `null`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
```diff
@@ -0,0 +1,64 @@
...
+package io.opentelemetry.instrumentation.api.field;
+
+import io.opentelemetry.instrumentation.api.internal.RuntimeVirtualFieldSupplier;
+import java.util.function.Supplier;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Represents a "virtual" field of type {@code F} that is added to type {@code T} in the runtime.
+ *
+ * <p>Field values are weakly referenced and will be garbage collected when their owner instance is
```
laurit, line 15: My understanding is that field values are not weakly referenced, if they were they could be collected if nobody else has a reference to them.
If I understand this correctly the happy path is that we transform the class and add a field to it so we really have a regular strong reference to the value. Things get messy when class is not transformed for some reason. Then we put  the values into a weak map where key is the object on which field is set. This has the annoying property that when value references key then we immediately have a memory leak because now key can be never collected. We could get around this by also making values weak which would mean that something else needs to keep the values alive. Basically afaik no matter what we do we can't get the same behaviour as when the value is just stored in a field.
Perhaps we could address this somehow. Maybe we could drop the support for using fallback map? Now that lambda classes are also transformed is there a case where it could be needed? There is a flag to disable field injection and there is hack that skips filed injection on classes annotated with `javax.decorator.Decorator` so might not be that easy to completely get rid of it. An alternative would be to have something in `VirtualField` api that lets user figure out whether it is field or map based. Or could add methods like `setWeak` that doesn't change anything in field backed implementation but for map based implementation makes the value also weak. This would allow using `VirtualField` in netty `FutureListenerWrappers` without leaking memory even when map based store is used.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
mateuszrzeszutek: Good point.
Now I'm starting to wonder whether the method should be named `setIfNullAndGet`...

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
mateuszrzeszutek: > My understanding is that field values are not weakly referenced, if they were they could be collected if nobody else has a reference to them.

Yeah that's true, I copied this Javadoc and didn't really think too deeply about it.

> Basically afaik no matter what we do we can't get the same behaviour as when the value is just stored in a field.
> Perhaps we could address this somehow. Maybe we could drop the support for using fallback map?

We probably have to keep the fallback implementation; I suppose it's better to have it in place and install a few more instrumentations than not to have it and skip instrumenting some pieces of code.

Perhaps we could fix the Javadoc so that it outright says that this has similar semantics to a weak-keys strong-values map and keeping values that might reference the original key is discouraged?

javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/tooling/muzzle/InstrumentationClassPredicateTest.groovy
```diff
@@ -37,6 +37,6 @@ class InstrumentationClassPredicateTest extends Specification {
     "Java SDK class"                      | "java.util.ArrayList"
     "javaagent-tooling class"             | "io.opentelemetry.javaagent.tooling.Constants"
     "instrumentation-api class"           | "io.opentelemetry.instrumentation.api.InstrumentationVersion"
-    "javaagent-instrumentation-api class" | "io.opentelemetry.javaagent.instrumentation.api.ContextStore"
+    "javaagent-instrumentation-api class" | "io.opentelemetry.instrumentation.api.field.ContextStore"
```
trask, line 40: is this right?

instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/JdbcData.java
```diff
@@ -5,55 +5,17 @@
...
 
-  private static final CacheFactory cacheFactory = getCacheFactory();
-
-  public static Cache<Connection, DbInfo> connectionInfo = cacheFactory.connectionInfoCache();
-  public static Cache<PreparedStatement, String> preparedStatement =
-      cacheFactory.preparedStatementCache();
+  public static VirtualField<Connection, DbInfo> connectionInfo =
+      VirtualField.find(Connection.class, DbInfo.class);
+  public static VirtualField<PreparedStatement, String> preparedStatement =
+      VirtualField.find(PreparedStatement.class, String.class);
```
trask, line 18: 💯

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
trask: > Maybe we could drop the support for using fallback map?

that would be nice, the memory leak has bit us before (at least #441 and #3027 I think)

> Now that lambda classes are also transformed is there a case where it could be needed? There is a flag to disable field injection and there is hack that skips filed injection on classes annotated with javax.decorator.Decorator so might not be that easy to completely get rid of it

I wonder if the `javax.decorator.Decorator` is needed since #3948

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
trask: I like `setIfNull`

trask (approved):

javaagent-tooling/src/test/groovy/io/opentelemetry/javaagent/tooling/muzzle/InstrumentationClassPredicateTest.groovy
mateuszrzeszutek: Nope - nice catch, thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
mateuszrzeszutek: 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
mateuszrzeszutek: This is the original DD PR that introduced this mathcer: https://github.com/DataDog/dd-trace-java/pull/1045

> In CDI (Contexts and Dependency Injection) environments, classes annotated with `@Decorator` wrap other classes.
Decorators implement the same interfaces but delegate to another class. See the Oracle docs.
> 
> In general, the tracer should instrument the delegate (decorated class) and not the decorator. When the tracer instruments the decorator, it cause the CDI container to be confused about the interfaces of the decorator and crash on startup.

If we can reproduce that problem and verify that #3948 solves it then we can start thinking of phasing out the weak map implementation.

trask: opened #4241 to track/discuss further

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/field/VirtualField.java
trask: i was thinking `setIfNull` without the `AndGet`, similar to `Map.putIfAbsent`, but now i realize the potential confusion because `putIfAbsent` returns the previous value or `null` if there was none, while this method always returns the stored value.

maybe:

`void setIfNull(T object, F fieldValue)`

and

`F computeIfNull(T object, Supplier<F> fieldValueSupplier)`

as it looks like we don't (currently at least) use the return value from the first method, though now I'm wondering if that method is even needed, i.e. will anything break if those callers just use `set` (and if the "set if null" behavior really does matter, they could call `get` first, as there's nothing atomic anyways if they aren't using the return value)

mateuszrzeszutek: > though now I'm wondering if that method is even needed, i.e. will anything break if those callers just use `set` (and if the "set if null" behavior really does matter, they could call `get` first, as there's nothing atomic anyways if they aren't using the return value)

Most of the `setIfNull` usages look like this method is used instead of plain `set` "just in case something happens"; even though there seems to be no real chance of the field being non-null. I can take a look at them and maybe remove the method in a follow-up PR.

PR #4221, opened 2021-09-28
Remove deprecated method helperResourceNames from InstrumentationModule

iNikem:


muzzle/src/test/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceCollectorTest.java
```diff
@@ -259,61 +259,6 @@ public void shouldCorrectlyFindExternalInstrumentationClasses() {
     assertThat(helperClasses).containsExactly(ExternalHelper.class.getName());
   }
 
-  @ParameterizedTest
-  @MethodSource
-  public void shouldCollectHelperClassesFromResourceFile(
```
mateuszrzeszutek, line 264: We should not remove that test; it should be rewritten to the new version instead.

trask (approved):

mateuszrzeszutek (approved):

Issue #3901, opened 2021-08-23
Replace deprecated adoptopenjdk docker images


trask:
Just noticed this at https://hub.docker.com/_/adoptopenjdk

> DEPRECATION NOTICE
This image is officially deprecated in favor of [the `eclipse-temurin` image](https://hub.docker.com/_/eclipse-temurin/), and will receive no further updates after 2021-08-01 (Aug 01, 2021). Please adjust your usage accordingly.

iNikem:
I tried to replace adopt jdks with temurin in our GitHub actions: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5287

Unfortunately, this fails, because temurin does not provide java 15, only 8, 11 and 17+. Ideally, we should stop running with java 15 anyway and move to 17 and 18. For that we need to use groovy4, because groovy3 still fails on java 16+

trask:
Current blocker is that there's no ibm-semeru-runtimes windows docker images, so still using adoptium openj9 on windows

trask:
No longer using adoptium anywhere, but still no ibm-semeru-runtimes windows docker images...

trask:
Closing in favor of #9379

PR #4222, opened 2021-09-28
Replace some AdoptOpenJDK J9 docker images with IBM Semeru

iNikem:
Part of #3901


iNikem:
Oh, they are not available yet for windows :( Will wait

PR #4229, opened 2021-09-28
Update patch release doc

trask:


RELEASING.md
```diff
@@ -40,7 +40,27 @@ In general, patch releases are only made for bug-fixes for the following types o
...
+Unfortunately, the automated branch creation in the patch release build workflow
+is currently failing, e.g.
+
+```
+Switched to a new branch 'v1.6.x'
+To https://github.com/open-telemetry/opentelemetry-java-instrumentation
+! [remote rejected]     v1.6.x -> v1.6.x (refusing to allow a GitHub App to create or update workflow `.github/workflows/pr-smoke-test-fake-backend-images.yml` without `workflows` permission)
+```
+
+so if the patch branch doesn't exist already, you will need to manually create it before proceeding,
```
anuraaga, line 54: Well - it doesn't fail every time, just many :P I think the condition is just if the yaml files differ between main and the release branch which is relatively common given dependabot. But I'd personally still try the patch workflow first to see if it works before going for the manual

anuraaga (approved):

Issue #4174, opened 2021-09-21
Support kafka-streams 3.0.0 (and above)


trask:
The current kafka-streams instrumentation is not passing muzzle or tests against the recently released kafka-streams 3.0.0.

(see #4173)

PR #4065, opened 2021-09-08
Add support for Kafka consumer and producer interceptors.

alesj:
Sending a PR after a short discussion #4061 

The goal is to have an equivalent interceptors implementation to this one:
* https://github.com/opentracing-contrib/java-kafka-client/tree/master/opentracing-kafka-client/src/main/java/io/opentracing/contrib/kafka

The main stuff is in TracingKafkaUtils class.

I'm not 100% I got all of the things, specially scopes and contexts right,
so a more experienced eye is welcome.

Plus some suggestions on what to check in the test(s) -- currently I just print out headers.

PR #4236, opened 2021-09-29
Support kafka streams 3

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4174
Apparently tests didn't pass with kafka-streams 1.x perhaps we should also test some intermediate versions?


instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/TracingIterator.java
```diff
@@ -42,6 +43,14 @@ public TracingIterator(
...
   }
 
+  public static <K, V> Iterator<ConsumerRecord<K, V>> wrap(
+      Iterator<ConsumerRecord<K, V>> delegateIterator, @Nullable SpanContext receiveSpanContext) {
+    if (KafkaTracingWrapperUtil.wrappingEnabled()) {
+      return new TracingIterator<>(delegateIterator, receiveSpanContext);
+    }
+    return delegateIterator;
+  }
+
```
trask, line 53: I think there's a similar use case that @mateuszrzeszutek recently ran into recently:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/be090e431a39f54c8d909ac74838ed0a19818322/instrumentation/kafka-streams-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkastreams/StreamTaskInstrumentation.java#L74-L78

would it work to use the same "workaround" for both? (either thread local suppression, or force unwrapping for both)

trask (commented):
let's give a day or two for us to wrap up #4065 with @alesj, that PR has been through enough merge conflicts already 😭

instrumentation/kafka-clients/kafka-clients-0.11/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafkaclients/TracingIterator.java
laurit: I think I can use thread local for both, I'll just have to move the thread local to a different method that exists in all kafka-streams versions

instrumentation/kafka-clients/kafka-clients-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kafka/KafkaTracingWrapperUtil.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.kafka;
+
+public final class KafkaTracingWrapperUtil {
```
mateuszrzeszutek, line 8: I wonder if we could rename it to sth like `KafkaClientsConsumerProcessTracing` - since it only affects the CONSUMER `operation = "process"` part of kafka-clients instrumentation

mateuszrzeszutek (approved):
Let's wait until #4065 is merged though (I think it should be this week)

trask:
@laurit fyi (so we don't duplicate) I'm resolving merge conflicts

laurit:
@trask too late. I think that it might be better to put `KafkaClientsConsumerProcessTracing` into a `bootstrap` module as when kafka-clients and kafka-streams are in different class loaders they would each get their own copy of that class. The same applies to `KafkaConsumerIterableWrapper` and `KafkaConsumerIteratorWrapper` that are used from other instrumentations, probably should have a `KafkaWrapper` interface in bootstrap module.

trask:
No worries, feel free to force push over my changes

mateuszrzeszutek (approved):

instrumentation/kafka-clients/kafka-clients-0.11/bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/kafka/KafkaClientsConsumerProcessTracing.java
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.bootstrap.kafka;
+
+public final class KafkaClientsConsumerProcessTracing {
```
trask, line 8: can you add a brief comment about why this lives in the bootstrap class loader?

instrumentation/kafka-clients/kafka-clients-0.11/bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/kafka/KafkaClientsConsumerProcessWrapper.java
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.bootstrap.kafka;
+
+public interface KafkaClientsConsumerProcessWrapper<T> {
```
trask, line 8: same here

trask (approved):

instrumentation/kafka-clients/kafka-clients-0.11/bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/kafka/KafkaClientsConsumerProcessTracing.java
laurit: added a comment, could you check whether it makes sense

trask: thx 👍

gradle-plugins/src/main/kotlin/io/opentelemetry/javaagent/muzzle/MuzzleDirective.kt
```diff
@@ -42,11 +42,11 @@ abstract class MuzzleDirective {
   /**
    * Adds extra dependencies to the current muzzle test.
    *
-   * @param compileString An extra dependency in the gradle canonical form:
-   * '<group_id>:<artifact_id>:<version_id>'.
+   * @param dependencyNotation An extra dependency in the gradle canonical form:
+   * '<group_id>:<artifact_id>:<version_id>' or a project dependency project(':some-project').
```
trask, line 46: 👍

PR #4237, opened 2021-09-29
Extract HTTP request/response headers as span attributes

mateuszrzeszutek:
This PR implements the [HTTP request/response headers as span attributes semantic convention](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md).

It includes:
* new methods added in the HTTP extractors, and a new "configuration" class that needs to be passed to HTTP extractors
* 4 new javaagent config properties (see `HttpHeadersConfig`)
* implementations of the new methods in all HTTP instrumentations that we have

Things that will be done in the followings PRs:
* remove `userAgent()` and `host()` implementations from all instrumentations, they're not needed anymore
* implement captured headers configuration for library instrumentations
* add tests that verify that headers were captured to `HttpClientTest` and `HttpServerTest`
* move client IP extraction from `ServerInstrumenter` to `HttpServerAttributesExtractor`
* maybe introduce a universal `HttpHeadersGetter` for all HTTP instrumentations? since we already have a method that returns values of any header in the extractor (it could implement https://github.com/open-telemetry/opentelemetry-specification/pull/1884 properly too, for all HTTP instrumentations at the same time)
* some documentation for the new config properties (other than Javadocs 😅 )

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/CapturedHttpHeaders.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/CapturedHttpHeaders.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.http;
+
+import static java.util.Collections.emptyList;
+
+import com.google.auto.value.AutoValue;
+import java.util.List;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+/**
+ * Represents the configuration that specifies which HTTP request/response headers should be
+ * captured as span attributes.
+ */
+@AutoValue
+public abstract class CapturedHttpHeaders {
+
+  private static final CapturedHttpHeaders EMPTY = create(emptyList(), emptyList());
+
+  /** Don't capture any HTTP headers as span attributes. */
+  public static CapturedHttpHeaders empty() {
+    return EMPTY;
+  }
+
+  /**
+   * Captures the configured HTTP request and response headers as span attributes as described in <a
+   * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers">HTTP
+   * semantic conventions</a>.
+   *
+   * <p>The HTTP request header values will be captured under the {@code http.request.header.<name>}
+   * attribute key. The HTTP response header values will be captured under the {@code
+   * http.response.header.<name>} attribute key. The {@code <name>} part in the attribute key is the
+   * normalized header name: lowercase, with dashes replaced by underscores.
+   *
+   * @param capturedRequestHeaders A list of HTTP request header names that are to be captured as
+   *     span attributes.
+   * @param capturedResponseHeaders A list of HTTP response header names that are to be captured as
+   *     span attributes.
+   */
+  public static CapturedHttpHeaders create(
+      List<String> capturedRequestHeaders, List<String> capturedResponseHeaders) {
+    return new AutoValue_CapturedHttpHeaders(
+        lowercase(capturedRequestHeaders), lowercase(capturedResponseHeaders));
+  }
+
+  private static List<String> lowercase(List<String> names) {
+    return names.stream().map(s -> s.toLowerCase(Locale.ROOT)).collect(Collectors.toList());
+  }
+
+  /** Returns the list of HTTP request header names that are to be captured as span attributes. */
+  public abstract List<String> requestHeaders();
+
+  /** Returns the list of HTTP response header names that are to be captured as span attributes. */
+  public abstract List<String> responseHeaders();
+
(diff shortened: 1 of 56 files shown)


javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/config/HttpHeadersConfig.java
```diff
@@ -0,0 +1,42 @@
...
+
+  private static final CapturedHttpHeaders CLIENT;
+  private static final CapturedHttpHeaders SERVER;
+
+  static {
+    Config config = Config.get();
+    CLIENT =
+        CapturedHttpHeaders.create(
+            config.getList(
+                "otel.instrumentation.common.experimental.capture-http-headers.client.request"),
```
iNikem, line 22: Shouldn't this part of `ExperimentalConfig`?

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/CapturedHttpHeaders.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Captures the configured HTTP request and response headers as span attributes as described in <a
+   * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-request-and-response-headers">HTTP
+   * semantic conventions</a>.
+   *
+   * <p>The HTTP request header values will be captured under the {@code http.request.header.<name>}
+   * attribute key. The HTTP response header values will be captured under the {@code
+   * http.response.header.<name>} attribute key. The {@code <name>} part in the attribute key is the
+   * normalized header name: lowercase, with dashes replaced by underscores.
```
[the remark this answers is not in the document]
iNikem: I see now.

javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/config/HttpHeadersConfig.java
mateuszrzeszutek: Hmm, you're right. Do you mind if I do that in a separate PR, after this one an https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4255 are both merged? I'll just add another item to the TODO list in the PR description (and create a tracking issue for all of them once this PR is merged)

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientHttpAttributesExtractor.java
```diff
@@ -68,4 +76,12 @@ protected Long responseContentLengthUncompressed(
       ApacheHttpClientRequest request, HttpResponse response) {
     return null;
   }
+
+  @Override
+  protected List<String> responseHeader(
+      ApacheHttpClientRequest request, HttpResponse response, String name) {
+    return Arrays.stream(response.getHeaders(name))
+        .map(Header::getValue)
+        .collect(Collectors.toList());
```
trask, line 85: i'm a bit concerned about streams API overhead (memory allocation in particular) for small collections like this

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpHeaderAttributes.java
```diff
@@ -0,0 +1,34 @@
...
+import io.opentelemetry.instrumentation.api.caching.Cache;
+import java.util.List;
+
+final class HttpHeaderAttributes {
+
+  private static final Cache<String, AttributeKey<List<String>>> requestKeysCache =
+      Cache.newBuilder().setMaximumSize(32).build();
+  private static final Cache<String, AttributeKey<List<String>>> responseKeysCache =
+      Cache.newBuilder().setMaximumSize(32).build();
+
```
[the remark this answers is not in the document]
mateuszrzeszutek: We'd create different `AttributeKey` values for same headers, if there were more than 1 instrumentations loaded.
And I wanted to avoid exposing the attribute keys in any public APIs (although now that I think of it, it could be implemented without doing so)

instrumentation/apache-httpasyncclient-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpasyncclient/ApacheHttpAsyncClientHttpAttributesExtractor.java
mateuszrzeszutek: I've removed those few streams API calls and replaced them with a simple loop

trask (approved):

Issue #4244, opened 2021-09-29
v1.6.1 crashes when extension project is built with muzzle plugin 0.7.0


edwardxia:
Commit https://github.com/open-telemetry/opentelemetry-java-instrumentation/commit/0d0d2ca09b0952340bf7cfeb4aa9b1cdcd423727 in v1.6.1 which rename tons of muzzle related package names causes crash with extensions repository with muzzle plugin 0.7.0, and meanwhile muzzle plugin 0.8.0 is not yet released.

```
Caused by: java.lang.NoClassDefFoundError: io/opentelemetry/javaagent/extension/muzzle/Flag
        at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
        at java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:3137)
        at java.base/java.lang.Class.getConstructor0(Class.java:3342)
        at java.base/java.lang.Class.getConstructor(Class.java:2151)
        at java.base/java.util.ServiceLoader$1.run(ServiceLoader.java:659)
        at java.base/java.util.ServiceLoader$1.run(ServiceLoader.java:656)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.base/java.util.ServiceLoader.getConstructor(ServiceLoader.java:667)
        ... 22 more
Caused by: java.lang.ClassNotFoundException: io.opentelemetry.javaagent.extension.muzzle.Flag
        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:471)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
        ... 30 more
```

edwardxia:
I don't understand how a breaking change get into a patch release. I see two ways out:

- Revert this commit and release 1.6.2 ASAP. - This option is probably more appropriate.
- Release muzzle plugin 0.8.0 and clarify that 1.6.1 will need muzzle plugin 0.8.0

trask:
oh no 😞, thanks for reporting this

unfortunately the 1.6.0 tag depended on a gradle-plugins snapshot, which made releasing 1.6.1 really painful. i think the best is to revert everything in the 1.6.x branch and release 1.6.2 which is essentially the same as 1.6.0, and give up on making a patch release for #4191

trask:
[...]

PR #4245, opened 2021-09-29
v1.6.x branch: revert breaking muzzle changes

trask:
See #4244

This PR has three commits:

* Straight revert of the problematic cherry-pick which introduced breaking changes for extensions
* Fresh and partial cherry-pick of #4183, including the composite build (which solves the problem of the v1.6.0 tag being built against a snapshot version of gradle-plugins), but excluding package renames for muzzle classes that are used at runtime.
* Cherry-pick of #4173 which is needed due to new kafka-streams release


trask:
I tested this manually against an extension built against 1.6.0

anuraaga (approved):

PR #4248, opened 2021-09-29
Sync gradle-plugins version with main project

trask:
I think releasing these simultaneously is important, since we are doing composite build + dependency substitution now, it seems like without this it would be easy for the substituted local dependency to drift from the published version that extension authors use(?)

I tried but failed to get nebula working in `gradle-plugins` because it's not in the git root, and `git.root` can't be relative, and what I really want is to set it in the gradle file so that you don't have to always specify it on the command line, but I couldn't figure out how to do that...


anuraaga:
For the plugins, since they aren't published on the main build and the automatic tag based versioning isn't needed there, does it make sense to just read the system property we set directly?

trask:
> For the plugins, since they aren't published on the main build and the automatic tag based versioning isn't needed there, does it make sense to just read the system property we set directly?

@anuraaga here's an attempt at this: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4248/commits/222935ad8eaa274aae185b8cb64fae84c1226ede, the examples projects still require some hard-coding or coordination (composite? 😬)

.github/workflows/ci.yml
```diff
@@ -271,10 +271,14 @@ jobs:
           path: ~/.gradle/wrapper
           key: ${{ runner.os }}-gradle-wrapper-cache-${{ hashFiles('examples/distro/gradle/wrapper/gradle-wrapper.properties') }}
 
-      - name: Local publish
+      - name: Local publish of artifacts
         # javadoc task fails sporadically fetching https://docs.oracle.com/javase/8/docs/api/
         run: ./gradlew publishToMavenLocal -x javadoc
 
+      - name: Local publish of gradle plugins
+        run: ../gradlew publishToMavenLocal -Pversion=1.7.0-alpha-SNAPSHOT
```
anuraaga, line 279: It looks like the new feature is this publishToMavenLocal, which we don't like due to the version requirement (and not being able to nebula the version).

It may work to do something like this in the examples settings.gradle

```
if (gradle.startParameters["includeGradlePlubins"]) {
  includeBuild("../gradle-plugins") {
  dependencySubstitution {
    substitute(module("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin")).using(project(":"))
    substitute(module("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin")).using(project(":"))
  }
}
```
a bit similar to https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/conventions/settings.gradle.kts#L1 but allowing examples to still build independently too.

trask:
ptal at the changes to the release instructions in case you have idea how to avoid any steps

trask:
This is ready for a fresh look

RELEASING.md
```diff
@@ -23,6 +32,14 @@ to release in the input field that pops up, and then press "Run workflow".
 This triggers the release process, which builds the artifacts, publishes the artifacts, and creates
 and pushes a git tag with the version number.
 
+After making the release:
+
+* Update the version (bump and add `-SNAPSHOT`) in these files:
+  * build.gradle.kts
+  * gradle-plugins/build.gradle.kts
```
mateuszrzeszutek, line 39: Nit: WDYT about keeping the version in some `version.properties` file that could be reused by all projects involved in the composite build?

.github/workflows/pr.yml
```diff
@@ -253,8 +253,6 @@ jobs:
 
   examples:
     runs-on: ubuntu-latest
-    # When we make PR against pre-release branch examples may point to yet-unpublished version
-    if: github.base_ref == 'main'
```
iNikem, line 257: Why is this not needed any more?

conventions/build.gradle.kts
```diff
@@ -29,8 +29,8 @@ dependencies {
   implementation(gradleApi())
   implementation(localGroovy())
 
-  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:0.8.0")
-  implementation("io.opentelemetry.instrumentation.muzzle-check:io.opentelemetry.instrumentation.muzzle-check.gradle.plugin:0.8.0")
+  implementation("io.opentelemetry.instrumentation.muzzle-generation:io.opentelemetry.instrumentation.muzzle-generation.gradle.plugin:${version}")
```
iNikem, line 32: conventions always include plugins as composite, probably can skip version altogether.

iNikem (approved):

PR #4249, opened 2021-09-30
Faster patch release build

trask:
I'm watching the patch release, and wondering why the final `release` step is taking so long, and then wondering why the final `release` step is running tests when they've already been run in prior steps.

Looks like this update was already made for the normal release build.

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -235,7 +235,7 @@ jobs:
         with:
           job-id: jdk11
           remote-build-cache-proxy-enabled: false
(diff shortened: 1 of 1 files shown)


anuraaga (approved):

iNikem:
More interesting question is why gradle cache does not work :(

iNikem (approved):

mateuszrzeszutek (approved):

anuraaga:
@iNikem It looks like we don't mount the S3 key to allow the `test` job to publish artifacts, which are probably quite unique to the release branch. And the burrunan action is set to read-only. Should we change one of those?

iNikem:
Ok, our release job does the same. Let it be so for now

PR #4254, opened 2021-09-30
AWS lambda - improvements in custom type handling in wrappers, SQS ev…

kubawach:
…en wrapper added

Two goals with this PR:
- allow users of lambda wrappers to use custom types in their handlers (along with simple types and build-in events)
- add wrapper for SQS event handler

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/LambdaParameters.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/LambdaParameters.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import java.lang.reflect.Method;
+import java.util.function.BiFunction;
+
+class LambdaParameters {
+
+  static <T> Object[] toArray(
+      Method targetMethod, T input, Context context, BiFunction<T, Class, Object> mapper) {
+    Class<?>[] parameterTypes = targetMethod.getParameterTypes();
+    Object[] parameters = new Object[parameterTypes.length];
+    for (int i = 0; i < parameterTypes.length; i++) {
+      Class clazz = parameterTypes[i];
+      boolean isContext = clazz.equals(Context.class);
+      if (isContext) {
+        parameters[i] = context;
+      } else if (i == 0) {
+        parameters[0] = (clazz.isInstance(input) ? input : mapper.apply(input, clazz));
+      }
+    }
+    return parameters;
+  }
+}
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
@@ -5,23 +5,58 @@
 
 package io.opentelemetry.instrumentation.awslambda.v1_0;
 
+import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.fasterxml.jackson.core.JsonProcessingException;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
+import java.util.function.BiFunction;
 
 /**
  * Wrapper for {@link TracingRequestHandler}. Allows for wrapping a lambda proxied through API
  * Gateway, enabling single span tracing and HTTP context propagation.
  */
 public class TracingRequestApiGatewayWrapper
-    extends TracingRequestWrapperBase<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+    extends TracingRequestWrapperBase<APIGatewayProxyRequestEvent, Object> {
 
   public TracingRequestApiGatewayWrapper() {
-    super();
+    super(TracingRequestApiGatewayWrapper::map);
   }
 
   // Visible for testing
-  TracingRequestApiGatewayWrapper(OpenTelemetrySdk openTelemetrySdk, WrappedLambda wrappedLambda) {
-    super(openTelemetrySdk, wrappedLambda);
+  TracingRequestApiGatewayWrapper(
+      OpenTelemetrySdk openTelemetrySdk,
+      WrappedLambda wrappedLambda,
+      BiFunction<APIGatewayProxyRequestEvent, Class, Object> mapper) {
+    super(openTelemetrySdk, wrappedLambda, mapper);
+  }
+
+  // Visible for testing
+  static Object map(APIGatewayProxyRequestEvent event, Class clazz) {
+    try {
+      return OBJECT_MAPPER.readValue(event.getBody(), clazz);
+    } catch (JsonProcessingException e) {
+      throw new IllegalStateException(
+          "Could not map API Gateway event body to requested parameter type: " + clazz, e);
+    }
+  }
+
+  @Override
+  protected APIGatewayProxyResponseEvent doHandleRequest(
+      APIGatewayProxyRequestEvent input, Context context) {
+    Object result = super.doHandleRequest(input, context);
+    APIGatewayProxyResponseEvent event = null;
+    // map to response event if needed
+    if (result instanceof APIGatewayProxyResponseEvent) {
+      event = (APIGatewayProxyResponseEvent) result;
+    } else {
+      try {
+        event = new APIGatewayProxyResponseEvent();
+        event.setBody(OBJECT_MAPPER.writeValueAsString(result));
+      } catch (JsonProcessingException e) {
+        throw new IllegalStateException("Could not serialize return value.", e);
+      }
+    }
+    return event;
   }
 }
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapper.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapper.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.instrumentation.awslambda.v1_0;
 
 import io.opentelemetry.sdk.OpenTelemetrySdk;
+import java.util.function.BiFunction;
 
 /**
  * Wrapper for {@link TracingRequestHandler}. Allows for wrapping a regular lambda, not proxied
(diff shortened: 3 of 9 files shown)


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
```diff
@@ -6,54 +6,44 @@
...
  */
 abstract class TracingRequestWrapperBase<I, O> extends TracingRequestHandler<I, O> {
 
+  protected static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
   private final WrappedLambda wrappedLambda;
+  private final BiFunction<I, Class, Object> parameterMapper;
 
-  protected TracingRequestWrapperBase() {
-    this(OpenTelemetrySdkAutoConfiguration.initialize(), WrappedLambda.fromConfiguration());
+  protected TracingRequestWrapperBase(BiFunction<I, Class, Object> parameterMapper) {
```
mateuszrzeszutek, line 26: Nit: can be package private
```suggestion
  TracingRequestWrapperBase(BiFunction<I, Class, Object> parameterMapper) {
```

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsEventWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+
+  // Visible for testing
+  TracingSqsEventWrapper(OpenTelemetrySdk openTelemetrySdk, WrappedLambda wrappedLambda) {
+    super(openTelemetrySdk, WrapperConfiguration.flushTimeout());
+    this.wrappedLambda = wrappedLambda;
+  }
+
+  @Override
+  protected void handleEvent(SQSEvent sqsEvent, Context context) {
+    Method targetMethod = wrappedLambda.getRequestTargetMethod();
```
mateuszrzeszutek, line 31: Can we do this in the constructor, instead of every function call?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsEventWrapper.java
```diff
@@ -0,0 +1,44 @@
...
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.events.SQSEvent;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.OpenTelemetrySdkAutoConfiguration;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class TracingSqsEventWrapper extends TracingSqsEventHandler {
```
mateuszrzeszutek, line 15: Can you add some tests for the SQS wrapper?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
kubawach: I wanted to leave this package private for extension in downstream distros.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsEventWrapper.java
kubawach: Yeah, definitely.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsEventWrapper.java
kubawach: Done!

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/LambdaParameters.java
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import java.lang.reflect.Method;
+import java.util.function.BiFunction;
+
+class LambdaParameters {
```
anuraaga, line 12:
```suggestion
final class LambdaParameters {
```

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/LambdaParameters.java
```diff
@@ -0,0 +1,29 @@
...
+      boolean isContext = clazz.equals(Context.class);
+      if (isContext) {
+        parameters[i] = context;
+      } else if (i == 0) {
+        parameters[0] = (clazz.isInstance(input) ? input : mapper.apply(input, clazz));
+      }
+    }
+    return parameters;
+  }
+}
```
anuraaga, line 29:
```suggestion
  private LambdaParameters() {}

}
```

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
```diff
@@ -5,23 +5,58 @@
...
+      APIGatewayProxyRequestEvent input, Context context) {
+    Object result = super.doHandleRequest(input, context);
+    APIGatewayProxyResponseEvent event = null;
+    // map to response event if needed
+    if (result instanceof APIGatewayProxyResponseEvent) {
+      event = (APIGatewayProxyResponseEvent) result;
+    } else {
+      try {
+        event = new APIGatewayProxyResponseEvent();
+        event.setBody(OBJECT_MAPPER.writeValueAsString(result));
```
anuraaga, line 55: Do we know that our OBJECT_MAPPER is configured the same as Lambda?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
```diff
@@ -5,23 +5,58 @@
...
+      } catch (JsonProcessingException e) {
+        throw new IllegalStateException("Could not serialize return value.", e);
```
anuraaga, line 57: These exceptions seem worrisome, are we sure we won't break the app?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
kubawach: Unfortunately no guarantees here. Haven't seen an official spec by AWS guys regarding how the JSON is mapped in their code.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
kubawach: The silver lining here is that if the wrapper is unable to map to a custom type, the user can still use the "APIGatewayProxyRequestEvent" and avoid mapping on the wrapper side entirely. So in the worst case it will work exactly the same way it does right now.

instrumentation/aws-lambda-1.0/library/src/test/groovy/io/opentelemetry/instrumentation/awslambda/v1_0/TracingSqsEventWrapperTest.groovy
```diff
@@ -0,0 +1,86 @@
...
+    environmentVariables.set(WrappedLambda.OTEL_LAMBDA_HANDLER_ENV_KEY, handler)
+    wrapper = wrapperConstructor.call(testRunner().openTelemetrySdk, WrappedLambda.fromConfiguration())
+  }
+
+  static class TestRequestHandler implements RequestHandler<SQSEvent, Void> {
+
+    @Override
+    Void handleRequest(SQSEvent input, Context context) {
+      return null
+      throw new IllegalArgumentException("bad argument")
```
mateuszrzeszutek, line 45: That's an unreachable statement here

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
anuraaga: I think it still means we have the possibility of breaking the lambda just by the presence of a lambda layer such as the `opentelemetry-lambda` ones. I don't think it's good to have the possibility of adding crashing.

That being said I guess I can't find any alternative unfortunately. It's too bad the Context object doesn't provide information about HTTP requests...

Is it at least beneficial to split into one where the target handler is proxy that doesn't have any mapping and this one that does for custom return type?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapper.java
```diff
@@ -6,18 +6,32 @@
...
-    super(openTelemetrySdk, wrappedLambda);
+  TracingRequestWrapper(
+      OpenTelemetrySdk openTelemetrySdk,
+      WrappedLambda wrappedLambda,
+      BiFunction<Object, Class, Object> mapper) {
+    super(openTelemetrySdk, wrappedLambda, mapper);
+  }
+
+  // Visible for testing
+  static Object map(Object jsonMap, Class clazz) {
```
anuraaga, line 29: I'm surprised this non-API Gateway class (so not populating HTTP attrs) is mapping, does it need to?

instrumentation/aws-lambda-1.0/library/src/test/java/io/opentelemetry/instrumentation/awslambda/v1_0/LambdaParametersTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.jupiter.api.Test;
+
+class LambdaParametersTest {
+
+  public void onlyContext(Context context) {}
+
+  public void contextOnThird(String one, String two, Context context) {}
+
+  @Test
+  public void shouldSetContextOnFirstPosition() throws NoSuchMethodException {
```
anuraaga, line 22: Dont' need `public` in JUnit5

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestApiGatewayWrapper.java
kubawach: It's implemented (kind of) this way - mapping ins only done if the return type of the handler is different than APIGatewayProxyResponseEvent. So if the wrapper mapping causes issues, the user can always switch to return the event. In other words both with input and result if the new wrapper breaks, user can implement own handler to input / return the event - and everything will work the same way it did (no mapping if no custom type used).

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapper.java
kubawach: With our wrapper requesting Object input type, AWS will (out of my testing):
- invoke with value in case of primitive values 
- invoke with Map<String, Object> in case of custom types
So in order to support pre-mapped custom types we need to map the Map into appropriate class.

PR #4256, opened 2021-09-30
Convert liberty dispatcher instrumentation to instrumenter api

laurit:
Part of #2713


mateuszrzeszutek (approved):

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherNetAttributesExtractor.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public String transport(LibertyRequest libertyRequest) {
+    return SemanticAttributes.NetTransportValues.IP_TCP;
+  }
+
+  @Override
+  public @Nullable String peerName(
+      LibertyRequest libertyRequest, @Nullable LibertyResponse libertyResponse) {
+    if (!libertyRequest.isCompleted()) {
```
trask, line 23: is checking `isCompleted()` a performance optimization? or is it needed for another reason?

trask (approved):

instrumentation/liberty/liberty-dispatcher/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/liberty/dispatcher/LibertyDispatcherNetAttributesExtractor.java
laurit: It is used to limit calling these methods to `onStart` because in `onEnd` they may throw a NPE.

PR #4257, opened 2021-09-30
Convert jsp integration to Instrumenter API

iNikem:


instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/HttpJspPageInstrumentationSingletons.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private HttpJspPageInstrumentationSingletons() {}
+
+  private static class RenderAttributesExtractor
+      extends AttributesExtractor<HttpServletRequest, Void> {
+
+    @Override
+    protected void onStart(AttributesBuilder attributes, HttpServletRequest request) {
+      if (!CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES) {
```
mateuszrzeszutek, line 59: Nit: you can just not add the extractor in the instrumenter building phase instead

instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/JspCompilationContextInstrumentationSingletons.java
```diff
@@ -0,0 +1,68 @@
...
+        Instrumenter.<JspCompilationContext, Void>newBuilder(
+                GlobalOpenTelemetry.get(),
+                "io.opentelemetry.jsp-2.3",
+                JspCompilationContextInstrumentationSingletons::spanNameOnCompile)
+            .addAttributesExtractor(new CompilationAttributesExtractor())
+            .newInstrumenter(SpanKindExtractor.alwaysInternal());
+  }
+
+  public static String spanNameOnCompile(JspCompilationContext jspCompilationContext) {
+    return jspCompilationContext == null
```
mateuszrzeszutek, line 35: Can `jspCompilationContext` even be null? The instrumenter gets passed the `@Advice.This` instance

instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/HttpJspPageInstrumentation.java
```diff
@@ -52,23 +52,20 @@ public static void onEnter(
         @Advice.Argument(0) HttpServletRequest req,
         @Advice.Local("otelContext") Context context,
         @Advice.Local("otelScope") Scope scope) {
-      context = tracer().startSpan(tracer().spanNameOnRender(req), SpanKind.INTERNAL);
-      tracer().onRender(context, req);
+
+      context = instrumenter().start(Java8BytecodeBridge.currentContext(), req);
```
mateuszrzeszutek, line 56: Can you add the `shouldStart()` check?

instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/JspCompilationContextInstrumentationSingletons.java
iNikem: It cannot, missed this place

instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/HttpJspPageInstrumentation.java
iNikem: Does it make sense here? When can it return `true` for INTERNAL spans?

instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/HttpJspPageInstrumentationSingletons.java
iNikem: I cannot decide which way is better. Adding extractors conditionally breaks the fluent interface of `InstrumenterBuilder`

instrumentation/jsp-2.3/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jsp/HttpJspPageInstrumentation.java
mateuszrzeszutek: I think we should do that for "future compatibility" (and uniformity in all instrumentations) - if we ever introduce some more elaborate suppresion rules, or have a `shouldStart()` customization interface, etc. There's also the `setDisabled` method.

mateuszrzeszutek (approved):

laurit (approved):
Out of cope for this pr but this instrumentation is a bit strange. Really there are 2 instrumentations one for jsp api and another for jasper (jsp engine used on tomcat and many other servers) which probably should be split. For jsp api there should be a separate instrumentation for jakarta api.

Issue #3479, opened 2021-07-02
Spring WebMVC spanName is contant


gquintana:
**Is your feature request related to a problem? Please describe.**
`WebMvcTracingFilter` always set spanName as `WebMVCTracingFilter.doFilterInternal` no matter what the HTTP endpoint is.

See the code https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/WebMvcTracingFilter.java#L33

**Describe the solution you'd like**
I would expect the `spanName` (which is then called `operationName` in Jaeger) which always contains the same value `WebMVCTracingFilter.doFilterInternal` to have distinct values for example `GET /foo`, `POST /bar` to be able to tell them appart  in  the Jaeger UI.

The HTTP method, path, status, etc. could also be added to the Span tags.

**Describe alternatives you've considered**
I tried adding an annotation `@WithSpan("http-get-foo")` it creates an internal span which is better that nothing. But still the WebMVCTracingFilter span is useless.

iNikem:
Do you use javaagent or just library instrumentation that you referenced?

gquintana:
@iNikem Just the libraries `opentelemetry-spring-starter` and `opentelemetry-exporter-jaeger`

mateuszrzeszutek:
We could try implementing the following idea: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4258#discussion_r720087815
By implementing the `ApplicationListener<ContextRefreshedEvent>` and retrieving the `DispatcherServlet` from the context (and using a bit of reflection) we should be able to retrieve the set of handler mappings needed to determine `http.route`.

ZawadaZSE:
Hey,

I'd be also interested in getting this done. Spans named with current library implementation are not easily browsable in jaeger.

reactivejson:
We're also interested to have this functionality for better tracing visibility.

PR #4258, opened 2021-09-30
Convert Spring Web MVC library instrumentation to Instrumenter API

mateuszrzeszutek:


instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcNetAttributesExtractor.java
```diff
@@ -0,0 +1,37 @@
...
+    extends NetAttributesExtractor<HttpServletRequest, HttpServletResponse> {
+  @Override
+  public String transport(HttpServletRequest request) {
+    return SemanticAttributes.NetTransportValues.IP_TCP;
+  }
+
+  @Override
+  public @Nullable String peerName(
+      HttpServletRequest request, @Nullable HttpServletResponse response) {
+    return null;
```
laurit, line 24: Could be filled with
```suggestion
    return request.getRemoteHost();
```

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcHttpAttributesExtractor.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @Override
+  protected @Nullable String flavor(
+      HttpServletRequest request, @Nullable HttpServletResponse response) {
+    return request.getProtocol();
+  }
+
+  @Override
+  protected Integer statusCode(HttpServletRequest request, HttpServletResponse response) {
+    return response.getStatus();
```
laurit, line 45: Would it make sense to mimic servlet instrumentation and set status to 500 when response isn't committed and there is an exception?

instrumentation/spring/spring-webmvc-3.1/library/README.md
```diff
@@ -79,12 +79,12 @@ import org.springframework.web.client.RestTemplate;
 public class WebMvcTracingFilterConfig {
 
    @Bean
-   public WebMvcTracingFilter webMvcTracingFilter(Tracer tracer) {
-      return new WebMvcTracingFilter(tracer);
+   public WebMvcTracingFilter webMvcTracingFilter(OpenTelemetry openTelemetry) {
```
laurit, line 82:
```suggestion
   public Filter webMvcTracingFilter(OpenTelemetry openTelemetry) {
```

laurit: I believe that out of the box this construct works only for spring boot, for plain spring mvc you'll need to also add the filter to web.xml. There is a sample in instrumentation/spring/spring-webmvc-3.1/wildfly-testing/src/test/resources/web.xml

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcHttpAttributesExtractor.java
mateuszrzeszutek: Can you point me to the class where it happens?

instrumentation/spring/spring-webmvc-3.1/library/README.md
mateuszrzeszutek: Yeah, that may be true - the whole instrumentations seems to be written with spring boot (and our spring-boot-autoconfigure module) in mind.
I'll add it to the README file.

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcTracingBuilder.java
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public SpringWebMvcTracing build() {
+    SpringWebMvcHttpAttributesExtractor httpAttributesExtractor =
+        new SpringWebMvcHttpAttributesExtractor();
+
+    Instrumenter<HttpServletRequest, HttpServletResponse> instrumenter =
+        Instrumenter.<HttpServletRequest, HttpServletResponse>newBuilder(
+                openTelemetry,
+                INSTRUMENTATION_NAME,
+                HttpSpanNameExtractor.create(httpAttributesExtractor))
```
laurit, line 55: If I understand correctly this just sets span name to `HTTP GET`, is it correct to say that this fixes #3479? I guess we could somehow adapt `OpenTelemetryHandlerMappingFilter` for library instrumentation that should give better names.

laurit (approved):

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcHttpAttributesExtractor.java
laurit: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/d55eee2ebf74b21ba9f63da44d511d7316d43de8/instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/ServletHttpAttributesExtractor.java#L112
To pass throwable response is wrapped, which happens here
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/d55eee2ebf74b21ba9f63da44d511d7316d43de8/instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/ServletHelper.java#L51

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcTracingBuilder.java
mateuszrzeszutek: > If I understand correctly this just sets span name to `HTTP GET`, is it correct to say that this fixes #3479?

🤷‍♂️  It's better than the class name I guess, and it fits the HTTP spec.
We can't really get better names on the `Filter` level, we'd probably have to implement some additional piece of code that injects itself into Spring machinery - I'm not really sure whether this is worth the effort though.

laurit: Although it is out of the scope of this pr we already have this beast https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/org/springframework/web/servlet/OpenTelemetryHandlerMappingFilter.java A filter that deduces which spring mapping is going to match and sets the name of the server span. I guess the intention behind it is that if request is denied by spring security then it doesn't reach to actual matching so to get a good span name we need to figure out what matches the request before the actual matching in spring takes place. The same approach could probably also be used in library instrumentation.

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcHttpAttributesExtractor.java
mateuszrzeszutek: Nice catch, I'll add this. Thanks!

instrumentation/spring/spring-webmvc-3.1/library/src/main/java/io/opentelemetry/instrumentation/spring/webmvc/SpringWebMvcTracingBuilder.java
mateuszrzeszutek: That's wild, but I think it's doable. I'll remove the association with #3479 and link your comment there - it's a good idea to at least try it out. Thanks!

trask (approved):

PR #4261, opened 2021-09-30
Sync otlp exporter grpc-netty-shaded version with opentelemetry-java

jack-berg:
`opentelemetry-java` uses [grpc version 1.40.1](https://github.com/open-telemetry/opentelemetry-java/blob/main/dependencyManagement/build.gradle.kts#L19). 

I know that its sometimes ok using different grpc api and implementation versions, but sometimes not. I think we ought to try to keep these in sync.

### DIFF
--- a/javaagent-exporters/build.gradle.kts
+++ b/javaagent-exporters/build.gradle.kts
@@ -19,5 +19,5 @@ dependencies {
   implementation("io.opentelemetry:opentelemetry-exporter-zipkin")
 
   // TODO(anuraaga): Move version to dependency management
-  implementation("io.grpc:grpc-netty-shaded:1.38.0")
+  implementation("io.grpc:grpc-netty-shaded:1.40.1")
 }


trask (approved):
thx @jack-berg!

I'm curious to get @anuraaga and @jkwatson's thoughts on this:

> I know that its sometimes ok using different grpc api and implementation versions, but sometimes not.

Would this be considered a breaking change if (hypothetically) SDK 1.5.0 worked successfully against gRPC 1.38.0 but SDK 1.6.0 had failures against gRPC 1.38.0?

mateuszrzeszutek (approved):

iNikem (approved):

jkwatson:
> Would this be considered a breaking change if (hypothetically) OTLP gRPC Exporter 1.5.0 worked successfully against gRPC 1.38.0 but OTLP gRPC Exporter 1.6.0 had failures against gRPC 1.38.0?

Although I don't think it's cut-and-dried, I think we wouldn't consider this a change that would necessitate revving a major version under semver. If we did, that could end up meaning we'd have to do a major version bump in the case of a security vulnerability (hypothetical) that was found in a dependent library.

trask:
> > Would this be considered a breaking change if (hypothetically) OTLP gRPC Exporter 1.5.0 worked successfully against gRPC 1.38.0 but OTLP gRPC Exporter 1.6.0 had failures against gRPC 1.38.0?
> 
> Although I don't think it's cut-and-dried, I think we wouldn't consider this a change that would necessitate revving a major version under semver. If we did, that could end up meaning we'd have to do a major version bump in the case of a security vulnerability (hypothetical) that was found in a dependent library.

thx! I discussed with @anuraaga in SIG also and this position makes sense 👍

I will add a step to RELEASING.md to ensure that the gRPC versions are in sync before releasing, at least until we figure out a way to enforce this automatically

Issue #4188, opened 2021-09-22
Occasionally get connection timeout exception occur in HTTP requests


GlowingRuby:
**Describe the bug**

Deadlocks were found through thread dump, and the thread stacks were

![image](https://user-images.githubusercontent.com/26060287/134326540-e6b15e67-595c-40bc-99b1-bd43bb4adfef.png)
![image](https://user-images.githubusercontent.com/26060287/134326675-d0a32b11-3c9d-4315-ad6a-8c6769e45bc5.png)

In sharding-jdbc-1.5.4.1, called com.dangdang.ddframe.rdb.sharding.jdbc.core.connection.ShardingConnection#getMetaData, If there is no cached connection(because of druid connection pool logic), the metadata is obtained after the connection is created. Then the mysql driver call com.mysql.jdbc.ConnectionImpl#loadServerVariables When creating a connection to execute SQL, Triggered the io.opentelemetry.instrumentation.jdbc.internal.JdbcUtils#extractDbInfo method, but the lock has already been acquired by the thread executing the HTTP request, and the thread executing the HTTP request is waiting for the connection to be created, A deadlock has been created.

**What version are you using?**
v1.4.1

trask:
@GlowingRuby thanks for reporting this!! I've sent a PR to address it #4191

GlowingRuby:
> @GlowingRuby thanks for reporting this!! I've sent a PR to address it #4191

You're welcome~ 
Do we need to consider whether deadlock can also occur with other uses of WeakLockFreeCache?

trask:
> Do we need to consider whether deadlock can also occur with other uses of WeakLockFreeCache?

that's an excellent question, I've opened #4192 to track and get others' feedback. I did do a cursory review of other calls to `computeIfAbsent()` and didn't see any other glaring callbacks to application code, but it's definitely worth a more detailed review.

PR #4262, opened 2021-10-01
Remove unnecessary VirtualField#setIfNull() method

mateuszrzeszutek:
I was used in 3 instrumentations:
* Hibernate -- the field value was set for freshly constructed objects (`@Advice.Return` instrumentation), there couldn't have been any value other than null
* Netty -- I replaced it with a much safer alternative that skips adding the handler altogether if there's a non-null value attached
* Servlet -- only used in `(Filter|Servlet).init()` calls, I doubt that a servlet instance will have its `init()` method called more than once (and even if it somehow does get called again, it's probably safer to construct the mappings again based on the most recent `ServletConfig` value)

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/context/FieldBackedProvider.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/context/FieldBackedProvider.java
@@ -882,21 +882,6 @@ public class FieldBackedProvider implements InstrumentationContextProvider {
     }
 
     @Override
-    public void setIfNull(Object object, Object fieldValue) {
-      Object oldFieldValue = realGet(object);
-      if (oldFieldValue != null) {
-        return;
-      }
-      synchronized (realSynchronizeInstance(object)) {
-        oldFieldValue = realGet(object);
-        if (oldFieldValue != null) {
-          return;
-        }
-        realPut(object, fieldValue);
-      }
-    }
-
-    @Override
     public Object computeIfNull(Object object, Supplier<Object> fieldValueSupplier) {
       Object existingContext = realGet(object);
       if (null != existingContext) {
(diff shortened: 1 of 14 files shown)


iNikem (approved):

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/context/FieldBackedProvider.java
```diff
@@ -881,21 +881,6 @@ public Object get(Object object) {
       return realGet(object);
     }
 
-    @Override
-    public void setIfNull(Object object, Object fieldValue) {
```
laurit, line 885: Now if we could also get rid of `computeIfNull`. There seems to be a small difference between `set` and `setIfNull`. `set` just writes the field, `setIfNull` synchronizes around setting the field. As `get` doesn't synchronize it probably wasn't correct anyway from the concurrency side. Should we make the field volatile to make it at least a bit safer to use? For `computeIfNull` should we synchronize at all, as `set` doesn't synchronize we can't get reasonable multi thread behaviour from this anyway. Or should `set` also synchronize?

mateuszrzeszutek: > Should we make the field volatile to make it at least a bit safer to use?

It wasn't? 😱 

We definitely should! Any changes to those fields should definitely be visible in other threads - they're often used to pass context between different threads.

> Now if we could also get rid of `computeIfNull`.

I guess we could, but it is useful in those several places that use it - it makes the code a bit clearer.

> For `computeIfNull` should we synchronize at all, as `set` doesn't synchronize we can't get reasonable multi thread behaviour from this anyway.

I don't think we care about atomicity in the case of `VirtualField` (it has even proved to be a problem in certain situations: #4188), the visibility of changes is what's really important. I think I'd be fine with dropping the synchronization and just making the field volatile.

trask (approved):
nice 👍

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/context/FieldBackedProvider.java
trask: I created 2 tracking issues to continue these discussions 👍 #4268 and #4269

PR #4267, opened 2021-10-01
Remove ContextStore/InstrumentationContext mentions from internal age…

mateuszrzeszutek:
…nt classes

This PR is 100% renames. I think I'll slowly start working on refactoring/cleaning up the `FieldBackedProvider` class after that, I've noticed several things that could be improved (and started analyzing the DD code)


trask (approved):
not overloading the term `context` quite as much anymore makes me 🥰

PR #4273, opened 2021-10-03
Convert Spymemcached integration to Instrumenter

iNikem:


instrumentation/spymemcached-2.12/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spymemcached/SpymemcachedAttributeExtractor.java
```diff
@@ -0,0 +1,36 @@
+package io.opentelemetry.javaagent.instrumentation.spymemcached;
+
+import io.opentelemetry.instrumentation.api.instrumenter.db.DbAttributesExtractor;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class SpymemcachedAttributeExtractor extends DbAttributesExtractor<SpymemcachedRequest, Object> {
+  @Override
+  protected @Nullable String system(SpymemcachedRequest spymemcachedRequest) {
```
trask, line 8:
```suggestion
  protected String system(SpymemcachedRequest spymemcachedRequest) {
```

trask (approved):

instrumentation/spymemcached-2.12/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spymemcached/SpymemcachedRequest.java
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  public abstract MemcachedConnection getConnection();
+
+  public abstract String getStatement();
+
+  public String dbOperation() {
+    char[] chars =
+        getStatement()
+            .replaceFirst("^async", "")
```
anuraaga, line 25: These look like they can have significant performance overhead on all statements, not just these two, because they compile the regex every single time. How about `startsWith` + `substring`?

anuraaga: I realize this is copy-pasted from the old code - please file an issue if don't want to fix in this PR. Very obvious when reviewing it though :)

instrumentation/spymemcached-2.12/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spymemcached/CompletionListener.java
```diff
@@ -26,10 +26,12 @@
   private static final String MISS = "miss";
 
   private final Context context;
+  private final SpymemcachedRequest request;
 
   protected CompletionListener(
       Context parentContext, MemcachedConnection connection, String methodName) {
-    context = tracer().startSpan(parentContext, connection, methodName);
+    request = SpymemcachedRequest.create(connection, methodName);
```
mateuszrzeszutek, line 33: Can we refactor this instrumentation to insert `shouldStart()` somewhere? E.g. not register the listener at all if `shouldStart()` returns false

iNikem: We still have to create `SpymemcachedRequest` for `shouldStart`. Plus we have `CallDepth` check before creating listener. I don't think we are going to have any performance gain from this 🤔

mateuszrzeszutek:
Oops, sorry, didn't mean to merge it -- I mistakenly clicked on the wrong PR 🙇

iNikem:
> Oops, sorry, didn't mean to merge it -- I mistakenly clicked on the wrong PR 🙇

:D Well, as a punishment you now have to create a separate issue for any follow-ups left unresolved in this PR :)

mateuszrzeszutek:
Good thing it was only my comment 😂

anuraaga:
LGTM :)

PR #4276, opened 2021-10-03
Avoid duplicative net attribute extraction

trask:
Motivated by https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4256#discussion_r720723396

But also motivated by performance, and not collecting net attributes twice (even if the underlying collection methods are fast, we are still allocating memory to store them in AttributesBuilders twice).

And in our current instrumentations, it looks like net attributes are either extracted from request or response but not both.


trask:
Closing, this was not great, will open an issue with some options

PR #4277, opened 2021-10-03
Check if span name is null before use

aras112:
Because of getSpanName can return null, we should check this before SERVER_SPAN_NAME uses spanName() which has requireNonNull. Otherwise we receive NPE in main app, when spanName is NULL.


trask:
hey @aras112! let us know if you are able to sign the CLA above, thx

aras112:
@trask Done

anuraaga (commented):
Do you know if it's possible to add a unit test case which will trigger the NPE without this change?

mateuszrzeszutek (approved):

trask (approved):
thx @aras112!

trask:
@aras112 looks like it needs a `./gradlew spotlessApply` to make CI go green

PR #4278, opened 2021-10-04
NetAttributesExtractor optimization

trask:
NetAttributesExtractor is currently often collecting net attributes twice.

Even if the underlying collection methods are fast, we are still allocating memory to store them in AttributesBuilders twice.

Also motivated by the issue where extracting from request can throw exception during onEnd: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4256#discussion_r720723396

Initially I tried to split it into two AttributesExtractors, one that just captures from request onStart, and one that just captures from response onEnd. This almost worked except for armeria instrumentation which captures from request onEnd (and there may be others, this was just the first one that failed tests after I split them up).

The duplicative hierarchy wasn't lovely either since it led to separate NetAttributesExtractors and separate InetSocketAddressNetAttributesExtractors.

I think this is a fairly low impact change to the extractors, and provides a good value.

This PR currently preserves the existing behavior by providing default NetAttributesExtractor and InetSocketAddressNetAttributesExtractor constructors which default to `ON_BOTH`. An option is to remove those so that users are more likely to think about which behavior they want. I'm not even sure if `ON_BOTH` is ever really needed, so could likely even remove that option.

### DIFF
--- a/benchmark/src/jmh/java/io/opentelemetry/benchmark/InstrumenterBenchmark.java
+++ b/benchmark/src/jmh/java/io/opentelemetry/benchmark/InstrumenterBenchmark.java
@@ -110,6 +110,10 @@ public class InstrumenterBenchmark {
     private static final InetSocketAddress ADDRESS =
         InetSocketAddress.createUnresolved("localhost", 8080);
 
+    protected ConstantNetAttributesExtractor() {
+      super(NetPeerAttributeExtraction.ON_START);
+    }
+
     @Override
     public @Nullable InetSocketAddress getAddress(Void unused, @Nullable Void unused2) {
       return ADDRESS;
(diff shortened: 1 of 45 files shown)


mateuszrzeszutek:
We could also have two methods responsible for extraction of a single attribute:

```java
public abstract String peerName(REQUEST request);

public abstract String peerName(REQUEST request, @Nullable RESPONSE response);
```

I'm not really sure if this is a better solution though.

If we end up making the behavior configurable (via an enum passed in the constructor) we should coordinate this with changes in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4235 - where we could potentially also make the peer service mappings configurable.

> I'm not even sure if `ON_BOTH` is ever really needed, so could likely even remove that option.

👍 
I don't think it's that useful to have that option, you usually extract these attributes either on start or on end.

trask:
> We could also have two methods responsible for extraction of a single attribute:
> 
> ```java
> public abstract String peerName(REQUEST request);
> 
> public abstract String peerName(REQUEST request, @Nullable RESPONSE response);
> ```

Ya, this PR feels a little weird because of using the same method in two different ways. And agree splitting the methods out in the same class doesn't seem ideal either. See alternative of splitting out different types in #4287 and let me know if that seems better.

trask:
Closing in favor of reworking #4287 around client/server split

Issue #4280, opened 2021-10-04
Update "Writing instrumentation" document to Instrumenters


iNikem:
Make sure that https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation.md properly describes current approach for writing instrumentations, including new Instrumenter API and recent changes in muzzle interfaces.

iNikem:
That documentation should also explain how to manually implement methods from `InstrumentationModuleMuzzle` interface if muzzle build-time plugin is not used for some reason.

PR #4281, opened 2021-10-04
Remove last muzzle generate method from InstrumentationModule

iNikem:
It is still possible to write custom instrumentations without muzzle, but that should be properly documented as part of #4280


mateuszrzeszutek (approved):

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/InstrumentationModuleMuzzle.java
```diff
@@ -22,11 +24,35 @@
...
+   */
+  List<String> getMuzzleHelperClassNames();
+
+  /** See {@link #getMuzzleHelperClassNames()} */
+  static List<String> getMuzzleHelperClassNames(InstrumentationModule module) {
+    if (module instanceof InstrumentationModuleMuzzle) {
+      return ((InstrumentationModuleMuzzle) module).getMuzzleHelperClassNames();
+    } else {
+      return Collections.emptyList();
+    }
```
mateuszrzeszutek, line 56: > It is still possible to write custom instrumentations without muzzle

Hold on, if you have to implement `InstrumentationModuleMuzzle` to get helper class names how can you use helper classes without the muzzle plugin now? Add `implements InstrumentationModuleMuzzle` manually to your code?

iNikem: Yes, that is my idea, that if you write custom instrumentation and you don't use muzzle plugin, then you have to apply `InstrumentationModuleMuzzle` interface to your module manually. I will move it back to `extension-api` module.

I still think it is a good idea to have muzzle interface separate from `InstrumentationModule`, even if we have to apply the same stability guarantees to it. To not confuse end-users in the common scenario.

mateuszrzeszutek: > I still think it is a good idea to have muzzle interface separate from `InstrumentationModule`, even if we have to apply the same stability guarantees to it. To not confuse end-users in the common scenario.

💯 
I definitely agree with that.

trask: how about `InstrumentationModule.getExtraHelperClassNames()`? it's advanced and therefore potentially confusing to users which is downside, but on the upside it would allow a cleaner separation of muzzle functionality, and we could remove this behavior:

> If your `InstrumentationModule` subclass defines a method with exact same signature as a method
from `InstrumentationModuleMuzzle`, the muzzle compile plugin will not override your code:
muzzle will only generate those methods that do not have a custom implementation.

trask (approved):
> Should I do this in this PR or keep it simple and do in the follow up? I personally prefer the latter :) Smaller PRs FTW.

👍

PR #4282, opened 2021-10-04
Retry strict context check failures

laurit:
Relates to #3916
Oftentimes strict context check fails because a request processing thread hasn't managed to exit scope yet. This pr tries to deal with this by retrying failed strict context checks after sleeping a bit.


trask (approved):
nice, I like this approach

PR #4285, opened 2021-10-04
Add functionality to generate API changes

### LINKED ISSUE #4116: Track api diff reports for soonish-to-be-stable artifacts
Following the example from otel-java repo: https://github.com/open-telemetry/opentelemetry-java/pull/3183

Artifacts
* `instrumentation-api`
* `instrumentation-api-annotation-support`
* `instrumentation-api-caching` (hopefully tracked by instrumentation-api diffs)
* `javaagent-instrumentation-api`
* `extension-api`
* `muzzle` - even though this artifact isn't used directly, the classes that are used in code that is generated inside of InstrumentationModule implementations must remain backwards compatible

see: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4087/files#r707181702

iNikem:
Closes #4116

Almost verbatim copy from opentelemetry-java repo


conventions/src/main/kotlin/otel.japicmp-conventions.gradle.kts
```diff
@@ -0,0 +1,87 @@
...
+  id("me.champeau.gradle.japicmp")
+}
+
+/**
+ * The latest *released* version of the project. Evaluated lazily so the work is only done if necessary.
+ */
+val latestReleasedVersion: String by lazy {
+  // hack to find the current released version of the project
+  val temp: Configuration = configurations.create("tempConfig")
+  // pick the api, since it's always there.
```
jkwatson, line 15:
```suggestion
  // pick the agent, since it's always there.
```

docs/apidiffs/current_vs_latest/opentelemetry-instrumentation-api-annotation-support.txt
```diff
@@ -0,0 +1,2 @@
+Comparing source compatibility of  against 
+No changes.
```
jkwatson, line 2: does this mean that this has no non-internal public classes/methods?

iNikem: This means that it did not change since the last release. At least I read id this way.

jkwatson: Was this marked stable in a previous release, then?

iNikem: No, why?

jkwatson: The gradle code is supposed to generate the full API in the case where there is no previously published stable release. If it's not doing that, then there might be something amiss without how the porting over from the java repo happened.

iNikem: A, indeed, I changed the code to search for `-alpha` version, because we don't have stable version yet. I am not sure which version is more useful. What benefit do you get from the full API listing?

jkwatson: Having a baseline of the API can be useful when looking at the first official stable release, as a documentation of the actual API that is now stable, rather than just seeing "no changes".

iNikem: I see. And under which folder such full API listing should live? I suspect not under "current vs latest"?

trask (approved):
> I see. And under which folder such full API listing should live? I suspect not under "current vs latest"?

I'm good with whatever you decide

mateuszrzeszutek (approved):

docs/apidiffs/current_vs_latest/opentelemetry-instrumentation-api-annotation-support.txt
iNikem: I removed my support for unstable versions and also removed this diffs, as they don't make sense for now.

jkwatson: > I see. And under which folder such full API listing should live? I suspect not under "current vs latest"?

I think it would belong there if the 'current' has been marked stable in the current snapshot. It's a description of what you're preparing to release as a new stable API.

iNikem: > > I see. And under which folder such full API listing should live? I suspect not under "current vs latest"?
> 
> I think it would belong there if the 'current' has been marked stable in the current snapshot. It's a description of what you're preparing to release as a new stable API.

Got it, makes sense. So we will add that API list when we are ready to declare it stable.

PR #4287, opened 2021-10-04
Split NetAttributesExtractor into NetClientAttributesExtractor and NetServerAttributesExtractor

trask:
Alternative to #4278 that implements parallel hierarchies:

* NetClientAttributesExtractor
* InetSocketAddressNetClientAttributesExtractor

and

* NetClientAttributesExtractor
* InetSocketAddressNetClientAttributesExtractor

Note: `peer.service` only applies to client spans ([afaict](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-remote-service-attributes)), so `PeerServiceAttributesExtractor` is not split, and only applies to `NetClientAttributesExtractor`.


mateuszrzeszutek:
Should we go ahead with the `NetAttributesServerExtractor`/`NetAttributesClientExtractor` idea? It matches the recently changes `flavor()` methods in HTTP extractors.

trask:
> Should we go ahead with the `NetAttributesServerExtractor`/`NetAttributesClientExtractor` idea? It matches the recently changes `flavor()` methods in HTTP extractors.

This PR is updated now with this approach 👍

instrumentation/jedis/jedis-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v1_4/JedisNetAttributesExtractor.java
```diff
@@ -5,10 +5,10 @@
 
 package io.opentelemetry.javaagent.instrumentation.jedis.v1_4;
 
-import io.opentelemetry.instrumentation.api.instrumenter.net.NetAttributesExtractor;
+import io.opentelemetry.instrumentation.api.instrumenter.net.NetAttributesServerExtractor;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
-final class JedisNetAttributesExtractor extends NetAttributesExtractor<JedisRequest, Void> {
+final class JedisNetAttributesExtractor extends NetAttributesServerExtractor<JedisRequest, Void> {
```
mateuszrzeszutek, line 11: Should we use a client extractor here? It's a redis client library after all (even if the net info is available on start)

trask:
> > please see [this commit](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4287/commits/298f306f1595f97202c6062755c46a434c15d131) for attempt to expose `peerService` to library instrumentation based on discussion in #4235
> 
> I also added similar for Dubbo

and gRPC

instrumentation/apache-dubbo-2.7/library/src/main/java/io/opentelemetry/instrumentation/apachedubbo/v2_7/DubboTracingBuilder.java
```diff
@@ -42,19 +53,37 @@ public DubboTracingBuilder addAttributesExtractor(
...
+          AttributesExtractor.constant(SemanticAttributes.PEER_SERVICE, peerService));
+    } else {
+      clientInstrumenterBuilder.addAttributesExtractor(
+          PeerServiceAttributesExtractor.create(netClientAttributesExtractor));
+    }
 
     return new DubboTracing(
-        builder.newServerInstrumenter(new DubboHeadersGetter()),
-        builder.newClientInstrumenter(new DubboHeadersSetter()));
+        clientInstrumenterBuilder.newServerInstrumenter(new DubboHeadersGetter()),
```
mateuszrzeszutek, line 86:
```suggestion
        serverInstrumenterBuilder.newServerInstrumenter(new DubboHeadersGetter()),
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
```diff
@@ -51,4 +51,9 @@ protected abstract void onEnd(
       attributes.put(key, value);
     }
   }
+
+  public static <REQUEST, RESPONSE, T> AttributesExtractor<REQUEST, RESPONSE> constant(
```
mateuszrzeszutek, line 55:
```suggestion
  /**
   * Returns an {@link AttributesExtractor} implementation that always extracts the provided constant value.
   */
  public static <REQUEST, RESPONSE, T> AttributesExtractor<REQUEST, RESPONSE> constant(
```

mateuszrzeszutek (approved):

PR #4290, opened 2021-10-04
Add comment about keeping grpc version in sync

trask:


dependencyManagement/build.gradle.kts
```diff
@@ -35,6 +37,7 @@ rootProject.extra["caffeine3Version"] = "3.0.4"
 val DEPENDENCY_BOMS = listOf(
   "com.fasterxml.jackson:jackson-bom:2.12.3",
   "com.google.guava:guava-bom:30.1.1-jre",
+  "io.grpc:grpc-bom:${grpcVersion}",
```
anuraaga, line 40: I wouldn't expect this to work well with the dependencies declared in our `grpc` instrumentation modules - can you verify with `:dependencies` and `:dependencies -PtestLatestDeps=true` that they are as expected? If not, we could just list the `grpc-netty-shaded` in `DEPENDENCIES` below

dependencyManagement/build.gradle.kts
```diff
@@ -11,7 +11,9 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
+// IMPORTANT when updating otelVersion, need to also update grpcVersion
```
anuraaga, line 14: The main requirement is alignment - so actually if grpc-bom were to work (it can definitely be made to work but I'm not sure it currently does without some tweaks in grpc modules, see other comment) then the versions of gRPC and otel become decoupled in our build and gRPC versions would be aligned, possibly on a different version than specified by OTel. The policy would then be always use the latest version of dependencies when releasing, i.e., there wouldn't be anything intrinsically different about grpc vs jackson.

dependencyManagement/build.gradle.kts
anuraaga: One example of the needed tweaks is this

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/netty/netty-4.0/javaagent/build.gradle.kts#L57

We do use netty-bom, mostly to avoid having dozens of netty versions loaded into our build which hasn't been as problematic for grpc. It requires this sort of syntax to make sure the BOM doesn't cause the wrong dependency in the instrumentation.

trask:
> we could just list the grpc-netty-shaded in DEPENDENCIES below

thx, went with this approach

PR #4300, opened 2021-10-05
Move Trie back to javaagent-tooling

mateuszrzeszutek:
As discussed in the SIG this morning - the trie class should not be exposed in any of our `*-api` modules, it should be kept in some internal javaagent module.

### DIFF
--- a/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/internal/InstrumentedTaskClasses.java
+++ b/javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/internal/InstrumentedTaskClasses.java
@@ -41,8 +41,8 @@ public final class InstrumentedTaskClasses {
   private static volatile Predicate<String> ignoredTaskClassesPredicate;
 
   /**
-   * Sets the configured ignored tasks predicate (that uses a trie to find the common prefix). This
-   * method is called internally from the agent classloader.
+   * Sets the configured ignored tasks predicate. This method is called internally from the agent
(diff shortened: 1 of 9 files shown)


javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/internal/InstrumentedTaskClasses.java
```diff
@@ -38,18 +38,18 @@ protected Boolean computeValue(Class<?> taskClass) {
         }
       };
 
-  private static volatile Trie<Boolean> ignoredTaskClasses;
+  private static volatile Predicate<String> ignoredTaskClassesPredicate;
 
   /**
-   * Sets the configured ignored tasks trie. This method is called internally from the agent
-   * classloader.
+   * Sets the configured ignored tasks predicate (that uses a trie to find the common prefix). This
```
iNikem, line 44: Why is it important to mention trie here?

iNikem (approved):

javaagent-instrumentation-api/src/main/java/io/opentelemetry/javaagent/instrumentation/api/internal/InstrumentedTaskClasses.java
mateuszrzeszutek: It's not, now that I think of it. Removed 👍

trask (approved):

trask:
closing/re-opening to trigger GHA

PR #4302, opened 2021-10-05
Add supported way to add additional helper classes to instrumentation without interfering with muzzle

iNikem:


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/instrumentation/InstrumentationModule.java
```diff
@@ -134,4 +135,18 @@ public void registerHelperResources(HelperResourceBuilder helperResourceBuilder)
...
+  /**
+   * Returns a list of additional instrumentation helper classes, which are not automatically
+   * detected during compile time.
+   *
+   * <p>If your instrumentation module does not apply and you see warnings about missing classes in
+   * the logs, you may need to override this method and provide fully qualified classes names of
+   * helper classes that your instrumentation uses.
+   *
+   * <p>These helper classes will be injected into the application classloader.
+   */
```
mateuszrzeszutek, line 148: Can you mention that those classes are injected after the muzzle-detected ones?

mateuszrzeszutek (approved):

trask (approved):

mateuszrzeszutek (approved):

PR #4310, opened 2021-10-06
Fix a bug in the field backed VirtualField implementation

mateuszrzeszutek:
When refactoring these classes I've noticed that generated fields (and getters and setters) are not named correctly; thus making it impossible to actually inject more than one field. It turned out to be a bug: when more than one `VirtualField` was used for a single class, only the first field would "really" be injected; the rest of them used the cache-based fallback implementation.

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/field/GeneratedVirtualFieldNames.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/field/GeneratedVirtualFieldNames.java
@@ -38,15 +38,15 @@ final class GeneratedVirtualFieldNames {
         + Utils.convertToInnerClassName(fieldTypeName);
   }
 
-  static String getRealFieldName(String typeName) {
-    return "__opentelemetryVirtualField$" + Utils.convertToInnerClassName(typeName);
+  static String getRealFieldName(String fieldTypeName) {
+    return "__opentelemetryVirtualField$" + Utils.convertToInnerClassName(fieldTypeName);
   }
 
-  static String getRealGetterName(String typeName) {
-    return "get" + getRealFieldName(typeName);
+  static String getRealGetterName(String fieldTypeName) {
+    return "get" + getRealFieldName(fieldTypeName);
   }
 
-  static String getRealSetterName(String typeName) {
-    return "set" + getRealFieldName(typeName);
+  static String getRealSetterName(String fieldTypeName) {
+    return "set" + getRealFieldName(fieldTypeName);
   }
 }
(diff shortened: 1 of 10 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/field/GeneratedVirtualFieldNames.java
```diff
@@ -38,15 +38,15 @@ static String getFieldAccessorInterfaceName(String typeName, String fieldTypeNam
         + Utils.convertToInnerClassName(fieldTypeName);
   }
 
-  static String getRealFieldName(String typeName) {
-    return "__opentelemetryVirtualField$" + Utils.convertToInnerClassName(typeName);
+  static String getRealFieldName(String fieldTypeName) {
```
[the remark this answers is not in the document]
laurit: What if I use different context classes like `VirtualField.find(Interface1.class, String.class)` and `VirtualField.find(Interface2.class, String.class)` and have `class Foo implements Interface1, Interface2`. Would the use the same field?

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/field/GeneratedVirtualFieldNames.java
```diff
@@ -38,15 +38,18 @@ static String getFieldAccessorInterfaceName(String typeName, String fieldTypeNam
         + Utils.convertToInnerClassName(fieldTypeName);
   }
 
-  static String getRealFieldName(String typeName) {
-    return "__opentelemetryVirtualField$" + Utils.convertToInnerClassName(typeName);
+  static String getRealFieldName(String typeName, String fieldTypeName) {
+    return "__opentelemetryVirtualField$"
```
laurit, line 42: Imo this isn't completely safe. `Utils.convertToInnerClassName` is used to replace `.` with `$` because `.` is not valid in field name. But `.` is not the only invalid character so are  `; [ /`. Out of these the only one we should be concerned about is `[`. I would guess that using `VirtualField.find(A.class, Object[].class)` will fail. Though even if this is fixed things will get a bit more complicated when you consider that since 1.6 (if I remember correctly) jvm spec was changed to be much more liberal about what it considers a valid type or field name. So if the field is injected into a class with version 1.5 or earlier it might fail because of other character too. Luckily these type names are extracted from calls to `VirtualField.find` where you won't be using generated classes with crazy names that contains newlines and stuff like that. I guess the chances of any of this happening aren't too high.

mateuszrzeszutek: > Luckily these type names are extracted from calls to `VirtualField.find` where you won't be using generated classes with crazy names that contains newlines and stuff like that. I guess the chances of any of this happening aren't too high.

Yep, except for the array scenario there it shouldn't be possible to get any weird characters in.

> Out of these the only one we should be concerned about is `[`. I would guess that using `VirtualField.find(A.class, Object[].class)` will fail.

We should probably also forbid using an array type as the "owner" of a `VirtualField`.  I'll fix that in a separate PR.

trask:
@mateuszrzeszutek @laurit 👍

PR #4320, opened 2021-10-07
Test captured HTTP headers - HTTP server tests, part 1

mateuszrzeszutek:
I've added a new test case in `HttpServerTest` just for testing the captured headers; this test is enabled in some Spring boot test (javaagent instrumentation) and Restlet (library instrumentation). I'll enable it in all other tests in the next PR.

### DIFF
--- a/instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.0/javaagent/build.gradle.kts
+++ b/instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.0/javaagent/build.gradle.kts
@@ -40,7 +40,7 @@ dependencies {
   testLibrary("org.jboss.resteasy:resteasy-undertow:3.0.4.Final") {
     exclude("org.jboss.resteasy", "resteasy-client")
   }
-  testLibrary("io.undertow:undertow-servlet:1.0.0.Final")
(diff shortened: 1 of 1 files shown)


instrumentation/jaxrs/jaxrs-2.0/jaxrs-2.0-resteasy-3.0/javaagent/build.gradle.kts
```diff
@@ -40,7 +40,7 @@ dependencies {
   testLibrary("org.jboss.resteasy:resteasy-undertow:3.0.4.Final") {
     exclude("org.jboss.resteasy", "resteasy-client")
   }
-  testLibrary("io.undertow:undertow-servlet:1.0.0.Final")
+  testLibrary("io.undertow:undertow-servlet:1.4.28.Final")
```
trask, line 43: > Undertow 1.0 had a bug where it thrown NPE on getHeaders()

😱

mateuszrzeszutek:
```java
    public Collection<String> getHeaders(String name) {
        return new ArrayList(this.exchange.getResponseHeaders().get(name));
    }
```

Guess what happens when `this.exchange.getResponseHeaders()` returns null because there are no headers in the response yet 🙈

PR #4327, opened 2021-10-08
Remove deprecated ServerSpanNaming.updateSource method

iNikem:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/servlet/ServerSpanNaming.java
```diff
@@ -125,7 +112,9 @@ private boolean isBetterName(String name) {
     // filter that is called
     FILTER(2, /* useFirst= */ false),
     SERVLET(3),
-    CONTROLLER(4);
+    CONTROLLER(4),
+    // JaxRS allows for nested paths and we want to select the longest one
+    JAXRS(5, false);
```
laurit, line 117: Instead of `JAXRS` perhaps we should try to come up with a generic name so that it could be used by other instrumentations.
As a sidenote all of this looks fine but really all usages of `serverSpan.updateName` should also have used `updateSource`. I wonder whether there are remaining usages that do something weird and could still benefit from having `updateSource`.

mateuszrzeszutek (approved):

PR #4339, opened 2021-10-10
Fix tomcat async spans

### LINKED ISSUE #4334: Duration of request if wrong if method returns Mono (spring webflux controller)
**Describe the bug**
When Spring WebFlux controller returns Mono, the span for request ends immediately, but deferred/delayed mono still runs.

**Steps to reproduce**
Create a simple weblux app, here's controller code, full example: 
[mono-duration-bug.zip](https://github.com/open-telemetry/opentelemetry-java-instrumentation/files/7314615/mono-duration-bug.zip)

```java

@RestController
@RequestMapping("/hello")
public class HelloController {

    @GetMapping("/messages")
    public Mono<ResponseEntity<String>> get() {
        return Mono.delay(Duration.ofMillis(100)).then(Mono.just(ResponseEntity.ok("Hello")));
    }
}
```

duration of hello/messages is longer - span ends after all deferred mono ends.

**What did you see instead?**
image

opentelemetry-javaagent - version: 1.6.2

**Environment**
Compiler:  AdoptOpenJDK 16.0.1
OS: win10

trask:
Resolves #4334

Adds test for all server instrumentation to verify that server spans end after all of their child spans (with a couple of exceptions).


anuraaga:
@trask Just curious does this have anything to do with the strict context stuff we talked about?

instrumentation/vertx-web-3.0/testing/src/main/java/server/AbstractVertxWebServer.java
```diff
@@ -39,60 +39,46 @@ public Router buildRouter() {
...
-                HttpServerTest.controller(
-                    QUERY_PARAM,
-                    () -> {
-                      end(
-                          ctx.response().setStatusCode(QUERY_PARAM.getStatus()),
-                          ctx.request().query());
-                      return null;
-                    }));
+            ctx -> {
+              HttpServerTest.controller(QUERY_PARAM, () -> null);
```
mateuszrzeszutek, line 62: WDYT about removing the second argument of `controller`? The passed lambda seems to be unneeded now

instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatHelper.java
```diff
@@ -32,7 +33,9 @@ public boolean shouldStart(Context parentContext, Request request) {
   }
 
   public Context start(Context parentContext, Request request) {
-    return instrumenter.start(parentContext, request);
+    Context context = instrumenter.start(parentContext, request);
+    request.setAttribute(HttpServerTracer.CONTEXT_ATTRIBUTE, context);
```
mateuszrzeszutek, line 37: Is this the actual fix?

trask: this would be nice if it leads to unifying app server behavior a bit, I created #4347 to revisit the idea

mateuszrzeszutek (approved):

PR #4340, opened 2021-10-11
Update to OTel 1.7

anuraaga:


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/LambdaUtils.java
```diff
@@ -5,17 +5,25 @@
...
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
 
 final class LambdaUtils {
 
   static void forceFlush(OpenTelemetrySdk openTelemetrySdk, long flushTimeout, TimeUnit unit) {
     CompletableResultCode traceFlush = openTelemetrySdk.getSdkTracerProvider().forceFlush();
-    CompletableResultCode metricsFlush = IntervalMetricReader.forceFlushGlobal();
+    MeterProvider meterProvider = GlobalMeterProvider.get();
```
iNikem, line 20: Is it expected that we pass here an instance of `OpenTelemetrySdk` but `MeterProvider` we take from the global? Why is it not possible to get the latter from the former?

anuraaga: Until metrics is stable, we haven't associated it with the `opentelemetry-sdk` API at all, which is stable. So we have to access metrics separately - the model for autoconfigure is it sets up the SDK object but only creates a global meter provider, as a "side channel" into the alpha API. Eventually, it'll be available through the SDK object though

PR #4349, opened 2021-10-12
Less surprising behavior

### LINKED ISSUE #2445: Consistent package for javaagent test artifacts
As re-discovered in #2440, some tests rely on their custom classes to be instrumented, but `io.opentelemetry.javaagent` package is excluded from instrumentation in `GlobalIgnoresMatcher` (to reduce the javaagent startup overhead).

Need to make a note about this in https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation.md#writing-java-agent-instrumentation.

But also (in addition to doc), people would be less likely to make this mistake if all javaagent tests used same package (e.g. `io.opentelemetry.test`) that is outside of `io.opentelemetry.javaagent`, so that when they copy/review how other javaagent tests work, they will avoid the problem.

trask:
Only ignore instrumenting `io.opentelemetry.javaagent...` classes in bootstrap class loader.

This way test classes under `io.opentelemetry.javaagent...` classes will be instrumented.

Resolves #2445

### DIFF
--- a/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ignore/GlobalIgnoredTypesConfigurer.java
+++ b/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ignore/GlobalIgnoredTypesConfigurer.java
@@ -55,10 +55,14 @@ public class GlobalIgnoredTypesConfigurer implements IgnoredTypesConfigurer {
     // clojure
     builder.ignoreClass("clojure.").ignoreClass("$fn__");
 
-    // this is to exclude javaagent helper classes that are injected into other class loaders
-    // "io.opentelemetry.javaagent" classes in the bootstrap class loader are excluded elsewhere
-    // and all classes in the AgentClassLoader are excluded elsewhere
+    // all classes in the AgentClassLoader are ignored separately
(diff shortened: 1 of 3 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/ignore/GlobalIgnoredTypesConfigurer.java
```diff
@@ -55,7 +55,10 @@ private static void configureIgnoredTypes(IgnoredTypesBuilder builder) {
     // clojure
     builder.ignoreClass("clojure.").ignoreClass("$fn__");
 
-    builder.ignoreClass("io.opentelemetry.javaagent.");
+    // this is to exclude javaagent helper classes that are injected into other class loaders
+    // "io.opentelemetry.javaagent" classes in the bootstrap class loader are excluded elsewhere
+    // and all classes in the AgentClassLoader are excluded elsewhere
+    builder.ignoreClass("io.opentelemetry.javaagent.shaded.");
```
iNikem, line 61: Why is this here and not in `AgentInstaller`? I don't understand by what criterium we decide between these two locations.

PR #4359, opened 2021-10-12
Convert akka-http-10.0 to instrumenter api

jack-berg:
Contributing to Issue #2713.

### DIFF
--- a/instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/HttpHeaderSetter.java
+++ b/instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/HttpHeaderSetter.java
@@ -5,10 +5,9 @@
 
 package io.opentelemetry.javaagent.instrumentation.akkahttp.client;
 
-import static io.opentelemetry.javaagent.instrumentation.api.Java8BytecodeBridge.currentContext;
-
 import akka.http.javadsl.model.headers.RawHeader;
 import akka.http.scaladsl.model.HttpRequest;
+import io.opentelemetry.context.Context;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.context.propagation.TextMapSetter;
 


instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/server/AkkaHttpServerInstrumentationModule.java
```diff
@@ -66,28 +72,32 @@ public AsyncWrapper(
...
-                    tracer().end(ctx, response);
+                    instrumenter().end(context, request, response, null);
                     return response;
                   }
                 },
                 new AbstractFunction1<Throwable, Throwable>() {
                   @Override
                   public Throwable apply(Throwable t) {
-                    tracer().endExceptionally(ctx, t);
+                    instrumenter().end(context, request, errorResponse(), t);
```
jack-berg, line 94: This is a bit strange. Here and on line 56, I needed to return a default error response so that the tests associated the attributes expected by the test with the span.

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/AkkaHttpClientSingletons.java
```diff
@@ -0,0 +1,46 @@
...
+    AkkaHttpClientAttributesExtractor httpAttributesExtractor =
+        new AkkaHttpClientAttributesExtractor();
+    INSTRUMENTER =
+        Instrumenter.<HttpRequest, HttpResponse>newBuilder(
+                GlobalOpenTelemetry.get(),
+                AkkaHttpUtil.instrumentationName(),
+                HttpSpanNameExtractor.create(httpAttributesExtractor))
+            .setSpanStatusExtractor(HttpSpanStatusExtractor.create(httpAttributesExtractor))
+            .addAttributesExtractor(httpAttributesExtractor)
+            .addAttributesExtractor(new AkkaHttpNetAttributesExtractor())
```
mateuszrzeszutek, line 33:
```suggestion
            .addAttributesExtractor(new AkkaHttpNetAttributesExtractor())
            .addRequestMetrics(HttpClientMetrics.get())
```

mateuszrzeszutek: Also, can you add `PeerServiceAttributesExtractor` to the client instrumenter too?

instrumentation/akka-http-10.0/javaagent/src/test/groovy/AkkaHttpClientInstrumentationTest.groovy
```diff
@@ -75,17 +72,7 @@ class AkkaHttpClientInstrumentationTest extends HttpClientTest<HttpRequest> impl
...
-          hasNoParent()
-          name "HTTP request"
-          kind CLIENT
-          status ERROR
-          errorEvent(NullPointerException, e.getMessage())
-        }
-      }
-    }
+    thrown NullPointerException
+    assertTraces(0) {}
```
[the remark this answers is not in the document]
mateuszrzeszutek: That's weird that we even had this test case. I'm in favor of removing it completely.

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/server/AkkaHttpServerSingletons.java
```diff
@@ -0,0 +1,39 @@
...
+    AkkaHttpServerAttributesExtractor httpAttributesExtractor =
+        new AkkaHttpServerAttributesExtractor();
+    INSTRUMENTER =
+        Instrumenter.<HttpRequest, HttpResponse>newBuilder(
+                GlobalOpenTelemetry.get(),
+                AkkaHttpUtil.instrumentationName(),
+                unused -> "akka.request")
+            .setSpanStatusExtractor(HttpSpanStatusExtractor.create(httpAttributesExtractor))
+            .addAttributesExtractor(httpAttributesExtractor)
+            .newServerInstrumenter(new AkkaHttpServerHeaders());
```
mateuszrzeszutek, line 29:
```suggestion
            .addRequestMetrics(HttpServerMetrics.get())
            .newServerInstrumenter(new AkkaHttpServerHeaders());
```

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/server/AkkaHttpServerInstrumentationModule.java
mateuszrzeszutek: Hmm, which test cases failed? Was it because of status code?

jack-berg: The `HttpServerTest."test exception"` cases fail for `AkkaHttpServerInstrumentationTestSync` and `AkkaHttpServerInstrumentationTestAsync` if these `errorResponse()` are missing.

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/HttpExtClientInstrumentation.java
```diff
@@ -56,15 +57,14 @@ public static void methodEnter(
       with way of continuing to reusing it.
        */
       Context parentContext = currentContext();
-      if (!tracer().shouldStartSpan(parentContext)) {
+      if (request == null || !instrumenter().shouldStart(parentContext, request)) {
```
trask, line 60: if null `request` is not valid input for the library method anyways, we typically (for better or worse?) don't validate the input and let the advice method throw NPE (which is then suppressed by `suppress = Throwable.class`)

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/HttpExtClientInstrumentation.java
```diff
@@ -56,15 +57,14 @@ public static void methodEnter(
...
       }
 
-      // Request is immutable, so we have to assign new value once we update headers
-      AkkaHttpHeaders headers = new AkkaHttpHeaders(request);
-      context = tracer().startSpan(parentContext, request, headers);
+      context = instrumenter().start(parentContext, request);
       scope = context.makeCurrent();
-      request = headers.getRequest();
+      // Request is immutable, so we have to assign new value once we update headers
+      request = setter().inject(request);
```
trask, line 67: can this be done before calling `start`? just so its the same request instance passed to both `start` and `end`

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpUtil.java
```diff
@@ -0,0 +1,49 @@
...
+        .map(httpHeader -> Collections.singletonList(httpHeader.value()))
+        .orElse(Collections.emptyList());
+  }
+
+  public static List<String> responseHeader(HttpResponse httpResponse, String name) {
+    return httpResponse
+        .getHeader(name)
+        .map(httpHeader -> Collections.singletonList(httpHeader.value()))
+        .orElse(Collections.emptyList());
+  }
```
trask, line 34: i'd suggest not using streams api here for performance sake

trask (approved):
thx!

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/server/AkkaHttpServerInstrumentationModule.java
trask: can you open an issue for this? probably we should revisit

the old tracer had this code which has its own problems:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/b0430f2b79d2ec8c1805c0889bdbc06e5e5f0af1/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpServerTracer.java#L137-L138

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpUtil.java
mateuszrzeszutek: It's actually `Optional`, not streams

mateuszrzeszutek (approved):

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/server/AkkaHttpServerInstrumentationModule.java
laurit: Servlet instrumenter has the same issue https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/b0430f2b79d2ec8c1805c0889bdbc06e5e5f0af1/instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/ServletHttpAttributesExtractor.java#L79-L93
If the actual response status is set outside the scope of our integration we'll need to widen our integration so that we can capture the actual response code or predict what it is going to be.
Just using `errorResponse()` would be incorrect in case it is possible to send response status and fail with exception after that. Servlet instrumentation uses `isResponseCommitted` to make sure that we don't replace the status that was already sent to client with `500`.

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/HttpExtClientInstrumentation.java
trask: oh, you are so right! 👍

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/client/HttpHeaderSetter.java
```diff
@@ -22,4 +29,26 @@ public void set(AkkaHttpHeaders carrier, String key, String value) {
           (HttpRequest) request.removeHeader(key).addHeader(RawHeader.create(key, value)));
     }
   }
+
+  public HttpRequest inject(HttpRequest original) {
+    AkkaHttpHeaders carrier = new AkkaHttpHeaders(original);
+    contextPropagators.getTextMapPropagator().inject(currentContext(), carrier, this);
```
trask, line 35: `Java8BytecodeBridge` is only needed inside of Advice, check out https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation-module.md

> You may have noticed that the example above does not use Context.current(), but a Java8BytecodeBridge method. This is intentional: if you are instrumenting a pre-Java 8 library, then inlining Java 8 default method calls (or static methods in an interface) into that library will result in a java.lang.VerifyError at runtime, since Java 8 default method invocations are not legal in Java 7 (and prior) bytecode. Because OpenTelemetry API has many common default/static interface methods (e.g. Span.current()), the javaagent-instrumentation-api artifact has a class Java8BytecodeBridge which provides static methods for accessing these default methods from advice. In fact, we suggest avoiding Java 8 language features in advice classes at all - sometimes you don't know what bytecode version is used by the instrumented class.

```suggestion
    contextPropagators.getTextMapPropagator().inject(Context.current(), carrier, this);
```

instrumentation/akka-http-10.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/akkahttp/server/AkkaHttpServerInstrumentationModule.java
trask: thx! i'm in favor of merging this as-is 👍

PR #4361, opened 2021-10-12
Update version in manual instrumentation (from 1.0.0 -> 1.7.0)

breedx-splk:
Not sure how frequently we want to update the docs like this, but it seemed that quite some time had passed. Do we have a strategy for updating these over time?


docs/manual-instrumentation.md
```diff
@@ -28,12 +28,12 @@ use the `@WithSpan` annotation, also include the `opentelemetry-extension-annota
     <dependency>
       <groupId>io.opentelemetry</groupId>
       <artifactId>opentelemetry-api</artifactId>
-      <version>1.0.0</version>
+      <version>1.7.0</version>
```
jkwatson, line 31: until we've updated to the 1.7.0 javaagent, we should keep this at 1.6.0, I think.

PR #4363, opened 2021-10-13
Convert AttributesExtractor to interface

### LINKED ISSUE #3131: Should AttributesExtractor be an interface?
Only bringing up because all other Extractors are interfaces, and Intellij has had to correct me on `extends` and `implements` a few times because it took me some time to notice the difference.

Related, is the `set` method in `AttributesExtractor` needed? It looks like `AttributesBuilder` is no-op already on `null` value, or is that subject to change?

iNikem:
Fixes #3131

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
@@ -24,18 +24,18 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  * @see HttpClientAttributesExtractor
  * @see NetServerAttributesExtractor
  */
-public abstract class AttributesExtractor<REQUEST, RESPONSE> {
+public interface AttributesExtractor<REQUEST, RESPONSE> {
   /**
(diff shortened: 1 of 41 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
```diff
@@ -45,8 +45,7 @@ protected abstract void onEnd(
    * Sets the {@code value} with the given {@code key} to the {@link AttributesBuilder} if {@code
    * value} is not {@code null}.
    */
-  protected static <T> void set(
-      AttributesBuilder attributes, AttributeKey<T> key, @Nullable T value) {
+  default <T> void set(AttributesBuilder attributes, AttributeKey<T> key, @Nullable T value) {
```
anuraaga, line 48: Since set isn't really part of the contract of AttributesExtractor, we should keep them off the method even as defaults. I'm thinking the use case of "fill attribute if available" is so common that it's fine to add the `@Nullable` annotation in the API and consider it "correct usage". So easiest is probably to rely on the implementation detail that `put` can be called directly, and I can work on that in the SDK to formalize it.

Can be another PR but we should do it

PR #4364, opened 2021-10-13
Mongo4 strict context check

laurit:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3916


mateuszrzeszutek (approved):

instrumentation/mongo/mongo-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/mongo/v4_0/AsyncWorkManagerInstrumentation.java
```diff
@@ -0,0 +1,51 @@
...
+        named("initUnlessClosed"),
+        AsyncWorkManagerInstrumentation.class.getName() + "$DisablePropagationAdvice");
+  }
+
+  @SuppressWarnings("unused")
+  public static class DisablePropagationAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static Scope onEnter() {
+      if (Java8BytecodeBridge.currentContext() != Java8BytecodeBridge.rootContext()) {
+        return Java8BytecodeBridge.rootContext().makeCurrent();
```
trask, line 39:
```suggestion
        // Prevent context from leaking by running this method under root context.
        // Root context is not propagated by executor instrumentation.
        return Java8BytecodeBridge.rootContext().makeCurrent();
```

PR #4365, opened 2021-10-13
PlayWS strict context check

laurit:
Part of #3916


instrumentation/play-ws/play-ws-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/playws/AbstractBootstrapInstrumentation.java
```diff
@@ -0,0 +1,49 @@
...
+        named("initAndRegister"),
+        AbstractBootstrapInstrumentation.class.getName() + "$DisablePropagationAdvice");
+  }
+
+  @SuppressWarnings("unused")
+  public static class DisablePropagationAdvice {
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static Scope onEnter() {
+      if (Java8BytecodeBridge.currentContext() != Java8BytecodeBridge.rootContext()) {
+        return Java8BytecodeBridge.rootContext().makeCurrent();
```
iNikem, line 37: Why?

mateuszrzeszutek: I believe it forcefully removes the current context for while the method executes; this way preventing the context leaking into a (potentially long-running) task that's started there.

PR #4366, opened 2021-10-13
AWS 1.11 strict context check

laurit:
Part of #3916
This is purely a test issue `shouldSetScopeIfClientSpanNotPresent` didn't close scope. `shouldNotSetScopeAndNotFailIfClientSpanAlreadyPresent` started failing when the scope was closed, this test didn't pass when it was run on its own.

### DIFF
--- a/instrumentation/aws-sdk/aws-sdk-1.11/javaagent-unit-tests/src/test/java/TracingRequestHandlerTest.java
+++ b/instrumentation/aws-sdk/aws-sdk-1.11/javaagent-unit-tests/src/test/java/TracingRequestHandlerTest.java
@@ -25,7 +25,13 @@ class TracingRequestHandlerTest {
   }
 
   private static Request<SendMessageRequest> request() {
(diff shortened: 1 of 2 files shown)


instrumentation/aws-sdk/aws-sdk-1.11/javaagent-unit-tests/src/test/java/TracingRequestHandlerTest.java
```diff
@@ -61,5 +67,8 @@ void shouldSetScopeIfClientSpanNotPresent() {
     // then - no exception and scope not set
     assertThat(request.getHandlerContext(TracingRequestHandler.SCOPE)).isNotNull();
     underTest.afterResponse(request, response(request));
```
mateuszrzeszutek, line 69: Hmm, why doesn't `afterResponse` close the scope on its own? Is it really okay that it's only closed on error? CC @anuraaga

trask (approved):

PR #4367, opened 2021-10-13
Ratpack strict context check

laurit:
Part of #3916


instrumentation/ratpack-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/ratpack/ActionWrapper.java
```diff
@@ -35,7 +35,11 @@ public void execute(T t) throws Exception {
     if (delegate instanceof ActionWrapper) {
       return delegate;
     }
+    Context context = Context.current();
+    if (context == Context.root()) {
```
iNikem, line 39: Can you add some comment to explain why wrapping actions depend on the context being root or not?

mateuszrzeszutek (approved):

laurit:
This pr depends on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4380 because ratpack tests are not spock tests and are flaky because of strict context checks.

PR #4368, opened 2021-10-13
Netty 38 server instrumenter

mateuszrzeszutek:


trask (approved):

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerResponseTracingHandler.java
```diff
@@ -19,24 +18,27 @@
...
+
     try (Scope ignored = context.makeCurrent()) {
       ctx.sendDownstream(msg);
-      tracer().end(context, (HttpResponse) msg.getMessage());
+      instrumenter().end(context, request, response, null);
     } catch (Throwable throwable) {
-      tracer().endExceptionally(context, throwable);
+      instrumenter().end(context, request, response, throwable);
       throw throwable;
     }
```
iNikem, line 43: Replace with `finally`?

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/NettyServerErrorHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.instrumenter.ErrorCauseExtractor;
+
+public final class NettyServerErrorHandler {
+
+  // copied from BaseTracer#onException()
+  public static void onError(Context context, Throwable error) {
```
iNikem, line 16: Is this needed? Shouldn't `Instrumenter` take care of that?

laurit: Alternatively could add a context key that can hold this exception and extract it for calling end. Idk whether it would be nicer.

mateuszrzeszutek: That'd be a pretty good idea. I'll do that for all netty versions in a separate PR (they all handle exceptions this way currently)

instrumentation/netty/netty-3.8/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v3_8/server/HttpServerResponseTracingHandler.java
iNikem: Oi, I got confused :) Current version is good.

PR #4369, opened 2021-10-13
Convert HibernateTracer to Instrumenter

iNikem:
This is one very strange instrumentation. Starting from the fact that it does not have any clear `Request` object.


instrumentation/hibernate/hibernate-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/HibernateSingletons.java
```diff
@@ -0,0 +1,25 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+
+public class HibernateSingletons {
+
+  private static final Instrumenter<String, Void> INSTANCE;
+
+  static {
+    INSTANCE =
+        Instrumenter.<String, Void>newBuilder(
+                GlobalOpenTelemetry.get(), "io.opentelemetry.hibernate-common", s -> s)
```
mateuszrzeszutek, line 18: Nit: WDYT about converting this to `HibernateInstrumenterFactory` with a factory method accepting the instrumentation name? Every hibernate module would pass its own name with the version

PR #4374, opened 2021-10-13
Prototype Log4j2 Appender

jack-berg:
This is meant to prompt discussion. Proves out what an `OpenTelemetryAppender` might look like for log4j2. The idea behind this appender is that it delivers log records to a opentelemetry `LogSink`, where they can be batched and shipped out of process via an exporter, such as the recently added OTLP grpc and http log exporters. 

One of the main problems I had to address was that log appenders are instantiated before the opentelemetry log sdk is available. To get around this, each `OpenTelemetryAppender` instance is statically registered after initialization. Once the otel log sdk is available, it is passed to each `OpnTelemetryAppender` instance via a static method, upon which log records start flowing.

Learnings so far:
- LogRecord trace id and span id need to be nullable. And potentially other properties too. The otlp grpc and http exporters currently fail to serialize if span id and trace id are left as their default `""` values.
- We need a `InMemoryLoggingExporter` for testing. Mocked it out in this code, and will contribute back to `opentelemetry-java`. 
- We need a `SimpleLogProcessor` for testing. Mocked it out in this code, and will contribute back to `opentelemetry-java`. 
- `LogSinkSdkProvider.builder()` is not public but needs to be for it to be useful. I'll make it so.
- `LogSinkSdkProvider` should be immutable after its created. Log processors should be registered with the builder, not with the instance. I'll make it so. 
- Thought needs to be put into how resource and instrumentation library get attached to the log records. Right now and producer to `LogSink` is responsible for creating `LogRecord`s with resource and instrumenation library attached. This means that components like the `OpenTelemetryAppender` need to maintain fields for these. Should `LogSink` accept `LogRecords` without resource and instrumenation library, and let the pipeline attach those?

### DIFF
--- a/instrumentation/log4j/log4j-2-testing/build.gradle.kts
+++ b/instrumentation/log4j/log4j-2-testing/build.gradle.kts
@@ -5,6 +5,8 @@ plugins {
 dependencies {
   api(project(":testing-common"))
 
+  api("io.opentelemetry:opentelemetry-sdk-extension-logging")
+
   api("org.apache.logging.log4j:log4j-core:2.7")
 
   implementation("com.google.guava:guava")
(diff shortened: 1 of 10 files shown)


jack-berg:
Oops. Meant to open as a draft. Re-opened [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4375).

PR #4375, opened 2021-10-13
Prototype Log4j2 Appender

jack-berg:
*EDIT* - keeping the description below around, but all the issues it mentions have been solved in the 1.9.0 release of the log SDK. This is no longer a prototype. 

This is meant to prompt discussion. Proves out what an `OpenTelemetryAppender` might look like for log4j2. The idea behind this appender is that it delivers log records to a opentelemetry `LogSink`, where they can be batched and shipped out of process via an exporter, such as the recently added OTLP grpc and http log exporters. 

One of the main problems I had to address was that log appenders are instantiated before the opentelemetry log sdk is available. To get around this, each `OpenTelemetryAppender` instance is statically registered after initialization. Once the otel log sdk is available, it is passed to each `OpnTelemetryAppender` instance via a static method, upon which log records start flowing.

Learnings so far:
- LogRecord trace id and span id need to be nullable. And potentially other properties too. The otlp grpc and http exporters currently fail to serialize if span id and trace id are left as their default `""` values. Solved [here](https://github.com/open-telemetry/opentelemetry-java/pull/3745).
- We need a `InMemoryLoggingExporter` for testing. Mocked it out in this code, and will contribute back to `opentelemetry-java`. Solved [here](https://github.com/open-telemetry/opentelemetry-java/pull/3757).
- We need a `SimpleLogProcessor` for testing. Mocked it out in this code, and will contribute back to `opentelemetry-java`. Solved [here](https://github.com/open-telemetry/opentelemetry-java/pull/3749).
- `LogSinkSdkProvider.builder()` is not public but needs to be for it to be useful. I'll make it so. Solved [here](https://github.com/open-telemetry/opentelemetry-java/pull/3745/files).
- `LogSinkSdkProvider` should be immutable after its created. Log processors should be registered with the builder, not with the instance. I'll make it so. Solved [here](https://github.com/open-telemetry/opentelemetry-java/pull/3751).
- Thought needs to be put into how resource and instrumentation library get attached to the log records. Right now and producer to `LogSink` is responsible for creating `LogRecord`s with resource and instrumenation library attached. This means that components like the `OpenTelemetryAppender` need to maintain fields for these. Should `LogSink` accept `LogRecords` without resource and instrumenation library, and let the pipeline attach those?


instrumentation/log4j/log4j-2.13.2/library/src/test/groovy/OpenTelemetryAppenderTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+  def setupSpec() {
+    logExporter = InMemoryLogExporter.create()
+    resource = Resource.getDefault()
+    instrumentationLibraryInfo = InstrumentationLibraryInfo.create(OpenTelemetryAppender.getName(), null)
+    attributes = Attributes.builder().put("logger.name", "TestLogger").put("thread.name", "Time-limited test").build()
+
+    def sdkSinkProvider = LogSinkSdkProvider.builder().build()
+    sdkSinkProvider.addLogProcessor(SimpleLogProcessor.create(logExporter))
+    def logSink = sdkSinkProvider.get(null, null)
+    OpenTelemetryLog4j.initialize(logSink, resource)
```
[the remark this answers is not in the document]
jkwatson: I think this is a reasonable approach for now. Is it possible to build more complex configuration that would allow log4j to run the initialization itself?

jack-berg:
Now that 1.9.0 has been released, I've updated this branch, refined the logic, and added more thorough tests.

Its now ready for review.

**NOTE** this still places an explicit dependency on the otel java 1.9.0 log SDK. I think the whole repo should be updated to use 1.9.0 before this is merged.

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
```diff
@@ -3,11 +3,15 @@ plugins {
 }
 
 dependencies {
+  api("io.opentelemetry:opentelemetry-sdk-logs:1.9.0-alpha")
```
iNikem, line 6: That will make all clients of this library instrumentation depend on SDK. I assume all users must depend only on API.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+
+    // time
+    builder.setEpoch(logEvent.getNanoTime(), TimeUnit.NANOSECONDS);
+
+    // level
+    Level level = logEvent.getLevel();
+    if (level != null) {
+      builder.setSeverity(
+          LEVEL_SEVERITY_MAP.getOrDefault(level, Severity.UNDEFINED_SEVERITY_NUMBER));
+      builder.setSeverityText(logEvent.getLevel().name());
```
iNikem, line 80:
```suggestion
      builder.setSeverityText(level.name());
```

And shouldn't this text come from our severity and not logging framework level?

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
jack-berg: Users need to call `OpenTelemetryLog4j.initialize(SdkLogEmitterProvider)` to wire opentelemetry into the appender so the log SDK is actually part of this module's API.

jack-berg: But this reminds me that I need to update the readme to show usage!

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: From the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-severitytext), the severity text field "is the original string representation of the severity as it is known at the source".

But this does bring up the point that I have made a number of subjective decisions in mapping between the Log4j2 LogEvent data model and the otel log data model. Please take a close look at the attribute name's I've chosen for various things and let me know if you agree / disagree.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryLog4j.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private static final Object LOCK = new Object();
+
+  @GuardedBy("LOCK")
+  private static LogEmitter logEmitter;
+
+  @GuardedBy("LOCK")
+  private static final List<OpenTelemetryAppender> APPENDERS = new ArrayList<>();
+
+  public static void initialize(SdkLogEmitterProvider sdkLogEmitterProvider) {
```
mateuszrzeszutek, line 24: How is this going to work as a javaagent instrumentation? In the agent we keep all the SDK classes hidden in the agent classloader, they're basically unavailable from the application classloader. The agent won't be able to register its emitterProvider in the application's `OpenTelemetryLog4j` class.
Was the idea about "log appender api" completely scrapped? What happened with that?

instrumentation/log4j/log4j-2-common/testing/build.gradle.kts
```diff
@@ -5,6 +5,8 @@ plugins {
 dependencies {
   api(project(":testing-common"))
 
+  api("io.opentelemetry:opentelemetry-sdk-logs:1.9.0-alpha")
```
anuraaga, line 8: Don't think we need version here since we apply BOM for io.opentelemetry. Similar for all usages

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
anuraaga: @iNikem Yeah - we have https://github.com/open-telemetry/opentelemetry-java/issues/3807 to track fixing this.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.message.Message;
+import org.apache.logging.log4j.util.ReadOnlyStringMap;
+
+final class LogEventMapper {
+
+  // Visible for testing
+  static final AttributeKey<String> ATTR_LOGGER_NAME = AttributeKey.stringKey("logger.name");
```
anuraaga, line 38: I don't see any semantic conventions for logs defined yet, can we leave them out until they're defined?

https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/logs/semantic_conventions

anuraaga: If leaving any, I guess logger.name, throwable.name, throwable.message are the only ones I'd keep as they may be necessary for playing with a prototype, but others seem too detailed, or unclear (e.g., I'm not familiar with term `ndc`)

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+      AttributeKey.stringKey("throwable.message");
+  static final AttributeKey<List<String>> ATTR_NDC = AttributeKey.stringArrayKey("ndc");
+  static final AttributeKey<String> ATTR_FQCN = AttributeKey.stringKey("fqcn");
+  static final AttributeKey<String> ATTR_MARKER = AttributeKey.stringKey("marker");
+
+  private static final Map<Level, Severity> LEVEL_SEVERITY_MAP;
+
+  static {
+    Map<Level, Severity> levelSeverityMap = new HashMap<>();
+    levelSeverityMap.put(Level.ALL, Severity.TRACE);
```
anuraaga, line 53: Use a method with a `switch` instead of a Map - complete enum switches get compiled into a jump table based on the enum ordinal, which is highly efficient. Errorprone verifying the switch is complete is added icing.

Alternative is `EnumMap`, but we'd only prefer it for an incomplete mapping which doesn't become a jump table.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+    AttributesBuilder attributes = Attributes.builder();
+
+    // message
+    Message message = logEvent.getMessage();
+    if (message != null) {
+      builder.setBody(message.getFormattedMessage());
+    }
+
+    // time
+    builder.setEpoch(logEvent.getNanoTime(), TimeUnit.NANOSECONDS);
```
anuraaga, line 73: My reading of the javadoc is that `getNanoTime` is just `System.nanoTime()`, not the epoch time

https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/LogEvent.html#getNanoTime--

`getInstant` is the epoch

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+    if (level != null) {
+      builder.setSeverity(
+          LEVEL_SEVERITY_MAP.getOrDefault(level, Severity.UNDEFINED_SEVERITY_NUMBER));
+      builder.setSeverityText(logEvent.getLevel().name());
+    }
+
+    // logger
+    attributes.put(ATTR_LOGGER_NAME, logEvent.getLoggerName());
+
+    // fully qualified class name
```
anuraaga, line 86: Hmm - I wonder what this is, I thought the only "class name" type of thing is the logger name, if you derive it from a class. Otherwise each log message would be quite expensive

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+    if (contextStackList.size() > 0) {
+      attributes.put(ATTR_NDC, contextStackList);
+    }
+
+    // mapped diagnostic context (included span context)
+    ReadOnlyStringMap contextData = logEvent.getContextData();
+    if (contextData != null) {
+      Map<String, String> contextMap = contextData.toMap();
+      if (contextMap != null) {
+        String traceId =
```
anuraaga, line 118: I would like to first try with `Context.current()` and then consider alternatives if we find, e.g. people really want to wrap OTel in a async appenders, it is a significant usability problem.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,141 @@
...
+                ? TraceFlags.fromHex(contextMap.remove(TRACE_FLAGS), 0)
+                : TraceFlags.getDefault();
+        Context context =
+            Context.root()
+                .with(
+                    Span.wrap(
+                        SpanContext.create(traceId, spanId, traceFlags, TraceState.getDefault())));
+        builder.setContext(context);
+
+        contextMap.forEach(attributes::put);
```
anuraaga, line 133: I don't think we should fill arbitrary attributes, I don't think we ever do anything similar for spans

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryAppender.java
```diff
@@ -0,0 +1,76 @@
...
+public class OpenTelemetryAppender extends AbstractAppender {
+
+  static final String PLUGIN_NAME = "OpenTelemetry";
+
+  @PluginBuilderFactory
+  public static <B extends Builder<B>> B newBuilder() {
+    return new Builder<B>().asBuilder();
+  }
+
+  public static class Builder<B extends Builder<B>> extends AbstractAppender.Builder<B>
```
anuraaga, line 35: Since `newBuilder` is public, I'm guessing this class doesn't have to be, but not sure

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryLog4j.java
```diff
@@ -0,0 +1,59 @@
...
+
+import io.opentelemetry.sdk.logs.LogEmitter;
+import io.opentelemetry.sdk.logs.SdkLogEmitterProvider;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.concurrent.GuardedBy;
+
+public final class OpenTelemetryLog4j {
+
+  private static final Object LOCK = new Object();
```
anuraaga, line 16: We generally use `lock` (the idea being that just because it's static final doesn't mean it's constant, as we know a lock has state that mutates)

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryLog4j.java
anuraaga: We'll need it - https://github.com/open-telemetry/opentelemetry-java/issues/3807

instrumentation/log4j/log4j-2.13.2/library/src/test/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapperTest.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.logging.log4j.spi.DefaultThreadContextStack;
+import org.apache.logging.log4j.spi.ThreadContextStack;
+import org.apache.logging.log4j.util.ReadOnlyStringMap;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+class LogEventMapperTest {
+
+  private LogEvent logEvent;
```
anuraaga, line 56: Better to use `MockitoExtension` and `@Mock` to get strict stubs

anuraaga: That being said, I think we'd generally verify mapping via `LogData` instead of using mocks, i.e., these tests could be transfered into the AppenderTest. Imagine adding some helper methods or something on the builder - mocks are nice for RPC stubs, but for data classes we may as well check the actual data.

anuraaga (commented):
I think we need a new artifact, maybe `log4j-logssdk-prototype` since we can't leak the SDK dependency into existing users of the log4j instrumentation.

If a temporary artifact in Maven Central is not palatable (it's not great TBH), I think we should hold off for the 1.9.0 release and prioritize https://github.com/open-telemetry/opentelemetry-java/issues/3807. Sorry, didn't notice this ordering constraint until now.

We could still get a snapshot out sooner than later by

- Adding `log4j-logssdk-prototype` after the instrumentation 1.9.0 release, and merging it in for 1.10, using the log appender API
- Adding a `compileOnly` dependency on `log4j` in the opentelemetry-sdk-logs artifact itself and put the code there.

The second option seems to allow the easiest iteration on snapshots.

instrumentation/log4j/log4j-2-common/testing/build.gradle.kts
jack-berg: Yes, I'm just waiting for this project to update its BOM dependency on 1.9.0 and then will drop the version here.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: ndc = Nested Diagnostic Context

Nested diagnostic context is thread context that acts like a stack. You push items onto it, and pop them off. Mapped diagnostic context is thread context that acts like a map. Frankly, I haven't used Nested Diagnostic Context, and have only used Mapped Diagnostic Context in a limited way in applications I've written.

I took the stance of making sure that all data in the log4j2 data model somehow was mapped into the otel log data model. You're suggesting the opposite, where only a limited part of the log4j2 data model is mapped. I see pros and cons of each stance:
- If we only do a partial mapping of fields well defined in semantic conventions, then we're less likely to make breaking changes, though we can't eliminate breaking changes since there are no stable semantic conventions. We're also going to disappoint folks using parts of the log4j2 data model that are unmapped. 
- If we do a fully mapping, we basically guarantee breaking changes, but at least all the data gets mapped. And by at least making an attempt at mapping each field we can provide better feedback for the semantic conventions.

Personally, I prefer the full mapping approach published in an "alpha" artifact.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: Wow I didn't know that about enum switch compilation. Cool!

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: Nice catch. Must have missed it because when I tested it with an app on my machine it was returning an epoch nanos.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: This [test](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4375/files#diff-c3aa4475c1045c52b212b54274b73f836e2cdca46381c0dd2e5601e87243bea9R135) demonstrates the value. In that  case it was `org.apache.logging.log4j.spi.AbstractLogger`. Not sure how useful it is, but as mentioned [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4375#discussion_r749527440), my stance was to map everything.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: Kind of related to the the conversation about [doing a complete mapping](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4375#discussion_r749527440) or not.

Not filling arbitrary attributes from mapped diagnostic context is limiting the usefulness of the log4j2 api. It means if you want to use otel logs in a more "machine generated events" type of way, that you'd want to use the otel log sdk in your application instead of the log4j2 api.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: Sounds like a good plan.

instrumentation/log4j/log4j-2.13.2/library/src/test/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapperTest.java
jack-berg: Yeah looking at this test closer, all the code being exercised is also exercised in `OpenTelemetryAppenderConfigTest`. So I feel comfortable getting rid of this test class altogether.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryLog4j.java
jack-berg: Took a stab at extracting out the appender API [here](https://github.com/open-telemetry/opentelemetry-java/pull/3880).

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
anuraaga: Sleeping on this, I realized we can keep this code here by making this `compileOnly` - anyone adding the appender has to add the SDK to their classpath themselves.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
anuraaga: I've seen it's easier to add an attribute and not really go anywhere in the spec with it than the opposite. By not including too much, we can hear about what users really need and then build the spec around it, rather than being ad-hoc. While I don't like to disappoint folks, actually hearing about that disappointment is very valuable feedback.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
anuraaga: I think we'd generally use baggage for more arbitrary attributes? Though we haven't even specced it for spans. But I'd still suggest keeping things smaller for now to make sure the spec grows rather than going all in

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
anuraaga: Though I still think it might be simpler to continue to prototype with the code in the SDK repo. What do you think?

jkwatson: I'm definitely wary about adding an API API to support this. I think `compileOnly` and a reflective check for the presence of an SDK is a better way to go until we have a spec for a new logging API.

jack-berg: `compileOnly` dependency on the SDK sounds reasonable to me. 

> Though I still think it might be simpler to continue to prototype with the code in the SDK repo. 

Prototyping in the SDK has the advantage that the SDK and the appender can develop in lockstep, rather than the appender waiting on SDK releases, which is nice. But I think a lot of the SDK flux has been flushed out (of course assuming no significant changes to the experimental log spec) and so I don't expect to see much value from having the appender in the SDK.

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
jack-berg: That resonates with me. Also, if we were to do a complete mapping and the required semantic conventions never materialized, we may have to pull support for that mapping, which would be an unpleasant breaking change for users. 

Updated to a minimalist mapping in new commit.

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
```diff
@@ -3,11 +3,16 @@ plugins {
 }
 
 dependencies {
+  compileOnly("io.opentelemetry:opentelemetry-sdk-logs:1.9.0-alpha")
```
anuraaga, line 6: The repo's updated to 1.9 so should be able to merge main and remove the versions now

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/LogEventMapper.java
```diff
@@ -0,0 +1,102 @@
...
+      case WARN:
+        return Severity.WARN;
+      case ERROR:
+        return Severity.ERROR;
+      case FATAL:
+        return Severity.FATAL;
+      case OFF:
+        return Severity.UNDEFINED_SEVERITY_NUMBER;
+    }
+    throw new IllegalStateException("Unrecognized level " + level.name());
```
anuraaga, line 98: While log4j probably doesn't ever add log levels, it's conceivable and could trigger this line if log4j is newer than what we've gotten to compile against. Let's return UNDEFINED_SEVERITY_NUMBER instead of throwing

anuraaga (approved):

mateuszrzeszutek (approved):

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryAppender.java
```diff
@@ -0,0 +1,78 @@
...
+@Plugin(
+    name = OpenTelemetryAppender.PLUGIN_NAME,
+    category = Core.CATEGORY_NAME,
+    elementType = Appender.ELEMENT_TYPE)
+public class OpenTelemetryAppender extends AbstractAppender {
+
+  static final String PLUGIN_NAME = "OpenTelemetry";
+
+  @PluginBuilderFactory
+  public static <B extends Builder<B>> B newBuilder() {
```
trask, line 31: we've recently standardized on `builder` over `newBuilder` (#4392)
```suggestion
  public static <B extends Builder<B>> B builder() {
```

instrumentation/log4j/log4j-2.13.2/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_13_2/OpenTelemetryLog4j.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @GuardedBy("lock")
+  private static final List<OpenTelemetryAppender> APPENDERS = new ArrayList<>();
+
+  public static void initialize(SdkLogEmitterProvider sdkLogEmitterProvider) {
+    List<OpenTelemetryAppender> instances;
+    synchronized (lock) {
+      if (OpenTelemetryLog4j.sdkLogEmitterProvider != null) {
+        throw new IllegalStateException("SdkLogEmitterProvider has already been set.");
+      }
```
trask, line 28: given trickiness around debugging init sequences, is it worth adding the stack trace of the original caller similar to GlobalOpenTelemetry?

https://github.com/open-telemetry/opentelemetry-java/blob/efbc87b4df38ea457c159b9360ae0db8a1791bb1/api/all/src/main/java/io/opentelemetry/api/GlobalOpenTelemetry.java#L82-L93

jack-berg: I like that idea.

jack-berg:
Can this PR be merged?

trask:
🎉

PR #4377, opened 2021-10-14
Fix reactor context propagation in latest reactor

### LINKED ISSUE #4372: Nightly build failed
But it looks like I broke the auto-issue creation, trying to fix that in #4371

In the meantime, this is a manual created issue for the nightly build failure: https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/1335848063

@lmolkova it looks like the new reactor test is failing against the latest reactor version, if you have time can you take a look?

it should repro it locally running:

```
./gradlew :instrumentation:reactor-3.1:library:test -PtestLatestDeps=true
```

lmolkova:
Fixes #4372 introduced in  #4159

The newer reactor does not support a hack ` dummy.flatMap(i -> publisher)` to force publisher in `onNext` (i.e. with `current()` span)

It seems there is optimization or some different path that causes the below branch to run, skipping context instrumentation

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/51cf64bb3fb06ef4d6ef9a3d8b60751674979135/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/TracingOperator.java#L91-L94

One way to solve it would be to enable context propagation for scalars, but it probably would be bad for performance.

Instead, I went down the road of creating a small custom Mono impl instead of dummy that implements context propagation in `subscribe` call.

### DIFF
--- a/instrumentation/reactor-3.1/library/src/test/groovy/io/opentelemetry/instrumentation/reactor/ReactorCoreTest.groovy
+++ b/instrumentation/reactor-3.1/library/src/test/groovy/io/opentelemetry/instrumentation/reactor/ReactorCoreTest.groovy
@@ -202,9 +202,9 @@ class ReactorCoreTest extends AbstractReactorCoreTest implements LibraryTestTrai
 
   def getDummy(def publisher) {
     if (publisher instanceof Mono) {
-      return ContextPropagationOperator.ScalarPropagatingMono.INSTANCE;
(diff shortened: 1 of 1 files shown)


iNikem (approved):

trask:
@lmolkova looks like `./gradlew :instrumentation:reactor-3.1:library:codenarcTest` is failing, usually for me that means I used semi-colons in groovy code 😢

PR #4381, opened 2021-10-14
Migrate Netty 4.x client instrumentations (except CONNECT) to Instrum…

mateuszrzeszutek:
…enter API

Connection spans are not included in this PR, I'll migrate them separately to make those PRs a bit easier to review.

Part of #2713

### DIFF
--- a/instrumentation/netty/netty-4.0/javaagent/src/test/groovy/Netty40ClientTest.groovy
+++ b/instrumentation/netty/netty-4.0/javaagent/src/test/groovy/Netty40ClientTest.groovy
@@ -139,4 +139,9 @@ class Netty40ClientTest extends HttpClientTest<DefaultFullHttpRequest> implement
   boolean testHttps() {
     false
   }
+
+  @Override
+  boolean testReadTimeout() {
+    true
+  }
 }
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/NettyCommonNetAttributesExtractor.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/NettyCommonNetAttributesExtractor.java
@@ -3,16 +3,15 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.javaagent.instrumentation.netty.common.server;
+package io.opentelemetry.javaagent.instrumentation.netty.common;
 
 import io.netty.handler.codec.http.HttpResponse;
 import io.opentelemetry.instrumentation.api.instrumenter.net.InetSocketAddressNetServerAttributesExtractor;
-import io.opentelemetry.javaagent.instrumentation.netty.common.HttpRequestAndChannel;
 import java.net.InetSocketAddress;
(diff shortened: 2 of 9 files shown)


instrumentation/netty/netty-4.0/javaagent/src/test/groovy/Netty40ClientTest.groovy
```diff
@@ -139,9 +139,4 @@ class Netty40ClientTest extends HttpClientTest<DefaultFullHttpRequest> implement
   boolean testHttps() {
     false
   }
-
-  @Override
-  boolean testReadTimeout() {
```
mateuszrzeszutek, line 144: Now that we've switched over to using `NetClientAttributesExtractor` in netty instrumentations the `net.peer.*` attributes are no longer captured (`channel.remoteAddress()` returns null in case of timeout) and this test case fails. 
There are no more tests in any instrumentations that enable it - we should verify how other clients behave and, if the lack of `net.*` attributes is a common thing, change the test scenario to reflect that. I'll create a separate issue for this once this PR gets merged.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/HttpClientRequestTracingHandler.java
```diff
@@ -30,27 +32,40 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise prm) {
...
+    // The AWS SDK uses Netty for asynchronous clients but constructs a request signature before
+    // beginning transport. This means we MUST suppress Netty spans we would normally create or
+    // they will inject their own trace header, which does not match what was present when the
+    // signature was computed, breaking the SDK request completely. We have not found how to
+    // cleanly propagate context from the SDK instrumentation, which executes on an application
+    // thread, to Netty instrumentation, which executes on event loops. If it's possible, it may
+    // require instrumenting internal classes. Using a header which is more or less guaranteed to
+    // always exist is arguably more stable.
+    return request.request().headers().contains("amz-sdk-invocation-id");
+  }
```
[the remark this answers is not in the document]
trask: can you open an issue? I agree this sounds promising to look through instrumentation to see how would many benefit

instrumentation/netty/netty-4.0/javaagent/src/test/groovy/Netty40ClientTest.groovy
mateuszrzeszutek: > btw, is the problem that `Channel.remoteAddress()` returns null in `onEnd` (while it returns non-null earlier in `onStart`)?

Yep, it's exactly that. I believe that we can do 2 things to fix that:
1. change the assertions in the timeout test so that they don't expect `net.*` attributes - this has a downside of losing some attributes that, as Lauri said, might be incredibly useful on a failing request span;
2. always capture `net.*` attributes on start - which would probably force us to re-think the net attributes extractors naming; it is also worth noticing that there were some HTTP clients that reported `net*` attributes on start, so it could be worth implementing anyway.

I think I'm leaning towards option 2, and just simply reusing the server net attributes extractor in this PR.

Issue #4370, opened 2021-10-13
Compute serialVersionUID for Serializable classes that don't have one and use VirtualField


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
When a `VirtualField` is added to a `Serializable` class that doesn't have a pre-computed `serialVersionUID` the added fields & interfaces will alter the default `serialVersionUID` computation.

**Describe the solution you'd like**
When a virtual field is being installed we could compute the UID ourselves - ASM even has a helper class for that: `SerialVersionUIDAdder`.
This behavior should be configurable (enabled by default); it is possible (though very unlikely) that this behavior might break the instrumented application - apparently it is possible to configure Java crypto so that it's not able to compute SHA hashes due to that.

**Additional context**
DataDog also does that in their agent: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/context/FieldBackedContextInjector.java#L487

PR #4400, opened 2021-10-15
Use byte-buddy-dep instead of byte-buddy

laurit:
Replaces `byte-buddy` which includes shaded asm with byte-buddy-dep that just depends on asm. As byte-buddy includes only the asm classes it needs it is missing some that we might need. For example for https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4370 we need `SerialVersionUIDAdder` which is not included in byte-buddy.

### DIFF
--- a/gradle-plugins/build.gradle.kts
+++ b/gradle-plugins/build.gradle.kts
@@ -23,9 +23,8 @@ configurations.named("compileOnly") {
 dependencies {
   implementation("com.google.guava:guava:30.1.1-jre")
   // we need to use byte buddy variant that does not shade asm
-  // for this we extract byte buddy gradle plugin inside our plugin
-  bbGradlePlugin("net.bytebuddy:byte-buddy-gradle-plugin:1.11.20") {
-    exclude("net.bytebuddy:byte-buddy")
+  implementation("net.bytebuddy:byte-buddy-gradle-plugin:1.11.20") {
+    exclude(group = "net.bytebuddy", module = "byte-buddy")
   }
   implementation("net.bytebuddy:byte-buddy-dep:1.11.20")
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -275,6 +275,10 @@ jobs:
         # javadoc task fails sporadically fetching https://docs.oracle.com/javase/8/docs/api/
         run: ./gradlew publishToMavenLocal -x javadoc
 
+      - name: Local publish gradle plugin
+        run: ../gradlew publishToMavenLocal -x javadoc
+        working-directory: gradle-plugins
+
```
trask, line 281: I think this won't be needed now that I finally got #4248 working

gradle-plugins/build.gradle.kts
```diff
@@ -72,4 +82,15 @@ tasks {
   publishPlugins {
     enabled = !version.toString().contains("SNAPSHOT")
   }
-}
\ No newline at end of file
+
+  val extractBbGradlePlugin by registering(Copy::class) {
+    from(zipTree(bbGradlePlugin.files.first { it.name.contains("byte-buddy-gradle-plugin") })) {
```
iNikem, line 87: Honestly, this looks like a hack. Why exactly do we need to unzip a dependency jar file into our `classes` folder instead of putting jar file to a classpath?

iNikem (commented):
Is this `SerialVersionUIDAdder` classes needed only for our gradle plugin? Compile time? Or runtime? Why do we have to switch dependency for the remaining project?

.github/workflows/ci.yml
trask: 👍 going to merge this PR since #4248 is getting reworked, and I'll take care of it in that PR

PR #4403, opened 2021-10-15
Don't report 400 level as error for SERVER spans

### LINKED ISSUE #4357: Don't set Span.Status for 4xx http status codes for SERVER spans
According to https://github.com/open-telemetry/opentelemetry-specification/pull/1998

breedx-splk:
This resolves #4357

I am assuming that this will likely break some tests, but let's get it going and I'll address those as they arise.

### DIFF
--- a/instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/AbstractServlet3MappingTest.groovy
+++ b/instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/AbstractServlet3MappingTest.groovy
@@ -16,7 +16,6 @@ import javax.servlet.http.HttpServletRequest
 import javax.servlet.http.HttpServletResponse
 
 import static io.opentelemetry.api.trace.StatusCode.ERROR
-import static io.opentelemetry.api.trace.StatusCode.UNSET
 
 abstract class AbstractServlet3MappingTest<SERVER, CONTEXT> extends AgentInstrumentationSpecification implements HttpServerTestTrait<SERVER> {
 
@@ -51,9 +50,6 @@ abstract class AbstractServlet3MappingTest<SERVER, CONTEXT> extends AgentInstrum
           kind SpanKind.SERVER
           if (!success && response.status().code() >= 500) {
             status ERROR
-          }
-          else if (!success && response.status().code() < 500) {
-            status UNSET
           }
         }
         if (!success) {
--- a/instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/AbstractServlet5MappingTest.groovy
+++ b/instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/AbstractServlet5MappingTest.groovy
@@ -15,7 +15,6 @@ import jakarta.servlet.http.HttpServletResponse
 import spock.lang.Unroll
 
 import static io.opentelemetry.api.trace.StatusCode.ERROR
-import static io.opentelemetry.api.trace.StatusCode.UNSET
 
 abstract class AbstractServlet5MappingTest<SERVER, CONTEXT> extends AgentInstrumentationSpecification implements HttpServerTestTrait<SERVER> {
 
@@ -50,9 +49,6 @@ abstract class AbstractServlet5MappingTest<SERVER, CONTEXT> extends AgentInstrum
           kind SpanKind.SERVER
           if (!success && response.status().code() >= 500) {
             status ERROR
-          }
-          else {
-            status UNSET
           }
         }
         if (!success) {


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/DefaultSpanStatusExtractor.java
```diff
@@ -15,7 +16,7 @@
 
   @Override
   public StatusCode extract(
-      REQUEST request, @Nullable RESPONSE response, @Nullable Throwable error) {
+      REQUEST request, @Nullable RESPONSE response, SpanKind kind, @Nullable Throwable error) {
```
mateuszrzeszutek, line 19: Instead of introducing a `kind` param, how about deciding the behavior basing on the attributes extractor used? I.e. you could check `extractor instanceof HttpServerAttributesExtractor` and choose the behavior based on that

iNikem: +1 to choosing right implementation ASAP. Although it does not make any difference performance-wise.

jkwatson: But, that would mean that no one could build a kind-oriented attributes extractor (independent of this particular one for http server spans), which still seems like a limitation that might not be desirable.

breedx-splk: I've never met an `instanceof` that I didn't despise (and later regret).  😁

trask: would splitting into `HttpClientSpanStatusExtractor` and `HttpServerSpanStatusExtractor` work?

we've been finding it useful to split out other extractors based on client/server, e.g. `Http(Client|Server)AttributesExtractor` and `Net(Client|Server)AttributesExtractor` (though there's an open discussion on the latter in #4408)

breedx-splk: I considered it but was too concerned it might cause problems with frameworks (and consequently instrumentations) that aren't purely client/server, right? Isn't there instrumentation that can generate both SERVER and CLIENT spans? I'm thinking about messaging systems and http/2 server push, for example.

This was also the thinking behind the "late binding", if you will (I think this is what @iNikem was getting at with the "ASAP" comment).

If this isn't true and we don't think it'll be a problem, we could try that out.

iNikem: > I considered it but was too concerned it might cause problems with frameworks (and consequently instrumentations) that aren't purely client/server, right? Isn't there instrumentation that can generate both SERVER and CLIENT spans? I'm thinking about messaging systems and http/2 server push, for example.

They already will have to provide the correct `SpanKindExtractor`. Which means that they will most probably have separate instrumenters per kind. Thus separate `SpanStatusExtractor` will not be a problem either.

trask: > Isn't there instrumentation that can generate both SERVER and CLIENT spans?

I think all of these are creating two instrumenters already, one to handle server spans and one to handle client spans.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -202,6 +201,7 @@ public void end(
       }
     }
 
+    SpanKind kind = spanKindExtractor.extract(request);
```
mateuszrzeszutek, line 204: No longer needed anymore:
```suggestion
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanStatusExtractor.java
```diff
@@ -19,30 +19,44 @@
...
 
+  private final HttpStatusConverter statusConverter;
+
+  /**
+   * Returns the {@link SpanStatusExtractor} for HTTP requests, which will use the HTTP status code
+   * to determine the {@link StatusCode} if available or fallback to {@linkplain #getDefault() the
+   * default status} otherwise.
+   */
+  public static <REQUEST, RESPONSE> SpanStatusExtractor<REQUEST, RESPONSE> createClient(
+      HttpCommonAttributesExtractor<? super REQUEST, ? super RESPONSE> attributesExtractor) {
```
mateuszrzeszutek, line 30: WDYT about requiring the `HttpClientAttributesExtractor` to be passed instead of encoding the convention in the method name?
```suggestion
  public static <REQUEST, RESPONSE> SpanStatusExtractor<REQUEST, RESPONSE> create(
      HttpClientAttributesExtractor<? super REQUEST, ? super RESPONSE> attributesExtractor) {
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpSpanStatusExtractor.java
```diff
@@ -19,30 +19,44 @@
...
+  }
+
   /**
    * Returns the {@link SpanStatusExtractor} for HTTP requests, which will use the HTTP status code
    * to determine the {@link StatusCode} if available or fallback to {@linkplain #getDefault() the
    * default status} otherwise.
    */
-  public static <REQUEST, RESPONSE> SpanStatusExtractor<REQUEST, RESPONSE> create(
+  public static <REQUEST, RESPONSE> SpanStatusExtractor<REQUEST, RESPONSE> createServer(
       HttpCommonAttributesExtractor<? super REQUEST, ? super RESPONSE> attributesExtractor) {
```
mateuszrzeszutek, line 40:
```suggestion
  public static <REQUEST, RESPONSE> SpanStatusExtractor<REQUEST, RESPONSE> create(
      HttpServerAttributesExtractor<? super REQUEST, ? super RESPONSE> attributesExtractor) {
```

instrumentation-api/src/test/groovy/io/opentelemetry/instrumentation/api/tracer/HttpServerStatusConverterTest.groovy
```diff
@@ -0,0 +1,94 @@
...
+    409        | StatusCode.UNSET
+    410        | StatusCode.UNSET
+    411        | StatusCode.UNSET
+    412        | StatusCode.UNSET
+    413        | StatusCode.UNSET
+    414        | StatusCode.UNSET
+    415        | StatusCode.UNSET
+    416        | StatusCode.UNSET
+    417        | StatusCode.UNSET
+    418        | StatusCode.UNSET
```
mateuszrzeszutek, line 66: I was hoping you won't miss that one 😂

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientStatusConverter.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.tracer;
+
+import io.opentelemetry.api.trace.StatusCode;
+
+final class HttpClientStatusConverter implements HttpStatusConverter {
```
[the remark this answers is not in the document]
breedx-splk: Good call, but yeah, this PR has grown and I'd prefer containing the blast radius. It's referenced by `HttpClientTracer` and `HttpServerTracer` but also used by a few tests and for the `HttpUrlConnectionInstrumentation` advice. A separate issue seems like a better way to handle it.

I'm happy to take on that work. Do you want to create the issue or should I?

trask (approved):

instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/AbstractServlet3MappingTest.groovy
```diff
@@ -48,9 +49,12 @@ abstract class AbstractServlet3MappingTest<SERVER, CONTEXT> extends AgentInstrum
         span(0) {
           name getContextPath() + spanName
           kind SpanKind.SERVER
-          if (!success) {
+          if (!success && response.status().code() >= 500) {
             status ERROR
           }
+          else if (!success && response.status().code() < 500) {
+            status UNSET
+          }
```
mateuszrzeszutek, line 57: Nit: no need to check for that explicitly; `SpanAssert` will verify that the status is `UNSET` in case you haven't done that in your test code

trask: @breedx-splk is this another vote for converting groovy tests to java? 😂

Issue #4291, opened 2021-10-04
Type annotation placement


trask:
This has been discussed previously: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1214#discussion_r490719668 and #3023 (and probably elsewhere).

The existing convention (and desire of majority maintainers I believe) is to not use "type annotation" placement for return type annotations.

However Intellij keeps using "type annotation" placement when implementing super type methods, and so things keep drifting (there are ~300 instances currently of "type annotation" placement).

I noticed that the otel-java repo doesn't seem to have this problem because it uses `javax.annotation.Nullable` instead of checker framework's `Nullable` (which specifically targets `TYPE_USE`).

I'd like to suggest that we either:

(1) switch to `javax.annotation.Nullable`, or
(2) switch to "type annotation" placement

trask:
This is done

PR #4406, opened 2021-10-17
Type annotation placement

MartinDai:
switch annotation `org.checkerframework.checker.nullness.qual.Nullable` to `javax.annotation.Nullable` #4291


trask (approved):
thx @MartinDai!

trask:
Going ahead and merging to avoid conflicts

iNikem:
Do I miss something? This PR changes which annotation we use, but it did not change the annotation location, right? We still have to do that somehow, right?

mateuszrzeszutek:
The annotation placement seems to have been changed though
```java
  @Nullable
  protected Long requestContentLength(
```
Not sure if in all places though

PR #4409, opened 2021-10-18
Migrate Grizzly to Instrumenter API

mateuszrzeszutek:
Part of #2713 

Grizzly was the last HTTP server instrumentation that used `HttpServerTracer` - now all of them are using `Instrumenter`

### DIFF
--- a/instrumentation/grizzly-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grizzly/DefaultFilterChainInstrumentation.java
+++ b/instrumentation/grizzly-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grizzly/DefaultFilterChainInstrumentation.java
@@ -47,7 +47,7 @@ public class DefaultFilterChainInstrumentation implements TypeInstrumentation {
       Context context = GrizzlyStateStorage.removeContext(ctx);
       HttpRequestPacket request = GrizzlyStateStorage.removeRequest(ctx);
       if (context != null && request != null) {
-        Throwable error = GrizzlyExceptionHolder.getOrDefault(context, throwable);
+        Throwable error = GrizzlyErrorHolder.getOrDefault(context, throwable);
         instrumenter().end(context, request, null, error);
       }
     }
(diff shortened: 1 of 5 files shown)


instrumentation/grizzly-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grizzly/DefaultFilterChainInstrumentation.java
```diff
@@ -43,9 +44,10 @@ public void transform(TypeTransformer transformer) {
     @Advice.OnMethodEnter(suppress = Throwable.class)
     public static void onFail(
         @Advice.Argument(0) FilterChainContext ctx, @Advice.Argument(1) Throwable throwable) {
-      Context context = tracer().getServerContext(ctx);
-      if (context != null) {
-        tracer().endExceptionally(context, throwable);
+      Context context = GrizzlyStateStorage.removeContext(ctx);
+      HttpRequestPacket request = GrizzlyStateStorage.removeRequest(ctx);
```
iNikem, line 48: Should we introduce `ContextAndRequest` here ;)

instrumentation/grizzly-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grizzly/GrizzlyErrorHandler.java
```diff
@@ -0,0 +1,23 @@
...
+
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.instrumentation.api.instrumenter.ErrorCauseExtractor;
+
+public final class GrizzlyErrorHandler {
+
+  // copied from BaseTracer#onException()
+  public static void onError(Context context, Throwable error) {
```
iNikem, line 16: Is here something that `io.opentelemetry.instrumentation.api.instrumenter.Instrumenter#end` does not already do?

instrumentation/grizzly-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grizzly/GrizzlyHttpAttributesExtractor.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+package io.opentelemetry.javaagent.instrumentation.grizzly;
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpServerAttributesExtractor;
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.Nullable;
```
iNikem, line 13: Can you replace that with `javax.annotation.Nullable`?

trask (approved):

PR #4417, opened 2021-10-19
Capture response headers for HttpURLConnection

trask:


instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -74,6 +74,11 @@ public static void methodEnter(
         // top-level HttpURLConnection calls
         return;
       }
+      // double increment on first entry is used to prevent infinite recursion in case end()
+      // captures response headers due to HttpUrlConnection.getHeaderField() calling
+      // HttpUrlConnection.getInputStream() which then enters this advice again
```
mateuszrzeszutek, line 79: 🤯

mateuszrzeszutek (approved):

Issue #4408, opened 2021-10-18
Re-think our net attributes extractors


mateuszrzeszutek:
**Describe the bug**
Currently we have a `NetServerAttributesExtractor` that extracts `net.*` attributes on start and a `NetClientAttributesExtractor` that extracts `net.*` attributes on end. As it turned out in #4381, some HTTP client instrumentations (netty) need to capture net attributes on start, because they become unavailable at the end of request processing in some scenarios (timeouts in this particular case).
Perhaps we should go back to the `NetOn(Start|End)AttributesExtractor` naming? Or something else, e.g. `Net(Request|Response)AttributesExtractor`?

PR #4420, opened 2021-10-19
Extract net attributes both on start and on end in netty HTTP client

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4408 - on the SIG meeting today we've decided to just extract those attributes both on start and end and stay with the current naming scheme.

### DIFF
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettyNetClientAttributesExtractor.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettyNetClientAttributesExtractor.java
@@ -19,12 +19,7 @@ final class NettyNetClientAttributesExtractor
   @Nullable
   public InetSocketAddress getAddress(
       HttpRequestAndChannel requestAndChannel, @Nullable HttpResponse response) {
-    SocketAddress address = requestAndChannel.channel().remoteAddress();
-    // if the remote address is null at the end of request processing (because of e.g. lost
-    // connection) use the one captured when the operation started
-    if (address == null) {
-      address = requestAndChannel.remoteAddress();
-    }
+    SocketAddress address = requestAndChannel.remoteAddress();
     if (address instanceof InetSocketAddress) {
       return (InetSocketAddress) address;
     }


iNikem (approved):

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettyClientInstrumenterFactory.java
```diff
@@ -25,16 +26,24 @@ public NettyClientInstrumenterFactory(String instrumentationName) {
...
+        // 1. the common one will extract net attributes on start of the operation; in case of
+        // timeouts or other connection issues netty may return null when calling
+        // Channel.remoteAddress() at the end of processing
+        // 2. the client one will extract full net attributes at the end of processing - it should
+        // be the fully resolved address at this point in time
         .addAttributesExtractor(new NettyCommonNetAttributesExtractor())
-        // TODO: add peer extractor attributes once Net*AttributesExtractors are refactored
-        // .addAttributesExtractor(PeerServiceAttributesExtractor.create(netClientAttributesExtractor))
+        .addAttributesExtractor(netClientAttributesExtractor)
+        .addAttributesExtractor(PeerServiceAttributesExtractor.create(netClientAttributesExtractor))
```
trask, line 46: `net.peer.service` won't be captured on read timeout? what about alt idea I think touched on yesterday of adding `remoteAddress` into `HttpRequestAndChannel`, and then using that in `NettyNetClientAttributesExtractor`?

mateuszrzeszutek: Hmm, I'll try that out - I haven't thought about `peer.service` in the timeout case.

laurit (approved):

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettyNetClientAttributesExtractor.java
```diff
@@ -20,6 +20,11 @@
   public InetSocketAddress getAddress(
       HttpRequestAndChannel requestAndChannel, @Nullable HttpResponse response) {
     SocketAddress address = requestAndChannel.channel().remoteAddress();
+    // if the remote address is null at the end of request processing (because of e.g. lost
+    // connection) use the one captured when the operation started
+    if (address == null) {
+      address = requestAndChannel.remoteAddress();
+    }
```
trask, line 27: I don't think remoteAddress will change since netty instrumentation is tracing a single low-level request/response

```suggestion
    SocketAddress address = requestAndChannel.remoteAddress();
```

PR #4437, opened 2021-10-20
Unify images configuration for smoke test matrix build

iNikem:


trask (approved):
thanks for doing this ❤️

PR #4438, opened 2021-10-20
Publish javaagent bootstrap modules

trask:


trask:
> Why? :)

our distro still pulls in individual instrumentations, so I need the new kafka one in particular. i'm thinking to change our mechanism (e.g. `otel.instrumentation.common.default-enabled=false` and enabling specific instrumentations), though I think these probably make sense to publish anyways since they are essentially dependencies of the javaagent instrumentation modules that we publish already

mateuszrzeszutek:
> I think these probably make sense to publish anyways since they are essentially dependencies of the javaagent instrumentation modules that we publish already

💯

PR #4446, opened 2021-10-20
Netty server instrumentation now captures http.scheme

osherv:


osherv:
@trask I'll fix them, thanks!!

trask (approved):
thx @osherv! (and looks like there's another test that needs updating)

osherv:
@trask Thanks :)
Finally looks like all of them were fixed

PR #4451, opened 2021-10-20
Move common testing modules to match convention a bit better(?)

trask:
See settings.gradle.kts for quick overview of renames.

Primarily moved `xyz-testing` modules to `xyz-common:testing`.

Because in several cases we already have `xyz-common:javaagent` or `xyz-common:library`, so seems to fit well here. And then for consistency moved the others.

`jaxws-2.0-common-testing` is a bit of an outlier since it's used by `jaxws-2.0-axis2-1.6`, `jaxws-2.0-cxf-3.0` and `jaxws-2.0-metro-2.2`


instrumentation/elasticsearch/elasticsearch-transport-5.3/javaagent/build.gradle.kts
```diff
@@ -1,59 +1,59 @@
 plugins {
-  id("otel.javaagent-instrumentation")
+    id("otel.javaagent-instrumentation")
```
mateuszrzeszutek: You might want to revert formatting changes in this file (and a few others in this PR)

PR #4454, opened 2021-10-20
Remove unshaded javax.annotation classes from bootstrap class loader

trask:
This probably deserves a 1.7.1 patch unless there's reason to believe that since they're just annotations they shouldn't cause problems.

Created issue #4455 to add a test to guard against this.


anuraaga:
@trask At the same time you can move it into java-conventions where it was supposed to be :P

instrumentation/rxjava/rxjava-2.0/library/src/main/java/io/opentelemetry/instrumentation/rxjava2/TracingAssembly.java
```diff
@@ -22,6 +22,7 @@
 
 package io.opentelemetry.instrumentation.rxjava2;
 
+import io.opentelemetry.api.internal.GuardedBy;
```
anuraaga, line 25: Can you copy this into instrumentation api instead?

PR #4461, opened 2021-10-21
Migrate Netty 4.x CONNECT instrumentation to Instrumenter API

mateuszrzeszutek:
Part of #2713


instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
```diff
@@ -30,54 +30,48 @@
...
-  @SuppressWarnings("unused")
-  public static class InitAdvice {
-    @Advice.OnMethodEnter
-    public static void enter() {
-      // Ensure that tracer is initialized. Connection failure handling is initialized in the static
-      // initializer of tracer which needs to be run before an attempt is made to establish
-      // connection.
-      tracer();
-    }
-  }
```
trask, line 49: 👍

Issue #1415, opened 2020-10-19
Play instrumentation tests are not capturing route


trask:
`PlayServerTest` overrides `expectedServerSpanName`, which shows it's not capturing the route (which `HttpServerTest` is expecting by default):

```
  String expectedServerSpanName(String method, ServerEndpoint endpoint) {
    return "netty.request"
  }
```

It seems that `PlayTracer.updateSpanName()` is not finding `ROUTE_PATTERN` needed to populate the route.

I think it may only be a problem with the tests, as it looks like `PlaySmokeTest` is capturing the route correctly.

mcmho:
@trask   Should I work on this and get my hands dirty with unit testing?

trask:
> @trask Should I work on this and get my hands dirty with unit testing?

@mcmho that would be great!

mcmho:
feel free to assign to me.

medicharlachiranjeevi:
Is this fixed ? . We are using play framework version 2.5 in Prometheus exporter there no route was included in http bucket metrics

mateuszrzeszutek:
Hey @medicharlachiranjeevi ,
No, this has not been fixed yet.

medicharlachiranjeevi:
Thanks for response @mateuszrzeszutek .Is there any alternate way to get those metrics ?

trask:
I think we still need someone to dig into this issue

mwy001:
+1 for this issue. Using the java agent to auto collect play 2.x http metrics, and no http_route label was included in the prometheus result.

medicharlachiranjeevi:
hi, is there any eta on it?

mateuszrzeszutek:
Hey @medicharlachiranjeevi ,
We don't have any ETA on this. Nobody from our core maintainer team is particularly familiar with Play, honestly we're hoping for an external contribution here.

gurudatta-carbon:
@trask 
it works for us with a custom metric (similar to http.server.duration metric with http_route) defined part of play.api.mvc.Filter trait, 
Here I have a hold on RequestHeader object instance.
but it can retrieved from Request object as well,

this is the impl that works for us,

```
              val requestHeader: RequestHeader = req.request
              val handlerDefOpt: Option[HandlerDef] = requestHeader.handlerDef
              handlerDefOpt.map(handlerDef => attr += ("http.route" -> handlerDef.path))
```
req is of type Request.

moznion:
[...]

PR #4478, opened 2021-10-22
Play instrumentation tests are not capturing route #1415

mcmho:
This PR not only fix the PlayServerTest.  It also fixes the PlayTracer to update the span name with the play request api path.


mcmho:
@trask ,  here is my attempt.  let's see what you think.

trask:
@mcmho ok to close this PR now that findings have been cycled back into the issue?

Issue #3913, opened 2021-08-23
Add Android API-friendliness checks to library instrumentation


jkwatson:
For library instrumentation that we expect to work on Android, we should add some build-time checks to make sure we're not adding java 8 features that aren't supported on Android 21+.

- [ ] OkHttp3 3.x
- [ ] gRPC
- [ ] Apache HTTP Client 4.3

PR #4505, opened 2021-10-26
Add Android API-friendliness checks

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3913

I added animalsniffer checks to `instrumentation-api`, OkHttp, gRPC and Apache HTTP client library instrumentations. `instrumentation-api` needed a lot of additional ignores, mostly because of shaded caffeine2 - but they all currently seem to be desugared properly so it's all good, probably.


trask:
I removed all of the ignores, and added reflective handling for `CompletionException` which is unfortunately not in Android 21. This should pass once #4667 is merged.

anuraaga (approved):

trask (approved):

instrumentation-api/build.gradle.kts
```diff
@@ -67,28 +65,4 @@ tasks {
...
-      "java.util.concurrent.ForkJoinPool",
-      "java.util.concurrent.ConcurrentHashMap",
-      // LongAdder is also referenced by our own SupportabilityMetrics; only if agent debug is enabled though
-      "java.util.concurrent.atomic.LongAdder",
-
-      // these standard Java classes seem to be desugared properly
-      "java.lang.ClassValue",
-      "sun.misc.Unsafe",
-    )
-  }
```
mateuszrzeszutek: 🚀

Issue #995, opened 2020-08-15
Rename the Setter classes which currently have old naming "*InjectAdapter"


trask:
The `HttpTextFormat.Setter` class names don't make much sense anymore, should probably rename them to something like `*Setter`.

dengliming:
@trask It looks like the `getter ` does the same thing. and I'd love to submit a PR for this. Can you assigne to me?

dengliming:
Just Make sure I understand exactly what you mean and I have some options for modification.
1. just rename `*InjectAdapter` to something like `*Setter`. like `HttpHeadersInjectAdapter ` to  `HttpHeadersSetter`.
The effect of the change is as follows：
```
public enum HttpHeadersSetter implements HttpTextFormat.Setter<HttpRequest> {
  public static final HttpHeadersSetter SETTER = new HttpHeadersSetter();

  @Override
  public void set(HttpRequest carrier, String key, String value) {
    carrier.setHeader(key, value);
  }
}
```
2. Most of  `*Setter` are used as singletons. Maybe we can use enum instead of class like the following code?
```
public enum HttpHeadersSetter implements HttpTextFormat.Setter<HttpRequest> {
  INSTANCE;

  @Override
  public void set(HttpRequest carrier, String key, String value) {
    carrier.setHeader(key, value);
  }
}
```
@trask @iNikem @anuraaga Can you share your opinion here? : )

trask:
Hey @dengliming!

> It looks like the `getter` does the same thing

👍

> rename `*InjectAdapter` to something like `*Setter`. like `HttpHeadersInjectAdapter ` to  `HttpHeadersSetter`

👍 

Can you move them to private nested classes inside of their respective Tracers at the same time? (I don't think they generally need visibility beyond their Tracer)

I'd propose to give them all the same class name (especially once they private nested inside of their Tracers), e.g. PropagationSetter / PropagationGetter

> Most of  `*Setter` are used as singletons. Maybe we can use `enum ` instead of  `class ` like the following code?

👍 (I think @anuraaga and @iNikem wanted to do this too https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/920#discussion_r467687078)

Lastly, I'd wait until #971 is merged, as there is lots of overlapping changes with that PR.

iNikem:
I am on the edge about using enum. I don't see any actual benefits comparing to `public static final` field. But the only argument against that I have is "tiny bit more surprising to the reader"

[...]

PR #4522, opened 2021-10-27
Convert TextMapSetters and TextMapGetters to enums

iNikem:
All effectively singletons setters and getters were converted to enums. A couple of them which accept a list of propagators for semi-manual propagation could not be converted and so remained.


iNikem:
Are there any downsides?

iNikem:
Based on https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/995#issuecomment-675145181 and the next several comments, I understood that we want to convert these setters and getters to enums. @anuraaga is clearly in favour of this change. Remaining @open-telemetry/java-instrumentation-maintainers , do you vote in favour, against or abstain?

trask:
> Based on [#995 (comment)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/995#issuecomment-675145181) and the next several comments, I understood that we want to convert these setters and getters to enums

that discussion predated the Instrumenter API

> @anuraaga is clearly in favour of this change. Remaining @open-telemetry/java-instrumentation-maintainers , do you vote in favour, against or abstain?

I would just ask you or @anuraaga to explain the reason for making them singletons, maybe it's obvious, but I am missing it. Then I would be happy to vote in favor.

anuraaga:
I guess the main reason would be just to make sure they're singletons - for example, while there often isn't a use case, a user could initialize a `*Tracing` module multiple times for some reason, and there is some benefit then to keep things singleton where possible. In many cases, there won't be a difference though, but it probably doesn't hurt to guarantee singleton just in case.

trask (approved):

mateuszrzeszutek:
> Remaining @open-telemetry/java-instrumentation-maintainers , do you vote in favour, against or abstain?

I think I'd like to abstain - it seems to me that this is purely a stylistic choice; the setter classes perform exactly the same no matter how many instances exist. I don't mind making them all singletons if we decide so.

Issue #4477, opened 2021-10-22
Avoid expected muzzle failures


trask:
Expected muzzle failures in cases like servlet 2.2/3.0 and netty 4.0/4.1 are confusing to users.

Hoping the approach in #4474 can be extended to other instrumentations, and enforce it by failing tests somehow if any muzzle failures are logged.

trask:
Closing, this has been mostly addressed.

PR #4545, opened 2021-10-29
Fail tests on muzzle failure

laurit:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4477
Perhaps instead of trying to avoid muzzle failures by adding `hasClassesNamed` checks we should instead change muzzle failures from warn to info and accept them as a part of normal execution.

### DIFF
--- a/instrumentation/hibernate/hibernate-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/HibernateInstrumentationModule.java
+++ b/instrumentation/hibernate/hibernate-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/hibernate/v4_0/HibernateInstrumentationModule.java
@@ -5,18 +5,29 @@
 
 package io.opentelemetry.javaagent.instrumentation.hibernate.v4_0;
 
+import static io.opentelemetry.javaagent.extension.matcher.AgentElementMatchers.hasClassesNamed;
 import static java.util.Arrays.asList;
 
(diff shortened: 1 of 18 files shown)


iNikem (approved):

trask (approved):
nice 👍

> Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4477

what is left from #4477?

> Perhaps instead of trying to avoid muzzle failures by adding `hasClassesNamed` checks we should instead change muzzle failures from warn to info and accept them as a part of normal execution.

Info level is on by default, and so I'm afraid users will report these messages as "issues". And our debug log is (expected) very verbose. I think this change is great from a supportability perspective.

laurit:
> what is left from #4477?

This is not enough to catch all of the expected failures, because not all of our tests try to apply all instrumentation versions (for example couchbase). To get better coverage perhaps we should also change muzzle plugin to verify that there are no unexpected failures. Though with this we'd basically be duplicating what muzzle already does by manually writing `classLoaderMatcher` methods. While I think it is good to get rid of the muzzle failures it might require too much effort, it might be better to make muzzle failures that we log look more like informative messages rather than warnings. Perhaps we should use `classLoaderMatcher` where it is easy and let muzzle deal with the rest of the failures.

> Info level is on by default, and so I'm afraid users will report these messages as "issues". And our debug log is (expected) very verbose. I think this change is great from a supportability perspective.

Sorry meant debug instead of info, still not used that info goes to stdout.

mateuszrzeszutek (approved):

PR #4571, opened 2021-11-02
Remove micrometer from armeria-shaded-for-testing

mateuszrzeszutek:
This will become really useful when we implement a micrometer->otel bridge instrumentation


trask (approved):
> This will become really useful when we implement a micrometer->otel bridge instrumentation

🥳

laurit (approved):

PR #4573, opened 2021-11-02
Inject helper resources only once

laurit:
Inject helper resources only when they haven't already been injected into given class loader.


trask (approved):
nice

mateuszrzeszutek (approved):

Issue #4552, opened 2021-10-31
Support JDK 17


harel-e:
**Is your feature request related to a problem? Please describe.**

The latest release (1.7.1) fails to start with JDK 17
I tested with OpenJDK 17 and Oracle JDK 17

```
[otel.javaagent 2021-10-31 19:52:09:746 +0200] [main] INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent - version: 1.7.1
[otel.javaagent 2021-10-31 19:52:11:687 +0200] [main] ERROR io.opentelemetry.javaagent.tooling.HelperInjector - Error preparing helpers while processing class sun.rmi.transport.ObjectTable for rmi. Failed to inject helper classes into instance <bootstrap>
java.lang.IllegalStateException: java.lang.IllegalAccessError: superinterface check failed: class io.opentelemetry.javaagent.instrumentation.rmi.context.server.ContextDispatcher (in unnamed module @0x5ac86ba5) cannot access class sun.rmi.server.Dispatcher (in module java.rmi) because module java.rmi does not export sun.rmi.server to unnamed module @0x5ac86ba5
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingUnsafe$Dispatcher$Enabled.defineClass(ClassInjector.java:2023)
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingUnsafe.injectRaw(ClassInjector.java:1798)

...

Caused by: java.lang.IllegalAccessError: superinterface check failed: class io.opentelemetry.javaagent.instrumentation.rmi.context.server.ContextDispatcher (in unnamed module @0x5ac86ba5) cannot access class sun.rmi.server.Dispatcher (in module java.rmi) because module java.rmi does not export sun.rmi.server to unnamed module @0x5ac86ba5
	at java.base/jdk.internal.misc.Unsafe.defineClass0(Native Method)
	at java.base/jdk.internal.misc.Unsafe.defineClass(Unsafe.java:1325)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #4577, opened 2021-11-03
Rmi instrumentation on jdk17

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4552

### DIFF
--- a/instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
+++ b/instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
@@ -56,7 +56,7 @@ public class ExposeRmiModuleInstrumentation implements TypeInstrumentation {
   public void transform(TypeTransformer transformer) {
     transformer.applyTransformer(
         (builder, typeDescription, classLoader, module) -> {
-          if (!instrumented.get() && module != null && module.isNamed()) {
+          if (module != null && module.isNamed()) {
             // using InstrumentationVersion because it's in the unnamed module in the bootstrap
             // loader, and that's where the rmi instrumentation helper classes will end up
             JavaModule helperModule = JavaModule.ofType(InstrumentationVersion.class);
--- a/instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
+++ b/instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
@@ -23,7 +23,8 @@ import org.slf4j.LoggerFactory;
 public class ExposeRmiModuleInstrumentation implements TypeInstrumentation {
   private static final Logger logger =
       LoggerFactory.getLogger(ExposeRmiModuleInstrumentation.class);
-  private static final AtomicBoolean instrumented = new AtomicBoolean();
+
+  private final AtomicBoolean instrumented = new AtomicBoolean();
 
   @Override
   public ElementMatcher<TypeDescription> typeMatcher() {
@@ -37,6 +38,18 @@ public class ExposeRmiModuleInstrumentation implements TypeInstrumentation {
         };
 
     return notInstrumented.and(nameStartsWith("sun.rmi"));
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
+    return new ElementMatcher.Junction.AbstractBase<ClassLoader>() {
+
+      @Override
+      public boolean matches(ClassLoader target) {
+        // runs only in bootstrap class loader
+        return JavaModule.isSupported() && target == null;
+      }
+    };
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
```diff
@@ -0,0 +1,67 @@
...
+import net.bytebuddy.dynamic.loading.ClassInjector;
+import net.bytebuddy.matcher.ElementMatcher;
+import net.bytebuddy.utility.JavaModule;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ExposeRmiModuleInstrumentation implements TypeInstrumentation {
+  private static final Logger logger =
+      LoggerFactory.getLogger(ExposeRmiModuleInstrumentation.class);
+  private static final AtomicBoolean instrumented = new AtomicBoolean();
```
mateuszrzeszutek, line 26: Nit: it doesn't have to be static, it can be an instance variable.

mateuszrzeszutek (approved):
Nice 👍

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+        new ElementMatcher.Junction.AbstractBase<TypeDescription>() {
+
+          @Override
+          public boolean matches(TypeDescription target) {
+            return !instrumented.get();
+          }
+        };
+
+    return notInstrumented.and(nameStartsWith("sun.rmi"));
+  }
```
trask, line 41: this is a good trick to file away 😄

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+        return JavaModule.isSupported() && target == null;
+      }
+    };
+  }
+
+  @Override
+  public void transform(TypeTransformer transformer) {
+    transformer.applyTransformer(
+        (builder, typeDescription, classLoader, module) -> {
+          if (!instrumented.get() && module != null && module.isNamed()) {
```
trask, line 59: is `!instrumented.get()` needed here since checked in typeMatcher?

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
```diff
@@ -0,0 +1,80 @@
...
+            // unnamed module in bootstrap loader, instead of InstrumentationVersion could use
+            // any class that is in agent bootstrap
+            JavaModule helperModule = JavaModule.ofType(InstrumentationVersion.class);
```
trask, line 62: I'm not following (because I'm still learning to speak JPMS), `InstrumentationVersion` is in bootstrap class loader, and I thought we want to expose `sun.rmi.server` to the helper class we are injecting... oh... typing this out made me realize... it's because we are injecting into the bootstrap class loader, will attempt to update comment slightly so that I will understand when I ask the question next time...
```suggestion
            // using InstrumentationVersion because it's in the unnamed module in the bootstrap
            // loader, and that's where the rmi instrumentation helper classes will end up
            JavaModule helperModule = JavaModule.ofType(InstrumentationVersion.class);
```

trask (approved):
any thoughts what it would take to run the normal rmi instrumentation tests using JPMS?

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
laurit: I did it this way mostly because for every class that matches this gets logged
```
io.opentelemetry.javaagent.tooling.AgentInstaller$TransformLoggingListener - Transformed sun.rmi.server.WeakClassHashMap -- null
```
What I really wanted was to make this module start before `RmiContextPropagationInstrumentationModule` so that `io.opentelemetry.javaagent.instrumentation.rmi.context.server.ContextDispatcher` could access `sun.rmi.server.Dispatcher`. Perhaps a cleaner alternative would be to have some sort of api for declaring dependency on an internal package so this could be handled in `HelperInjector`.

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/jpms/ExposeRmiModuleInstrumentation.java
laurit: removed

PR #4579, opened 2021-11-03
Migrate RMI instrumentation to Instrumenter API

mateuszrzeszutek:
Part of #2713


instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/client/UnicastRefInstrumentation.java
```diff
@@ -46,28 +46,33 @@ public static void onEnter(
...
 
-      // TODO replace with client span check
-      if (!Java8BytecodeBridge.currentSpan().getSpanContext().isValid()) {
+      Context parentContext = Java8BytecodeBridge.currentContext();
+
+      // TODO replace with client span check - this is a bit different condition though, can we
+      // remove it?
+      if (!Java8BytecodeBridge.spanFromContext(parentContext).getSpanContext().isValid()) {
+        return;
+      }
```
trask, line 55: looks like tests pass without it. +1 to remove 🤷‍♂️

trask (approved):

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/client/UnicastRefInstrumentation.java
iNikem: But this is very different condition, no? I read it that we don't want to create top level spans for rmi calls. Nothing in our instrumenters does such check.

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/ContextPayload.java
```diff
@@ -37,7 +34,9 @@ public ContextPayload(Map<String, String> context) {
 
   public static ContextPayload from(Context context) {
     ContextPayload payload = new ContextPayload();
-    tracer().inject(context, payload, SETTER);
+    GlobalOpenTelemetry.getPropagators()
+        .getTextMapPropagator()
+        .inject(context, payload, ContextPayloadSetter.INSTANCE);
```
iNikem, line 39: I think we should add `inject` method to `Instrumenter`. My problem with approach like this is that it is very easy to end up using different instances of `OpenTelemetry` .

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/server/RemoteServerInstrumentation.java
```diff
@@ -43,38 +42,45 @@ public void transform(TypeTransformer transformer) {
...
         @Advice.Local("otelScope") Scope scope) {
+
       callDepth = CallDepth.forClass(RemoteServer.class);
       if (callDepth.getAndIncrement() > 0) {
         return;
       }
 
       // TODO review and unify with all other SERVER instrumentation
       Context parentContext = THREAD_LOCAL_CONTEXT.getAndResetContext();
+      request = ClassAndMethod.create(declaringClass, methodName);
```
iNikem, line 59: You can get an instance of `Method` here, then you don't need a separate `ClassAndMethod` class

mateuszrzeszutek: ByteBuddy returns a new `Method` instance each time the instrumentation is executed -- method name and class are constants. I believe @laurit has found that issue, I can't find the PR for that though

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/ContextPayload.java
mateuszrzeszutek: Can you create an issue for this?

iNikem: Done.

PR #4580, opened 2021-11-03
Migrate opentelemetry-api testing module to Instrumenter API

mateuszrzeszutek:
Part of #2713


trask (approved):

instrumentation/opentelemetry-api/opentelemetry-api-1.0/testing/src/main/java/AgentSpanTestingInstrumenter.java
```diff
@@ -0,0 +1,89 @@
...
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.api.instrumenter.SpanKey;
+import java.lang.reflect.Field;
+
+public final class AgentSpanTestingInstrumenter {
```
iNikem, line 15: I understand that this is testing instrumenter, but shouldn't it still follow all the same patterns that we use for all other instrumenters? E.g. instance methods vs static ones, calling `shouldStart` etc?

mateuszrzeszutek: Hmm, I don't really have an opinion on that. Since it's just test code, I implemented it in the simplest way possible. I can change that to match the common patterns of Instrumenter usage if you want.

iNikem: I think that would be better, but I don't feel very strongly about that.

iNikem (approved):

PR #4587, opened 2021-11-04
Trace DNS resolution in Netty 4.1

mateuszrzeszutek:
Previously, the netty 4.1 instrumentation created a single `CONNECT` span for both DNS resolution and the actual connection. This PR splits it into two spans: `RESOLVE` that captures DNS resolution and `CONNECT` capturing the actual connection.
I'm planning to implement the same change in reactor netty next - it will share a lot of classes with netty 4.1 (e.g. the instrumented `AddressResolverGroup` implementattion). Netty 3.8 and 4.0 don't seem to support any resolution abstraction (other than passing a `InetSocketAddress`) so I won't implement this change there.


instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/NettyClientSingletons.java
```diff
@@ -24,21 +24,21 @@
       Config.get().getBoolean("otel.instrumentation.netty.always-create-connect-span", false);
```
mateuszrzeszutek, line 24: Can we still change property names? According to VERSIONING.md

> Changes to configuration properties (...) will be considered breaking changes (unless they only affect telemetry produced by instrumentation)

This only affects generated telemetry, so maybe it'd be okay to rename it? I've been thinking about renaming this to sth like `otel.instrumentation.<module>.connection-telemetry.enabled`

trask: Yes we can rename since it only affects telemetry shape

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/BootstrapInstrumentation.java
```diff
@@ -31,35 +37,68 @@
...
 
+  @SuppressWarnings("unused")
+  public static class ConstructorAdvice {
+
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void onExit(@Advice.This Bootstrap bootstrap) {
+      // this is already the default value, but we're calling the resolver() method to invoke its
+      // instrumentation
+      bootstrap.resolver(DefaultAddressResolverGroup.INSTANCE);
+    }
```
trask, line 63: below alternative feels less strange, though I'm not sure what protection we get from muzzle for structure dependency like `FieldValue`? e.g. if the field had a different name for netty 4.1.20, would the CI muzzle check catch that? (or we'd only catch it if we ran tests against that version)
```suggestion
    public static void onExit(
        @Advice.FieldValue(value = "resolver", readOnly = false) AddressResolverGroup<?> resolver) {
      resolver = InstrumentedAddressResolverGroup.wrap(connectionInstrumenter(), resolver);
    }
```

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/InstrumentedAddressResolverGroup.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Override
+  public void close() {
+    delegate.close();
+  }
+
+  @Override
+  protected AddressResolver<T> newResolver(EventExecutor eventExecutor) {
+    throw new UnsupportedOperationException("This method should never be called");
```
trask, line 54:
```suggestion
    // this method is called from the super class's implementation of `getResolver` which is
    // overridden by this class
    throw new UnsupportedOperationException("This method should never be called");
```

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettyClientInstrumenterFactory.java
```diff
@@ -46,12 +46,12 @@ public NettyClientInstrumenterFactory(
...
-  public NettyConnectInstrumenter createConnectInstrumenter() {
+  public NettyConnectionInstrumenter createConnectionInstrumenter() {
     NettyConnectNetAttributesExtractor netAttributesExtractor =
         new NettyConnectNetAttributesExtractor();
-    Instrumenter<NettyConnectRequest, Channel> instrumenter =
-        Instrumenter.<NettyConnectRequest, Channel>builder(
-                GlobalOpenTelemetry.get(), instrumentationName, rq -> "CONNECT")
+    Instrumenter<NettyConnectionRequest, Channel> instrumenter =
+        Instrumenter.<NettyConnectionRequest, Channel>builder(
+                GlobalOpenTelemetry.get(), instrumentationName, NettyConnectionRequest::spanName)
```
trask, line 54: I sort of see these as two different instrumenters, one for resolve and one for connect, but ok with this too

trask (approved):
👍

cc @lmolkova as example of more http layering in practice

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/BootstrapInstrumentation.java
mateuszrzeszutek: No protection at all, I'm afraid; ByteBuddy would just fail when applying this particular advice class and apply everything else.

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettyClientInstrumenterFactory.java
mateuszrzeszutek: That was my first idea too, but then I noticed that they're basically the same; so I ended up with just reusing it and changing just the span name.

instrumentation/netty/netty-4.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_1/client/NettyClientSingletons.java
mateuszrzeszutek: Nice. I'll do that in a separate PR, together with docs for the new property.

PR #4591, opened 2021-11-04
Improve muzzle check for constructors

mateuszrzeszutek:
Muzzle should fail if our helper class references a library base class constructor that's absent on the classpath. For example imagine that muzzle codegen has captured references to the following class hierarchy: `Helper extends LibraryBaseClass extends Object`, all with default no-arg constructors. Now, if the actual `LibraryBaseClass` on the application classpath had a constructor that requires a `long` the muzzle check would still pass - because it would find the default constructor in the `Object` class. This PR fixes muzzle so that it recognizes this situation.


trask (approved):
nice

PR #4600, opened 2021-11-05
Update garbage collector instruments to be async counters

jack-berg:
A couple fellow New Relic engineers, Mike LaSpina and Erika Arnold, noticed that the garbage collection metrics are being reported as gauges when the values being observed are actually cumulative counts.

This PR fixes that, and as a result, causes the metrics to be exported as sums by default instead of gauges.


trask (approved):
👍

laurit (approved):

PR #4603, opened 2021-11-06
Enable per-try tracing for OkHttp3

lmolkova:


lmolkova:
oops, this was a prototype, I did not intend to create the PR here

PR #4604, opened 2021-11-06
Trace SSL handshakes in netty 4.1

mateuszrzeszutek:
This PR adds telemetry to netty's SSL handshakes. It only includes netty 4.1 changes -- I'll implement the same thing in netty 4.0 and reactor-netty once this PR is merged. 

CC @lmolkova

### DIFF
--- a/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettySslInstrumentationHandler.java
+++ b/instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettySslInstrumentationHandler.java
@@ -19,7 +19,6 @@ public final class NettySslInstrumentationHandler extends ChannelDuplexHandler {
   private Context parentContext;
   private NettySslRequest request;
   private Context context;
-  private boolean handshakeDone = false;
 
   public NettySslInstrumentationHandler(NettySslInstrumenter instrumenter) {
     this.instrumenter = instrumenter;
@@ -31,14 +30,6 @@ public final class NettySslInstrumentationHandler extends ChannelDuplexHandler {
     // this happens inside Bootstrap#connect()
     parentContext = Context.current();
     ctx.fireChannelRegistered();
-  }
-
-  @Override
-  public void channelReadComplete(ChannelHandlerContext ctx) {
-    if (!this.handshakeDone) {
-      ctx.read();
-    }
-    ctx.fireChannelReadComplete();
   }
 
   @Override
@@ -68,7 +59,6 @@ public final class NettySslInstrumentationHandler extends ChannelDuplexHandler {
   @Override
   public void userEventTriggered(ChannelHandlerContext ctx, Object evt) {
     if (evt instanceof SslHandshakeCompletionEvent) {
-      this.handshakeDone = true;
       if (ctx.pipeline().context(this) != null) {
         ctx.pipeline().remove(this);
       }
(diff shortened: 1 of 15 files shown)


instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ClientSslTest.groovy
```diff
@@ -0,0 +1,225 @@
...
+          attributes {
+            "${SemanticAttributes.NET_TRANSPORT.key}" IP_TCP
+            "${SemanticAttributes.NET_PEER_NAME.key}" uri.host
+            "${SemanticAttributes.NET_PEER_PORT.key}" uri.port
+            "${SemanticAttributes.NET_PEER_IP.key}" { it == null || it == "127.0.0.1" }
+          }
+        }
+        span(3) {
+          name "SSL handshake"
+          kind INTERNAL
```
trask, line 178: should `RESOLVE`, `CONNECT` and `SSL handshake` be `CLIENT` spans? (users would need to enable `outgoing-span-suppression-by-type=true`, or we could consider making suppression-by-type the default)

sort of related, what do you think of converting the "error-only" spans into span events on the current span? this way those would still get captured even when using the "no nested CLIENT spans" setting

I can create an issue to discuss both of these further since they have broader impact than just this PR

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettySslInstrumentationHandler.java
```diff
@@ -0,0 +1,84 @@
...
+    // remember the parent context from the time of channel registration;
+    // this happens inside Bootstrap#connect()
+    parentContext = Context.current();
+    ctx.fireChannelRegistered();
+  }
+
+  @Override
+  public void channelReadComplete(ChannelHandlerContext ctx) {
+    if (!this.handshakeDone) {
+      ctx.read();
```
trask, line 39: can you add a comment about why we need to call `read()` and potentially why it's safe to do so (since more than just observing)?

instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ClientSslTest.groovy
mateuszrzeszutek: > I can create an issue to discuss both of these further since they have broader impact than just this PR

Yes, please! There's a lot of things going on here that we could discuss before establishing a "common pattern" for tracing these things.

> should `RESOLVE`, `CONNECT` and `SSL handshake` be `CLIENT` spans?

I think it makes sense for them to be `CLIENT` spans. I was following the current implementation, which used `INTERNAL` everywhere though.

> users would need to enable `outgoing-span-suppression-by-type=true`, or we could consider making suppression-by-type the default)

In plain netty the connection phase and actually sending the HTTP request are completely separate; so all those low-level spans are not the children of the `CLIENT` span. In case of "normal"/higher-level HTTP clients I think it'd make perfect sense for them to be nested under the HTTP `CLIENT` span.

> sort of related, what do you think of converting the "error-only" spans into span events on the current span? this way those would still get captured even when using the "no nested CLIENT spans" setting

Hmm, do we want that in all cases? Or only when nested spans are disabled?
Using spans for these operations (as opposing to events) enables gathering timing and exception information; e.g. you know that SSL handshake took 10 seconds and threw an exception, and it failed before the HTTP request was sent.

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/client/NettySslInstrumentationHandler.java
mateuszrzeszutek: This one turned out to be not needed at all; it's a remnant of the reactor-netty's `SslReadHandler` implementation, which handles things in a slightly different way.

instrumentation/netty/netty-4.1/javaagent/src/test/groovy/Netty41ClientSslTest.groovy
trask: 👍 created issue to continue discussion #4617

trask (approved):

PR #4605, opened 2021-11-06
Support ratpack functional tests

jsalinaspolo:
By making `public` `OpenTelemetryExecInterceptor` and `OpenTelemetryServerHandler` allows to manually initialize RatpackTracing. 

Initializing manually `OpenTelemetry` adds support for Ratpack functional tests because we can create impositions and/or access the Ratpack registry to get the `SpanExporter`.

Also, being able to decide the position in the Ratpack chain of `OpenTelemetryServerHandler` helps not to require custom `Sampler` to ignore handlers where instrumentation is not needed.

This PR, will allow to add instrumentation to Ratpack HttpClient

### DIFF
--- a/instrumentation/ratpack-1.4/library/build.gradle.kts
+++ b/instrumentation/ratpack-1.4/library/build.gradle.kts
@@ -7,7 +7,7 @@ dependencies {
   library("io.ratpack:ratpack-core:1.4.0")
 
   testImplementation(project(":instrumentation:ratpack-1.4:testing"))
-  testImplementation("io.ratpack:ratpack-guice:1.4.0")
+  testLibrary("io.ratpack:ratpack-guice:1.4.0")
 
   if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_11)) {
     testImplementation("com.sun.activation:jakarta.activation:1.2.2")
--- a/instrumentation/ratpack-1.4/library/build.gradle.kts
+++ b/instrumentation/ratpack-1.4/library/build.gradle.kts
@@ -7,6 +7,7 @@ dependencies {
   library("io.ratpack:ratpack-core:1.4.0")
 
   testImplementation(project(":instrumentation:ratpack-1.4:testing"))
+  testImplementation("io.ratpack:ratpack-guice:1.4.0")
 
   if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_11)) {
     testImplementation("com.sun.activation:jakarta.activation:1.2.2")
--- a/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
+++ b/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
@@ -40,7 +40,7 @@ public final class RatpackTracing {
     return new RatpackTracingBuilder(openTelemetry);
   }
 
-  private final OpenTelemetryServerHandler serverHandler;
+  public final OpenTelemetryServerHandler serverHandler;
 
   RatpackTracing(Instrumenter<Request, Response> serverInstrumenter) {
     serverHandler = new OpenTelemetryServerHandler(serverInstrumenter);
--- a/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
+++ b/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
@@ -8,7 +8,6 @@ package io.opentelemetry.instrumentation.ratpack;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 import ratpack.exec.ExecInterceptor;
-import ratpack.handling.Handler;
 import ratpack.handling.HandlerDecorator;
 import ratpack.http.Request;
 import ratpack.http.Response;
@@ -48,7 +47,7 @@ public final class RatpackTracing {
     serverHandler = new OpenTelemetryServerHandler(serverInstrumenter);
   }
 
-  public Handler getOpenTelemetryServerHandler() {
+  public OpenTelemetryServerHandler getOpenTelemetryServerHandler() {
     return serverHandler;
   }
 
--- a/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
+++ b/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
@@ -7,6 +7,8 @@ package io.opentelemetry.instrumentation.ratpack;
 
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import ratpack.exec.ExecInterceptor;
+import ratpack.handling.Handler;
 import ratpack.handling.HandlerDecorator;
 import ratpack.http.Request;
 import ratpack.http.Response;
@@ -40,10 +42,18 @@ public final class RatpackTracing {
     return new RatpackTracingBuilder(openTelemetry);
   }
 
-  public final OpenTelemetryServerHandler serverHandler;
+  private final OpenTelemetryServerHandler serverHandler;
 
   RatpackTracing(Instrumenter<Request, Response> serverInstrumenter) {
     serverHandler = new OpenTelemetryServerHandler(serverInstrumenter);
+  }
+
+  public Handler getOpenTelemetryServerHandler() {
+    return serverHandler;
+  }
+
+  public ExecInterceptor getOpenTelemetryExecInterceptor() {
+    return OpenTelemetryExecInterceptor.INSTANCE;
   }
 
   /** Configures the {@link RegistrySpec} with OpenTelemetry. */
--- a/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
+++ b/instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
@@ -47,20 +47,12 @@ public final class RatpackTracing {
     serverHandler = new OpenTelemetryServerHandler(serverInstrumenter);
   }
 
-  /**
-   * This method returns the {@link OpenTelemetryServerHandler} instance to support Ratpack Registry binding
-   *
-   * @return The OpenTelemetryServerHandler instance
-   */
+  /** Returns instance of {@link OpenTelemetryServerHandler} to support Ratpack Registry binding. */
   public OpenTelemetryServerHandler getOpenTelemetryServerHandler() {
     return serverHandler;
   }
 
-  /**
-   * This method returns the {@link ExecInterceptor} instance to support Ratpack Registry binding
-   *
-   * @return The ExecInterceptor instance
-   */
+  /** Returns instance of {@link ExecInterceptor} to support Ratpack Registry binding. */
   public ExecInterceptor getOpenTelemetryExecInterceptor() {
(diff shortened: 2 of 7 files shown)


instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy
```diff
@@ -0,0 +1,105 @@
...
+class OpenTelemetryModule extends AbstractModule {
+  @Override
+  protected void configure() {
+    bind(SpanExporter).toInstance(InMemorySpanExporter.create())
+  }
+
+  @Singleton
+  @Provides
+  OpenTelemetryServerHandler ratpackServerHandler(OpenTelemetry openTelemetry) {
+    return RatpackTracing.create(openTelemetry).serverHandler
```
iNikem, line 71: Should this be just `new OpenTelemetryServerHandler(openTelemetry)`?

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
```diff
@@ -40,7 +40,7 @@ public static RatpackTracingBuilder builder(OpenTelemetry openTelemetry) {
     return new RatpackTracingBuilder(openTelemetry);
   }
 
-  private final OpenTelemetryServerHandler serverHandler;
+  public final OpenTelemetryServerHandler serverHandler;
```
iNikem, line 43: This change is quite suspicious in my opinion. Why do we want this field public?

instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy
jsalinaspolo: `OpenTelemetryServerHandler` requires `Instrumenter<Request, Response>` 

Using `RatpackTracingBuilder` we are consistent and allow the user to decide calling `configureServerRegistry` or manually bind `OpenTelemetryExecInterceptor` and `OpenTelemetryServerHandler`

instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
jsalinaspolo: Making `public` the Handler, allows the user to manually `bind` the Handler in Ratpack chain and not having to use `configureServerRegistry` that enforce to have the `Handler` prepended

```registry.add(HandlerDecorator.prepend(serverHandler));```

Also, it allows to add an imposition of `opentelemetry` in the functional tests to use `InMemorySpanExporter` and be able to assert on it

Comment (laurit) [reply instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java:43]: What do you think of adding an getter method instead of making the field public. If you make the getter return `Handler` instead of `OpenTelemetryServerHandler` you can keep that class as package private. You could use the same approach of exposing `OpenTelemetryExecInterceptor`.

Comment (jsalinaspolo) [reply instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java:43]: I am not sure if I am following. I can make the `getter` in `RatpackTracing` but the class `OpenTelemetryServerHandler` and `OpenTelemetryExecInterceptor` needs to be public. Otherwise, can not be binded manually, like it is done [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4605/files#diff-8b624528a353047ed474284b3c872de150725789538671005043a02626ecd27fR93) and [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4605/files#diff-8b624528a353047ed474284b3c872de150725789538671005043a02626ecd27fR99)

Comment (iNikem) [reply instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java:43]: Actually you don't need those classes to be public for your use-case. You return `ServerHandler` and `ExecInterceptor` and use them. The actual implementation classes don't need to be public

**Comment by iNikem** [comment]
Our library instrumentation support for Ratpack "assumed" that all clients can and will use `RatpackTracing.configureServerRegistry` to "hook" OpenTelemetry into Ratpack machinery. If you require more granular approach by having the separate access to Otel `ServerHandler` and `ExecInterceptor`, you can get them from `RatpackTracing` without making classes/fields public :)

File: instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java (hunk 2 above)
Comment (jsalinaspolo) [reply instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java:43]: The problem is that we are not gonna be able to bind them and add impositions in the functional tests.
`OpenTelemetryServerHandler` and `OpenTelemetryExecInterceptor` are Ratpack specific, so I am not sure why it is a problem to make them accesible.

The Handler, we could get it from the `Registry` and add it to the chain, but the `OpenTelemetryExecInterceptor` only can be binded by calling `RatpackTracing.configureServerRegistry` outside of the Registry, which will make not possible to add an imposition.

Am I missing something?

**Comment by jsalinaspolo** [comment]
Thanks, @iNikem, for the feedback.

I've changed the implementation to use the `getters`, even though I think that it gets a little bit more complicated [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4605/files#diff-8b624528a353047ed474284b3c872de150725789538671005043a02626ecd27fR111)

At the end, `OpenTelemetryServerHandler` and `OpenTelemetryExecInterceptor` are Ratpack specific. Would not be better to allow users to call `RatpackTracing.configureServerRegistry ` or manual binding of these Ratpack specific implementations? 

Otherwise, it looks like a work-around and fragile; those getters are only there to be used by the work-around. I am not sure if these `getters` will be removed at some point, thinking they are not required?

**Comment by iNikem** [comment]
> Thanks, @iNikem, for the feedback.
> 
> I've changed the implementation to use the `getters`, even though I think that it gets a little bit more complicated [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4605/files#diff-8b624528a353047ed474284b3c872de150725789538671005043a02626ecd27fR111)
> 
> At the end, `OpenTelemetryServerHandler` and `OpenTelemetryExecInterceptor` are Ratpack specific. Would not be better to allow users to call `RatpackTracing.configureServerRegistry ` or manual binding of these Ratpack specific implementations?
> 
> Otherwise, it looks like a work-around and fragile design; those getters are only there to be used by the work-around.

We want to have as small public surface as possible. The less public classes/methods we have, the lesser is our maintenance burden and the more changes we can make without breaking backward compatibility. That is why we had only 1 public class, `RatpackTracing` (plus its builder, but that does not count), which did everything needed inside its single method `configureServerRegistry`. 

I understand that your actual use-case is this, am I correct?:

> being able to decide the position in the Ratpack chain of OpenTelemetryServerHandler 

Can this be achieved without exposing its class? e.g. by saying somehow "RatpackTracing, please insert start tracing from this position in the chain"?

File: instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy [hunk 3]
```diff
@@ -0,0 +1,118 @@
...
+          Guice.registry { bindings ->
+            bindings
+              .module(OpenTelemetryModule)
+          }
+        )
+        .handlers { chain ->
+          chain
+            .get("ignore") { ctx -> ctx.render("ignored") }
+            .all {
+              it.insert(it.get(RatpackTracing).getOpenTelemetryServerHandler())
```
anuraaga, line 111: From what I understand, the desire for exposing the `OpenTelemetryServerHandler` class is to be able to reference it directly from the Guice DI context here. Is that correct?

I'm curious whether it's valid to have the handler both in the DI context and here - won't the handler get applied twice?

instrumentation/ratpack-1.4/library/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ dependencies {
   library("io.ratpack:ratpack-core:1.4.0")
 
   testImplementation(project(":instrumentation:ratpack-1.4:testing"))
+  testImplementation("io.ratpack:ratpack-guice:1.4.0")
```
mateuszrzeszutek, line 10: Shouldn't it be a `testLibrary` instead? (version gets updated to the latest one when `latestDepTest` is executed)
```suggestion
  testLibrary("io.ratpack:ratpack-guice:1.4.0")
```

instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy
jsalinaspolo: No, it will not.
The handler need to be instanciated in Guice DI context (aka Ratpack registry), but the Handler needs to be added to the Ratpack Chain.

```it.get(RatpackTracing).getOpenTelemetryServerHandler()``` gets `RatpackTracing` from the Registry and access the Handler instance

File: instrumentation/ratpack-1.4/library/build.gradle.kts (hunk 4 above)
Comment (jsalinaspolo) [reply instrumentation/ratpack-1.4/library/build.gradle.kts:10]: Ok, I was not familiar with `testLibrary` - it is something custom of the repository?

File: instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy (hunk 3 above)
Comment (anuraaga) [reply instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy:111]: Ok - I think to allow usage through DI like you want it makes sense to make the classes, but not the constructor, public. So we could have the classes and the current getters that return the concrete type. Does that sound ok?

Comment (jsalinaspolo) [reply instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy:111]: That sounds great @anuraaga . I think it makes sense becauase then we enforce to construct them via `RatpackTracing`  🙇

File: instrumentation/ratpack-1.4/library/build.gradle.kts (hunk 4 above)
Comment (mateuszrzeszutek) [reply instrumentation/ratpack-1.4/library/build.gradle.kts]: Yeah, `library`/`testLibrary` are custom gradle configurations we use to facilitate testing against different versions of the instrumented lib.

File: instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy (hunk 3 above)
Comment (jsalinaspolo) [reply instrumentation/ratpack-1.4/library/src/test/groovy/io/opentelemetry/instrumentation/ratpack/server/RatpackServerApplicationTest.groovy]: @anuraaga applied the changes, now binding the Handler in Ratpack Chain is nicer but the constructions is enforced via `RatpackTracing `

File: instrumentation/ratpack-1.4/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java [hunk 5]
```diff
@@ -46,6 +47,14 @@ public static RatpackTracingBuilder builder(OpenTelemetry openTelemetry) {
     serverHandler = new OpenTelemetryServerHandler(serverInstrumenter);
   }
 
+  public OpenTelemetryServerHandler getOpenTelemetryServerHandler() {
```
anuraaga, line 50: Can you add some simple javadoc?

anuraaga (approved):
Thanks!

iNikem (approved):

trask:
thx @jsalinaspolo!

PR #4607, opened 2021-11-08
Automatic AWS library instrumentor

steven-aerts:
Like AWS X-Ray, provide an instrumentor which automatically registers
opentelemetry instrumentation in the AWS SDK without any code changes.
Those instrumentors are separate libraries published as
opentelemetry-aws-sdk-1.11-instrumentor and opentelemetry-aws-sdk-2.2-instrumentor


instrumentation/aws-sdk/aws-sdk-1.11/library-instrumentor/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
```
anuraaga, line 1: Can you call the new projects / artifacts here `-autoconfigure` instead of `-instrumentor`? It's sort of a naming scheme we have in OTel.

instrumentation/aws-sdk/aws-sdk-1.11/library-instrumentor/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+  id("otel.library-instrumentation")
+}
+
+base.archivesName.set("${base.archivesName.get()}-instrumentor")
+project.description = "Instruments calls to AWS services made with AWS SDK for Java clients by adding a tracing client as a request handler."
```
anuraaga, line 6: We don't have descriptions in most of our projects. I'd like to leave them out for this PR rather than having just a couple of libraries with descriptions

instrumentation/aws-sdk/aws-sdk-2.2/library-instrumentor/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+plugins {
```
anuraaga, line 1: Can we try importing this module in the javaagent module and deleting the duplicate interceptor? Don't think we can do it for aws sdk v1 since there is some custom logic there but guess we may as well for v2 if possible.

anuraaga (commented):
Thanks @steven-aerts! Sorry for the huge delay on reviewing

instrumentation/aws-sdk/aws-sdk-2.2/library-instrumentor/build.gradle.kts
steven-aerts: It took me some time to work around the java-agent package shader but it works now.

anuraaga (approved):
Thanks!

instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsSdkTracingBuilder.java
```diff
@@ -6,16 +6,20 @@
...
+  private static final boolean CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES_DEFAULT =
+      Config.get().getBoolean("otel.instrumentation.aws-sdk.experimental-span-attributes", false);
 
   private final OpenTelemetry openTelemetry;
 
   private boolean captureExperimentalSpanAttributes;
 
   AwsSdkTracingBuilder(OpenTelemetry openTelemetry) {
     this.openTelemetry = openTelemetry;
+    this.captureExperimentalSpanAttributes = CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES_DEFAULT;
```
trask, line 22: other library instrumentations only allow programmatic configuration of `captureExperimentalSpanAttributes`, I'd suggest doing the same here for consistency (setting this from the autoconfigure module), and opening an issue to discuss doing this across all instrumentations if there's interest

trask (approved):
thx @steven-aerts!

instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsSdkTracingBuilder.java
steven-aerts: Hi @trask ,

The `captureExperimentalSpanAttributes` could already be configured by these system properties/environment variables.  I only moved this mechanism to make sure this can be shared between the java-agent and the autoconfigure library.  
I understand that the default for the builder are now also changing based on these variables.
Would it be a solution if I add an extra method to the `AwsSdkTracingBuilder`, which is then called by both the java-agent as the autoconfigure class?
Or do you prefer the PR where I copy this behaviour to all other classes which have these `Config` variables for the `java-agent`?

trask: > Would it be a solution if I add an extra method to the `AwsSdkTracingBuilder`, which is then called by both the java-agent as the autoconfigure class?

`AwsSdkTracingBuilder` already has `setCaptureExperimentalSpanAttributes`, can you call this from both javaagent and autoconfigure modules? I think that would be most consistent with other instrumentations currently.

steven-aerts: @trask I tried to update it, while still keeping the Config parameter in one place for v1.1.
Let me know what you think.

instrumentation/aws-sdk/aws-sdk-1.11/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v1_11/AwsSdkTracing.java
```diff
@@ -37,6 +40,16 @@ public static AwsSdkTracing create(OpenTelemetry openTelemetry) {
...
 
+  /**
+   * Returns a new {@link AwsSdkTracing} configured with the given {@link OpenTelemetry} and from
+   * the global agent {@link Config}.
+   */
+  public static AwsSdkTracing fromAgentConfig(OpenTelemetry openTelemetry) {
+    return builder(openTelemetry)
+        .setCaptureExperimentalSpanAttributes(CAPTURE_EXPERIMENTAL_SPAN_ATTRIBUTES_DEFAULT)
+        .build();
+  }
```
trask, line 51: I don't think we want to expose "agent" methods in library instrumentation, can the agent call `setCaptureExperimentalSpanAttributes` directly instead?

steven-aerts: @trask I inlined the function on the two places where it is used.
This currently means that also `Config.get().getBoolean("otel.instrumentation.aws-sdk.experimental-span-attributes", false))` is duplicated.
Let me know if you prefer this to be put in a constant on the `AwsSdkTracing` class or its builder.

trask: thx! we have some related work to do (and will likely end up where you were going originally with adding Config default directly to library instrumentation) but appreciate keeping it consistent with existing instrumentation for now 👍

PR #4611, opened 2021-11-08
Add version to the InstrumenterBuilder and Instrumenter

### LINKED ISSUE #4599: Hardcoded Tracer version in Instrumenter
In https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/d314d76fce9400976d1dfb8c326ab3d9394991d4/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java#L84-L85

The `Instrumenter` constructor has a hardcoded version to retrieve the `Tracer`. While this works fine for instrumentation libraries inside the project, it forces external libraries to use the same hardcoded version if you want to retrieve the same `Tracer` used by your `Instrumenter`.

Have the `Tracer` version be part of the `InstrumenterBuilder` so it can be set by a consumer. If not set in the builder it defaults to the current behavior (use the OTel version).

None.

**Additional context**
I'm happy to provide a PR if this is an acceptable change.

radcortez:
- Fixes #4599

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
@@ -58,7 +58,34 @@ public class Instrumenter<REQUEST, RESPONSE> {
       OpenTelemetry openTelemetry,
       String instrumentationName,
       SpanNameExtractor<? super REQUEST> spanNameExtractor) {
-    return new InstrumenterBuilder<>(openTelemetry, instrumentationName, spanNameExtractor);
+    return new InstrumenterBuilder<>(
+        openTelemetry, instrumentationName, InstrumentationVersion.VERSION, spanNameExtractor);
+  }
+
+  /**
+   * Returns a new {@link InstrumenterBuilder}.
+   *
+   * <p>The {@code instrumentationName} is the name of the instrumentation library, not the name of
+   * the instrument*ed* library. The value passed in this parameter should uniquely identify the
+   * instrumentation library so that during troubleshooting it's possible to pinpoint what tracer
+   * produced problematic telemetry.
+   *
+   * <p>The {@code instrumentationVersion} is the version of the instrumentation library, not the
+   * version of the instrument*ed* library.
+   *
+   * <p>In this project we use a convention to encode the minimum supported version of the
+   * instrument*ed* library into the instrumentation name, for example {@code
+   * io.opentelemetry.apache-httpclient-4.0}. This way, if there are different instrumentations for
+   * different library versions it's easy to find out which instrumentations produced the telemetry
+   * data.
+   */
+  public static <REQUEST, RESPONSE> InstrumenterBuilder<REQUEST, RESPONSE> builder(
+      OpenTelemetry openTelemetry,
+      String instrumentationName,
+      String instrumentationVersion,
+      SpanNameExtractor<? super REQUEST> spanNameExtractor) {
+    return new InstrumenterBuilder<>(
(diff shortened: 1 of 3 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -58,7 +58,17 @@
       OpenTelemetry openTelemetry,
       String instrumentationName,
       SpanNameExtractor<? super REQUEST> spanNameExtractor) {
-    return new InstrumenterBuilder<>(openTelemetry, instrumentationName, spanNameExtractor);
+    return new InstrumenterBuilder<>(
+        openTelemetry, InstrumentationVersion.VERSION, instrumentationName, spanNameExtractor);
```
laurit, line 62: I think you have switched name and version here

iNikem: @radcortez Please add new tests as well

iNikem (changes requested):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
radcortez: Ops.

iNikem (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -58,7 +58,17 @@
       OpenTelemetry openTelemetry,
       String instrumentationName,
       SpanNameExtractor<? super REQUEST> spanNameExtractor) {
-    return new InstrumenterBuilder<>(openTelemetry, instrumentationName, spanNameExtractor);
+    return new InstrumenterBuilder<>(
+        openTelemetry, instrumentationName, InstrumentationVersion.VERSION, spanNameExtractor);
+  }
+
+  public static <REQUEST, RESPONSE> InstrumenterBuilder<REQUEST, RESPONSE> builder(
```
mateuszrzeszutek, line 65: Can you add some Javadoc to this method? Just copy everything from the 3-arg `builder()` and mention that `instrumentationVersion` refers to the instrumentation library version, not instrumented library version.

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
radcortez: Sure.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -58,7 +58,35 @@
...
+   * <p>The {@code instrumentationVersion} is the version of the instrumentation library, not the
+   * version of the instrument*ed* library.
+   *
+   * <p>In this project we use a convention to encode the minimum supported version of the
+   * instrument*ed* library into the instrumentation name, for example {@code
+   * io.opentelemetry.apache-httpclient-4.0}. This way, if there are different instrumentations for
+   * different library versions it's easy to find out which instrumentations produced the telemetry
+   * data.
+   */
+
```
laurit, line 82:
```suggestion

```
We don't usually have a blank line between javadoc and method declaration

laurit (approved):

trask:
Thanks @radcortez!

Issue #4593, opened 2021-11-04
SqlSanitizer removes double-quoted "delimited" table and field names


dboreham:
## Background
SQL, as defined by [SQL92](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt) and as implemented by the popular relational database servers, defines a syntax for identifier names (including the names of tables and fields) that is somewhat bizarre by modern standards. Identifiers specified in the regular way are subject to implementation-dependent case folding. For example:
`select * from foo` in Oracle is interpreted as `select * from FOO` while `select * from FOO` in PostgreSQL is interpreted as `select * from foo'.

SQL92 defines a double-quoted _delimited_ _identifier_ syntax the purpose of which is to override any case folding applied to identifiers. For example `select * from "Foo"` is interpreted as `select * from Foo`, preserving the mixed case name.

This is the only use of double quotes allowed in SQL92. String literals are delimited with single quotes like: `select * from foo where bar = 'baz'` and string literals delimited with double quotes constitute a syntax error.

Popular practice is to avoid the use of mixed case in database schema identifiers in order to avoid the need to use delimited identifiers. However, its use is not prohibited. It turns out that some ORMs do not support automatic case folding between camel-case application code identifiers and database table and field names. Therefore deployments using mixed-case database identifiers exist, and as a result all the queries executed against those databases feature double-quoted identifiers. Queries in these systems look like:
```
select "userId" from "adminUsers" where "lastName" = 'jones'
```

## The Problem
OpenTelemetry's database statement sanitizer function, the purpose of which is to remove potentially sensitive information from logged query data, removes delimited identifiers. The result is that the example query above is captured as:
```
select ? from ? where ? = ?
```
This is not the desired outcome because insufficient query information is captured to be practically useful. Operators of such a system would like instead to see:
```
select "userId" from "adminUsers" where "lastName" =  ?
```
The desired behavior can be achieved by sanitizing only single-quoted strings, rather than both single and double-quoted strings.

## The Exception to the Rule
[...]

PR #4595, opened 2021-11-04
SQL92 and CouchDB sanitizers (proof of concept)

dboreham:
This is an example of a solution for https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4593
A parameterized sanitizer would be a better solution (in this PR I just cloned the original into a second CouchDB-specific sanitizer). I couldn't figure out quickly how to parameterize a jflex parser.

PR #4615, opened 2021-11-09
Sql sanitizer: sanitize double quoted strings only in couchbase queries

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4593
While looking at https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4595 I started thinking about how we could avoid copying the whole sanitizer and came up with this.


trask (approved):
👍

@dboreham can you give this a look over? thx!

dboreham (approved):
This looks good. It's what I wanted to do but didn't have the codebase knowledge/time to code up. Thanks!

Issue #4612, opened 2021-11-09
NoClassDefFoundError: Could not initialize class io.opentelemetry.javaagent.instrumentation.netty.v4_0.AttributeKeys


qiansheng91:
My application had started failure when I used the OpenTelemetry agent.

Here are the pieces of information: 

Here is my java agent version:
![image](https://user-images.githubusercontent.com/81615037/140847289-8ad7f7e5-7a67-4aa4-a84c-d6c7751c46af.png)

Here is  the error message:
```
 ERROR 15548 [     parallel-5] reactor.core.scheduler.Schedulers        : Scheduler worker in group main failed with an uncaught exception
java.lang.NoClassDefFoundError: Could not initialize class io.opentelemetry.javaagent.instrumentation.netty.v4_0.AttributeKeys
	at io.netty.channel.AbstractChannel.writeAndFlush(AbstractChannel.java:305)
	at io.lettuce.core.protocol.DefaultEndpoint.channelWriteAndFlush(DefaultEndpoint.java:342)
	at io.lettuce.core.protocol.DefaultEndpoint.writeToChannelAndFlush(DefaultEndpoint.java:282)
	at io.lettuce.core.protocol.DefaultEndpoint.write(DefaultEndpoint.java:142)
	at io.lettuce.core.protocol.CommandExpiryWriter.write(CommandExpiryWriter.java:112)
	at io.lettuce.core.RedisChannelHandler.dispatch(RedisChannelHandler.java:184)
	at io.lettuce.core.StatefulRedisConnectionImpl.dispatch(StatefulRedisConnectionImpl.java:152)
	at io.lettuce.core.RedisPublisher$RedisSubscription.dispatchCommand(RedisPublisher.java:390)
	at io.lettuce.core.RedisPublisher$CommandDispatch$1.dispatch(RedisPublisher.java:477)
	at io.lettuce.core.RedisPublisher$RedisSubscription.checkCommandDispatch(RedisPublisher.java:385)
	at io.lettuce.core.RedisPublisher$State$2.request(RedisPublisher.java:543)
	at io.lettuce.core.RedisPublisher$RedisSubscription.request(RedisPublisher.java:242)
	at reactor.core.publisher.MonoNext$NextSubscriber.request(MonoNext.java:102)
	at reactor.core.publisher.MonoFlatMap$FlatMapInner.onSubscribe(MonoFlatMap.java:230)
[...]

PR #4626, opened 2021-11-10
Ensure that netty 4.0 instrumentation is not applied to 4.1

laurit:
Hopefully resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4612
Not completely sure how the linked issue happened, perhaps there are netty jars from different versions on the classpath, perhaps something else. This pr adds checking of method references to muzzle so that usages like `AttributeKey::new` would also be verified.

### DIFF
--- a/instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
+++ b/instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
@@ -7,6 +7,7 @@ package io.opentelemetry.javaagent.instrumentation.netty.v4_0;
 
 import static io.opentelemetry.javaagent.instrumentation.netty.v4_0.client.NettyClientSingletons.connectionInstrumenter;
 import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.returns;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
 import io.netty.channel.ChannelFuture;
@@ -31,7 +32,9 @@ public class BootstrapInstrumentation implements TypeInstrumentation {
   @Override
   public void transform(TypeTransformer transformer) {
     transformer.applyAdviceToMethod(
-        named("doConnect").and(takesArgument(0, SocketAddress.class)),
+        named("doConnect")
+            .and(takesArgument(0, SocketAddress.class))
+            .and(returns(named("io.netty.channel.ChannelFuture"))),
         BootstrapInstrumentation.class.getName() + "$ConnectAdvice");
   }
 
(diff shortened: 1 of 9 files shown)


instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
```diff
@@ -31,7 +32,9 @@
   @Override
   public void transform(TypeTransformer transformer) {
     transformer.applyAdviceToMethod(
-        named("doConnect").and(takesArgument(0, SocketAddress.class)),
+        named("doConnect")
+            .and(takesArgument(0, SocketAddress.class))
+            .and(returns(named("io.netty.channel.ChannelFuture"))),
```
laurit, line 37: In netty 4.1 this method returns `void`

trask (commented):
can you try adding `testInstrumentation` dependencies also, to cross-test the netty instrumentations against each other?

muzzle/src/test/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceCollectorTest.java
```diff
@@ -132,7 +132,26 @@ public void invokedynamicCreatesReferences() {
...
+        "()V",
+        PROTECTED_OR_HIGHER,
+        OwnershipFlag.STATIC);
+    assertThat(references).containsKey("muzzle.TestClasses$MethodBodyAdvice$A");
+    assertMethod(
+        references.get("muzzle.TestClasses$MethodBodyAdvice$A"),
+        "<init>",
+        "()V",
+        PROTECTED_OR_HIGHER,
+        OwnershipFlag.NON_STATIC);
```
trask, line 154: 👍

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/ReferenceCollectingClassVisitor.java
```diff
@@ -435,13 +436,35 @@ public void visitInvokeDynamicInsn(
...
                       computeMinimumClassAccess(
-                          refSourceType, Type.getObjectType(handle.getOwner())))
-                  .build());
+                          refSourceType, Type.getObjectType(handle.getOwner())));
+
+          if (handle.getTag() == Opcodes.H_INVOKEVIRTUAL
+              || handle.getTag() == Opcodes.H_INVOKESTATIC
+              || handle.getTag() == Opcodes.H_INVOKESPECIAL
+              || handle.getTag() == Opcodes.H_NEWINVOKESPECIAL
+              || handle.getTag() == Opcodes.H_INVOKEINTERFACE) {
```
trask, line 450: 😬 👍

trask (approved):
nice

mateuszrzeszutek (approved):
Nice 👍

PR #4630, opened 2021-11-10
fix incorrect delegate method

### LINKED ISSUE #4628: Instrumentation version on application tracer not set correctly when using agent
**Describe the bug**
When an application using the Java agent creates a tracer, the instrumentation version is incorrectly set as the schema URL instead. Browsing the source code for the Java agent, it appears there is a copy'n'paste bug on [this line](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/8b7c09791283addefb3687ab9025c2eb1b2cf052/instrumentation/opentelemetry-api/opentelemetry-api-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/v1_4/trace/ApplicationTracerBuilder.java#L28)
It appears
```java
agentTracerBuilder.setSchemaUrl(s);
```
should instead be
```java
agentTracerBuilder.setInstrumentationVersion(s);
```

**Steps to reproduce**
Using code similar to the following for an application run with the Java agent results in spans created by the application to have incorrect instrumentation library attributes:
```java
OpenTelemetry openTelemetry = GlobalOpenTelemetry.get();
TracerBuilder tracerBuilder = openTelemetry.tracerBuilder("my-library");
tracerBuilder.setInstrumentationVersion("1.0.0");
Tracer tracer = tracerBuilder.build();
```

I expected to see spans (in Jaeger) created by the application to have the tag `otel.library.name` with value `my-library` and tag `otel.library.version` with value `1.0.0`

**What did you see instead?**
I see spans created by the application with tag `otel.library.name` as expected, but no tag `otel.library.version`.
On further investigation using a Java debugger, I see that fields for a `SdkTracerBuilder` instance are incorrectly set, where field `instrumentationVersion` is null and `schemaUrl` is `1.0.0`.
Note that spans created by the agent are fine, they have expected values for tags `otel.library.name` and `otel.library.version`.

Java agent version 1.7.2

**Environment**
Compiler: openjdk version "11.0.11" 2021-04-20
OS: Ubuntu 20.04
Runtime (if different from JDK above): OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OS (if different from OS compiled on): Windows 10, running WSL 2

**Additional context**
My application is actually written in Clojure, so I've had to simplify some of the details above. Apologies in advance if my Java syntax is a little off!

breedx-splk:
resolves #4628 

I tried writing a test, but the classloading stuff was too perilous and I gave up.

### DIFF
--- a/instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
+++ b/instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
@@ -336,7 +336,7 @@ class TracerTest extends AgentInstrumentationSpecification {
   // this test uses opentelemetry-api-1.4 instrumentation
   def "test tracer builder"() {
     when:
-    def tracer = GlobalOpenTelemetry.get().tracerBuilder("test").build()
+    def tracer = GlobalOpenTelemetry.get().tracerBuilder("test").setInstrumentationVersion("1.2.3").build()
     def testSpan = tracer.spanBuilder("test").setSpanKind(PRODUCER).startSpan()
     testSpan.end()
 
@@ -347,6 +347,7 @@ class TracerTest extends AgentInstrumentationSpecification {
           name "test"
           kind PRODUCER
           hasNoParent()
+          instrumentationLibraryVersion "1.2.3"
           attributes {
           }
         }
--- a/instrumentation/opentelemetry-api/opentelemetry-api-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/v1_4/trace/ApplicationTracerBuilder.java
+++ b/instrumentation/opentelemetry-api/opentelemetry-api-1.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/v1_4/trace/ApplicationTracerBuilder.java
@@ -18,14 +18,14 @@ class ApplicationTracerBuilder implements TracerBuilder {
   }
 
   @Override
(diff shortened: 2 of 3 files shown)


laurit:
I believe you can modify this test https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/4719e4cc797b06d0b6e24cf7ab3e84f3d2d961af/instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy#L337-L347
to verify the fix. Just add something like
```groovy
assert it.span.instrumentationLibraryInfo.name == "test"
assert it.span.instrumentationLibraryInfo.version == "1.0"
```
inside the `span(0) {}` block and set the version on tracer builder.

laurit (approved):

breedx-splk:
> I believe you can modify this test <snip>

Interesting. I was just trying to write a vanilla unit test, but that really did fail miserably, so this seems like a good consolation prize. I got it going, thanks for the suggestion.

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/test/groovy/TracerTest.groovy
```diff
@@ -347,6 +347,7 @@ class TracerTest extends AgentInstrumentationSpecification {
           name "test"
           kind PRODUCER
           hasNoParent()
+          instrumentationLibraryVersion "1.2.3"
```
trask, line 350: 👍

trask (approved):

anuraaga (approved):

mateuszrzeszutek (approved):

Issue #4617, opened 2021-11-09
How to model network telemetry under HTTP client spans


trask:
From @mateuszrzeszutek and my discussion in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4604#discussion_r744148786

> should `RESOLVE`, `CONNECT` and `SSL handshake` be `CLIENT` spans?

I think it makes sense for them to be `CLIENT` spans. I was following the current implementation, which used `INTERNAL` everywhere though.

> users would need to enable `outgoing-span-suppression-by-type=true`, or we could consider making suppression-by-type the default)

In plain netty the connection phase and actually sending the HTTP request are completely separate; so all those low-level spans are not the children of the `CLIENT` span. In case of "normal"/higher-level HTTP clients I think it'd make perfect sense for them to be nested under the HTTP `CLIENT` span.

> sort of related, what do you think of converting the "error-only" spans into span events on the current span? this way those would still get captured even when using the "no nested CLIENT spans" setting

Hmm, do we want that in all cases? Or only when nested spans are disabled?
Using spans for these operations (as opposing to events) enables gathering timing and exception information; e.g. you know that SSL handshake took 10 seconds and threw an exception, and it failed before the HTTP request was sent.

cc @lmolkova @denisivan0v

trask:
Closing, I think we have reasonable clarity on this now (and further work should be done in semantic-convention repo)

PR #4635, opened 2021-11-12
Trace SSL handshakes in netty 4.0

mateuszrzeszutek:
Continuation of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4604

### DIFF
--- a/instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
+++ b/instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
@@ -7,10 +7,9 @@ package io.opentelemetry.javaagent.instrumentation.netty.v4_0;
 
 import static io.opentelemetry.javaagent.instrumentation.netty.v4_0.client.NettyClientSingletons.connectionInstrumenter;
 import static net.bytebuddy.matcher.ElementMatchers.named;
-import static net.bytebuddy.matcher.ElementMatchers.returns;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
-import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelPromise;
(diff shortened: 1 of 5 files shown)


instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
```diff
@@ -32,16 +31,16 @@
   @Override
   public void transform(TypeTransformer transformer) {
     transformer.applyAdviceToMethod(
-        named("doConnect")
-            .and(takesArgument(0, SocketAddress.class))
-            .and(returns(named("io.netty.channel.ChannelFuture"))),
+        named("doConnect0")
```
iNikem, line 34: Just out of curiosity - why this change?

iNikem (commented):
Have we already discussed having this as event as opposed to a span? Is there any separation between start/end for that span?

instrumentation/netty/netty-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/v4_0/BootstrapInstrumentation.java
mateuszrzeszutek: The `doConnect` actually traces more than just connection - it starts with tracing channel registration (which happens before) and ends after handler `channelActive()` has returned (which is way after SSL handshake is finished). The `doConnect0` method seems to be doing just the connection, which is more accurate.

mateuszrzeszutek:
> Have we already discussed having this as event as opposed to a span? Is there any separation between start/end for that span?

We've briefly touched upon this idea in the last PR, and we have https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4617 where we can discuss it further. To sum up the previously mentioned points: SSL handshake takes time and has non-zero duration, and can fail - we can record the exception and set the span status.

iNikem (approved):

Issue #4629, opened 2021-11-10
Kotlin coroutines + Java Http Client + otel agent generating memory leak


brunojcm:
**Describe the bug**
I'm currently using the Java Http Client thorugh the ktor client (I think ktor is irrelevant here, any coroutines code would cause the same problem), and when the otel java agent is installed, I have a memory leak.

**Steps to reproduce**
1. Create a Ktor client like this:
```
val client = HttpClient(Java.create())
val r = client.get("$baseUrl/any/endpoint")
```
2. Run multiple http calls using the client

**What did you expect to see?**
Many http request calls with no memory leak.

**What did you see instead?**
Many `JobImpl` instances leak:
<img width="1526" alt="image" src="https://user-images.githubusercontent.com/251695/141166468-a5bb9906-7be2-4b74-9f7a-4b843b4e4fc2.png">

GC Root goes all the way back to the first job created:
<img width="780" alt="image" src="https://user-images.githubusercontent.com/251695/141166535-5691b7a9-fac0-4296-b2f7-770c334e1a3b.png">
<img width="892" alt="image" src="https://user-images.githubusercontent.com/251695/141166561-06352977-8b07-40ec-89ba-b4154e42ad5c.png">

**What version are you using?**
`opentelemetry-javaagent-all-1.7.2`

**Environment**
Compiler: Temurin 11.0.13
OS: macOS

**Additional context**

Weirdly enough, even though the object leaking seems to be related to coroutines (`JobImp`), if I add `-Dotel.instrumentation.java-http-client.enabled=false` to my startup command, the leak stops.

I also tried various combinations of `-Dotel.instrumentation.executor.enabled=false` and `-Dotel.instrumentation.kotlinx-coroutines.enabled=false`, no luck, the trigger for the issue seems to be `java-http-client` instrumentation.

brunojcm:
I just tested the Ktor client with with the `ktor-client-apache` and the `ktor-client-cio` engines, none of those causes the memory leak. `ktor-client-apache` even uses Apache Http, which is also intrumented, but for some reason that does not trigger the bug.

brunojcm:
I just realized Apache and CIO clients are not generating spans (CIO is expected, but I thought Apache should be instrumented), so maybe that's why we don't see the memory leak when those are used.

laurit:
A small sample app that reproduces the issue would help a lot.

brunojcm:
I'll try to work on that this week.

[...]

PR #4637, opened 2021-11-13
Fix memory leak when using ktor-client-java

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4629

### DIFF
--- a/instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/BodyHandlerWrapper.java
+++ b/instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/BodyHandlerWrapper.java
@@ -30,7 +30,7 @@ public class BodyHandlerWrapper<T> implements BodyHandler<T> {
     if (subscriber instanceof BodySubscriberWrapper) {
       return subscriber;
     }
-    return new BodySubscriberWrapper<>(delegate.apply(responseInfo), context);
+    return new BodySubscriberWrapper<>(subscriber, context);
   }
 
   public static class BodySubscriberWrapper<T> implements BodySubscriber<T> {


instrumentation/java-http-client/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpclient/BodyHandlerWrapper.java
```diff
@@ -30,7 +30,7 @@ public BodyHandlerWrapper(BodyHandler<T> delegate, Context context) {
     if (subscriber instanceof BodySubscriberWrapper) {
       return subscriber;
     }
-    return new BodySubscriberWrapper<>(delegate.apply(responseInfo), context);
+    return new BodySubscriberWrapper<>(subscriber, context);
```
iNikem, line 33: Can you please explain, why the previous version caused the memory leak?

laurit: On the line before the diff there is `BodySubscriber<T> subscriber = delegate.apply(responseInfo);` Apparently in case of ktor the result of `delegate.apply` which is a JavaHttpResponseBodySubscriber https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpResponseBodyHandler.kt can't just be thrown away without closing it. Anyway I believe that I'm the author of this bug, it should have looked like this from the start.

iNikem (approved):

mateuszrzeszutek (approved):

trask (approved):

brunojcm:
any changes of having this backported to 1.7.x?

laurit:
> any changes of having this backported to 1.7.x?

Considering that the release of 1.9 is imminent (yeah there is a jump from 1.7 to 1.9) any more releases in 1.7 series are unlikely.

brunojcm:
when is it planned for?

on our perspective, it's always harder to pitch for a major upgrade in comparison to a bug fix one, so it would be nice to have such a safe change backported.

let me know if there is anything I can do to make it happen.

trask:
this issue does fit our [patch release criteria](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/RELEASING.md#patch-release), so I have added it to tomorrow's SIG meeting agenda to discuss

1.9.0 is targeted for end of this week, but currently looking likely to slip into next week

brunojcm:
Thank you very much for chasing that :)

trask:
@brunojcm I tried to release a patch, but our patch release workflow is broken https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/4268334982?check_suite_focus=true, I think because maven snapshots were recently pruned https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/javaagent/opentelemetry-javaagent/, and our build process was not resilient to that. The problem is solvable in the 1.7.x branch, but I'm not sure it's the best use of time given that we've done a lot of work in the upcoming release to improve our build process (in particular #4248) which should avoid this issue (and others we've had related to releasing patches) going forwards.

brunojcm:
No worries, thanks for trying, @trask!

Looking forward to any release containing this fix, I can do the extra work on my side to do a major update, I'd have to do it eventually anyway.

trask:
@brunojcm 1.9.0 is released, thanks again for reporting this!

brunojcm:
thank you very much, @trask, I'll give it a try today!

PR #4650, opened 2021-11-16
Added Http4SpanDecorator to apache-camel

osherv:


instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/Http4SpanDecorator.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.camel.Endpoint;
+import org.apache.camel.Exchange;
+
+class Http4SpanDecorator extends HttpSpanDecorator {
+  @Nullable
+  @Override
+  protected String getHttpUrl(Exchange exchange, Endpoint endpoint) {
+    Object url = exchange.getIn().getHeader(Exchange.HTTP_URL);
+    if (url instanceof String) {
+      return ((String) url).replace("http4", "http");
```
laurit, line 18: In https://camel.apache.org/components/2.x/http4-component.html they also mention urls starting with `https4://`

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/Http4SpanDecorator.java
```diff
@@ -0,0 +1,33 @@
...
+package io.opentelemetry.javaagent.instrumentation.apachecamel.decorators;
+
+import javax.annotation.Nullable;
+import org.apache.camel.Endpoint;
+import org.apache.camel.Exchange;
+
+class Http4SpanDecorator extends HttpSpanDecorator {
+  @Nullable
+  @Override
+  protected String getHttpUrl(Exchange exchange, Endpoint endpoint) {
```
iNikem, line 15: This is mostly a copy-paste from the super-class. Can you instead introduce a new protected method into `HttpSpanDecorator`, something like `getProtocol`. Super-class will return `http`, this class can override it and return `http4`.

Even easier way is, probably, to pass this protocol prefix as constructor parameter to `HttpSpanDecorator`

osherv: Hey good point! thanks!
Fixed

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/Http4SpanDecorator.java
osherv: Nice catch! Thanks!
fixed

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/Http4SpanDecorator.java
```diff
@@ -10,23 +10,17 @@
...
-      } else {
-        // Try to obtain from endpoint
-        int index = endpoint.getEndpointUri().lastIndexOf("http4:");
-        if (index != -1) {
-          return endpoint.getEndpointUri().substring(index).replace("http4", "http");
-        }
-      }
+    String url = super.getHttpUrl(exchange, endpoint);
+    if (url != null) {
+      return url.replace(getProtocol(), getProtocol().substring(0, getProtocol().length() - 1));
```
iNikem, line 23: Why is this still needed?

osherv: This is where the replacement from ("http4"/"https4") to "http/https" happens.
It's only uniq to http4 protocol, so it has to be done here i think

iNikem: But why this replacement is needed? If Camel uses such protocols, shouldn't we preserve this information?

osherv: Maybe we should, but the actual url of the request  is not http4, but http.
"http4://url" is not a valid address. Apache translates the url to "http://url".
I thought that the "http.url" tag should hold the real URL..
Maybe we could add another tag? Or you're thinking that it should remain http.url="http4://some-url"?

iNikem: I am actually not sure that I understand what is going on here at all :) To me it seems that `io.opentelemetry.javaagent.instrumentation.apachecamel.decorators.HttpSpanDecorator#getHttpUrl` method should return url that is present in either exchange or endpoint. If this is the url that was requested from Camel, then I would think that whatever protocol that url uses should remain in `http.url` attribute value without modifications.

@kubawach can you please express your opinion as the original author of Camel integration?

osherv: That's an intersting point, i'm actually not sure what is the right answer XD
Let's wait to @kubawach opinion than.
Thanks @iNikem for noice that!

trask: I think `http.url` should only be used for `http` and `https` schemes. E.g. we wouldn't put `ftp://host` into `http.url`.

If we want this instrumentation to follow http semantic conventions (and suppress underlying http client instrumentation), then I think it's ok to coerce the scheme into one of `http` or `https`, if we know that's what is produced by the library anyways.

Alternatively, we could make this instrumentation emit INTERNAL spans and not suppress underlying http client instrumentation.

My guess is that the former (following http semantic conventions) is probably more helpful for users.

iNikem: > I think `http.url` should only be used for `http` and `https` schemes. E.g. we wouldn't put `ftp://host` into `http.url`.

Why? :) If my http request understands different schemes and I make a request to `mySchema://blab-bla`, I would expect to see that full url in `http.url` attribute of that span.

E.g. `feed://example.com/rss.xml` for rss reader or `webcal://example.com/calendar.ics` for a calendar. Or there is even old [`shttp` schema](https://en.wikipedia.org/wiki/Secure_Hypertext_Transfer_Protocol)

trask: > > I think `http.url` should only be used for `http` and `https` schemes. E.g. we wouldn't put `ftp://host` into `http.url`.
> 
> Why? :) If my http request understands different schemes and I make a request to `mySchema://blab-bla`, I would expect to see that full url in `http.url` attribute of that span.
> 
> E.g. `feed://example.com/rss.xml` for rss reader or `webcal://example.com/calendar.ics` for a calendar. Or there is even old [`shttp` schema](https://en.wikipedia.org/wiki/Secure_Hypertext_Transfer_Protocol)

Can you open a spec issue to clarify / update the spec if there's agreement on that? Currently [the spec says](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#semantic-conventions-for-http-spans):

> This document defines semantic conventions for HTTP client and server Spans. They can be used for http and https schemes and various HTTP versions like 1.1, 2 and SPDY.

osherv: @iNikem In your examples the protocols are not http tho. Am i wrong? It is the same as websocket:// or grpc://
which in these cases the use of 'http.url' is incorrect because they are not http urls but another urls of another protocols. 
In the case of 'http4' it is only apache camel component and it tells the program when to ask for external http resources.
Anyway, i'd be happy to update the issue if i need to. The issue is still relevant evan without the replacement because we are not catch http4 urls at all today.
What is the decision? :)

osherv: @trask @iNikem 
Hey :) any updates?

iNikem: @osherv @trask I created a spec issue: https://github.com/open-telemetry/opentelemetry-specification/issues/2171. Let's wait couple of days for the initial response.

mateuszrzeszutek: From what I understand `http4` is the name of the [camel component](https://camel.apache.org/components/2.x/http4-component.html), not the protocol/scheme - these are separate concepts. The actual used HTTP scheme used is still either `http` or `https`.
WDYT about having both `getProtocol()` (maybe renamed to `getScheme()` to match OTEL naming) and `getComponent()` methods, and using the second for searching in endpointUri and the first for constructing `http.url`?

osherv: @mateuszrzeszutek That is correct.

trask: @iNikem what are your latest thoughts on this? I can see the benefit of capturing the url as it was passed in by the user (in this case `http4://`). But I also see the benefit of modeling the actual downstream request (in this case `http://`).

Since it could some time to get feedback on your spec issue, let's just pick one so we can unblock @osherv, and I will open a tracking issue in this repo to make sure we sync up once there's clarification in the spec.

iNikem: If we are going to follow up after spec clarification, then it does not matter for me which of the two options we choose now. We can go ahead with what is already in this PR.

iNikem (approved):

instrumentation/apache-camel-2.20/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachecamel/decorators/Http4SpanDecorator.java
kubawach: Just found this notification so to reinforce @mateuszrzeszutek findings - the idea was that http is the actual url, while http4 only camel component.

PR #4663, opened 2021-11-17
Start using Gradle Enterprise instance

iNikem:


settings.gradle.kts
```diff
@@ -28,13 +29,13 @@ dependencyResolutionManagement {
 val isCI = System.getenv("CI") != null
 val skipBuildscan = System.getenv("SKIP_BUILDSCAN").toBoolean()
 gradleEnterprise {
+  server = "https://ge.opentelemetry.io"
```
laurit, line 32: Previously scan was published only when building from ci but this check is now removed. Does this mean that scan is uploaded even for local builds?

trask (approved):
🥳

(+1 to @laurit's question above)

settings.gradle.kts
```diff
@@ -14,8 +14,9 @@ pluginManagement {
 }
 
 plugins {
-  id("com.gradle.enterprise") version "3.6.3"
+  id("com.gradle.enterprise") version "3.7.1"
   id("com.github.burrunan.s3-build-cache") version "1.2"
+  id("com.gradle.common-custom-user-data-gradle-plugin") version "1.2.1"
```
anuraaga, line 19: https://github.com/gradle/gradle-enterprise-build-config-samples/tree/master/common-custom-user-data-gradle-plugin

```
A tag representing the operating system
A tag representing how the build was invoked, be that from your IDE (IDEA, Eclipse, Android Studio) or from the command-line
A tag representing builds run on CI, together with a set of tags, links and custom values specific to the CI server running the build
For Git repositories, information on the commit id, branch name, status, and whether the checkout is dirty or not
```

I'm not sure what "tag representing OS" means here, but it seems a bit scary for an OSS project, naturally we will be publishing scans from local corp laptops sometimes to troubleshoot builds. This plugin smells like it's intended for corporate, not OSS use.

Can we leave it out for now?

settings.gradle.kts
iNikem: That is Gradle's suggestion, yes. The goal is to better understand builds (especially build failures) on local machines. If we are concerned about privacy we may change that to publish scans only on failures.

settings.gradle.kts
iNikem: Sure, but I will also ask Gradle people to comment here.

settings.gradle.kts
anuraaga: I think it's too surprising if OSS builds phone home without an opt-in. Eventually someone will notice and complain hard :P

iNikem: Ok, will change to publish only on CI or if opted in

settings.gradle.kts
iNikem: @anuraaga As publish from local machine now require opt-in, do you still want to remove this plugin?

anuraaga: It's probably OK (I'm guessing OS tag is just "mac" type of string)

anuraaga (approved):

settings.gradle.kts
runningcode: It will add `Mac OS X` to your builds (amongst other tags). it is very useful to debug issues to know if a build came from Mac or windows for example because line endings or encoding may cause issues.
Here is an example of what that tagging looks like on the JUnit Gradle Enterprise OSS instance: https://ge.junit.org/s/katto6gtipkwe#info

runningcode:
```suggestion
  id("com.gradle.common-custom-user-data-gradle-plugin") version "1.5"
```

settings.gradle.kts
```diff
@@ -26,15 +27,15 @@ dependencyResolutionManagement {
...
 
 val isCI = System.getenv("CI") != null
-val skipBuildscan = System.getenv("SKIP_BUILDSCAN").toBoolean()
+val publishBuildscan = System.getenv("PUBLISH_BUILDSCAN").toBoolean()
 gradleEnterprise {
+  server = "https://ge.opentelemetry.io"
   buildScan {
-    termsOfServiceUrl = "https://gradle.com/terms-of-service"
-    termsOfServiceAgree = "yes"
+    publishAlwaysIf(isCI || publishBuildscan)
```
runningcode, line 34:
```suggestion
    publishAlways()
    publishIfAuthenticated()
```

iNikem: @runningcode Do we need both `always` and `ifAuthenticated`?

runningcode: Yes, its undocumented but this is the recommended set up for OSS projects.

mateuszrzeszutek (approved):

trask:
@iNikem is this expected prior to merging this PR?

```
A build scan was not published as you have not authenticated with server 'ge.opentelemetry.io'.
```

iNikem:
> @iNikem is this expected prior to merging this PR?
> 
> ```
> A build scan was not published as you have not authenticated with server 'ge.opentelemetry.io'.
> ```

Unfortunately. As our GE instance is public, it is recommended (by Gradle folks and, probably, common sense) to allow only authenticated users to publish build scans. This means only builds initiated by our repo (and not PRs from forks, because GH secrets are not passed to them) and maintainers who wish to configure authentication for their local machines. This limits the benefits of GE, but we don't know a way around that.

We could continue to publish build scan for all other builds to `scans.gradle.org` as we did before, but as it was pointed out to me this means that we accept Gradle scans' TOS in the name of anybody who happens to run a build/create a PR. Which is not legally right thing to do. But maybe doing so only for PRs is OK?

trask:
> Unfortunately. As our GE instance is public, it is recommended (by Gradle folks and, probably, common sense) to allow only authenticated users to publish build scans. This means only builds initiated by our repo (and not PRs from forks, because GH secrets are not passed to them) and maintainers who wish to configure authentication for their local machines. This limits the benefits of GE, but we don't know a way around that.

got it. I think that's ok since all of the non-PR builds will still give us plenty of builds to track flaky tests across. and we'll still get the gradle cache benefits for PRs and local builds.

> We could continue to publish build scan for all other builds to `scans.gradle.org` as we did before, but as it was pointed out to me this means that we accept Gradle scans' TOS in the name of anybody who happens to run a build/create a PR. Which is not legally right thing to do. But maybe doing so only for PRs is OK?

this is a good question.

we currently recommend reviewing the gradle scan for troubleshooting PR build failures: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/CONTRIBUTING.md#troubleshooting-pr-build-failures

for this PR, can you keep publishing PR build scans to `scans.gradle.org`, since that was prior behavior, and send a separate PR to remove it (and to remove the above doc)? then we can request CNCF legal review, and bring it back depending on their guidance

iNikem:
Current state:

- If GE authentication present, then publish to ge.opentelemetry.io
- if GE authentication is not present and we are in CI, then publish to scans.gradle.com (will be removed in future)
- if GE cache username is present and we are in CI, then push to remote GE cache

anuraaga:
Oh I didn't know this would only work with explicit opt-in via adding credentials. In that case we can have the scan published automatically if you'd like, it's not implicit phone home like I thought before.

iNikem:
My initial version was "publish always" :) But turned out, it could not work :) So yes, only if authenticated.

Issue #4659, opened 2021-11-17
How Embed extensions used in jdk8 ?


chenjazz:
Embed extensions in the OpenTelemetry Agent (opentelemetry-java-instrumentation\examples\extension\build\libs\opentelemetry-javaagent.jar ) can only used in jdk 11+ version. How can I build it,then I can use in jdk8 project. 
![image](https://user-images.githubusercontent.com/11512994/142159865-0a660ebc-6101-4ca9-83ea-3ed0ba2b857e.png)

When I change gradle `options.release.set(8)`  or jdk version to build agent,it fail:
![image](https://user-images.githubusercontent.com/11512994/142160070-a2b3b370-e7be-4952-b86d-5ce4b974326d.png)
 Please help me ,thank you.

trask:
hi @chenjazz, thanks for reporting this!

I think it would be better for the examples to target Java 8 for maximum compatibility (and as a good test to ensure extensions are supported on Java 8). I have sent a PR to update them #4665.

laurit:
@chenjazz currently when you set `options.release.set(8)` you'll still need to build with jdk 11

chenjazz:
@trask @laurit Thank you 👍

trask:
@chenjazz is your issue resolved now?

PR #4665, opened 2021-11-18
Build and test extensions with Java 8

trask:
Related to #4659

### DIFF
--- a/examples/distro/build.gradle
+++ b/examples/distro/build.gradle
@@ -47,7 +47,7 @@ subprojects {
     }
 
     compileJava {
-      options.release.set(11)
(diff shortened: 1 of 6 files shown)


anuraaga (approved):

iNikem (approved):
😿

mateuszrzeszutek (approved):

laurit (approved):
I think the real issue in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4659 is that extension doesn't build with jdk8. I guess it could be solved by stating that jdk11 is required in readme.

PR #4667, opened 2021-11-18
Drop instrumentation-api-caching module and move weak cache implementation to instrumentation-api

iNikem:
This if first draft PR to verify my understand of what we agreed upon on Tuesday.

We have couple of usages that still don't work with `WeakLockFreeCache`:

- `FutureListenerWrappers` wants weak values
- `ClassLoaderHasClassesNamedMatcher` and `AgentCachingPoolStrategy` want to specify max capacity

How did we want to solve that?


iNikem:
The bigger problem is `AgentCachingPoolStrategy.SharedResolutionCacheAdapter.find`, which currently cannot work with comparison-based cache keys

mateuszrzeszutek:
I think `FutureListenerWrappers` should actually work correctly with a `VirtualField` now - a couple of weeks ago @laurit has introduced a lambda instrumentation that makes them go through `ClassFileTransformer`.

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/ClassLoaderHasClassesNamedMatcher.java
```diff
@@ -5,15 +5,16 @@
...
 import io.opentelemetry.javaagent.instrumentation.api.internal.ClassLoaderMatcherCacheHolder;
 import io.opentelemetry.javaagent.instrumentation.api.internal.InClassLoaderMatcher;
 import net.bytebuddy.matcher.ElementMatcher;
 
 class ClassLoaderHasClassesNamedMatcher extends ElementMatcher.Junction.AbstractBase<ClassLoader> {
 
+  //TODO .setMaximumSize(25)
   private final Cache<ClassLoader, Boolean> cache =
-      Cache.builder().setWeakKeys().setMaximumSize(25).build();
+      Cache.builder().build();
```
mateuszrzeszutek, line 17: Maybe we could just remove the max size here? This usage is very similar to the one in `MuzzleMatcher`, and we don't have any limit there.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/Cache.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+/** A cache from keys to values.
+ *
+ * Keys are always referenced weakly and are compared using identity comparison, not
+ * {@link Object#equals(Object)}.
+ */
+public interface Cache<K, V> {
```
mateuszrzeszutek, line 16: Do we want to keep the name `Cache`? It's now just a weak-keyed map, since we removed caffeine et al.

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/ClassLoaderHasClassesNamedMatcher.java
trask: 👍 for just weak keys (no max) here

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/Cache.java
iNikem: I did not want to rename in this PR, to avoid too many changes at once. We can rename later, if we want

iNikem:
Only 2 problems remain.

One is simple: `SqlStatementSanitizer` wants to use a bounded cache. I am not sure if bringing in the whole CLHM just for one use case makes sense.

Another one is more complicated: `AgentCachingPoolStrategy.SharedResolutionCacheAdapter` relies on cache keys being compared by `equal`. But I don't totally grasp why do we need this custom implementation of ByteBuddy cache at all.

trask:
> `SqlStatementSanitizer` wants to use a bounded cache. I am not sure if bringing in the whole CLHM just for one use case makes sense.

This is an important cache for an important use case (jdbc). I'm not sure what alternative we have? I think LRU (CLHM) should be enough, not sure if we need more advanced LFU eviction (caffeine), which would avoid the complexity of caffeine2/3 for now at least (in both library and agent).

> Another one is more complicated: `AgentCachingPoolStrategy.SharedResolutionCacheAdapter` relies on cache keys being compared by `equal`. But I don't totally grasp why do we need this custom implementation of ByteBuddy cache at all.

My guess based on prior work is that this is used to avoid re-analyzing classes during instrumentation (e.g. looking up structure/hierarchy of parent class/interfaces). This can help a lot with startup overhead. I recall the DataDog folks spent a good amount of time tuning this particular cache. I'd be hesitant to mess with it without doing a good amount of testing. I think CLHM should be good here also (no need for more advanced eviction).

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/Cache.java
iNikem: Now there are two different implementations, so we better keep this generic name

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpHeaderAttributes.java
```diff
@@ -6,15 +6,15 @@
...
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
 
 final class HttpHeaderAttributes {
 
-  private static final Cache<String, AttributeKey<List<String>>> requestKeysCache =
-      Cache.builder().setMaximumSize(32).build();
-  private static final Cache<String, AttributeKey<List<String>>> responseKeysCache =
-      Cache.builder().setMaximumSize(32).build();
+  private static final ConcurrentHashMap<String, AttributeKey<List<String>>> requestKeysCache =
```
iNikem, line 14: Should I replace this with bounded cache now?

iNikem:
CLHM uses unsafe and thus fails our jpms tests. Should I switch to `cache2k`? I am a little bit concerned by its size, 400K is not that small.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpHeaderAttributes.java
trask: these are naturally limited, I just pushed comment to this file

instrumentation-api/build.gradle.kts
```diff
@@ -14,27 +14,23 @@ sourceSets {
...
 dependencies {
-  compileOnly(project(":instrumentation-api-caching"))
-
   api("io.opentelemetry:opentelemetry-api")
   api("io.opentelemetry:opentelemetry-semconv")
 
   implementation("io.opentelemetry:opentelemetry-api-metrics")
   implementation("org.slf4j:slf4j-api")
+  implementation("com.blogspot.mydailyjava:weak-lock-free")
+  implementation("com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru:1.4.2")
```
trask, line 29: let's shade these in. less transitive dependencies will make `instrumentation-api` more palatable for libraries to embed directly

trask:
> CLHM uses unsafe and thus fails our jpms tests. Should I switch to `cache2k`? I am a little bit concerned by its size, 400K is not that small.

I keep forgetting all the branches of this problem 😞

This is what using Caffeine 3 on Java 11 would solve for us.

Cache2k seems like a reasonable option to keep things simple for now.

As long as it's all hidden / shaded dependency we can improve on it later.

ben-manes:
CLHM has a backport of CHM v8 which uses unsafe. Most just copy the code into their project and replace with the standard CHM. For example ms-jdbc, groovy, micronaut. This is pretty much equivalent to shading.

trask:
> CLHM has a backport of CHM v8 which uses unsafe. Most just copy the code into their project and replace with the standard CHM. For example ms-jdbc, groovy, micronaut. This is pretty much equivalent to shading.

thx @ben-manes! 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/BoundedCache.java
```diff
@@ -0,0 +1,45 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.cache;
+
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+final class BoundedCache<K, V> implements Cache<K, V> {
```
trask, line 12: `BoundedCache` -> `MapBackedCache`? (since nothing bounded about this particular class)

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/FutureListenerWrappers.java
```diff
@@ -44,21 +46,32 @@ public static boolean shouldWrap(GenericFutureListener<? extends Future<?>> list
...
+                wrapper =
+                    new WrappedProgressiveFutureListener(
+                        context, (GenericProgressiveFutureListener<ProgressiveFuture<?>>) delegate);
+              } else {
+                wrapper =
+                    new WrappedFutureListener(context, (GenericFutureListener<Future<?>>) delegate);
+              }
+              return new WeakReference<>(wrapper);
+            });
+    return resultReference.get();
```
trask, line 64: we may need to be careful here, I think the cache can return a "cleared" weak value reference now (previously caffeine took care of not making "cleared" values visible).

Instead, I think we could use VirtualField instead of Cache, essentially reverting #3059 now that #4182 solves the underlying problem with VirtualField not being applied to lambdas.

trask: I sent two commits converting from Cache to VirtualField
* [Convert Netty wrapper cache to VirtualField](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4667/commits/dac1522a3f6fef096f379554e84e99f206229302)
* [Work around lambda instrumentation failure](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4667/commits/6d63815b44ca4721e9ee417f4f18314ee703bd80)

The second commit was needed to workaround a problem caused by the VirtualField instrumenting some of our own helper class lambdas. Normally I don't think helper classes are instrumented since they are injected from inside of ClassFileTransformer.transform (which is not recursive). But lambdas are different since they don't go through the normal class definition / transformation process.

I was already having second thoughts about the first commit above after typing up the warning about the memory leak that will occur if this code is ever used inside of library instrumentation.

I did add a [commit to handle cleared weak values](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4667/commits/9ab70f4d2aa52abcdb1a27ed917494cfe1276b69).

instrumentation-api/build.gradle.kts
trask: we copied in CLHM in order to update code to use default CHM

I copied in weak-lock-free also (instead of shading), just to avoid gradle shadow problems (only 4 files)

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/BoundedCache.java
anuraaga: Seems like a worthwhile rename

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/Cache.java
```diff
@@ -0,0 +1,59 @@
...
+package io.opentelemetry.instrumentation.api.cache;
+
+import io.opentelemetry.instrumentation.api.cache.concurrentlinkedhashmap.ConcurrentLinkedHashMap;
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+/**
+ * A cache from keys to values.
+ *
+ * <p>Keys are always referenced weakly and are compared using identity comparison, not {@link
```
anuraaga, line 15: Looks like drift

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/concurrentlinkedhashmap/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
anuraaga, line 1: This is the same as our top level LICENSE file so don't need it I think

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/concurrentlinkedhashmap/NOTICE
```diff
@@ -0,0 +1,7 @@
+ConcurrentLinkedHashMap
+Copyright 2008, Ben Manes
```
anuraaga, line 2: I think we should add this line above the Copyright 2010 Google Inc. into the file license headers. We could probably remove the NOTICE then (otherwise I would put the NOTICE in resources, not java)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/weaklockfree/AbstractWeakConcurrentMap.java
```diff
@@ -0,0 +1,362 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
anuraaga, line 4: Need license header like

https://github.com/open-telemetry/opentelemetry-java/blob/main/context/src/main/java/io/opentelemetry/context/internal/shaded/WeakConcurrentMap.java#L6

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/weaklockfree/DetachedThreadLocal.java
```diff
@@ -0,0 +1,141 @@
+/*
```
anuraaga, line 1: A couple of unused files, even when vendoring we may as well leave out completely unused ones I guess. Probably easier to copy from here actually :P

https://github.com/open-telemetry/opentelemetry-java/tree/main/context/src/main/java/io/opentelemetry/context/internal/shaded

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/weaklockfree/WeakConcurrentSet.java
```diff
@@ -0,0 +1,135 @@
+/*
```
anuraaga, line 1: Ditto

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/weaklockfree/AbstractWeakConcurrentMap.java
trask: I didn't find that license header anywhere, but sync'd entire weaklockfree package with sdk repo (so is now included) 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/concurrentlinkedhashmap/NOTICE
anuraaga: By the way I noticed micronaut doesn't reference the NOTICE, nor update license headers

https://github.com/micronaut-projects/micronaut-core/tree/813923882bcfed158d136a27bf8010edf70bb613/core/src/main/java/io/micronaut/core/util/clhm

IANAL, but I think the `author` tag satisfies "readable copy of attribution notices contained within such NOTICE file...within the Source form or documentation"

https://github.com/ben-manes/concurrentlinkedhashmap/blob/master/LICENSE#L111

Mostly not a fan of having license type files within the `java` source - converting to a docstring on a `package-info.java` would also be better to me

anuraaga (approved):
Just point about NOTICE but not a huge deal

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/concurrentlinkedhashmap/NOTICE
trask: > converting to a docstring on a `package-info.java` would also be better to me if needing to include it

went with this 👍 

I also re-generated the license report, which pulls in NOTICE files if they're in the dependencies jar file, but in this case it's not.

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/FutureListenerWrappers.java
```diff
@@ -11,20 +11,25 @@
...
-  // weak keys where original listener is key and wrapper is value. As wrapper has a strong
-  // reference to original listener this causes a memory leak.
-  // Also note that it's ok if the value is collected prior to the key, since this cache is only
-  // used to remove the wrapped listener from the netty future, and if the value is collected prior
-  // to the key, that means it's no longer used (referenced) by the netty future anyways.
+  // note: it's ok if the value is collected prior to the key, since this cache is only used to
+  // remove the wrapped listener from the netty future, and if the value is collected prior to the
+  // key, that means it's no longer used (referenced) by the netty future anyways.
+  //
+  // also note: this is not using VirtualField in case this is ever converted to library
```
laurit, line 22: Alternatively we could add an option to use weak values when `VirtualField` fallback map is used. When we know that `VirtualField` value references key we could just use `VirtualField` with weak values and would not need to worry whether it is going to use fallback map because something wasn't instrumented or because it is used inside library instrumentation.

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/matcher/ClassLoaderHasClassesNamedMatcher.java
```diff
@@ -5,15 +5,14 @@
...
+import io.opentelemetry.instrumentation.api.cache.Cache;
 import io.opentelemetry.javaagent.instrumentation.api.internal.ClassLoaderMatcherCacheHolder;
 import io.opentelemetry.javaagent.instrumentation.api.internal.InClassLoaderMatcher;
 import net.bytebuddy.matcher.ElementMatcher;
 
 class ClassLoaderHasClassesNamedMatcher extends ElementMatcher.Junction.AbstractBase<ClassLoader> {
 
-  private final Cache<ClassLoader, Boolean> cache =
-      Cache.builder().setWeakKeys().setMaximumSize(25).build();
+  private final Cache<ClassLoader, Boolean> cache = Cache.weak();
```
laurit, line 15: I like this, in my opinion 25 was way too low.

laurit (approved):

instrumentation/netty/netty-4-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/netty/common/FutureListenerWrappers.java
trask: I'll open a issue for this 👍

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/Cache.java
```diff
@@ -3,17 +3,34 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-package io.opentelemetry.instrumentation.api.caching;
+package io.opentelemetry.instrumentation.api.cache;
 
+import io.opentelemetry.instrumentation.api.cache.concurrentlinkedhashmap.ConcurrentLinkedHashMap;
```
mateuszrzeszutek, line 8: Should we put CLHM and WeakLockFree in `internal` package(s)? They shouldn't be a part of our public API.

trask: 👍 created #4735 to track

Issue #1131, opened 2020-08-30
Finatra controller spans are not capturing exceptions


trask:
See this comment in `FinatraServerTest.groovy` (also in `FinatraServerLatestTest.groovy`):

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/770502eed9b0b65e0802c4088b0ebef974096b0c/instrumentation/finatra-2.9/src/test/groovy/FinatraServerTest.groovy#L91-L95

PR #4669, opened 2021-11-18
Capture exception on finatra controller span

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1131

### DIFF
--- a/instrumentation/finatra-2.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/finatra/FinatraExceptionManagerInstrumentation.java
+++ b/instrumentation/finatra-2.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/finatra/FinatraExceptionManagerInstrumentation.java
@@ -7,6 +7,7 @@ package io.opentelemetry.javaagent.instrumentation.finatra;
 
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
 import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.returns;
 import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 
 import com.twitter.finagle.http.Response;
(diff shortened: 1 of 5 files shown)


instrumentation/finatra-2.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/finatra/FinatraExceptionManagerInstrumentation.java
```diff
@@ -0,0 +1,49 @@
...
+        isMethod().and(named("toResponse")).and(takesArgument(1, Throwable.class)),
+        this.getClass().getName() + "$HandleExceptionAdvice");
+  }
+
+  @SuppressWarnings("unused")
+  public static class HandleExceptionAdvice {
+
+    @Advice.OnMethodExit(onThrowable = Throwable.class, suppress = Throwable.class)
+    public static void handleException(
+        @Advice.Return Response response, @Advice.Argument(1) Throwable throwable) {
```
mateuszrzeszutek, line 38: WDYT about adding `returns(named("com.twitter.finagle.http.Response"))` to the matcher?

mateuszrzeszutek (approved):

iNikem (approved):

instrumentation/finatra-2.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/finatra/FinatraResponseListener.java
```diff
@@ -23,7 +24,10 @@ public FinatraResponseListener(Context context, Class<?> request) {
 
   @Override
   public void onSuccess(Response response) {
-    instrumenter().end(context, request, null, null);
+    VirtualField<Response, Throwable> virtualField =
+        VirtualField.find(Response.class, Throwable.class);
```
trask, line 28: move to static field?

trask (approved):

instrumentation/finatra-2.9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/finatra/FinatraResponseListener.java
laurit: thanks, moved

PR #4671, opened 2021-11-18
Remove LongAdder and ClassValue usages from instrumentation-api

### LINKED ISSUE #4533: Remove `LongAdder` and `ClassValue` usages from instrumentation-api
**Describe the bug**
Animalsniffer detected that these classes are not supported by Android API level 21 - they're used inside our own code in the `instrumentation-api` module.

**Steps to reproduce**
See #4505

mateuszrzeszutek:
Resolves #4533

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/SupportabilityMetrics.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/SupportabilityMetrics.java
@@ -11,7 +11,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.LongAdder;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/SupportabilityMetrics.java
```diff
@@ -100,44 +100,44 @@ private CounterNames() {}
 
   // this class is threadsafe.
   private static class KindCounters {
-    private final LongAdder server = new LongAdder();
-    private final LongAdder client = new LongAdder();
-    private final LongAdder internal = new LongAdder();
-    private final LongAdder consumer = new LongAdder();
-    private final LongAdder producer = new LongAdder();
+    private final AtomicLong server = new AtomicLong();
```
anuraaga, line 103: I think `LongAdder` is important in normal code. We should have a wrapper around the two types based on classpath check, and can use `@IgnoreJreRequirement` to allow LongAdder

trask: `AtomicLong` counters seem acceptable here since they are only used when agent debug is enabled. Not sure whether we'll need our own high-throughput counters in instrumentation, since typically that kind of work will be delegated to the Metrics API.

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/SupportabilityMetrics.java
anuraaga: Ah thanks for pointing that out - agree if only when debug is enabled, AtomicLong is fine

anuraaga (approved):

PR #4673, opened 2021-11-19
Fix spring-integration context leak

trask:


instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -37,6 +37,23 @@
...
+      // interceptors for every bean name / channel pair, which means it's possible that this
+      // interceptor is added twice to the same channel if the channel is registered twice under
+      // different bean names
+      //
+      // there's an option for this class to implement VetoCapableInterceptor and prevent itself
+      // from being registered if it's already registered, but the VetoCapableInterceptor interface
+      // broke backwards compatibility in 5.2.0, and the version prior to 5.2.0 takes a parameter
+      // of type ChannelInterceptorAware which doesn't exist after 5.2.0, and while it's possible to
+      // implement both at the same time (since we compile using 4.1.0), muzzle doesn't like the
+      // missing class type when running testLatestDeps
```
trask, line 53: I can remove this part if it seems too much, I just didn't want others to go down this road, at least without knowing the pitfalls ahead of time

anuraaga (approved):

laurit (approved):

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
iNikem: This is good description 👍

iNikem (approved):

PR #4677, opened 2021-11-19
Record exception for async servlet invocations

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807 this should be the last remaining instrumentation in that issue.

### DIFF
--- a/instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/JettyServlet3Test.groovy
+++ b/instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/JettyServlet3Test.groovy
@@ -42,9 +42,13 @@ abstract class JettyServlet3Test extends AbstractServlet3Test<Server, ServletCon
     ServletException
   }
 
+  boolean isAsyncTest() {
+    false
+  }
+
   @Override
   boolean hasResponseSpan(ServerEndpoint endpoint) {
-    return (IS_BEFORE_94 && endpoint == EXCEPTION) || super.hasResponseSpan(endpoint)
+    return (IS_BEFORE_94 && endpoint == EXCEPTION && !isAsyncTest()) || super.hasResponseSpan(endpoint)
   }
 
   @Override
@@ -140,9 +144,8 @@ class JettyServlet3TestAsync extends JettyServlet3Test {
   }
 
   @Override
-  boolean testException() {
-    // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
-    return false
+  boolean isAsyncTest() {
+    true
   }
 }
 
@@ -155,8 +158,9 @@ class JettyServlet3TestFakeAsync extends JettyServlet3Test {
 
   @Override
   boolean testException() {
-    // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
-    return false
+    // we expect this request to fail with http 500 but is succeeds with http 200
+    // when using -PtestLatestDeps=true
+    false
   }
 }
 
@@ -224,6 +228,11 @@ class JettyServlet3TestDispatchImmediate extends JettyDispatchTest {
   }
 
   @Override
+  boolean isAsyncTest() {
+    true
+  }
+
+  @Override
   protected void setupServlets(ServletContextHandler context) {
     super.setupServlets(context)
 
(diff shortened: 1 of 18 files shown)


iNikem (approved):

instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/JettyServlet3Test.groovy
```diff
@@ -155,8 +158,9 @@ class JettyServlet3TestFakeAsync extends JettyServlet3Test {
 
   @Override
   boolean testException() {
-    // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
-    return false
+    // we expect this request to fail with http 500 but is succeeds with http 200
+    // when using -PtestLatestDeps=true
```
trask, line 162: do u know if this is a limitation of jetty, our instrumentation, or our tests?

instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/JettyServlet5Test.groovy
```diff
@@ -137,8 +131,8 @@ class JettyServlet5TestFakeAsync extends JettyServlet5Test {
 
   @Override
   boolean testException() {
-    // https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/807
-    return false
+    // we expect this request to fail with http 500 but is succeeds with http 200
```
trask, line 134: same question here

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/AsyncRequestCompletionListener.java
```diff
@@ -35,7 +35,8 @@ public void onComplete(RESPONSE response) {
     if (responseHandled.compareAndSet(false, true)) {
       ServletResponseContext<RESPONSE> responseContext =
           new ServletResponseContext<>(response, null);
-      instrumenter.end(context, requestContext, responseContext, null);
+      Throwable throwable = servletHelper.getAsyncException(requestContext.request());
+      instrumenter.end(context, requestContext, responseContext, throwable);
```
trask, line 39: does an exception thrown from a `Runnable` executed under `AsyncContext.start(Runnable)` always lead to servlet failure?

instrumentation/servlet/servlet-3.0/javaagent/src/test/groovy/JettyServlet3Test.groovy
laurit: It is hard to say. My guess is that our test is weird and could very well depend on unspecified behaviour. Test first completes an AsyncContext and then lets exception propagate out of servlet. It might be better to not complete AsyncContext and rely on timeout (if AsyncContext isn't completed container will close it with timeout). In that case the behaviour should be better defined, but timeout has its own issues. There is always the problem that we'd need to make the timeout as short as possible to not waste time but still have it long enough so that test doesn't timeout before the exception is thrown even when running under load.
The tests that are currently failing also exhibit weirdness, apparently on tomcat10 exception test can break the subsequent test.

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/AsyncRequestCompletionListener.java
laurit: This is a good question. If the request has already been sent then exception can not turn it back. This behaviour is the same for regular servlets. So it is possible to have requests that were ok from http clients perspective, but failed (got an exception) from servers perspective.

trask: good point 👍

trask (approved):
👍

instrumentation/tomcat/tomcat-10.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/tomcat/v10_0/AsyncServlet.groovy
```diff
@@ -37,41 +40,36 @@ class AsyncServlet extends HttpServlet {
...
               break
             case EXCEPTION:
-              resp.status = endpoint.status
-              resp.writer.print(endpoint.body)
-              context.complete()
               throw new ServletException(endpoint.body)
           }
         }
       } finally {
+        // complete at the end so the server span will end after the controller span
```
trask, line 69: 👍

trask:
💯

PR #4682, opened 2021-11-21
Temporary fix for spring-webmvc test latest deps

trask:
Not sure what the problem is, just trying to stabilize `main` (and find out the full scope of instrumentations that are broken in spring boot 2.6.0). Will open tracking issue for real fix.


iNikem (approved):

trask:
Going ahead to merge this so we can see if any other testLatestDeps are failing. Real fix is close behind thx to @laurit #4687 👍

Issue #4683, opened 2021-11-21
Remove testLatestDeps restriction from spring-webmvc instrumentation


trask:
See #4682

PR #4687, opened 2021-11-21
Fix spring webmvc latest dep test

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4683

### DIFF
--- a/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/org/springframework/web/servlet/OpenTelemetryHandlerMappingFilter.java
+++ b/instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/org/springframework/web/servlet/OpenTelemetryHandlerMappingFilter.java
@@ -12,6 +12,9 @@ import io.opentelemetry.instrumentation.api.servlet.ServerSpanNameSupplier;
 import io.opentelemetry.instrumentation.api.servlet.ServerSpanNaming;
 import io.opentelemetry.javaagent.instrumentation.springwebmvc.SpringWebMvcServerSpanNaming;
 import java.io.IOException;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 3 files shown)


instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/org/springframework/web/servlet/OpenTelemetryHandlerMappingFilter.java
```diff
@@ -90,31 +116,16 @@ private boolean findMapping(HttpServletRequest request) {
...
-      //
-      // and then the application calls UrlMappingsHandlerMapping.getHandler() to route the request
-      // but it finds servlet request attribute "org.grails.url.match.info" already populated (by
-      // above) and so it short cuts the matching process and uses the wrong route that the agent
-      // populated caused to be populated into the request attribute above
+      // Originally we ran findMapping at the very beginning of the request. This turned out to have
+      // application-crashing side-effects with grails. That is why we don't add all HandlerMapping
+      // classes here. Although now that we run findMapping after the request, and only when server
+      // span name has not been updated by a controller, the probability of bad side-effects is much
+      // reduced.
```
trask, line 123: just confirming my understanding (missed this advantage to moving it to the end 👍)
```suggestion
      // reduced even if we did add all HandlerMapping classes here
```

instrumentation/spring/spring-webmvc-3.1/javaagent/src/main/java/org/springframework/web/servlet/OpenTelemetryHandlerMappingFilter.java
```diff
@@ -28,18 +31,41 @@
 import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping;
 
 public class OpenTelemetryHandlerMappingFilter implements Filter, Ordered {
+  // copied from ServletRequestPathUtils
+  private static final String PATH_ATTRIBUTE =
+      "org.springframework.web.util.ServletRequestPathUtils.PATH";
```
trask, line 36: I happened to click into an earlier version of `ServletRequestPathUtils` and noticed they used lowercase `.path` there, but uppercase `.PATH` in later versions.

maybe we can get the constant via reflection instead?

trask (commented):
do you know why it broke with spring boot 2.6.0? i looked and spring-webmvc was only bumped from 5.3.12 to 5.3.13. mostly wondering if that was the breaking point (5.3.12 to 5.3.13) or if we have a testing problem that made us miss an earlier breakage

laurit:
> do you know why it broke with spring boot 2.6.0? i looked and spring-webmvc was only bumped from 5.3.12 to 5.3.13. mostly wondering if that was the breaking point (5.3.12 to 5.3.13) or if we have a testing problem that made us miss an earlier breakage

I believe that it broke because previously `HandlerMapping.usesPathPatterns()` returned false, but now it returns true. Something must have changed in the default configuration of spring boot. Most likely this can be made to fail with earlier versions(`usesPathPatterns` was added in 5.3).

trask (approved):
thx 👍

trask:
@open-telemetry/java-instrumentation-approvers the spring-webmvc span naming is a hairy bit of instrumentation, would be great to get one more review on this PR before merging

PR #4691, opened 2021-11-21
Use set prefix on builders

### LINKED ISSUE #4564: Builder setter naming
I noticed that our builders (and other places) don't use `set` prefix on setters. I like this, but it is different from the SDK convention (e.g. [SpanBuilder](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/SpanBuilder.java)). Should we sync to the SDK convention, or forge our own way?

(while I prefer the convention we are using, I personally have slightly higher preference for not using a different convention in this repo from the SDK)

trask:
Closes #4564

### DIFF
--- a/instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
+++ b/instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
@@ -53,7 +53,7 @@ public final class ApacheHttpClientTracingBuilder {
    * @param capturedHttpHeaders An instance of {@link CapturedHttpHeaders} containing the configured
    *     HTTP request and response names.
    */
-  public ApacheHttpClientTracingBuilder captureHttpHeaders(
+  public ApacheHttpClientTracingBuilder setCaptureHttpHeaders(
       CapturedHttpHeaders capturedHttpHeaders) {
     this.capturedHttpHeaders = capturedHttpHeaders;
     return this;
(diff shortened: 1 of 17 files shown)


iNikem:
It is ugly :( Does Java API have builders and what convention they use?

trask:
> It is ugly :(

ya, but I think this ship has sailed and we should be consistent with OTel SDK

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
```diff
@@ -53,7 +53,7 @@ public ApacheHttpClientTracingBuilder addAttributeExtractor(
    * @param capturedHttpHeaders An instance of {@link CapturedHttpHeaders} containing the configured
    *     HTTP request and response names.
    */
-  public ApacheHttpClientTracingBuilder captureHttpHeaders(
+  public ApacheHttpClientTracingBuilder setCaptureHttpHeaders(
```
anuraaga, line 56: Think this / and the type needs to be renamed to `setCapturedHttpHeaders`

anuraaga (commented):
I'm neutral on the need to add `set` here - one thing that comes to mind for this one specifically is it will be awkward to add a version that captures all headers.

`captureHttpHeaders()` captures all and implemented as `captureHttpHeaders(CapturedHttpHeaders.all())` or something
`captureHttpHeaders(CapturedHttpHeaders)` captures specified

is pretty clean. `setCapturedHttpHeaders()` with no parameters is basically a no-go I think so user always needs to pass in the `all` parameter

This is likely not actually a use case here and set seems to make sense to :-) But an example of how we probably do need to be flexible on this topic in general.

instrumentation/apache-httpclient/apache-httpclient-4.3/library/src/main/java/io/opentelemetry/instrumentation/apachehttpclient/v4_3/ApacheHttpClientTracingBuilder.java
trask: makes sense, I will apply everywhere

trask:
👍 will hold on this PR until @mateuszrzeszutek is back and weighs in. I'm happy to make exceptions to "convention" as long as they're intentional.

mateuszrzeszutek:
My intention was to always use a verb in builder methods: `addThing()`, `setThing()`, `removeThing()`, `enableThing()`, ...
I think that the SDK repo (and instrumentation repo too) has adhered to the same convention, although maybe it wasn't explicitly stated anywhere.
With that in mind, `captureHttpHeaders()` is a valid builder method name - even though it doesn't contain `set`/`add` it still has a verb.

trask:
> `captureHttpHeaders()` is a valid builder method name - even though it doesn't contain `set`/`add` it still has a verb.

makes sense 👍

PR #4703, opened 2021-11-24
Capture servlet request parameters

laurit:
This pr implements capturing HTTP request parameters as span attributes when request is handled by servlet api. Similarly to capturing HTTP headers only parameters that are explicitly configured will be captured.

### DIFF
--- a/instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/JettyServletHandlerTest.groovy
+++ b/instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/JettyServletHandlerTest.groovy
@@ -16,6 +16,9 @@ class JettyServletHandlerTest extends AbstractServlet5Test<Object, Object> {
 
   @Override
   String expectedServerSpanName(ServerEndpoint endpoint) {
+    if (ServerEndpoint.CAPTURE_PARAMETERS == endpoint) {
+      return "HTTP POST"
+    }
     "HTTP GET"
   }
 
--- a/docs/config/common.md
+++ b/docs/config/common.md
@@ -49,3 +49,16 @@ These configuration options are supported by all HTTP client and server instrume
 
 > **Note**: The property/environment variable names listed in the table are still experimental,
 > and thus are subject to change.
+
+## Capturing servlet request parameters
+
+You can configure the agent to capture predefined HTTP request parameter as span attributes for
+requests that are handled by Servlet API.
+Use the following property to define which servlet request parameters you want to capture:
+
+| System property                                                        | Environment variable                                                   | Description |
+| ---------------------------------------------------------------------- | ---------------------------------------------------------------------- | ----------- |
+| `otel.instrumentation.servlet.experimental.capture-request-parameters` | `OTEL_INSTRUMENTATION_SERVLET_EXPERIMENTAL_CAPTURE_REQUEST_PARAMETERS` | A comma-separated list of request parameter names.
+
+> **Note**: The property/environment variable names listed in the table are still experimental,
+> and thus are subject to change.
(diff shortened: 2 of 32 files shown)


instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/JettyServletHandlerTest.groovy
```diff
@@ -16,6 +16,9 @@ class JettyServletHandlerTest extends AbstractServlet5Test<Object, Object> {
 
   @Override
   String expectedServerSpanName(ServerEndpoint endpoint) {
+    if (ServerEndpoint.CAPTURE_PARAMETERS == endpoint) {
+      return "HTTP POST"
```
iNikem, line 20: Why does this request have to be POST?

iNikem (approved):

mateuszrzeszutek (approved):

instrumentation/servlet/servlet-5.0/javaagent/src/test/groovy/JettyServletHandlerTest.groovy
laurit: It doesn't. I chose POST because GET parameters can be extracted from url and could also be captured on other http server frameworks. POST parameters, in contrast, are easy to capture with servlet api, but hard on all other http server frameworks. This is also the reason why it is only implemented for servlet.

trask: How important is it to capture parameters from `application/x-www-form-urlencoded` bodies vs limiting this feature to query parameters?

I think the `application/x-www-form-urlencoded` body support makes this feature harder to explain to users and (as you mentioned) harder to implement across all http server frameworks. I think it may also make this feature harder to upstream into the spec.

laurit: I believe that from end users perspective it makes sense to capture a request parameter regardless of whether it was passed in url, post or in multipart request, because the same parameter could be passed by different means even in the same application. I suspect that it will be hard to explain why we capture a parameter only when it is passed in the url if we choose to ignore post parameters. Whilst it might be hard to get access to post or multipart parameters in most http server frameworks it would require addition effort to exclude these using servlet api.
I agree that specifying this will be hard because with some languages and frameworks it might be too hard to capture post parameters, while in other it will be easy and excluding them just to implement the feature the same way across languages and frameworks will feel wrong. If this is going to be specified it will need to allow flexibility to allow capturing extra info when it is available and not force capturing post parameters where it is hard.

trask: What do you think of renaming from "request parameters" to "query parameters"? (and still applying to both query strings and `application/x-www-form-urlencoded` bodies)

For some reason, when I see "request parameters", I'm not really sure what that means.

Though, I just went looking and I'm having a hard time finding a canonical definition for either "request parameter" or "query parameter" 😭 🤷‍♂️.

(I can live with "request parameters" if that's what others prefer, and of course the spec will have the last word on the naming..)

laurit: Spring mvc uses `@RequestParam`. JAX-RS has `@QueryParam` (for parameters in query string) and `@FormParam` (for post).

trask: thx, let's merge as is 👍

PR #4709, opened 2021-11-25
Update release process

trask:
Will update patch release process later since there's an open question of whether we should automate cherry-picking or not for patch releases since it requires requesting a CLA exception:

> Bot accounts are not recommended, as they require special work to pass CLA checks. If pushing automatically-generated, non-copyrightable code using a bot account is required, an explanation should be sent to the Governance Committee for review and forwarding to the EasyCLA team to exempt the bot's commits from the CLA requirement.

https://github.com/open-telemetry/community/blob/main/docs/using-github-extensions.md#writing-your-github-actions-pipelines

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -15,6 +15,10 @@ Before making the release:
 
 * Merge a PR to `main` updating the `CHANGELOG.md`
 * Create a release branch, e.g. `v1.9.x`
+  ```
+  git checkout -b v1.9.x upstream/main
+  git push upstream v1.9.x
+  ```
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -13,23 +13,24 @@ as the last step, which publishes a snapshot build to
 
 Before making the release:
 
-* Update `CHANGELOG.md`
-* Update the version (remove `-SNAPSHOT`) in these files:
+* Merge a PR to `main` updating the `CHANGELOG.md`
+* Create a release branch, e.g. `v1.9.x`
```
anuraaga, line 17: Think needs documentation

anuraaga (approved):

PR #4722, opened 2021-11-26
Remove virtual field interfaces from reflection results

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4370
I initially planned to use asm `SerialVersionUIDAdder` to add `serialVersionUID` field to serializable classes that don't declare it. This is needed to make sure that instrumented version of class has the same `serialVersionUID` as uninstrumented version. This approach failed with one of spring-batch tests because inside spring there is code that copies instances of an internal spring class via reflection and fails because it can't write to `serialVersionUID` which is a static final field. We can't use the same reflection filtering code as we use for fields & methods added by virtual field implementetion because that filtering completely removes these members from reflection results. For `serialVersionUID` we'd need it to be visible from `getDeclaredField`, so that `ObjectStreamClass` can read it, and hidden from `getDeclaredFields`, so that code that isn't explicitly looking for it wouldn't stumble upon it. Instead of attempting to remove `serialVersionUID` from `getDeclaredFields` I chose to remove interfaces added by virtual field implementation from the result of `getInterfaces`. This makes the default `serialVersionUID` computation algorithm return the correct result on instrumented classes and there is no need to precompute `serialVersionUID`  for them.

### DIFF
--- a/instrumentation/internal/internal-reflection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/ReflectionHelper.java
+++ b/instrumentation/internal/internal-reflection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/ReflectionHelper.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.javaagent.instrumentation.internal.reflection;
 
+import io.opentelemetry.javaagent.bootstrap.VirtualFieldAccessorMarker;
 import io.opentelemetry.javaagent.bootstrap.VirtualFieldInstalledMarker;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
@@ -18,6 +19,7 @@ public final class ReflectionHelper {
   public static Field[] filterFields(Class<?> containingClass, Field[] fields) {
     if (fields.length == 0
         || !VirtualFieldInstalledMarker.class.isAssignableFrom(containingClass)) {
+      // nothing to filter when class does not have any added virtual fields
       return fields;
     }
     List<Field> result = new ArrayList<>(fields.length);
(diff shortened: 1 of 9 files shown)


iNikem (approved):
I hope you know what you are doing :D Instrumenting `java.lang.Class` seems scary

trask:
@laurit do you think we could get rid of these interfaces and use method handles to access the generated methods?

laurit:
@trask Firstly we would need to figure out a new way to detect whether virtual fields have been added to a class which is currently handled by testing for `VirtualFieldInstalledMarker` interface or virtual field accessor interface. This could be implemented using a map with class name+class loader as key, but as the whole point of virtual fields is to avoid using maps we'd probably need to come up with something else. If we somehow end up using method handles then one thing to consider is that we can't always initialize them in static initializer because presence of `<clinit>` method also changes the default serial version uid.

instrumentation/internal/internal-reflection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/ReflectionHelper.java
```diff
@@ -36,6 +38,7 @@ private ReflectionHelper() {}
       return methods;
     } else if (containingClass.isInterface()
```
laurit, line 39: If filtering of the interfaces works then this branch shouldn't evaluate to true. Similarly I suspect that  https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/internal/internal-proxy/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/proxy/ProxyHelper.java might not be necessary any more.

mateuszrzeszutek (approved):

instrumentation/internal/internal-reflection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/ReflectionHelper.java
trask: nice, I'll open an issue to try this 👍

trask (approved):
> presence of `<clinit>` method also changes the default serial version uid.

😭

Issue #4645, opened 2021-11-15
Serialisation exception for ScheduledEvent


acm19:
**Describe the bug**
When instrumenting a Lambda triggered by `ScheduledEvent` a parsing exception is produced:

```
at io.opentelemetry.instrumentation.awslambda.v1_0.TracingRequestHandler.handleRequest(TracingRequestHandler.java:89)
Caused by: java.lang.IllegalArgumentException: Unrecognized field "version" (class com.amazonaws.services.lambda.runtime.events.ScheduledEvent), not marked as ignorable (8 known properties: "detail", "region", "resources", "detailType", "account", "source", "time", "id"])
```

**Steps to reproduce**
Create a lambda triggered by a `ScheduledEvent` and instrument it using OpenTelemetry following this tutorial: https://aws-otel.github.io/docs/getting-started/lambda/lambda-java.

**What did you expect to see?**
I expect no exception.

**What did you see instead?**
The lambda failed with an exception.

**What version are you using?**
v1.7.0

**Environment**
Java 11 Corretto running in AWS Lambda.

**Additional context**
I think a good way to fix the problem would be to ignore unknown properties by this object mapper: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6063a16f54453c9c0cff8b754e492b142d091c45/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java#L22. I'm happy to create a PR if that would speed up having it fixed.

laurit:
@acm19 a pr would be appreciated.

acm19:
@laurit https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4724

PR #4724, opened 2021-11-26
Fix serialisation exception on default lambda events

acm19:
Adds `FAIL_ON_UNKNOWN_PROPERTIES` to the paramter parser on the lambda
as events coming from AWS have more fields than those represented in the
libraries provided by AWS. Adds `JodaModule` to the same parser to
support `ScheduledEvent` which for now uses Joda Time.

Resolves: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4645

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/build.gradle.kts
+++ b/instrumentation/aws-lambda-1.0/library/build.gradle.kts
@@ -22,8 +22,10 @@ dependencies {
 
   implementation("io.opentelemetry:opentelemetry-extension-aws")
 
-  // 1.2.0 allows to get the function ARN
-  testLibrary("com.amazonaws:aws-lambda-java-core:1.2.0")
+  // allows to get the function ARN
+  testLibrary("com.amazonaws:aws-lambda-java-core:1.2.1")
+  // allows to get the default events
+  testLibrary("com.amazonaws:aws-lambda-java-events:3.10.0")
 
   testImplementation("com.fasterxml.jackson.core:jackson-databind")
   testImplementation("commons-io:commons-io:2.2")
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
@@ -6,6 +6,7 @@
 package io.opentelemetry.instrumentation.awslambda.v1_0;
 
 import com.amazonaws.services.lambda.runtime.Context;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
@@ -19,7 +20,10 @@ import java.util.function.BiFunction;
  */
 abstract class TracingRequestWrapperBase<I, O> extends TracingRequestHandler<I, O> {
 
-  protected static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  protected static final ObjectMapper OBJECT_MAPPER =
+      new ObjectMapper()
+          .registerModule(new CustomJodaModule())
+          .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
   private final WrappedLambda wrappedLambda;
   private final Method targetMethod;
   private final BiFunction<I, Class, Object> parameterMapper;
--- a/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/CustomJodaModule.java
+++ b/instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/CustomJodaModule.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import io.opentelemetry.testing.internal.jackson.core.JsonTokenId;
+import java.io.IOException;
+import javax.annotation.Nullable;
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+
+/**
+ * Jackson module to be able to properly parse standard events as provided by <a
+ * href="https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-events">AWS</a>.
+ *
+ * <p>It enables parsing of {@link DateTime} which is used by some standard events. A custom module
+ * was used as opposed to Jackson standard module for parsing Joda to avoid adding more libraries to
+ * the Java Agent.
+ *
+ * <p>Supporting custom POJOs using Joda is out of the scope of this class.
+ */
+class CustomJodaModule extends SimpleModule {
+  public CustomJodaModule() {
+    super();
+    addDeserializer(DateTime.class, new DateTimeDeserialiser());
+  }
+
+  @Override
+  public String getModuleName() {
+    return getClass().getSimpleName();
+  }
+
+  @Override
+  public int hashCode() {
+    return getClass().hashCode();
+  }
+
+  @Override
+  public boolean equals(@Nullable Object o) {
+    return this == o;
+  }
+
+  private static class DateTimeDeserialiser extends JsonDeserializer<DateTime> {
+    private final DateTimeFormatter dateFormatter =
+        DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ssZ");
+
+    @Override
+    public DateTime deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+      if (JsonTokenId.ID_STRING != p.getCurrentTokenId()) {
+        throw new IllegalArgumentException("Only stream input is accepted");
+      }
+
+      String value = p.getText().trim();
+      return value.isEmpty() ? null : dateFormatter.parseDateTime(value);
+    }
+  }
+}
--- a/instrumentation/aws-lambda-1.0/library/src/test/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperStandardEventsTest.java
+++ b/instrumentation/aws-lambda-1.0/library/src/test/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperStandardEventsTest.java
@@ -1,0 +1,262 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.awslambda.v1_0;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.mock;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.KinesisEvent;
+import com.amazonaws.services.lambda.runtime.events.S3Event;
+import com.amazonaws.services.lambda.runtime.events.SNSEvent;
+import com.amazonaws.services.lambda.runtime.events.SQSEvent;
+import com.amazonaws.services.lambda.runtime.events.ScheduledEvent;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class TracingRequestWrapperStandardEventsTest {
+  private static final String SUCCESS = "success";
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  private static final Map<Class<?>, String> EVENTS_JSON = buildEventExamples();
+
+  private final OpenTelemetrySdk sdk = OpenTelemetrySdk.builder().build();
+  private final Context context = mock(Context.class);
+  private TracingRequestWrapper wrapper;
+
+  private static Map<Class<?>, String> buildEventExamples() {
+    Map<Class<?>, String> events = new HashMap<>();
+    events.put(
+        ScheduledEventRequestHandler.class,
+        "{\n"
+            + "    \"version\": \"0\",\n"
+            + "    \"id\": \"53dc4d37-cffa-4f76-80c9-8b7d4a4d2eaa\",\n"
+            + "    \"detail-type\": \"Scheduled Event\",\n"
+            + "    \"source\": \"aws.events\",\n"
+            + "    \"account\": \"123456789012\",\n"
+            + "    \"time\": \"2015-10-08T16:53:06Z\",\n"
+            + "    \"region\": \"us-east-1\",\n"
+            + "    \"resources\": [\n"
+            + "        \"arn:aws:events:us-east-1:123456789012:rule/my-scheduled-rule\"\n"
+            + "    ],\n"
+            + "    \"detail\": {}\n"
+            + "}");
+    events.put(
+        KinesisEventRequestHandler.class,
+        "{\n"
+            + "    \"Records\": [\n"
+            + "        {\n"
+            + "            \"kinesis\": {\n"
+            + "                \"kinesisSchemaVersion\": \"1.0\",\n"
+            + "                \"partitionKey\": \"1\",\n"
+            + "                \"sequenceNumber\": \"49590338271490256608559692538361571095921575989136588898\",\n"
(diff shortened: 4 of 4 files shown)


anuraaga (commented):
@kubawach Can you take a look at this PR? Thanks

instrumentation/aws-lambda-1.0/library/build.gradle.kts
```diff
@@ -9,21 +9,24 @@ dependencies {
   compileOnly("com.google.auto.value:auto-value-annotations")
   annotationProcessor("com.google.auto.value:auto-value")
 
-  library("com.amazonaws:aws-lambda-java-core:1.0.0")
+  library("com.amazonaws:aws-lambda-java-core:1.2.1")
```
laurit, line 12: Usually the name of the module (in this case `aws-lambda-1.0`) includes the earliest supported library version. Are you sure that it still works with 1.0? If it does consider not changing this version.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
```diff
@@ -9,21 +9,24 @@ dependencies {
...
   // First version to includes support for SQSEvent, currently the most popular message queue used
   // with lambda.
   // NB: 2.2.0 includes a class called SQSEvent but isn't usable due to it returning private classes
   // in public API.
   library("com.amazonaws:aws-lambda-java-events:2.2.1")
 
   compileOnly("com.fasterxml.jackson.core:jackson-databind")
+  implementation("com.fasterxml.jackson.datatype:jackson-datatype-joda")
```
laurit, line 20: You have to be careful with adding implementation dependencies because they are usually shaded inside agent jar. Could you verify that it is not shaded into agent jar, if you need it shaded inside the agent then please explain why.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
```diff
@@ -19,7 +21,10 @@
  */
 abstract class TracingRequestWrapperBase<I, O> extends TracingRequestHandler<I, O> {
 
-  protected static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+  protected static final ObjectMapper OBJECT_MAPPER =
+      new ObjectMapper()
+          .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+          .registerModule(new JodaModule());
```
laurit, line 27: Adding `JodaModule` this way might not be a good idea. The problem here is that if the instrumented application does not contain `JodaModule` then the instrumentation can not be applied. Are you sure that all `aws-lambda-java-core` users also use `jackson-datatype-joda`?

instrumentation/aws-lambda-1.0/library/build.gradle.kts
acm19: This artefact is used to deploy the AWS lambda layer. By using the latest version you have access to more feature of the lambda context: https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-core/RELEASE.CHANGELOG.md. I see nothing to gain from leaving the old version in, but it's out of scope for this PR and I don't want it to block it, so let me know if you want me to roll it back.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
acm19: `aws-lambda-java-events` uses Joda: https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/pom.xml#L49. It uses it in `ScheduledEvent` for example. If you don't use `JodaModule` your lambda layer will fail to parse it properly as described in the bug report. Add that module shouldn't do anything apart from properly parsing classes using Joda. Joda is basically dead so hopefully AWS will release an updated version of the events with Java Time API but this project still needs to support older versions of the events library, so I'd expect it to remain.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
acm19: I explained it here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4724/files#r758320858. Please let me know that's clear and you want a comment in the code as well.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
mateuszrzeszutek: > By using the latest version you have access to more feature of the lambda context

The thing is, we don't necessarily want to use the newer features - the javaagent instrumentation of AWS lambdas captures references to used library code at compile time and verifies that they're present in the runtime (read [muzzle docs](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/muzzle.md) for more info on that); if we instrument a lambda that is compiled against 1.0 and doesn't have the methods we use in the instrumentation, the whole instrumentation will be turned off to prevent linking errors.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
laurit: I believe you are missing a few details here. `aws-lambda-1.0/library` is for manual instrumentation and `aws-lambda-1.0/javaagent` that is used for autoinstrumentaiton. The `javaagent` module depends on the `library` module. All runtime dependencies of `javaagent` module will be embedded inside javaagent jar under `inst` directory, check the files under `javaagent/build/libs`. During runtime classes from javaagent instrumentation are injected into the class loader that contains the library, but we only injects the so called helper classes (see https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/writing-instrumentation-module.md). Everything else is taken from the application class loader, so the content of `jackson-datatype-joda` that is embedded into agent jar isn't really used.

acm19: What details exactly am I missing? The auto-instrumentation of lambda functions doesn't work well. It doesn't parse the default Amazon events, then the pull request. You need the `jackson-datatype-joda` in the agent so that it parses the events to the correct type defined in the instrumented lambda function. Please notice `joda-time` is transitively included by `aws-lambda-java-events` but `jackson-datatype-joda` isn't. Do you have a better idea on how to address the issue?

instrumentation/aws-lambda-1.0/library/build.gradle.kts
acm19: Reverted.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
anuraaga: I think in the original PR that added the JSON deserialization I had commented on how it's a bit tricky that we aren't using AWS's same mechanism for deserialization, which presumably isn't using jackson-datatype-joda.

More worringly is that the intent was that the instrumentation wouldn't cause exceptions in the user's function if the serialization did have problems but that seems to be happening here. I think we need to fix that. Otherwise we may want to give up on handling JSON - Joda types are one corner case but it's hard to be confident we don't continue to run into more and more corner cases for this. @kubawach Do you happen to be around to provide thoughts on this?

acm19: > we aren't using AWS's same mechanism for deserialization, which presumably isn't using jackson-datatype-joda.

AWS might not even use Jackson I presume.
>More worringly is that the intent was that the instrumentation wouldn't cause exceptions in the user's function if the serialization did have problems but that seems to be happening here. I think we need to fix that.

I agree but I think this issue needs fixing nevertheless as it's one of AWS standard events provided by their library. Not failing on missing properties is a step in this direction BTW.
> Joda types are one corner case but it's hard to be confident we don't continue to run into more and more corner cases for this.

I think the instrumentation should parse all events in `aws-lambda-java-events`, so I wouldn't call it a corner case. I added tests for 3 of them but we can add test for **all** of them to avoid surprises.

kubawach: Hey, yes, @anuraaga, you remember well latest discussion on that and general skepticism regarding wrapper's ability to deserialize a required POJO from JSON. 
My two cents here are:

1. In general the wrapper is a customer-centric "nice to have" component. Users can always use appropriate handler directly if the wrapper does not work in their particular case. POJO-handling wrapper is even more so. It has a built-in fallback mechanism - if user declares own lambda handler as <Object, Object> the wrapper will not try to map coming data. In such case it will work exactly as it worked before adding POJO mapping extension. 

2. Interestingly AWS seems to publish the mapping code on GitHub - see https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/factories/JacksonFactory.java It's even in the M2 repo (albeit not up to date with changes on GitHub - last change 7 days ago, last release a year ago). So there could be a way to either implement the mapping same way as AWS does or even re-use their code by depending on the artifact.

To sum up:
1. Adding `FAIL_ON_UNKNOWN_PROPERTIES` will certainly not hurt (AWS does the same).
2. Date/time conversion is done with reflection, we could use the same to avoid adding Joda etc (see https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/util/SerializeUtil.java)
3. Overall strategy towards this can ultimately go one of two routes (as already signed by @anuraaga):
a) try to support everything same way as AWS (which may be tricky due to version changes) for example by using AWS artifact / re-using their mapping code
b) clearly state in the docs that POJO-mapping wrapper is a `convenience solution` that may not support all cases (various types etc) in which case users should implement the mapping on the handler side (and declare it as <Object, Object>)

I *personally* lean towards option `b)` as it gives more control to users and relives the community from burden of supporting external (AWS) deserialization implementation.

acm19: @kubawach I think having to use `<Object, Object>` is a terrible user experience. Supporting a random POJO might be a nice to have but supporting default AWS types should not. It forces you to deserialise objects in your app consuming CPU and time unnecessarily, which might be a big drawback in many cases. Ideally you should be able to auto instrument your lambda without changes in the code.

About the code you shared, that's not their actual logic. It's a serialisation logic used by `aws-lambda-java-runtime-client` which is used for containerised your lambda for development. It's not designed for standalone use. You could copy the logic but I'd expect the crude reflection to be less performant than using the Jackson module, I haven't run any performance test to validate that, but it shouldn't be hard. Also you get rid of any side effects of that reflection. I feel like a good solid implementation is worth 74 KB, but it's not my call.

About the 2 routes you mentions my personal view as a user is they are not mutually exclusive. You can provide a good support for mapping POJO, helping people instrument their code with minimal changes and impact. At the same time you can document that this is a best effort solution and how to work around the issues you might experience. Please let me know if I can help.

kubawach: @acm19 ability to use the wrapper without code changes was the primary reason to introduce the changes we're discussing right now ;) However I side with @anuraaga that we may expect more corner cases in the future if we decide we (as a community) want to allow wrapper usage in each and every case. In this respect - this does not sound viable. But it's a "strategic" side of things. In this particular case I'd:
- set the option `FAIL_ON_UNKNOWN_PROPERTIES`
- implement types-via-reflection handling as documented in the AWS code
..but I'm not an approver so it's up to @open-telemetry/java-instrumentation-approvers to decide.

Wrt reflection and performance - I meant dynamic invocation in general, not exactly reflection. Static MethodHandles are common in the instrumentation code and known to be quite fast. Good way to avoid another dependency especially in the quite tirvial case.

acm19: > ability to use the wrapper without code changes was the primary reason to introduce the changes we're discussing right now

I think the discussion is not on the purpose but the usefulness. My point is, from the user perspective this is an essential feature. Not only for convenience but also because having to parse objects is not only an unnecessary hassle but impractical for the reasons I explained before. Again, not my call, but I think this feature should at least:

1. Address **all** events POJOs in `aws-lambda-java-events`.
2. Support edge cases as they arrive whenever sensible.

> implement types-via-reflection handling as documented in the AWS code
..but I'm not an approver so it's up to @open-telemetry/java-instrumentation-approvers to decide.

If you prefer this way, I'm happy to change the PR accordingly. I think supporting "default" events is the essential bit missing that should be added in this PR anyway.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
anuraaga: I'm afraid of dependency conflict if we depend on a separate jackson library here - the function will have a version of jackson through aws-lambda-java-events, but this one may not be aligned and I think Jackson is always supposed to be aligned to risk problems.

Instead, we could add a compileOnly dependency on joda-time and vendor in just the parts of datatype-joda that we need, would that work?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
anuraaga: > Address all events POJOs in aws-lambda-java-events.
> Support edge cases as they arrive whenever sensible.

This seems to make sense to me, thanks for discussing. The more event types we can cover in unit tests the safer it'll be.

acm19: I'll be off for a few days, I'll update the PR with additional tests when I'm back and I'll address the rest of the comments. Thanks.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
acm19: > I'm afraid of dependency conflict if we depend on a separate jackson library here - the function will have a version of jackson through aws-lambda-java-events, but this one may not be aligned and I think Jackson is always supposed to be aligned to risk problems.

The version of Jackson is defined by the library's own dependencyManagement: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/dependencyManagement/build.gradle.kts#L33. That makes sure it's properly aligned with the Jackson Joda module.

> Instead, we could add a compileOnly dependency on joda-time and vendor in just the parts of datatype-joda that we need, would that work?

I might not understand what you have in mind but I see 2 options, using the module or the approached mentioned by @kubawach in the other thread. I have preference for the module, I explained the reasons in the other thread. But I'm not a decision maker in this project and I'm happy to change it in order to get the issue fixed.

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
acm19: I added SNS and S3 as well.

instrumentation/aws-lambda-1.0/library/build.gradle.kts
anuraaga: > The version of Jackson is defined by the library's own dependencyManagement

We use a `compileOnly` dependency on `jackson-databind` currently - this is to make sure the version of Jackson used is the one in the user's app / lambda framework dependency since we generally don't want instrumentation to control dependency versions in the app. So currently we aren't defining a version for Jackson here. It isn't hard for jackson-databind and jackson-joda to have inconsistent versions in the final app - simplest solution is probably to avoid jackson-joda

acm19: > this is to make sure the version of Jackson used is the one in the user's app / lambda framework dependency since we generally don't want instrumentation to control dependency versions in the app

The thing is, for the lambda layer, which is where this module is relevant, it doesn't matter if the app includes the dependency or not. Actually many lambdas don't include jackson and it still works. I think the important thing is that this dependency makes it into the lambda layer: https://github.com/open-telemetry/opentelemetry-lambda/tree/main/java/layer-wrapper.

So I could change it to `compileOnly` and `testImplementation` in this repository and make the changes where necessary so it gets included in the layer. If that works for you, please let me know where should we add that dependency so it gets added there in the way you prefer so that it's consistent with your way of working.

anuraaga: Right - if an app doesn't include jackson, then it will be using the one included in the lambda runtime itself. I'm not sure but I guess this one

https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/pom.xml#L41

If jackson-databind used by the app / lambda runtime is 2.10.0 but we include 2.12.0, there is reasonable chance of a conflict causing breakage as the version alignment is outside our control. As instrumentation, we don't want to influence the version of non-OTel libraries as much as possible.

Since we only need to marshal a couple of joda types, I suspect we can add here a small amount of code to worry less about dependency conflicts, and it would probably have a reasonable effect on reducing added binary size.

acm19: Sorry for the long silence, I've been off for holiday and was very busy when I came back. I've uploaded a version of the code without JodaModule.

anuraaga (approved):
Thanks!

acm19:
@anuraaga I've rebased it and added some Javadoc to the custom Joda module.

kubawach (approved):
Great job, thanks for perseverance here :)

Issue #4731, opened 2021-11-29
Vertx http client 3 instrumentation populating relative URL in http.url


anuraaga:
We currently populating the confusingly named relative URI field as http.url

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vertx/v3_0/client/Vertx3HttpAttributesExtractor.java#L17

We hadn't noticed it because this line is incorrect.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/test/groovy/client/VertxHttpClientTest.groovy#L34

It should be `httpClient.requestAbs(HttpMethod.valueOf(method), "$uri")`

It seems like Jetty and older versions of Armeria handled an absolute URL as a path leniently in a way that worked. When trying to update Armeria I found this test failing with HTTP 400. It did make clear that when using the methods correctly, the instrumentation populates a relative path into that attribute.

Unfortunately there doesn't seem to be a way to access hostname publicly. Should we instrument the constructor to store it in a virtual field? Or any other tips appreciated.

PR #4739, opened 2021-11-29
Fix http.url handing in vert.x 3 http client

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4731

### DIFF
--- a/instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vertx/v3_0/client/HttpClientImplInstrumentation.java
+++ b/instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vertx/v3_0/client/HttpClientImplInstrumentation.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.vertx.v3_0.client;
+
+import static net.bytebuddy.matcher.ElementMatchers.isConstructor;
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+import io.opentelemetry.instrumentation.api.field.VirtualField;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeTransformer;
+import io.vertx.core.http.HttpClientOptions;
+import io.vertx.core.http.impl.HttpClientImpl;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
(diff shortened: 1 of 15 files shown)


instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/test/groovy/client/VertxHttpClientTest.groovy
```diff
@@ -78,6 +78,11 @@ class VertxHttpClientTest extends HttpClientTest<HttpClientRequest> implements A
     false
   }
 
+  @Override
+  String nonRoutableAddress() {
```
anuraaga, line 82: Does it make sense to check testHttps instead of adding a similar knob?

instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vertx/v3_0/client/HttpClientImplInstrumentation.java
```diff
@@ -0,0 +1,41 @@
...
+    transformer.applyAdviceToMethod(
+        isConstructor(), HttpClientImplInstrumentation.class.getName() + "$AttachStateAdvice");
+  }
+
+  public static class AttachStateAdvice {
+    @Advice.OnMethodExit(suppress = Throwable.class)
+    public static void attachHttpClientOptions(
+        @Advice.This HttpClientImpl client,
+        @Advice.FieldValue("options") HttpClientOptions options) {
+      VirtualField.find(HttpClientImpl.class, HttpClientOptions.class).set(client, options);
```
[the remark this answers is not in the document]
laurit: In my initial attempt I used `@FieldValue` and tried calling `client.getOptions()`, but this is a package private method, so to make it compile I had to move this call into a helper class that is in the same package as `HttpClientImpl`. Turns out that even that isn't enough. I think that currently muzzle tags all method calls with protected or higher modifiers expected which makes calling package private method from another class impossible.

mateuszrzeszutek (approved):

anuraaga (approved):

instrumentation/vertx-http-client/vertx-http-client-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/vertx/v3_0/client/Vertx3HttpAttributesExtractor.java
```diff
@@ -5,16 +5,42 @@
...
+    }
+    return uri;
+  }
+
+  private static boolean isAbsolute(String uri) {
+    return uri.startsWith("http://") || uri.startsWith("https://");
+  }
+
+  private static String absoluteUri(VertxRequestInfo requestInfo, String uri) {
+    String result = requestInfo.isSsl() ? "https://" : "http://";
```
anuraaga, line 35: This code is probably not called commonly since it is misuse of vertx API but we should probably still stick to StringBuilder as a general pattern for this sort of string building

PR #4745, opened 2021-11-30
Use containerIpAddress instead of hard-coded 127.0.0.1 in rabbit tests

mateuszrzeszutek:
This failed locally for me (probably because I'm using minikube-backed docker); but it'll probably be useful for testcontainers cloud too.


instrumentation/rabbitmq-2.7/javaagent/src/test/groovy/RabbitMqTest.groovy
```diff
@@ -340,7 +340,7 @@ class RabbitMqTest extends AgentInstrumentationSpecification implements WithRabb
 
       attributes {
         "${SemanticAttributes.NET_PEER_NAME.key}" { it == null || it instanceof String }
-        "${SemanticAttributes.NET_PEER_IP.key}" { "127.0.0.1" }
+        "${SemanticAttributes.NET_PEER_IP.key}" { it == null || it == rabbitMqContainer.containerIpAddress }
```
iNikem, line 343: Can it be null?

iNikem (approved):

instrumentation/rabbitmq-2.7/javaagent/src/test/groovy/RabbitMqTest.groovy
mateuszrzeszutek: Yes, apparently some of these tests expect null here. (just for the record, `{ "127.0.0.1" }` was treated as a predicate that always returns "truthy" value so it didn't check the IP address at all 🙈 )

trask (approved):

instrumentation/rabbitmq-2.7/javaagent/src/test/groovy/WithRabbitMqTrait.groovy
```diff
@@ -28,4 +28,13 @@ trait WithRabbitMqTrait {
   def stopRabbit() {
     rabbitMqContainer?.stop()
   }
+
+  static boolean isContainerIpAddress(Object ip) {
+    def containerIp = rabbitMqContainer.containerIpAddress
+    // getContainerIpAddress() can return "localhost", which obviously is not an IP address
```
iNikem: Should we report this as bug to Testcontainers?

mateuszrzeszutek: https://github.com/testcontainers/testcontainers-java/issues/4715

mateuszrzeszutek: Turns out this method is supposed to work this way and will be deprecated 😅 
I changed the assertion slightly so that it matches any IP address.

laurit (approved):

PR #4746, opened 2021-11-30
Move cache implementations to internal package

### LINKED ISSUE #4735: Move concurrentlinkedhashmap and weaklockfree to `internal` packages
From https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4667#pullrequestreview-817719290

> Should we put CLHM and WeakLockFree in `internal` package(s)? They shouldn't be a part of our public API.

iNikem:
Closes #4735

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/WeakLockFreeCache.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/WeakLockFreeCache.java
@@ -5,8 +5,8 @@
 
 package io.opentelemetry.instrumentation.api.cache;
 
-import java.util.function.Function;
 import io.opentelemetry.instrumentation.api.cache.internal.weaklockfree.WeakConcurrentMap;
+import java.util.function.Function;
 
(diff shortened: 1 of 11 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/WeakLockFreeCache.java
```diff
@@ -5,7 +5,7 @@
 
 package io.opentelemetry.instrumentation.api.cache;
 
-import io.opentelemetry.instrumentation.api.cache.weaklockfree.WeakConcurrentMap;
+import io.opentelemetry.context.internal.shaded.WeakConcurrentMap;
```
laurit, line 8: This import looks wrong. I guess sdk also has a copy of weaklockfree

iNikem: It is wrong. Thanks for noticing

laurit (approved):

mateuszrzeszutek (approved):

trask (approved):

anuraaga (approved):

Issue #4740, opened 2021-11-29
Try to remove ProxyHelper and ReflectionHelper conditional branch


trask:
See @laurit's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4722#pullrequestreview-817605399

PR #4749, opened 2021-11-30
Remove internal-proxy instrumentation

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4740
As far as I understand internal-proxy instrumentation shouldn't be needed any more after patching `Class.getInterfaces` because now application code can't easily stumble upon our virtual field interfaces.


iNikem (approved):

mateuszrzeszutek (approved):

trask (approved):
nice

PR #4751, opened 2021-11-30
Remove usage of rmic

iNikem:
All tests pass, anybody knows any reasons we actually need this?


trask (commented):
> All tests pass, anybody knows any reasons we actually need this?

Looks like it's been there since original rmi instrumentation was contributed: https://github.com/DataDog/dd-trace-java/pull/1128

> This PR in its current form adds instrumentation of client and server endpoints used in Java RMI communication.
Both new RMI dynamically generated communication stubs and `rmic` generated Stubs are supported.

If we delete it, we should also delete `ServerLegacy.java` and associated test.

It's not clear to me if any of the instrumentation is specific for this "legacy" use case, or if this is just an extra test to verify the normal instrumentation also works in the "legacy" case.

iNikem:
My original intent is to make this instrumentation a better Gradle citizen. Currently `rmic` task does not declare any output and thus plays badly with task avoidance.  I am happy to remove our support for deprecates static stubs. But if this is considered too bold of a move, I can just make some changes in gradle script.

trask:
It looks to me like there's no instrumentation that is specific to the legacy rmi case, so I think test coverage should be good enough without the legacy rmi test.

iNikem:
Removed `ServerLegacy`

mateuszrzeszutek (approved):

trask (approved):

PR #4756, opened 2021-12-01
Update patch release docs

trask:
My thought about removing the auto- cherry picking is that the manual cherry pick PRs will give us review/approval/traceability on the patch release process, and then no need for checking before hitting the patch release button.


anuraaga:
I guess I generally don't like reviewing the same code twice. If it were clear from a PR title whether it is a clean cherry pick or not and just 100% rubber-stamp than it could be better but otherwise still tend towards not review for clean cherry picks since it's like reviewing the exact same code twice

anuraaga (approved):

PR #4764, opened 2021-12-01
Develop/rmi instrumentation

Samudraneel24:
- Removing the top level span check as discussed.
- Changing `extendsClass(named("java.rmi.server.RemoteServer"))` to `implementsInterface(named("java.rmi.Remote"))` because any remote object must directly or indirectly implement the `Remote` interface. So this matcher covers a larger set of possible remote object implementations.


mateuszrzeszutek (approved):
LGTM 👍

iNikem (approved):

PR #4765, opened 2021-12-01
Disable jar task in javaagent module

iNikem:
Are there any reasons why we cannot do this?

### DIFF
--- a/conventions/src/main/kotlin/otel.publish-conventions.gradle.kts
+++ b/conventions/src/main/kotlin/otel.publish-conventions.gradle.kts
@@ -9,8 +9,10 @@ publishing {
       plugins.withId("java-platform") {
         from(components["javaPlatform"])
       }
-      plugins.withId("java-library") {
-        from(components["java"])
+      if(project.name != "javaagent"){
+        plugins.withId("java-library") {
+          from(components["java"])
+        }
       }
 
       versionMapping {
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek:
I think there were come complications with the artifacts produced by the `shadowJar` plugin, and leaving the `jar` task seemingly fixed them - if I remember correctly the no-classifier jar disappeared when you turned off `jar` and tried to publish to maven.

iNikem:
> I think there were come complications with the artifacts produced by the `shadowJar` plugin, and leaving the `jar` task seemingly fixed them - if I remember correctly the no-classifier jar disappeared when you turned off `jar` and tried to publish to maven.

A, what this build failure means... I will look into that, thanks! :)

iNikem:
Well... I think we have some sort of a pickle here...

First, I think it is wrong that we apply `java-library` plugin to `javaagent` module. This also leads to us trying to publish "common" jar from that module. I don't think we should do it. We should publish only two shadow jars from it.

Second, we actually never asked gradle specifically to publish our shadow jar. I think it worked accidentally, because we overwrite the result of `jar` file with a shadow jar without classifier.

trask:
It's nice to see the `examples` build catching jar publishing problems now prior to releasing 😅

trask (approved):

mateuszrzeszutek (approved):

Issue #4483, opened 2021-10-25
Consolidate "benchmark" modules


iNikem:
We currently have 5 modules with the word "benchmark" in their names. We probably can consolidate them and remove some duplication. 

Do we still need `benchmark-e2e` when we have newer `benchmark-overhead`? 

What is the difference between `benchmark` and `benchmark-overhead-jmh`? Can they be one module?

Do we need a whole separate `benchmark-jfr-analyzer` module for just one class?

PR #4766, opened 2021-12-01
Remove unhelpful benchmarks

iNikem:
Part of #4483

I reviewed benchmarks in `benchmark` module and the majority of them did not make sense to me. I moved a couple of useful ones to a module that contains classes they benchmark and removed the whole remaining module.


benchmark/src/jmh/java/io/opentelemetry/benchmark/IgnoredTypesMatcherBenchmark.java
```diff
@@ -1,52 +0,0 @@
...
-import org.openjdk.jmh.annotations.Mode;
-import org.openjdk.jmh.annotations.OutputTimeUnit;
-import org.openjdk.jmh.annotations.Warmup;
-
-@Fork(1)
-@Warmup(iterations = 3, time = 1)
-@Measurement(iterations = 10, time = 1)
-@OutputTimeUnit(TimeUnit.MICROSECONDS)
-@BenchmarkMode(Mode.AverageTime)
-public class IgnoredTypesMatcherBenchmark {
```
trask, line 28: this one seems worth keeping (cc @mateuszrzeszutek)

trask (approved):

benchmark/src/jmh/java/io/opentelemetry/benchmark/IgnoredTypesMatcherBenchmark.java
iNikem: Then I will ask some clarification about what it does :) What is the difference between `springType` and `appType`?

mateuszrzeszutek: The length of the matched prefix -- `com.example.myapp.Main` is dropped after `com.`, `org.springframework.test.SomeClass` after `org.springframework.`. Maybe there's a better example now, but when I wrote this benchmark and compared against the previous implementation it had to be spring - there were additional ifs, nesting and string comparisons for spring classes.

iNikem: I restored this benchmark

mateuszrzeszutek (approved):

PR #4768, opened 2021-12-01
Ask Animalsniffer to always produce report

iNikem:
The main result of this change is that Animalsniffer tasks now always has declared output and thus properly participate in tasks up-to-date checks

### DIFF
--- a/conventions/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
+++ b/conventions/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
@@ -1,3 +1,5 @@
+import ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer
+
 plugins {
   `java-library`
 
@@ -10,4 +12,7 @@ dependencies {
 
 animalsniffer {
   sourceSets = listOf(java.sourceSets.main.get())
+}
+tasks.withType<AnimalSniffer> {
+  reports.text.required.set(true)
 }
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

conventions/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
```diff
@@ -10,4 +12,7 @@ dependencies {
 
 animalsniffer {
   sourceSets = listOf(java.sourceSets.main.get())
+}
+tasks.withType<AnimalSniffer> {
+  reports.text.required.set(true)
```
trask, line 17:
```suggestion
  // always having declared output makes this task properly participate in tasks up-to-date checks
  reports.text.required.set(true)
```

trask (approved):

conventions/src/main/kotlin/otel.animalsniffer-conventions.gradle.kts
```diff
@@ -10,4 +12,8 @@ dependencies {
 
 animalsniffer {
   sourceSets = listOf(java.sourceSets.main.get())
+}
+tasks.withType<AnimalSniffer> {
```
anuraaga, line 16: `tasks.withType<Foo>.configureEach {`

iNikem: What's the difference to the current way?

PR #4770, opened 2021-12-01
Make AgentClassLoaderAccess methods public

mateuszrzeszutek:
This class used to have a public getter that exposed the agent/extension classloader and we were using it in our distro's tests; the `loadClass()` method is pretty much equivalent to that.

### DIFF
--- a/testing-common/src/main/java/io/opentelemetry/javaagent/testing/common/AgentClassLoaderAccess.java
+++ b/testing-common/src/main/java/io/opentelemetry/javaagent/testing/common/AgentClassLoaderAccess.java
@@ -24,6 +24,7 @@ public final class AgentClassLoaderAccess {
     }
   }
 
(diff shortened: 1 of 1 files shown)


testing-common/src/main/java/io/opentelemetry/javaagent/testing/common/AgentClassLoaderAccess.java
```diff
@@ -24,7 +24,7 @@
     }
   }
 
-  static Class<?> loadClass(String name) {
+  public static Class<?> loadClass(String name) {
```
trask, line 27: something like this, just so we don't accidentally make it package-private in the future?
```suggestion
  // public for use by downstream distros
  public static Class<?> loadClass(String name) {
```

trask (approved):

Issue #4718, opened 2021-11-26
AWS SDK fails to initialize


anuraaga:
It seems #4607 has caused an issue with the AWS SDK instrumentation. I guess muzzle may not be parsing classes out of the SPI resource file which is in a different artifact.

I think our javaagent tests don't catch this because of the `implementation` dependency on the library - presumably for safety against these issues we need to change the pattern to something like `compileOnly` and `shadowBundle` so that the test classpath does not have the library included in it. Well I guess https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts#L116 is supposed to be handling it but maybe it doesn't for some reason.

```
[docker-java-stream-1118179615] INFO logback - STDOUT: Caused by: java.lang.ClassNotFoundException: io.opentelemetry.instrumentation.awssdk.v2_2.autoconfigure.TracingExecutionInterceptor
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at java.base/java.lang.Class.forName0(Native Method)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at java.base/java.lang.Class.forName(Unknown Source)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at software.amazon.awssdk.core.internal.util.ClassLoaderHelper.loadClass(ClassLoaderHelper.java:112)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at software.amazon.awssdk.core.internal.util.ClassLoaderHelper.loadClass(ClassLoaderHelper.java:71)
[docker-java-stream-1118179615] INFO logback - STDOUT: 	at software.amazon.awssdk.core.interceptor.ClasspathInterceptorChainFactory.createExecutionInterceptor(ClasspathInterceptorChainFactory.java:123)
[...]

PR #4771, opened 2021-12-01
Support testing library/agent instrumentation interoperability

trask:
I don't think the removed code is needed anymore since the library instrumentation is shaded inside of the agent even during tests now. But let's see what CI says...

Removing this code will allow agent tests to be written that use the library instrumentation and verify interoperability between the two.


anuraaga:
@trask Looks like CI might not like this PR, but also the tests not catching https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4718 makes me think that either an improvement or alternative to this snippet is still needed

anuraaga:
@trask Thanks for looking at this :) By the way, the idea I had in mind to remove this snippet and make things hopefully more robust is a custom `configuration` type for instrumentation dependencies we bundle. We currently have `library`, which is like `compileOnly` + `testImplementation` (with the ability to raise the version number to latest), we could have a `bundled` which is like `compileOnly` + `shadowInclude` (semantically, we don't shade each individual project so this configuration would need to be recognized at the bundle site). And this would then make sure the files aren't in the test classpath in a more Gradle-native way than my hack

trask:
> the idea I had in mind to remove this snippet and make things hopefully more robust is a custom `configuration` type for instrumentation dependencies we bundle. We currently have `library`, which is like `compileOnly` + `testImplementation` (with the ability to raise the version number to latest), we could have a `bundled` which is like `compileOnly` + `shadowInclude` (semantically, we don't shade each individual project so this configuration would need to be recognized at the bundle site). And this would then make sure the files aren't in the test classpath in a more Gradle-native way than my hack around our current usage of `implementation`

This sounds cool, though I'll probably leave this part for the gradle magicians ✨

I think I've about worked out what should / should not be on the class path in order to enable writing tests that verify interoperability between agent and library instrumentation (see also #4779).

trask:
CI still hasn't passed yet, but I'm feeling good about this next run, so optimistically marking it ready for review 😂

conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-testing.gradle.kts
```diff
@@ -106,22 +106,24 @@ afterEvaluate {
...
       // plugin identification instead.
 
-      // If agent depends on some shared instrumentation module that is not a testing module, it will
-      // be packaged into the testing jar so we need to make sure to exclude from the test classpath.
       val lib = it.absoluteFile
-      val instrumentationDir = file("${rootDir}/instrumentation/").absoluteFile
-      if (lib.startsWith(instrumentationDir) &&
-        lib.extension == "jar" &&
-        !lib.absolutePath.substring(instrumentationDir.absolutePath.length).contains("testing")) {
+      if (lib.name.startsWith("opentelemetry-javaagent-")) {
```
iNikem, line 119: If `lib` is an `absoluteFile`, how can it start with `opentelemetry-javaagent`?

trask: it's `lib.name`, e.g. `File.getName()`, not the full path

mateuszrzeszutek (approved):

PR #4772, opened 2021-12-01
Try to make tomcat async servlet exception test less flaky

laurit:
According to https://ge.opentelemetry.io/scans/tests?search.relativeStartTime=P7D&search.tags=CI&search.timeZoneId=Europe/Tallinn&tests.sortField=FLAKY&tests.unstableOnly=true `test exception` with async servlet on tomcat is the most flaky test we have. Here is an attempt to make it stable by forcing response to be sent before exception is thrown which could trigger tomcat resetting connection to the client.


trask (approved):
👍🤞

(more weirdness with exceptions thrown from `AsyncContext.start()` 😅)

iNikem (approved):

mateuszrzeszutek (approved):

PR #4774, opened 2021-12-01
Use versioned download url

### LINKED ISSUE #3516: Inconsistencies with version tags

trask:
IIRC this was raised when we were having problems depending on collector `latest` images, and there was interest encouraging the "best practice" of using versioned download urls.

Closes #3516

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ if you are looking for documentation on using those.
 
 ## Getting Started
 
-Download the [latest version](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/latest/download/opentelemetry-javaagent.jar).
+Download the [latest version](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v1.9.1/opentelemetry-javaagent.jar).
 
 This package includes the instrumentation agent as well as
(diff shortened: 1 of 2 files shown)


iNikem (approved):
We have to find a way to automate all this version management.

mateuszrzeszutek (approved):

mateuszrzeszutek:
> We have to find a way to automate all this version management.

... bash scripts and `sed`? 😅 
That's what I did in the Splunk distro, and it does make releasing it easier.

Issue #4246, opened 2021-09-29
Reproducible tag builds


trask:
1.5.x and 1.6.x patch releases have required several cherry-picks to "fix the build".

This is caused by a few things:
* Use of collector `latest` docker image - this is solved recently by #4186
* Use of gradle-plugins SNAPSHOT - this is solved recently by #4183
* New library version releases, which can cause `muzzle` verification and `testLatestDeps` to fail - short of pinning all "latest library versions"(?), this can potentially be solved by not running `muzzle` and `testLatestDeps` as part of patch releases, or we can live with this, as it's not hard to limit versions if needed (and shouldn't require backporting)

trask:
Closing, all of the above have been resolved, as we are no longer running `testLatestDeps` or `muzzle` as part of release jobs.

The only thing that causes problems still is that the release builds run using the the yaml configuration from `main`. I will open a separate issue to track that, in case there's something that can be done.

trask:
> The only thing that causes problems still is that the release builds run using the the yaml configuration from `main`. I will open a separate issue to track that, in case there's something that can be done.

oh! somehow never noticed this 🥳 I will update the patch release instructions

![image](https://user-images.githubusercontent.com/218610/144327443-d2ef6ab1-38ba-49b7-af19-8114cca88d49.png)

PR #4776, opened 2021-12-01
v1.9.x: cherry pick patch release yaml updates

trask:
Clean cherry-picks of #4713, #4741, #4756, #4758

These were the fixes to the `patch-release-build.yml` that were made in `main` in order to make the `1.9.1` release, but now that we want to use the workflow from the release branch (#4775), we need these fixes in the `v1.9.x` release branch in case we need to make a `1.9.2` patch release.

### DIFF
--- a/.github/workflows/patch-release-build.yml
+++ b/.github/workflows/patch-release-build.yml
@@ -252,7 +252,7 @@ jobs:
         with:
           job-id: jdk11
           remote-build-cache-proxy-enabled: false
-          arguments: assemble final closeAndReleaseSonatypeStagingRepository -Prelease.version=${{ github.event.inputs.version }}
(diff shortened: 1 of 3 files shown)


anuraaga (approved):
>  we need these fixes in the v1.9.x release branch in case we need to make a 1.9.2 patch release.

I think this can be good bookkeeping anyways, but technically this isn't true since workflows always run from main right?

trask:
> I think this can be good bookkeeping anyways, but technically this isn't true since workflows always run from main right?

see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4246#issuecomment-984132856 🥳

anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

PR #4779, opened 2021-12-02
Rename autoconfigure modules

### LINKED ISSUE #4778: Use new autoconfigure module naming convention where applicable
Following the recent new convention added to separate aws-sdk `library` and `library-autoconfigure` instrumentation, we should rename other library instrumentation modules that autoconfigure themselves.

This is a useful distinction because
* `-autoconfigure` modules have to use `GlobalOpenTelemetry`, and there's no way for the user to supply their own `OpenTelemetry` instance, so we may want to add `library` (non-autoconfigure) modules later on if there's user demand
* `javaagent` module tests can run with `library` instrumentation on the test classpath (which can be useful to test interoperability), but they should not generally be run with `library-autoconfigure` instrumentation on the test classpath, since that will get applied and the tests may pass even if the javaagent instrumentation fails to apply

I think these are the two existing modules that fall into this category and should be renamed:

* `apache-dubbo-2.7:library` --> `apache-dubbo-2.7:library-autoconfigure`
* `log4j:log4j-2.13.2:library` --> `log4j:log4j-2.13.2:library-autoconfigure`

trask:
Resolves #4778 (and needed for #4771)


settings.gradle.kts
```diff
@@ -264,7 +264,7 @@ include(":instrumentation:liberty:liberty-dispatcher:javaagent")
 include(":instrumentation:log4j:log4j-1.2:javaagent")
 include(":instrumentation:log4j:log4j-2.7:javaagent")
 include(":instrumentation:log4j:log4j-2.13.2:javaagent")
-include(":instrumentation:log4j:log4j-2.13.2:library")
+include(":instrumentation:log4j:log4j-2.13.2:library-autoconfigure")
```
anuraaga, line 267: I think so but just want to confirm we're comfortable with this naming for libraries that don't use `GlobalOpenTelemetry`.

anuraaga (approved):

settings.gradle.kts
anuraaga: Also - ok for followup but I guess we do need to split the appender out sooner than later for this one

trask: Good point. Maybe we define `autoconfigure` as any module that is enabled automatically just by being on the class path.

trask: I think the `-autoconfigure` naming is a nice indication to users that there's nothing they have to do to register/configure it, so seems reasonably worth it even if there's would never be a non-autoconfigure module.

trask: > Also - ok for followup but I guess we do need to split the appender out sooner than later for this one

oh yes, created #4780 and set milestone for `1.10.0` to limit breakage

jack-berg: Just seeing this now. This case is interesting because parts of this module are enabled automatically (i.e. the context provider) while other have to be manually enabled (i.e. the log4j2 log appender for the log sdk). 

Thoughts?

anuraaga: I think we would have the appender only in library and the context provider in auto, so basically split in half

Issue #4317, opened 2021-10-07
Move io.opentelemetry.instrumentation.api.servlet to servlet-common


iNikem:
`io.opentelemetry.instrumentation.api.servlet` package from `instrumentation-api` module is used only by servlets and some methods from it are used only from `servlet-common` module. We can move that package closer to its usage and hide those methods from general public, who should not be concerned with them.

Thoughts? I can do it if get enough support.

laurit:
Wouldn't it make sense to split api into public and internal. Internal api could contain classes that are not ready to be used outside the agent like `ServerSpanNaming` which wouldn't work without bridging context key. Or we could move internal apis under `javaagent-bootstrap`.

PR #4785, opened 2021-12-02
Reorganize shared servlet code (intro to #4317)

mateuszrzeszutek:
#4317 will introduce a bootstrap module that'll contain classes moved from `instrumentation-api`. Library modules should not depend on a bootstrap module, because that's a javaagent only concept -- in fact, the whole servlet instrumentation should not have any library modules because a servlet library instrumentation does not exist.
In this PR I moved some of those shared "library" classes around and removed the servlet library modules.


trask (approved):
👍

laurit (approved):

iNikem (approved):

Issue #4666, opened 2021-11-18
Context key shading: multiple instances


lmolkova:
**What I want to achieve:** 

Libraries that are not shaded into agent should be able to use instrumentation-api and other instrumentation libraries directly and it should work seamlessly with the agent.

**Scenario 1:**
- library instrumentation does not ship with agent
- library wants to pass spans in Reactor context using `ContextPropagationOperator.store...` so that nested instrumentations would pick up parent context.
- I'm trying to make it work along with agent instrumentation (i.e. nested instrumentation correlates properly)

**Scenario 2:**
~~Same as 1, but I want to set [SpanKey.ALL_CLIENTS](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6d9e3618d3876f91fac59f87f83cffd7fb9ecc5e/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/SpanKey.java#L52) to suppress nested CLIENTS~~

**Update:** SpanKeys just work because of context key bridging 

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/30b231645f7f5902bd4c48db8a8995af3eac24dd/instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/context/AgentContextStorage.java#L158

**What doesn't work: shaded context keys**

Interaction between an external library and everything the agent is hard and fragile because of shaded context keys and context.

E.g. `io.opentelemetry.instrumentation.reactor.ContextPropagationOperator` from `opentelemetry-reactor-3.1` is different from what agent uses `io.opentelemetry.javaagent.shaded.instrumentation.reactor.ContextPropagationOperator`, i.e. static context key is another instance

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/6d9e3618d3876f91fac59f87f83cffd7fb9ecc5e/instrumentation/reactor-3.1/library/src/main/java/io/opentelemetry/instrumentation/reactor/ContextPropagationOperator.java#L52

[Here's an example](https://gist.github.com/lmolkova/af083090744eea9eabc6f9d19fc16b16) for reactor that shows the problem (need to run with agent)

**Describe the solution you'd like**

[...]

PR #4786, opened 2021-12-02
Instrument ContextPropagationOperator to bridge lib/agent calls

lmolkova:
Fixes reactor part of #4666

### DIFF
--- a/instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
+++ b/instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
@@ -53,7 +53,9 @@ public class ContextPropagationOperatorInstrumentation implements TypeInstrument
             .and(isPublic())
             .and(isStatic())
             .and(named("runWithContext"))
-            .and(takesArgument(0, namedOneOf("reactor.core.publisher.Mono", "reactor.core.publisher.Flux")))
+            .and(
+                takesArgument(
+                    0, namedOneOf("reactor.core.publisher.Mono", "reactor.core.publisher.Flux")))
             .and(takesArgument(1, named("application.io.opentelemetry.context.Context")))
             .and(returns(namedOneOf("reactor.core.publisher.Mono", "reactor.core.publisher.Flux"))),
         ContextPropagationOperatorInstrumentation.class.getName() + "$RunWithAdvice");
--- a/instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
+++ b/instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.javaagent.instrumentation.reactor;
 
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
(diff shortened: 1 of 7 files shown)


instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+import io.opentelemetry.javaagent.extension.instrumentation.TypeTransformer;
+import io.opentelemetry.javaagent.instrumentation.opentelemetryapi.context.AgentContextStorage;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class ContextPropagationOperatorInstrumentation implements TypeInstrumentation {
+  @Override
+  public ElementMatcher<TypeDescription> typeMatcher() {
+    return named("application.io.opentelemetry.instrumentation.reactor.ContextPropagationOperator");
```
trask, line 27: 👍

trask (approved):
thx @lmolkova!

@mateuszrzeszutek @anuraaga would be great to get your thoughts on this. It's similar to the bridging we do on the OpenTelemetry API, but applied to a library instrumentation, which feels a little odd since typically I think of users using either library instrumentation -or- javaagent instrumentation, but not needing to use both. Maybe we could split out part of the reactor instrumentation module into an `opentelemetry-extension-reactor` module (similar to [opentelemetry-extension-kotlin](https://github.com/open-telemetry/opentelemetry-java/tree/main/extensions/kotlin))?

iNikem:
> which feels a little odd since typically I think of users using either library instrumentation -or- javaagent instrumentation, but not needing to use both.

Library itself can provide native/library instrumentation and then app developer brings in the agent. So this scenario is totally possible.

instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
```diff
@@ -0,0 +1,110 @@
...
+            .and(takesArgument(0, named("reactor.util.context.Context")))
+            .and(takesArgument(1, named("application.io.opentelemetry.context.Context")))
+            .and(returns(named("application.io.opentelemetry.context.Context"))),
+        ContextPropagationOperatorInstrumentation.class.getName() + "$GetAdvice");
+    transformer.applyAdviceToMethod(
+        isMethod()
+            .and(isPublic())
+            .and(isStatic())
+            .and(named("runWithContext"))
+            .and(takesArgument(0, named("reactor.core.publisher.Mono")))
```
mateuszrzeszutek, line 55: What about the `Flux` variant? Shouldn't it be instrumented too?

mateuszrzeszutek:
> @mateuszrzeszutek @anuraaga would be great to get your thoughts on this. It's similar to the bridging we do on the OpenTelemetry API, but applied to a library instrumentation, which feels a little odd since typically I think of users using either library instrumentation -or- javaagent instrumentation, but not needing to use both.

I like this PR (and muzzle changes in #4797); perhaps we'll need to implement a similar bridge for the `instrumentation-api` in the future (other than the `SpanKey`s).

instrumentation/reactor-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/reactor/ContextPropagationOperatorInstrumentation.java
lmolkova: it's currently package-private, so there is no use-case where instrumentation is needed. I can make it public and instrument or maybe add a comment that instrumentation is needed before it can be made public.

mateuszrzeszutek: It is used indirectly by the `@WithSpan` instrumentation though; and I guess it should probably be made public anyway, the `Mono` variant turned out to be super useful in reactor-netty instrumentation.

lmolkova: `@WithSpan` instrumentation in  this case deals with the shaded version of the context key and context, so it should be fine. But I agree, making `Flux` version public makes total sense - I'll update the PR.

mateuszrzeszutek (approved):
👍 
Thanks!

PR #4787, opened 2021-12-02
Ratpack httpclient

jsalinaspolo:
Proposal to support Ratpack HttpClient Instrumentation. The intention of the PR is to start a conversation about the proposal

🎉 Ratpack `1.7.0` supports HttpClient request/response/error interceptors, therefore ratpack version have been upgraded

* Does Ratpack HttpClient needs `ContextPropagation`, I have been unable to see the need but I am probably missing something.
* Should `RatpackTracingBuilder` support `RatpackHttpTracing` or should be treated independently


anuraaga:
Thanks #4787 - I think bumping the version makes sense, 1.7 is pretty old and I haven't heard any demand for supporting older versions. Do you mind sending an initial PR with just the version bump to make these easier to review?

jsalinaspolo:
Better to leave the renaming of the module for after the review 🙇

iNikem:
As I understand this introduces new library instrumentation for Ratpack and this requires v1.7. Do we actually need to upgrade our server instrumentation to 1.7 as well or we can leave it to continue to work with 1.4?

jsalinaspolo:
We can create a new module to ratpack-1.7 and leave the ratpack-1.4 as it is, but I am not sure if there is any need to support ratpack-1.4 that is over 4 years old. Something I have been thinking is if we should be validating that the instrumentation works with newer version as well. What is the current approach for other library instrumentation?

laurit:
@jsalinaspolo 
> Something I have been thinking is if we should be validating that the instrumentation works with newer version as well. What is the current approach for other library instrumentation?

When you run tests with`-PtestLatestDeps=true` the version used in `library` and `testLibrary` dependencies is automatically changed to latest available version. If the latest available version doesn't work for some reason you can use `latestDepTestLibrary`. Tests against latest version are run on merge to main and during nightly builds.

jsalinaspolo:
> Thanks #4787 - I think bumping the version makes sense, 1.7 is pretty old and I haven't heard any demand for supporting older versions. Do you mind sending an initial PR with just the version bump to make these easier to review?

Sorry @anuraaga, I misunderstood you and undo the renaming of the module so the PR is easier to review. Do you think that makes sense to keep this PR to discuss the instrumentation approach of the Ratpack HttpClient and once it is reviewed, we can rename the module to `1.7`?

@iNikem the instrumentation of the Ratpack HttpClient requires interceptors, therefore we need version `1.7`. If you think that we need to keep version `1.4`, then we could create a new module with `ratpack-1.7` with the HttpClient instrumentation. Personally, I think would be better to bump to `1.7` altogether, anyway `1.7` is over 2 years old

anuraaga:
I was hoping to do the version bump first, including directory rename, if there is consensus for it. It's a bit tricky discerning changes made for the bump vs ones for client instrumentation.

> We can create a new module to ratpack-1.7 and leave the ratpack-1.4 as it is, but I am not sure if there is any need to support ratpack-1.4 that is over 4 years old.

My vote is to bump for now - that being said I think we need a proper process for sunsetting instrumentation, which we  haven't really done before. Perhaps we should target the next release for this change, first updating current instrumentation to log a warning that the version floor will change. Anyone have any thoughts?

trask (commented):
Can you add http client library tests that extends `AbstractRatpackHttpClientTest`? or at least tests that extends `AbstractHttpClientTest` if the `AbstractRatpackHttpClientTest` is not appropriate for library instrumentation? the base http client tests run a large array of standard tests to ensure that all of our http client instrumentations generate consistent telemetry.

jsalinaspolo:
> Can you add http client library tests that extends `AbstractRatpackHttpClientTest`? or at least tests that extends `AbstractHttpClientTest` if the `AbstractRatpackHttpClientTest` is not appropriate for library instrumentation? the base http client tests run a large array of standard tests to ensure that all of our http client instrumentations generate consistent telemetry.

I've been digging into the `AbstractHttpClientTest` and figuring out how was working. I've added propagation to the HttpClient but there are a few tests that are failing. Most of them are related to having more traces than expected, can anyone have a look to see if are familiar why it might be happening?

instrumentation/ratpack/ratpack-1.7/library/src/main/java/io/opentelemetry/instrumentation/ratpack/client/OpenTelemetryHttpClient.java
```diff
@@ -0,0 +1,62 @@
...
+                Context parentOtelCtx = Context.current();
+                if (!instrumenter.shouldStart(parentOtelCtx, requestSpec)) {
+                  return;
+                }
+
+                Context otelCtx = instrumenter.start(parentOtelCtx, requestSpec);
+                Span span = Span.fromContext(otelCtx);
+                String path = requestSpec.getUri().getPath();
+                span.updateName(path);
+//                span.updateName("HTTP " + requestSpec.getMethod().getName()); // TODO use path instead of [HTTP method]
```
jsalinaspolo, line 40: Based on the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#name) should be possible to leave name like `/api/users/{user_id}` instead of `HTTP GET` ?

trask:
> I've been digging into the `AbstractHttpClientTest` and figuring out how it is working. I've added context propagation to the HttpClient but still, a few tests are failing. Most of them are related to having more traces than expected, but I am not sure if it is related to suppressing nested CLIENT spans or something else.

thx @jsalinaspolo! i'll have a look at the failing tests tomorrow 👍

instrumentation/ratpack/ratpack-1.7/library/src/main/java/io/opentelemetry/instrumentation/ratpack/client/OpenTelemetryHttpClient.java
trask: it depends on if path is a template (e.g. `/api/users/{user_id}`) or if it's a real url path (e.g. `/api/users/5551234`)

trask:
hey @jsalinaspolo, thanks for your patience. the problem in the tests appears to be that they are losing (local) context propagation across threads, somewhere between where the (synthetic) "parent" span is created and where the http client span is created.

the reason the same tests probably work in the javaagent tests is because the javaagent instruments java executors in order to propagate context across threads

I think your other library test `InstrumentedHttpClientTest` is working because you are using `RatpackTracing.create(openTelemetry)` over there which looks like it registers `OpenTelemetryExecInterceptor.INSTANCE` to take care of context propagation for ratpack threads.

trask:
hey @jsalinaspolo, if you add these printlns to AbstractRatpackHttpClientTest and look at `successfulRequestWithParent`, you can see that the current context is still getting lost during the `yield`:

```
  @Override
  protected final int sendRequest(Void request, String method, URI uri, Map<String, String> headers)
    throws Exception {
    println "==========outer==============="
    println Span.current()
    println "========================="
    return exec.yield {
      println "==========inner==============="
      println Span.current()
      println "========================="
      internalSendRequest(client, method, uri, headers)
    }.getValueOrThrow()
  }
```

jsalinaspolo:
Thanks @trask 

I think the problem is that Ratpack HttpClient is not design to be used outside of Ratpack managed threads. I believe the other tests are working because starts from Ratpack Server and then propagates it to the HttpClient.

Let me dig a little bit to see if there is any "hacky" way to pass the context to the Executor from the ExecHarness 🤔

jsalinaspolo:
Thanks @trask for looking into it and the hint.

Finally, only was required to pass the current `Context` to the test execution like [here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4787/commits/910d7dd592e65c4883ad3ca13f1277bafa7aa4f1#diff-7b8d26a80b980df052011049f471f169b6801c0335165d36209fb700730eff65R89)

trask:
@jsalinaspolo thanks for working through the tests! there's one more test issue, can you run this locally and check the failure?

```
./gradlew :instrumentation:ratpack:ratpack-1.7:library:test -PtestLatestDeps=true
```

`testLatestDeps` will run the tests against the latest version of ratpack in maven central (there are ways to configure this if there's a reason we intentionally don't support the latest version yet).

sometimes failures against the latest dependency point to issues in the instrumentation, and sometimes issues with the tests themselves. if it's not clear how to make the instrumentation and tests work against both the base version and latest version, go ahead and post your findings, and we can help you sort it out (we have a variety of ways to deal with this common issue)

jsalinaspolo:
@trask looks like an issue in the functional test to start the `otherApp`. 
Should be fix now by [this](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4787/commits/c559781e63bb72cb49bfe8885d82ce84f1054202) 🙇

instrumentation/ratpack/ratpack-1.7/library/src/main/java/io/opentelemetry/instrumentation/ratpack/client/RatpackHttpTracing.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import ratpack.exec.ExecInitializer;
+import ratpack.http.client.HttpClient;
+import ratpack.http.client.HttpResponse;
+import ratpack.http.client.RequestSpec;
+
+/** Entrypoint for tracing OkHttp clients. */
+public final class RatpackHttpTracing {
```
anuraaga, line 16: Because Ratpack HTTP client can really only be used in a Ratpack server, do you think it makes sense to just keep one package and combine the entrypoints into one? Notably `configureServerRegistry` can add the exec initializer too to keep things simpler.

jsalinaspolo: As far as we are happy to add the exec initializer at all times, I think might be easier to configure.

The only downside would be having too many classes in a package without structure while having `client` `server` and `internal` might help to clarify. 

Having different packages with a single `RatpackTracing` have the problem that a few classes would need to be `public`, which I understood was not ideal.

What do you think would be nicer? @anuraaga

jsalinaspolo: I made a refactor to have a flatter package structure

jsalinaspolo:
@anuraaga @trask is the PR good to go?

instrumentation/ratpack/ratpack-1.7/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTracing.java
```diff
@@ -57,9 +65,20 @@ public ExecInterceptor getOpenTelemetryExecInterceptor() {
...
+
   /** Configures the {@link RegistrySpec} with OpenTelemetry. */
   public void configureServerRegistry(RegistrySpec registry) {
     registry.add(HandlerDecorator.prepend(serverHandler));
     registry.add(OpenTelemetryExecInterceptor.INSTANCE);
+    registry.add(OpenTelemetryExecInitializer.INSTANCE);
+  }
+
+  /** Returns instrumented instance of {@link HttpClient} with OpenTelemetry. */
+  public HttpClient instrumentedHttpClient(HttpClient httpClient) throws Exception {
```
anuraaga, line 81: Let's name it `instrumentHttpClient`

anuraaga (approved):
Just small point thanks!

trask:
thx @jsalinaspolo!

PR #4797, opened 2021-12-04
Run muzzle check against OpenTelemetry API instrumentation (bridge)

trask:
Addresses part of #3044

This PR changes muzzle checks to run against shaded artifacts (similar to how tests are run against shaded artifacts). The main reason for doing this is to enable running muzzle checks against OpenTelemetry API instrumentation.

This will help unblock PR #4786 which is currently failing on `:instrumentation:reactor-3.1:javaagent:muzzle` due to `java.lang.ClassNotFoundException: application.io.opentelemetry.context.ContextStorage`


gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
```diff
@@ -38,14 +41,37 @@ val muzzleTooling: Configuration by configurations.creating {
...
   isCanBeResolved = true
 }
+
 val muzzleBootstrap: Configuration by configurations.creating {
   isCanBeConsumed = false
   isCanBeResolved = true
 }
 
+val muzzleShadow by tasks.registering(ShadowJar::class) {
+  from(tasks.jar.get().archiveFile.get())
```
iNikem, line 51:
```suggestion
  from(tasks.jar)
```

gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
```diff
@@ -38,14 +41,37 @@ val muzzleTooling: Configuration by configurations.creating {
...
+
+  configurations = listOf(project.configurations.runtimeClasspath.get())
```
iNikem, line 53: this is default, no?

gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
```diff
@@ -38,14 +41,37 @@ val muzzleTooling: Configuration by configurations.creating {
   isCanBeConsumed = false
   isCanBeResolved = true
 }
+
 val muzzleBootstrap: Configuration by configurations.creating {
   isCanBeConsumed = false
   isCanBeResolved = true
 }
 
+val muzzleShadow by tasks.registering(ShadowJar::class) {
```
iNikem, line 50: I don't understand why this shadow jar is needed?

gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
trask: defaults are only applied to the default "shadowJar" task, see https://github.com/johnrengelman/shadow/issues/108#issuecomment-62418005

gradle-plugins/src/main/kotlin/io.opentelemetry.instrumentation.muzzle-check.gradle.kts
trask: this shades the javaagent module classes themselves

iNikem: Rename it to `shadedModule` then? Or something similar. We don't shade muzzle here, do we?

mateuszrzeszutek (approved):

iNikem (approved):

PR #4802, opened 2021-12-06
Add guidance about dropping support for older library versions

trask:
This has come up a couple of times:

* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3780#issuecomment-895674756
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4796#issuecomment-986195360

time to try to write something down.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -72,6 +72,6 @@ that do not require bumping the minimum supported library version include:
 * Create a new javaagent instrumentation module to support the new library version. This requires
   configuring non-overlapping versions in the muzzle check and applying `assertInverse` to confirm
   that the two instrumentations are never be applied to the same library version (see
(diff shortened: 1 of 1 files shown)


iNikem (approved):
By this rules, we should not just drop support for ratpack 1.4 in javaagent instrumentation :)

trask:
I wouldn't call it a rule, as it still allows dropping old versions in javaagent instrumentation on a case-by-case basis. But I do generally think the bar is pretty high (and I think it's hard to define what exactly that bar is, which is why I left it as "case-by-case").

Moving javaagent instrumentations that supports old library versions to `contrib` is one potential mitigation to the growing maintenance burden, while still being usable by users and downstream distros. Hopefully those old instrumentation versions won't require active maintenance once all of the APIs they depend on are marked stable.

mateuszrzeszutek (approved):

laurit (approved):

anuraaga (approved):

PR #4803, opened 2021-12-06
Fix gradle warning

trask:
Fixes gradle warning:

```
> Task :gradle-plugins:compileKotlin
'compileJava' task (current target is 11) and 'compileKotlin' task (current target is 1.8) jvm target compatibility should be set to the same Java version.
```

### DIFF
--- a/gradle-plugins/build.gradle.kts
+++ b/gradle-plugins/build.gradle.kts
@@ -21,6 +21,10 @@ configurations.named("compileOnly") {
   extendsFrom(bbGradlePlugin)
 }
 
+java {
+  toolchain.languageVersion.set(JavaLanguageVersion.of(11))
+}
+
 dependencies {
   implementation("com.google.guava:guava:30.1.1-jre")
   // we need to use byte buddy variant that does not shade asm


anuraaga (approved):

gradle-plugins/build.gradle.kts
```diff
@@ -21,6 +21,10 @@ configurations.named("compileOnly") {
   extendsFrom(bbGradlePlugin)
 }
 
+java {
+  toolchain.languageVersion.set(JavaLanguageVersion.of(11))
+}
+
```
anuraaga, line 27: Should we also have this?

https://github.com/open-telemetry/opentelemetry-java/blob/main/extensions/kotlin/build.gradle.kts#L36

trask: hm, after re-testing, it doesn't seem like my change gets rid of this warning.

adding the setting for kotlin doesn't seem to help either:

```
tasks.withType<KotlinCompile>().configureEach {
  kotlinOptions {
    jvmTarget = "11"
  }
}
```

googling around I found https://github.com/gradle/gradle/issues/18935 which seems to be the issue here

trask:
Closing this, I don't think I solved anything, probably need to wait for  gradle/gradle#18935

PR #4806, opened 2021-12-06
Propagate Context instead of SpanContext

trask:
I haven't noticed any problems with the existing code, just changing from `SpanContext` to `Context` to be consistent with other instrumentations where we propagate `Context` where possible.


trask:
there's a little wrinkle with this approach around suppressing nested spans, see last 2 commits

iNikem (approved):

mateuszrzeszutek (approved):

PR #4811, opened 2021-12-06
Add requested java and smoke test suite as build tags

iNikem:


mateuszrzeszutek (approved):

settings.gradle.kts
```diff
@@ -59,6 +59,12 @@ if (useScansGradleCom) {
       capture {
         isTaskInputFiles = true
       }
+
+      gradle.startParameter.projectProperties["testJavaVersion"]?.let { tag(it) }
+      gradle.startParameter.projectProperties["testJavaVM"]?.let { tag(it) }
+      gradle.startParameter.projectProperties["smokeTestSuite"]?.let {
+        value("Smoke test suite", it)
```
trask, line 66: @iNikem what's the difference between `tag` and `value`? thx

trask: nm I found it: https://blog.gradle.org/custom-data-in-build-scans

Issue #4694, opened 2021-11-22
opentelemetry-api not mentioned in suppressing-instrumentation.md


tomasol:
Document [suppressing-instrumentation.md](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/suppressing-instrumentation.md) describes how to suppress instrumentations. 

It does mention the other way around, disabling everything and selectively enabling some instrumentations, but fails to mention `opentelemetry-api` which seems to be responsible for installing the tracer into `GlobalOpenTelemetry`.

PR #4818, opened 2021-12-06
Update supported frameworks and suppressing instrumentation docs

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4694
An alternative would be to make `opentelemetry-api` instrumentation an internal instrumentation that would not be disabled by default enabled flag.

### DIFF
--- a/docs/suppressing-instrumentation.md
+++ b/docs/suppressing-instrumentation.md
@@ -75,7 +75,7 @@ corresponding instrumentation `name`:
 | OkHttp | okhttp|
 | OpenLiberty | liberty |
 | OpenTelemetry Trace annotations | opentelemetry-annotations |
-| OpenTelemetry SDK | opentelemetry-api |
+| OpenTelemetry API | opentelemetry-api |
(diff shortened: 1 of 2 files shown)


docs/suppressing-instrumentation.md
```diff
@@ -72,9 +75,11 @@ corresponding instrumentation `name`:
 | OkHttp | okhttp|
 | OpenLiberty | liberty |
 | OpenTelemetry Trace annotations | opentelemetry-annotations |
+| OpenTelemetry SDK | opentelemetry-api |
```
trask, line 78: since we are instrumenting specific portions of the OpenTelemetry SDK (e.g. also `opentelemetry-annotations` above)
```suggestion
| OpenTelemetry API | opentelemetry-api |
```

trask (approved):

mateuszrzeszutek (approved):

iNikem (approved):

PR #4820, opened 2021-12-07
Add javadoc note about metrics timestamps

trask:
I wasn't sure where to add the note, so added it to two places, happy to remove one.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/TimeExtractor.java
```diff
@@ -14,6 +14,13 @@
   /** Returns the timestamp marking the start of the request processing. */
   Instant extractStartTime(REQUEST request);
 
-  /** Returns the timestamp marking the end of the response processing. */
+  /**
+   * Returns the timestamp marking the end of the response processing.
+   *
+   * <p>Note: if metrics are generated by the Instrumenter, the start and end times from the {@code
```
mateuszrzeszutek, line 20: WDYT about moving it to class level? There are some metrics (`http.server.active_requests`) that receive new values on both `start()` and `end()`.

mateuszrzeszutek (approved):

Issue #442, opened 2020-05-26
`http.route` attribue


iNikem:
[Semantic convention for HTTP server spans](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions) defines attribute `http.route`: 

> The matched route (path template). E.g. "/users/:userID".

Currently instrumentations don't create this attribute. I think we should, but I am not quite sure which span should have this attribute. In case of a typical Java web application the first span created will usually come from generic middleware, such as Servlets or similar. It has no notion of route yet, just the full URL. Then, when request processing reaches actual web framework, such as Spring or jax-rs, second span will be created (potentially with `kind=INTERNAL`). At this point instrumentation can get used route information from the framework. 

Where this attribute should now go? To parent span of kind SERVER (which has all other http attributes like `http.url` and `http.method`) or to a current span of kind INTERNAL, which has no other `http` attributes?

trask:
I believe it should go on the `SERVER` span.

We already have instrumentation to capture the route as part of the `SERVER` span name ([to conform with the _current_ span name guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#name)), so I think this would be added in the same places in the code, calling now `span.setAttribute("http.route", ...)` in addition to `span.updateName(...)`.

iNikem:
@tylerbenson said during last Java SIG meeting that he really dislikes this inter-instrumentation communication, when we modify already existing spans. And I tend to agree with him. If we continue to do this, we HAVE to do that in more explicit and transparent way. Which I don't know yet :)

Can we always be sure that SERVER span is still present when we create child spans? Can it happen that in some async processing server span will be already sent out to collector?

iNikem:
Also, modifying existing spans goes against e.g. #388. Also https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span says:
> It is highly discouraged to update the name of a Span after its creation. Span name is often used to group, filter and identify the logical groups of spans. And often, filtering logic will be implemented before the Span creation for performance reasons. Thus the name update may interfere with this logic.

trask:
[...]

PR #4824, opened 2021-12-07
Move `ServletContextPath` to `servlet-common:bootstrap`

### LINKED ISSUE #4317: Move io.opentelemetry.instrumentation.api.servlet to servlet-common

mateuszrzeszutek:
Resolves #4317

Changes include:
* A couple of `jaxws` "library" instrumentations (which weren't _real_ library instrumentations) were removed and their code moved into their respective javaagent module
* Renamed `jsf-common:library` to `jsf-common:javaagent`
* Restlet library instrumentations are not using `ServletContextPath` anymore (there's no servlet library instrumentation so it makes no sense to use it there anyway)

`instrumentation-api` still contains `ServerSpanNaming`, but this particular class should perhaps be refactored into a general `http.route` utility in #442


instrumentation/jaxws/jaxws-2.0-axis2-1.6/javaagent/build.gradle.kts
```diff
@@ -17,14 +17,15 @@ muzzle {
 }
 
 configurations.configureEach {
-  if (!name.contains("muzzle")) {
-    // axis has a dependency on servlet api, get rid of it
+  if (name.contains("test")) {
```
trask, line 20: 👍

instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/RestletSingletons.java
```diff
@@ -16,9 +18,22 @@
   private static final Instrumenter<Request, Response> INSTRUMENTER =
       RestletTracing.create(GlobalOpenTelemetry.get()).getServerInstrumenter();
 
+  private static final ServerSpanNameSupplier<String> SERVER_SPAN_NAME =
+      (context, pattern) -> {
+        if (pattern == null || pattern.equals("")) {
+          return null;
+        }
+
```
trask, line 26: is this needed? ServletContextPath.prepend handles this case (and still prepends servlet context root)
```suggestion
```

instrumentation/restlet/restlet-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v2_0/RestletSingletons.java
```diff
@@ -16,9 +18,22 @@
   private static final Instrumenter<Request, Response> INSTRUMENTER =
       RestletInstrumenterFactory.newServerInstrumenter(GlobalOpenTelemetry.get());
 
+  private static final ServerSpanNameSupplier<String> SERVER_SPAN_NAME =
+      (context, pattern) -> {
+        if (pattern == null || pattern.equals("")) {
+          return null;
+        }
+
```
trask, line 26: same

instrumentation/spring/spring-webmvc-3.1/javaagent/build.gradle.kts
```diff
@@ -20,6 +20,8 @@ muzzle {
 val versions: Map<String, String> by project
 
 dependencies {
+  bootstrap(project(":instrumentation:servlet:servlet-common:bootstrap"))
```
trask, line 23: it's nice to see/understand these dependencies better now

trask (approved):

instrumentation/restlet/restlet-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/restlet/v1_0/RestletSingletons.java
mateuszrzeszutek: Good point -- I'll simplify that.

PR #4825, opened 2021-12-07
Always forbid system.out by errorprone

iNikem:
This `if` currently breaks CI-Local gradle cache


mateuszrzeszutek (approved):

trask (approved):
interesting. too bad. but makes sense

PR #4826, opened 2021-12-07
Retry strict context check failures for library instrumentation tests

laurit:
Hopefully resolves https://ge.opentelemetry.io/scans/tests?search.relativeStartTime=P7D&search.tags=CI&search.timeZoneId=Europe/Tallinn&tests.container=io.opentelemetry.instrumentation.ratpack.server.RatpackAsyncHttpServerTest&tests.sortField=FLAKY&tests.unstableOnly=true

### DIFF
--- a/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/util/ContextStorageCloser.java
+++ b/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/util/ContextStorageCloser.java
@@ -30,7 +30,7 @@ public final class ContextStorageCloser {
   }
 
   private static void cleanup(AutoCloseable storage) throws Exception {
-    ContextRestorer restorer = ContextRestorer.create();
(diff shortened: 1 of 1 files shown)


testing-common/src/main/java/io/opentelemetry/instrumentation/testing/util/ContextStorageCloser.java
```diff
@@ -43,6 +43,7 @@ private static void cleanup(AutoCloseable storage) throws Exception {
         .ignoreException(AssertionError.class)
         .atMost(Duration.ofSeconds(10))
         .pollInterval(Duration.ofSeconds(1))
+        .pollDelay(Duration.ZERO)
```
laurit, line 46: apparently poll delay is set to poll interval by default which will just slow down the build

mateuszrzeszutek (approved):

trask (approved):
👍

PR #4831, opened 2021-12-07
Expose AutoConfiguredOpenTelemetrySdk to AgentListener

### LINKED ISSUE #4791: AgentListener access to AutoConfiguredOpenTelemetrySdk
As discussed in [this](https://github.com/open-telemetry/opentelemetry-java/issues/3924) `opentelemetry-java` ticket, there is often a need for agent listeners to have access to the autoconfigured `Resource`. To facilitate this and additional future use cases for agent listeners wanting access to the `OpenTelemetrySdk`, we should extend the agent listener interface to allow access to `AutoConfiguredOpenTelemetrySdk`, which has accessors to the things that are autoconfigured.

jack-berg:
Resolves #4791

### DIFF
--- a/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
+++ b/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
@@ -8,13 +8,14 @@ package io.opentelemetry.javaagent.extension;
 import io.opentelemetry.instrumentation.api.cache.Cache;
 import io.opentelemetry.instrumentation.api.config.Config;
 import io.opentelemetry.instrumentation.api.field.VirtualField;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
 import java.lang.instrument.Instrumentation;
 import net.bytebuddy.agent.builder.AgentBuilder;
 
 /**
  * {@link AgentListener} can be used to execute code before/after Java agent installation, for
- * example to install any implementation providers that are used by instrumentations. For instance,
- * this project uses this SPI to install OpenTelemetry SDK.
+ * example to install any implementation providers that are used by instrumentations. Can also be
+ * used to obtain the {@link AutoConfiguredOpenTelemetrySdk}.
  *
  * <p>This is a service provider interface that requires implementations to be registered in a
  * provider-configuration file stored in the {@code META-INF/services} resource directory.
@@ -22,18 +23,36 @@ import net.bytebuddy.agent.builder.AgentBuilder;
 public interface AgentListener extends Ordered {
 
   /**
-   * Runs before the {@link AgentBuilder} construction, before any instrumentation is added.
+   * Runs before {@link AgentBuilder} construction, before any instrumentation is added.
    *
-   * <p>Execute only a minimal code because any classes loaded before the agent installation will
-   * have to be retransformed, which takes extra time, and more importantly means that fields can't
-   * be added to those classes - which causes {@link VirtualField} to fall back to the less
-   * performant {@link Cache} implementation for those classes.
+   * <p>Execute only minimal code because any classes loaded before the agent installation will have
+   * to be retransformed, which takes extra time, and more importantly means that fields can't be
+   * added to those classes - which causes {@link VirtualField} to fall back to the less performant
+   * {@link Cache} implementation for those classes.
    */
   default void beforeAgent(Config config) {}
+
+  /**
+   * Runs before {@link AgentBuilder} construction, before any instrumentation is added. Not called
+   * if noop api enabled via {@code otel.javaagent.experimental.use-noop-api}.
+   *
+   * <p>Execute only minimal code because any classes loaded before the agent installation will have
+   * to be retransformed, which takes extra time, and more importantly means that fields can't be
+   * added to those classes - which causes {@link VirtualField} to fall back to the less performant
+   * {@link Cache} implementation for those classes.
+   */
+  default void beforeAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {}
 
   /**
    * Runs after instrumentations are added to {@link AgentBuilder} and after the agent is installed
    * on an {@link Instrumentation}.
    */
   default void afterAgent(Config config) {}
+
+  /**
+   * Runs after instrumentations are added to {@link AgentBuilder} and after the agent is installed
+   * on an {@link Instrumentation}. Not called if noop api enabled via {@code
+   * otel.javaagent.experimental.use-noop-api}.
+   */
+  default void afterAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {}
 }
--- a/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
+++ b/javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
@@ -29,7 +29,10 @@ public interface AgentListener extends Ordered {
    * to be retransformed, which takes extra time, and more importantly means that fields can't be
    * added to those classes - which causes {@link VirtualField} to fall back to the less performant
    * {@link Cache} implementation for those classes.
+   *
+   * @deprecated Use {@link #beforeAgent(Config, AutoConfiguredOpenTelemetrySdk)}
    */
+  @Deprecated
   default void beforeAgent(Config config) {}
 
   /**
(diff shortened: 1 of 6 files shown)


javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
```diff
@@ -8,32 +8,51 @@
...
-   * <p>Execute only a minimal code because any classes loaded before the agent installation will
-   * have to be retransformed, which takes extra time, and more importantly means that fields can't
-   * be added to those classes - which causes {@link VirtualField} to fall back to the less
-   * performant {@link Cache} implementation for those classes.
+   * <p>Execute only minimal code because any classes loaded before the agent installation will have
+   * to be retransformed, which takes extra time, and more importantly means that fields can't be
+   * added to those classes - which causes {@link VirtualField} to fall back to the less performant
+   * {@link Cache} implementation for those classes.
    */
   default void beforeAgent(Config config) {}
```
jack-berg, line 33: I considered deprecating this but `beforeAgent(AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk)` doesn't get called when there is a noop OpenTelemetry installed. If we think its ok to not have any agent listener hooks when noop is used, then I can go ahead and deprecate this because folks can access `AutoConfiguredOpenTelemetrySdk#getConfig()`.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/OpenTelemetryInstaller.java
```diff
@@ -5,79 +5,73 @@
...
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@AutoService(AgentListener.class)
-public class OpenTelemetryInstaller implements AgentListener {
+public class OpenTelemetryInstaller {
   private static final Logger logger = LoggerFactory.getLogger(OpenTelemetryInstaller.class);
 
-  static final String JAVAAGENT_ENABLED_CONFIG = "otel.javaagent.enabled";
```
jack-berg, line 30: This constant was a duplicate of of `AgentInstaller#JAVAAGENT_ENABLED_CONFIG`.

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -110,7 +112,9 @@ public static ResettableClassFileTransformer installBytebuddyAgent(
 
     setBootstrapPackages(config);
 
-    runBeforeAgentListeners(agentListeners, config);
+    AutoConfiguredOpenTelemetrySdk autoConfiguredSdk = installOpenTelemetrySdk(config);
```
jack-berg, line 115: `installOpenTelemetrySdk(Config)` is `@Nullable` because config may dictate a noop `OpenTelemetry`. But how is a noop `OpenTelemetry` different than just deactivating the agent altogether via `otel.javaagent.enabled=false`?

trask: see #3053 for history / purpose of noop

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
trask: it would be really nice to deprecate this. not sure what to do about `noop`. @jkwatson @breedx-splk are you finding `noop` useful?

breedx-splk: I do think the noop is useful.

Over in [the original issue in the other repo](https://github.com/open-telemetry/opentelemetry-java/issues/3924) we chatted about the two arg `(config, sdk)` version being preferred. This two-version approach requires the SPI user to override both and hold state if they need both. It also looks weird (to me) at the call site below, because, depending on the implementation, `beforeAgent()` can be called twice.

My preference would be to just have the two-arg version and then later when the config problem (Config vs. ConfigProperties) is solved we can remove the config arg.

How's that sound?

trask: > My preference would be to just have the two-arg version and then later when the config problem (Config vs. ConfigProperties) is solved we can remove the config arg.

If we continue supporting `noop`, how would we remove the config arg once the config problem is solved? since in the `noop` case `AutoConfiguredOpenTelemetrySdk` is null, so it seems like we still need to pass around agent configuration somehow

breedx-splk: What about a noop implementation equivalent of the `AutoConfiguredOpenTelemetrySdk`?

mateuszrzeszutek: > What about a noop implementation equivalent of the `AutoConfiguredOpenTelemetrySdk`?

I like that idea 👍 
There are some instrumentations that are implemented as agent listener (e.g. metrics), or other features (like the profiler in splunk's case) that should be initialized even if the SDK is not initialized - so that we're able to measure the impact of these without the SDK. Not calling the listeners if noop OpenTelemetry is installed is kinda hacky and unreliable, and will produce wrong results in overhead tests.

How about implementing the listener this way:

```java
interface AgentListener {
  
  default void beforeAgent(OpenTelemetry openTelemetry, Resource resource, Config config) {
    beforeAgent(config);
  }

  @Deprecated
  default void beforeAgent(Config config) {}
  
  default void afterAgent(OpenTelemetry openTelemetry, Resource resource, Config config) {
    afterAgent(config);
  }

  @Deprecated
  default void afterAgent(Config config) {}
}
```

`AutoConfiguredOpenTelemetrySdk` contains `OpenTelemetrySdk` and we won't be able to fill that in with `OpenTelemetry.noop()`, so we can just pass its fields to the listener.

mateuszrzeszutek: > the config problem (Config vs. ConfigProperties)

About that: I think that currently I prefer using `Config` everywhere - it has all those get-or-default variants that will never throw and return the default value even when an error occurs during parsing. We don't want the agent to break the application because of some random property using wrong syntax after all.

trask: sound good 👍

jack-berg: What if we keep `noop`, but only call `AgentListener` methods if the agent is enabled and the sdk is installed? Is there a use case for having the `AgentListener` when `noop` is installed? If we did that, we could define the `AgentListener` as:
```
interface AgentListener {
  
  default void beforeAgent(Config config, AutoConfiguredOpenTelemetrySdk autoConfiguredSdk) {
    beforeAgent(config);
  }

  @Deprecated
  default void beforeAgent(Config config) {}
  
  default void afterAgent(Config config, AutoConfiguredOpenTelemetrySdk autoConfiguredSdk) {
    afterAgent(config);
  }

  @Deprecated
  default void afterAgent(Config config) {}
}
```
I prefer having access to `AutoConfiguredOpenTelemetrySdk` versus `OpenTelemetry` and `Resource` because: 
- We have access to `OpenTelemetrySdk` instead of `OpenTelemetry`, which gives access to `SdkTracerProvider` and `SdkMeterProvider` instead of just `TracerProvider` and `MeterProvider`. This also means that we could give access to `SdkLogEmitterProvider` which will probably never be accessible via `OpenTelemetry`. 
- We give the user access to holder of all the autoconfigured things, which means that as the set of autoconfigured things changes, the user already access without further API changes.

trask: > Is there a use case for having the AgentListener when noop is installed?

this is a good question. I'm guessing ideally we would still install things like `RuntimeMetricsInstaller` in the `noop` case, but maybe we can live without that?

mateuszrzeszutek: > I prefer having access to `AutoConfiguredOpenTelemetrySdk` versus `OpenTelemetry` and `Resource` because:
> 
> * We have access to `OpenTelemetrySdk` instead of `OpenTelemetry`, which gives access to `SdkTracerProvider` and `SdkMeterProvider` instead of just `TracerProvider` and `MeterProvider`. This also means that we could give access to `SdkLogEmitterProvider` which will probably never be accessible via `OpenTelemetry`.
> * We give the user access to holder of all the autoconfigured things, which means that as the set of autoconfigured things changes, the user already access without further API changes.

Yeah, these are all perfectly valid reasons. Okay, consider me convinced -- I suppose we can live without running a few listeners in case of `noop`; all bytecode manipulations are still done, we're just not installing a couple of metrics.

trask (approved):
thx!

javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/AgentInstaller.java
```diff
@@ -238,9 +245,7 @@ private static void runAfterAgentListeners(
     } else {
       for (AgentListener agentListener : agentListeners) {
         agentListener.afterAgent(config);
-        if (autoConfiguredSdk != null) {
-          agentListener.afterAgent(autoConfiguredSdk);
-        }
+        agentListener.afterAgent(config, autoConfiguredSdk);
```
mateuszrzeszutek, line 248: The `afterAgent()` listener method is also executed from the `DelayedAfterAgentCallback` (under some specific conditions), can you pass the `autoConfiguredSdk` there too?

javaagent-extension-api/src/main/java/io/opentelemetry/javaagent/extension/AgentListener.java
jack-berg: @trask if noop `OpenTelemetry` is installed then the runtime metrics won't have any affect, right?

mateuszrzeszutek (approved):

PR #4833, opened 2021-12-07
Enhancements to overhead tests

breedx-splk:
This adds a few enhancements to the overhead tests:

* Add the test run duration to the output. This is a nice sanity check that helps us to make sure that we're comparing apples and oranges and to get increased confidence in the tests being similar.
* Add `averageMachineCpuTotal` to the output. This can help us to detect if the machine itself is overloaded.
* Add `totalGcPauseNanos` to the output. This can help us to quantify how instrumentation may contribute to GC pauses.
* Consolidate agent run configs. Previously we had 3 configs, each ran with two different agent configs. This was silly, because we ended up running the no-agent case 3 times and the stable twice. This PR just combines them into a single configuration with 3 agents (none, released, snapshot). As a result, the output is also less complicated and the thing runs faster.
* Improve CSV field ordering. Previously, all the metrics for a single agent were grouped together. This made comparing the same metric across multiple agents difficult (it required moving stuff around manually). Fields are grouped by metric, and the order of agents is consistent with the definition in the Config object.

When we're ready for this, I'd like to set the existing CSV files aside and start fresh with the new field groupings and data.


trask (approved):
👍

iNikem (approved):

benchmark-overhead/src/test/java/io/opentelemetry/config/Configs.java
```diff
@@ -13,23 +13,9 @@
   RELEASE(
```
mateuszrzeszutek, line 13: Is there any need to have more than 1 `TestConfig` ever? Maybe we could get rid of this class and just construct the config object in test.

benchmark-overhead/src/test/java/io/opentelemetry/results/AppPerfResults.java
```diff
@@ -139,45 +147,60 @@ Builder peakThreadCount(long peakThreadCount) {
       return this;
     }
 
-    Builder averageNetworkRead(long averageNetworkRead) {
+    Builder averageNetworkRead(long averageNetworkRead){
```
mateuszrzeszutek, line 150: Revert formatting?
```suggestion
    Builder averageNetworkRead(long averageNetworkRead) {
```

benchmark-overhead/src/test/java/io/opentelemetry/results/PrintStreamPersister.java
```diff
@@ -4,66 +4,64 @@
...
+    display(results, "Avg. CPU (user)", res -> String.valueOf(res.averageJvmUserCpu));
+    display(results, "Max. CPU (user)", res -> String.valueOf(res.maxJvmUserCpu));
+    display(results, "Avg. mch tot cpu", res -> String.valueOf(res.averageMachineCpuTotal));
+    display(results, "Startup time (ms)", res -> String.valueOf(res.startupDurationMs));
+    display(results, "Total allocated MB", res -> format(res.getTotalAllocatedMB()));
+    display(results, "Heap (min)", res -> String.valueOf(res.heapUsed.min));
+    display(results, "Heap (max)", res -> String.valueOf(res.heapUsed.max));
+    display(results, "Thread switch rate", res -> String.valueOf(res.maxThreadContextSwitchRate));
+    display(results, "GC time", res -> String.valueOf(res.totalGCTime));
+    display(results, "GC pause time", res -> String.valueOf(NANOSECONDS.toMillis(res.totalGcPauseNanos)));
```
mateuszrzeszutek, line 45: Can we add unit to those metrics that don't have them (and are not totally obvious)?
```suggestion
    display(results, "GC time (ms)", res -> String.valueOf(res.totalGCTime));
    display(results, "GC pause time (ms)", res -> String.valueOf(NANOSECONDS.toMillis(res.totalGcPauseNanos)));
```

breedx-splk: Sure. It's in the main table, but it's probably helpful to add it here so I'll do that. 👍🏻

mateuszrzeszutek (approved):

PR #4838, opened 2021-12-08
add metrics rpc conventions implement

yangtaoran:


yangtaoran:
hi @trask,the EasyCLA check have passed,thank you.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.RequestMetrics;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * guide from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-client
+ */
+@UnstableApi
+public class RpcClientMetrics implements RequestListener {
```
mateuszrzeszutek, line 25:
```suggestion
public final class RpcClientMetrics implements RequestListener {
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.instrumentation.api.annotations.UnstableApi;
+import io.opentelemetry.instrumentation.api.instrumenter.RequestListener;
+import io.opentelemetry.instrumentation.api.instrumenter.RequestMetrics;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * guide from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-client
```
mateuszrzeszutek, line 22:
```suggestion
 * {@link RequestListener} which keeps track of <a
 * href="https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-client">RPC
 * client metrics</a>.
 *
 * <p>To use this class, you may need to add the {@code opentelemetry-api-metrics} artifact to your
 * dependencies.
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+        .setDescription("measures duration of outbound RPC")
+        .setUnit("milliseconds")
+        .ofLongs().build();
+  }
+
+  /**
+   * Returns a {@link RequestMetrics} which can be used to enable recording of {@link
+   * RpcClientMetrics} on an {@link
+   * io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder}
+   * method addRequestMetrics().
```
mateuszrzeszutek, line 49:
```suggestion
   * io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder}.
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * measures duration of outbound RPC.
+   */
+  private final LongHistogram clientDurationHistogram;
+
+  private RpcClientMetrics(Meter meter) {
+    clientDurationHistogram = meter
+        .histogramBuilder("rpc.client.duration")
+        .setDescription("measures duration of outbound RPC")
```
mateuszrzeszutek, line 40:
```suggestion
        .setDescription("The duration of an outbound RPC invocation")
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,82 @@
...
+  @Override
+  public void end(Context context, Attributes endAttributes, long endNanos) {
+    State state = context.get(RPC_CLIENT_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset RPC request metrics.", context);
+    }
+    clientDurationHistogram.record(
+        TimeUnit.MILLISECONDS.convert(
+            endNanos - state.startTimeNanos(), TimeUnit.NANOSECONDS),
```
mateuszrzeszutek, line 71: I think we're using the `to<Unit>()` methods way more often, they're a bit simpler I guess:

```suggestion
        TimeUnit.NANOSECONDS.toMillis(endNanos - state.startTimeNanos()),
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/MetricsView.java
```diff
@@ -0,0 +1,64 @@
...
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.BiConsumer;
+
+/**
+ * filter rpc metrics unnecessary attributes.
+ */
+public class MetricsView {
```
mateuszrzeszutek, line 19: We probably don't want to expose this class as public API:
```suggestion
final class MetricsView {
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/MetricsView.java
```diff
@@ -0,0 +1,64 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.function.BiConsumer;
+
+/**
+ * filter rpc metrics unnecessary attributes.
+ */
```
mateuszrzeszutek, line 18: Can you add a similar comment as in `...http.TemporaryMetricsView`? These "view" classes are still kind of a hack, they're supposed to be replaced by the SDK views at some point.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.instrumentation.api.annotations.UnstableApi;
+import io.opentelemetry.instrumentation.api.instrumenter.RequestListener;
+import io.opentelemetry.instrumentation.api.instrumenter.RequestMetrics;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * guide from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-server
```
mateuszrzeszutek, line 22:
```suggestion
 * {@link RequestListener} which keeps track of <a
 * href="ttps://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-server">RPC
 * server metrics</a>.
 *
 * <p>To use this class, you may need to add the {@code opentelemetry-api-metrics} artifact to your
 * dependencies.
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+
+  /**
+   * measures duration of inbound RPC.
+   */
+  private final LongHistogram serverDurationHistogram;
+
+  private RpcServerMetrics(Meter meter) {
+    serverDurationHistogram = meter
+        .histogramBuilder("rpc.server.duration")
+        .setDescription("measures duration of inbound RPC")
```
mateuszrzeszutek, line 40:
```suggestion
        .setDescription("The duration of an inbound RPC invocation")
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+        .setUnit("milliseconds")
+        .ofLongs().build();
+  }
+
+  /**
+   * Returns a {@link RequestMetrics} which can be used to enable recording of {@link
+   * RpcServerMetrics} on an {@link
+   * io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder}
+   * method addRequestMetrics().
```
mateuszrzeszutek, line 49:
```suggestion
   * io.opentelemetry.instrumentation.api.instrumenter.InstrumenterBuilder}.
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public void end(Context context, Attributes endAttributes, long endNanos) {
+    State state = context.get(RPC_SERVER_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset RPC request metrics.", context);
+    }
+    serverDurationHistogram.record(
+        TimeUnit.MILLISECONDS.convert(
+            endNanos - state.startTimeNanos(), TimeUnit.NANOSECONDS),
```
mateuszrzeszutek, line 71:
```suggestion
        TimeUnit.NANOSECONDS.toMillis(endNanos - state.startTimeNanos()),
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.RequestMetrics;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * guide from https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#rpc-server
+ */
+@UnstableApi
+public class RpcServerMetrics implements RequestListener {
```
mateuszrzeszutek, line 25:
```suggestion
public final class RpcServerMetrics implements RequestListener {
```

mateuszrzeszutek (approved):
LGTM 👍 
Thanks! I've left a couple of comments here, but they're all fairly minor things.

yangtaoran:
Thanks for @mateuszrzeszutek suggestions,and I have resolve them.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+        RPC_CLIENT_REQUEST_METRICS_STATE,
+        new AutoValue_RpcClientMetrics_State(startAttributes, startNanos));
+  }
+
+  @Override
+  public void end(Context context, Attributes endAttributes, long endNanos) {
+    State state = context.get(RPC_CLIENT_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset RPC request metrics.", context);
```
trask, line 70:
```suggestion
      logger.debug(
          "No state present when ending context {}. Cannot record RPC request metrics.", context);
      return;
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Override
+  public void end(Context context, Attributes endAttributes, long endNanos) {
+    State state = context.get(RPC_SERVER_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset RPC request metrics.", context);
+    }
+    serverDurationHistogram.record(
+        TimeUnit.NANOSECONDS.toMillis(endNanos - state.startTimeNanos()),
```
trask, line 73: same comment as in `RpcClientMetrics`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,88 @@
...
+        RPC_SERVER_REQUEST_METRICS_STATE,
+        new AutoValue_RpcServerMetrics_State(startAttributes, startNanos));
+  }
+
+  @Override
+  public void end(Context context, Attributes endAttributes, long endNanos) {
+    State state = context.get(RPC_SERVER_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset RPC request metrics.", context);
```
trask, line 70:
```suggestion
      logger.debug(
          "No state present when ending context {}. Cannot record RPC request metrics.", context);
      return;
```

trask (commented):
thx @yangtaoran!

can you add tests similar to `HttpClientMetricsTest` and `HttpServerMetricsTest`?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/MetricsView.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static Set<AttributeKey> buildOptional() {
+    // the list of optional metrics attributes is from
+    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc.md#attributes
+    Set<AttributeKey> view = new HashSet<>();
+    view.add(SemanticAttributes.NET_PEER_IP);
+    view.add(SemanticAttributes.NET_PEER_NAME);
+    view.add(SemanticAttributes.NET_PEER_PORT);
+    view.add(SemanticAttributes.NET_TRANSPORT);
+    return view;
```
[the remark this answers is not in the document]
jsuereth: According to the linked doc:

```
To avoid high cardinality, implementations should prefer the most stable of net.peer.name or net.peer.ip, depending on expected deployment profile. For many cloud applications, this is likely net.peer.name as names can be recycled even across re-instantiation of a server with a different ip.
```

I.e. you can *fallback* to net.peer.ip if it's you're only option, but you should JUST use net.peer.name if you can.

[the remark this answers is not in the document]
trask: @jsuereth so we should go ahead and capture these same attributes on server metrics (unlike on http server metrics)? I guess the situation is not as bad in RPC-land as in browser-land (except for maybe android/grpc?)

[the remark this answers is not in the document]
jsuereth: We may want to re-evaluate server side using a notion of endpoint vs. net.peer.ip for similar reasons to http.  However the specification is based on what gRPC does.  I'll do some homework here to see if this is still how that's instrumented.

Key thing though, we should be using hostname over IP because hostname should be more stable, especially in the presence of a load balancer.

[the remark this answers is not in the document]
yangtaoran: I understanding is  usering net.peer.name instead of net.peer.ip if the attribute  net.peer.name exist.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+    clientDurationHistogram.record(
+        TimeUnit.NANOSECONDS.toMillis(endNanos - state.startTimeNanos()),
```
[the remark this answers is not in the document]
yangtaoran: Thank you very much for this bug.I had use time division like HttpClientMetrics,and use DoubleHistogram instead of LongHistogram

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
yangtaoran: done.I'd appreciate your input on this.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
yangtaoran: done.like RpcClientMetrics.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
yangtaoran: done.like RpcClientMetrics.

yangtaoran:
> thx @yangtaoran!
> 
> can you add tests similar to `HttpClientMetricsTest` and `HttpServerMetricsTest`?
hi @trask  I have added RpcClientMetricsTest and RpcServerMetricsTest similar to HttpClientMetricsTest and HttpServerMetricsTest

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/MetricsView.java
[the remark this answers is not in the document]
jsuereth: Yeah, the other issue here is around whether we *should* be using `net.peer.*` on RPC server-metrics.   

We translated the specification on instrumentation from OpenCensus gRPC that lead to this as a basis.  HOWEVER, there's a few caveats:

1. In OpenCensus instrumentation + Views were separately specified.  Instrumentation was thorough, views were more conservative.   When this was originally written for OTEL the View spec wasn't known, so that part is unspecified.   In OpenCensus, views are where the high cardinality issues were dealt with by use case, and where you could select styles of latency metrics to export (e.g. if you wanted SLO monitoring).   I think the balance we struck 13 months ago is likely now wrong given the evolution of our APIs/SDKs *unless* we provide out-of-the-box VIew configuration for these metrics in addition to the rich instrumentation attributes.
2. RPC semantic conventions now diverge from HTTP semantic conventions.  Effectively....   "client" metrics use `net.peer.name` and "host" metrics should use `host.name` in HTTP, but RPC both use `net.peer`.

For both of these reasons, we may want to re-evaluate RPC metrics.  Specifically RPC server metrics.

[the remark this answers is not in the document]
jsuereth: Opened https://github.com/open-telemetry/opentelemetry-specification/issues/2214

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetricsTest.java
```diff
@@ -0,0 +1,156 @@
...
+                    SpanContext.create(
+                        "ff01020304050600ff0a0b0c0d0e0f00",
+                        "090a0b0c0d0e0f00",
+                        TraceFlags.getSampled(),
+                        TraceState.getDefault())));
+
+    Context context1 = listener.start(parent, requestAttributes, nanos(100));
+
+    // TODO(anuraaga): Remove await from this file after 1.8.0 hopefully fixes
+    // https://github.com/open-telemetry/opentelemetry-java/issues/3725
```
mateuszrzeszutek, line 73: @anuraaga I tried removing awaits from `HttpClientMetricsTest`/`HttpServerMetricsTest` and it failed for me; looks like this issue is still not solved.

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetricsTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    Attributes responseAttributes1 =
+        Attributes.builder()
+            .put(SemanticAttributes.NET_PEER_NAME, "example.com")
+            .put(SemanticAttributes.NET_PEER_IP, "127.0.0.1")
+            .put(SemanticAttributes.NET_PEER_PORT, 8080)
+            .put(SemanticAttributes.NET_TRANSPORT, "ip_tcp")
+            .build();
+
+    Attributes responseAttributes2 =
```
mateuszrzeszutek, line 53: WDYT about splitting this into 2 test cases? One for host name, one for IP.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+        RPC_CLIENT_REQUEST_METRICS_STATE,
+        new AutoValue_RpcClientMetrics_State(startAttributes, startNanos));
+  }
+
+  @Override
+  public void end(Context context, Attributes endAttributes, long endNanos) {
+    State state = context.get(RPC_CLIENT_REQUEST_METRICS_STATE);
+    if (state == null) {
+      logger.debug(
+          "No state present when ending context {}. Cannot reset RPC request metrics.", context);
```
trask, line 74:
```suggestion
          "No state present when ending context {}. Cannot record RPC request metrics.", context);
```

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcServerMetrics.java
```diff
@@ -0,0 +1,90 @@
...
```
trask, line 74:
```suggestion
          "No state present when ending context {}. Cannot record RPC request metrics.", context);
```

trask (approved):
this looks great, thx!

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/RpcClientMetricsTest.java
yangtaoran: Yeah,according to @jsuereth suggestion,we should pefer to net.peer.name, we can fallback to net.peer.ip if net.peer.name is  unavailable.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/rpc/MetricsView.java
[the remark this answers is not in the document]
yangtaoran: So I can modify the RPC conventions implement like this:
Client metrics use net.peer.name (or if unavailable net.peer.ip),net.peer.port,net.transport.
Server metrics use net.host.name (or if unavailable nethost.ip),net.transport,exclude net.host.port.
Do you think this is ok?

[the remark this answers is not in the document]
jsuereth: That's effectively what I'd do.

For the purpose of this PR, I think if you just don't fill out any `net.*` attributes on the server side you're fine (because of Resource attributes).   As we discussed in the Java-SiG yesterday, it'd be good to mark this instrumentation as "experimental" in some fashion so you can try a few options out and we can see how they work in practice.

trask:
Thanks again @yangtaoran!

PR #4842, opened 2021-12-08
Enable Armeria server instrumentation notFound() test

mateuszrzeszutek:
I noticed that there was a bit of code in the `OpenTelemetryService` class that seems to contradict our general approach to how we treat 404 request routes/span names.


iNikem (approved):

laurit (approved):

trask:
👍

Issue #4840, opened 2021-12-08
Trace information from Istio ingress gateway not propagated


danhngo-lx:
**Describe the bug**
I have Istio ingress gateway to add b3 headers in requests sent to my Java service with automatic instrumentation, but trace id from the ingress gateway not propagated. The service is creating new trace id when creating request to another service. The value of trace_id in the logs is empty too.

**Steps to reproduce**
I enabled tracing for Istio and could confirm it is adding b3 headers (x-request-id, x-b3-traceid, x-b3-spanid, x-b3-parentspanid, x-b3-sampeld) when forwarding the request from the outside to my service.
Here's the related env vars:
```
        - name: APISERVER_OPTS
          value: -javaagent:/app/opentelemetry-javaagent.jar  -Dotel.javaagent.debug=true
        - name: OTEL_PROPAGATORS
          value: b3multi
        - name: OTEL_TRACES_EXPORTER
          value: zipkin
        - name: OTEL_EXPORTER_ZIPKIN_ENDPOINT
          value: http://otel-collector.tracing.svc:9411
        - name: OTEL_RESOURCE_ATTRIBUTES
          value: service.name=api-server,job=core/api-server
```

In my service I have this line to record the request when forwarding it to another service:
```
    logger.debug("Forward http request to {} ", request);
```
But the traceID is empty in the logs:
```
2021-12-08 08:45:21,965 [apiserver-akka.actor.default-dispatcher-26] DEBUG c.l.a.common.http.CipherRouter traceID= - Forward http request to HttpRequest(HttpMethod(GET),http://example.com/v1/profiles?userId=123&profileLevel=USER&pageNumber=1&pageSize=1,List(Authorization),HttpEntity.Strict(none/none,0 bytes total),HttpProtocol(HTTP/1.1))
```

**What did you expect to see?**
TraceID from b3 headers propagated correctly to another downstream service. Also trace_id in logs of the service not empty

**What did you see instead?**
The trace id is not propagated, but opentelemetry-javaagent is generating a new trace id for downstream service.

**What version are you using?**
1.9.1

**Environment**
Compiler: (e.g., "OpenJDK 11")
Docker image: openjdk:11-jdk-slim

**Additional context**

danhngo-lx:
This is working fine for my grpc services: the request that goes through the ingress gateway and then forwarded to a gprc service has the same trace id.
I wonder if this has anything to do with the akka actor library in my service.

[...]

PR #4847, opened 2021-12-08
Verify case-insensitive TextMapGetters

trask:
One possible explanation for #4840 (if CI fails...)


trask:
oh well, seems like CI passed (except for a sporadic failure). I think it's still a good test(?)

iNikem (commented):
I am weakly against it. I feel that mixing several test-cases in one test in this way muddies waters...

anuraaga (approved):

iNikem (approved):

mateuszrzeszutek (approved):

PR #4852, opened 2021-12-09
Refactor `ServerSpanNaming` (in preparation for `http.route`)

mateuszrzeszutek:
First part of #442 

I've refactored `ServerSpanNaming` (mostly in the servlet instrumentation) so that it's called every time we include the route in the span name. In next PRs I want to remove/deprecate the `HttpServerAttributesExtractor#route()` method and only use `ServerSpanNaming` (which will be renamed to `HttpServerRoute` probably) to set the route as span name & attribute.


laurit:
Have you considered that what we use as server span name is not always a route. For example wicket instrumentation name sets server span name to an arbitrary value whose only purpose is to identify the visited page.

mateuszrzeszutek:
> For example wicket instrumentation sets server span name to an arbitrary value whose only purpose is to identify the visited page.

Oh, didn't know that. Most of the web libraries that we instrument use route to update the name I think.
How many of these "special" instrumentations that do not update the span name with route do we have?

laurit:
> Oh, didn't know that. Most of the web libraries that we instrument use route to update the name I think. How many of these "special" instrumentations that do not update the span name with route do we have?

A few: grails, jaxws, spring-ws, tapestry, vaadin, wicket. For non-rest frameworks we need to invent some kind of name that describes the operation. This name isn't necessarily even deduced from url. For example in case of jax-ws it is a soap xml document that is posted that determines what is invoked.

mateuszrzeszutek:
> A few: grails, jaxws, spring-ws, tapestry, vaadin, wicket.

Thanks, I'll take a look at these. 
The spec definition of `http.route` is pretty loose (`The matched route (path template)`), so I wonder if it's fine to leave at least some of these as they are (template file names?).

Anyway, this PR is still valid even with these reservations -- in the worst case I'll just have to think of how to split the server name from the route so that the code still makes any sense 😅

instrumentation/jetty/jetty-11.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jetty/v11_0/Jetty11Singletons.java
```diff
@@ -26,10 +23,7 @@
       INSTRUMENTER =
           ServletInstrumenterBuilder.<HttpServletRequest, HttpServletResponse>create()
               .addContextCustomizer(
-                  (context, request, attributes) -> {
-                    context = ServerSpanNaming.init(context, CONTAINER);
-                    return new AppServerBridge.Builder().init(context);
-                  })
+                  (context, request, attributes) -> new AppServerBridge.Builder().init(context))
```
trask, line 26: should `ServerSpanNaming.get()` context customizer be added here? I notice it's missing in some cases and included in others, but not clear to me why

mateuszrzeszutek: It actually is added, via the `ServletInstrumenterBuilder` class.
All server instrumentations should add `ServerSpanNaming.get()`

instrumentation/servlet/servlet-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/v3_0/Servlet3Advice.java
```diff
@@ -39,52 +38,41 @@ public static void onEnter(
...
-    helper().setAsyncListenerResponse(httpServletRequest, (HttpServletResponse) response);
+    // Update context with info from current request to ensure that server span gets the best
+    // possible name.
+    // In case server span was created by app server instrumentations calling updateContext
+    // returns a new context that contains servlet context path that is used in other
+    // instrumentations for naming server span.
+    MappingResolver mappingResolver = Servlet3Singletons.getMappingResolver(servletOrFilter);
+    boolean servlet = servletOrFilter instanceof Servlet;
+    contextToUpdate =
+        helper().updateContext(contextToUpdate, httpServletRequest, mappingResolver, servlet);
```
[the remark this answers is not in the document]
mateuszrzeszutek: > the previous behavior was nice of setting everything up correctly in `startSpan` (e.g. correct span name and correct `ServerSpanNaming.Source`)

Honestly I found setting the `Source` a bit confusing: what if your span name was `HTTP GET` with `FILTER` as a source? If you had another filter trying to update the name to `/abc` it wouldn't, because it was a shorter name - which I'm not so sure is correct; setting the source seems to be closely connected to the route, not necessarily the span name (even if they're the same in most cases). (Although I'm not sure how "real" this example is 😅 )

Also, the only 2 instrumentations I found that have access to the route in the beginning are armeria and servlets; and servlet span name will usually be updated by something like spring or jax-rs. Which implies that for most cases, the route should be extracted on `end()` (and span name updated somewhere in the meantime).

trask (approved):

PR #4853, opened 2021-12-09
Convert play instrumentation to instrumenter api

laurit:
Part of #2713


instrumentation/play/play-2.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/play/v2_4/Play24Singletons.java
```diff
@@ -0,0 +1,53 @@
...
+
+    Span.fromContext(context).updateName(route);
+    // set the span name on the upstream akka/netty span
+    Span serverSpan = ServerSpan.fromContextOrNull(context);
+    if (serverSpan != null) {
+      serverSpan.updateName(route);
+    }
+  }
+
+  private static String getRoute(Request<?> request) {
```
laurit, line 42: I couldn't find any tests that would expect this method to not return null

instrumentation/play/play-2.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/play/v2_4/Play24Singletons.java
```diff
@@ -0,0 +1,53 @@
...
+    String route = getRoute(request);
+    if (route == null) {
+      return;
+    }
+
+    Span.fromContext(context).updateName(route);
+    // set the span name on the upstream akka/netty span
+    Span serverSpan = ServerSpan.fromContextOrNull(context);
+    if (serverSpan != null) {
+      serverSpan.updateName(route);
```
mateuszrzeszutek, line 38: WDYT about using `ServerSpanNaming` here? Netty 4.x instrumentations actually initialize it, and I think ratpack instrumentation uses it to update server span names -- I think it should probably work here too.

mateuszrzeszutek (approved):

instrumentation/play/play-2.4/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/play/v2_4/Play24Singletons.java
laurit: When I realized that nothing tests this I decided to modify it as little as possible.

PR #4866, opened 2021-12-10
Update to OTel 1.10

anuraaga:


dependencyManagement/build.gradle.kts
```diff
@@ -11,7 +11,8 @@ data class DependencySet(val group: String, val version: String, val modules: Li
 val dependencyVersions = hashMapOf<String, String>()
 rootProject.extra["versions"] = dependencyVersions
 
-val otelVersion = "1.9.1"
+val otelVersion = "1.10.0-rc.1"
+val otelAlphaVersion = "1.10.0-alpha-rc.1"
```
anuraaga, line 15: @jkwatson Oops

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/InstrumenterBuilder.java
```diff
@@ -67,8 +66,7 @@
       String instrumentationVersion,
       SpanNameExtractor<? super REQUEST> spanNameExtractor) {
     this.openTelemetry = openTelemetry;
-    // TODO(anuraaga): Retrieve from openTelemetry when not alpha anymore.
-    this.meter = GlobalMeterProvider.get().get(instrumentationName);
+    this.meter = openTelemetry.getMeterProvider().get(instrumentationName);
```
trask, line 69: 🎉

dependencyManagement/build.gradle.kts
trask: 🤷‍♂️

instrumentation/runtime-metrics/library/src/main/java/io/opentelemetry/instrumentation/runtimemetrics/GarbageCollector.java
```diff
@@ -5,17 +5,17 @@
...
-import io.opentelemetry.api.metrics.GlobalMeterProvider;
 import io.opentelemetry.api.metrics.Meter;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
- * Registers observers that generate metrics about JVM garbage collectors.
+ * Registers recordrs that generate metrics about JVM garbage collectors.
```
trask, line 18:
```suggestion
 * Registers observers that generate metrics about JVM garbage collectors.
```

trask (approved):

instrumentation/opentelemetry-api/opentelemetry-api-1.10/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/v1_10/OpenTelemetryInstrumentation.java
```diff
@@ -0,0 +1,39 @@
...
+        none(), OpenTelemetryInstrumentation.class.getName() + "$InitAdvice");
+  }
+
+  @SuppressWarnings("unused")
+  public static class InitAdvice {
+    @Advice.OnMethodEnter
+    public static void init() {
+      // the sole purpose of this advice is to ensure that ApplicationOpenTelemetry110 is recognized
+      // as helper class and injected into class loader
+      ApplicationOpenTelemetry110.class.getName();
```
anuraaga, line 36: By the way I wonder if we can provide a official API for this pattern, it seems to be a combination of muzzle classpath check + helper class injection @mateuszrzeszutek @laurit

mateuszrzeszutek: Yeah, we're not doing anything with `additionalHelperClasses` in muzzle codegen - this should be fixed, I'll create an issue for that.
Still, even if we did that, that still wouldn't solve the problem: helpers aren't injected when you have no type instrumentations, you must have at least one.

anuraaga: Yeah I guess like this one

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/aws-sdk/aws-sdk-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awssdk/v2_2/AwsSdkInstrumentationModule.java#L55

I guess both aren't great, but I think that pattern of no-op transform is still clearer than no-op advice.

javaagent/src/test/groovy/io/opentelemetry/javaagent/classloading/ShadowPackageRenamingTest.groovy
```diff
@@ -5,7 +5,7 @@
```
anuraaga, line 5: @trask @mateuszrzeszutek Can you confirm my changes to this file? It was needed since no more Guava in agent.

javaagent/src/test/groovy/io/opentelemetry/javaagent/classloading/ShadowPackageRenamingTest.groovy
```diff
@@ -16,9 +16,7 @@ class ShadowPackageRenamingTest extends Specification {
   static final String[] AGENT_PACKAGE_PREFIXES = [
```
anuraaga, line 16: I'm also wondering what this list is for - it doesn't seem to be updated (was missing jackson, has disruptor which we don't use)

trask: > I'm also wondering what this list is for

ya, me too, see the comment below 😄

```
    // TODO (trask) agentClasspath.getAllClasses() is empty
    //  so this part of the test doesn't verify what it thinks it is verifying
```

trask: @laurit any idea if there's anything worth salvaging from the `"agent jar contains no bootstrap classes"` test, or should we just delete it (and `AGENT_PACKAGE_PREFIXES`)

javaagent/src/test/groovy/io/opentelemetry/javaagent/classloading/ShadowPackageRenamingTest.groovy
```diff
@@ -39,12 +37,12 @@ class ShadowPackageRenamingTest extends Specification {
...
-    URL userGuava =
-      MapMaker.getProtectionDomain().getCodeSource().getLocation()
-    URL agentGuavaDep =
+    URL userJackson =
+      JSON.getProtectionDomain().getCodeSource().getLocation()
+    URL agentJacksonep =
       clazz
         .getClassLoader()
-        .loadClass("com.google.common.collect.MapMaker")
+        .loadClass("com.fasterxml.jackson.jr.ob.JSON")
```
trask, line 45: I think using `io.opentelemetry.sdk.OpenTelemetrySdk` would make this more clear (my first thought was, we have jackson in the agent? until I found that this is being pulled in via the jaeger exporter)

trask (approved):
(re-approving 👍)

mateuszrzeszutek (approved):

PR #4867, opened 2021-12-10
Fix reactor-netty memory/connection leak

### LINKED ISSUE #4862: Memory & connection leak in reactor-netty instrumentation
**Describe the bug**
When using `WebClient` in Spring WebFlux, a `SimpleDequePool` is created on every HTTP request and never gets disposed.

**Steps to reproduce**
1. Included `spring-boot-starter-webflux` as a dependency
2. Create a `org.springframework.web.reactive.function.client.WebClient`
3. Use this client to send some HTTP requests

1. Sending requests with one WebClient should create only one connection pool and reuse connection if possible.

**What did you see instead?**
1. Lots of `SimpleDequePool` live in heap and lots of TCP connections are established. None of these will get disposed after GC.
2. Server gets OOM or reaches limit of num of FD eventually.

`1.9.1`

**Environment**
Compiler: openjdk 1.8.0_312
OS: macOS 12.0.1

**Additional context**
- `PooledConnectionProvider` use `PooledConnectionProvider.PoolKey` as key for its internal channelPools (ConcurrentHashMap).
- Hashcode of `PooledConnectionProvider.PoolKey` is computed with `pipelineKey` and other fields.
- `pipelineKey` is assigned with `TransportConfig.channelHash()`
- `channelHash()` is computed with `observer` and other fields.
- `observer` is a different instance of `HttpResponseReceiverInstrumenter.EndOperation` on each HTTP request and it may cause `PooledConnectionProvider` to create a new pool every time

mateuszrzeszutek:
Fixes #4862

### DIFF
--- a/instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/reactornetty/v1_0/AbstractReactorNettyHttpClientTest.groovy
+++ b/instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/reactornetty/v1_0/AbstractReactorNettyHttpClientTest.groovy
@@ -191,6 +191,32 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
     }
   }
 
+  def "should not leak connections"() {
+    given:
+    def uniqueChannelHashes = new HashSet<>()
+    def httpClient = createHttpClient()
+      .doOnConnect({ uniqueChannelHashes.add(it.channelHash())})
+    def uri = "http://localhost:${server.httpPort()}/success"
+
+    def count = 100
+
+    when:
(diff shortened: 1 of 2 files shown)


laurit (approved):

instrumentation/reactor-netty/reactor-netty-1.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/reactornetty/v1_0/AbstractReactorNettyHttpClientTest.groovy
```diff
@@ -191,6 +191,32 @@ abstract class AbstractReactorNettyHttpClientTest extends HttpClientTest<HttpCli
     }
   }
 
+  def "should not leak connections"() {
```
trask, line 194: 👍

trask (approved):

PR #4870, opened 2021-12-10
Don't use HttpStatusConverter outside of instrumentation-api

mateuszrzeszutek:


instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentation.java
```diff
@@ -137,7 +132,7 @@ public static void methodExit(
                 .end(
                     httpUrlState.context,
                     connection,
-                    responseCode > 0 ? responseCode : null,
+                    responseCode > 0 ? responseCode : httpUrlState.statusCode,
```
trask, line 135: do we know that `httpUrlState.statusCode` will be initialized at this point, or should we change it from primitive int to nullable Integer?

iNikem (approved):

instrumentation/http-url-connection/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/httpurlconnection/HttpUrlConnectionInstrumentation.java
mateuszrzeszutek: Using `Integer` here causes NPE - I think Java tries to unbox it to evaluate the ternary operator, and before boxing it again. Instead, I added an explicit initialization of the `statusCode` field to 0 (and a comment).

trask (approved):

PR #4877, opened 2021-12-12
Fix Intellij (re-)import

trask:
Partial revert of #4863

Intellij (re-)import is currently failing with spotless 6.0.3 and 6.0.4

Would be good to get confirmation that this is reproducible by someone else and not just a local issue.

image

### DIFF
--- a/conventions/build.gradle.kts
+++ b/conventions/build.gradle.kts
@@ -1,7 +1,8 @@
 plugins {
   `kotlin-dsl`
   // When updating, update below in dependencies too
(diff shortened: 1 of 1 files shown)


anuraaga (approved):
I'm getting an even worse problem from the commandline too

https://github.com/diffplug/spotless/issues/1034

anuraaga:
Oops - forgot this is reverting to the version I had already been having problems with. It's the spotpocalypse!

PR #4887, opened 2021-12-13
Use Gradle GitHub Action

iNikem:


anuraaga:
Don't think we need burrunan and maybe actions/cache, actions now

iNikem:
> Don't think we need burrunan and maybe actions/cache, actions now

Yes, I did not want to lump all of that into single PR for some reason. But I can rework caching also :)

anuraaga (approved):

mateuszrzeszutek (approved):

PR #4893, opened 2021-12-13
Add startup check to all rabbitmq container usages

laurit:
Continuation of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4813
Previous pr added startup check to some rabbitmq usages, if I interpret the flaky test results correct this seems to have worked. This pr adds the same startup check to other rabbitmq container usages in the hope that this will get rid of flakiness like https://ge.opentelemetry.io/s/qu73cs57tjqr6/tests/:instrumentation:spring:spring-integration-4.1:javaagent:test/SpringCloudStreamRabbitTest/should%20propagate%20context%20through%20RabbitMQ?top-execution=1


trask (approved):
nice

PR #4901, opened 2021-12-14
Fix NPE in RocketMQ instrumentation

### LINKED ISSUE #4899: RocketMQ send message async cause the NullPointerException
**Describe the bug**
RocketMQ send message async will cause the NullPointerException.

**The exception stacktrace is that:**
exception_stack

When we use RocketMQ send the message async, the SendResult will be NULL.  The part of RocketMQ Client code is:
ope

After send message, RocketMQ Client will execute the method 'executeSendMessageHookAfter' to do the method 'sendMessageAfter' of all 'SendMessageHook'.  The SendMessageHook of Opentelemetry will get the messageId from SendResult (but it's NULL), so the NullPointerException occurs:
null

mateuszrzeszutek:
Fixes #4899

I couldn't really reproduce that issue (we're already testing async calls), perhaps it only affects some versions?
Anyway, I've added several nullchecks to all places that seemed suspicious.

### DIFF
--- a/instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetry/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
+++ b/instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetry/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
@@ -85,8 +85,10 @@ abstract class AbstractRocketMqClientTest extends InstrumentationSpecification {
             "$SemanticAttributes.MESSAGING_SYSTEM" "rocketmq"
             "$SemanticAttributes.MESSAGING_DESTINATION" sharedTopic
             "$SemanticAttributes.MESSAGING_DESTINATION_KIND" "topic"
+            "$SemanticAttributes.MESSAGING_MESSAGE_ID" String
             "messaging.rocketmq.tags" "TagA"
             "messaging.rocketmq.broker_address" String
+            "messaging.rocketmq.send_result" "SEND_OK"
           }
         }
         span(1) {
--- a/instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetry/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
+++ b/instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetry/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
@@ -85,10 +85,8 @@ abstract class AbstractRocketMqClientTest extends InstrumentationSpecification {
             "$SemanticAttributes.MESSAGING_SYSTEM" "rocketmq"
             "$SemanticAttributes.MESSAGING_DESTINATION" sharedTopic
             "$SemanticAttributes.MESSAGING_DESTINATION_KIND" "topic"
-            "$SemanticAttributes.MESSAGING_MESSAGE_ID" String
             "messaging.rocketmq.tags" "TagA"
             "messaging.rocketmq.broker_address" String
-            "messaging.rocketmq.send_result" "SEND_OK"
           }
         }
         span(1) {
--- a/instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RockerMqConsumerAttributeExtractor.java
+++ b/instrumentation/rocketmq-client-4.8/library/src/main/java/io/opentelemetry/instrumentation/rocketmq/RockerMqConsumerAttributeExtractor.java
@@ -61,9 +61,11 @@ class RockerMqConsumerAttributeExtractor extends MessagingAttributesExtractor<Me
     return null;
   }
 
+  @Nullable
   @Override
   protected Long messagePayloadSize(MessageExt request) {
-    return (long) request.getBody().length;
+    byte[] body = request.getBody();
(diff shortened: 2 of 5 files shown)


instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetry/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
```diff
@@ -85,10 +85,8 @@ abstract class AbstractRocketMqClientTest extends InstrumentationSpecification {
             "$SemanticAttributes.MESSAGING_SYSTEM" "rocketmq"
             "$SemanticAttributes.MESSAGING_DESTINATION" sharedTopic
             "$SemanticAttributes.MESSAGING_DESTINATION_KIND" "topic"
-            "$SemanticAttributes.MESSAGING_MESSAGE_ID" String
             "messaging.rocketmq.tags" "TagA"
             "messaging.rocketmq.broker_address" String
-            "messaging.rocketmq.send_result" "SEND_OK"
```
trask, line 91: does this mean that you were able to repro the (or at least "a") problem?

trask (approved):

instrumentation/rocketmq-client-4.8/testing/src/main/groovy/io/opentelemetry/instrumentation/rocketmq/AbstractRocketMqClientTest.groovy
mateuszrzeszutek: That part I seriously don't understand - those assertions have clearly passed before, but when I added these null checks they suddenly started failing with missing attributes.
Maybe there are some internal try-catch retries at work there? No idea honestly.

laurit: Had a quick peek, seems to be caused by the NPE fix in `messageId(...)`. The unexpected thing here is that `TracingSendMessageHookImpl.sendMessageAfter` is called twice. Once before send result is set and second time after. Before this change the code flow was so that it got NPE at the first call when extracting message id and succeeded on the second call. Maybe we should change the condition at the start of `TracingSendMessageHookImpl.sendMessageAfter` to `if (context == null || (context.getSendResult() == null && context.getException() == null)) {` also should pass the value of `context.getException()` to `instrumenter.end(...)`.

mateuszrzeszutek: > The unexpected thing here is that `TracingSendMessageHookImpl.sendMessageAfter` is called twice.

Hm, is it always called twice? It looks like it's just called in a retry loop: https://github.com/apache/rocketmq/blob/master/client/src/main/java/org/apache/rocketmq/client/impl/producer/DefaultMQProducerImpl.java#L887

> also should pass the value of `context.getException()` to `instrumenter.end(...)`.

Yeah that's a great idea 👍

laurit: I have no idea whether it is always called twice. Maybe only for async requests? Idk. For `test rocketmq produce callback` first call is from
```
	at io.opentelemetry.javaagent.shaded.instrumentation.rocketmq.TracingSendMessageHookImpl.sendMessageAfter(TracingSendMessageHookImpl.java:47)
	at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.executeSendMessageHookAfter(DefaultMQProducerImpl.java:962)
	at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.sendKernelImpl(DefaultMQProducerImpl.java:871)
	at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.sendDefaultImpl(DefaultMQProducerImpl.java:584)
	at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.access$300(DefaultMQProducerImpl.java:97)
	at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl$4.run(DefaultMQProducerImpl.java:511)
```

Second call is from
```
	at io.opentelemetry.javaagent.shaded.instrumentation.rocketmq.TracingSendMessageHookImpl.sendMessageAfter(TracingSendMessageHookImpl.java:47)
	at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.executeSendMessageHookAfter(DefaultMQProducerImpl.java:962)
	at org.apache.rocketmq.client.impl.MQClientAPIImpl$1.operationComplete(MQClientAPIImpl.java:549)
	at org.apache.rocketmq.remoting.netty.ResponseFuture.executeInvokeCallback(ResponseFuture.java:54)
	at org.apache.rocketmq.remoting.netty.NettyRemotingAbstract$2.run(NettyRemotingAbstract.java:319)
```

mateuszrzeszutek: Yeah it looks like it's only happening in async calls. I'll try using the modified condition you posted.

laurit (approved):

anuraaga (approved):

PR #4907, opened 2021-12-15
Split out log4j appender

### LINKED ISSUE #4780: Split out log4j appender from log4j autoconfigure module
Follow up from https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4779#discussion_r760776336

trask:
Resolves #4780


anuraaga (approved):

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
```diff
@@ -0,0 +1,13 @@
+plugins {
+  id("otel.library-instrumentation")
+}
+
+dependencies {
+  compileOnly("io.opentelemetry:opentelemetry-sdk-logs")
+
+  library("org.apache.logging.log4j:log4j-core:2.13.2")
```
iNikem, line 8: As nobody in its right mind is going to use this version of log4j, should we just start with `2.16`? :)

iNikem (approved):

instrumentation/log4j/log4j-2.13.2/library-autoconfigure/build.gradle.kts
```diff
@@ -3,16 +3,11 @@ plugins {
 }
 
 dependencies {
```
mateuszrzeszutek, line 5: Should this module include `implementation(project("...:log4j-2.13.2:library"))`?

instrumentation/log4j/log4j-2.13.2/library/build.gradle.kts
trask: ya, I think this is the only instrumentation that targets a "patch" version too, will do this in follow-up 👍

instrumentation/log4j/log4j-2.13.2/library-autoconfigure/build.gradle.kts
trask: this case is a bit odd. the `autoconfigure` instrumentation doesn't enable the appender (`library`) instrumentation. another option is to rename `library-autoconfigure` to `library-context-data`.... I'll open an issue to discuss and track

PR #4910, opened 2021-12-15
Wait for request to return result in http client async concurrency test

laurit:
Currently we run all 50 request concurrently. Waiting for the request to complete limits the number of concurrent requests to 4 (thread pool size) as it is also in the synchronous http client concurrency test.
Hopefully this will reduce test flakiness. There are some runs where this test fails because there is a duplicate server span and in the log there is the following warning where `test-request-id` in warning matches the problematic request.
```
21:20:30.765 [armeria-common-worker-epoll-2-1] WARN  i.o.t.i.a.s.logging.LoggingService - [sreqId=179c4c01, chanId=92a0d047, raddr=127.0.0.1:43778, laddr=127.0.0.1:44289][h1c://fv-az201-310/success#GET] Request: {startTime=2021-12-13T21:20:30.703Z(1639430430703000), length=0B, duration=19656µs(19656028ns), cause=io.opentelemetry.testing.internal.io.netty.channel.unix.Errors$NativeIoException: writevAddresses(..) failed: Broken pipe, scheme=none+h1c, name=GET, headers=[:method=GET, :path=/success, :scheme=http, content-type=*/*, accept=text/plain, test-request-id=25, traceparent=00-fc1d841e9cff07d68398caab2be695d0-1756394b5a1f61be-01, user-agent=Apache CXF 3.1.0, cache-control=no-cache, pragma=no-cache, host=localhost:44289]}
21:20:30.826 [armeria-common-worker-epoll-2-1] WARN  i.o.t.i.a.s.logging.LoggingService - [sreqId=179c4c01, chanId=92a0d047, raddr=127.0.0.1:43778, laddr=127.0.0.1:44289][h1c://fv-az201-310/success#GET] Response: {startTime=2021-12-13T21:20:30.721Z(1639430430721000), length=6B, duration=23115µs(23115950ns), totalDuration=41349µs(41349069ns), cause=io.opentelemetry.testing.internal.io.netty.channel.unix.Errors$NativeIoException: writevAddresses(..) failed: Broken pipe, headers=[:status=200, test-request-id=25, content-length=6]}
io.opentelemetry.testing.internal.io.netty.channel.unix.Errors$NativeIoException: writevAddresses(..) failed: Broken pipe
```

### DIFF
--- a/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
+++ b/testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpClientTest.java
@@ -754,18 +754,24 @@ public abstract class AbstractHttpClientTest<REQUEST> {
                       throw new AssertionError(e);
                     }
                     try {
-                      testing.runWithSpan(
-                          "Parent span " + index,
-                          () -> {
-                            Span.current().setAttribute("test.request.id", index);
-                            doRequestWithCallback(
-                                method,
-                                uri,
-                                Collections.singletonMap("test-request-id", String.valueOf(index)),
-                                () -> testing.runWithSpan("child", () -> {}));
-                          });
-                    } catch (Exception e) {
-                      throw new AssertionError(e);
+                      RequestResult result =
+                          testing.runWithSpan(
+                              "Parent span " + index,
+                              () -> {
+                                Span.current().setAttribute("test.request.id", index);
+                                return doRequestWithCallback(
+                                    method,
+                                    uri,
+                                    Collections.singletonMap(
+                                        "test-request-id", String.valueOf(index)),
+                                    () -> testing.runWithSpan("child", () -> {}));
+                              });
+                      assertThat(result.get()).isEqualTo(200);
+                    } catch (Throwable throwable) {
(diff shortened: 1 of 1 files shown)


iNikem:
One of the use-cases that our concurrent tests verify is http pipelining, when several requests are sent before any responses are received. It seems that this change will remove this verification.

laurit:
> One of the use-cases that our concurrent tests verify is http pipelining, when several requests are sent before any responses are received. It seems that this change will remove this verification.

Requests are still concurrently sent from 4 threads. This behaviour is in line with `highConcurrency` test. This test (`highConcurrencyWithCallback`) is just a copy of `highConcurrency` that uses `doRequestWithCallback` instead of `doRequest`. If it test http pipelining then this is accidental. When I added it I only wished to exercise `doRequestWithCallback` the same way as the existing `highConcurrency` test used `doRequest`.

trask (approved):
> One of the use-cases that our concurrent tests verify is http pipelining, when several requests are sent before any responses are received. It seems that this change will remove this verification.

@iNikem isn't `highConcurrencyOnSingleConnection()` for verifying http pipelining?

iNikem:
Indeed, I stand corrected

iNikem (approved):

Issue #4885, opened 2021-12-13
Muzzle should collect references from `additionalHelperClasses` during codegen


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Currently we don't collect references to classes defined in the `InstrumentationModule#getAdditionalHelperClasses()` method (nor do we collect references made by these classes' contents). 

**Additional context**
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4866#discussion_r767483772

PR #4911, opened 2021-12-15
Collect references from additionalHelperClasses during muzzle codegen

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4885

### DIFF
--- a/instrumentation/internal/internal-lambda/javaagent/build.gradle.kts
+++ b/instrumentation/internal/internal-lambda/javaagent/build.gradle.kts
@@ -8,3 +8,8 @@ dependencies {
 
   testImplementation(project(":javaagent-bootstrap"))
 }
+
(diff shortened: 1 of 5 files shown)


iNikem (approved):

mateuszrzeszutek:
Okay, looks like this change is causing the Java 8 tests failures on lambda transformation:
```
java.lang.UnsupportedClassVersionError: io/opentelemetry/javaagent/instrumentation/internal/lambda/Java9LambdaTransformer has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```
I don't really have any good idea how to fix this issue in a non-horrible way, so I'll close this PR. I suppose that this feature is kinda minor and we can live without that.

PR #4913, opened 2021-12-15
Fix npe in apache http client 4.0 instrumentation

laurit:
NPE is logged when running some apache http client 4.0 tests. For example `trace request with callback and parent` in  `ApacheClientHostAbsoluteUriRequest`.


trask (approved):
how did this ever work?!??

laurit:
> how did this ever work?!??

the method with broken instrumentation chains to another `execute` method that has working instrumentation

anuraaga (approved):

PR #4915, opened 2021-12-16
Update log4j instrumentation docs

trask:
Doc updates for #4914. This PR is split out if we want to merge during/after 1.10 release.

Or should I just include the doc updates in #4914 and tags can be used now for viewing stable release docs?

PR #4917, opened 2021-12-16
Add a logging appender API

### LINKED ISSUE #4790: Create log appender API
The agent can't take a dependency on the log SDK, but there is no log API. There should be an log appender API that log instrumentation can call without needing a dependency on the log sdk. 

As discussed in the SIG and [here](https://github.com/open-telemetry/opentelemetry-java/pull/3880), the preference is to add an appender API in `opentelemetry-java-instrumentation`, where the instrumentation that will consume it lives.

trask:
Resolves #4790

Just a shim on top of the Log SDK
* LogEmitterProvider
* LogEmitterBuilder
* LogEmitter
* LogBuilder

with an additional no-op implementation, and support for an (optional) global `LogEmitterProvider`.


trask:
A couple of thoughts:
* maybe this should just be part of the `instrumentation-api` artifact?
* maybe instead of calling it "appender api", call it "log emitter api"? (with similarly named package, e.g. `io.opentelemetry.instrumentation.api.logemitter`)

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/GlobalLogEmitterProvider.java
```diff
@@ -0,0 +1,63 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.appender.api;
+
+import io.opentelemetry.api.internal.GuardedBy;
+import javax.annotation.Nullable;
+
+public final class GlobalLogEmitterProvider {
```
jack-berg, line 11: Needs a private constructor.

instrumentation-appender-api/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
+plugins {
+  id("otel.java-conventions")
+  id("otel.animalsniffer-conventions")
+  id("otel.jacoco-conventions")
+  id("otel.japicmp-conventions")
+  id("otel.publish-conventions")
```
jack-berg, line 6: Do we plan on publishing this? I would think it would be strictly internal.

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/SdkLogEmitter.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.appender.api;
+
+final class SdkLogEmitter implements LogEmitter {
```
jack-berg, line 8: `DelegatingLogEmitter`?

instrumentation/log4j/log4j-2.16/library/src/main/java/io/opentelemetry/instrumentation/log4j/v2_16/OpenTelemetryAppender.java
```diff
@@ -59,20 +52,12 @@ private OpenTelemetryAppender(
...
-    if (logEmitterProvider == null) {
-      // appender hasn't been initialized
-      return;
-    }
     LogBuilder builder =
-        logEmitterProvider.logEmitterBuilder(event.getLoggerName()).build().logBuilder();
+        GlobalLogEmitterProvider.get()
+            .logEmitterBuilder(event.getLoggerName())
+            .build()
+            .logBuilder();
```
jack-berg, line 59: This cleans up nicely!

instrumentation-appender-api/build.gradle.kts
jack-berg: Oh I suppose yes, since users now need to call `GlobalLogEmitterProvider.set(..)`.

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/SdkLogEmitter.java
jack-berg: Same advice applies to `SdkLogBuilder`, `SdkLogEmitterBuilder`, `SdkLogEmitterProvider`.

jack-berg:
This [README](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/log4j/log4j-2.16/library-autoconfigure/README.md) has references to using the `OpenTelemetryAppender`. Mind splitting off the section related to library instrumentation and updating to describe the new pattern?

jack-berg (commented):
Few comments, but looks good!

instrumentation-appender-api/build.gradle.kts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+group = "io.opentelemetry.instrumentation"
+
+dependencies {
+  implementation("io.opentelemetry:opentelemetry-sdk-logs")
```
trask, line 12:
```suggestion
  compileOnly("io.opentelemetry:opentelemetry-sdk-logs")
```

trask:
> This [README](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/log4j/log4j-2.16/library-autoconfigure/README.md) has references to using the `OpenTelemetryAppender`. Mind splitting off the section related to library instrumentation and updating to describe the new pattern?

👍 I'll add this to #4915 so we can postpone the doc update until 0.10.0 release

trask:
> * maybe this should just be part of the `instrumentation-api` artifact?

discussed this with @anuraaga and this appender api is (for the most part) is not really an end-user api (just used for a few appenders), compared to the instrumentation api.

although now thinking about it, the appender api may also be useful as a "event api" for end users to emit events.

which could be a reason to rename "appender api" to "log emitter api"...

(I'd like to gather feedback on these questions, but implement any renames/restructurings as follow-ups, so I can unblock #4927 and start writing some javaagent logging instrumentation in parallel)

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/GlobalLogEmitterProvider.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private static volatile LogEmitterProvider globalLogEmitterProvider = null;
+
+  @GuardedBy("mutex")
+  @Nullable
+  private static Throwable setGlobalCaller;
+
+  /** Returns the registered global {@link LogEmitterProvider}. */
+  public static LogEmitterProvider get() {
+    if (globalLogEmitterProvider == null) {
```
anuraaga, line 23: https://github.com/open-telemetry/opentelemetry-java/pull/4009

anuraaga (approved):

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/DelegatingLogEmitterProvider.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.appender.api;
+
+import io.opentelemetry.sdk.logs.SdkLogEmitterProvider;
+
+public final class DelegatingLogEmitterProvider implements LogEmitterProvider {
```
mateuszrzeszutek, line 10: In the agent these `Delegating...` classes will have to be loaded by the agent classloader (the api classes will be loaded by bootstrap), right? Does it make sense to at least put them into a separate package? (a separate module would probably be "cleaner", but that may be a bit too much for library instrumentations)

mateuszrzeszutek (approved):

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/DelegatingLogEmitterProvider.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.appender.api;
```
mateuszrzeszutek, line 6: WDYT about using `...instrumentation.api.appender`? Or even `...instrumentation.api.logs.appender`; the appender api would under the same parent package as `LoggingContextConstants` then

jack-berg (approved):

instrumentation-api-appender/src/main/java/io/opentelemetry/instrumentation/api/appender/GlobalLogEmitterProvider.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @GuardedBy("mutex")
+  @Nullable
+  private static Throwable setGlobalCaller;
+
+  /** Returns the registered global {@link LogEmitterProvider}. */
+  public static LogEmitterProvider get() {
+    LogEmitterProvider logEmitterProvider = globalLogEmitterProvider;
+    if (logEmitterProvider == null) {
+      synchronized (mutex) {
```
breedx-splk, line 22: I've stared at this way longer that I should have, and I can't poke any holes in its correctness.  😁 

It was initially concerning to me that there were two stateful variables being modified, and only one was officially guarded by the mutex. Having symmetry there would be nice for readability, but I assumed the `get()` case was trying to avoid the mutex entirely, probably for performance reasons. As we know, the double check that exists here only works correctly when the variable is `volatile` (which it is here), but we pay a performance penalty for that as well...

So how about changing `globalLogEmitterProvider` to be an atomic reference, get rid of the mutex, make `setGlobalCaller` volatile, and use `compareAndSet()` in the `set()` implementation and just `get()` in the `get()` implementation? Definitely cleaner.

The docs say that the memory semantics of `AtomicReference.get()` are the same as a volatile variable, but maybe you have deeper insight about the performance not being the same?

instrumentation-api-appender/src/main/java/io/opentelemetry/instrumentation/api/appender/LogEmitter.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * A {@link LogEmitter} is the entry point into a log pipeline.
+ *
+ * <p>Obtain a log builder via {@link #logBuilder()}, add properties using the setters, and emit it
+ * via {@link LogBuilder#emit()}.
+ */
+@ThreadSafe
+public interface LogEmitter {
+
+  /** Return a {@link LogBuilder} to emit a log. */
```
breedx-splk, line 19:
```suggestion
  /** Return a new {@link LogBuilder} to emit a log. */
```

Nitpicky, but I think calling out that it's a new instance helps the confusing api a little.

instrumentation-api-appender/src/main/java/io/opentelemetry/instrumentation/api/appender/LogEmitter.java
```diff
@@ -0,0 +1,21 @@
...
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * A {@link LogEmitter} is the entry point into a log pipeline.
+ *
+ * <p>Obtain a log builder via {@link #logBuilder()}, add properties using the setters, and emit it
+ * via {@link LogBuilder#emit()}.
+ */
+@ThreadSafe
```
breedx-splk, line 16: We have a deep precedent of adding `@ThreadSafe` to interfaces, but it continues to be weird to me because it's just a hint and it's always possible to implement a very thread unsafe implementation of these interfaces. 🤷🏻

instrumentation-api-appender/src/main/java/io/opentelemetry/instrumentation/api/appender/Severity.java
```diff
@@ -0,0 +1,44 @@
...
+  WARN3(15),
+  WARN4(16),
+  ERROR(17),
+  ERROR2(18),
+  ERROR3(19),
+  ERROR4(20),
+  FATAL(21),
+  FATAL2(22),
+  FATAL3(23),
+  FATAL4(24);
```
breedx-splk, line 33: are these in the sdk (still?). Maybe we could use the numeric constants from the sdk? I dunno.

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/DelegatingLogEmitterProvider.java
trask: oh no, this is absolutely right!

> a separate module would probably be "cleaner", but that may be a bit too much for library instrumentations

I went the separate module route so we don't have to do any weird shadow configuration for now, but I'll add this to the list of things to revisit.

instrumentation-appender-api/src/main/java/io/opentelemetry/instrumentation/appender/api/DelegatingLogEmitterProvider.java
trask: I flipped it, but didn't add the `logs` level, will add this to list of things to revisit.

breedx-splk (commented):
Looks good to me. A couple of comments.

instrumentation-api-appender/src/main/java/io/opentelemetry/instrumentation/api/appender/GlobalLogEmitterProvider.java
trask: I copied this from `GlobalOpenTelemetry`, would prefer to stay consistent with whatever approach is blessed in the SDK (and see @anuraaga's recent https://github.com/open-telemetry/opentelemetry-java/pull/4009) 😄

I did take a look at your suggestion, but it seemed like I still needed a coordinating synchronized block in both `get` and `set`. I'm likely missing something though.

instrumentation-api-appender/src/main/java/io/opentelemetry/instrumentation/api/appender/Severity.java
trask: ya, this is copied from sdk. the problem is that we this artifact needs to not depend on the sdk.

trask:
I have opened #4935 to track several follow-ups

PR #4919, opened 2021-12-16
Micrometer bridge instrumentation

mateuszrzeszutek:
I started working on upstreaming splunk's [micrometer bridge instrumentation](https://github.com/signalfx/splunk-otel-java/tree/main/instrumentation/micrometer-1.5/src/main/java/com/splunk/opentelemetry/instrumentation/micrometer/v1_5), now that metrics API is stable (well, almost stable). This PR contains the first few instruments  - and some questions about how we want to handle translating micrometer -> otel calls.

CC @jsuereth @jack-berg


instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryCounter.java
```diff
@@ -0,0 +1,69 @@
...
+            .build();
+    this.attributes = toAttributes(id.getTags());
+  }
+
+  @Override
+  public void increment(double v) {
+    if (removed) {
+      return;
+    }
+    otelCounter.add(v, attributes);
```
mateuszrzeszutek, line 45: Micrometer meters can be removed from the registry - that is not possible with otel metrics API; I had to use this little hack to emulate that behavior.
Why don't OTel metrics support this scenario? It seems to be a good idea to be able to turn off instrument, especially the asynchronous ones.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,167 @@
...
+        otelMeter
+            .histogramBuilder(id.getName())
+            .setDescription(description(id))
+            .setUnit("milliseconds")
+            .ofLongs()
+            .build();
+    this.attributes = toAttributes(id.getTags());
+
+    if (isUsingMicrometerHistograms()) {
+      measurements = new MicrometerHistogramMeasurements(clock, distributionStatisticConfig);
```
mateuszrzeszutek, line 53: Micrometer histograms are configurable via the API, the OTel ones aren't - in this PR I left the possibility to define "micrometer-style" histograms using micrometer settings (this is very similar to how `DropwizardMeterRegistry` works), completely separately from otel histograms.

instrumentation/micrometer/micrometer-1.5/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/TimerTest.java
```diff
@@ -0,0 +1,138 @@
...
+                                    .attributes()
+                                    .containsEntry("le", "1000"))
+                // TODO: I have absolutely no idea why otel drops these points below
+                /*
+                                        .anySatisfy(
+                                            point ->
+                                                assertThat(point)
+                                                    .hasValue(2)
+                                                    .attributes()
+                                                    .containsEntry("le", "10000"))
```
mateuszrzeszutek, line 122: It looks like OTel drops all values for a particular gauge after the first one is registered, even if other observations have completely different attributes. Is that correct behavior?

instrumentation/micrometer/micrometer-1.5/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/TimerTest.java
```diff
@@ -0,0 +1,138 @@
...
+    testing.waitAndAssertMetrics(
+        INSTRUMENTATION_NAME,
+        "testTimer",
+        metrics ->
+            metrics.anySatisfy(
+                metric ->
+                    assertThat(metric)
+                        .hasDescription("This is a test timer")
+                        .hasUnit("milliseconds")
+                        .hasDoubleHistogram()
```
mateuszrzeszutek, line 50: This is actually a long histogram -- perhaps we could improve our assertions so that they reflect that?

instrumentation/micrometer/micrometer-1.5/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/TimerTest.java
mateuszrzeszutek: Okay, this is actually this: https://github.com/open-telemetry/opentelemetry-java/issues/4005

mateuszrzeszutek:
I figured this PR is pretty much ready for review -- there are only 3 instruments implemented right now, but all the other ones are derivative and do not introduce any new concepts. And this PR is already kinda big and I don't really want to make it any bigger.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistry.java
```diff
@@ -0,0 +1,119 @@
...
+  private final AsyncInstrumentRegistry asyncInstrumentRegistry;
+
+  private OpenTelemetryMeterRegistry(Clock clock, io.opentelemetry.api.metrics.Meter otelMeter) {
+    super(clock);
+    this.otelMeter = otelMeter;
+    this.asyncInstrumentRegistry = new AsyncInstrumentRegistry(otelMeter);
+  }
+
+  @Override
+  protected <T> Gauge newGauge(Meter.Id id, T t, ToDoubleFunction<T> toDoubleFunction) {
```
jack-berg, line 49: Could use `@Nullable` annotations and nullaway throughout. For example, `t` is nullable here, and propagates in a lot of places which make it easy to forget.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryGauge.java
```diff
@@ -0,0 +1,58 @@
...
+      Id id,
+      T obj,
+      ToDoubleFunction<T> objMetric,
+      AsyncInstrumentRegistry asyncInstrumentRegistry) {
+    this.id = id;
+    this.attributes = toAttributes(id.getTags());
+    this.asyncInstrumentRegistry = asyncInstrumentRegistry;
+
+    asyncInstrumentRegistry.buildGauge(
+        id.getName(), description(id), baseUnit(id), attributes, obj, objMetric);
```
jack-berg, line 34: Could pass `Id` here instead of `name, `description`, and `baseUnit` to reduce the arguments.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryCounter.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Collections;
+
+final class OpenTelemetryCounter implements Counter, RemovableMeter {
+
+  private final Id id;
+  // TODO: use bound instruments when they're available
+  private final DoubleCounter otelCounter;
+  private final Attributes attributes;
+
+  volatile boolean removed = false;
```
jack-berg, line 26: Can be private.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,167 @@
...
+      DistributionStatisticConfig distributionStatisticConfig,
+      PauseDetector pauseDetector,
+      Meter otelMeter) {
+    super(id, clock, distributionStatisticConfig, pauseDetector, TimeUnit.MILLISECONDS, false);
+
+    this.otelHistogram =
+        otelMeter
+            .histogramBuilder(id.getName())
+            .setDescription(description(id))
+            .setUnit("milliseconds")
```
jack-berg, line 47: Do we keep a constants class anywhere with units like this?

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+final class OpenTelemetryTimer extends AbstractTimer implements RemovableMeter {
+
+  // TODO: use bound instruments when they're available
+  private final LongHistogram otelHistogram;
+  private final Attributes attributes;
+  private final Measurements measurements;
+
+  volatile boolean removed = false;
```
jack-berg, line 33: Can be private.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,167 @@
...
+      return Double.NaN;
+    }
+  }
+
+  // calculate count, totalTime and max value for the use of micrometer histograms
+  // kinda similar to how DropwizardTimer does that
+  private static final class MicrometerHistogramMeasurements implements Measurements {
+
+    private final AtomicLong count = new AtomicLong(0);
+    private final AtomicLong totalTime = new AtomicLong(0);
```
jack-berg, line 136: Do we need to be worried about performance problems with high contention here? Should we use `LongAdder` instead?

The usage of `TmeWindowMax` suggests that performance is a consideration.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/Bridging.java
```diff
@@ -0,0 +1,41 @@
...
+package io.opentelemetry.javaagent.instrumentation.micrometer.v1_5;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Tag;
+import io.opentelemetry.api.common.AttributeKey;
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.instrumentation.api.cache.Cache;
+
+final class Bridging {
```
jack-berg, line 15: BridgingUtil?

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,167 @@
...
+    // OpenTelemetry metrics bridge does not support reading measurements
+    return Collections.emptyList();
+  }
+
+  @Override
+  public void onRemove() {
+    removed = true;
+  }
+
+  private interface Measurements {
```
jack-berg, line 97: I struggled to understand the purpose of this interface based on its name alone. Is there something that better captures the abstraction it represents?

instrumentation/micrometer/micrometer-1.5/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/TimerTest.java
jsuereth: We only export double histograms from an aggregation standpoint.  Practically speaking, the only difference is whether the "sum" field is a double or long.

instrumentation/micrometer/micrometer-1.5/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
+plugins {
+  id("otel.javaagent-instrumentation")
```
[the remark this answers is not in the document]
jsuereth: +100

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryCounter.java
jsuereth: That's a great question.  I think it's worth opening a feature request.

Short answer: outside of plumbing, we do some gc of metric storage.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryCounter.java
```diff
@@ -0,0 +1,69 @@
...
+    if (removed) {
+      return;
+    }
+    otelCounter.add(v, attributes);
+  }
+
+  @Override
+  public double count() {
+    // OpenTelemetry metrics bridge does not support reading measurements
+    return Double.NaN;
```
jsuereth, line 51: Do we know how this is commonly used?

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
jsuereth: Unit should be "ms" according to spec.

instrumentation/micrometer/micrometer-1.5/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/GaugeTest.java
```diff
@@ -0,0 +1,143 @@
...
+                        .satisfiesExactly(
+                            point ->
+                                assertThat(point)
+                                    .hasValue(42)
+                                    .attributes()
+                                    .containsOnly(attributeEntry("tag", "value")))));
+    testing.clearData();
+
+    // when
+    Metrics.globalRegistry.remove(gauge);
```
trask, line 66:
```suggestion
    // when
    Metrics.globalRegistry.remove(gauge);
    Thread.sleep(10); // give time for any inflight metric export to be received
    testing.clearData();
```

trask (approved):
nice!

instrumentation/micrometer/micrometer-1.5/javaagent/build.gradle.kts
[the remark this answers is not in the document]
mateuszrzeszutek: I realized that this would make a very useful library instrumentation some time after opening this PR 😅 
I'll split it out in one of the next PRs - once all meters/instruments are implemented.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
mateuszrzeszutek: Micrometer itself is a bit inconsistent here: `DropwizardTimer` uses `AtomicLong` for total time, but `StepTimer` uses `LongAdder`s. Perhaps it doesn't really matter that much for the dropwizard integration? Anyway, using adders is probably better here, so I'll do just that.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
mateuszrzeszutek: It quite literally represents the [`Measurement`s](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/Measurement.java) made by the timer, just grouped together. Do you have any suggestion on how to name this?

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/Bridging.java
mateuszrzeszutek: I don't think that `Util` adds anything useful to the name; and we already have a couple of `*Bridging` classes in the codebase. I don't really have a strong opinion though, both names are fine.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/AsyncInstrumentRegistry.java
```diff
@@ -0,0 +1,90 @@
...
+                  .buildWithCallback(recorder);
+              return recorder;
+            })
+        .addGaugeMeasurement(attributes, obj, objMetric);
+  }
+
+  void removeGauge(String name, Attributes attributes) {
+    GaugeMeasurementsRecorder recorder = gauges.get(name);
+    if (recorder != null) {
+      recorder.removeGaugeMeasurement(attributes);
```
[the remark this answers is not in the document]
mateuszrzeszutek: Oh, great point 💯

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryCounter.java
mateuszrzeszutek: I think it's only supposed to be used in the `measure()` method - which in turn is only called in actual `MeterRegistry` implementations (like the Signalfx one, for example). But since this is public API anyone can call it -- I don't really know why though, it's not particularly useful.
The `Timer` is a bit of an exception to that - its `count()`, `totalTime()` and `max()` methods are also called to compute the histogram.

> should we log a one-time warning/debug in case it's being used?

👍

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryCounter.java
mateuszrzeszutek: Done:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4919
https://github.com/open-telemetry/opentelemetry-specification/issues/2232

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/Bridging.java
jack-berg: I don't feel strongly. If there's precedent for "Bridging", let's keep it.

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
jack-berg: Can't think of a good one.

jack-berg (approved):

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/Bridging.java
```diff
@@ -0,0 +1,41 @@
...
+
+  private static final Cache<String, AttributeKey<String>> tagsCache = Cache.bounded(64);
```
anuraaga, line 17: Think this could be higher, maybe 1024, since 64 seems a bit limiting, and the values are tiny

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
```diff
@@ -0,0 +1,170 @@
...
+      PauseDetector pauseDetector,
+      Meter otelMeter) {
+    super(id, clock, distributionStatisticConfig, pauseDetector, TimeUnit.MILLISECONDS, false);
+
+    this.otelHistogram =
+        otelMeter
+            .histogramBuilder(id.getName())
+            .setDescription(description(id))
+            .setUnit("ms")
+            .ofLongs()
```
anuraaga, line 48: I think this should be double and we should be recording the fractional part of ms, currently we seem to be truncating

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/UnsupportedReadLogger.java
```diff
@@ -0,0 +1,24 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class UnsupportedReadLogger {
+
+  private static final Logger logger = LoggerFactory.getLogger(OpenTelemetryMeterRegistry.class);
+  private static final AtomicBoolean done = new AtomicBoolean(false);
+
+  static void logWarning() {
+    if (done.compareAndSet(false, true)) {
```
anuraaga, line 18: Instead you can put the log message in a static initializer, which makes sure it's only logged once on classload, which happens lazily while calling from the Unsupported method

anuraaga (approved):

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/UnsupportedReadLogger.java
mateuszrzeszutek: Yeah, that's a really good idea, will do that. Thanks!

instrumentation/micrometer/micrometer-1.5/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/micrometer/v1_5/OpenTelemetryTimer.java
mateuszrzeszutek: 👍  added

instrumentation/micrometer/micrometer-1.5/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+  pass {
+    group.set("io.micrometer")
+    module.set("micrometer-core")
+    versions.set("[1.5.0,)")
+    assertInverse.set(true)
+  }
+}
+
+dependencies {
+  library("io.micrometer:micrometer-core:1.5.0")
```
jonatan-ivanov, line 15: This is a pretty old version, it is not supported anymore, I would go with `1.8.x` (latest).

anuraaga: This is a compile dependency, users bring in the version that they want. Our standard practice is to target the lowest version we support as the compile dependency, while running tests on that and the latest version

jonatan-ivanov: I get that but do you want to support a version that reached its end of life and nobody should use it in prod?
Currently the oldest version that we support is the 1.7.x line which will reach its EOL in May.

mateuszrzeszutek: We will support all versions from 1.5 upwards, 1.8 included. As Anuraag mentioned, we generally aim to support the lowest version possible - the javaagent is often attached to applications that haven't been updated for some time; and it is supposed to work with no code changes (like upgrading the micrometer dependency version) at all.

PR #4932, opened 2021-12-17
Create producer span from spring integration instrumentation

laurit:
Currently we create producer span only from rabbitmq, kafka, jms and rocketmq instrumentations. Due to this producer spans are not created when an unsupported messaging library or a homegrown solution is used. This pr adds a flag that allows creating producer span in spring integration instrumentation which would create the span regardless of the underlying messaging library. The downside of this is that in spring instrumentation channel interceptor we have very little info available and can't fill many of the messaging semantic attributes. Also we can't enable this flag by default as it would suppress the producer span created by the messaging library that would have access to more semantic attributes. Another surprising issue with creating producer span in spring integration channel interceptor is that it is hard to tell whether the current channel is an input or output channel, thus we don't generally know whether we should be creating consumer or producer span. Luckily spring cloud stream adds a `type` attribute to its channels which describes the direction of the channel. We only create the producer span when the current channel is from spring cloud stream and we know that it is an output channel. If needed we could extend this to support more channel types by using a similar logic to what spring cloud sleuth uses, if channel is a DirectChannel then treat it as consume, otherwise producer.


instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -175,4 +199,66 @@ private static void ensureNativeHeadersAreMutable(MessageHeaderAccessor headerAc
...
+    } catch (ClassNotFoundException ignore) {
+      return null;
+    }
+  }
+
+  private static MethodHandle getChannelAttributeMh(Class<?> directWithAttributesChannelClass) {
+    try {
+      return MethodHandles.lookup()
+          .findVirtual(
+              directWithAttributesChannelClass,
```
mateuszrzeszutek, line 221: Nit: missing null-check for class

instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java
```diff
@@ -23,16 +29,23 @@
 
 final class TracingChannelInterceptor implements ExecutorChannelInterceptor {
 
+  private static final boolean PRODUCER_SPAN_ENABLED =
+      Config.get().getBoolean("otel.instrumentation.spring-integration.producer.enabled", false);
```
mateuszrzeszutek, line 33: Can we document this setting? Like in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4981

mateuszrzeszutek (approved):

instrumentation/spring/README.md
```diff
@@ -17,8 +17,9 @@ In this guide we will be using a running example. In section one and two, we wil
 | System property | Type | Default | Description |
 |---|---|---|---|
 | `otel.instrumentation.spring-integration.global-channel-interceptor-patterns` | List | `*` | An array of Spring channel name patterns that will be intercepted. See [Spring Integration docs](https://docs.spring.io/spring-integration/reference/html/channel.html#global-channel-configuration-interceptors) for more details. |
+| `otel.instrumentation.spring-integration.producer.enabled` | Boolean | `false` | Create producer span in spring integration instrumentation when message is written into an output channel. Detecting output channels works only for [Spring Cloud Stream](https://spring.io/projects/spring-cloud-stream) `DirectWithAttributesChannel`. Use this flag only when you are using a messaging library that does not have instrumentation to create producer spans. |
```
mateuszrzeszutek, line 20: CC @theletterf

mateuszrzeszutek (approved):

instrumentation/spring/README.md
theletterf, line 20:
```suggestion
| `otel.instrumentation.spring-integration.producer.enabled` | Boolean | `false` | Create producer spans when  messages are sent to an output channel. Enable when you're using a messaging library that doesn't have its own instrumentation for generating producer spans. Note that the detection of output channels only works for [Spring Cloud Stream](https://spring.io/projects/spring-cloud-stream) `DirectWithAttributesChannel`.  |
```

instrumentation/spring/spring-integration-4.1/javaagent/build.gradle.kts
```diff
@@ -45,11 +45,24 @@ tasks {
...
 
+  val testWithProducerInstrumentation by registering(Test::class) {
+    filter {
+      includeTestsMatching("SpringCloudStreamProducerTest")
+      isFailOnNoMatchingTests = false
+    }
+    include("**/SpringCloudStreamProducerTest.*")
+    jvmArgs("-Dotel.instrumentation.rabbitmq.enabled=false")
+    jvmArgs("-Dotel.instrumentation.spring-rabbit.enabled=false")
+    jvmArgs("-Dotel.instrumentation.spring-integration.producer.enabled=true")
```
trask, line 56: 👍

trask (approved):

PR #4933, opened 2021-12-17
Verify System.out.printlns using checkstyle

### LINKED ISSUE #4926: disable("SystemOut") in errorprone
Since #4825, I'm constantly battling with this during development / troubleshooting.

I agree with not breaking gradle cache, so that leaves the option of disabling this rule unconditionally (both local and CI).

Or any other ideas?

trask:
Resolves #4926


buildscripts/checkstyle.xml
```diff
@@ -375,6 +375,14 @@
     <module name="SuppressWarningsHolder"/>
     <!-- this enables CHECKSTYLE:OFF and CHECKSTYLE:ON comments -->
     <module name="SuppressionCommentFilter"/>
+    <module name="Regexp">
+      <!-- see comment in otel.errorprone-conventions.gradle.kts for why we disable the SystemOut
+           rule there and rely on this check -->
+      <property name="id" value="SystemOut"/>
+      <!-- . matches any character, so we need to escape it and use \. to match dots. -->
+      <property name="format" value="System\.out\.println"/>
```
laurit, line 383: I believe errorprone also detects `System.err.println()` and `printStackTrace()` with the same rule. It might be better to check for `print` instead of `println`.

iNikem: +1 for `print` to catch `printf` also

anuraaga: Let's add Thread.dumpStack too

iNikem (approved):

anuraaga (approved):

PR #4936, opened 2021-12-18
Simplify the thread-safe handling of the globalLogEmitterProvider

breedx-splk:
Follow-up on some comments in #4917 

@trask This should show what I was getting at with my comments before. There are two slight semantic differences with this approach: 1) calling `get()` before `set()` is ok, you just get the NoOp impl. 2) calling `set()` multiple times with the NoOp impl is fine. I think this is acceptable, but others could differ.

I completely understand if we don't want to do this because it's inconsistent with the others.

### DIFF
--- a/instrumentation-api-appender/src/test/java/io/opentelemetry/instrumentation/api/appender/GlobalLogEmitterProviderTest.java
+++ b/instrumentation-api-appender/src/test/java/io/opentelemetry/instrumentation/api/appender/GlobalLogEmitterProviderTest.java
@@ -7,6 +7,8 @@ package io.opentelemetry.instrumentation.api.appender;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.mockito.Mockito.mock;
 
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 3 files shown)


trask:
oh, yes, this is *much* simpler!

> 1) calling `get()` before `set()` is ok, you just get the NoOp impl.

this behavior could be better, e.g. in cases where you aren't able to set the global early enough, though I'm not sure how it will work with autoconfiguration (though to be honest I don't know how any of this will work with autoconfiguration yet)

> 2) calling `set()` multiple times with the NoOp impl is fine. I think this is acceptable, but others could differ.

the NoOp impl is package-private, so I don't think this is an issue

instrumentation-api-appender/src/test/java/io/opentelemetry/instrumentation/api/appender/GlobalLogEmitterProviderTest.java
```diff
@@ -32,26 +34,31 @@ void testGlobalBeforeSet() {
...
+    GlobalLogEmitterProvider.set(NoopLogEmitterProvider.INSTANCE);
+    GlobalLogEmitterProvider.set(NoopLogEmitterProvider.INSTANCE);
+    GlobalLogEmitterProvider.set(NoopLogEmitterProvider.INSTANCE);
     GlobalLogEmitterProvider.set(NoopLogEmitterProvider.INSTANCE);
+    // pass
   }
 
-  private static LogEmitterProvider getLogEmitterProvider() {
-    return GlobalLogEmitterProvider.get();
+  private void setLogEmitterProvider() {
```
mateuszrzeszutek, line 61: Errorprone sure can be annoying sometimes:
>: warning: [MethodCanBeStatic] A private method that does not reference the enclosing instance can be static

```suggestion
  private static void setLogEmitterProvider() {
```

mateuszrzeszutek (approved):

trask (approved):
thx!

PR #4940, opened 2021-12-19
Temurin JDK

sullis:
https://github.com/actions/setup-java/blob/main/README.md

NOTE: Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt to temurin to keep receiving software and security updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up JDK 11 for running Gradle
         uses: actions/setup-java@v2
         with:
-          distribution: adopt
+          distribution: temurin
           java-version: 11
 
       - name: Build
@@ -39,7 +39,7 @@ jobs:
       - name: Set up JDK 11 for running Gradle
         uses: actions/setup-java@v2
         with:
-          distribution: adopt
+          distribution: temurin
           java-version: 11
 
       - name: Build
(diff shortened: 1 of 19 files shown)


.github/workflows/ci.yml
```diff
@@ -67,13 +67,13 @@ jobs:
         name: Set up JDK ${{ matrix.test-java-version }}-${{ matrix.vm }} for running tests
         uses: actions/setup-java@v2
         with:
-          distribution: adopt-${{ matrix.vm }}
+          distribution: temurin-${{ matrix.vm }}
```
mateuszrzeszutek, line 70: Looks like it fails on this part:

```
Error: No supported distribution was found for input temurin-hotspot
```

We should probably keep using `adopt-openj9` for openj9-based jvms - I don't think the `setup-java` action supports anything else: https://github.com/actions/setup-java#supported-distributions

mateuszrzeszutek (commented):
Thanks!

.github/workflows/ci.yml
trask: hey @sullis! let us know if you can get to fixing the build, otherwise one of us can do that

sullis: @trask I don't have time to fix it. Can you finish it?

iNikem: I can take it over

laurit:
Thanks, we have completed the switch to Eclipse Temurin.

PR #4944, opened 2021-12-20
Add log4j2 appender javaagent instrumentation

trask:
There's no autoconfiguration of log exporter (yet), so this instrumentation doesn't benefit end users unless they explicitly configure a logging exporter via javaagent extension (this is how the tests work).


instrumentation/log4j/log4j-appender-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/appender/v2_0/Log4jAppenderInstrumentation.java
```diff
@@ -0,0 +1,115 @@
...
+    return extendsClass(named("org.apache.logging.log4j.spi.AbstractLogger"));
+  }
+
+  @Override
+  public ElementMatcher<ClassLoader> classLoaderOptimization() {
+    return hasClassesNamed("org.apache.logging.log4j.spi.AbstractLogger");
+  }
+
+  @Override
+  public void transform(TypeTransformer transformer) {
```
anuraaga, line 41: Had forgotten about the log4j appender we have. Is the reason we can't use logic from the library instrumentation to support old versions of log4j? Log transformation is much more complicated than ID injection so we shouldn't duplicate instrumentation I think if at all possible

trask: yeah, there are some things we can share and some we can't due to 2.0 vs 2.16. I'll send a follow-up that extracts the part(s) that are shareable so we can see if it's worth an extra module for those.

anuraaga: By the way - do we need to support <2.16? :P

trask: let me think longer on this, I'm actually not sure...

I will update this PR to target 2.16 and share mapping code with the library instrumentation. I can always propose a separate 2.0 javaagent version later (that in addition to capturing logs, logs a warning about using a bad log4j version).

anuraaga: Thanks a lot!

instrumentation/log4j/log4j-appender-2.16/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+dependencies {
+  library("org.apache.logging.log4j:log4j-core:2.16.0")
+
+  compileOnly(project(":instrumentation-api-appender"))
+
+  implementation(project(":instrumentation:log4j:log4j-appender-2.16:library"))
+
+  testImplementation("org.awaitility:awaitility")
+
+  testImplementation("com.lmax:disruptor:3.4.2")
```
iNikem, line 23: Why this is needed?

iNikem (approved):

instrumentation/log4j/log4j-appender-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/appender/v2_0/Log4jAppenderInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
...
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class Log4jAppenderInstrumentationModule extends InstrumentationModule {
+
+  public Log4jAppenderInstrumentationModule() {
+    super("log4j-appender", "log4j-appender-2.0");
```
mateuszrzeszutek, line 19: Shouldn't it be 2.16?
```suggestion
    super("log4j-appender", "log4j-appender-2.16");
```

instrumentation/log4j/log4j-appender-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/appender/v2_0/Log4jAppenderInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.log4j.appender.v2_0;
```
mateuszrzeszutek, line 6:
```suggestion
package io.opentelemetry.javaagent.instrumentation.log4j.appender.v2_16;
```

instrumentation/log4j/log4j-appender-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/appender/v2_0/Log4jAppenderInstrumentation.java
```diff
@@ -0,0 +1,79 @@
...
+        isMethod()
+            .and(isProtected().or(isPublic()))
+            .and(named("log"))
+            .and(takesArguments(6))
+            .and(takesArgument(0, named("org.apache.logging.log4j.Level")))
+            .and(takesArgument(1, named("org.apache.logging.log4j.Marker")))
+            .and(takesArgument(2, named("java.lang.String")))
+            .and(takesArgument(3, named("java.lang.StackTraceElement")))
+            .and(takesArgument(4, named("org.apache.logging.log4j.message.Message")))
+            .and(takesArgument(5, named("java.lang.Throwable"))),
```
mateuszrzeszutek, line 52: Nit
```suggestion
            .and(takesArgument(0, named("org.apache.logging.log4j.Level")))
            .and(takesArgument(1, named("org.apache.logging.log4j.Marker")))
            .and(takesArgument(2, String.class))
            .and(takesArgument(3, StackTraceElement.class))
            .and(takesArgument(4, named("org.apache.logging.log4j.message.Message")))
            .and(takesArgument(5, Throwable.class)),
```

instrumentation/log4j/log4j-appender-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/appender/v2_0/Log4jHelper.java
```diff
@@ -0,0 +1,29 @@
...
+package io.opentelemetry.javaagent.instrumentation.log4j.appender.v2_0;
+
+import io.opentelemetry.instrumentation.api.appender.GlobalLogEmitterProvider;
+import io.opentelemetry.instrumentation.api.appender.LogBuilder;
+import io.opentelemetry.instrumentation.log4j.appender.v2_16.internal.LogEventMapper;
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.Message;
+
+public class Log4jHelper {
```
mateuszrzeszutek, line 15:
```suggestion
public final class Log4jHelper {
```

instrumentation/log4j/log4j-appender-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/appender/v2_0/Log4jHelper.java
```diff
@@ -0,0 +1,29 @@
...
+        GlobalLogEmitterProvider.get().logEmitterBuilder(logger.getName()).build().logBuilder();
+
+    LogEventMapper.setBody(builder, message);
+    LogEventMapper.setSeverity(builder, level);
+    LogEventMapper.setThrowable(builder, throwable);
+    LogEventMapper.setContext(builder);
+
+    builder.emit();
+  }
+}
```
mateuszrzeszutek, line 29:
```suggestion

  private Log4jHelper() {}
}
```

instrumentation/log4j/log4j-appender-2.16/javaagent/build.gradle.kts
trask: I added a comment to explain, it's needed for the async logger test

Issue #4023, opened 2021-08-30
Instrument RxJava 3.1.1+


mateuszrzeszutek:
RxJava 3.1.1 has moved some classes between packages (internal -> non-internal, see the [v3.1.1 release notes](https://github.com/ReactiveX/RxJava/releases/tag/v3.1.1)) and as a result fo that our instrumentation stopped working (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4022).
We should probably introduce a new (very similar I guess) instrumentation for 3.1.1+

PR #4954, opened 2021-12-20
Instrumentation for rxjava3.1.1

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4023


iNikem (approved):

anuraaga (approved):

trask (approved):
we finally got rid of our one "patch release" base version (log4j 2.13.2) only to get a new one 😢

instrumentation/rxjava/rxjava-3.1.1/library/src/main/java/io/opentelemetry/instrumentation/rxjava3/TracingAssembly.java
```diff
@@ -0,0 +1,324 @@
...
+ *
+ * <p>In order to enable RxJava3 instrumentation one has to call the {@link
+ * TracingAssembly#enable()} method.
+ *
+ * <p>Instrumentation uses <code>on*Assembly</code> and <code>on*Subscribe</code> RxJavaPlugin hooks
+ * to wrap RxJava3 classes in their tracing equivalents.
+ *
+ * <p>Instrumentation can be disabled by calling the {@link TracingAssembly#disable()} method.
+ */
+public final class TracingAssembly {
```
mateuszrzeszutek, line 57: This class has exactly the same name & package as the 3.0 one - won't they conflict inside the agent?

laurit: Thanks, renamed 3.1.1 classes.

laurit:
> we finally got rid of our one "patch release" base version (log4j 2.13.2) only to get a new one 😢

I don't like it either. What alternatives do we have? The whole problem stems from rxjava moving `ConditionalSubscriber` and `QueueSubscription` interfaces to a different package. For `QueueSubscription` we could use reflection to invoke `requestFusion` method. With `ConditionalSubscriber` it is more troublesome because we have  `TracingConditionalSubscriber extends BasicFuseableSubscriber` where `BasicFuseableSubscriber` takes `ConditionalSubscriber` as constructor argument. I guess we could use a jdk proxy that implements the same interfaces as `BasicFuseableSubscriber` and copy all the code from `TracingConditionalSubscriber` and `BasicFuseableSubscriber` into proxies invocation handler.

trask:
ya, no worries, I think it's a lost cause trying too hard to limit our version splitting...

eugeniyk:
@trask will this change apply to `3.1.6` and other higher than `3.1.1` versions automatically? The reason I'm asking is that after `1.32.0` -> `2.15.0` bump of javaagent, context stop being propagated for some reason

laurit:
@eugeniyk commenting on a 4 year PR is a sure way for your question to get ignored. If you believe you have found a bug then please open a new issue. Include a minimal application that reproduces the problem.

PR #4955, opened 2021-12-20
Update library autoconfigure artifact names

trask:
Update artifact names for the `library-autoconfigure` artifacts:
* apache-dubbo-2.7-autoconfigure -> apache-dubbo-autoconfigure-2.7
* aws-sdk-1.11-autoconfigure -> aws-sdk-autoconfigure-1.11
* aws-sdk-2.2-autoconfigure -> aws-sdk-autoconfigure-2.2
* log4j-thread-context-2.16-autoconfigure -> log4j-thread-context-autoconfigure-2.16

The previous name could make sense too (it's `-autoconfigure` appended on to the artifact name that it autoconfigures, except I guess for dubbo which doesn't have a non-autoconfigure artifact yet).

But I think keeping the base version at the end seems better for consistency of this (already weird) artifact naming.

### DIFF
--- a/instrumentation/apache-dubbo-2.7/library-autoconfigure/build.gradle.kts
+++ b/instrumentation/apache-dubbo-2.7/library-autoconfigure/build.gradle.kts
@@ -2,6 +2,8 @@ plugins {
   id("otel.library-instrumentation")
 }
 
+base.archivesName.set("apache-dubbo-autoconfigure-2.7")
+
 dependencies {
   compileOnly("com.google.auto.value:auto-value-annotations")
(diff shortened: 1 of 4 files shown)


anuraaga (approved):

iNikem:
I like new names even less than the old ones :(

> The previous name could make sense too (it's -autoconfigure appended on to the artifact name that it autoconfigures, except I guess for dubbo which doesn't have a non-autoconfigure artifact yet).

This makes sense to me

trask:
Ok, works for me, the existing naming does keep the base version next to the library name which is good, and I don't think we have any other precedence for having something between the library name and the base version.

PR #4961, opened 2021-12-21
Remove test that doesn't seem to check what is intended

anuraaga:
Happened to randomly read this file and could make no sense of this test.

`onTestClasspath` is set to true even though the class isn't on the test classpath, it's in the agent (using debugger on the result of `getResource()` it is obviously reading from the agent jar). This probably predates our usage of a real javaagent for tests.

`.concurrent.State` doesn't exist anymore so I tried `PropagatedContext` in a debugger, and saw the same behavior with Java8BytecodeBridge, can load class, can get resource. The title of the test says "can find classes but not resources" which doesn't seem to ever apply at least in this test code.

Blame shows glassfish-related commits from 3 years ago, I suspect our glassfish smoke tests are doing the needed job while this test doesn't seem to test anything important.


trask (approved):
👍

iNikem (approved):

PR #4962, opened 2021-12-21
Fix some gauge types

anuraaga:


trask (approved):

visonliamazon:
Hi @anuraaga , @trask ,

May I know the why we change the `gauge` type to `unDownCounter` here?

I think it doesn't make sense for the metrics like Java JVM memory metrics to have `increments/decrements` values.
Instead, it should use `actual value` in the metrics as I mentioned in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8748.


Feel free to give any feedbacks. Thanks!

PR #4966, opened 2021-12-22
Support log4j map messages

trask:
Introduces experimental setting `otel.instrumentation.log4j-appender.experimental.capture-map-message-attributes` to capture log4j `MapMessage` key/value pairs as log attributes.

It doesn't add any prefix to those keys because this is going to be the primary mechanism for users to report arbitrary key/value attributes on logs.

I would support enabling the default to true for this setting, since it is critical for supporting structured events, but it is disabled by default in this PR for now.


instrumentation/log4j/log4j-appender-2.16/library/src/main/java/io/opentelemetry/instrumentation/log4j/appender/v2_16/internal/LogEventMapper.java
```diff
@@ -76,17 +86,22 @@ public void mapLogEvent(
       @Nullable Instant timestamp,
       T contextData) {
 
+    AttributesBuilder attributes = Attributes.builder();
+
     if (message != null) {
-      builder.setBody(message.getFormattedMessage());
+      if (message instanceof MapMessage) {
+        builder.setBody(message.getFormat());
```
anuraaga, line 93: Weird API :O

Looks like `getFormat` returns the message for `StructuredDataMessage` and empty string for `StringMapMessage`. From what I've seen, it's conventional to use a map message with an attribute `message` for the body. How about doing `instanceof StructuredDataMessage` for `message.getFormat()`, and checking the attribute for StringMapMessage, leaving Body empty if it's not set?

instrumentation/log4j/log4j-appender-2.16/library/src/main/java/io/opentelemetry/instrumentation/log4j/appender/v2_16/internal/LogEventMapper.java
```diff
@@ -105,6 +118,42 @@ public void mapLogEvent(
...
+      builder.setBody(message.getFormattedMessage());
+      return;
+    }
+
+    MapMessage<?, ?> mapMessage = (MapMessage<?, ?>) message;
+
+    String body = mapMessage.getFormat();
+    boolean checkSpecialMapMessageAttribute = (body == null || body.isEmpty());
+    if (checkSpecialMapMessageAttribute) {
+      body = mapMessage.get(SPECIAL_MAP_MESSAGE_ATTRIBUTE);
```
mateuszrzeszutek, line 136: Can you add tests that verify this scenario?

mateuszrzeszutek (approved):

Issue #4947, opened 2021-12-20
Add logback appender instrumentation module


trask:
Similar to the log4j2 appender instrumentation module.

jack-berg:
I'm working on this.

PR #4968, opened 2021-12-22
Add support for capturing logback mdc attributes

trask:
Following convention from #4959 (log4j)

Split out the mapper class into `library` instrumentation to enable unit testing of it, and we will want logback appender (library) shortly anyways (#4947)


anuraaga (approved):

instrumentation/logback/logback-appender-1.0/library/src/main/java/io/opentelemetry/instrumentation/logback/appender/v1_0/internal/LoggingEventMapper.java
```diff
@@ -0,0 +1,163 @@
...
+import java.io.StringWriter;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+public final class LoggingEventMapper {
+
+  public static final LoggingEventMapper INSTANCE = new LoggingEventMapper();
+
+  private static final Cache<String, AttributeKey<String>> mdcAttributeKeys = Cache.bounded(100);
```
anuraaga, line 33: Do you think there'd be a benefit in having a shared cache in an `.internal.` class in instrumentation-api?

jack-berg (approved):
Looks good. Adding an appender is really trivial at this point. Will open a PR once this is merged.

instrumentation/logback/logback-appender-1.0/library/src/main/java/io/opentelemetry/instrumentation/logback/appender/v1_0/internal/LoggingEventMapper.java
trask: Created #5003 to discuss/track

PR #4975, opened 2021-12-26
Generate jApiCmp report under the build folder

iNikem:
Otherwise those reports live in some limbo space. Being outside of the build folder, they are not cleaned by `gradle clean`. At the same time they are not added to git. Until we establish a proper workflow around them, let them be generated under the `build` folder.

### DIFF
--- a/conventions/src/main/kotlin/otel.japicmp-conventions.gradle.kts
+++ b/conventions/src/main/kotlin/otel.japicmp-conventions.gradle.kts
@@ -73,7 +73,7 @@ tasks {
     packageExcludes = listOf("*.internal", "*.internal.*")
     val baseVersionString = if (apiBaseVersion == null) "latest" else baselineVersion
     val newVersionString = if (apiNewVersion == null) "current" else apiNewVersion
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek:
Aren't we supposed to keep these in git? Like the SDK repo does; you can easily check for the differences between versions this way. If we store them in `build/` then we'll be basically discarding them without using.

iNikem:
> Aren't we supposed to keep these in git? Like the SDK repo does; you can easily check for the differences between versions this way. If we store them in `build/` then we'll be basically discarding them without using.

We are. But I think that the process of adding specific diff to git should be manual. And it may then include copying files from the `build` folder to their final destination.

iNikem:
@open-telemetry/java-instrumentation-approvers ping :)

trask:
I'm probably missing something, will this require manually copying over on each build?

The good thing about generating it directly into source control is that we could potentially fail the build if someone forgets (which seems like it is easy to do).

Let's discuss with SDK repo folks on Thu to get their thoughts since they've been doing this for a while already?

Issue #4972, opened 2021-12-23
Add support for Ktor


yonigibbs:
Would it be possible to add support for Ktor, both as a server and a client, please?

PR #4983, opened 2021-12-28
Add library instrumentation for ktor server

anuraaga:
For #4972

This seemed like a fun instrumentation to try to get more practice with Kotlin. It wasn't fun when trying to integrate it with Spock :P


instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorServerTracing.kt
```diff
@@ -0,0 +1,161 @@
...
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter
+import io.opentelemetry.instrumentation.api.instrumenter.SpanStatusExtractor
+import io.opentelemetry.instrumentation.api.instrumenter.http.CapturedHttpHeaders
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpServerMetrics
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpSpanNameExtractor
+import io.opentelemetry.instrumentation.api.instrumenter.http.HttpSpanStatusExtractor
+import io.opentelemetry.instrumentation.api.servlet.ServerSpanNaming
+import kotlinx.coroutines.withContext
+
+class KtorServerTracing private constructor(
```
anuraaga, line 28: Reference while writing this

https://github.com/ktorio/ktor/blob/1.6.7/ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/MicrometerMetrics.kt

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorServerTracing.kt
```diff
@@ -0,0 +1,161 @@
...
+          }
+        } else {
+          proceed()
+        }
+      }
+
+      pipeline.environment.monitor.subscribe(Routing.RoutingCallStarted) { call ->
+        val context = call.attributes.getOrNull(contextKey)
+        if (context != null) {
+          ServerSpanNaming.updateServerSpanName(context, ServerSpanNaming.Source.SERVLET, { _, arg -> arg.route.parent.toString() }, call)
```
iNikem, line 154: Should this set `http.route` attribute as well?

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorServerTracing.kt
```diff
@@ -0,0 +1,161 @@
...
+) {
+
+  class Configuration {
+    internal lateinit var openTelemetry: OpenTelemetry
+
+    internal var capturedHttpHeaders = CapturedHttpHeaders.server(Config.get())
+
+    internal val additionalExtractors = mutableListOf<AttributesExtractor<in ApplicationRequest, in ApplicationResponse>>()
+
+    internal var statusExtractor:
```
iNikem, line 39: I don't understand why is this a function and not just an instance of `SpanStatusExtractor`?

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorServerTracing.kt
anuraaga: Yeah I think so - do you think we should do it here or should ServerSpanName, or a similar helper, set them both?

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorServerTracing.kt
anuraaga: This is the same pattern as many of the tracing builders. It's because it's very common to want to customize one specific case and then delegate to default for others.

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorServerTracing.kt
iNikem: I believe we discussed this already somewhere... I don't remember our decision :(

mateuszrzeszutek: `http.route` is going to be set by `ServerSpanNaming` (well, once I find enough time to implement that...)

mateuszrzeszutek (approved):

iNikem (approved):

PR #5004, opened 2022-01-03
Fix multi-layer JDBC driver call depth calculation

marcobjorge:
The #2756 request introduced a check on the call depth of the JDBC instrumentation that limits tracing with multi-layered drivers. For example, with Apache Calcite, the Calcite JDBC driver is caught on the call stack and none of the underlying connections are traced.

For a better support of different use cases the check should be anchored to the specific connection instead of the generic JDBC interface.

### DIFF
--- a/instrumentation/jdbc/javaagent/build.gradle.kts
+++ b/instrumentation/jdbc/javaagent/build.gradle.kts
@@ -33,4 +33,7 @@ dependencies {
 
 tasks.withType<Test>().configureEach {
   jvmArgs("-Dotel.instrumentation.jdbc-datasource.enabled=true")
+  // this is needed to guarantee that if a layered JDBC driver is excluded then
+  // the underlying driver is still correctly instrumented
+  jvmArgs("-Dotel.javaagent.exclude-classes=io.opentelemetry.javaagent.instrumentation.jdbc.excluded.*")
 }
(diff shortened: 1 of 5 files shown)


iNikem:
@marcobjorge please sign CLA

marcobjorge:
@iNikem I'm on it, but I need to jump some hoops in my org first though (corporate contributor).

laurit:
@marcobjorge just curious, would it work if you'd use `-Dotel.javaagent.exclude-classes` to exclude calcite classes from instrumentation when running without this change?
Perhaps an alternative would be to fix #2756 in a slightly different way. Would it work if we surrounded https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/935cb88ff429f81fa7d70b6fff60ff8bc6fdf455/instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/JdbcUtils.java#L89 with a thread local (or `CallDepth`) and only suppressed statement instrumentation when the call originates from there? Then you wouldn't need to worry whether `statement.getConnection()` throws exception or returns a new proxy on each call.

marcobjorge:
@laurit I was **sure** it could not be done with `-Dotel.javaagent.exclude-classes`, but after your ask I set out to prove it and arrived at the conclusion **it can** and this PR is **not needed** (oopsy and thank you!).

TLDR: In my tests was just not excluding ALL Calcite `java.sql.{Driver,Connection,Statement,PreparedStatement}` in the classpath and the javaagent instrumentation was picking up something it shouldn't and therefore instrumenting the top level driver instead of the underlying drivers.

Issue #4848, opened 2021-12-08
Sporadic openj9 test failure


trask:
```
[otel.javaagent 2021-12-08 21:39:25:510 +0000] [main] DEBUG io.opentelemetry.javaagent.tooling.AgentInstaller$RedefinitionLoggingListener - Exception while retransforming 3 classes: [class java.util.concurrent.ForkJoinTask$AdaptedCallable, class java.util.concurrent.ForkJoinTask$AdaptedRunnable, class java.util.concurrent.ForkJoinTask$AdaptedRunnableAction]
[otel.javaagent 2021-12-08 21:39:25:510 +0000] [otel-javaagent-transform-safe-logger] DEBUG io.opentelemetry.javaagent.tooling.AgentInstaller$TransformLoggingListener - Transformed java.util.concurrent.ForkJoinTask$AdaptedRunnableAction -- null
java.lang.UnsupportedOperationException: class redefinition failed: attempted to change superclass or interfaces
	at java.instrument/sun.instrument.InstrumentationImpl.retransformClasses0(Native Method)
	at java.instrument/sun.instrument.InstrumentationImpl.retransformClasses(InstrumentationImpl.java:167)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at net.bytebuddy.utility.Invoker$Dispatcher.invoke(Unknown Source)
	at net.bytebuddy.utility.dispatcher.JavaDispatcher$Dispatcher$ForNonStaticMethod.invoke(JavaDispatcher.java:1013)
	at net.bytebuddy.utility.dispatcher.JavaDispatcher$ProxiedInvocationHandler.invoke(JavaDispatcher.java:1142)
	at net.bytebuddy.agent.builder.$Proxy18.retransformClasses(Unknown Source)
	at net.bytebuddy.agent.builder.AgentBuilder$RedefinitionStrategy$Collector$ForRetransformation.doApply(AgentBuilder.java:7982)
[...]

PR #5009, opened 2022-01-04
Fix redefinition failure on openj9

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4848 The redefinition failure in the linked case isn't what made the test fail, but as it seems to be impossible to tell what the actual cause is it is best to close it.
After hiding our interfaces from `Class.getInterfaces` virtual field implementation isn't able to tell that class has been transformed and thus skips transforming it which breaks redefinition as new bytes don't have the same interfaces as the redefined class.
Although this stack trace has been only observed on openj9 it could happen on hotspot too.


instrumentation/internal/internal-reflection/bootstrap/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/RealInterfaces.java
```diff
@@ -0,0 +1,36 @@
...
+   * interfaces that the class directly implements including the interfaces that we have added.
+   *
+   * @param clazz class or interface
+   * @return interfaces directly implemented by given class
+   */
+  public static Class<?>[] get(Class<?> clazz) {
+    // instrumentation calls RealInterfaces#set() from Class#getInterfaces()
+    Class<?>[] interfaces = clazz.getInterfaces();
+    Class<?>[] real = realInterfaces.get(clazz);
+    return real != null ? real : interfaces;
```
trask, line 30:
```suggestion
    Class<?>[] real = realInterfaces.get(clazz);
    return real != null ? real : clazz.getInterfaces();
```

instrumentation/internal/internal-reflection/bootstrap/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/RealInterfaces.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.opentelemetry.javaagent.instrumentation.internal.reflection;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+public final class RealInterfaces {
+  private static final Map<Class<?>, Class<?>[]> realInterfaces =
+      Collections.synchronizedMap(new WeakHashMap<>());
```
trask, line 14: I'm worried about the strong reference to the classes in the value array, can you add a comment here explaining why this is ok (if it is?)

trask (commented):
great find!

instrumentation/internal/internal-reflection/bootstrap/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/RealInterfaces.java
laurit: `clazz.getInterfaces()` needs to be called before reading from `realInterfaces` as the call to `clazz.getInterfaces()` trigger adding `clazz` to that map

trask: oh indeed! can you add this as a comment in the code?

laurit: there is `// instrumentation calls RealInterfaces#set() from Class#getInterfaces()` just before `Class<?>[] interfaces = clazz.getInterfaces();` should the comment be improved?

instrumentation/internal/internal-reflection/bootstrap/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/RealInterfaces.java
laurit: I initially thought that it would be ok because value doesn't reference key, but this logic doesn't work when classes are in the same class loader as value keeps the loader alive which prevents the key from being collected. I'll try to verify this tomorrow. It is probably easiest to change this to just keep the flag whether class has virtual fields or not instead of keeping the list of interfaces. Or if you prefer could also wrap all the interfaces with weakreferences.

trask: > It is probably easiest to change this to just keep the flag whether class has virtual fields or not instead of keeping the list of interfaces

👍

instrumentation/internal/internal-reflection/bootstrap/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/RealInterfaces.java
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * Returns the interfaces directly implemented by the given class or interface. We instrument
+   * Class#getInterfaces() to remove interfaces added by us. This method can be used to get the
+   * interfaces that the class directly implements including the interfaces that we have added.
+   *
+   * @param clazz class or interface
+   * @return interfaces directly implemented by given class
+   */
+  public static Class<?>[] get(Class<?> clazz) {
+    // instrumentation calls RealInterfaces#set() from Class#getInterfaces()
```
trask, line 27: oh got it, ya, how about
```suggestion
    // clazz.getInterfaces() needs to be called before reading from realInterfaces
    // as the call to clazz.getInterfaces() trigger adding clazz to that map
    // via instrumentation calling RealInterfaces#set() from Class#getInterfaces()
```

instrumentation/internal/internal-reflection/bootstrap/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/reflection/VirtualFieldDetector.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.internal.reflection;
```
mateuszrzeszutek, line 6: WDYT about putting this class in `javaagent-bootstrap` instead of a new bootstrap module? It's used by the `javaagent-tooling` module, so it makes sense for it to be in the `javaagent-bootstrap` module.

mateuszrzeszutek (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/VirtualFieldDetector.java
```diff
@@ -0,0 +1,41 @@
...
+
+package io.opentelemetry.javaagent.bootstrap;
+
+import io.opentelemetry.instrumentation.api.cache.Cache;
+import java.util.Arrays;
+
+/** Helper class for detecting whether given class has virtual fields. */
+public final class VirtualFieldDetector {
+
+  private static final Cache<Class<?>, Boolean> classesWithVirtualFields = Cache.weak();
```
trask, line 14: ClassValue?

trask (approved):

iNikem (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/VirtualFieldDetector.java
laurit: As ClassValue is not mutable, it works best if you can compute the value on first use. Here we can't really compute the value, so the best we could do would be to use a mutable object as value of ClassValue, initialize it to false, and update it when the actual value is known. With current solution we can ignore the classes that don't have virtual fields, with ClassValue we would probably end up initializing a ClassValue for these too.

trask: makes sense, thx!

PR #5014, opened 2022-01-05
Fix ktor testLatestDeps

trask:


anuraaga (approved):
Oops forgot to run that, thanks

Issue #592, opened 2020-06-28
Remove final from local variables


trask:
We decided against `final` on local variables recently, and updated the Intellij save actions doc recently (#512).

It would probably be good to go ahead and remove `final` from existing local variables so that it's not confusing which style we are following.

heyams:
Please assign this issue to me. Thanks.

PR #5016, opened 2022-01-05
Update errorprone

anuraaga:


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
```diff
@@ -192,7 +192,7 @@ private void setUrl(AttributeSetter setter, REQUEST request) {
       URI url = url(request);
       if (url != null) {
         netPeerAttributes.setNetPeer(setter, url.getHost(), null, url.getPort());
-        final URI sanitized;
+        URI sanitized;
```
iNikem, line 195: Why this is bad? If I want to guarantee that local variable is read-only, why is it bad to do so?

anuraaga: @iNikem Do you mean specifically about this case where the variable is declared separately from initialization or any usage of `final` on local variable?

iNikem: Any usage of `final` on local variable. I agree that our codebase probably overdo it, but I don't like the idea of forbidding it.

anuraaga: We decided against final on local variables a long time ago

#592

I think many of us agree with errorprone's sentiment that "it adds a fair amount of noise for minimal benefit."

https://errorprone.info/bugpattern/UnnecessaryFinal

I think it's worth being consistent on this and this prevents needing a yearly version of #592 going forward

iNikem: :( 😿

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
trask: we had carved out an exception for this particular case though:

> Local variables should only be declared final if they are not initialized inline (declaring these vars final can help prevent accidental double-initialization).

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/style-guideline.md#final-keyword-usage

While I think it's a nice exception, if errorprone can't differentiate, I'm ok with removing the exception.

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/HttpClientTracer.java
anuraaga: Ah forgot we documented it. Indeed I lean towards sticking with the error prone check to guarantee consistency instead.

Through this I learned that error prone actually tries to only use syntax in the rarer case - there is a `Var` check which will fail if the variable is assigned to multiple times, what this final was intended to avoid. For legitimate cases you add a `@Var` errorprone annotation. It's not clear to me yet if it's worth bringing in errorprone annotations just for this but could be a good approach - currently we have the Var check disabled.

trask:
merging before more conflicts

PR #5019, opened 2022-01-05
Capture servlet request parameters at the end of the request

laurit:
Move capturing servlet request parameters as span attributes from the beginning of the request to the end of the request. Reading request parameters forces parsing of the request body. When we do it at the beginning there is a risk that request will be parsed before request encoding is set which could result in incorrect parameter values.


trask (approved):
👍

I have one question, but it's on pre-existing code, so will explore that separately (#5021)

laurit:
@trask I added a comment, could you see if it helps.

trask:
> @trask I added a comment, could you see if it helps.

thx!

PR #5027, opened 2022-01-06
ktor net attribute extraction

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5022
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5023

ktor behaves slightly differently between versions so had to resort to regular expressions to figure out whether we got an ip address or hostname :(

### DIFF
--- a/instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/IpAddressUtil.kt
+++ b/instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/IpAddressUtil.kt
@@ -1,3 +1,8 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 package io.opentelemetry.instrumentation.ktor.v1_0
 
 import java.util.regex.Pattern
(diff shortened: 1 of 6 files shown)


instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/IpAddressUtil.kt
```diff
@@ -0,0 +1,28 @@
+package io.opentelemetry.instrumentation.ktor.v1_0
+
+import java.util.regex.Pattern
+
+// Source: Regular Expressions Cookbook 2nd edition - 8.17.
+// Matching IPv6 Addresses
+private val ipv6 = Pattern.compile( // Non Compressed
+  "^(?:(?:(?:[A-F0-9]{1,4}:){6}" // Compressed with at most 6 colons
```
laurit, line 8: copied from [JdbcConnectionUrlParser](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/9d68a9f004d0c4c729e3543eff7966fec6563e08/instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/JdbcConnectionUrlParser.java#L119)

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/KtorNetServerAttributesExtractor.kt
```diff
@@ -16,15 +15,23 @@ internal class KtorNetServerAttributesExtractor : NetServerAttributesExtractor<A
     return SemanticAttributes.NetTransportValues.IP_TCP
   }
 
-  override fun peerName(request: ApplicationRequest): String {
-    return request.origin.host
+  override fun peerName(request: ApplicationRequest): String? {
+    var remote = request.local.remoteHost
```
laurit, line 19: https://api.ktor.io/ktor-http/io.ktor.http/-request-connection-point/remote-host.html states that it is either client address or host name. What it returns seems to have changed between 1.0 and the latest + there are multiple implementations of `RequestConnectionPoint` interface that don't behave in the same way.

trask (approved):
👍

trask:
merging now to fix the build

instrumentation/ktor-1.0/library/src/main/kotlin/io/opentelemetry/instrumentation/ktor/v1_0/IpAddressUtil.kt
```diff
@@ -0,0 +1,36 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.ktor.v1_0
+
+import java.util.regex.Pattern
+
+// Source: Regular Expressions Cookbook 2nd edition - 8.17.
+// Matching IPv6 Addresses
+private val ipv6 = Pattern.compile( // Non Compressed
```
anuraaga, line 12: Just a pointer to Netty's isIpAddress methods

https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/NetUtil.java#L270

There're probably faster than regex, but also tons of code so not worth it but for reference in case it seems useful

trask: good to know, could be worth vendoring in to instrumentation-api at some point if we find it useful on more instrumentations

mateuszrzeszutek: Also: we already have a ipv6 parser in [`JdbcConnectionUrlParser`](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/jdbc/library/src/main/java/io/opentelemetry/instrumentation/jdbc/internal/JdbcConnectionUrlParser.java#L119), so there are now 2 instrumentations that would benefit from it.

PR #5030, opened 2022-01-07
NetClientAttributesAdapter - favor composition over inheritance

breedx-splk:
The goal here is to reduce the number of responsibilities in the `NetClientAttributesExtractor` (single responsibility principle) and to favor composition/delegation over inheritance. To do this, we extracted the abstract methods to an interface  named `NetClientAttributesAdapter`, which the `NetClientAtrributesExtractor` delegates to.  This allows it to now be made concrete (and final). The existing implementations are migrated to the new interface instead of extending `NetClientAttributesAdapter`. 

To reduce the scope of this change (which already touches a lot of code), this does not include a similar change to the `NetServerAttributesExtractor`, but I think we should have parity via a follow-up PR.


trask:
hey @breedx-splk! there was some related discussion in #3131

it didn't seem like having the instrumentation library deal with two types was worth it (e.g. in this case NetClientAttributesExtractor and NetClientAttributesAdapter), though @anuraaga [had a thought to mitigate that](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3131#issuecomment-850939172)

also, if we decide to do this for Net(Client|Server)AttributesExtractor, I assume we would do it for Http(Client|Server)AttributesExtractor, DbAttributesExtractor, CodeAttributesExtractor, RpcAttributesExtractor, and MessagingAttributesExtractor too?

breedx-splk:
> hey @breedx-splk! there was some related discussion in #3131

Thanks, I didn't realize this had come up before.

> it didn't seem like having the instrumentation library deal with two types was worth it (e.g. in this case NetClientAttributesExtractor and NetClientAttributesAdapter), though @anuraaga [had a thought to mitigate that](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3131#issuecomment-850939172)

I dunno, I think it's worth it. Maybe it's worth getting @anuraaga to chime in again.

> also, if we decide to do this for Net(Client|Server)AttributesExtractor, I assume we would do it for Http(Client|Server)AttributesExtractor, DbAttributesExtractor, CodeAttributesExtractor, RpcAttributesExtractor, and MessagingAttributesExtractor too?

I think definitely that would be the longer term goal....to be consistent in eliminating inheritance as an api mechanism.

trask:
@breedx-splk we discussed in yesterday's SIG and there is support for this change 👍

one suggestion is to rename `NetClientAttributesAdapter` to `NetClientAttributesGetter` (naming sort of matches `TextMapGetter`)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/PeerServiceAttributesExtractor.java
```diff
@@ -29,24 +29,24 @@
       Config.get().getMap("otel.instrumentation.common.peer-service-mapping", emptyMap());
 
   private final Map<String, String> peerServiceMapping;
-  private final NetClientAttributesExtractor<REQUEST, RESPONSE> netClientAttributesExtractor;
+  private final NetClientAttributesGetter<REQUEST, RESPONSE> attributesAdapter;
```
mateuszrzeszutek, line 32: You probably need to `s/Adapter/Getter/g` everywhere in this file.

instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/ApacheHttpClientSingletons.java
```diff
@@ -28,16 +29,17 @@
         HttpSpanNameExtractor.create(httpAttributesExtractor);
     SpanStatusExtractor<? super HttpMethod, ? super HttpMethod> spanStatusExtractor =
         HttpSpanStatusExtractor.create(httpAttributesExtractor);
-    ApacheHttpClientNetAttributesExtractor netAttributesExtractor =
-        new ApacheHttpClientNetAttributesExtractor();
-
+    ApacheHttpClientNetAttributesGetter netAttributesAdapter =
+        new ApacheHttpClientNetAttributesGetter();
+    NetClientAttributesExtractor<HttpMethod, HttpMethod> netAttributesExtractor =
+        NetClientAttributesExtractor.create(netAttributesAdapter);
```
mateuszrzeszutek, line 35: Nit: could this one be inlined?

instrumentation/lettuce/lettuce-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v4_0/LettuceSingletons.java
```diff
@@ -38,14 +39,14 @@
             .addAttributesExtractor(attributesExtractor)
             .newInstrumenter(SpanKindExtractor.alwaysClient());
 
-    LettuceConnectNetAttributesExtractor connectNetAttributesExtractor =
-        new LettuceConnectNetAttributesExtractor();
+    NetClientAttributesExtractor<RedisURI, Void> netClientAttributesExtractor =
+        NetClientAttributesExtractor.create(new LettuceConnectNetAttributesGetter());
```
mateuszrzeszutek, line 43: Can that `LettuceConnectNetAttributesGetter` instance be reused? The extractor itself can probably be just inlined.

mateuszrzeszutek (approved):
Nice 👍

instrumentation/apache-httpclient/apache-httpclient-2.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/apachehttpclient/v2_0/ApacheHttpClientSingletons.java
breedx-splk: sure, it's just a massive wall of text anyway..... 🙃

instrumentation/lettuce/lettuce-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/lettuce/v4_0/LettuceSingletons.java
breedx-splk: Yup! Good call.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/NetClientAttributesGetter.java
```diff
@@ -0,0 +1,31 @@
...
+package io.opentelemetry.instrumentation.api.instrumenter.net;
+
+import javax.annotation.Nullable;
+
+/**
+ * A type adapter for client-based network attributes. It adapts from a type-specific request and
+ * response into the 4 common network attribute values.
+ *
+ * <p>Instrumentation authors will create implementations of this interface for their specific
+ * library/framework. It will be used by the NetworkClientAttributesExtractor to obtain the various
```
trask, line 15:
```suggestion
 * library/framework. It will be used by the NetClientAttributesExtractor to obtain the various
```

trask (commented):
can you do one more pass through to change remaining adapter to getter terminology?

breedx-splk:
I think I got them all now.

trask:
go to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5030/files and search for `adapter`, I still see 96 matches

trask (approved):
there's a couple remaining references to "adapter" term, if you don't mind updating those too, then let's get this merged!

PR #5033, opened 2022-01-07
Spring RMI instrumentation

Samudraneel24:
Added instrumentation for Spring RMI.


settings.gradle.kts
```diff
@@ -367,6 +367,7 @@ include(":instrumentation:spring:spring-integration-4.1:library")
 include(":instrumentation:spring:spring-integration-4.1:testing")
 include(":instrumentation:spring:spring-kafka-2.7:javaagent")
 include(":instrumentation:spring:spring-rabbit-1.0:javaagent")
+include(":instrumentation:spring:spring-rmi:javaagent")
```
trask, line 370: the convention here is to add the min version that the instrumentation supports (the reason the rmi instrumentation has none is because it's supported on all Java versions that are supported by the javaagent)

```suggestion
include(":instrumentation:spring:spring-rmi-5.0:javaagent")
```

instrumentation/spring/spring-rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/server/ServerSingletons.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private static final Instrumenter<ClassAndMethod, Void> INSTRUMENTER;
+
+  static {
+    ServerAttributesExtractor attributesExtractor = new ServerAttributesExtractor();
+
+    INSTRUMENTER =
+        Instrumenter.<ClassAndMethod, Void>builder(
+                GlobalOpenTelemetry.get(),
+                "io.opentelemetry.springrmi",
```
trask, line 24: to match the module name
```suggestion
                "io.opentelemetry.spring-rmi-5.0",
```

instrumentation/spring/spring-rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/server/ServerInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
...
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class ServerInstrumentationModule extends InstrumentationModule {
+
+  public ServerInstrumentationModule() {
+    super("spring-rmi", "spring-rmi-server");
```
trask, line 19: the rmi instrumentation itself is slightly different from normal convention, here's a good example similar to this one: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/39d20b5fbc79ab38b7ce685fd5f2610e961ba1e6/instrumentation/spring/spring-webflux-5.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/webflux/client/WebfluxClientInstrumentationModule.java#L19

```suggestion
    super("spring-rmi", "spring-rmi-5.0", "spring-rmi-server");
```

instrumentation/spring/spring-rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/client/ClientSingletons.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private static final Instrumenter<Method, Void> INSTRUMENTER;
+
+  static {
+    ClientAttributesExtractor attributesExtractor = new ClientAttributesExtractor();
+
+    INSTRUMENTER =
+        Instrumenter.<Method, Void>builder(
+                GlobalOpenTelemetry.get(),
+                "io.opentelemetry.springrmi",
```
trask, line 24:
```suggestion
                "io.opentelemetry.spring-rmi-5.0",
```

instrumentation/spring/spring-rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/client/ClientInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
...
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class ClientInstrumentationModule extends InstrumentationModule {
+
+  public ClientInstrumentationModule() {
+    super("spring-rmi", "spring-rmi-client");
```
trask, line 19:
```suggestion
    super("spring-rmi", "spring-rmi-5.0", "spring-rmi-client");
```

instrumentation/spring/spring-rmi/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
+plugins {
+  id("otel.javaagent-instrumentation")
+}
+
+muzzle {
+  pass {
+    group.set("org.springframework")
+    module.set("spring-context")
+    versions.set("[5.0.4.RELEASE,5.4)")
```
trask, line 9: since there's no 5.4 yet, it's good to leave unbounded, that way CI will fail when 5.4 is released if our instrumentation/tests fail on it

also, can you briefly describe why it only works starting at 5.0.4?

```suggestion
    versions.set("[5.0.4.RELEASE,)")
```

instrumentation/spring/spring-rmi/javaagent/build.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+dependencies {
+  compileOnly("com.google.auto.value:auto-value-annotations")
+  annotationProcessor("com.google.auto.value:auto-value")
+
+  bootstrap(project(":instrumentation:rmi:bootstrap"))
+  testInstrumentation(project(":instrumentation:rmi:javaagent"))
+
+  library("org.springframework:spring-context:5.2.9.RELEASE")
```
trask, line 20: by default we run the tests against the "base" version, and [there's a separate task](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/contributing/running-tests.md#executing-tests-against-the-latest-versions-of-libraries-under-instrumentation) that runs the tests against the "latest" version that it can find in maven central
```suggestion
  library("org.springframework:spring-context:5.0.4.RELEASE")
```

trask (commented):
thx @Samudraneel24!

settings.gradle.kts
Samudraneel24: Got it, and thanks for the explanation. I didn't know of this convention.

instrumentation/spring/spring-rmi/javaagent/build.gradle.kts
Samudraneel24: Actually nothing significant, the application on which I was testing the instrumentation is using `spring-starter-web`, and I tested till it's version 2.0.0.RELEASE, which comes bundled with `spring-context` version 5.0.4.RELEASE.
Now I tested it with earlier versions and the instrumentation works with `spring-context` version 4.0.0.RELEASE.
I will put out a new commit with the minimum version changed to 4.

Samudraneel24:
@trask Thanks for the suggestions. I've fixed them and pushed a new commit but EasyCLA seems to be just stuck for me. Why could this be happening? I had already signed it earlier for this PR.

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/server/RemoteServerInstrumentation.java
```diff
@@ -28,7 +29,8 @@
 public class RemoteServerInstrumentation implements TypeInstrumentation {
   @Override
   public ElementMatcher<TypeDescription> typeMatcher() {
-    return implementsInterface(named("java.rmi.Remote"));
+    return implementsInterface(named("java.rmi.Remote"))
+        .and(not(nameStartsWith("org.springframework.remoting")));
```
mateuszrzeszutek, line 33: This may seem like a somewhat stupid question, but what's the reason for a separate spring-rmi instrumentations? Did it not work properly before? It looks like you have to exclude the spring RMI classes from the "standard" Java RMI instrumentation.

instrumentation/spring/spring-rmi-4.0/javaagent/src/test/groovy/SpringRmiTest.groovy
```diff
@@ -0,0 +1,163 @@
...
+          attributes {
+            "$SemanticAttributes.RPC_SYSTEM" "spring_rmi"
+            "$SemanticAttributes.RPC_SERVICE" "springrmi.app.SpringRmiGreeter"
+            "$SemanticAttributes.RPC_METHOD" "exceptional"
+          }
+        }
+        span(2) {
+          name "springrmi.app.SpringRmiGreeterImpl/exceptional"
+          kind SpanKind.SERVER
+          status ERROR
```
mateuszrzeszutek, line 144: Is this span a `childOf span(1)`? The spring-rmi client instrumentation does not seem to implement context propagation; does the context get properly passed to the server?

instrumentation/spring/spring-rmi-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/client/ClientInstrumentationModule.java
```diff
@@ -0,0 +1,26 @@
...
+import com.google.auto.service.AutoService;
+import io.opentelemetry.javaagent.extension.instrumentation.InstrumentationModule;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import java.util.List;
+
+@AutoService(InstrumentationModule.class)
+public class ClientInstrumentationModule extends InstrumentationModule {
+
+  public ClientInstrumentationModule() {
+    super("spring-rmi", "spring-rmi-4.0", "spring-rmi-client");
```
mateuszrzeszutek, line 19: I know that the base RMI instrumentation has several separate instrumentation modules for every part, but honestly I'm not sure whether this is a good idea in general - WDYT about keeping a single `InstrumentationModule` with all the instrumentations here?

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/server/RemoteServerInstrumentation.java
Samudraneel24: The standard Java RMI instrumentation didn't work properly for Spring RMI.
Client side spans don't get generated and in Server side span it instruments an Spring RMI internal class: `RmiInvocationHandler` because it extends `java.rmi.Remote`, which wasn't the original intent in the Java RMI instrumentation as there the user defined class extends it. Due to this it captures `rpc.service` and `rpc.method` incorrectly.
And to suppress this unwanted instrumentation I had to include that condition in Java RMI instrumentation.

instrumentation/spring/spring-rmi-4.0/javaagent/src/test/groovy/SpringRmiTest.groovy
Samudraneel24: Yes, it is a `childOf span(1)`. For context propagation I'm using the Java RMI context propagation module, works perfectly fine using that.

instrumentation/spring/spring-rmi-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/client/ClientInstrumentationModule.java
Samudraneel24: In other instrumentations I saw the Client and Server `InstrumentationModule` are usually kept separate, and there's merit in that. But we can also have all of them in a single module, I'm fine with both.

instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/server/RemoteServerInstrumentation.java
mateuszrzeszutek: Got it, thanks for the explanation!

instrumentation/spring/spring-rmi-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springrmi/client/ClientInstrumentationModule.java
mateuszrzeszutek: Usually we try to have a single `InstrumentationModule` for a single library that we instrument - this makes the runtime reference matching process a bit more efficient; and makes it harder to just partially instrument the code (usually resulting in invalid telemetry). There are some instrumentations that do not follow that rule, usually because they consist of features that need to be turned on/off separately -- but the general recommendation is to keep everything in a single `InstrumentationModule`.

Samudraneel24: Got it, thanks for the explanation.
I'll transfer all the instrumentation in a single `InstrumentationModule` and push a commit.

Samudraneel24:
Thanks for the suggestions @mateuszrzeszutek! I've made the changes.

mateuszrzeszutek (approved):
Thanks!

trask (approved):
👍

PR #5043, opened 2022-01-07
Rename some packages, move some classes

iNikem:
- `HttpStatusConverter` and related moved to `instrumenter.http` package
- `servlet` package renamed to `server` package as it better described it
- `ServerSpan` and `ServerSpanNaming` and related were moved to  `server` package


trask (approved):
👍

PR #5046, opened 2022-01-07
Allow running smoke tests locally on windows

trask:


smoke-tests/src/test/groovy/io/opentelemetry/smoketest/CrashEarlyJdk8Test.groovy
```diff
@@ -17,10 +16,14 @@ import org.testcontainers.utility.MountableFile
 import spock.lang.IgnoreIf
 import spock.lang.Specification
 
+import java.time.Duration
```
anuraaga, line 19: Formatting looks off

anuraaga (approved):

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/CrashEarlyJdk8Test.groovy
trask: I think this is the correct formatting (#1708)

smoke-tests/src/test/java/io/opentelemetry/smoketest/TestContainerManager.java
```diff
@@ -28,4 +28,9 @@
       String[] command);
 
   void stopTarget();
+
+  static boolean useLinuxContainers() {
```
iNikem, line 32: This method's result is always reverted, should we revert the method itself instead?

trask: makes sense 👍

PR #5054, opened 2022-01-09
Add level config to logback instrumentation

trask:
Adds an experimental config `otel.instrumentation.logback-appender.experimental.level` that serves as a minimum threshold for the logs that are captured.

The current default in this PR is `ALL` (matching existing behavior), but I could also see `INFO` being a good default.

I will add similar config to log4j and java.util.logging instrumentation if this is approved.

### DIFF
--- a/instrumentation/logback/logback-appender-1.0/javaagent/src/test/resources/logback-test.xml
+++ b/instrumentation/logback/logback-appender-1.0/javaagent/src/test/resources/logback-test.xml
@@ -12,10 +12,11 @@
   <logger name="abc" level="INFO"/>
   <logger name="def" level="WARN"/>
 
+  <logger name="DebugLogger" level="DEBUG"/>
+
   <root level="INFO">
     <appender-ref ref="console"/>
   </root>
 
-  <logger name="org.testcontainers" level="WARN"/>
   <logger name="io.opentelemetry.testing.internal" level="WARN"/>
 </configuration>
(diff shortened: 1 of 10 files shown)


instrumentation/logback/logback-appender-1.0/javaagent/src/test/resources/logback-test.xml
```diff
@@ -12,10 +12,11 @@
   <logger name="abc" level="INFO"/>
   <logger name="def" level="WARN"/>
 
+  <logger name="DebugLogger" level="DEBUG"/>
+
   <root level="INFO">
     <appender-ref ref="console"/>
   </root>
 
-  <logger name="org.testcontainers" level="WARN"/>
```
trask, line 19: unrelated to this PR - there's no testcontainers logging in this module, not sure how this snuck in there originally

testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/AgentTestingLogsCustomizer.java
```diff
@@ -24,8 +24,7 @@ public void beforeAgent(
     SdkLogEmitterProvider logEmitterProvider =
         SdkLogEmitterProvider.builder()
             .setResource(autoConfiguredOpenTelemetrySdk.getResource())
-            .addLogProcessor(
-                BatchLogProcessor.builder(AgentTestingExporterFactory.logExporter).build())
+            .addLogProcessor(SimpleLogProcessor.create(AgentTestingExporterFactory.logExporter))
```
trask, line 27: unrelated to this PR - testing should use simple log processor so we don't have to sleep

trask:
closing, I think I'll implement this as part of custom export pipeline for now...

PR #5058, opened 2022-01-09
Add spotless to standalone subprojects

trask:
Adds spotless to these standalone projects:

* benchmark-overhead
* examples/distro
* examples/extension
* smoke-tests/images/*

Split out into separate commits (adding configuration vs applying spotless) for easier review.

### DIFF
--- a/benchmark-overhead/build.gradle.kts
+++ b/benchmark-overhead/build.gradle.kts
@@ -1,5 +1,14 @@
 plugins {
   id("java")
+  id("com.diffplug.spotless") version "6.1.2"
+}
+
+spotless {
+  java {
+    googleJavaFormat()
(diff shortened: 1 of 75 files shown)


iNikem:
Should we publish our `otel.spotless-conventions` and just reuse it instead of duplicating spotless configuration in every module?

mateuszrzeszutek (approved):

mateuszrzeszutek:
> Should we publish our `otel.spotless-conventions` and just reuse it instead of duplicating spotless configuration in every module?

We use the same spotless config in the SDK repo, don't we? We could reuse the same published plugin in all our projects.

trask:
> Should we publish our `otel.spotless-conventions` and just reuse it instead of duplicating spotless configuration in every module?

I opened https://github.com/open-telemetry/opentelemetry-java/issues/4076 to discuss

Issue #3478, opened 2021-07-02
Akka http server span names are always `akka.request`


mattypb:
**Describe the bug**
I've added the Open Telemetry java instrumentation to a Scala Akka service. Traces are created when the service's api's are hit, but no matter which api is hit they all have a span name of `akka.request`.

Looking at the Akka http server instrumentation, this is appears to be the current expected behaviour - [AkkaHttpServerInstrumentationTest](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/akka-http-10.0/javaagent/src/test/groovy/AkkaHttpServerInstrumentationTest.groovy#L24). However some other java instrumentations use the api path as the span name, for example Tomcat 7.0 - [TomcatHandlerTest](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/tomcat/tomcat-7.0/javaagent/src/test/groovy/io/opentelemetry/javaagent/instrumentation/tomcat/v7_0/TomcatHandlerTest.groovy#L35-L40).

Why does the Akka server instrumentation have `akka.request` hard coded for the span name instead of using the api path? Will this change in the future?

**What did you expect to see?**
Hitting an api on my service, e.g. `GET /v1/{user}`, would generate a trace with span name `/v1/{user}`

**What did you see instead?**
Span name of `akka.request`

**What version are you using?**
1.3.0

**Environment**
Scala: 2.13.5
Akka-Http: 10.2.4
Compiler: AdoptOpenJDK 11.0.10
OS: Mac 10.15.7
Runtime (if different from JDK above): Oracle  11.0.11+9-LTS-194
OS (if different from OS compiled on): RHEL 7.1

nibhuran:
@mattypb , @trask Can I take a crack at this issue.

trask:
> Can I take a crack at this issue.

sure!

mcmho:
@trask, I am looking into this a bit.  Does it make sense to add a SpanNameExtractor for this case and pass it to the Instrumenter?

trask:
> @trask, I am looking into this a bit. Does it make sense to add a SpanNameExtractor for this case and pass it to the Instrumenter?

yes, that makes sense 👍

mcmho:
@trask, @mateuszrzeszutek,   I see that there are many span name extractors having null implementation for route().   
```
  @Override
  @Nullable
  protected String route(Request request) {
    return null;
  }
```

[...]

PR #5065, opened 2022-01-10
Akka http server span names are always akka.request #3478

mcmho:
Open Telemetry java instrumentation to a Scala Akka service.  Akka server span name is hardcoded with akka.request.
Make sure AkkaHttpServerInstrumentationTest is retrofitted after the fix.   also double check TomcatHandlerTest.
Expected to see  GET /v1/{user} would generate a span name /v1/{user}


mcmho:
unintentional

PR #5067, opened 2022-01-11
Predeclare spotless deps to allow high concurrency

anuraaga:
I haven't been able to succeed with recent spotless versions on a 16-core machine, but it seems the very latest version added this workaround which causes tools to only be resolved in a single, root project, instead of every project causing deadlocks.

### DIFF
--- a/conventions/src/main/kotlin/otel.spotless-conventions.gradle.kts
+++ b/conventions/src/main/kotlin/otel.spotless-conventions.gradle.kts
@@ -1,3 +1,5 @@
+import com.diffplug.gradle.spotless.SpotlessExtension
+
 plugins {
   id("com.diffplug.spotless")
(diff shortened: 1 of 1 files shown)


conventions/src/main/kotlin/otel.spotless-conventions.gradle.kts
```diff
@@ -46,3 +48,24 @@ spotless {
     endWithNewline()
   }
 }
+
+if (project == rootProject) {
+  spotless {
+    predeclareDeps()
+  }
```
trask, line 55: since it's a workaround(?), a brief comment here would be nice

trask (approved):

mateuszrzeszutek (approved):

Issue #5061, opened 2022-01-10
Jedis and Redisson instrumentation recording duration equals 0


ralphgj:
**Describe the bug**
The duration of spans created by jedis or redisson instrumentation equals 0, it seems incorrectly.
![image](https://user-images.githubusercontent.com/7887053/148730721-38c34859-7cd7-49a2-967e-b01f6829f4db.png)

**Steps to reproduce**
Use the latest agent and observe the spans created by jedis or redisson instrumentation

**What did you expect to see?**
Jedis or redisson instrumentation recording duration with the actual value.

**What did you see instead?**
Jedis or redisson instrumentation recording duration with 0.

**What version are you using?**
v1.9.1

laurit:
I wasn't able to reproduce this with test in this repository. Both redisson and jedis seem to report a non-zero duration.

ralphgj:
> They both do not capture the whole execution though: Jedis instrumentations have a comment that explicitly says that only sending the request is instrumented, processing/getting the response is not; and it looks like the instrumented Redisson method returns a netty `ChannelFuture` that we completely ignore and finish the span on return, not when the future completes. So I guess that a zero duration is quite possible with both of these.

Thank you for your reply, I read the code and that you said above should be the root cause. So would you have some plan or proposals to resolve this problem?

mateuszrzeszutek:
> Thank you for your reply, I read the code and that you said above should be the root cause. So would you have some plan or proposals to resolve this problem?

Redisson was quite simple to fix, I've fixed it here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5061

And I just noticed that you've implemented the fix for jedis, that's great, thanks! Looks like that should cover the whole issue.

ralphgj:
> > Thank you for your reply, I read the code and that you said above should be the root cause. So would you have some plan or proposals to resolve this problem?
> 
> Redisson was quite simple to fix, I've fixed it here: #5073
> 
> And I just noticed that you've implemented the fix for jedis, that's great, thanks! Looks like that should cover the whole issue.

Yes, thank you for your help.

trask:
thx for reporting this @ralphgj!

I believe this is closed now by #5073 and #5256

PR #5072, opened 2022-01-11
Correct jedis instrumentation span's duration

ralphgj:
Try to fix the problem related jedis span's duration in issue #5061


trask (commented):
thx @ralphgj!

how did you decide which methods to include/exclude in `getJedisMethodMatcher()`?

ralphgj:
> thx @ralphgj!
> 
> how did you decide which methods to include/exclude in `getJedisMethodMatcher()`?

Now I just choose all the public methods in Jedis class except the ping.  Maybe I should choose the methods that only match in `https://redis.io/commands`. Could you give me some suggestions？

instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisMethodInstrumentation.java
```diff
@@ -8,53 +8,46 @@
...
   public void transform(TypeTransformer transformer) {
     transformer.applyAdviceToMethod(
-        isMethod()
-            .and(named("sendCommand"))
-            .and(takesArguments(2))
-            .and(takesArgument(0, named("redis.clients.jedis.commands.ProtocolCommand")))
-            .and(takesArgument(1, is(byte[][].class))),
-        this.getClass().getName() + "$SendCommandAdvice");
-    // FIXME: This instrumentation only incorporates sending the command, not processing the result.
+        isMethod().and(JedisMethodMatch.getJedisMethodMatcher()),
```
trask, line 33: got it, you should be able to do this then:

```suggestion
        isMethod().and(isPublic()).and(not(named("ping"))),
```

ralphgj: thx, I will resubmit it.

anuraaga:
Just for reference, I noticed opentracing does create a span for `ping`

https://github.com/opentracing-contrib/java-redis-client/blob/master/opentracing-redis-jedis/src/main/java/redis/clients/jedis/TracingJedisWrapper.java#L2117

I do tend to agree that it's probably too noisy and ok to skip but wanted to share to confirm.

trask:
@ralphgj can you check if the previous instrumentation (prior to your PR) captured `ping`?

ralphgj:
> @ralphgj can you check if the previous instrumentation (prior to your PR) captured `ping`?

Yes , it did capturing `ping`.

instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisMethodInstrumentation.java
ralphgj: I noticed jedis in 3.8.0 version added some new public methods like `sendCommand`, `xinfoConsumers`, etc. Maybe the specified method name is better.

trask:
cool, I'd recommend continuing to capture `ping` then, if it's too noisy let's discuss/address that separately

ralphgj:
I resubmit this pr and put `ping` method back. And I excluded the public methods with `close` and `setDataSource` for they are not the command methods. 
In consider of Jedis would add new public methods in the latest version, I prefer to use the specified method name to instrument. 
Would you think this proposal is ok? @trask @anuraaga

instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisMethodInstrumentation.java
```diff
@@ -8,53 +8,46 @@
...
-import redis.clients.jedis.Connection;
-import redis.clients.jedis.commands.ProtocolCommand;
+import redis.clients.jedis.Jedis;
 
-public class JedisConnectionInstrumentation implements TypeInstrumentation {
+public class JedisMethodInstrumentation implements TypeInstrumentation {
   @Override
   public ElementMatcher<TypeDescription> typeMatcher() {
-    return named("redis.clients.jedis.Connection");
+    return named("redis.clients.jedis.Jedis");
```
laurit, line 27: Not all commands seem to be handled in `Jedis`, for example there is `slaveof` is in `BinaryJedis` which is the super class of `Jedis`. In Jedis 4 `JedisSharding` and `JedisCluster` don't delegate to `Jedis` so there will be more classes to instrument.

instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisMethodInstrumentation.java
```diff
@@ -8,53 +8,46 @@
...
+        @Advice.This Jedis jedis,
+        @Advice.AllArguments Object[] args,
+        @Advice.Origin Method method,
         @Advice.Local("otelJedisRequest") JedisRequest request,
         @Advice.Local("otelContext") Context context,
         @Advice.Local("otelScope") Scope scope) {
       Context parentContext = currentContext();
-      request = JedisRequest.create(connection, command, asList(args));
+      request =
+          JedisRequest.create(jedis.getClient(), method.getName().toUpperCase(), asList(args));
```
laurit, line 50: The problem with using method name here is that method names don't always match command names. For example method `pubsubNumPat` does not exactly match a command (command is `PUBSUB`). Not getting the command right will mess with `RedisCommandSanitizer`.
I guess we could try to find the methods that don't match a command and add a mapping for them. It should be possible to add a test for it by comparing method names to `redis.clients.jedis.Protocol.Command` enum.

instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisMethodMatch.java
```diff
@@ -0,0 +1,204 @@
...
+package io.opentelemetry.javaagent.instrumentation.jedis.v3_0;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+import net.bytebuddy.description.method.MethodDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class JedisMethodMatch {
+
+  public static ElementMatcher<MethodDescription> getJedisMethodMatcher() {
```
laurit, line 15: If the goal is to get all methods that correspond to jedis commands then could also experiment with something like
```
        isMethod()
            .and(isPublic())
            .and(hasSuperMethod(methodIsDeclaredByType(nameEndsWith("Commands"))))
```

ralphgj: > If the goal is to get all methods that correspond to jedis commands then could also experiment with something like
> 
> ```
>         isMethod()
>             .and(isPublic())
>             .and(hasSuperMethod(methodIsDeclaredByType(nameEndsWith("Commands"))))
> ```

Hi, @laurit I used the code above and it seems not working with `slaveOfNoOne` and `select` methods that inherited from `BinaryJedis `.

instrumentation/jedis/jedis-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v3_0/JedisMethodInstrumentation.java
ralphgj: Hi, @laurit thanks for your suggestions above. I am trying to rewrite the code follow your comments. And now I will resubmit it but it is still not finished. There are two problems in the implementation. Could you give me some more suggestions?
1. the methods in `Jedis` inherited from `SentinelCommands` and `ClusterCommands` lack of some tests. It seems need a redis cluster or master/slave test containers and I have no idea with it now.
2. the instrument seems not working with methods like `slaveofNoOne` or `select` in `Jedis` that inherited from `BinaryJedis`. I added two tests but they were not passed.

laurit:
@ralphgj Here is my attempt to tackle this https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5256

trask:
Closing as #5256 has been merged.

@ralphgj thx for helping us explore options on this tricky PR!

PR #5073, opened 2022-01-11
End redisson span when the operation actually ends

mateuszrzeszutek:
Part of https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5061


laurit (approved):

trask (approved):
👍

PR #5080, opened 2022-01-11
Move appender api to internal

### LINKED ISSUE #5066: Hide instrumentation appender api/sdk
Discussed in #4935 and elsewhere

TL;DR there is no desire to create anything public that could be mistaken for a log api

Proposal to rename modules:
* instrumentation-api-appender -> instrumentation-appender-api-internal
* instrumentation-sdk-appender -> instrumentation-appender-sdk-internal

and packages:
* io.opentelemetry.instrumentation.api.appender -> io.opentelemetry.instrumentation.appender.api.internal
* io.opentelemetry.instrumentation.sdk.appender -> io.opentelemetry.instrumentation.appender.sdk.internal

trask:
Resolves #5066

Renames in the PR:
* instrumentation-api-appender -> instrumentation-appender-api-internal
* instrumentation-sdk-appender -> instrumentation-appender-sdk-internal
* io.opentelemetry.instrumentation.api.appender -> io.opentelemetry.instrumentation.api.appender.internal
  * keeping "io.opentelemetry.instrumentation.api" prefix so that existing `HelperClassPredicate` understands that these are "bootstrap classes" (alternatively could add additional condition to `HelperClassPredicate`)
* io.opentelemetry.instrumentation.sdk.appender -> io.opentelemetry.instrumentation.appender.internal

Changes to library instrumentation (logback and log4j2 appender):
* Users now do not need to (and should not) interact with `instrumentation-appender-api-internal` or `instrumentation-appender-sdk-internal`
* Logback and log4j2 each have their own "global" SdkLogEmitterProvider. A single global could be used but then you can't set it on both without getting an exception that it was set already (alternatively could relax the condition that doesn't allow updating the global)


instrumentation-appender-api-internal/src/main/java/io/opentelemetry/instrumentation/api/appender/internal/AgentLogEmitterProvider.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.api.appender.internal;
```
anuraaga, line 6: Just confirming this makes sense in instrumentation-api and not javaagent-instrumentation-api?

anuraaga (approved):

instrumentation-appender-api-internal/src/main/java/io/opentelemetry/instrumentation/api/appender/internal/AgentLogEmitterProvider.java
trask: good idea, done 👍

mateuszrzeszutek (approved):

PR #5083, opened 2022-01-12
Update dependencies

anuraaga:


settings.gradle.kts
```diff
@@ -1,6 +1,6 @@
 pluginManagement {
   plugins {
-    id("com.github.ben-manes.versions") version "0.39.0"
+    id("com.github.ben-manes.versions") version "0.41.0"
```
laurit, line 3: I suspect that we aren't using anything from ben-manes any more

laurit (approved):

settings.gradle.kts
anuraaga: This isn't caffeine, it's the versions plugin which gives `:dependencyUpdates` task :)

mateuszrzeszutek (approved):

iNikem (approved):

Issue #5048, opened 2022-01-07
Add smoke test for logging signal


trask:
Now that we are updated to SDK 1.10.0, and have logging exporter autoconfiguration (https://github.com/open-telemetry/opentelemetry-java/pull/4015), we should add smoke test for logging signal to make sure that it's all working together.

breedx-splk:
I can take this.

breedx-splk:
Just to make sure we're on the same page @trask  -- is our expectation right now that we should be able to just set `OTEL_LOGS_EXPORTER=otlp` and any app that does logging with log4j or logback should start exporting logs? Is there an additional step of hooking up a specific exporter?

Do we think we can get away with (re)using one of the existing smoke test app images? It looks like `grpc` uses log4j and `play` uses logback...

trask:
> is our expectation right now that we should be able to just set `OTEL_LOGS_EXPORTER=otlp` and any app that does logging with log4j or logback should start exporting logs?

yes, I think (hope) so 👍

> Do we think we can get away with (re)using one of the existing smoke test app images? It looks like `grpc` uses log4j and `play` uses logback...

I think the spring-boot smoke test is a good catch-all. We only need a smoke test for one of the logging instrumentations (just to make sure the signal hookup is working, the instrumentation-specific tests should cover the rest).

breedx-splk:
> I think the spring-boot smoke test is a good catch-all. We only need a smoke test for one of the logging instrumentations (just to make sure the signal hookup is working, the instrumentation-specific tests should cover the rest).

Yup, and spring-boot smoke test is the one I used first but didn't get logs out. 🤔 

I'll keep after it.

PR #5088, opened 2022-01-12
Connect AgentLogEmitterProvider global during OpenTelemetryInstaller 

breedx-splk:
I was having some challenges around #5048 related to this.

Instrumentation uses the global holder for the log emitter. The `LogEmitterProvider` from the autoconfigured SDK was not being stitched to the global holder during install time, so instrumentations were not yet able to actually emit logs. This remedies that.

I confirmed that this this change allows logs to be exported (when the logs exporter is set).

Thanks to @mateuszrzeszutek for help spelunking to find this disconnect.

### DIFF
--- a/instrumentation/java-util-logging/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jul/JavaUtilLoggingHelper.java
+++ b/instrumentation/java-util-logging/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jul/JavaUtilLoggingHelper.java
@@ -31,11 +31,12 @@ public final class JavaUtilLoggingHelper {
       return;
     }
 
+    String instrumentationName = logRecord.getLoggerName();
+    if (instrumentationName == null || instrumentationName.isEmpty()) {
+      instrumentationName = "ROOT";
+    }
     LogBuilder builder =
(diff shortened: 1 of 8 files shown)


trask (approved):
👍

trask:
@breedx-splk I pushed a commit 🤞

breedx-splk:
> @breedx-splk I pushed a commit 🤞

Excellent, thanks!

anuraaga (approved):

trask:
payara smoke test has failed a couple of times, the log file is 600mb full of:

```
    App - STDERR: [#|2022-01-13T00:33:44.512+0000|SEVERE|Payara 5.2020.6||_ThreadID=30;_ThreadName=Logging output pump;_TimeMillis=1642034024512;_LevelValue=1000;|
    App - STDERR:   [otel.javaagent 2022-01-13 00:33:44:512 +0000] [Logging output pump] DEBUG io.opentelemetry.sdk.logs.SdkLogEmitterProvider - LogEmitter requested without instrumentation name.|#]
```

I'm going to push a commit to default the instrumentation name to `ROOT` when the logger name is null/empty.

mateuszrzeszutek (approved):

PR #5089, opened 2022-01-12
Apply gradle github action everywhere

trask:
Follow-up to #4887

Also adds gradle wrapper to `gradle-plugins` for consistency since all other standalone gradle modules have their own wrappers.


trask:
targeting this to 1.10.0 milestone so I can test out the changes to release.yml while still fresh in my mind

anuraaga (approved):

PR #5091, opened 2022-01-13
Update changelog for 1.10.0

trask:
~(depends on #5090)~


iNikem:
For some reason I feel that we have to call out 2 things specifically on the top of the changelog with a little bit more explanation:

- That Tracers and the whole package is deprecated and will be removed in 1.11 or 1.12
- What all these log libraries instrumentations mean. This is quite confusing atm, I think

CHANGELOG.md
```diff
@@ -1,5 +1,126 @@
...
+### 🌟 New library instrumentation
+
+- Logback appender
+  ([#4984](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4984))
+- Log4j 2.x appender
+  ([#4375](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4375),
+   [#4907](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4907))
+- Ktor instrumentation
+  ([#4983](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4983))
+
```
mateuszrzeszutek, line 32: There's a new micrometer library instrumentation too

mateuszrzeszutek:
I would probably be worth mentioning explicitly that the new Micrometer instrumentation will now capture metrics already defined in the instrumented application (e.g. some libraries have built-in micrometer metrics, like spring boot for example) and send them out through the agent metrics pipeline.

trask:
> For some reason I feel that we have to call out 2 things specifically on the top of the changelog with a little bit more explanation:
> 
> * That Tracers and the whole package is deprecated and will be removed in 1.11 or 1.12

I added this at the top of the release notes draft as we discussed in SIG meeting today.

> * What all these log libraries instrumentations mean. This is quite confusing atm, I think

I added some clarification in the change log, let me know if this helps.

trask:
> I would probably be worth mentioning explicitly that the new Micrometer instrumentation will now capture metrics already defined in the instrumented application (e.g. some libraries have built-in micrometer metrics, like spring boot for example) and send them out through the agent metrics pipeline.

I added a bit of clarification to the change log, and added a more explicit note about this at the top of the release notes draft.

PR #5100, opened 2022-01-13
Remove unnecessary outer micrometer directory

trask:


anuraaga (approved):

mateuszrzeszutek:
Adding that one directory was intentional - the Splunk distro has a [Micrometer 1.3](https://github.com/signalfx/splunk-otel-java/tree/main/instrumentation/micrometer-1.3) instrumentation that covers [1.3, 1.5) (since there were some incompatible API changes in 1.5). I was planning to extract the common parts and add the 1.3 instrumentation a bit later, after the release.

anuraaga (changes requested):
Thanks for the context @mateuszrzeszutek

Issue #5103, opened 2022-01-13
Sporadic micrometer library test failure


trask:
I saw this on the original micrometer library instrumentation PR: https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/4797134585

And again on next PR: https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/4798756656

Not sure what's going on with the lack of details about the error...

```
> Task :instrumentation:micrometer-1.5:library:compileTestJava
> Task :instrumentation:micrometer-1.5:library:compileTestGroovy NO-SOURCE
> Task :instrumentation:micrometer-1.5:library:processTestResources NO-SOURCE
> Task :instrumentation:micrometer-1.5:library:testClasses
> Task :instrumentation:micrometer-1.5:library:test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':instrumentation:micrometer-1.5:library:test'.
> java.lang.NullPointerException (no error message)

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3m 20s
408 actionable tasks: 104 executed, 304 from cache

A build scan cannot be produced as an error occurred gathering build data.
Please report this problem via https://gradle.com/help/plugin and include the following via copy/paste:

----------
Gradle version: 7.3.3
Plugin version: 3.8

java.lang.NullPointerException
	at com.gradle.scan.plugin.internal.i.g.b(SourceFile:61)
	at com.gradle.scan.plugin.internal.b.t.b.a(SourceFile:74)
	at com.gradle.scan.plugin.internal.b.t.b.a(SourceFile:47)
	at com.gradle.scan.plugin.internal.b.h.a.a(SourceFile:92)
	at com.gradle.scan.plugin.internal.b.h.a.f(SourceFile:79)
	at com.gradle.scan.plugin.internal.b.h.a.g(SourceFile:103)
```

trask:
enabling `--debug` in #5104 printed out the real error:

```
2022-01-13T05:59:46.5141084Z 2022-01-13T05:59:46.481+0000 [DEBUG] [TestEventLogger] FunctionTimerTest > functionTimersWithSameNameAndDifferentTags() FAILED
[...]

Issue #5051, opened 2022-01-08
Sporadic failure on openj9 Java 8


trask:
Seen this a couple of times on openj9, Java 8, specifically on `:instrumentation:log4j:log4j-appender-2.16:library:test`

I'm not sure if the NullPointerException is from the gradle scan plugin, or from the test...

Links to failures
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/4746375721?check_suite_focus=true
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/4747895431?check_suite_focus=true

```
* What went wrong:
Execution failed for task ':instrumentation:log4j:log4j-appender-2.16:library:test'.
> java.lang.NullPointerException (no error message)

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1h 26m 27s

A build scan cannot be produced as an error occurred gathering build data.
Please report this problem via https://gradle.com/help/plugin and include the following via copy/paste:

----------
Gradle version: 7.3.3
Plugin version: 3.8

java.lang.NullPointerException
	at com.gradle.scan.plugin.internal.i.g.b(SourceFile:61)
	at com.gradle.scan.plugin.internal.b.t.b.a(SourceFile:74)
	at com.gradle.scan.plugin.internal.b.t.b.a(SourceFile:47)
	at com.gradle.scan.plugin.internal.b.h.a.a(SourceFile:92)
	at com.gradle.scan.plugin.internal.b.h.a.f(SourceFile:79)
	at com.gradle.scan.plugin.internal.b.h.a.a(SourceFile:41)
	at com.gradle.scan.plugin.internal.b.v.g$b.a(SourceFile:162)
	at com.gradle.scan.plugin.internal.b.v.g$b.a(SourceFile:105)
	at com.gradle.scan.plugin.internal.b.v.g.a(SourceFile:42)
	at com.gradle.scan.plugin.internal.l.a.f.a(SourceFile:12)
	at com.gradle.scan.plugin.internal.l.a$c.finished(SourceFile:154)
	at com.gradle.scan.plugin.internal.l.a.a(SourceFile:65)
	at com.gradle.scan.plugin.internal.l.l.a(SourceFile:56)
	at com.gradle.scan.plugin.internal.l.c.a(SourceFile:98)
[...]

PR #5106, opened 2022-01-13
Fix flaky micrometer tests

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5103

This PR presents yet another reason why we'd like to be able to use multiple callbacks in async instruments (https://github.com/open-telemetry/opentelemetry-specification/issues/2249) -- turns out the library instrumentation tests were randomly failing because each of them created a new `OpenTelemetryMeterRegistry`, and they did not share state - and the `OpenTelemetry` used in library tests is shared between all test classes.

### DIFF
--- a/instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AsyncInstrumentRegistry.java
+++ b/instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AsyncInstrumentRegistry.java
@@ -24,7 +24,10 @@ import javax.annotation.Nullable;
 // TODO: refactor this class, there's too much copy-paste here
 final class AsyncInstrumentRegistry {
 
-  private final Meter meter;
+  // using a weak ref so that the AsyncInstrumentRegistry (which is stored in a static maps) does
+  // not hold strong references to Meter (and thus make it impossible to collect Meter garbage).
+  // in practice this should never return null
+  private final WeakReference<Meter> meter;
 
   // we're always locking lock on the whole instrument map; the add/remove methods aren't called
   // that often, so it's probably better to opt for correctness in that case - there is a small
@@ -45,7 +48,7 @@ final class AsyncInstrumentRegistry {
   private final Map<String, DoubleMeasurementsRecorder> upDownDoubleCounters = new HashMap<>();
 
   AsyncInstrumentRegistry(Meter meter) {
-    this.meter = meter;
+    this.meter = new WeakReference<>(meter);
   }
 
   <T> AsyncMeasurementHandle buildGauge(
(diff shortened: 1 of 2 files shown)


mateuszrzeszutek:
Hmm, apparently that's still not it, still getting these
```
java.lang.NullPointerException
	at com.gradle.scan.plugin.internal.i.g.b(SourceFile:61)
```

instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AsyncInstrumentRegistry.java
```diff
@@ -21,10 +21,15 @@
 import java.util.function.ToLongFunction;
 import javax.annotation.Nullable;
 
+// lalala test
```
anuraaga, line 24: fafafa

anuraaga (approved):

mateuszrzeszutek:
Okay, I think these cryptic NPEs I observed earlier may be related to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5051, not micrometer tests (_probably_), so I'm reopening it.

instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AsyncInstrumentRegistry.java
```diff
@@ -24,7 +24,10 @@
 // TODO: refactor this class, there's too much copy-paste here
 final class AsyncInstrumentRegistry {
 
-  private final Meter meter;
+  // using a weak ref so that the AsyncInstrumentRegistry (which is stored in a static maps) does
+  // not hold strong references to Meter (and thus make it impossible to collect Meter garbage).
+  // in practice this should never return null
```
trask, line 29: it's not obvious to me why `Meter` can't be GC'd before `AsyncInstrumentRegistry`, can you explain briefly, or what do you think about using weak value in the cache instead `Cache<Meter, WeakReference<AsyncInstrumentRegistry>>`?

mateuszrzeszutek: It can't be GC'd because the `OpenTelemetryMeterRegistry` maintains a strong reference to both `Meter` and `AsyncInstrumentRegistry` -- so if the registry instance is collected then the registry cannot possibly be used anymore. I'll add a comment.

> or what do you think about using weak value in the cache instead `Cache<Meter, WeakReference<AsyncInstrumentRegistry>>`?

Hmm, in case the meter registry is GC'd the `WeakReference<AsyncInstrumentRegistry>` could get nulled out even if the `Meter` is not GC'd yet (because the `OpenTelemetry` instance is still used) - this hypothetical scenario would break the async instruments, cause if you create a new `OpenTelemetryMeterRegistry` for the same `OpenTelemetry`/`Meter` the async instrument registry will get cleared out.

trask: oh, got it 👍

> I'll add a comment

❤️

trask (approved):

PR #5108, opened 2022-01-13
Add smoke test for logs signal

### LINKED ISSUE #5048: Add smoke test for logging signal

breedx-splk:
Resolves #5048 

Intentionally left very simple. We can add additional scenarios/run configs as we see fit. Might also be nice to make some additional assertions around the data contents, especially the Resource.

### DIFF
--- a/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LogsSmokeTest.groovy
+++ b/smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LogsSmokeTest.groovy
@@ -15,9 +15,8 @@ import static java.util.stream.Collectors.toList
 @IgnoreIf({ useWindowsContainers() })
 class LogsSmokeTest extends SmokeTest {
 
-  @Override
   protected String getTargetImage(String jdk) {
-    "ghcr.io/open-telemetry/opentelemetry-java-instrumentation/smoke-test-spring-boot:jdk11-20211213.1570880324"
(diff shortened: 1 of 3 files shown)


breedx-splk:
I just realized that we probably also want a GHA workflow for this? If so, I can add that here...

breedx-splk:
> I just realized that we probably also want a GHA workflow for this? If so, I can add that here...

Just verified that it's already running OOTB inside the "other" suite.

smoke-tests/src/test/groovy/io/opentelemetry/smoketest/LogsSmokeTest.groovy
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  protected TargetWaitStrategy getWaitStrategy() {
+    return new TargetWaitStrategy.Log(Duration.ofMinutes(1), ".*Started SpringbootApplication in.*")
+  }
+
+  def "Should export logs"() {
+    setup:
+    startTarget(11)
```
trask, line 35: probably worth running on 8, 11, 17 similar to `SpringBootSmokeTest`

trask (approved):
💯

mateuszrzeszutek (approved):

PR #5112, opened 2022-01-14
Convert InstrumentationTestRunner from interface to abstract class

anuraaga:
I am looking into enabling java lint on this repo like the other ones and one particularly hairy one was the `@SafeVarargs` on the default method in InstrumentationTestRunner. It can only be applied on a final method, which is impossible for interfaces. I figured there isn't a big advantage to using an interface since it's more of a "bag of logic" for only two implementations that we would ever write, and in fact it allowed me to share some more code between the two.


iNikem (approved):
Sealed classes!

mateuszrzeszutek (approved):

PR #5113, opened 2022-01-14
Fix workflow_dispatch

trask:
The github actions seem to run fine, but without these quotes, the workflow_dispatch ones are not showing up as manually runnable in the GitHub U/X 🤷‍♂️


anuraaga (approved):
😿

PR #5118, opened 2022-01-14
One more micrometer AsyncInstrumentRegistry fix

mateuszrzeszutek:
This is pretty much a continuation of https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5106

This morning I realized that we can't ever remove the recorder callback (even if there are no measurements being recorded), because that way we'd lose the information that a particular async instrument was already created - and OTel Meter still remembers that, so the next time somebody would try to create a gauge (e.g.) with the same name they'd not get their metrics, and receive an error log from OTel SDK instead.


trask (approved):
👍

PR #5120, opened 2022-01-14
Fix some java lint warnings

anuraaga:
Looking into enabling lint on this repo like the others, for now sending a first set of lint fixes. 

Depends on #5112


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/config/ConfigParsingException.java
```diff
@@ -6,6 +6,8 @@
 package io.opentelemetry.instrumentation.api.config;
 
 public class ConfigParsingException extends RuntimeException {
+  private static final long serialVersionUID = 384182080843382041L;
```
laurit, line 9: I would use `serialVersionUID = 1L;` otherwise it is hard to tell whether this is just a random number or deliberately computed with serialver

mateuszrzeszutek (approved):

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/tracer/DatabaseClientTracer.java
```diff
@@ -30,13 +29,17 @@
     extends BaseTracer {
   private static final String DB_QUERY = "DB Query";
 
-  protected final NetPeerAttributes netPeerAttributes;
+  protected final io.opentelemetry.instrumentation.api.tracer.net.NetPeerAttributes
```
iNikem, line 32: Why this change, @anuraaga ?

anuraaga: Since `NetPeerAttributes` is deprecated and can't be used in import statement

Issue #5124, opened 2022-01-14
Autoclose TracerProvider in Spring-Boot Application


frommeyerc:
**Is your feature request related to a problem? Please describe.**
According to `SdkTracerProvider#close()` JavaDoc this method should be called before the application is shut down in order to make sure all spans are reported. With the current spring-boot autoconfig this is not the case.

**Describe the solution you'd like**
While the current autoconfiguration create the `SdkTracerProvider` as a POJO baked into the `OpenTelemetry` `@Bean` it should become a `@Bean` in it's own right. This will allow the Spring-Context to call it's `#close()` method when the Context is destroyed.

**Describe alternatives you've considered**
As there seam to be no other `@Bean`s that would do the shutdown instead I didn't see another option.

I'm already preparing a PR for this change.

trask:
Closed by #5125

PR #5125, opened 2022-01-14
SdkTracerProvider auto closed as separate Context Bean (#5124)

frommeyerc:
Providing the `SdkTracerProvider` as a separate `@Bean` allows to auto close it together with the Spring-Context. This change also allows to provide a custom `SdkTracerProvider` if desired. If a user intends to simply create a custom `OpenTelemetry` object with an arbitrary `SkdTracerProvider` non of both will be auto created.

### DIFF
--- a/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
+++ b/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
@@ -33,29 +33,37 @@ import org.springframework.context.annotation.Configuration;
 @EnableConfigurationProperties(SamplerProperties.class)
 public class OpenTelemetryAutoConfiguration {
 
-  @Bean
-  @ConditionalOnMissingBean
-  public OpenTelemetry openTelemetry(
-      SamplerProperties samplerProperties,
-      ObjectProvider<ContextPropagators> propagatorsProvider,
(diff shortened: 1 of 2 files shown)


trask (approved):

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
```diff
@@ -33,29 +33,37 @@
...
-
-    return OpenTelemetrySdk.builder()
-        .setTracerProvider(tracerProvider)
-        .setPropagators(propagators)
-        .build();
+  @Configuration
+  @ConditionalOnMissingBean(OpenTelemetry.class)
+  public static class OpenTelemetryBeanConfig {
+
+    @Bean(destroyMethod = "close")
```
anuraaga, line 40: FWIU, `close` will be called automatically without needing to set it here

anuraaga (approved):

mateuszrzeszutek (approved):

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
trask: it looks like there's one other place with `destroyMethod = "close"`, going to merge and I'll send a separate PR to update both spots

PR #5131, opened 2022-01-14
Update RELEASING.md

trask:


buildscripts/generate-release-contributors.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -e
```
anuraaga, line 1: This file doesn't seem to be documented, intended to add it?

trask: sort of intentional, I wanted to get feedback on whether we should formalize this as part of the processes, but better to remove from this PR and add to SIG agenda

buildscripts/draft-change-log-entries.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+git log --reverse --pretty=format:"- %s" "$1"..HEAD \
+  | sed -r 's,\(#([0-9]+)\),\n  ([#\1](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/\1)),'
```
anuraaga, line 4:
```suggestion
  | sed -r 's,\(#([0-9]+)\),([#\1](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/\1)),'
```

As discussed on Slack, maybe it's simplest to not add this newline since rendered MD will be fine. We could probably remove the other script then since it'd be a 1:1 copy-pasta

anuraaga (approved):

buildscripts/draft-change-log-entries.sh
trask: > We could probably remove the other script then since it'd be a 1:1 copy-pasta

ya, that's a good point, will do 👍

buildscripts/generate-release-contributors.sh
```diff
@@ -0,0 +1,16 @@
...
+
+last_version=$1
+auth_token=$2
+
+git log --pretty=format:"%s" $last_version..HEAD | grep -oE "\(#[0-9]{4,}\)" | grep -oE "[0-9]{4,}" | xargs -I{} curl -s -H "Authorization: token $auth_token" https://api.github.com/repos/open-telemetry/opentelemetry-java-instrumentation/issues/{} | grep login | sed 's/.*"login": "\([^"]*\)".*/\1/' | sort -uf > pr-authors
```
anuraaga: @trask I'm good with going all in on expecting the `gh` CLI available. It also removes the need to deal with auth token in scripts.

It'll let you use `-q` for far more enjoyment than these sed expressions I think :)

e.g. https://cli.github.com/manual/gh_pr_view

anuraaga: Also if interested in adventures in graphql-land, queries would look something like this one

```
gh api graphql --paginate -F q="repo:open-telemetry/opentelemetry-java-instrumentation is:pr is:merged merged:>=2021-11-26" -f query='

query($q: String!, $endCursor: String) {
  search(query: $q, type: ISSUE, first: 100, after: $endCursor) {
    edges {
      node {
        ... on PullRequest {
          author { login }
        }
      }
    }
  }
}
' -q '.data.search.edges.[].node.author.login' | sort -uf
```

trask: > I'm good with going all in on expecting the `gh` CLI available. It also removes the need to deal with auth token in scripts.

sounds good, I'll try this out and submit generate-release-contributors.sh in a separate PR

PR #5132, opened 2022-01-14
Fix publishing gradle plugins

trask:
I'm not sure if we should publish io.opentelemetry.instrumentation.javaagent-shadowing publicly. It's just needed here as an implementation dependency of the public `muzzle-check` gradle plugin. "Vendoring" it in for now.

EDIT: this was already merged into v1.10.x branch in #5133

### DIFF
--- a/conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-shadowing.gradle.kts
+++ b/conventions/src/main/kotlin/io.opentelemetry.instrumentation.javaagent-shadowing.gradle.kts
@@ -4,6 +4,8 @@ plugins {
   id("com.github.johnrengelman.shadow")
 }
 
+// NOTE: any modifications below should also be made in
+//       io.opentelemetry.instrumentation.muzzle-check.gradle.kts
 tasks.withType<ShadowJar>().configureEach {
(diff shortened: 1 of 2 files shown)


laurit (approved):

anuraaga (approved):
FWIW, I was actually looking forward to replace this

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/build.gradle.kts#L158

though perhaps we should just focus a more wholistic distro plugin

mateuszrzeszutek (approved):

iNikem:
> I'm not sure if we should publish io.opentelemetry.instrumentation.javaagent-shadowing publicly

Why not, @trask ?

trask:
> > I'm not sure if we should publish io.opentelemetry.instrumentation.javaagent-shadowing publicly
> 
> Why not, @trask ?

I like @anuraaga's suggestion above

> perhaps we should just focus a more wholistic distro plugin

(basically a rebranding)

another option we discussed is publishing things that are only needed for use inside of otel projects (e.g. spotless conventions) under an "internal" groupId (but still published so we can share across the 3 otel repos)

PR #5143, opened 2022-01-16
Add workflow to prepare release branch

trask:
Automate a couple of the manual release steps

I tested this on my fork, you can look there to see example of the two PRs that this workflow creates: https://github.com/trask/opentelemetry-java-instrumentation/pulls

### DIFF
--- a/.github/workflows/prepare-release-branch.yml
+++ b/.github/workflows/prepare-release-branch.yml
@@ -1,0 +1,77 @@
+name: Prepare Release Branch
+on:
+  workflow_dispatch:
+
+jobs:
+  prepare-release-branch:
+    runs-on: ubuntu-latest
+    outputs:
+      release-branch-name: ${{ steps.set-release-branch-name.outputs.release-branch-name }}
+    steps:
+      - uses: actions/checkout@v2.3.4
+
+      - name: Set release branch name
+        id: set-release-branch-name
+        run: |
+          release_branch_name=$(grep -Eo "[0-9.]+-SNAPSHOT" version.gradle.kts | sed -E 's/([0-9]+)\.([0-9]+)\.([0-9]+)-SNAPSHOT/v\1.\2.x/')
+          echo "::set-output name=release-branch-name::$release_branch_name"
+
+      - name: Create release branch
+        run: |
+          git checkout -b ${{ steps.set-release-branch-name.outputs.release-branch-name }}
+          git push origin ${{ steps.set-release-branch-name.outputs.release-branch-name }}
+
+  create-pull-request-against-release-branch:
+    needs: prepare-release-branch
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+        with:
+          ref: ${{ needs.prepare-release-branch.outputs.release-branch-name }}
+
+      - name: Remove SNAPSHOT from version
+        run: |
+          v=$(grep -Eo "[0-9]+.[0-9]+.[0-9]+-SNAPSHOT" version.gradle.kts | sed 's/-SNAPSHOT//')
+          .github/scripts/update-versions.sh "$v-SNAPSHOT" "$v-alpha-SNAPSHOT" "$v" "$v-alpha"
+
+      - name: Bump download link version
+        run: |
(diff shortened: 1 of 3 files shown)


laurit (approved):

anuraaga (approved):

.github/workflows/prepare-release-branch.yml
```diff
@@ -0,0 +1,77 @@
...
+            minor="${BASH_REMATCH[2]}"
+          else
+            echo "unexpected version: $v"
+            exit 1
+          fi
+          bump="$major.$((minor + 1)).0"
+          .github/scripts/update-versions.sh "$v-SNAPSHOT" "$v-alpha-SNAPSHOT" "$bump-SNAPSHOT" "$bump-alpha-SNAPSHOT"
+
+      - name: Create pull request
+        uses: peter-evans/create-pull-request@v3
```
anuraaga, line 71: If wanting to avoid a third party action, might be easy enough to use gh CLI

https://cli.github.com/manual/gh_pr_create

.github/workflows/prepare-release-branch.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          commit-message: Bump snapshot version
+          branch: bump-snapshot-version
+          delete-branch: true
```
anuraaga, line 75: FWIW, I was trying to see how delete-branch could possibly work when creating a pull request, and indeed it doesn't

https://github.com/peter-evans/create-pull-request#action-behaviour

It would delete the PR if it had already been merged when running this step. So to prevent confusion, we should delete this field if sticking with the action. But I'd say the CLI command is simple and all we need, this action seems to support some intense workflows :O

https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#how-the-action-works

RELEASING.md
```diff
@@ -14,17 +14,8 @@ as the last step, which publishes a snapshot build to
...
-  git push upstream v1.9.x
-  ```
-* Merge a PR to the release branch with the following changes
-  * Remove `-SNAPSHOT` from the version in these files:
-    * version.gradle.kts
-    * examples/distro/build.gradle
-    * examples/extension/build.gradle
-  * Bump the version in the download link in the root `README.md` file
+* Run the [Prepare Release Branch workflow](https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/workflows/prepare-release-branch.yml).
+* Review and merge the two PRs that it creates (one is targeted to the release branch and one is targeted to the `main` branch)
```
anuraaga, line 18: Make sure to delete the branch after merging

.github/workflows/prepare-release-branch.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Create pull request
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          msg="Bump snapshot version"
+          git add -u
+          git commit -m "$msg"
+          git push origin HEAD:bump-snapshot-version
+          gh pr create --title "$msg" --body "$msg" --head bump-snapshot-version --base main
```
anuraaga, line 92: I just noticed the `**/**` CNCF branch protection rule seems to be gone, did we accidentally delete it?

Though either way, even if we can make it work for release branch, this PR creation to `main` looks like it'll be problematic due to CLA check :(

anuraaga: Though we do explicitly have the checkbox to let us merge even if status checks aren't all passing. So that could be our workaround

trask: we have branch protection on both `main` and `v**`, so CLA check will be a problem for both PRs 😞

trask: it looks there are [complexities to approving github-actions bot account for EasyCLA](https://github.com/open-telemetry/community/issues/809#issuecomment-905869941) and so that path could take some time

it seems that the current recommendation is to use a Personal Access Token generated by one of the maintainers: https://github.com/open-telemetry/community/blob/main/docs/using-github-extensions.md#writing-your-github-actions-pipelines

trask: reverted back to using github-actions bot for now at least

mateuszrzeszutek (approved):

PR #5144, opened 2022-01-16
Simplify release workflow

trask:
Two things in this PR:
* Detect version based on the branch that the workflow is being run on
* Delete the patch release workflow, as the normal release workflow should work for patch releases now


laurit (approved):

anuraaga (approved):

RELEASING.md
```diff
@@ -73,12 +69,10 @@ Before making the release:
...
-To make a patch release, open the patch release build workflow in your browser
-[here](https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/workflows/patch-release-build.yml).
+Open the [Release workflow](https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/workflows/release.yml).
 
-You will see a button that says "Run workflow". Press the button, then enter the following:
-* Use workflow from: <select the branch from dropdown list, e.g. `v1.9.x`>
-* The release branch to use: <e.g. `v1.9.x`>
-* The version of the release: <e.g. `1.9.1`>
+Press the "Run workflow" button, then select the release branch from the dropdown list,
+e.g. `v1.9.x`, and click the "Run workflow" button below that.
```
mateuszrzeszutek, line 75: Before running the workflow, should be bump the version in `version.gradle.kts` and others?

trask: ya, this is doc'd above

PR #5145, opened 2022-01-16
Remove fetch-depth

trask:
I thought `fetch-depth: 0` meant fetch only the most recent commit, but it means fetch all history for all branches, which seems unnecessary. Not sure if there's a reason to keep it?


laurit (approved):
Nice, I think we should remove it.

anuraaga (approved):
This was needed for nebula to fetch tags, but not anymore without nebula

mateuszrzeszutek (approved):

Issue #1338, opened 2020-10-07
Set net peer for gRPC client


anuraaga:
When separating out library instrumentation from auto, we removed an auto-specific workaround for setting net peer. Clients should provide the net peer in general, and gRPC does do so with an "ExperimentalApi". Going to dig for ways without that, or to find a way with the existing APIs, before requiring users to pass in the net peer themselves.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/1329#discussion_r500031829

mateuszrzeszutek:
`net.peer.*` attributes are now added by `GrpcNetAttributesExtractor` -- `peer.service` was still missing though, will be added in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/3357

ralphgj:
Hi, I used agent of v1.10.0 and then I found the grpc client spans don't capture the attributes of `net.peer`. But in this issue, it seems being already resolved.
I use the [grpc-java examples](https://github.com/grpc/grpc-java/tree/master/examples/src/main/java/io/grpc/examples/helloworld) as the demo.

trask:
hi @ralphgj, can you open a new issue for this? thx!

ralphgj:
> hi @ralphgj, can you open a new issue for this? thx!

yeah, I will do it.

PR #5151, opened 2022-01-17
feat: add grpc attribute "rpc.grpc.authority"

ralphgj:


instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
```diff
@@ -43,6 +43,7 @@
   public <REQUEST, RESPONSE> ClientCall<REQUEST, RESPONSE> interceptCall(
       MethodDescriptor<REQUEST, RESPONSE> method, CallOptions callOptions, Channel next) {
     GrpcRequest request = new GrpcRequest(method, null, null);
+    request.setAuthority(next.authority());
```
mateuszrzeszutek, line 46: The Javadocs for `authority()` say the following:
```
The authority of the destination this channel connects to. Typically this is in the format host:port.
```
Is there any significant difference between capturing `authority()` and the `net.peer.*` attributes?

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcRequest.java
```diff
@@ -17,6 +17,8 @@
 
   @Nullable private volatile SocketAddress remoteAddress;
 
+  @Nullable private volatile String authority;
```
mateuszrzeszutek, line 20: This does not change during this object's life, it's set once just after the construction - can you make it a final field instead?

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcAttributesExtractor.java
```diff
@@ -14,7 +14,9 @@
 final class GrpcAttributesExtractor implements AttributesExtractor<GrpcRequest, Status> {
   @Override
   public void onStart(AttributesBuilder attributes, GrpcRequest grpcRequest) {
-    // No request attributes
+    if (grpcRequest.getAuthority() != null) {
+      attributes.put(GrpcHelper.RPC_GRPC_AUTHORITY, grpcRequest.getAuthority());
```
mateuszrzeszutek, line 18: This new attribute does not seem to be present in the [RPC semantic conventions spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#grpc). Would you mind opening a PR/issue against the spec repo that introduces this attribute?
And, for now, how about moving it to a separate `AttributesExtractor` that's disabled by default? We're using the `experimental-span-attributes` for these, e.g. `otel.instrumentation.aws-sdk.experimental-span-attributes`

anuraaga: Note that I agree this attribute seems to already be captured as part of `net.peer` so unless there is a difference, we probably wouldn't want to go forward with even an experimental attribute.

ralphgj: Thank you for your reply. I found in grpc client spans there were no `net.peer` attributes. So I add the `authority` attribute by this PR and then we can know the invoked remote service's host and port. Is this a bug in grpc client instrumentation?

trask: @ralphgj there should be `net.peer` attributes captured for grpc client spans since version 1.4.0 (#1338), if you're on latest and not seeing them it could be a bug

ralphgj: thx @trask. I will close this PR and comment in #1338

Issue #5152, opened 2022-01-17
Proposal: Remove log4j-context-data-2.7


hdost:
Given it's currently experimental anyhow and with the most recent CVEs around log4j
https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/log4j/log4j-context-data/log4j-context-data-2.7
It seems this should just be removed entirely to prevent encouraging people to use vulnerable versions of log4j.

**CVEs**
https://www.cvedetails.com/cve/CVE-2021-45046
https://www.cvedetails.com/cve/CVE-2021-44832
https://www.cvedetails.com/cve/CVE-2021-44228
https://www.cvedetails.com/cve/CVE-2021-45105

trask:
hi @hdost! this would definitely make sense for library instrumentation, but is a more difficult decision for javaagent instrumentation, check out https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/VERSIONING.md#dropping-support-for-older-library-versions

mateuszrzeszutek:
Closing this for the same reason as #5153 : https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5153#issuecomment-1023887602

PR #5153, opened 2022-01-17
Remove log4j-context-data-2.7

hdost:
It doesn't make sense to encourage people to use vulnerable software for new projects.

Affected by Log4j Major CVEs
https://www.cvedetails.com/cve/CVE-2021-45046
https://www.cvedetails.com/cve/CVE-2021-44832
https://www.cvedetails.com/cve/CVE-2021-44228
https://www.cvedetails.com/cve/CVE-2021-45105

Relates to #5152


trask:
> for new projects

This is a key difference between library instrumentation and javaagent instrumentation. Often javaagent instrumentation is applied to (very) old projects / existing systems.

hdost:
Perhaps merging the code into just the java agent then to prevent it from being used as a library elsewhere?

trask:
> Perhaps merging the code into just the java agent then to prevent it from being used as a library elsewhere?

hey @hdost, it's already only a javaagent module, so I'm not quite sure what you mean?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/log4j/log4j-context-data/log4j-context-data-2.7

hdost:
> > Perhaps merging the code into just the java agent then to prevent it from being used as a library elsewhere?
> 
> hey @hdost, it's already only a javaagent module, so I'm not quite sure what you mean?
> 
> https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/log4j/log4j-context-data/log4j-context-data-2.7

It's available as a JAR https://search.maven.org/artifact/io.opentelemetry.javaagent.instrumentation/opentelemetry-javaagent-log4j-context-data-2.7/1.10.1-alpha/jar which was more my point, make it **only** available as part of the javaagent. So essentially merge it into the java agent module.

trask:
that maven artifact doesn't have a transitive dependency on log4j, so shouldn't pull in any old versions of log4j or trigger any CVE detection. i'm curious if that addresses your concern, of if your concern was from a different perspective? thx

trask:
@hdhost let us know if you still consider this an issue, thx!

PR #5168, opened 2022-01-18
Fix flaky micrometer retries in javaagent test

mateuszrzeszutek:
I found an interesting thing here: https://ge.opentelemetry.io/s/4duiw4mxhj6e4/tests/:instrumentation:micrometer:micrometer-1.5:javaagent:test/io.opentelemetry.javaagent.instrumentation.micrometer.v1_5.FunctionTimerTest/functionTimersWithSameNameAndDifferentTags()?top-execution=1

The first execution failed on missing 1 measurement out of 2 (I don't know why yet 🙈 ), but the rest of them missed all measurements. It turns out the gradle retry plugin creates a new CL with all the test code for each retry and executes it on the same JVM; so the OpenTelemetry `Meter` stays the same (because it's bootstrap/agent CL) but the `AsyncInstrumentRegistry` is created anew. This PR moves `AsyncInstrumentRegistry` to `instrumentation-api` so that it's in the same CL as the `Meter` interface.

~~I still have no idea why that test has failed the first time, but I _strongly_ suspect it's also the fault of the `AsyncInstrumentRegistry` hackery (and having to maintain one callback with multiple measurements).~~


trask (commented):
> This PR moves AsyncInstrumentRegistry to instrumentation-api so that it's in the same CL as the Meter interface.

Can we use a micrometer instrumentation bootstrap module instead of moving it to instrumentation-api?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/AsyncInstrumentRegistry.java
```diff
@@ -21,14 +19,28 @@
 import javax.annotation.Nullable;
 
 // TODO: refactor this class, there's too much copy-paste here
-final class AsyncInstrumentRegistry {
+public final class AsyncInstrumentRegistry {
```
trask, line 22: can you add a comment on this class about why it needs to live in the bootstrap class loader?

trask (approved):

mateuszrzeszutek:
> Can we use a micrometer instrumentation bootstrap module instead of moving it to instrumentation-api?

No, unfortunately not - micrometer library instrumentation needs to have access to this class too. Moving the class to `instrumentation-api` seems to be the most suitable solution here (aside from waiting for multiple callbacks support in the SDK 😅 )

laurit (approved):

anuraaga (approved):

Issue #5156, opened 2022-01-17
NoSuchMethodError when using agent and modules (JPMS) and selenium


valodzka:
### Discussed in https://github.com/open-telemetry/opentelemetry-java-instrumentation/discussions/5126

<div type='discussions-op-text'>

<sup>Originally posted by **valodzka** January 14, 2022</sup>
I'm trying to add agent to my project but see the issue during startup (stack trace below). I think issue is that:
 
1. Project uses java module system (JPMS)
2. Project has selenium-remote as dependency which has opentelementry as dependency

In combination it causes that incorrect version of autoconfigure loaded (shaded vs unshaded). 

```
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at io.opentelemetry.javaagent.bootstrap.AgentInitializer.initialize(AgentInitializer.java:40)
	at io.opentelemetry.javaagent.OpenTelemetryAgent.agentmain(OpenTelemetryAgent.java:51)
	at io.opentelemetry.javaagent.OpenTelemetryAgent.premain(OpenTelemetryAgent.java:44)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:491)
	at java.instrument/sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:503)
[...]

PR #5169, opened 2022-01-18
Always set parent of agent class loader to bootstrap class loader

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5156

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
@@ -99,7 +99,7 @@ public final class AgentInitializer {
    * @return Agent Classloader
    */
   private static ClassLoader createAgentClassLoader(String innerJarFilename, File javaagentFile) {
-    return new AgentClassLoader(javaagentFile, innerJarFilename, null);
+    return new AgentClassLoader(javaagentFile, innerJarFilename);
   }
 
   private static AgentStarter createAgentStarter(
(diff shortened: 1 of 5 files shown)


javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -100,31 +98,8 @@ public static ClassLoader getExtensionsClassLoader() {
...
-     compatible with java 8.
-    */
-    Method method = ClassLoader.class.getDeclaredMethod("getPlatformClassLoader");
-    return (ClassLoader) method.invoke(null);
-  }
-
-  public static boolean isJavaBefore9() {
-    return System.getProperty("java.version").startsWith("1.");
+  private static ClassLoader createAgentClassLoader(String innerJarFilename, File javaagentFile) {
+    return new AgentClassLoader(javaagentFile, innerJarFilename, null);
```
trask, line 102: any reason to keep the `parent` argument in the constructor?

trask (approved):
👍

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
```diff
@@ -100,31 +98,8 @@ public static ClassLoader getExtensionsClassLoader() {
...
    * @return Agent Classloader
    */
-  private static ClassLoader createAgentClassLoader(String innerJarFilename, File javaagentFile)
-      throws Exception {
-    ClassLoader agentParent;
-    if (isJavaBefore9()) {
-      agentParent = null; // bootstrap
-    } else {
-      // platform classloader is parent of system in java 9+
-      agentParent = getPlatformClassLoader();
```
anuraaga, line 110: Any idea why we were doing this before?

anuraaga (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentInitializer.java
trask: this was inherited, my best guess is for access to `java.sql` package for some reason(?)

Issue #5163, opened 2022-01-18
Micrometer flaky test


trask:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/4849075486

```
FunctionTimerTest > functionTimersWithSameNameAndDifferentTags() FAILED
    org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a io.opentelemetry.instrumentation.testing.junit.InstrumentationExtension 
    Expecting any element of:
      []
    to satisfy the given assertions requirements but none did:

     within 10 seconds.
```

PR #5171, opened 2022-01-18
Fix flaky micrometer tests

mateuszrzeszutek:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5163

Looks like on OpenJ9 JVMs these local variables were GC'd just after creating a meter, and it resulted in nulled-out weak refs and no measurements being observed.


trask (approved):
👍

PR #5172, opened 2022-01-18
Hopefully fix flaky spring-rmi test

laurit:


trask (approved):
were you still seeing flakes after #5164?

laurit:
@trask I didn't notice that pr. If I remember correctly we had something similar going on with regular rmi test which was fixed by adding this jvm argument https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2738

trask:
I just got clean runs after merging #5164 into #5165 and #5160. I'm good with either waiting to see if there is additional flakiness, or proactively merging this given we have similar config for rmi tests.

mateuszrzeszutek (approved):

mateuszrzeszutek:
I think we should probably merge this one too, there may be more than 1 issue here.

PR #5182, opened 2022-01-20
Enable java lint

anuraaga:
Need help! I get the same lint warning as javaagent-tooling-java9, but after suppressing it, the lint failure goes away and instead it just fails to compile with absolutely no error message even with `--debug` 😿 

Bumping to Java 11 does succeed.

Can anyone help debug this? I have no leads


laurit:
@anuraaga See whether this helps https://github.com/anuraaga/opentelemetry-java-instrumentation/pull/34

mateuszrzeszutek (approved):

trask (approved):

Issue #5155, opened 2022-01-17
java.lang.LinkageError thrown at startup after updating from 1.7 to 1.9


gsmirnov-splk:
**Describe the bug**
After updating splunk-otel-java from 1.5.0 to 1.6.1 (comes with `opentelemetry-java-instrumentation` updated to 1.9), the instrumented application (Apache Druid) fails to start with this error:

```
java.lang.LinkageError: loader (instance of  sun/misc/Launcher$AppClassLoader): attempted  duplicate class definition for name: "org/jboss/netty/channel/ChannelFutureListener"
```

After initial troubleshooting, @iNikem suggested that I open an issue in this repo. Subsequent troubleshooting confirmed that the vendor-neutral upstream `opentelemetry-javaagent - version: 1.10.0` also fails with the same error.

**Steps to reproduce**
1. Make sure you have Java 8 selected
2. Run Apache Druid quickstart: (Steps 1&2 [here](https://druid.apache.org/docs/latest/tutorials/index.html))
3. Ensure that the Historical is up and running, `http://localhost:8083/status` should return a JSON. It may take a minute to start
4. Download a [splunk-otel-java agent](https://github.com/signalfx/splunk-otel-java/releases/download/v1.6.1/splunk-otel-javaagent-all.jar) and place it into the `apache-druid-0.22.1` folder
4. Attach the agent to the historical node: `echo "-javaagent:splunk-otel-javaagent-all.jar" >> conf/druid/single-server/micro-quickstart/historical/jvm.config`
5. Start druid again: `./bin/start-micro-quickstart`

**What did you expect to see?**
Historical starts up like it did before.

**What did you see instead?**
Historical fails to start:
```
[Mon Jan 17 13:47:01 2022] Running command[historical], logging to[apache-druid-0.22.1/var/sv/historical.log]: bin/run-druid historical conf/druid/single-server/micro-quickstart
[Mon Jan 17 13:47:15 2022] Command[historical] exited (pid = 2747, exited = 1)
[Mon Jan 17 13:47:15 2022] Command[historical] failed, see logfile for more details: apache-druid-0.22.1/var/sv/historical.log
```
```
2022-01-17T11:46:41,236 ERROR [main] org.apache.druid.cli.CliHistorical - Error when starting up.  Failing.
[...]

PR #5185, opened 2022-01-20
Recover from duplicate class definition errors

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5155
If there is a duplicate class definition find the class with `findLoadedClass` and return it instead of throwing a `LinkageError`. Currently no attempt is made to detect where the duplicate definition was caused by us or not. If needed we could add a `ThreadLocal` to detect whether helper classes were injected while defining a class and only suppress duplicate class definition errors then.

### DIFF
--- a/instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/DefineClassInstrumentation.java
+++ b/instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/DefineClassInstrumentation.java
@@ -75,11 +75,18 @@ public class DefineClassInstrumentation implements TypeInstrumentation {
       MethodVisitor mv = super.visitMethod(access, name, descriptor, signature, exceptions);
       // apply the following transformation to defineClass method
       /*
+      DefineClassContext.enter();
       try {
         // original method body
+        DefineClassContext.exit();
+        return result;
       } catch (LinkageError error) {
+        boolean helpersInjected = DefineClassContext.exitAndGet();
         Class<?> loaded = findLoadedClass(className);
-        return DefineClassUtil.handleLinkageError(error, loaded);
+        return DefineClassUtil.handleLinkageError(error, helpersInjected, loaded);
+      } catch (Throwable throwable) {
+        DefineClassContext.exit();
+        throw throwable;
       }
        */
       if ("defineClass".equals(name)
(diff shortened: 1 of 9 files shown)


mateuszrzeszutek (approved):

trask (approved):
👍

this issue has caused headaches for injecting SPI classes (since ideally we would inject them when we see the SPI interface loaded) e.g.

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/88121c1583df18af7c67fd63d4792351546c4ea4/instrumentation/log4j/log4j-context-data/log4j-context-data-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_16/Log4j2InstrumentationModule.java#L55-L58

I'm in favor of adding the ThreadLocal check if that doesn't add too much complexity.

mateuszrzeszutek (approved):

instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/DefineClassInstrumentation.java
```diff
@@ -93,12 +94,31 @@ public MethodVisitor visitMethod(
 
               @Override
               public void visitCode() {
+                mv.visitMethodInsn(
+                    Opcodes.INVOKESTATIC,
+                    Type.getInternalName(DefineClassContext.class),
+                    "enter",
+                    "()V",
+                    false);
```
trask, line 102: can you update the pseudo code above (which is really helpful) with the new enter/exit/exitAndGet?

trask (approved):
thx!

instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/DefineClassInstrumentation.java
```diff
@@ -75,11 +75,18 @@ public MethodVisitor visitMethod(
...
+        DefineClassContext.exit();
+        return result;
       } catch (LinkageError error) {
+        boolean helpersInjected = DefineClassContext.exitAndGet();
         Class<?> loaded = findLoadedClass(className);
-        return DefineClassUtil.handleLinkageError(error, loaded);
+        return DefineClassUtil.handleLinkageError(error, helpersInjected, loaded);
+      } catch (Throwable throwable) {
+        DefineClassContext.exit();
+        throw throwable;
```
trask, line 89: 👍

trask (approved):
thx again, I'm looking forward to trying this out by removing a couple of SPI interface instrumentation hacks

Issue #5186, opened 2022-01-20
Spring boot cloud gateway, context propagation broken


dynatrace-peterlang:
**Describe the bug**
Since release `v1.10.0` the context propagation of a reactor-core context within spring cloud gateway is broken. I would only expect one trace for my sample scenario, but since v.1.10.0 two traces are produced.

**Steps to reproduce**
In a controller for spring cloud gateway I use a `flatMap` operation to chain two web service calls. 
Simplified sample code in gateway controller:

```
@GetMapping(value = "backendmap")
public Mono<Pojo> backendmap() {
    return requestPojo().flatMap(it -> requestMojo());
}
```
full reproducer in https://github.com/dynatrace-peterlang/springapigateway-reproducer 

**What did you expect to see?**
For agents until version 1.9.2 in jaeger I observe von trace for the whole operation. Simplified structure:

```
unknown_service:java /controller/backendmap
    unknown_service:java ApiGatewayController.backendmap
    unknown_service:java HTTP GET
        unknown_service:java /controller/pojo
            unknown_service:java ApiGatewayController.getPojo
    unknown_service:java HTTP GET
        unknown_service:java /controller/mojo
            unknown_service:java ApiGatewayController.getMojo
```

**What did you see instead?**
With the new agent version `1.10.0` this trace is split up in two. It seems the context propagation in operation `flatMap` does not work anymore

```
unknown_service:java /controller/backendmap
    unknown_service:java ApiGatewayController.backendmap
    unknown_service:java HTTP GET
        unknown_service:java /controller/pojo
            unknown_service:java ApiGatewayController.getPojo
```
and a second trace

```         
unknown_service:java HTTP GET
    unknown_service:java /controller/mojo
        unknown_service:java ApiGatewayController.getMojo
```

**What version are you using?**
broken since `v1.10.0`

**Environment**
Compiler: Eclipse Temurin 11 `OpenJDK Runtime Environment Temurin-11.0.13+8 (build 11.0.13+8)`
OS: Windows

**Additional context**
It seems context propagation was broken by changes in https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4786

trask:
@dynatrace-peterlang thanks for reporting this! it's now fixed in the [1.10.1 patch release](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.10.1)

PR #5188, opened 2022-01-20
Move reactor context propagation operator instrumentation to separate…

laurit:
… instrumentation module
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5186
Context propagation operator instrumentation depends on the presence of opentelemetry api. Move it to separate instrumentation module so that muzzle wouldn't fail for the whole reactor instrumentation when opentelemetry api isn't present.


mateuszrzeszutek (approved):

trask (approved):
👍

PR #5194, opened 2022-01-20
Factor out NetServerAttributesGetter

breedx-splk:
Continuing our series on favoring composition over inheritance. This is a follow-up from #5030, but geared at the server attributes.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/NetServerAttributesGetter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/NetServerAttributesGetter.java
@@ -12,8 +12,8 @@ import javax.annotation.Nullable;
  * type into the 4 common attributes (transport, peerName, peerPort, peerIp).
  *
  * <p>Instrumentation authors will create implementations of this interface for their specific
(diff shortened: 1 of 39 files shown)


trask (approved):
I like that `NetServerAttributesGetter` is only parameterized by `REQUEST`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/NetServerAttributesGetter.java
```diff
@@ -0,0 +1,31 @@
...
+package io.opentelemetry.instrumentation.api.instrumenter.net;
+
+import javax.annotation.Nullable;
+
+/**
+ * An interface for getting server-based network attributes. It adapts a vendor-specific request
+ * type into the 4 common attributes (transport, peerName, peerPort, peerIp).
+ *
+ * <p>Instrumentation authors will create implementations of this interface for their specific
+ * server library/framework. It will be used by the NetServerAttributesExtractor to obtain the
```
mateuszrzeszutek, line 15: Nit

```suggestion
 * server library/framework. It will be used by the {@link NetServerAttributesExtractor} to obtain the
```

mateuszrzeszutek (approved):

trask:
@breedx-splk looks like needs another `spotlessApply` then will merge

PR #5195, opened 2022-01-21
Use CLA approved user

trask:
Based on @anuraaga's observation that we can set arbitrary user on the commit, don't need a Personal Access Token after all.

Will replace this with @opentelemetry-java-bot once github unflags the account I created and CNCF whitelists it for EasyCLA.


anuraaga (approved):
Unless you were able to confirm somehow on a fork just to set expectations it is a hypothesis :)

mateuszrzeszutek (approved):

laurit (approved):

Issue #5154, opened 2022-01-17
Webflux + Kotlin lossing span


xsmrcek:
**Describe the bug**
Instrumentation using agent is lossing span.
Every client request have new/different trecaparent.

**Steps to reproduce**
Build application from attachment, build docker image with app and nevest javaagent.
Configuration I use:
ENV OTEL_TRACES_EXPORTER="none"
ENV OTEL_METRICS_EXPORTER="none"
ENV OTEL_PROPAGATORS="tracecontext,b3,baggage"
I am testing just propagation now. ( I wan to use application insights javaagent in future/ As far as I know it is build on top of opentelemetry javaagent, so first I need this one working)
**What did you expect to see?**
Trace headers (traceparent, traceid, spanin etc is generated/propagated correctly)
**What did you see instead?**
New trace headers for every request.
**What version are you using?**
newest opentelemetry-javaagent (1.10.0)

**Environment**
openjdk 15 2020-09-15
spring-boot-starter-parent 2.5.5 (webflux)
kotlin 1.5.31
logbook (for logging) 2.4.1

xsmrcek:
I wan't able to upload an attachment, so I upload minimal exmaple here: https://github.com/xsmrcek/minimal-example

PR #5196, opened 2022-01-21
kotlinx-coroutines-reactor context propagation

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5154


mateuszrzeszutek (approved):

trask (approved):

instrumentation/kotlinx-coroutines/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlinxcoroutines/reactor/KotlinCoroutinesFluxInstrumentation.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
+import io.opentelemetry.javaagent.extension.instrumentation.TypeTransformer;
+import io.opentelemetry.javaagent.instrumentation.kotlinxcoroutines.KotlinCoroutinesInstrumentationHelper;
+import kotlin.coroutines.CoroutineContext;
+import net.bytebuddy.asm.Advice;
+import net.bytebuddy.description.type.TypeDescription;
+import net.bytebuddy.matcher.ElementMatcher;
+
+public class KotlinCoroutinesFluxInstrumentation implements TypeInstrumentation {
```
anuraaga, line 20: This propagates context when the flux is created as opposed to when it is subscribed to, which I believe differs from our general pattern in reactor / rxjava. Is there anything about kotlin coroutines that would mean we should not follow that pattern?

anuraaga (commented):
Is this something that would only affect agent users and not library users

anuraaga:
It's always hard to context switch back into reactive land, but I think our general thought has been for patterns like this one

https://github.com/xsmrcek/minimal-example/blob/master/src/main/kotlin/org/example/minimalexample/Application.kt#L21

Where a flux is returned to a framework that subscribes to it, we're supposed to instrument the framework (spring) rather than flux creation, the latter not being applicable to all cases (for example a cached flux used in several different requests).

anuraaga (commented):
The current agent instrumentation was written by me without knowing kotlin coroutines well :)

Wondering if we should change the current instrumentation to instead instrument `newCoroutineContext`?

https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/Builders.kt#L49
https://github.com/Kotlin/kotlinx.coroutines/blob/master/reactive/kotlinx-coroutines-reactor/src/Mono.kt#L87

instrumentation/kotlinx-coroutines/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlinxcoroutines/reactor/KotlinCoroutinesFluxInstrumentation.java
laurit: Hopefully fixed now. I moved the instrumentation so that it should now be called from reactive flow and propagate the context that is active there into coroutine.
For library users context propagation is possible using kotlin coroutine library instrumentation and reactor `ContextPropagationOperator`
```kotlin
      val mono = mono(dispatcher) {
        // extract context from reactor and propagate it coroutine
        val reactorContext = coroutineContext[ReactorContext.Key]?.context
        val otelContext = ContextPropagationOperator.getOpenTelemetryContext(reactorContext, Context.current())
        withContext(otelContext.asContextElement()) {
          tracedChild("child")
        }
      }
      ContextPropagationOperator.runWithContext(mono, Context.current()).awaitSingle()
```

anuraaga:
@laurit Any thoughts on https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5196#pullrequestreview-861820292 ? Was hoping to instrument the lowest level building block to not have to instrument every type of coroutines library they build on it

laurit:
> Any thoughts on [#5196 (review)](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5196#pullrequestreview-861820292) ? Was hoping to instrument the lowest level building block to not have to instrument every type of coroutines library they build on it

@anuraaga seems to work nicely. I removed everything else and just kept instrumentation for `newCoroutineContext`.

mateuszrzeszutek (approved):

instrumentation/kotlinx-coroutines/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlinxcoroutines/KotlinCoroutinesInstrumentationHelper.java
```diff
@@ -13,11 +13,16 @@
 
   public static CoroutineContext addOpenTelemetryContext(CoroutineContext coroutineContext) {
     Context current = Context.current();
+    return addOpenTelemetryContext(coroutineContext, current);
+  }
+
+  public static CoroutineContext addOpenTelemetryContext(
```
anuraaga, line 19: Might be missing it but looks like we can revert this file? Don't think I see this used

anuraaga (approved):
Thanks for the iterations!

instrumentation/kotlinx-coroutines/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/kotlinxcoroutines/KotlinCoroutinesInstrumentationHelper.java
laurit: Thanks, reverted

PR #5197, opened 2022-01-21
Expected server span name for null route should be HTTP GET

mateuszrzeszutek:
Another one of `HttpServerTest` simplifications from #442


trask (approved):
nice

PR #5201, opened 2022-01-21
Actually fix redisson instrumentation after 3.16.8 release

mateuszrzeszutek:
Implemented Lauri's idea of using `VirtualField`s to avoid calling `getPromise()`


breedx-splk (approved):
Looks good to me. It took me a bit to figure out the `CommandData` vs. `CommandsData` difference. Does it make sense to combine those into a single class that takes the class and name? Might be a little clearer. 🤷🏻

mateuszrzeszutek:
You still need to have different advice classes for these two - except the method matcher there's pretty much nothing in common.

laurit (approved):

breedx-splk:
> You still need to have different advice classes for these two - except the method matcher there's pretty much nothing in common.

Ah yeah, didn't consider that we still need two advices because we don't control creation of them (and just pass the name as a string). Gotcha.

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonInstrumentationModule.java
```diff
@@ -21,6 +21,9 @@ public RedissonInstrumentationModule() {
 
   @Override
   public List<TypeInstrumentation> typeInstrumentations() {
-    return singletonList(new RedisConnectionInstrumentation());
+    return asList(
+        new CommandDataInstrumentation(),
+        new CommandDataInstrumentation(),
```
trask, line 26:
```suggestion
        new CommandsDataInstrumentation(),
```

trask (commented):
what do you think about using reflection to call `getPromise()` instead?

mateuszrzeszutek:
> what do you think about using reflection to call `getPromise()` instead?

That could work too! I reworked this PR to use `MethodHandle`s instead

instrumentation/redisson-3.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/redisson/RedissonRequest.java
```diff
@@ -23,6 +29,11 @@
 @AutoValue
 public abstract class RedissonRequest {
 
+  private static final VirtualField<CommandData<?, ?>, CompletionStage<?>> commandDataPromiseField =
```
anuraaga, line 32: Think can remove these

anuraaga (approved):

trask (approved):

PR #5207, opened 2022-01-23
Add release contributor script

trask:
Next step, populate the release notes during the release workflow using the change log and this script (the occasional highlight at the top of the release notes would still be manual).

### DIFF
--- a/.github/scripts/generate-release-contributors.sh
+++ b/.github/scripts/generate-release-contributors.sh
@@ -1,0 +1,80 @@
+#!/bin/bash -e
+
+# shellcheck disable=SC2016
+# shellcheck disable=SC2086
+
+from_version=$1
+to_version=$2
+
+# get the date of the first commit on main that wasn't in the from_version
+from=$(git log --reverse --pretty=format:"%cI" $from_version..HEAD | head -1)
+
+# get the last commit on main that was in the to_version
+to=$(git merge-base HEAD $to_version | xargs git log -1 --pretty=format:"%cI")
+
+contributors1=$(gh api graphql --paginate -F q="repo:open-telemetry/opentelemetry-java-instrumentation is:pr base:main is:merged merged:$from..$to" -f query='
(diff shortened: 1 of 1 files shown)


.github/scripts/generate-release-contributors.sh
```diff
@@ -0,0 +1,80 @@
...
+      hasNextPage
+      endCursor
+    }
+  }
+}' -q '.data.search.edges.[].node.author.login,
+       .data.search.edges.[].node.reviews.nodes.[].author.login,
+       .data.search.edges.[].node.comments.nodes.[].author.login,
+       .data.search.edges.[].node.closingIssuesReferences.nodes.[].author.login')
+
+contributors2=$(gh api graphql --paginate -F q="repo:open-telemetry/opentelemetry-java-instrumentation is:pr base:main is:merged merged:$from..$to" -f query='
```
anuraaga, line 50: Let's add a comment on what this query does, I guess it's to catch referenced issues that are not part of `closingIssuesReferences`

anuraaga (approved):

.github/scripts/generate-release-contributors.sh
```diff
@@ -0,0 +1,80 @@
...
+    pageInfo {
+      hasNextPage
+      endCursor
+    }
+  }
+}
+' -q '.data.search.edges.[].node.body' \
+  | grep -oE "#[0-9]{4,}|issues/[0-9]{4,}" \
+  | grep -oE "[0-9]{4,}" \
+  | xargs -I{} gh api "repos/{owner}/{repo}/issues/{}" --jq '[.user.login, .html_url]' \
```
anuraaga, line 69: Can this one be just `gh issue view`?

anuraaga: Another nit, we use `-q` above and `--jq` here, they're aliases so we should consolidate on either

PR #5211, opened 2022-01-24
Rename ServerSpanNaming to HttpRouteHolder

mateuszrzeszutek:
... and `ServerSpanNameSupplier`/`ServerSpanNameTwoArgSupplier` to `HttpRouteGetter`/`HttpRouteGetter2` - I used the functional libs/languages as an inspiration here, they usually have interfaces with names like `Function2` (e.g. Kotlin).

Another part of #442


trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteGetter2.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package io.opentelemetry.instrumentation.api.instrumenter.http;
+
+import io.opentelemetry.context.Context;
+import javax.annotation.Nullable;
+
+/** An interface for getting the {@code http.route} attribute. */
+@FunctionalInterface
+public interface HttpRouteGetter2<T, U> {
```
anuraaga, line 13: It's not obvious to me that it's better but as a reminder, the JDK idiom is to use `Bi`, not `2`, e.g. `BiFunction` `BiConsumer`. I guess here our naming is constrained by having a fixed `Context` param along with the arbitrary ones

anuraaga: That being said, I guess having these types is to avoid capturing lambdas, but are we sure the performance improvement is really enough to avoid the simpler `Supplier<String>` or `Function<Context, String>`? With inlining and such it's always hard for me to really understand how bad a capturing lambda is

mateuszrzeszutek: > It's not obvious to me that it's better but as a reminder, the JDK idiom is to use `Bi`, not `2`, e.g. `BiFunction` `BiConsumer`

Changed it `HttpRouteBiGetter` (because of `ToDoubleBiFunction` in the JDK)

trask: > With inlining and such it's always hard for me to really understand how bad a capturing lambda is

I opened #5215 (targeted to Stable API project) to make sure that this is an optimization worth keeping

PR #5228, opened 2022-01-25
Fix vertx-http-client testLatestDeps Xlint errors

trask:


instrumentation/vertx/vertx-http-client/vertx-http-client-4.0/javaagent/build.gradle.kts
```diff
@@ -14,8 +14,9 @@ muzzle {
 dependencies {
   library("io.vertx:vertx-core:4.0.0")
 
-  compileOnly("io.vertx:vertx-codegen:4.0.0")
-  testCompileOnly("io.vertx:vertx-codegen:4.0.0")
+  // vertx-codegen and vertx-docgen dependencies are needed for Xlint's annotation checking
+  library("io.vertx:vertx-codegen:4.0.0")
+  testLibrary("io.vertx:vertx-docgen:4.0.0")
```
trask, line 19: this change wasn't necessary, but I think is good so that these library versions will also vary with the `vertx-core` version

anuraaga (approved):

PR #5239, opened 2022-01-26
Use HttpRouteHolder in spring-webflux instrumentation

mateuszrzeszutek:
Part of #442

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
@@ -50,6 +50,24 @@ public final class HttpRouteHolder {
    * Updates the {@code http.route} attribute in the received {@code context}.
    *
    * <p>If there is a server span in the context, and the context has been customized with a {@link
+   * HttpRouteHolder}, then this method will update the route using the provided {@code httpRoute}
+   * if and only if the last {@link HttpRouteSource} to update the route using this method has
+   * strictly lower priority than the provided {@link HttpRouteSource}, and the pased value is
+   * non-null.
+   *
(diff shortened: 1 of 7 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
```diff
@@ -46,6 +46,24 @@
...
+   * if and only if the last {@link HttpRouteSource} to update the route using this method has
+   * strictly lower priority than the provided {@link HttpRouteSource}, and the pased value is
+   * non-null.
+   *
+   * <p>If there is a server span in the context, and the context has NOT been customized with a
+   * {@link HttpRouteHolder}, then this method will update the route using the provided value if it
+   * is non-null.
+   */
+  public static void updateHttpRoute(
+      Context context, HttpRouteSource source, @Nullable String httpRoute) {
```
trask, line 63: do you think this will be used by other instrumentations? also, wondering if HttpRouteGetter0(?) would be more general for this case, which takes Context but no other args, then could use (non-capturing) simple lambda `context -> httpRoute` in this case (though this requires naming, so ...). or maybe Consumer<Context>? sorry for just throwing random ideas here, I don't have any strong opinions on this

trask (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
mateuszrzeszutek: There are a few instrumentations that would make use of this method: ratpack, restlet. I thought that just passing the string would be the simplest solution in this case (and there's no naming problems with just passing a string 😄 ).

PR #5244, opened 2022-01-26
Set custom gRPC client/server span name extractor

eugeniyk:
Typical use case - set short operation name instead of full description name


instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcTracingBuilder.java
```diff
@@ -49,9 +51,28 @@ public GrpcTracingBuilder addAttributeExtractor(
...
+   */
+  public GrpcTracingBuilder setServerSpanNameExtractor(
+      SpanNameExtractor<? super GrpcRequest> serverSpanNameExtractor) {
+    this.serverSpanNameExtractor = serverSpanNameExtractor;
+    return this;
+  }
+
   /** Sets the {@code peer.service} attribute for http client spans. */
-  public void setPeerService(String peerService) {
+  public GrpcTracingBuilder setPeerService(String peerService) {
```
eugeniyk, line 73: Fixing existing builder method

trask (approved):
would like @mateuszrzeszutek or @anuraaga's review

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/GrpcTracingBuilder.java
```diff
@@ -49,9 +52,28 @@ public GrpcTracingBuilder addAttributeExtractor(
     return this;
   }
 
+  /**
+   * Sets custom client {@link SpanNameExtractor}
+   */
+  public GrpcTracingBuilder setClientSpanNameExtractor(
+      SpanNameExtractor<? super GrpcRequest> clientSpanNameExtractor) {
```
mateuszrzeszutek, line 59: Instead of just passing the `SpanNameExtractor`, WDYT about passing a `Function<SpanNameExtractor<GrpcRequest>, ? extends SpanNameExtractor<? super GrpcRequest>>`? (similar to how [the Armeria instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaTracingBuilder.java#L55) allows modifying the `SpanStatusExtractor`)
With a transformer function you can still make use of the original span name extractors - or reject them altogether and return your own, if you wish to do so.

eugeniyk: Sounds interesting
Chaining status code transformations makes more sense imo than extracting span name (which is a string, doubt you'd like to have pattern matching on it)

Or what's the use case you have in mind? One could be if length is more that X use shorter version of method name; since default impl is pretty straight forward I'd stick to the current version, + it gives api consistency with other methods

mateuszrzeszutek: > Or what's the use case you have in mind? One could be if length is more that X use shorter version of method name;

That's one option. I was thinking of a case where you might decide to fall back on the default naming scheme based on the request, e.g. `(request) -> shouldRename(request) ? customName(request) : originalSpanNameExtractor.extract(request)`

eugeniyk: @trask what do you think? current way as api consistency vs suggested flexibility? 

I'm ok with both

trask: let's go with @mateuszrzeszutek's proposal, thx!

trask (approved):
thx @eugeniyk!

if you merge `main` into your branch one more time, it should clear up the windows smoke test failures (#5388)

Issue #5227, opened 2022-01-25
Open issues to track instrumentations that do not run tests against the latest stable library versions


trask:
Discussed in yesterday's SIG, we should start opening issues to track when we do not support latest (stable) releases of an instrumented library.

We should label these, e.g. `test latest deps`(?)

Will close this issue once existing instrumentations have been reviewed and issues have been opened as needed.

trask:
Closing, I think documenting on supported-libraries.md is good enough tracking, and this will be "enforced" by #12945

PR #5258, opened 2022-01-27
Don't run testLatestDeps on alpha/beta/rc versions

### LINKED ISSUE #5225: Don't run testLatestDeps on alpha/etc releases
This would reduce the number of testLatestDeps caps that we need to put in place, and then we won't miss when stable is released.

Muzzle already doesn't run against alpha/etc releases:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/5f4fc25a370de73504acfb122a181557b1ae2b93/gradle-plugins/src/main/kotlin/io/opentelemetry/javaagent/muzzle/AcceptableVersions.kt#L23-L35

trask:
Resolves #5225

Will apply to more places as part of reviewing all testLatestDeps caps in #5227.

Sources for gradle magic:
* https://stackoverflow.com/questions/24995517/how-to-use-gradles-dynamic-versions-and-avoid-betas
* https://docs.gradle.org/current/userguide/component_metadata_rules.html#sec:custom_status_scheme

### DIFF
--- a/conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
+++ b/conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
@@ -30,6 +30,16 @@ plugins {
 val testLatestDeps = gradle.startParameter.projectProperties["testLatestDeps"] == "true"
 extra["testLatestDeps"] = testLatestDeps
 
+@CacheableRule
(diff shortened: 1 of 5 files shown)


anuraaga (approved):
Nice!!

iNikem (approved):

mateuszrzeszutek (approved):

PR #5264, opened 2022-01-28
Log a warning whenever GlobalOpenTelemetry.set() is called

mateuszrzeszutek:
I decided to implement this while I still remember about it 😅 

CC @jack-berg

### DIFF
--- a/instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/OpenTelemetryInstrumentation.java
+++ b/instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/OpenTelemetryInstrumentation.java
@@ -8,6 +8,8 @@ package io.opentelemetry.javaagent.instrumentation.opentelemetryapi;
 import static net.bytebuddy.matcher.ElementMatchers.isMethod;
 import static net.bytebuddy.matcher.ElementMatchers.isStatic;
 import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.returns;
+import static net.bytebuddy.matcher.ElementMatchers.takesArgument;
 import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
 
 import io.opentelemetry.javaagent.extension.instrumentation.TypeInstrumentation;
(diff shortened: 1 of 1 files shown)


laurit (approved):

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/OpenTelemetryInstrumentation.java
```diff
@@ -47,4 +61,18 @@ public static void methodExit(
...
     }
   }
+
+  @SuppressWarnings("unused")
+  public static class SetAdvice {
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter() {
+      LoggerFactory.getLogger(application.io.opentelemetry.api.GlobalOpenTelemetry.class)
+          .warn(
```
trask, line 71: maybe info? (which is the javaagent's default log level, so will still get logged)

iNikem (approved):

instrumentation/opentelemetry-api/opentelemetry-api-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/opentelemetryapi/OpenTelemetryInstrumentation.java
mateuszrzeszutek: I think that `warn` makes slightly more sense in this case (since the log _warns_ you about incorrect API usage), but I don't really have a strong opinion on this. I'll change that if we decide so.

trask: my only concern is about users adding the javaagent later to an existing app that was previously manually instrumented, and then they're stuck with this warning message which is not really actionable

let's keep at `warn` and we can always revisit later

iNikem: > my only concern is about users adding the javaagent later to an existing app that was previously manually instrumented, and then they're stuck with this warning message which is not really actionable
> 
> let's keep at `warn` and we can always revisit later

Why not actionable? They can remove their manual usages of `GlobalOpenTelemetry.set`?

trask: sorry, I should have been more specific, I was thinking about ops users

PR #5266, opened 2022-01-28
Use `http.route` in `HttpServerMetrics`

### LINKED ISSUE #442: `http.route` attribue

mateuszrzeszutek:
This PR actually resolves #442 , i.e. it finishes the implementation of the `http.route` attribute in both metrics and traces. It's still a quite messy solution though; I'll try to refactor it and implement necessary Instrumenter API changes in the following PRs.

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
@@ -146,7 +146,6 @@ public final class HttpRouteHolder {
     return name.length() > routeLength;
   }
 
-  // TODO: use that in HttpServerMetrics
   /**
    * Returns the {@code http.route} attribute value that's stored in the passed {@code context}, or
    * null if it was not set before.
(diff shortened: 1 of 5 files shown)


mateuszrzeszutek:
@trask Do you think we can remove the fallback from `TemporaryMetricsView` now?

trask (approved):
> @trask Do you think we can remove the fallback from `TemporaryMetricsView` now?

👍

PR #5269, opened 2022-01-28
test latest deps cleanup

trask:
part of #5227

and applying some benefits from #5258

this PR adds trailing comments to about half of all `latestDepTestLibrary` usages, categorizing them as

* ` // documented limitation` - version limitation is in [official docs](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/supported-libraries.md#libraries--frameworks)
* ` // issue #X` - issue tracking the limitation, issue should be tagged with `test latest deps`
* ` // see X module` - more recent versions supported by another instrumentation module

will tackle more in subsequent PRs and will see if any additional categories are needed still


instrumentation/jetty/jetty-8.0/javaagent/src/test/groovy/JettyContinuationHandlerTest.groovy
```diff
@@ -1,67 +0,0 @@
...
-import org.eclipse.jetty.server.handler.AbstractHandler
-
-import javax.servlet.ServletException
-import javax.servlet.http.HttpServletRequest
-import javax.servlet.http.HttpServletResponse
-import java.util.concurrent.ExecutorService
-import java.util.concurrent.Executors
-
-// FIXME: We don't currently handle jetty continuations properly (at all).
-abstract class JettyContinuationHandlerTest extends JettyHandlerTest {
```
trask, line 18: Jetty `Continuation` was deprecated in jetty 9.4 and removed in jetty 10, so this test doesn't compile against jetty 10 (see latestDepTestLibrary bump in the gradle file).

Since this test doesn't run anyways (it's abstract) and based on the FIXME we don't support Continuation anyways, it seems ok to just delete it.

instrumentation/elasticsearch/elasticsearch-transport-5.0/javaagent/build.gradle.kts
```diff
@@ -37,8 +37,11 @@ dependencies {
   testImplementation("org.apache.logging.log4j:log4j-core:2.11.0")
   testImplementation("org.apache.logging.log4j:log4j-api:2.11.0")
 
-  testImplementation("org.elasticsearch.plugin:transport-netty3-client:5.0.0")
-  testImplementation("org.elasticsearch.client:transport:5.0.0")
+  testLibrary("org.elasticsearch.plugin:transport-netty3-client:5.0.0")
+  testLibrary("org.elasticsearch.client:transport:5.0.0")
+
+  latestDepTestLibrary("org.elasticsearch.plugin:transport-netty3-client:5.0.0")
```
iNikem, line 43: Fixed patch version?

instrumentation/grails-3.0/javaagent/build.gradle.kts
```diff
@@ -43,8 +43,4 @@ dependencies {
 
   testLibrary("org.springframework.boot:spring-boot-autoconfigure:$springBootVersion")
   testLibrary("org.springframework.boot:spring-boot-starter-tomcat:$springBootVersion")
-
-  // spring boot 3 uses spring 6 which requires java 17
-  latestDepTestLibrary("org.springframework.boot:spring-boot-autoconfigure:2.+")
```
iNikem, line 48: This will still be a problem when Spring 3 is released

instrumentation/lettuce/lettuce-4.0/javaagent/build.gradle.kts
```diff
@@ -14,7 +14,7 @@ muzzle {
 dependencies {
   library("biz.paluch.redis:lettuce:4.0.Final")
 
-  latestDepTestLibrary("biz.paluch.redis:lettuce:4.+")
+  latestDepTestLibrary("biz.paluch.redis:lettuce:4.+") // see lettuce-4.0 module
```
iNikem, line 17:
```suggestion
  latestDepTestLibrary("biz.paluch.redis:lettuce:4.+") // see lettuce-5.0 module
```

iNikem (approved):

mateuszrzeszutek (approved):

instrumentation/elasticsearch/elasticsearch-transport-5.0/javaagent/build.gradle.kts
trask: tests were failing to pass on `5.2.+` so I bailed on it for this PR

instrumentation/grails-3.0/javaagent/build.gradle.kts
trask: is your preference to keep this and open a tracking issue now?

iNikem: I think yes, that will remove a time pressure from us.

instrumentation/elasticsearch/elasticsearch-transport-5.0/javaagent/build.gradle.kts
iNikem: What about `5.1+`?

instrumentation/grails-3.0/javaagent/build.gradle.kts
trask: by "remove time pressure" do you mean it will give us more time to fix the issue and support spring boot 3 before it's GA'd? if this is a concern, maybe we should add some exceptions to #5258?

iNikem: No, I mean that if we keep this constraint, then when Spring Boot 3 is released we will NOT be forced to fix it ASAP because our build broke.

trask: I personally think that's a good thing, it's sort of an alarm bell that Spring Boot 3 has GA'd. It's easy enough to snooze the alarm by adding the suppression + logging the issue at that time.

trask: going to merge, I think removing it is most consistent from #5258 perspective, and I think the alarm bell when unsupported GAs are released is a good thing for us, but we can revisit if this feels problematic

PR #5270, opened 2022-01-28
Add option to run testLatestDeps on PRs

trask:
This will be helpful for the current testLatestDeps cleanup I'm working on.

And seems like a nice alternative on some PRs instead of merging and waiting to see if CI fails on `main`.

We could also use this on brand new instrumentations since we often find out that testLatestDeps fails on those once they are merged.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -114,12 +114,29 @@ jobs:
           path: "**/javacore.*.txt"
           if-no-files-found: ignore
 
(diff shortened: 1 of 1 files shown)


trask:
Another option that may be good is always running `testLatestDeps` for folks in @open-telemetry/java-instrumentation-approvers, since the only reason we don't always run it is to not confuse new contributors

iNikem (approved):

PR #5277, opened 2022-01-29
Add InternalJavadoc custom error prone check

trask:
Error prone looks promising for custom checks.

I was able to hook it up to an individual module, but don't have any intuition for how to hook it up in the `conventions` module, in case anyone has tips...


custom-checks/src/main/java/io/opentelemetry/javaagent/customchecks/InternalJavadoc.java
```diff
@@ -0,0 +1,66 @@
+/*
```
anuraaga, line 1: If I were adding this to the SDK repo (which I do want to do, this looks sweet), I would add it to `buildSrc` and just keep it with our conventions (they are conventions). Can we do that here too?

custom-checks/src/test/java/io/opentelemetry/javaagent/customchecks/InternalJavadocTest.java
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright The OpenTelemetry Authors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.customchecks;
+
+import com.google.errorprone.CompilationTestHelper;
+import org.junit.jupiter.api.Test;
+
+public class InternalJavadocTest {
```
anuraaga, line 11:
```suggestion
class InternalJavadocTest {
```

custom-checks/src/test/java/io/opentelemetry/javaagent/customchecks/InternalJavadocTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+  @Test
+  public void test() {
```
anuraaga, line 14:
```suggestion
  void test() {
```

custom-checks/src/test/java/io/opentelemetry/javaagent/customchecks/InternalJavadocTest.java
```diff
@@ -0,0 +1,24 @@
...
+    test("internal/InternalJavadocPositiveCases.java");
+    test("internal/InternalJavadocNegativeCases.java");
+  }
+
+  private static void test(String path) {
```
anuraaga, line 19: I think it's worth having a different method name maybe `doTest(String path)`

iNikem:
We already have `@UnstableApi`. Should we verify and require its presence instead of the javadoc?

trask:
> We already have `@UnstableApi`. Should we verify and require its presence instead of the javadoc?

To me, `@UnstableApi` implies a public API that is working towards stability, so go ahead and use it if you really want to, and feel free to give us feedback if you run into issues.

And so I think the stronger message of "not for public use" is useful.

But really, I just want all the otel-java* repos to follow the same convention, whatever it is, so we'd need buy-in from @open-telemetry/java-maintainers since this is just codifying the convention that they follow already.

jack-berg:
I agree with @trask's assessment - there are classes in the SDK which are not ever intended for public use, but are marked `public` because code reuse necessitates it. Also, we don't have `@UnstableApi` in the SDK.

custom-checks/src/main/java/io/opentelemetry/javaagent/customchecks/InternalJavadoc.java
trask: I tried but was unable to get that to work, see commit history, the problem I had was that errorprone doesn't see the custom check just because its on the build classpath, it needs to be added as an annotation processor, and I couldn't figure out how to do that.

build.gradle.kts
```diff
@@ -35,3 +35,13 @@ nexusPublishing {
 // tasks.release.finalizedBy tasks.closeAndReleaseRepository
 
 description = "OpenTelemetry instrumentations for Java"
+
+subprojects {
+  if (path != ":instrumentation:liberty:compile-stub" && !path.startsWith(":instrumentation:internal:")) {
```
trask, line 40: applying it to `:instrumentation:internal:` modules made me put the comment on Advice classes which felt weird, so figured I should at least do that in a separate PR to get feedback separately

iNikem:
Then maybe new `@Internal` annotation? My point is that annotation is friendlier to tooling, javadoc is easy to miss.

iNikem:
> Then maybe new `@Internal` annotation? My point is that annotation is friendlier to tooling, javadoc is easy to miss.

After today's meeting I withdraw my proposal. We already have package name for any future tooling to check.

custom-checks/src/main/java/io/opentelemetry/javaagent/customchecks/InternalJavadoc.java
```diff
@@ -0,0 +1,70 @@
...
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.PackageTree;
+import com.sun.tools.javac.api.JavacTrees;
+import java.util.regex.Pattern;
+import javax.annotation.Nullable;
+import javax.lang.model.element.Modifier;
+
+@AutoService(BugChecker.class)
+@BugPattern(
+    name = "InternalJavadoc",
```
mateuszrzeszutek, line 25: Nit: WDYT about "OtelInternalJavadoc"? We could prefix all our custom rules with "Otel"

mateuszrzeszutek (approved):

custom-checks/src/main/java/io/opentelemetry/javaagent/customchecks/InternalJavadoc.java
trask: I think this is a great idea, will update 👍

trask:
going ahead and merging, happy to address any additional feedback after the fact

Issue #5265, opened 2022-01-28
javaagent causes big heap on high load for Netty instrumentation 


ozoz03:
**Javaagent produces a critical heap increase for Netty instrumentation on our prod env (heap grows persistently over days and is not cleaned up by GC).** 
We was able to reproduce the issue by a [test app with artillery load](https://github.com/ozoz03/netty-otel-memory-leak). 

<img width="858" alt="Item-0_and_Item-1_and_Item-0_and_Item-0" src="https://user-images.githubusercontent.com/7068169/151535736-62840710-ca51-483c-ae79-68bd1c4e2e8d.png">

**Looks like cache that contains `FutureListenerWrappers` is not properly maintained by GC:**
<img width="1076" alt="Eclipse_Memory_Analyzer" src="https://user-images.githubusercontent.com/7068169/151538039-b916f213-ffa9-4da4-8e26-547addf47321.png">

**Steps to reproduce**
Here are steps to reproduce: https://github.com/ozoz03/netty-otel-memory-leak/blob/master/README.md

**What version are you using?**
`v1.10.0`

**Environment**
Compiler: openjdk version "11.0.11" 2021-04-20 LTS
OS: macOS 11.6.2, Alpine Linux 3.12.8

iNikem:
How large heap do you use in your production?

ozoz03:
@iNikem 6g

ozoz03:
@iNikem 6g

laurit:
@ozoz03 If possible could you verify whether the issue you observed is resolved in nightly build https://oss.sonatype.org/content/repositories/snapshots/io/opentelemetry/javaagent/opentelemetry-javaagent/1.11.0-SNAPSHOT/

ozoz03:
@laurit thank you a lot.
going to test the build.
will be back in 2-3 days.

ozoz03:
@laurit hey, sorry for the delay.
 
from what we could see in our graphs the memory leak was still in place with the new snapshot version.
but we believe we found the root cause for the issue and proposed the fix in [this PR](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5316)

PR #5282, opened 2022-01-31
Use VirtualField for associating netty listener with wrapper

laurit:
Probably resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5265
To support removing netty listeners we need to be able to find the wrapper that we register instead of the original listener from the original listener. Currently we use a static weak cache for it. Removal of the elements from this cache depends on the gc, which might confuse users who inspect the heap dump and make them suspect that there is a memory leak.

### DIFF
--- a/instrumentation/internal/internal-lambda-java9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/lambda/Java9LambdaTransformer.java
+++ b/instrumentation/internal/internal-lambda-java9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/lambda/Java9LambdaTransformer.java
@@ -5,6 +5,7 @@
 
 package io.opentelemetry.javaagent.instrumentation.internal.lambda;
 
+import io.opentelemetry.javaagent.bootstrap.InjectedHelperClassDetector;
 import java.lang.instrument.ClassFileTransformer;
 import java.lang.instrument.IllegalClassFormatException;
 
@@ -19,6 +20,10 @@ public final class Java9LambdaTransformer {
       String slashClassName,
       Class<?> targetClass)
       throws IllegalClassFormatException {
+    // Skip transforming lambdas of injected helper classes.
+    if (InjectedHelperClassDetector.isHelperClass(targetClass)) {
+      return classBytes;
+    }
     return transformer.transform(
         targetClass.getModule(),
         targetClass.getClassLoader(),
--- a/instrumentation/internal/internal-lambda-java9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/lambda/Java9LambdaTransformer.java
+++ b/instrumentation/internal/internal-lambda-java9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/lambda/Java9LambdaTransformer.java
@@ -5,7 +5,6 @@
 
 package io.opentelemetry.javaagent.instrumentation.internal.lambda;
 
-import io.opentelemetry.javaagent.bootstrap.InjectedHelperClassDetector;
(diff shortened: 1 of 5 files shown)


mateuszrzeszutek (approved):

instrumentation/internal/internal-lambda-java9/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/lambda/Java9LambdaTransformer.java
```diff
@@ -19,6 +20,10 @@ private Java9LambdaTransformer() {}
       String slashClassName,
       Class<?> targetClass)
       throws IllegalClassFormatException {
+    // Skip transforming lambdas of injected helper classes.
+    if (InjectedHelperClassDetector.isHelperClass(targetClass)) {
+      return classBytes;
+    }
```
mateuszrzeszutek, line 26: Don't we need to check that in the `LambdaTransformer` too? (or instead)

laurit: Yes, it should be moved to `LambdaTransformer`

mateuszrzeszutek (approved):

trask (commented):
I didn't follow the connection to skipping transforming lambdas of injected helper classes?

laurit:
@trask I should have added a comment for this. As far as I understand we don't transform helper classes, so not transforming their lambdas should really be the expected behaviour. Transforming lambdas of helper classes can fail because some bytebuddy matcher tries to resolved the class file of containing class of lambda, which does not exist, because it is an injected class. We make tests fail when there is a transformation failure. Adding a virtual field for `GenericFutureListener` causes such failures.

trask:
oh, i c, and non-lambda injected classes are skipped already because they live under "io.opentelemetry.javaagent.shaded" which we skip in `GlobalIgnoredTypesConfigurer`

would it work to move the `isHelperClass` check up to `GlobalIgnoredTypesConfigurer`?

laurit:
> oh, i c, and non-lambda injected classes are skipped already because they live under "io.opentelemetry.javaagent.shaded" which we skip in `GlobalIgnoredTypesConfigurer`
> 
> would it work to move the `isHelperClass` check up to `GlobalIgnoredTypesConfigurer`?

I thought it was because the injection happens during transformation and transformation isn't re-entered. Only helper classes from library modules have "shaded".
Using `GlobalIgnoredTypesConfigurer` would require some kind of heuristics as you need to configure it before classes are actually injected. We would need to use something like class name starts with "io.opentelemetry.javaagent." and ends with "$Lambda*"

trask:
oh of course, I forgot that `GlobalIgnoredTypesConfigurer` is just used to build up the trie 👍

trask (approved):

trask:
oh yes, and this:

> I thought it was because the injection happens during transformation and transformation isn't re-entered

and I missed that the check isn't whether the lambda class itself is a helper class, but rather it's whether the lambda's `targetClass` is a helper class.

btw, what is the `targetClass`? is it the "outer" class where the lambda lives? or the target interface of the lambda?

thx 😅

laurit:
> btw, what is the `targetClass`? is it the "outer" class where the lambda lives? or the target interface of the lambda?
> 
> thx 😅

`targetClass` is the class that declares the lambda. Not the best name. It has the same value as https://github.com/openjdk/jdk11/blob/37115c8ea4aff13a8148ee2b8832b20888a5d880/src/java.base/share/classes/java/lang/invoke/AbstractValidatingLambdaMetafactory.java#L54

PR #5284, opened 2022-01-31
Set jackson.databind as 'implementation' dep for Lambda Instrumentation

### LINKED ISSUE #5285: Lambda instrumentation requires user to add jackson dependency
It seems that the lambda instrumentation doesn't work without the user adding jackson to their classpath

https://github.com/open-telemetry/opentelemetry-lambda/pull/211

My understanding which seems to have been off was that jackson is already part of the lambda runtime so this wouldn't be needed. If that's incorrect, either our instrumentation should have the dependency in the POM or we should try to allow the code to work by possibly skipping the jackson codepath if it's not available.

NathanielRN:
## Description

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4254 added new imports but missed including them as `compileOnly` dependencies in the `io.opentelemetry.instrumentation:opentelemetry-aws-lambda-1.0` instrumentation package.

We add them here so downstream packages which pull it don't have to add them explicitly.

cc @kubawach

Fixes #5285

### DIFF
--- a/instrumentation/aws-lambda-1.0/library/build.gradle.kts
+++ b/instrumentation/aws-lambda-1.0/library/build.gradle.kts
@@ -16,10 +16,10 @@ dependencies {
   // in public API.
   library("com.amazonaws:aws-lambda-java-events:2.2.1")
 
-  compileOnly("com.fasterxml.jackson.core:jackson-databind")
   compileOnly("commons-io:commons-io:2.2")
   compileOnly("org.slf4j:slf4j-api")
 
+  implementation("com.fasterxml.jackson.core:jackson-databind")
   implementation("io.opentelemetry:opentelemetry-extension-aws")
 
   // allows to get the function ARN


anuraaga:
@kubawach Can you review this and context in #5285? Ideally we didn't need to require the jackson dependency, I'm wondering 1) Was lambda runtime supposed to have included Jackson? and 2) Should we instead change the codepath to only parse JSON if jackson is available rather than requiring Jackson

NathanielRN:
Ping for @kubawach when you have time to review this and the issue!

anuraaga:
@NathanielRN I ended up incorporating this change into #5326 which fixes this and another bigger issue #5325

NathanielRN:
@anuraaga Thanks for taking that on! I didn’t realize how big a change would be needed 😅 Looking forward to the fix!

PR #5287, opened 2022-02-01
Switch to temurin distributions where possible in GitHub actions

iNikem:
Supersedes #4940


laurit:
Thanks, we have completed the switch to Eclipse Temurin.

PR #5288, opened 2022-02-01
Refactor AttributesExtractor so that it extracts route from Context

mateuszrzeszutek:
I added the `Context` as a parameter to the `AttributesExtractor` methods - some of the other interfaces already accept `Context` (e.g. `SpanLinksExtractor`, `RequestListener`) so it doesn't that look out of place. Still, `onEnd()` now accepts 5 params...

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
@@ -7,6 +7,7 @@ package io.opentelemetry.instrumentation.api.instrumenter;
 
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.context.Context;
 import io.opentelemetry.instrumentation.api.instrumenter.db.DbAttributesExtractor;
 import io.opentelemetry.instrumentation.api.instrumenter.http.HttpClientAttributesExtractor;
 import io.opentelemetry.instrumentation.api.instrumenter.net.NetServerAttributesExtractor;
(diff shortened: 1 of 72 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/AttributesExtractor.java
```diff
@@ -25,21 +26,56 @@
...
+    onStart(attributes, request);
+  }
+
   /**
    * Extracts attributes from the {@link REQUEST} into the {@link AttributesBuilder} at the
    * beginning of a request.
    */
-  void onStart(AttributesBuilder attributes, REQUEST request);
+  // * @deprecated Use {@link #onStart(AttributesBuilder, Context, Object)} instead.
+  // @Deprecated
```
mateuszrzeszutek, line 45: Temporarily commented out to appease errorprone

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerAttributesExtractor.java
```diff
@@ -66,12 +74,14 @@ public void onStart(AttributesBuilder attributes, REQUEST request) {
...
       AttributesBuilder attributes,
+      Context context,
       REQUEST request,
       @Nullable RESPONSE response,
       @Nullable Throwable error) {
 
-    super.onEnd(attributes, request, response, error);
+    super.onEnd(attributes, context, request, response, error);
     set(attributes, SemanticAttributes.HTTP_SERVER_NAME, getter.serverName(request, response));
+    set(attributes, SemanticAttributes.HTTP_ROUTE, httpRouteHolderGetter.apply(context));
```
mateuszrzeszutek, line 84: I thought about adding a `route(context, request, response)` to the getter interface (with a default `HttpRouteHolder` implementation) but I decided against it: `HttpRouteHolder` not only stores the route, but also renames the server span whenever the route is updated; if we were to allow somebody to implement a `route(context, request, response)` method that overrode the default implementation (and skipped the `HttpRouteHolder` altogether), that instrumentation would not rename the server span after start -- and that would differ from how the `HttpRouteHolder` handles things.

In other words: using `route(request)` is fine, because the `HttpSpanNameExtractor` handles that, but if you want to set the route during/at the end of the request you have to use `HttpRouteHolder`.

mateuszrzeszutek:
I made it a draft PR for now -- if this looks acceptable I'll uncomment the `@Deprecated` annotations and fix all other attributes extractors.

anuraaga (commented):
LGTM

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpRouteHolder.java
```diff
@@ -111,7 +111,9 @@ public static void updateHttpRoute(
     if (httpRouteHolder == null) {
       String httpRoute = httpRouteGetter.get(context, arg1, arg2);
       if (httpRoute != null && !httpRoute.isEmpty()) {
-        updateSpanData(serverSpan, httpRoute);
+        // update both span and name, since there's no HttpRouteHolder in the context
```
trask, line 114:
```suggestion
        // update both span name and attribute, since there's no HttpRouteHolder in the context
```

trask (commented):
makes sense 👍

instrumentation/lettuce/lettuce-5.1/library/src/main/java/io/opentelemetry/instrumentation/lettuce/v5_1/OpenTelemetryTracing.java
```diff
@@ -194,11 +193,22 @@ private OpenTelemetrySpan nextSpan(Context context) {
...
-        fillEndpoint(span, spanBuilder, (OpenTelemetryEndpoint) endpoint);
+        fillEndpoint((OpenTelemetryEndpoint) endpoint);
       }
       return this;
     }
 
+    private void fillEndpoint(OpenTelemetryEndpoint endpoint) {
+      AttributesBuilder attributesBuilder = Attributes.builder();
+      Context currentContext = span == null ? context : context.with(span);
+      netAttributesExtractor.onEnd(attributesBuilder, currentContext, endpoint, null, null);
```
trask, line 204: using `Context.root()` seems like it would be ok for this odd usage, but ok with this too

trask (approved):
nice

PR #5297, opened 2022-02-02
Fix triggering of DNS lookup

### LINKED ISSUE #5295: 10s latency introduced when no external network connectivity
**Describe the bug**
There appears to be some sort of blocking occurring when there is no connection available for the application(disconnect WIFI...)

**Steps to reproduce**
[REPRO](https://github.com/necrolyte2/ratpack-kotlin-otel/releases/tag/7.0.0)

I tested using the above with versions 1.5.2, 1.6.2, 1.7.0, 1.7.2, 1.9.2, and 1.10.1
1.5.2 and 1.6.2 did not have the issue, but all other versions >= 1.7.0 did

Calling the endpoint should take less than 100ms

**What did you see instead?**
A 10s latency is introduced

Any version >= 1.7.0

**Environment**
Compiler: "AdoptOpenJDK 11

**Additional context**
The environment where our apps run have scenarios where they lose connectivity to the internet frequently which is where this bug was uncovered

trask:
Resolves #5295

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetClientAttributesGetter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetClientAttributesGetter.java
@@ -29,9 +29,6 @@ public abstract class InetSocketAddressNetClientAttributesGetter<REQUEST, RESPON
     if (address == null) {
       return null;
     }
-    if (address.getAddress() != null) {
-      return address.getAddress().getHostName();
-    }
     return address.getHostString();
   }
 
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetServerAttributesGetter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetServerAttributesGetter.java
@@ -29,9 +29,6 @@ public abstract class InetSocketAddressNetServerAttributesGetter<REQUEST>
     if (address == null) {
       return null;
     }
-    if (address.getAddress() != null) {
-      return address.getAddress().getHostName();
-    }
     return address.getHostString();
   }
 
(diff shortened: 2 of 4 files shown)


iNikem (approved):

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetClientAttributesGetter.java
```diff
@@ -29,9 +29,6 @@ public final String peerName(REQUEST request, @Nullable RESPONSE response) {
     if (address == null) {
       return null;
     }
-    if (address.getAddress() != null) {
-      return address.getAddress().getHostName();
-    }
     return address.getHostString();
```
laurit, line 32: I believe `getHostString` may return ip address. I hope spec is ok with this, if it isn't I guess we could compare it with `getHostAddress` and return null in case they match.

laurit (approved):
`RabbitChannelNetAttributesGetter`, `RabbitReceiveNetAttributesGetter` and `ReactorNettyHttpClientAttributesGetter` might have the same issue.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/InetSocketAddressNetClientAttributesGetter.java
mateuszrzeszutek: We're already doing that: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/net/NetClientAttributesExtractor.java#L51-L53

necrolyte2 (approved):

trask:
> `RabbitChannelNetAttributesGetter`, `RabbitReceiveNetAttributesGetter` and `ReactorNettyHttpClientAttributesGetter` might have the same issue.

thx! I'll fix those in a follow-up, since PR builds for those should be faster than PR builds on this one which touches core classes

PR #5300, opened 2022-02-02
Use IBM Semeru OpenJ9 docker images

trask:
Unfortunately doesn't help with #5051


iNikem:
https://github.com/docker-library/official-images/pull/10666#issuecomment-1028652046

laurit:
Imo not testing with ibm vm on windows is also a valid option.

PR #5302, opened 2022-02-03
Avoid having two tasks producing the same file

iNikem:


javaagent/build.gradle.kts
```diff
@@ -218,6 +218,12 @@ tasks {
   named("generateLicenseReport").configure {
     dependsOn(cleanLicenses)
   }
+
+  // We don't publish this jar file, so the actual classifier value is not important.
+  // But we should not have two tasks (jar and shadow) that produce the same file.
+  jar {
+    archiveClassifier.set("slim")
```
iNikem, line 225: Taking suggestions for the classifier value :)

mateuszrzeszutek: `empty-do-not-use`? 😄

mateuszrzeszutek (approved):

javaagent/build.gradle.kts
laurit: Considering https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5251 slim doesn't look like the best name for now. If this is an intermediate jar that isn't really functional then maybe `-tmp` or something else that hints that this is not supposed to be used.

iNikem: Now only to understand, why smoke test crashes...

iNikem:
@trask I think that #5129 was a mistake:

First, `jar` is NOT needed for producing javadoc and source jars, they are totally independent.
Second, by having `jar` task with a classifier, that jar gets published and thus examples get that thin jar when they request `io.opentelemetry.javaagent:opentelemetry-javaagent`.
Third, if we have `jar` without any classifier, then things kinda work, but only by accident. See (https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4765#issuecomment-983679435).

What was the failure that #5129 tried to solve? I still think that #4765 was and is correct.

trask:
> What was the failure that #5129 tried to solve?

maven central publishing failed during release because there was no `-javadoc` or `-sources` artifact for the `opentelemetry-javaagent` artifact, so I just reverted #4765, but I'm good with solving it a different way

iNikem:
I will postpone this change. Don't want to mess with release process while my brain is 90% elsewhere...

anuraaga:
I had issues with jmx-metrics project in contrib being slow so looked into this, should be superceded by #5334

Issue #5292, opened 2022-02-01
Cannot get Micrometer custom metrics


irizzant:
**Describe the bug**
I have a Java application instrumented to create custom Micrometer metrics.
The very same Java application uses the OpenTelemetry Java Instrumentation agent to collect metrics using the Prometheus exporter.

I can see Micrometer metrics and OpenTelemetry metrics being exported just fine, but I cannot see Micrometer metrics exported via OpenTelemetry Prometheus exporter.

It's my understanding that from 1.10 the agent should export Micrometer metrics but maybe I get this wrong.

Agent is configured like this:
```
-Dotel.service.name=webportal
-Dotel.exporter.otlp.endpoint=http://tempo.tempo.svc.cluster.local:4317
-Dotel.traces.exporter=otlp
-Dotel.exporter.otlp.compression=gzip
-Dotel.metrics.exporter=prometheus
```

Querying port 9464 produces OTEL metrics like the following:
```
http_client_duration_sum{http_flavor="1.1",http_method="GET",http_status_code="200",http_url="http://localhost:8080/webtng/ediMonitorREST_1_0_0/status",} 301.605403 1643730489470
http_client_duration_bucket{http_flavor="1.1",http_method="GET",http_status_code="200",http_url="http://localhost:8080/webtng/ediMonitorREST_1_0_0/status",le="5.0",} 0.0 1643730489470
http_client_duration_bucket{http_flavor="1.1",http_method="GET",http_status_code="200",http_url="http://localhost:8080/webtng/ediMonitorREST_1_0_0/status",le="10.0",} 0.0 1643730489470
http_client_duration_bucket{http_flavor="1.1",http_method="GET",http_status_code="200",http_url="http://localhost:8080/webtng/ediMonitorREST_1_0_0/status",le="25.0",} 0.0 1643730489470
```

Querying the Mircometer port I get the following:
```
edimonitor_inmessage_incorrect_total{application="edimonitor",edimonitorinstance="0",edimonitorjob="JOB",} 0.0
```
but I cannot find the latter metrics inside the former ones.

**Steps to reproduce**
1. instrument a Java application with custom Java Micrometer metrics
2. auto-instrument a Java application with OTEL agent
3. enable Prometheus exporter for OTEL
4. verify that Micrometer metrics are exported via OTEL agent

**What did you expect to see?**
Metrics from Micrometer reported in the Prometheus exporter endpoint

**What did you see instead?**
[...]

PR #5303, opened 2022-02-03
Add max measurements to Micrometer Timer & DistributionSummary

mateuszrzeszutek:
Fixes one of the issues discovered in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5292

### DIFF
--- a/instrumentation/micrometer/micrometer-1.5/testing/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AbstractTimerTest.java
+++ b/instrumentation/micrometer/micrometer-1.5/testing/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AbstractTimerTest.java
@@ -13,6 +13,7 @@ import io.micrometer.core.instrument.Timer;
 import io.opentelemetry.instrumentation.testing.junit.InstrumentationExtension;
 import java.time.Duration;
 import java.util.concurrent.TimeUnit;
+import org.assertj.core.api.AbstractIterableAssert;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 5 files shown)


trask (approved):

instrumentation/micrometer/micrometer-1.5/testing/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AbstractTimerTest.java
```diff
@@ -197,4 +198,49 @@ void testMicrometerPercentiles() {
...
+                            .anySatisfy(
+                                point ->
+                                    assertThat(point)
+                                        .hasValue(4_000)
+                                        .attributes()
+                                        .containsEntry("tag", "value"))));
+
+    // when
+    Metrics.globalRegistry.remove(timer);
+    Thread.sleep(10); // give time for any inflight metric export to be received
```
anuraaga, line 237: Will be able to remove this after #5314 I think

anuraaga (approved):

instrumentation/micrometer/micrometer-1.5/testing/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/AbstractTimerTest.java
trask: I'm guessing there may still be a tiny window while the metric sdk is still collecting metrics but before it has sent them to the metric exporter(?)

anuraaga: At time of writing, that PR was causing clearData to first flush the metrics SDK. But realized it's not necessary or helpful in most tests so changed it back, so indeed this comment is now invalid

trask: oh i c, it flushed the full collection cycle, not just the export? that does seem useful for all of these micrometer unregister tests that we have

PR #5305, opened 2022-02-03
Avoid potential for reverse name lookup

trask:
Follow-up on @laurit's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5297#pullrequestreview-871697661

### DIFF
--- a/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitChannelNetAttributesGetter.java
+++ b/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitChannelNetAttributesGetter.java
@@ -20,7 +20,8 @@ public class RabbitChannelNetAttributesGetter
   @Nullable
   @Override
   public String peerName(ChannelAndMethod channelAndMethod, @Nullable Void unused) {
-    return channelAndMethod.getChannel().getConnection().getAddress().getHostName();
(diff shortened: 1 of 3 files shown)


anuraaga:
I have a feeling this method doesn't have a use case in our codebase - is it possible to add a custom error prone check for it?

instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitChannelNetAttributesGetter.java
```diff
@@ -20,7 +20,8 @@ public String transport(ChannelAndMethod channelAndMethod, @Nullable Void unused
   @Nullable
   @Override
   public String peerName(ChannelAndMethod channelAndMethod, @Nullable Void unused) {
-    return channelAndMethod.getChannel().getConnection().getAddress().getHostName();
```
laurit, line 23: Apparently it is possible to read the cached hostname from `InetAddress` with `new InetSocketAddress(channelAndMethod.getChannel().getConnection().getAddress(), 0).getHostString()`. In this case it doesn't seem to change anything because the address really only has the ip. Added this note in case you find some other place where this could come handy.

laurit (approved):
All server instrumentations that implement `peerName` probably also cause a name lookup.

mateuszrzeszutek (approved):

trask:
> All server instrumentations that implement `peerName` probably also cause a name lookup.

I opened #5317 to track this

PR #5308, opened 2022-02-04
Apply conventions to custom-checks project

anuraaga:
The custom-checks project is missing important conventions such as the `release` version of the artifact, so currently the project doesn't build when running the build with a newer version of Java than 11.

Luckily it's not too tedious to solve the circular reference.


custom-checks/src/main/java/io/opentelemetry/javaagent/customchecks/InternalJavadoc.java
```diff
@@ -60,8 +62,9 @@ private static boolean isInternal(VisitorState state) {
     if (packageTree == null) {
       return false;
     }
-    String packageName = packageTree.getPackageName().toString();
-    return INTERNAL_PACKAGE_PATTERN.matcher(packageName).find()
+    String packageName = state.getSourceForNode(packageTree.getPackageName());
```
trask, line 65: 👍

trask (approved):

PR #5309, opened 2022-02-04
Allow lambda instrumentation to work without events on classpath.

anuraaga:
This feels like spaghetti to me but nothing better comes to mind


instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestHandler.java
```diff
@@ -66,7 +67,7 @@ protected TracingRequestHandler(
 
   private Map<String, String> getHeaders(I input) {
     Map<String, String> result = null;
-    if (input instanceof APIGatewayProxyRequestEvent) {
+    if (AwsLambdaUtil.hasEvents() && input instanceof APIGatewayProxyRequestEvent) {
```
laurit, line 70: `javaagent` still requires events, is this something we care about?

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
```diff
@@ -21,9 +21,18 @@
 abstract class TracingRequestWrapperBase<I, O> extends TracingRequestHandler<I, O> {
 
   protected static final ObjectMapper OBJECT_MAPPER =
-      new ObjectMapper()
-          .registerModule(new CustomJodaModule())
-          .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+      new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
laurit, line 24: `aws-lambda-java-core` doesn't depend on jackson, would this also be an issue?

laurit (approved):

instrumentation/aws-lambda-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambda/v1_0/TracingRequestWrapperBase.java
anuraaga: Yeah think so. It looks like the best thing is to go ahead and switch this into lambda core and lambda events instrumentation, otherwise things get too hectic. Will try that instead.

anuraaga: Realized jackson is OK since the lambda runtime itself includes jackson, but it doesn't include events or joda.

Issue #4697, opened 2021-11-23
HttpUrlConnection instrument exception


wangzlei:
**Describe the bug**
opentelemetry-javaagent gets HttpUrlConnection exception on AWS EC2 host.

**Steps to reproduce**
Collector config, enable AWS XRay proxy to fetch remote sampling rules
```yml
extensions:
  awsproxy:
    endpoint: 0.0.0.0:2000

receivers:
  otlp:
    protocols:
      grpc:

exporters:
  logging:
    loglevel: debug
  awsxray:

service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [logging, awsxray]
  extensions: [awsproxy]
  telemetry:
    logs:
      level: "debug"
```
docker composite, launches a simple spring server with OTel javaagent
```yml
version: "3.6"
services:
  otel:
    image: otel/opentelemetry-collector-contrib:latest
    command: --config /config/collector3.yml
    volumes:
      - /home/ec2-user:/config
    environment:
      AWS_REGION: us-west-2
    ports:
      - '4317:4317'
      - '13133:13133'
      - '2000:2000'
  app:
    image: myapp:latest
    command: java -Dotel.javaagent.debug=true -javaagent:./aws-opentelemetry-agent.jar -jar ./build/libs/spring-boot-0.0.1-SNAPSHOT.jar Application
    environment:
      OTEL_SERVICE_NAME: MyEC2App
      OTEL_TRACES_SAMPLER: xray
      OTEL_TRACES_SAMPLER_ARG: 'endpoint=http://otel:2000'
      OTEL_TRACES_EXPORTER: otlp
      OTEL_EXPORTER_OTLP_ENDPOINT: http://otel:4317
    ports:
      - '8080:8080'
    depends_on:
      - otel
```

**What did you expect to see?**
No exception for HttpUrlConnection.
Collector should relay SamplingRules and SamplingTargets requests from application to AWS XRay 

**What did you see instead?**
application error
```
[otel.javaagent 2021-11-23 02:02:52:862 +0000] [xray-rules-poller] DEBUG io.opentelemetry.sdk.internal.JavaVersionSpecific - Using the APIs optimized for: Java 9+
[...]

PR #5314, opened 2022-02-07
Migrate from gRPC to armeria for testing agent in memory exporter

anuraaga:
There is currently a big gap between the published agent and our tests because for historical reasons we include gRPC in the testing exporter. This switches it to Armeria to avoid the gRPC dependency, which automatically switches the tests to using the okhttp export codepath.


testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/AgentTestingLogsCustomizer.java
```diff
@@ -11,21 +11,24 @@
...
 import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import io.opentelemetry.sdk.logs.LogProcessor;
 import io.opentelemetry.sdk.logs.SdkLogEmitterProvider;
-import io.opentelemetry.sdk.logs.export.BatchLogProcessor;
+import io.opentelemetry.sdk.logs.export.SimpleLogProcessor;
 
 @AutoService(AgentListener.class)
 public class AgentTestingLogsCustomizer implements AgentListener {
 
+  static final LogProcessor logProcessor =
```
anuraaga, line 21: Changed this to simple log processor, seems to have sped up the log tests a bit

testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/AgentTestingExporterFactory.java
```diff
@@ -25,6 +28,15 @@
     return logExporter.getCollectedExportRequests();
   }
 
+  public static void forceFlush() {
+    // TODO(anuraaga): Flush metrics too.
+    List<CompletableResultCode> results =
+        Arrays.asList(
+            AgentTestingTracingCustomizer.spanProcessor.forceFlush(),
+            AgentTestingLogsCustomizer.logProcessor.forceFlush());
+    CompletableResultCode.ofAll(results).join(10, TimeUnit.SECONDS);
```
mateuszrzeszutek, line 37: WDYT about using `OpenTelemetrySdkAccess` instead? It flushes the meter provider too

mateuszrzeszutek (approved):

trask (approved):
grpc without grpc 🤯

testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/AgentTestingExporterFactory.java
anuraaga: Oops realized that we don't want to flush metrics here anyways (there's no such thing as pending metric exports really)

anuraaga:
This PR seems to cause a core dump on Java 15 OpenJ9 reliably, anyone have an idea what could cause it?

https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/5104586391?check_suite_focus=true
https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/5103974755?check_suite_focus=true

trask:
> This PR seems to cause a core dump on Java 15 OpenJ9 reliably, anyone have an idea what could cause it?

ya this is weird. it seems to consistently seg fault on the dubbo and camel tests, and none of the others.

and there's no consistency in the "current thread" in the core dumps between runs, which is normally what i'd use to search on to see if it's a known bug

and there's a lot of [openj9 segfault issues](https://github.com/eclipse-openj9/openj9/issues?q=is%3Aissue+label%3Asegfault)

and even if we could run integration tests on Java 17 today, I don't think we could run them on openj9 17 due to https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5051#issuecomment-1028979378 and https://adoptopenjdk.net/archive.html?variant=openjdk8&jvmVariant=openj9

maybe it's best to go with the `@SuppressWarnings` route you mentioned yesterday for updating to OTel SDK 1.11.0 and postponing this PR for a bit?

i'm also ok with skipping those two tests on openj9 15 if we want to move forward with this PR

laurit:
I played around with this a bit. Firstly reporting this to openj9 is probably futile. Afaik they build all their runtimes from the same code base and if the bug was still there then there is a good chance 8 & 11 would also fail similarly.
The bug goes away (tested with `:instrumentation:apache-dubbo-2.7:javaagent:test`) when executor instrumentation is disabled or when in `AgentTestingCustomizer`
```java
    autoConfigurationCustomizer.addMeterProviderCustomizer(
        (meterProvider, config) ->
            meterProvider.registerMetricReader(
                PeriodicMetricReader.builder(AgentTestingExporterFactory.metricExporter)
                    .setInterval(Duration.ofMillis(300))
                    .newMetricReaderFactory()));
```
is commented out or when in `AgentInstaller`
```java
    // If noop OpenTelemetry is enabled, autoConfiguredSdk will be null and AgentListeners are not
    // called
    AutoConfiguredOpenTelemetrySdk autoConfiguredSdk = null;
    if (config.getBoolean(JAVAAGENT_NOOP_CONFIG, false)) {
      logger.info("Tracing and metrics are disabled because noop is enabled.");
      GlobalOpenTelemetry.set(NoopOpenTelemetry.getInstance());
    } else {
      autoConfiguredSdk = installOpenTelemetrySdk(config);
    }

    if (autoConfiguredSdk != null) {
      runBeforeAgentListeners(agentListeners, config, autoConfiguredSdk);
    }
```
is moved after byte-buddy instrumenter is set up or when in `AbstractExecutorInstrumentation` `java.util.concurrent.ForkJoinPool` is removed from included executors list.
The key to this crash probably has something to to with `java.util.concurrent.ForkJoinPool`. While commenting out `addMeterProviderCustomizer` seems somewhat random I think it is still indirectly related to `ForkJoinPool`. Metrics send data to armeria which uses caffeine bounded cache, which uses `ForkJoinPool`. Changing interval `Duration.ofSeconds(3)` seems to also avoid the crash (idk maybe just makes it not repro on each run).
Similarly moving code around in `AgentInstaller` doesn't have an obvious relation to `ForkJoinPool`. `ForkJoinPool` is loaded early in `AgentInstaller.installBytebuddyAgent` from call to `List<AgentListener> agentListeners = loadOrdered(AgentListener.class);`. Seems like moving this block ensures that first metrics are sent after byte-buddy transformer is installed and byte-buddy has redefined `ForkJoinPool`.
The most unintrusive fix for this crash that I could come up with is to modify `OtlpInMemoryMetricExporter` so that it would fail export attempts before `AgentListener.afterAgent` is called.
To avoid similar issues in the future I'd urge maintainers to consider altering agent initialization sequence so that byte-buddy transformer is installed before sdk is initialized.

mateuszrzeszutek:
> To avoid similar issues in the future I'd urge maintainers to consider altering agent initialization sequence so that byte-buddy transformer is installed before sdk is initialized.

I don't think we can do that either - that could cause some instrumentations to initialize with the no-op `OpenTelemetry` implementation (since the global is set after SDK is initialized)

anuraaga:
Thanks @laurit for the great digging! Delayed the metric export to after agent initializes

testing/agent-exporter/src/main/java/io/opentelemetry/javaagent/testing/exporter/AgentTestingCustomizer.java
```diff
@@ -30,9 +43,73 @@ public void customize(AutoConfigurationCustomizer autoConfigurationCustomizer) {
...
+  }
+
+  @Override
+  public void afterAgent(
+      Config config, AutoConfiguredOpenTelemetrySdk autoConfiguredOpenTelemetrySdk) {
+    StartableMetricReader.INSTANCE.start();
+  }
+
+  @SuppressWarnings("ImmutableEnumChecker")
+  private enum StartableMetricReader implements MetricReaderFactory, MetricReader {
```
trask, line 56: can you add a brief comment explaining why this is needed (or just pointing to the PR discussion)?

trask (approved):

anuraaga:
For reference, I was expecting test slowdown with this PR going from in-memory transport to using localhost network traffic. After getting some builds through, dunno if I like it. Instead sent #5332

laurit:
> > To avoid similar issues in the future I'd urge maintainers to consider altering agent initialization sequence so that byte-buddy transformer is installed before sdk is initialized.
> 
> I don't think we can do that either - that could cause some instrumentations to initialize with the no-op `OpenTelemetry` implementation (since the global is set after SDK is initialized)

I didn't say it would be easy :) There wouldn't be too many instrumentations that are affected, these could use an extra agent started check and bail out when sdk isn't ready yet. Perhaps this agent started check could even be baked into instrumenter api somehow? For another case where running initial retransformation concurrently with background thread started sdk code produces strangeness see https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4697 Basically it boils down to replacing one set of problems with another, hopefully more manageable, set of problems.

PR #5315, opened 2022-02-07
Migrate AWS Lambda tests to Java

anuraaga:
I've realized I need to try some surgery on the lambda instrumentation, first migrating tests to Java for easier refactoring.


anuraaga:
Currently this fails because the library instrumentation for `AwsLambdaSqsEventHandlerTest` produces links with a remote parent as expected, but the javaagent version produces non-remote links. Anyone have an idea what could conceivably cause such a difference?

trask (approved):

PR #5324, opened 2022-02-09
fix: add attributes of net.peer.* for grpc client span

### LINKED ISSUE #5178: Grpc client spans doesn't capture the attributes of net.peer.*
**Describe the bug**
`net.peer.*` attributes should be captured in all spans, but in grpc client spans there are seems missing.

**Steps to reproduce**
I use the [grpc-java examples](https://github.com/grpc/grpc-java/tree/master/examples/src/main/java/io/grpc/examples/helloworld) as the demo.

HelloWorldClient: 
```
-javaagent:v1.10.0/opentelemetry-javaagent.jar
-Dotel.traces.exporter=logging
-Dotel.instrumentation.grpc.experimental-span-attributes=true
```
HelloWorldServer: 
```
-javaagent:v1.10.0/opentelemetry-javaagent.jar
-Dotel.traces.exporter=logging
-Dotel.instrumentation.grpc.experimental-span-attributes=true
```

grpc client spans contain attributes of `net.peer.*`

**What did you see instead?**
```
INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'helloworld.Greeter/SayHello' : f88623cfadc94f0ebf83a36876ce098b 00e46f859704f762 CLIENT [tracer: io.opentelemetry.grpc-1.6:1.10.0] AttributesMap{data={rpc.method=SayHello, thread.id=1, rpc.service=helloworld.Greeter, rpc.grpc.status_code=0, net.transport=ip_tcp, rpc.system=grpc, thread.name=main}, capacity=128, totalAddedValues=7}
```

v1.10.0

**Additional context**
other discussion reference: #1338 https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5151

ralphgj:
Try to resolve #5178

### DIFF
--- a/instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
+++ b/instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
@@ -11,7 +11,6 @@ import io.grpc.ClientCall;
 import io.grpc.ClientInterceptor;
 import io.grpc.ForwardingClientCall;
 import io.grpc.ForwardingClientCallListener;
-import io.grpc.Grpc;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
 import io.grpc.Status;
@@ -21,7 +20,9 @@ import io.opentelemetry.context.Context;
 import io.opentelemetry.context.Scope;
 import io.opentelemetry.context.propagation.ContextPropagators;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import java.net.InetSocketAddress;
 import java.net.SocketAddress;
+import java.net.URI;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
 final class TracingClientInterceptor implements ClientInterceptor {
@@ -55,8 +56,13 @@ final class TracingClientInterceptor implements ClientInterceptor {
         throw e;
       }
     }
-
-    SocketAddress address = result.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR);
+    SocketAddress address = null;
+    try {
+      URI uri = new URI(null, next.authority(), null, null, null);
+      address = InetSocketAddress.createUnresolved(uri.getHost(), uri.getPort());
+    } catch (Throwable e) {
+      // do nothing
+    }
     request.setRemoteAddress(address);
 
     return new TracingClientCall<>(result, parentContext, context, request);
--- a/instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcStreamingTest.groovy
+++ b/instrumentation/grpc-1.6/testing/src/main/groovy/io/opentelemetry/instrumentation/grpc/v1_6/AbstractGrpcStreamingTest.groovy
@@ -116,6 +116,8 @@ abstract class AbstractGrpcStreamingTest extends InstrumentationSpecification {
             "$SemanticAttributes.RPC_METHOD" "Conversation"
             "$SemanticAttributes.NET_TRANSPORT" SemanticAttributes.NetTransportValues.IP_TCP
             "$SemanticAttributes.RPC_GRPC_STATUS_CODE" Status.OK.code.value()
+            "$SemanticAttributes.NET_PEER_NAME" "localhost"
+            "$SemanticAttributes.NET_PEER_PORT" port
           }
           (1..(clientMessageCount * serverMessageCount + clientMessageCount)).each {
             def messageId = it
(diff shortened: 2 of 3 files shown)


trask (approved):
nice! I didn't notice that we had a repro for this issue in the tests themselves 👍

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
```diff
@@ -56,7 +58,8 @@
       }
     }
 
-    SocketAddress address = result.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR);
+    URI uri = GrpcUtil.authorityToUri(next.authority());
```
anuraaga, line 61: Let's just call `URI` constructor itself (and not throw on an invalid one)

https://github.com/grpc/grpc-java/blob/5b6ee1c1766165c85bb58e930a1b8dbbb705b629/core/src/main/java/io/grpc/internal/GrpcUtil.java#L495

anuraaga (approved):
Thanks!

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
```diff
@@ -55,8 +56,13 @@
         throw e;
       }
     }
-
-    SocketAddress address = result.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR);
+    SocketAddress address = null;
+    try {
+      URI uri = new URI(null, next.authority(), null, null, null);
+      address = new InetSocketAddress(uri.getHost(), uri.getPort() == -1 ? 80 : uri.getPort());
```
mateuszrzeszutek, line 62: Won't calling the `InetSocketAddress` constructor cause DNS resolution here? Should we use `InetSocketAddress.createUnresolved()` instead?
On the other hand - it's a client instrumentation, so in theory the address should already be resolved - is that the case here? Or is the intercept method called earlier than that?

ralphgj: > On the other hand - it's a client instrumentation, so in theory the address should already be resolved - is that the case here? Or is the intercept method called earlier than that?

Yes, I found that invoked `result.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR)` will return `null`. But in `TracingClientCallListener#onMessage` method, invoked it will get the expected result. So the root cause should be `result.getAttributes()` called too early.

trask: @ralphgj I think the concern here is that calling [new InetSocketAddress](https://docs.oracle.com/javase/8/docs/api/java/net/InetSocketAddress.html#InetSocketAddress-java.lang.String-int-) will trigger the host name to be resolved to an IP address, which we try to avoid triggering from instrumentation

`InetSocketAddress.createUnresolved()` is probably safer unless we are clear this won't be a problem

ralphgj: Got it. I tried and then found the attributes in spans without `net.peer.ip`. Shall we go on using `InetSocketAddress.createUnresolved()` instead?

trask: ya, it's ok not to capture `net.peer.ip`

instrumentation/grpc-1.6/library/src/main/java/io/opentelemetry/instrumentation/grpc/v1_6/TracingClientInterceptor.java
```diff
@@ -55,8 +56,15 @@
...
-    SocketAddress address = result.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR);
+    SocketAddress address = null;
+    try {
+      URI uri = new URI(null, next.authority(), null, null, null);
+      address =
+          InetSocketAddress.createUnresolved(
+              uri.getHost(), uri.getPort() == -1 ? 80 : uri.getPort());
+    } catch (Throwable e) {
+      // do nothing
+    }
```
trask, line 67: I'm not sure about defaulting to port 80, it seems like missing port could also mean 443, so better to leave it -1, which will lead to no `net.peer.port` captured (which is better than an incorrect port)

```suggestion
    try {
      URI uri = new URI(null, next.authority(), null, null, null);
      request.setRemoteAddress(InetSocketAddress.createUnresolved(uri.getHost(), uri.getPort()));
    } catch (URISyntaxException ignored) {
    }
```

ralphgj: `InetSocketAddress.createUnresolved` will check the port and throw an exception when port is -1. So maybe giving the default port with 80 or 443 is better.

PR #5326, opened 2022-02-09
Split lambda instrumentation into core and events

### LINKED ISSUE #5285: Lambda instrumentation requires user to add jackson dependency

### LINKED ISSUE #5325: AWS Lambda instrumentation requires lambda-events on the classpath
lambda-events is an optional library for users that want access to structured AWS events, but instrumentation should work without it. Currently we directly reference it in various places.

To address this, going to try splitting out core and events instrumentation for Lambda, an alternative approach gets to be a bit hairy

https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5309

anuraaga:
Fixes #5325 
Fixes #5285 

Sorry for huge PR couldn't see a way to split up the PR since it's actually splitting up the code.

I had one big misunderstanding about the Java Lambda runtime, that it would include libraries like Jackson for our use. But it turns out unlike some other languages like Python, Java doesn't provide any libraries, presumably since Java provides good classloader isolation techniques.

This PR makes several changes

- Splits instrumentation into two parts.
  - Target aws-lambda-java-core, which is mostly for supporting `RequestStreamHandler`. Notably, most lambda functions which use an actual web framework built on top of Lambda (using aws-serverless-java-container or AWS's internal framework) only include this dependency and do not include -events nor should it need to. Core is mostly about instrumenting and does not contain features like serialization.
  - Target aws-lambda-java-events, which is mostly for the lambda wrapper of `RequestHandler`, including support for events. Because Lambda does not know about the user's function type when invoking our wrapper, we must reproduce Lambda's serialization logic when invoking the actual user function.
    - Instrumenter logic only applying to events is only in this module

- Include Jackson as a dependency
  - lambda-core depends on Jackson core for extracting headers for HTTP propagation. Ideally instrumentation doesn't need external dependencies but this feature is too important to lose in core I think
  - lambda-events depends on Jackson databind. In the future, we will replace that with depending on [aws-lambda-java-serialization](https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-serialization). I've been concerned about the compatibility of our object mapping and finally found that we can reuse Lambda's serialization logic through that logic. I suspect that all of [these types](https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/events/mixins) are currently incompatible (except SQS which we have a special wrapper for).


instrumentation/aws-lambda/aws-lambda-core-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdacore/v1_0/AwsLambdaInstrumentationHelper.java
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.javaagent.instrumentation.awslambdacore.v1_0;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.instrumentation.awslambdacore.v1_0.internal.AwsLambdaFunctionInstrumenter;
+import io.opentelemetry.instrumentation.awslambdacore.v1_0.internal.AwsLambdaFunctionInstrumenterFactory;
+
+public final class AwsLambdaInstrumentationHelper {
```
anuraaga, line 12: Unchanged except for removing reference to SQS instrumenter

instrumentation/aws-lambda/aws-lambda-core-1.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdacore/v1_0/AwsLambdaRequestHandlerInstrumentation.java
```diff
@@ -0,0 +1,88 @@
...
+  @SuppressWarnings("unused")
+  public static class HandleRequestAdvice {
+
+    @Advice.OnMethodEnter(suppress = Throwable.class)
+    public static void onEnter(
+        @Advice.Argument(value = 0, typing = Typing.DYNAMIC) Object arg,
+        @Advice.Argument(1) Context context,
+        @Advice.Local("otelInput") AwsLambdaRequest input,
+        @Advice.Local("otelContext") io.opentelemetry.context.Context otelContext,
+        @Advice.Local("otelScope") Scope otelScope) {
```
anuraaga, line 60: Unchanged except removing references to SQS instrumenter

instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/LambdaParameters.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package io.opentelemetry.instrumentation.awslambdacore.v1_0;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import java.lang.reflect.Method;
+
+final class LambdaParameters {
+
+  static <T> Object[] toArray(Method targetMethod, T input, Context context) {
```
anuraaga, line 13: Removed mapper for core version of this util

instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/TracingRequestHandler.java
```diff
@@ -98,4 +88,12 @@ public final O handleRequest(I input, Context context) {
   }
 
   protected abstract O doHandleRequest(I input, Context context);
+
+  /**
+   * Returns the HTTP headers for the request extracted from the request handler's input. By
+   * default, returns empty headers.
+   */
+  protected Map<String, String> extractHttpHeaders(I input) {
```
anuraaga, line 96: This is the most invasive change, since a handler that uses ApiGatewayProxyEvent would need to have its HTTP headers extracted. That being said, headers could be represented by input in other unknown ways so giving a hook into that could be considered generally useful (with limited practical use case).

instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/internal/ApiGatewayProxyRequest.java
```diff
@@ -38,17 +40,22 @@ private static boolean xrayPropagationFieldsOnly(Collection<String> fields) {
...
     if (noHttpPropagationNeeded()) {
       return new NoopRequest(source);
     }
 
     if (source.markSupported()) {
       return new MarkableApiGatewayProxyRequest(source);
     }
-    // fallback
-    return new CopiedApiGatewayProxyRequest(source);
+    // It is known that the Lambda runtime passes ByteArrayInputStream's to functions, so gracefully
```
anuraaga, line 51: I believe this is reasonable. Notably, if we were actually hitting the fallback codepath, it would often be failing with ClassNotFoundException since commons-io is not referenced at all by lambda libraries and would only have to be available by luck

instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/internal/HeadersFactory.java
```diff
@@ -0,0 +1,60 @@
...
+        return Collections.emptyMap();
+      }
+      while (parser.nextToken() != JsonToken.END_OBJECT) {
+        parser.nextToken();
+        if (!parser.getCurrentName().equals("headers")) {
+          parser.skipChildren();
+          continue;
+        }
+
+        if (!parser.isExpectedStartObjectToken()) {
```
anuraaga, line 39: Migrated from OBJECT_MAPPER to all jackson-core

instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaInstrumentationHelper.java
```diff
@@ -0,0 +1,35 @@
...
+
+package io.opentelemetry.javaagent.instrumentation.awslambdaevents.v2_2;
+
+import com.amazonaws.services.lambda.runtime.events.SQSEvent;
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
+import io.opentelemetry.instrumentation.awslambdaevents.v2_2.internal.AwsLambdaEventsInstrumenterFactory;
+import io.opentelemetry.instrumentation.awslambdaevents.v2_2.internal.AwsLambdaSqsInstrumenterFactory;
+
+public final class AwsLambdaInstrumentationHelper {
```
anuraaga, line 14: Same as helper in current HEAD code

instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaInstrumentationModule.java
```diff
@@ -0,0 +1,37 @@
...
+import net.bytebuddy.matcher.ElementMatcher;
+
+@AutoService(InstrumentationModule.class)
+public class AwsLambdaInstrumentationModule extends InstrumentationModule {
+  public AwsLambdaInstrumentationModule() {
+    super("aws-lambda", "aws-lambda-events-2.2");
+  }
+
+  @Override
+  public ElementMatcher.Junction<ClassLoader> classLoaderMatcher() {
```
anuraaga, line 24: Added this matcher, verified by testInstrumentation in other project

instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/awslambdaevents/v2_2/AwsLambdaRequestHandlerInstrumentation.java
```diff
@@ -63,18 +61,24 @@ public static void onEnter(
         @Advice.Local("otelMessageContext") io.opentelemetry.context.Context messageContext,
         @Advice.Local("otelMessageScope") Scope messageScope) {
```
anuraaga, line 62: This class is same as current instrumentation in HEAD

It's probably not enough to figure out a way to share logic in the advice classes between the two javaagent instrumentations at this time

instrumentation/aws-lambda/aws-lambda-events-2.2/library/src/main/java/io/opentelemetry/instrumentation/awslambdaevents/v2_2/TracingRequestWrapperBase.java
```diff
@@ -40,7 +48,10 @@ protected TracingRequestWrapperBase(BiFunction<I, Class<?>, Object> parameterMap
       OpenTelemetrySdk openTelemetrySdk,
       WrappedLambda wrappedLambda,
       BiFunction<I, Class<?>, Object> parameterMapper) {
-    super(openTelemetrySdk, WrapperConfiguration.flushTimeout());
+    super(
+        openTelemetrySdk,
+        WrapperConfiguration.flushTimeout(),
+        AwsLambdaEventsInstrumenterFactory.createInstrumenter(openTelemetrySdk));
```
anuraaga, line 54: Key change - the instrumenter factory in this module is aware of event objects

instrumentation/aws-lambda/aws-lambda-events-2.2/library/src/main/java/io/opentelemetry/instrumentation/awslambdaevents/v2_2/internal/AwsLambdaEventsInstrumenterFactory.java
```diff
@@ -3,27 +3,29 @@
...
-package io.opentelemetry.instrumentation.awslambda.v1_0.internal;
+package io.opentelemetry.instrumentation.awslambdaevents.v2_2.internal;
 
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.instrumentation.api.instrumenter.Instrumenter;
 import io.opentelemetry.instrumentation.api.instrumenter.SpanKindExtractor;
-import io.opentelemetry.instrumentation.awslambda.v1_0.AwsLambdaRequest;
+import io.opentelemetry.instrumentation.awslambdacore.v1_0.AwsLambdaRequest;
+import io.opentelemetry.instrumentation.awslambdacore.v1_0.internal.AwsLambdaFunctionAttributesExtractor;
```
anuraaga, line 13: We use the core instrumentation logic, augmenting with event-specific logic

anuraaga (commented):
Tried to explain some more interesting parts or large diffs in moved files

instrumentation/aws-lambda/aws-lambda-core-1.0/library/src/main/java/io/opentelemetry/instrumentation/awslambdacore/v1_0/internal/AwsLambdaFunctionAttributesExtractor.java
```diff
@@ -12,13 +12,17 @@
...
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
 import javax.annotation.Nullable;
 
-class AwsLambdaFunctionAttributesExtractor
+/**
+ * This class is internal and is hence not for public use. Its APIs are unstable and can change at
+ * any time.
+ */
```
trask, line 24: 😄

instrumentation/aws-lambda/aws-lambda-events-2.2/library/build.gradle.kts
```diff
@@ -16,25 +18,19 @@ dependencies {
   // in public API.
   library("com.amazonaws:aws-lambda-java-events:2.2.1")
 
-  compileOnly("com.fasterxml.jackson.core:jackson-databind")
-  compileOnly("commons-io:commons-io:2.2")
-  compileOnly("org.slf4j:slf4j-api")
-
+  implementation("com.fasterxml.jackson.core:jackson-databind")
```
trask, line 21: and copy in comment here from your PR comment would be nice I think

instrumentation/aws-lambda/aws-lambda-events-2.2/javaagent/build.gradle.kts
```diff
@@ -13,7 +13,12 @@ muzzle {
 }
 
 dependencies {
-  implementation(project(":instrumentation:aws-lambda-1.0:library"))
+  implementation(project(":instrumentation:aws-lambda:aws-lambda-core-1.0:library"))
+
+  implementation(project(":instrumentation:aws-lambda:aws-lambda-events-2.2:library")) {
+    // Only needed by wrappers, not the javaagent. Muzzle will catch if we accidentally change this.
```
trask, line 19: 👍

trask (approved):

PR #5327, opened 2022-02-09
Collect more crash dump files

laurit:
Last time I tried reporting to openj9 the first thing they asked was for other crash dump files besides the `javacore` file.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -36,12 +36,17 @@ jobs:
           path: /tmp/deadlock-detector-*
           if-no-files-found: ignore
 
-      - name: Upload OpenJ9 javacore files if any
+      - name: Upload jvm crash dump files if any
         if: always()
         uses: actions/upload-artifact@v2
         with:
           name: javacore-build
-          path: "**/javacore.*.txt"
+          path: |
+            "**/hs_err_pid*.log"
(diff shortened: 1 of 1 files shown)


mateuszrzeszutek (approved):

.github/workflows/pr.yml
```diff
@@ -36,12 +36,17 @@ jobs:
...
         if: always()
         uses: actions/upload-artifact@v2
         with:
           name: javacore-build
-          path: "**/javacore.*.txt"
+          path: |
+            "**/hs_err_pid*.log"
+            "**/javacore.*.txt"
+            "**/Snap.*.trc"
+            "**/core.*.dmp"
```
laurit, line 48: Collecting core dump will probably fail as test output has
```
JVMPORT030W /proc/sys/kernel/core_pattern setting "|/usr/share/apport/apport %p %s %c %d %P %E" specifies that the core dump is to be piped to an external program.  Attempting to rename either core or core.3403.

JVMDUMP012E Error in System dump: The core file created by child process with pid = 3403 was not found. Expected to find core file with name "/home/runner/work/opentelemetry-java-instrumentation/opentelemetry-java-instrumentation/instrumentation/apache-dubbo-2.7/javaagent/core"
```

trask (approved):

PR #5331, opened 2022-02-10
Remove separate exporter configuration which we always have in distro…

### LINKED ISSUE #5254: Remove javaagent-exporters module?
I'm not sure if this module is needed anymore.

I think we could embed the list of exporters directly into javaagent/build.gradle.kts, e.g.

```
  exporterLibs("io.opentelemetry:opentelemetry-exporter-jaeger")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-otlp")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-otlp-metrics")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-otlp-logs")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-otlp-http-trace")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-otlp-http-metrics")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-otlp-http-logs")
  exporterLibs("io.opentelemetry:opentelemetry-exporter-logging-otlp")

  exporterLibs("io.opentelemetry:opentelemetry-exporter-prometheus")
  exporterLibs("io.prometheus:simpleclient")
  exporterLibs("io.prometheus:simpleclient_httpserver")

  exporterLibs("io.opentelemetry:opentelemetry-exporter-zipkin")
```

anuraaga:
… now.

Don't think I see a difference between using javaagentLibs vs the previous exporterLibs (I think that would have been true before too though so not sure)

Fixes #5254


javaagent/build.gradle.kts
```diff
@@ -69,7 +65,20 @@ dependencies {
   baseJavaagentLibs(project(":instrumentation:internal:internal-reflection:javaagent"))
   baseJavaagentLibs(project(":instrumentation:internal:internal-url-class-loader:javaagent"))
 
-  exporterLibs(project(":javaagent-exporters"))
```
trask, line 72: can we delete this module now?

anuraaga: Oops that was the whole point wasn't it. Also ended up defining in javaagent-tooling instead since we already had an exporter there

anuraaga:
Might need #5314

trask (approved):

PR #5333, opened 2022-02-10
Publish gradle-plugins to Maven Central

### LINKED ISSUE #5330: Publish gradle-plugins to Maven Central
Currently we only publish gradle-plugins to Gradle Plugin Portal I think, but if they were on Maven Central it would be much easier for me to use internally, and maybe for other organizations that can only mirror Maven Central, not Gradle Plugin Portal.

The config makes me think it's all ready to go and we just need to run the task

https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/gradle-plugins/build.gradle.kts#L71

anuraaga:
Also makes sure to target Java 8 bytecode so the plugin doesn't require other builds to be using Java 11

Fixes #5330

### DIFF
--- a/.github/workflows/release-gradle-plugins.yml
+++ b/.github/workflows/release-gradle-plugins.yml
@@ -156,7 +156,6 @@ jobs:
           distribution: adopt
           java-version: 11
 
-        # TODO (trask) cache gradle wrapper?
       - name: Build and publish gradle plugins
         uses: gradle/gradle-build-action@v2
         env:
--- a/.github/workflows/release-gradle-plugins.yml
+++ b/.github/workflows/release-gradle-plugins.yml
@@ -158,8 +158,15 @@ jobs:
 
         # TODO (trask) cache gradle wrapper?
       - name: Build and publish gradle plugins
+        uses: gradle/gradle-build-action@v2
         env:
+          SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
+          SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
           GRADLE_PUBLISH_KEY: ${{ secrets.GRADLE_PUBLISH_KEY }}
           GRADLE_PUBLISH_SECRET: ${{ secrets.GRADLE_PUBLISH_SECRET }}
-        run: ../gradlew build publishPlugins
-        working-directory: gradle-plugins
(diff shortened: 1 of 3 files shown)


.github/workflows/release-gradle-plugins.yml
```diff
@@ -158,8 +158,15 @@ jobs:
 
         # TODO (trask) cache gradle wrapper?
```
trask, line 159:
```suggestion
```

.github/workflows/release-gradle-plugins.yml
```diff
@@ -158,8 +158,15 @@ jobs:
...
+          SONATYPE_KEY: ${{ secrets.SONATYPE_KEY }}
           GRADLE_PUBLISH_KEY: ${{ secrets.GRADLE_PUBLISH_KEY }}
           GRADLE_PUBLISH_SECRET: ${{ secrets.GRADLE_PUBLISH_SECRET }}
-        run: ../gradlew build publishPlugins
-        working-directory: gradle-plugins
+          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+          GPG_PASSWORD: ${{ secrets.GPG_PASSWORD }}
+        with:
+          # Don't use publishToSonatype since we don't want to publish the marker artifact
+          arguments: build publishPlugins publishPluginMavenPublicationToSonatypeRepository closeAndReleaseSonatypeStagingRepository
```
trask, line 171: where does the `publishPluginMavenPublicationToSonatypeRepository` task come from?

gradle-plugins/build.gradle.kts
```diff
@@ -86,3 +103,49 @@ tasks {
     enabled = !version.toString().contains("SNAPSHOT")
   }
 }
+
+afterEvaluate {
+  publishing {
+    publications {
+      named<MavenPublication>("pluginMaven") {
```
trask, line 110: does `named` imply that this task already exists?

.github/workflows/release-gradle-plugins.yml
anuraaga: Nexus publish plugin

gradle-plugins/build.gradle.kts
anuraaga: Yeah Gradle plugin plugin creates it

PR #5347, opened 2022-02-11
Add builders for setting optional attributes on HTTP extractors

mateuszrzeszutek:


instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientAttributesExtractorTest.java
```diff
@@ -104,10 +104,12 @@ void normal() {
...
-        HttpClientAttributesExtractor.create(
-            new TestHttpClientAttributesGetter(),
-            CapturedHttpHeaders.create(
-                singletonList("Custom-Request-Header"), singletonList("Custom-Response-Header")));
+        HttpClientAttributesExtractor.builder(new TestHttpClientAttributesGetter())
+            .captureHttpHeaders(
+                CapturedHttpHeaders.create(
+                    singletonList("Custom-Request-Header"),
+                    singletonList("Custom-Response-Header")))
+            .build();
```
trask, line 112: what do you think of hiding `CapturedHttpHeaders` and exposing `captureHttpRequestHeaders` and `captureHttpResponseHeaders` on the builders?

trask (approved):

instrumentation-api/src/test/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientAttributesExtractorTest.java
mateuszrzeszutek: I think I'd probably prefer leaving this as it is, it's a bit simpler this way.

If I added `captureHttpRequestHeaders(List<String>)` I'd have to take the following things into account:
* is it additive? does `. captureHttpRequestHeaders(["abc"]).captureHttpRequestHeaders(["def"])` capture `["abc", "def"]` or does the second call overwrite the first one?
* do the default (configured) values get overwritten, or appended to?

With `captureHttpHeaders(CapturedHttpHeaders)` it's clear (I hope so? 😅 ) that the previous setting is overwritten. (Maybe I should rename it to `setCapturedHttpHeaders` after all...)

trask: cc @anuraaga for any API thoughts

anuraaga: How about

`setCapturedRequestHeaders`
`setCapturedResponseHeaders`

mateuszrzeszutek: I'll do that in a separate PR - this requires changing all our HTTP library instrumentations, so the amount of changes will be pretty significant.

Issue #3480, opened 2021-07-02
Spring Boot Starter service-name is constant 


gquintana:
**Is your feature request related to a problem? Please describe.**
I added the `opentelemetry-spring-starter` and `opentelemetry-exporter-jaeger` to 2 Spring Boot applications. The OpenTelemetry instrumentation always set `serviceName` as `unknown_service:java`. I am not able to tell which span belongs to which application.

These values are the default ones https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L50

**Describe the solution you'd like**
Use Spring Boot `application.name` setting as service name

**Describe alternatives you've considered**
Be able to override from from `application.properties` file with something like `otel.resource.attributes=service.name=myapp-backend`. This would be the equivalent of `-Dotel.resource.attributes=service.name=myapp-backend` when running as a Java agent. Or even better  `otel.resource.attributes.service.name=myapp-backend`

AlexWang-CN:
I have the same problem

jkwatson:
it might be nice to support a pattern-based resource configuration in the application.properties like this:

```
otel.resource.attributes.environment=dev
otel.resource.attributes.xyz=foo
```

aschugunov:
Are there any updates, plans?

oceanstars:
```yaml
otel:
  springboot:
    resource:
      attributes:
        service.name: my_service_name
```

This configuration in "application. yaml" is valid for setting service name.

PR #5359, opened 2022-02-11
Spring Boot Starter service-name is constant

aschugunov:
Pattern-based resource attributes configuration
fix [3480](https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/3480)


aschugunov:
@mateuszrzeszutek could you provide some feedback, please?

instrumentation/spring/spring-boot-autoconfigure/README.md
```diff
@@ -1,18 +1,24 @@
 # OpenTelemetry Spring Auto-Configuration
 
-Auto-configures OpenTelemetry instrumentation for [spring-web](../spring-web-3.1), [spring-webmvc](../spring-webmvc-3.1), and [spring-webflux](../spring-webflux-5.0). Leverages Spring Aspect Oriented Programming, dependency injection, and bean post-processing to trace spring applications. To include all features listed below use the [opentelemetry-spring-starter](../starters/spring-starter/README.md).
+Auto-configures OpenTelemetry instrumentation for [spring-web](../spring-web-3.1)
+, [spring-webmvc](../spring-webmvc-3.1), and [spring-webflux](../spring-webflux-5.0). Leverages
+Spring Aspect Oriented Programming, dependency injection, and bean post-processing to trace spring
+applications. To include all features listed below use
+the [opentelemetry-spring-starter](../starters/spring-starter/README.md).
```
trask, line 7: can you undo the formatting changes in this file? totally fine with formatting changes, but much easier to review those in a separate PR

instrumentation/spring/spring-boot-autoconfigure/README.md
```diff
@@ -392,6 +440,13 @@ If an exporter is present in the classpath during runtime and a spring bean of t
 
 <!-- Slf4j Log Correlation  otel.springboot.loggers.slf4j.enabled		true   		org.slf4j.MDC -->
 
+##### Resource Properties
+
+| Feature  | Property                                         | Default Value        | ConditionalOnClass |
+|----------|--------------------------------------------------|----------------------|--------------------|
+| Resource | otel.springboot.resource.enabled                 | false                | OsResource         |
```
trask, line 447: I think it may be ok to make this default to `true`, similar to the otel autoconfigure module

instrumentation/spring/spring-boot-autoconfigure/README.md
```diff
@@ -392,6 +440,13 @@ If an exporter is present in the classpath during runtime and a spring bean of t
...
+|          | otel.springboot.resource.attributes.service.name | unknown_service:java |                    |
```
trask, line 448: i'm unsure of the default value, it's sort of `spring.application.name`?

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+      }
+      Attributes attributes = attributesBuilder.build();
+      return Resource.create(attributes);
+    };
+  }
+
+  @Bean
+  public Supplier<Resource> otelOsResourceProvider() {
+    return OsResource::get;
+  }
```
trask, line 54: what do you think of loading the resources via SPI? https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/resources/src/main/resources/META-INF/services/io.opentelemetry.sdk.autoconfigure.spi.ResourceProvider

or are there advantages to having them as separate named beans?

trask (commented):
thx @aschugunov! sorry for the delay in reviewing

instrumentation/spring/spring-boot-autoconfigure/README.md
aschugunov: no, it is a default name of resource taken from https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L48
So Resource is created with this name when `spring.application.name` and `otel.springboot.resource.attributes.service.name` are not set

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
aschugunov: To be honest, I dont deeply understand how SPI works.
Currently, trace does not contain any inforamtion about os, container and other resources.
Main idea is combine all resources via supplier resource beans and then inject complex resource into tracerProvider.
maybe it is possbile to do with SPI, I will try to research. Thanks.

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfigurationTest.java
```diff
@@ -0,0 +1,89 @@
...
+  private final ApplicationContextRunner contextRunner =
+      new ApplicationContextRunner()
+          .withConfiguration(
+              AutoConfigurations.of(
+                  OtelResourceAutoConfiguration.class, OpenTelemetryAutoConfiguration.class));
+
+  @AfterEach
+  void tearDown() {
+    GlobalOpenTelemetry.resetForTest();
+  }
```
trask, line 30: I don't think `resetForTest` is needed (anymore?) in spring boot autoconfigure tests, see #5566

instrumentation/spring/spring-boot-autoconfigure/README.md
```diff
@@ -392,6 +392,13 @@ If an exporter is present in the classpath during runtime and a spring bean of t
 
 <!-- Slf4j Log Correlation  otel.springboot.loggers.slf4j.enabled		true   		org.slf4j.MDC -->
 
+##### Resource Properties
+
+| Feature  | Property                                         | Default Value          |
+|----------|--------------------------------------------------|------------------------|
+| Resource | otel.springboot.resource.enabled                 | `true`                 |
+|          | otel.springboot.resource.attributes.service.name | `unknown_service:java` |
+
```
trask, line 401: can you add doc about `spring.application.name` and about adding custom attributes?

trask (approved):

instrumentation/spring/spring-boot-autoconfigure/src/test/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfigurationTest.java
aschugunov: Yes, it looks like that. I deleted `resetForTest` in new test files.

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
aschugunov: I tried to load resource objects via spring.factories and it seems spring does not load bean by interface or I did it in wrong manner. 
If I understand correctly java loads objects via SPI when implementation is located in classpath, spring can customize context via spring.factories file. But this way does not work
```
io.opentelemetry.sdk.autoconfigure.spi.ResourceProvider=\
io.opentelemetry.sdk.extension.resources.ContainerResourceProvider,\
io.opentelemetry.sdk.extension.resources.HostResourceProvider,\
io.opentelemetry.sdk.extension.resources.OsResourceProvider,\
io.opentelemetry.sdk.extension.resources.ProcessResourceProvider,\
io.opentelemetry.sdk.extension.resources.ProcessRuntimeResourceProvider
```

I added `@ConditionalOnClass` over appropriate bean, maybe it will be more suitable

instrumentation/spring/spring-boot-autoconfigure/README.md
```diff
@@ -392,6 +392,26 @@ If an exporter is present in the classpath during runtime and a spring bean of t
...
+`unknown_service:java` will be used as the service-name if no value has been specified to the
+property `spring.application.name` or `otel.springboot.resource.attributes.service.name` (which has
+the highest priority)
+
+`otel.springboot.resource.attributes` supports a pattern-based resource configuration in the
+application.properties like this:
+
+```
+otel.resource.attributes.environment=dev
+otel.resource.attributes.xyz=foo
```
trask, line 412:
```suggestion
otel.springboot.resource.attributes.environment=dev
otel.springboot.resource.attributes.xyz=foo
```

trask (approved):

instrumentation/spring/spring-boot-autoconfigure/README.md
aschugunov: exactly, thanks a lot

trask:
@open-telemetry/java-instrumentation-approvers last chance to review!

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceProperties.java
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.spring.autoconfigure;
+
+import java.util.Collections;
+import java.util.Map;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "otel.springboot.resource")
```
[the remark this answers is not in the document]
trask: since it's not spec'd (or at least supported by autoconfiguration) I think the springboot namespace makes some sense for now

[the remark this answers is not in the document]
aschugunov: I followed idea from this comment https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2770#issue-855124477
so there is some confusing between agent settings and springboot configuration

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+      Environment env, OtelResourceProperties otelResourceProperties) {
+    return () -> {
+      AttributesBuilder attributesBuilder = Attributes.builder();
+      for (Map.Entry<String, String> entry : otelResourceProperties.getAttributes().entrySet()) {
+        attributesBuilder.put(entry.getKey(), entry.getValue());
+      }
+      String otelServiceName = env.getProperty("otel.springboot.resource.attributes.service.name");
+      if (otelServiceName != null) {
+        attributesBuilder.put(ResourceAttributes.SERVICE_NAME, otelServiceName);
+      }
```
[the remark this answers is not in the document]
aschugunov: Unfortunately, `otel.springboot.resource.attributes.service.name`  does not set name into attributes
`dot` between service.name does not allow parse name properly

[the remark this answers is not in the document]
trask: that's too bad, this will be a problem for lots of other resource attributes https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#service

do you have a specific use-case or need for including generic resource attributes?

if not, maybe we should skip that until we have a solution that will generalize for all resource attributes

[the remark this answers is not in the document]
aschugunov: I verified again and I was wrong
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-Configuration-Binding#map-based-binding 
it is possible to define attributes with dotes, thank you @mateuszrzeszutek 
I added test for such cases (it seems I wrote wrong test before and it was a cause of wrong understanding)

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
aschugunov: Could you explain a bit more, please? 
Did you mean to create `resourceProvider` beans in addition to `resource` beans?

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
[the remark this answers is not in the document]
mateuszrzeszutek: Awesome, thanks!

instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OtelResourceAutoConfiguration.java
mateuszrzeszutek: So, instead of all the `Supplier<Resource>` beans in this class, you could use `ResourceProvider` instead:
```java
  @Bean
  @ConditionalOnClass(OsResource.class)
  public ResourceProvider osResourceProvider() {
    return new OsResourceProvider();
  }

  @Bean
  @ConditionalOnClass(ProcessResource.class)
  public ResourceProvider processResourceProvider() {
    return new ProcessResourceProvider();
  }

 // ...
```
Then, in the `OpenTelemetryAutoConfiguration` class you'd change the `otelResource` method so that it uses these beans:
```java
    @Bean
    @ConditionalOnMissingBean
    public Resource otelResource(
        Environment env, ObjectProvider<List<ResourceProvider>> resourceProviders) {
      String applicationName = env.getProperty("spring.application.name");
      Resource resource = defaultResource(applicationName);
      ConfigProperties config = SpringConfigProperties(env);
      for (ResourceProvider provider : resourceProviders.getIfAvailable(Collections::emptyList)) {
        resource = resource.merge(provider.createResource(config));
      }
      return resource;
    }
```
`ResourceProvider#createResource` requires a `ConfigProperties` argument, but it would be quite simple to implement one that just simply delegates to Spring's `Environment` class.

This way you could reuse the SDK `ResourceProvider` interface in a way that's compatible with Spring beans -- instead of using SPI, the user of the instrumented application could just configure a `ResourceProvider` bean in their app and it'd get picked up automatically by this lib.

WDYT about that?

aschugunov: Yes it is a great idea to reuse existed interface for getting all available resources.
I see benefit of interface, it can be more clear for whole open telemetry project because interface provides consistent concept.
On the other hand, supplier follows the same idea - to provide resource. I mean there is no difference how spring will provide beans - via `ResourceProvider` interface or via `Supplier<Resource>`. Also interface enforce to use `ConfigProperties` that is not used in any current resourceProviders implementations. 
`OtelResourceProperties` is such type of config. It is created by spring automatically.
If I understand correctly I should implement something like SpringResourceProvider and encapsulate there logic of determination of application name and attributes. 

I will add these changes in second commit in this PR in a day or two. So you will be able to evaluate changes and choose more appropriate implementation.

mateuszrzeszutek: > I will add these changes in second commit in this PR in a day or two. So you will be able to evaluate changes and choose more appropriate implementation.

👍  Thanks!

mateuszrzeszutek:
Thanks @aschugunov !

PR #5364, opened 2022-02-13
Improve the testLatestDeps PR job

trask:
... so that it won't show up as `canceled` and prevent the build section from collapsing when everything (else) is green

@anuraaga I tried splitting it out as a separate workflow (see initial commit on this PR), but didn't like it as much (it would only run once on the labeling action)


anuraaga:
> (it would only run once on the labeling action)

:1:

anuraaga (approved):

PR #5369, opened 2022-02-14
Updates to http.server_name

trask:
[from the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions), `http.server_name` is:

> The primary server name of the matched virtual host. This should be obtained via configuration. If no such configuration can be obtained, this attribute MUST NOT be set ( `net.host.name` should be used instead).
>
> ...
>
> For example, http.server_name has shown great value in practice, as bogus HTTP Host headers occur often in the wild.
>
> It is strongly recommended to set http.server_name to allow associating requests with some logical server entity.

The servlet and tomcat instrumentation currently capture `http.server_name` based on HTTP Host (e.g. https://docs.oracle.com/javaee/7/api/javax/servlet/ServletRequest.html#getServerName--), so this PR removes the capturing of `http.server_name` for those two instrumentations.

The only other instrumentation that captures `http.server_name` is armeria, which does appear to be capturing a configuration-based value.

I'm not sure if the one instrumentation justifies keeping `serverName()` in `HttpServerAttributesGetter`?

For this PR though, at least it seems the `RESPONSE` parameter is not needed for `serverName()` (and based on the spec definition it doesn't seem like this should be something that we need to wait for the `RESPONSE` for), so I have removed the `RESPONSE` parameter from the `serverName()` method.

### DIFF
--- a/instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpServerAttributesGetter.java
+++ b/instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpServerAttributesGetter.java
@@ -96,7 +96,7 @@ final class ArmeriaHttpServerAttributesGetter
   @Nullable
   public String serverName(RequestContext ctx) {
     if (ctx instanceof ServiceRequestContext) {
-      return ((ServiceRequestContext) ctx).config().virtualHost().hostnamePattern();
+      return ((ServiceRequestContext) ctx).config().virtualHost().defaultHostname();
     }
     return null;
   }
--- a/instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpServerAttributesGetter.java
+++ b/instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpServerAttributesGetter.java
@@ -94,7 +94,7 @@ final class ArmeriaHttpServerAttributesGetter
 
   @Override
   @Nullable
-  public String serverName(RequestContext ctx, @Nullable RequestLog requestLog) {
+  public String serverName(RequestContext ctx) {
     if (ctx instanceof ServiceRequestContext) {
       return ((ServiceRequestContext) ctx).config().virtualHost().hostnamePattern();
     }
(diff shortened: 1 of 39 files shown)


instrumentation/armeria-1.3/library/src/main/java/io/opentelemetry/instrumentation/armeria/v1_3/ArmeriaHttpServerAttributesGetter.java
```diff
@@ -94,9 +94,9 @@ public Long responseContentLengthUncompressed(RequestContext ctx, RequestLog req
 
   @Override
   @Nullable
-  public String serverName(RequestContext ctx, @Nullable RequestLog requestLog) {
+  public String serverName(RequestContext ctx) {
     if (ctx instanceof ServiceRequestContext) {
-      return ((ServiceRequestContext) ctx).config().virtualHost().hostnamePattern();
+      return ((ServiceRequestContext) ctx).config().virtualHost().defaultHostname();
```
trask, line 99: @anuraaga can you ack this change?

instrumentation/servlet/servlet-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/servlet/ServletHttpAttributesGetter.java
```diff
@@ -139,9 +139,7 @@ public String route(ServletRequestContext<REQUEST> requestContext) {
 
   @Override
   @Nullable
-  public String serverName(
-      ServletRequestContext<REQUEST> requestContext,
-      @Nullable ServletResponseContext<RESPONSE> responseContext) {
-    return accessor.getRequestServerName(requestContext.request());
+  public String serverName(ServletRequestContext<REQUEST> requestContext) {
+    return null;
```
trask, line 143: ^^ the servlet change

instrumentation/tomcat/tomcat-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/tomcat/common/TomcatHttpAttributesGetter.java
```diff
@@ -101,7 +101,7 @@ public String route(Request request) {
 
   @Override
   @Nullable
-  public String serverName(Request request, @Nullable Response response) {
-    return request.serverName().toString();
+  public String serverName(Request request) {
+    return null;
```
trask, line 105: ^^ the tomcat change

anuraaga (approved):

laurit:
> [from the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions), `http.server_name` is:
> 
> > The primary server name of the matched virtual host. This should be obtained via configuration. If no such configuration can be obtained, this attribute MUST NOT be set ( `net.host.name` should be used instead).
> > ...

currently we don't set `net.host.name` at all, should we?

trask:
> currently we don't set `net.host.name` at all, should we?

ya, probably, opened #5375 to discuss/track

PR #5371, opened 2022-02-14
Fix flaky jdk http client test

laurit:
https://ge.opentelemetry.io/scans/tests?search.relativeStartTime=P7D&search.tags=CI&search.timeZoneId=Europe/Tallinn&tests.container=JdkHttpClientTest&tests.sortField=FLAKY&tests.test=trace%20request%20with%20callback%20and%20no%20parent&tests.unstableOnly=true
Apparently sometimes http client call context is propagated to callback so we need to always set the context for the callback.


laurit:
> > Apparently sometimes http client call context is propagated to callback so we need to always set the context for the callback.
> 
> Is this a context leak? We should probably have an issue or at least TODO to track fixing root cause if there is one

I wouldn't call this a context leak, it is just executor instrumentation working in mysterious ways. My understanding is that as our instrumentation wraps the whole `sendAsync` context really should be propagating to `CompletableFuture` callbacks. Usually this undesired propagation does not happen, but as proved by the flaky test there exists a path where the propagation works. The fact that this worked before this pr could also be said to be an issue with the executor instrumentation.

PR #5372, opened 2022-02-14
Exclude broken oshi release

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5370

### DIFF
--- a/instrumentation/oshi/javaagent/build.gradle.kts
+++ b/instrumentation/oshi/javaagent/build.gradle.kts
@@ -7,6 +7,8 @@ muzzle {
     group.set("com.github.oshi")
     module.set("oshi-core")
     versions.set("[5.3.1,)")
+    // Could not parse POM https://repo.maven.apache.org/maven2/com/github/oshi/oshi-core/6.1.1/oshi-core-6.1.1.pom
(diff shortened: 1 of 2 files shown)


instrumentation/oshi/javaagent/build.gradle.kts
```diff
@@ -15,9 +17,13 @@ dependencies {
 
   compileOnly("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure")
 
-  library("com.github.oshi:oshi-core:5.3.1")
+  compileOnly("com.github.oshi:oshi-core:5.3.1")
+  testImplementation("com.github.oshi:oshi-core:5.3.1")
```
trask, line 21: why this change?

laurit: when using `library` it still attempted to parse the broken pom. Here is the failure with `library` https://github.com/open-telemetry/opentelemetry-java-instrumentation/runs/5185602614?check_suite_focus=true Anyway I expect that `library` and `latestDepTestLibrary` changes in this pr can be reverted tomorrow.

trask (approved):

PR #5376, opened 2022-02-14
Alter flaky redisson test

laurit:
https://ge.opentelemetry.io/scans/tests?search.relativeStartTime=P7D&search.tags=CI&search.timeZoneId=Europe/Tallinn&tests.container=RedissonAsyncClientTest&tests.sortField=FLAKY&tests.test=test%20future%20whenComplete&tests.unstableOnly=true
is currently the most flaky test that we have. It has been flaky since https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5073 but I believe that pr just brought out the issue. The problem is that the `strings.add("a")` call sometimes hangs. When we ended redisson span right after sending the command without waiting for the response it wasn't a problem because this hang didn't change the expected outcomes. Now that we end redisson span when operation completes it is an issue as test will fail because operation failed (timed out) or after increasing timeout test will fail because of missing span. I tried pretty hard, but wasn't able to replicate this locally. I believe the root cause of the issue is that this test misuses redisson api. `whenComplete` runs on a netty thread and should not invoke blocking operations. In later version redisson will actually throw an `IllegalStateException` when `strings.add("a")` is called on netty thread, this exception gets ignored and isn't visible in test output. As we typically just test that the callback was called in correct context I just removed the extra redisson call, an alternative that might fix this would be to replace `whenComplete` with `whenCompleteAsync`.

### DIFF
--- a/instrumentation/redisson-3.0/javaagent/src/test/groovy/RedissonAsyncClientTest.groovy
+++ b/instrumentation/redisson-3.0/javaagent/src/test/groovy/RedissonAsyncClientTest.groovy
@@ -8,7 +8,6 @@ import io.opentelemetry.semconv.trace.attributes.SemanticAttributes
 import org.redisson.Redisson
 import org.redisson.api.RBucket
 import org.redisson.api.RFuture
-import org.redisson.api.RList
 import org.redisson.api.RSet
 import org.redisson.api.RedissonClient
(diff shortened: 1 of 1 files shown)


trask (approved):
> is currently the most flaky test that we have

thx for persisting through all of these ❤️

Issue #5381, opened 2022-02-15
OpenTelemetryHandlerMappingFilter.class cannot be opened When logback.xml level="DEBUG"


domyway:
Enable the instrumentation agent using the -javaagent flag to the JVM
> -javaagent:/opentelemetry-javaagent-1.10.1.jar

And config logback.xml
```
<?xml version="1.0" encoding="UTF-8"?>
<configuration debug="false">
    <appender name="console" class="ch.qos.logback.core.ConsoleAppender">
        <encoder>
            <pattern>
                %d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n
            </pattern>
        </encoder>
    </appender>
    <root level="DEBUG">
        <appender-ref ref="console"/>
    </root>
</configuration>
```

run my app, console output : 
```
18:11:14.872 [main] DEBUG o.s.c.a.ConfigurationClassUtils - Could not find class file for introspecting configuration annotations: org.springframework.web.servlet.OpenTelemetryHandlerMappingFilter
java.io.FileNotFoundException: class path resource [org/springframework/web/servlet/OpenTelemetryHandlerMappingFilter.class] cannot be opened because it does not exist
	at org.springframework.core.io.ClassPathResource.getInputStream(ClassPathResource.java:180)
	at org.springframework.core.type.classreading.SimpleMetadataReader.getClassReader(SimpleMetadataReader.java:55)
	at org.springframework.core.type.classreading.SimpleMetadataReader.<init>(SimpleMetadataReader.java:49)
	at org.springframework.core.type.classreading.SimpleMetadataReaderFactory.getMetadataReader(SimpleMetadataReaderFactory.java:103)
	at org.springframework.boot.type.classreading.ConcurrentReferenceCachingMetadataReaderFactory.createMetadataReader(ConcurrentReferenceCachingMetadataReaderFactory.java:86)
	at org.springframework.boot.type.classreading.ConcurrentReferenceCachingMetadataReaderFactory.getMetadataReader(ConcurrentReferenceCachingMetadataReaderFactory.java:73)
	at org.springframework.core.type.classreading.SimpleMetadataReaderFactory.getMetadataReader(SimpleMetadataReaderFactory.java:81)
	at org.springframework.context.annotation.ConfigurationClassUtils.checkConfigurationClassCandidate(ConfigurationClassUtils.java:112)
[...]

PR #3101, opened 2021-05-27
Fix potential ClassNotFoundException

trask:
Resolves #3065.

I have also seen this error on jboss with a complex ear deployment, and this same fix resolved that issue as well (though I was unable to obtain a repro in that case - thank you @ryanrupp for the excellent repro attached to #3065!).

PR #5387, opened 2022-02-16
Setting bean class name overwrites bean class

laurit:
Hopefully resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5381
Spring bean definition keeps bean class and bean class name in the same field and uses instanceof to figure out which it has.


trask (approved):
this looks familiar 😄 #3101

PR #5388, opened 2022-02-16
Use windows-2019 for smoke tests

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5390
`windows-latest` seems to have changed to windows 2022


trask (approved):
Can you open an issue to track the issue with `windows-latest`?

anuraaga (approved):

Issue #5401, opened 2022-02-18
Error while starting prometheus metrics exporter


BodySplash:
**Describe the bug**
While using opentelemtry agent, when configured to use prometheus as metrics exporter, it crashed with these error : 

```
ERROR io.opentelemetry.javaagent.OpenTelemetryAgent
java.lang.NoClassDefFoundError: com/sun/net/httpserver/HttpHandler
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:375)
        at io.opentelemetry.sdk.autoconfigure.ClasspathUtil.checkClassExists(ClasspathUtil.java:15)
        at io.opentelemetry.sdk.autoconfigure.MetricExporterConfiguration.configurePrometheusMetricReader(MetricExporterConfiguration.java:164)
        at io.opentelemetry.sdk.autoconfigure.MetricExporterConfiguration.configureExporter(MetricExporterConfiguration.java:40)
        at io.opentelemetry.sdk.autoconfigure.MeterProviderConfiguration.configureMeterProvider(MeterProviderConfiguration.java:43)
        at io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdkBuilder.build(AutoConfiguredOpenTelemetrySdkBuilder.java:313)
        at io.opentelemetry.javaagent.tooling.OpenTelemetryInstaller.installOpenTelemetrySdk(OpenTelemetryInstaller.java:43)
        at io.opentelemetry.javaagent.tooling.AgentInstaller.installBytebuddyAgent(AgentInstaller.java:124)
        at io.opentelemetry.javaagent.tooling.AgentInstaller.installBytebuddyAgent(AgentInstaller.java:95)
        at io.opentelemetry.javaagent.tooling.AgentStarterImpl.start(AgentStarterImpl.java:65)
        at io.opentelemetry.javaagent.bootstrap.AgentInitializer.initialize(AgentInitializer.java:35)
        at io.opentelemetry.javaagent.OpenTelemetryAgent.startAgent(OpenTelemetryAgent.java:55)
        at io.opentelemetry.javaagent.OpenTelemetryAgent.premain(OpenTelemetryAgent.java:44)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #5406, opened 2022-02-19
Revert "Always set parent of agent class loader to bootstrap class loader (#5169)"

trask:
Resolves #5401 by reverting #5169. I will re-open #5156. I would open an issue to create a smoke test for the prometheus exporter, but we already have one #4516 😞.

Unfortunately, this PR is creating a regression to fix a regression 😞. But the regression it's fixing is a high-priority long existing use case, while the regression its causing is a more recently discovered issue related to JPMS support.

PR #5408, opened 2022-02-19
Load jdk http server classes from platform class loader

laurit:
Alternative to https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5406
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5401
Note that when using jpms user will need to use `--add-modules jdk.httpserver` so it would be best if prometheus exporter did not use these classes.


trask (approved):
nice 👍

Issue #4516, opened 2021-10-27
Add a smoke test verifying that the prometheus metrics exporter works


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
We don't have any smoke tests that verify that the prometheus metrics are properly exported.

**Additional context**
https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/4515
https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4460

PR #5417, opened 2022-02-21
Add prometheus smoke test

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/4516

### DIFF
--- a/smoke-tests/src/test/java/io/opentelemetry/smoketest/windows/WindowsTestContainerManager.java
+++ b/smoke-tests/src/test/java/io/opentelemetry/smoketest/windows/WindowsTestContainerManager.java
@@ -76,8 +76,6 @@ public class WindowsTestContainerManager extends AbstractTestContainerManager {
             .exec()
             .getId();
 
-    String backendSuffix = "-windows-20210611.927888723";
-
     String backendImageName =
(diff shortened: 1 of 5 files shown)


smoke-tests/src/test/java/io/opentelemetry/smoketest/windows/WindowsTestContainerManager.java
```diff
@@ -134,8 +132,13 @@ public int getTargetMappedPort(int originalPort) {
       String jvmArgsEnvVarName,
       Map<String, String> extraEnv,
       List<ResourceMapping> extraResources,
+      List<Integer> extraPorts,
       TargetWaitStrategy waitStrategy,
       String[] cmd) {
+    if (extraPorts != null && !extraPorts.isEmpty()) {
+      throw new UnsupportedOperationException("extra ports not supported");
```
trask, line 139: what was the issue with passing it to `withExposedPorts()` below?

trask (approved):
❤️

smoke-tests/src/test/java/io/opentelemetry/smoketest/windows/WindowsTestContainerManager.java
laurit: There is also the `withPortBindings` part. I don't have an easy way to test on windows so I chose to leave it unimplemented instead of adding something that might not work. This test doesn't run on windows anyway.

PR #5419, opened 2022-02-22
More release automation

trask:


.github/workflows/release.yml
```diff
@@ -240,3 +242,28 @@ jobs:
...
+
+  create-pull-request-against-main:
+    needs: release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+
+      - name: Bump download link version
+        run: |
+          sed -Ei "s,https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v[0-9]+.[0-9]+.[0-9]+/,https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/download/v${{ needs.release.outputs.release-version }}/," README.md
```
anuraaga, line 254: Was wondering whether for main branch it doesn't make sense to just use `latest`

trask: I think that makes sense, it does revert the decision in #4774, but we've recently decided that `main` docs reflect SNAPSHOT state anyways (for some reason those things connect for me, even though I'm not sure they really do)

mateuszrzeszutek (approved):

trask:
@mateuszrzeszutek is your approval a vote to stay with versioned download link on `README.md` vs using `latest`?

mateuszrzeszutek:
I think it makes slightly more sense to have versioned links -- READMEs on release branches and tags (that stem out from `main`) should have proper versions then.

trask:
oh, I was not clear, @anuraaga's proposal is to keep versioned links on the tags, only using `latest` link on `main` (which would eliminate the need for bumping the version on `main`)

mateuszrzeszutek:
Oh, so we'll replace the latest link with versioned link when the branch is cut? In that case, using `latest` seems to be a better choice 👍

trask:
Closing, will send new PR to change to `latest` link on `main` branch, and update the release branch automation as needed for that change.

PR #5420, opened 2022-02-22
Add possibility to configure Resource from Spring Boot properties / environment

### LINKED ISSUE #5414: Setting Service Name from Spring Boot application properties
I would like to set service name from Spring Boot properties so that when the spans are sent to Jaeger the service name matches the actual spring boot service name and is not `unknown`. As far as I can see Spring Boot autoconfiguration in `spring-boot-autoconfigure` module does not give any opportunity to set `ResourceAttributes.SERVICE_NAME`. If it is possbile to set it somehow with current implementation I would be glad to know the way. If not I would be happy to maybe contribute it.

mkrzywanski:
Would resolve #5414.

Added possibility to configure `Resource` with Spring boot properties so that I can change the service name for example so that it is exported to Jaeger. For example : 
```
otel:
  resource:
    attributes:
      service:
        name: test
```
Or through environment variable like : `OTEL_RESOURCE_ATTRIBUTES_SERVICE_NAME=test`
All properties under `otel.resource.attributes` will be passed to configured `Resource`.

If this change is ok I can also modify README and add information about this.

### DIFF
--- a/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
+++ b/instrumentation/spring/spring-boot-autoconfigure/src/main/java/io/opentelemetry/instrumentation/spring/autoconfigure/OpenTelemetryAutoConfiguration.java
@@ -21,7 +21,6 @@ import io.opentelemetry.sdk.trace.samplers.Sampler;
 import java.util.Collections;
 import java.util.List;
 import org.springframework.beans.factory.ObjectProvider;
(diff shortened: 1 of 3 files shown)


trask (commented):
thx @mkrzywanski! sorry for the delay in reviewing. can you review #5359 and see if that has everything you need?

mkrzywanski:
thanks @trask for having a look. Yes it looks fine for me. This PR and the issue that I created can be closed as soon as the PR that you linked is merged.

trask:
thanks @mkrzywanski! closing now that #5359 is merged

Issue #5261, opened 2022-01-28
Cross test javaagent instrumentation against other instrumentations targeted at the same library?


trask:
E.g. netty-3.8 instrumentation is tested with netty-4.0 and netty-4.1 instrumentations applied.

Should we apply this pattern to all instrumentations with multiple versions, or decide that the muzzle range checks are all we need and abandon this practice?

iNikem:
I already expressed my opinion, that we should test with the whole agent, not partial agents prepared specifically for a given module :)

mateuszrzeszutek:
> I already expressed my opinion, that we should test with the whole agent, not partial agents prepared specifically for a given module :)

That indeed sounds like a good idea; especially in light of our recent Armeria vs. Netty discussion 😄 
And if we ever need to test a particular instrumentation in isolation, we can create a test suite/set that disables unwanted instrumentations.

trask:
> we should test with the whole agent, not partial agents prepared specifically for a given module

this sounds good to me, any thoughts how to implement?

iNikem:
> > we should test with the whole agent, not partial agents prepared specifically for a given module
> 
> this sounds good to me, any thoughts how to implement?

`io.opentelemetry.instrumentation.javaagent-testing.gradle.kts` uses a minimal `agent-for-testing` and adds module-specific extension. It may depend on the full javaagent instead.

iNikem:
@anuraaga was against this idea in the past because allegedly building the full agent will take too much time.

trask:
sounds worth giving it a try and seeing what the perf impact is

trask:
Closing, I think we are doing this now.

PR #5430, opened 2022-02-23
Remove JAXRS HttpClient instrumentations

trask:
The Jersey, CXF and RESTEasy http client instrumentations are producing duplicate CLIENT spans currently, because they use HttpURLConnection by default (which looks pluggable at least [in some cases](https://docs.jboss.org/resteasy/docs/3.6.0.Final/userguide/html/RESTEasy_Client_Framework.html#transport_layer)).

(btw another reason for https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5261#issuecomment-1023924026)

The standard solution of course is to make the client spans created by these instrumentations "active" before they delegate to underlying http client library.

This is surprisingly difficult because of the layer at which this instrumentation is written.

The instrumentation is written as a JAX-RS `ClientRequestFilter` and `ClientResponseFilter`, I suspect in order to get access to the underlying HTTP url and response code, which aren't readily available at the JAXRS layer.

But this makes things hard already because the `ClientResponseFilter` doesn't get called for exceptions (and thus the various and slightly odd JAXRS implementation-specific instrumentations).

And it makes it hard to propagate context because it's not clear to me that these two filters will be called in the same thread (e.g. JAXRS supports async http client calls returning `Future`).

But then I am really wondering what's the value of these JAXRS client spans over the underlying http client spans.

If the JAXRS client were captured at the top level where the application code interacted with it, I would see value. But as it is, the spans are captured in between where the application code interacts with it and the underlying http client, which seems to have very limited value.

Maybe we should capture these JAXRS spans where the application code interacts with it (e.g. `javax.ws.rs.client.AsyncInvoker`) and capture them as INTERNAL spans with their own convention. I think this could parallel the hibernate instrumentation, since they are both "object" interfaces on top of underlying http/sql.

But I'm not personally convinced that these INTERNAL spans would have a lot of value (similar to my opinion about the hibernate INTERNAL spans), so I'm proposing we just delete this instrumentation (but keep the tests, which still pass!) and wait for more user feedback to justify value in capturing INTERNAL spans.

EDIT on third thought, I'm sort of not sure what the point of keeping only the tests is, although they did uncover a bug in the HttpURLConnection instrumentation, and they are already written....


trask:
@laurit curious your thoughts on this

laurit (approved):
@trask I agree with your analysis. The duplicate client span is definitely a bug. I wasn't able to find additional value over what `http-url-connection` or `apache-httpclient` instrumentation provides.

anuraaga (approved):

PR #5342, opened 2022-02-10
Split out CodeAttributesGetter

mateuszrzeszutek:
Part of #5291

PR #5436, opened 2022-02-24
Fix regression in spring-scheduling span name

trask:
This fixes (and adds test for) a small (unreleased) regression introduced in #5342.

### DIFF
--- a/instrumentation/spring/spring-scheduling-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/scheduling/SpringSchedulingCodeAttributesGetter.java
+++ b/instrumentation/spring/spring-scheduling-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/scheduling/SpringSchedulingCodeAttributesGetter.java
@@ -14,7 +14,7 @@ public class SpringSchedulingCodeAttributesGetter implements CodeAttributesGette
   public Class<?> codeClass(Runnable runnable) {
     if (runnable instanceof ScheduledMethodRunnable) {
       ScheduledMethodRunnable scheduledMethodRunnable = (ScheduledMethodRunnable) runnable;
-      return scheduledMethodRunnable.getTarget().getClass();
+      return scheduledMethodRunnable.getMethod().getDeclaringClass();
     } else {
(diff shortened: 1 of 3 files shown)


instrumentation/spring/spring-scheduling-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/scheduling/SpringSchedulingCodeAttributesGetter.java
```diff
@@ -14,7 +14,7 @@
   public Class<?> codeClass(Runnable runnable) {
     if (runnable instanceof ScheduledMethodRunnable) {
       ScheduledMethodRunnable scheduledMethodRunnable = (ScheduledMethodRunnable) runnable;
-      return scheduledMethodRunnable.getTarget().getClass();
+      return scheduledMethodRunnable.getMethod().getDeclaringClass();
```
laurit, line 17: I suspect that it might be better to attempt to unwrap the target object.
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/9f00cf38d5388cc1526744999ee1bf0576c1b8c7/instrumentation/spring/spring-integration-4.1/library/src/main/java/io/opentelemetry/instrumentation/spring/integration/TracingChannelInterceptor.java#L252-L268
Unfortunately this is a common issue with getting class from object, there is no easy way to tell whether we got a proxy and how to unwrap it.

laurit (approved):

instrumentation/spring/spring-scheduling-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/spring/scheduling/SpringSchedulingCodeAttributesGetter.java
trask: I gave this a try, but the unwrapped class in the test I added was still `EnhancedClassTaskConfig$$EnhancerByCGLIB$$b6a5b321@4ac197a7`

PR #5437, opened 2022-02-24
Add missing entry to change log

trask:
From @anuraaga's https://github.com/open-telemetry/opentelemetry-lambda/pull/221#discussion_r813433955

Updated release notes already.


anuraaga (commented):
We probably need to highlight the artifact rename at the top to make it easier to migrate. Not sure we did that for other renames like logging instrumentation but probably should be doing it.

laurit (approved):

PR #5446, opened 2022-02-24
Add migration notes to change log

trask:
let me know what you think, and I'll update the release notes with the same


anuraaga (approved):
Thanks! If my build breaks when updating a version release notes are the first thing I go to so this is really helpful

laurit (approved):

PR #5450, opened 2022-02-25
Add markdown spell check

trask:
This PR is built on top of #5449 (to avoid conflicts).

This `misspell` check is used in the spec repo, so thought I'd try it out here.

I'm not sure if it's worth adding or not, it only found 4 misspellings https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5450/commits/2c77ef13f1c586702b475af4c39d8cca7bc902ec.

It's kind of cool though, since it only looks for common misspellings it solves the custom dictionary problem that a normal spell checker would require for a project like this.

I'm weakly in favor of adding it.


laurit (approved):

.github/workflows/ci.yml
```diff
@@ -191,13 +191,42 @@ jobs:
...
+  markdown-misspell-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+
+      - name: Check markdown for common misspellings
+        run: |
+          curl -L -o ./install-misspell.sh https://git.io/misspell
+          sh ./install-misspell.sh
+          ./bin/misspell -error ./**/*.md
```
anuraaga, line 218: I think spell checking on Java code could be nice, not sure if it will work well if not restricted to only Java comments though.

anuraaga (approved):

.github/workflows/ci.yml
trask: great idea, found several more, definitely makes the tool more worth it

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/cache/internal/weaklockfree/WeakConcurrentMap.java
```diff
@@ -74,7 +74,9 @@
 
   private final boolean reuseKeys;
 
-  /** @param cleanerThread {@code true} if a thread should be started that removes stale entries. */
+  /**
+   * @param cleanerThread {@code true} if a thread should be started that removes stale entries.
```
anuraaga, line 78: Apparently misspell actually understands java well enough...to be incompatible with spotless...

Issue #5050, opened 2022-01-08
Add peer service support back to v2_6/CouchbaseNetworkInstrumentation.java


iNikem:
It was removed when migrating to Instrumenter codebase

PR #5451, opened 2022-02-25
Add peer service support back to couchbase26

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5050


settings.gradle.kts
```diff
@@ -162,7 +162,8 @@ include(":instrumentation:internal:internal-reflection:javaagent-integration-tes
 include(":instrumentation:internal:internal-url-class-loader:javaagent")
 include(":instrumentation:internal:internal-url-class-loader:javaagent-integration-tests")
 include(":instrumentation:couchbase:couchbase-2.0:javaagent")
-include(":instrumentation:couchbase:couchbase-2.0:javaagent-unit-tests")
+include(":instrumentation:couchbase:couchbase-2.0-common:javaagent")
+include(":instrumentation:couchbase:couchbase-2.0-common:javaagent-unit-tests")
```
trask, line 166: similar to `log4j-context-data-2-common`
```suggestion
include(":instrumentation:couchbase:couchbase-2-common:javaagent")
include(":instrumentation:couchbase:couchbase-2-common:javaagent-unit-tests")
```

instrumentation/couchbase/couchbase-2.0-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_0/CouchbaseRequestInfoHolder.java
```diff
@@ -0,0 +1,46 @@
...
+import io.opentelemetry.context.ContextKey;
+import io.opentelemetry.context.ImplicitContextKeyed;
+import javax.annotation.Nullable;
+
+public final class CouchbaseRequestInfoHolder implements ImplicitContextKeyed {
+
+  private static final ContextKey<CouchbaseRequestInfoHolder> KEY =
+      named("opentelemetry-couchbase-request-key");
+
+  private final CouchbaseRequestInfo couchbaseRequest;
```
trask, line 20: since this is the only field and it's final, do we need the holder instance?

instrumentation/couchbase/couchbase-2-common/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/couchbase/v2_0/CouchbaseRequestInfoHolder.java
```diff
@@ -0,0 +1,29 @@
...
+
+package io.opentelemetry.javaagent.instrumentation.couchbase.v2_0;
+
+import static io.opentelemetry.context.ContextKey.named;
+
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.ContextKey;
+import javax.annotation.Nullable;
+
+public final class CouchbaseRequestInfoHolder {
```
trask, line 14: the name doesn't seem right anymore now that it's not exactly a "holder", maybe just move the static stuff into `CouchbaseRequestInfo`? it looks like we have that pattern a couple other places, e.g. `AppServerBridge`, `ServletContextPath`, `GrizzlyErrorHolder`

trask (approved):

PR #5455, opened 2022-02-27
Fix aws-core testLatestDeps

### LINKED ISSUE #5454: Nightly build without cache #266 failed
<a href="https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions/runs/1905128784">
Nightly build without cache #266</a> failed. Please take a look and fix it ASAP.

trask:
Closes #5454


laurit:
This test seems to fail because the way retries are configured gets ignored in the lates release. So instead of 1 retry the test does 3 retries which makes it fail unexpectedly with 500 because the test is set up to only reply to 2 request.

trask:
let's pretend this didn't happen. sending fresh PR

PR #5456, opened 2022-02-27
Add gradle cache to muzzle jobs

### LINKED ISSUE #5440: Try adding gradle cache for muzzle CI jobs
These jobs flake from time to time downloading from maven central. It would be nice if we could use gradle cache for them to limit this flakiness. The concern is that these jobs may fill up our github action cache limit. I'm not sure how to monitor the github action cache, but the cache limit was recently increased from 5gb to 10gb, so it may be worth giving it a try and monitoring build performance for any impact to other jobs due to overflowing the cache limit.

trask:
Still seeing occasional muzzle failure due to maven central downloads: #5453

Closes #5440

I think this is worth trying, but need to monitor CI build times closely for a while to make sure this doesn't cause problems with overflowing the github action cache size limit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,6 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: ${{ matrix.module }}:muzzle
-          cache-read-only: true
 
   examples:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


trask:
Closing, I think we should investigate existing gradle cache failures first, to see if this is reall needed and also to make sure this doesn't make them worse.

Some existing gradle cache failures:

```
Warning: Failed to restore v6-gradle|Windows|smoke-test[fc53b6f6f757636c2b68de70cc9ce7cf]-81594fb52d86517e1eedefb86682e12a04d56f8a: Error: Cache service responded with 429
```

and

```
Warning: Failed to restore dependencies-2be958cf54776179ee716a8ae5443a13: RangeError [ERR_OUT_OF_RANGE]: The value of "length" is out of range. It must be >= 0 && <= 2147483647. Received 2332659880
```

PR #5459, opened 2022-02-28
Clean up examples github actions jobs

trask:
the diff is much more readable if you hide whitespace

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,37 +208,29 @@ jobs:
           distribution: adopt
           java-version: 11
 
-      - name: Local publish of artifacts
+      - name: Set up gradle cache
         uses: gradle/gradle-build-action@v2
-        with:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -208,37 +208,29 @@ jobs:
...
-        with:
-          arguments: build --init-script ../../.github/scripts/local.init.gradle.kts
-          build-root-directory: examples/extension
+        run: ./gradlew build --init-script ../../.github/scripts/local.init.gradle.kts
+        working-directory: examples/extension
 
       - name: Run muzzle check against extension
-        uses: gradle/gradle-build-action@v2
-        with:
-          cache-read-only: true
```
trask, line 239: I realized that `cache-read-only` can't be applied separately, another good reason to move to a single `gradle/gradle-build-action@v2` step

laurit (approved):

PR #5460, opened 2022-02-28
First reusable workflow

trask:
https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow


anuraaga (approved):
🎉

laurit (approved):

Issue #5466, opened 2022-02-28
Duplicate class definition of ContextDataProvider 


pleeplop:
**Describe the bug**
Log4j2InstrumentationModule used in javaagent version 1.11.0 throws java.lang.ClassNotFoundException when loading org.apache.logging.log4j.core.util.ContextDataProvider.
It was working in version 1.10.0

Log4j2InstrumentationModule defined as TypeMatcher the class ContextDataProvider but a class injected also implements ContextDataProvider causing the interface to be loaded while it's being transformed, which leads to duplicate class definition error after the interface is transformed and the triggering class loader tries to load it.

This issue was already known (see the comments of the previous version) https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5216/files#diff-0225235d7840055cce2f3782f412db7bae6c574d862c59d5e78e53669994cfd3

Instrumentation works again if i revert this change https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5216/ 

**Steps to reproduce**
You should be able to reproduce it with javaagent and log4j2. Let me know if you need more details.

**What did you expect to see?**
I will expect to be able to load log4j-context-data-2.16 instrumentation module with javaagent.

**What did you see instead?**
```
[otel.javaagent 2022-02-28 10:41:48:273 +0100] [Start Level: Equinox Container: 752c254d-46c0-4b48-9525-2d6dea8a0e2f] ERROR io.opentelemetry.javaagent.tooling.HelperInjector - Error preparing helpers while processing interface org.apache.logging.log4j.core.util.ContextDataProvider for log4j-context-data. Failed to inject helper classes into instance org.eclipse.osgi.internal.loader.EquinoxClassLoader@81f7bf7[org.ops4j.pax.logging.pax-logging-log4j2:2.0.14(id=7)]
java.lang.IllegalStateException: java.lang.NoClassDefFoundError: org/apache/logging/log4j/core/util/ContextDataProvider
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection$Dispatcher$UsingUnsafeInjection.defineClass(ClassInjector.java:1044)
	at net.bytebuddy.dynamic.loading.ClassInjector$UsingReflection.injectRaw(ClassInjector.java:283)
	at io.opentelemetry.javaagent.tooling.HelperInjector.injectClassLoader(HelperInjector.java:262)
[...]

PR #5467, opened 2022-02-28
Add Azure SDK instrumentation

trask:


instrumentation/azure-core/azure-core-1.14/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/azurecore/v1_14/AzureSdkInstrumentationModule.java
```diff
@@ -0,0 +1,76 @@
...
+      //
+      // this is a list of all classes that call one of these:
+      // * ServiceLoader.load(AfterRetryPolicyProvider.class)
+      // * ServiceLoader.load(Tracer.class)
+      return named("com.azure.core.http.policy.HttpPolicyProviders")
+          .or(named("com.azure.core.util.tracing.TracerProxy"))
+          .or(named("com.azure.cosmos.CosmosAsyncClient"))
+          .or(named("com.azure.messaging.eventhubs.EventHubClientBuilder"))
+          .or(named("com.azure.messaging.eventhubs.EventProcessorClientBuilder"))
+          .or(named("com.azure.messaging.servicebus.ServiceBusClientBuilder"));
```
trask, line 68: I thought we could simplify this similar to #5216 (thanks to #5185), but an issue (#5466) was just opened about #5216 causing a regression

lmolkova (commented):
As per offline discussion, we should keep suppression in 1.19 version. Otherwise looks great!

anuraaga (approved):
Wouldn't block it but want to point out that there seems to be quite a small span between 1.14 (8/2018) and 1.19 (1/2019) that having multiple instrumentation seems a bit overkill.

trask:
> Wouldn't block it but want to point out that there seems to be quite a small span between 1.14 (8/2018) and 1.19 (1/2019) that having multiple instrumentation seems a bit overkill.

those dates don't look right to me: https://search.maven.org/artifact/com.azure/azure-core

anuraaga:
Hmm, I got it from mvnrepository.com which surprisingly is more accurate it seems

https://mvnrepository.com/artifact/com.microsoft.azure/azure/1.14.0
https://repo1.maven.org/maven2/com/microsoft/azure/azure/

trask:
oh, you're looking at the groupId/artifactId

anuraaga:
Oops that explains it. So I guess 1.14 being less than a year old, so pretty new, is motivation to instrument it that seems reasonable.

trask:
Ooops, I forgot was going to convert tests to Java, will send separate PR.

PR #5468, opened 2022-02-28
Fix android desugaring for HashMap.forEach

laurit:
See https://github.com/open-telemetry/opentelemetry-java/issues/4161 and https://github.com/open-telemetry/opentelemetry-java/pull/4221
Android desugaring is not smart enough to handle calling HashMap.forEach through Attributes.forEach. This pr helps out desugaring by adding an explicit call to HashMap.forEach that desugaring can rewrite it.


trask (approved):
👍

anuraaga (approved):

PR #5469, opened 2022-02-28
Remove unnecessary fetch-depth

trask:
And follow convention for actions/checkout (not having a `name`)

### DIFF
--- a/.github/workflows/nightly-benchmark-overhead.yml
+++ b/.github/workflows/nightly-benchmark-overhead.yml
@@ -9,10 +9,8 @@ jobs:
   run-overhead-tests:
     runs-on: ubuntu-latest
     steps:
-      - name: checkout
-        uses: actions/checkout@v2.3.4
-      - name: checkout
-        uses: actions/checkout@v2.3.4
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly-benchmark-overhead.yml
```diff
@@ -9,10 +9,8 @@ jobs:
   run-overhead-tests:
     runs-on: ubuntu-latest
     steps:
-      - name: checkout
-        uses: actions/checkout@v2.3.4
-      - name: checkout
-        uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v2.3.4
+      - uses: actions/checkout@v2.3.4
```
laurit, line 13:
```suggestion
```
looks like a duplicate line

laurit (approved):

.github/workflows/nightly-benchmark-overhead.yml
trask: the second one is checking out `gh-pages` branch to a subdirectory

PR #5470, opened 2022-02-28
Convert smoke tests into reusable workflow

trask:


laurit (approved):
the only change I'd make would be to skip running windows smoke tests for pull requests

trask:
> the only change I'd make would be to skip running windows smoke tests for pull requests

just confirming, because they're basically redundant, or some other reason?

laurit:
> > the only change I'd make would be to skip running windows smoke tests for pull requests
> 
> just confirming, because they're basically redundant, or some other reason?

for faster feedback, windows tests just take more time to run and the chance of breaking agent on only windows isn't too high

trask:
👍 got it, I'm going to merge and will try that in a follow-up

PR #5483, opened 2022-03-02
Move build to reusable workflow

trask:


.github/workflows/build.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Set up JDK 11 for running Gradle
+        uses: actions/setup-java@v2
+        with:
+          distribution: adopt
+          java-version: 11
+
+      - name: Start deadlock detector
+        run: .github/scripts/deadlock-detector.sh
+
+      - name: Build
```
anuraaga, line 35: Is this a good time to change this to assemble instead of build? IIUC, we run tests on Java 11 twice, once as part of build and once as part of test

anuraaga (approved):

.github/workflows/build.yml
trask: oh, I think I need to add `check` in there too...

anuraaga: `check` would run the tests - I had a proposal to change `test` to `check` in the test workflow to make defining additional tests easier, without our current boilerplate when defining special tests. Haven't done that yet but don't believe it affects the build workflow.

`test`
`check` -> `test` + spotless/checkstyle/etc
`assemble` -> `jar`, `javadoc`, `sources`
`build` -> `assemble` + `check`

anuraaga: Oh with my table I get it now, heh. Let's just leave it as build for now then

mateuszrzeszutek (approved):

PR #5488, opened 2022-03-03
Better GitHub action cache management

trask:
~Currently built on top of #5482, #5483, #5486, #5487~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,7 +57,8 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: test -PtestJavaVersion=${{ matrix.test-java-version }} -PtestJavaVM=${{ matrix.vm }} -Porg.gradle.java.installations.paths=${{ steps.setup-test-java.outputs.path }} -Porg.gradle.java.installations.auto-download=false ${{ inputs.no-build-cache && ' --no-build-cache' || '' }}
-          cache-read-only: ${{ inputs.cache-read-only }}
+          # only push cache for one matrix option since github action cache space is limited
+          cache-read-only: ${{ inputs.cache-read-only || matrix.test-java-version != 11 || matrix.vm != 'hotspot' }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -57,7 +57,8 @@ jobs:
         uses: gradle/gradle-build-action@v2
         with:
           arguments: test -PtestJavaVersion=${{ matrix.test-java-version }} -PtestJavaVM=${{ matrix.vm }} -Porg.gradle.java.installations.paths=${{ steps.setup-test-java.outputs.path }} -Porg.gradle.java.installations.auto-download=false ${{ inputs.no-build-cache && ' --no-build-cache' || '' }}
-          cache-read-only: ${{ inputs.cache-read-only }}
+          # only push cache for one matrix option since github action cache space is limited
+          cache-read-only: ${{ inputs.cache-read-only || matrix.test-java-version != 11 || matrix.vm != 'hotspot' }}
```
anuraaga, line 61: I'm not too optimistic about this helping with Maven Central failures due to this

https://github.com/gradle/gradle-build-action#cache-keys

```
To reduce duplication between cache entries, certain artifacts are cached independently based on their identity. Artifacts that are cached independently include downloaded dependencies, downloaded wrapper distributions and generated Gradle API jars. For example, this means that all jobs executing a particular version of the Gradle wrapper will share common entries for wrapper distributions and for generated Gradle API jars.
```

downloaded dependencies seem to be shared among all jobs

trask: the Gradle User Home cache (which is specific to the job) is ~500mb.

the smoke-test matrix currently produces 16 of these, which will fill up 8gb out of the 10gb cache just by itself.

I'm hoping that reducing these will leave more room for dependency cache, and reduce the number of dependency cache failures:

```
Did not restore dependencies with key dependencies-36a648976de3a91328824fba3af79370 to /home/runner/.gradle/caches/modules-*/files-*/*/*/*/*
```

anuraaga (approved):

Issue #5385, opened 2022-02-16
Consider changing the default for suppress-messaging-receive-spans to true


trask:
Capturing receive spans and having consumer spans parent the receive span instead of parenting the producer span has been confusing for users since most(?) backends do not support links well yet and so it appears that distributed tracing is broken.

User reported issues that were resolved by setting this flag: #4509 #4582 #4783 #4854 #5340

cc @pyohannes

anuraaga:
A couple more
https://github.com/aws-observability/aws-otel-java-instrumentation/issues/153
https://github.com/aws-observability/aws-otel-java-instrumentation/issues/155
and 2+ internal tickets

PR #5500, opened 2022-03-04
Disable the messaging receive span telemetry by default

mateuszrzeszutek:
Fixes https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5385

I introduced a new property for this (`otel.instrumentation.messaging.experimental.receive-telemetry.enabled`) - I figured that in case somebody wants to turn it on it probably makes more sense to say `enabled=true` than `suppress=false`. And the majority of our properties use the `...enabled` form, there's only a few using `...suppress`.


trask (approved):

trask:
closing and re-opening, b/c retriggering build manually doesn't seem to pick up the `test latest deps` label

trask:
oh I forgot, `test latest deps` is currently failing due to #5503

anuraaga (approved):

PR #5501, opened 2022-03-04
Write http server tests in java

laurit:
Add an option to write http server tests in java as junit tests and convert grails3 test to java. Being able to write tests in java will hopefully allow us to upgrade to groovy 4 (currently grails tests don't run with groovy 4) which hopefully will let us run some tests on jdk17.


testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
```diff
@@ -0,0 +1,707 @@
...
+  protected void assertTheTraces(
+      int size,
+      String traceId,
+      String parentId,
+      String method,
+      ServerEndpoint endpoint,
+      String errorMessage,
+      AggregatedHttpResponse response) {
+    List<Consumer<TraceAssert>> assertions = new ArrayList<>();
+    for (int i = 0; i < size; i++) {
```
laurit, line 438: This doesn't have all the span types that groovy version has.

testing-common/src/main/groovy/io/opentelemetry/instrumentation/test/base/HttpServerTest.groovy
```diff
@@ -282,257 +176,137 @@ abstract class HttpServerTest<SERVER> extends InstrumentationSpecification imple
...
+        String method,
+        ServerEndpoint endpoint,
+        String errorMessage,
+        AggregatedHttpResponse response) {
+      HttpServerTest.this.assertTheTraces(size, traceId, parentId, method, endpoint, errorMessage, response)
+    }
+
+    void assertHighConcurrency(int count) {
+      HttpServerTest.this.assertHighConcurrency(count)
+    }
```
trask, line 231: can we get rid of the local groovy implementations of these two methods and only use the java implementations of them?
```suggestion
```

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
trask: ah, this is why we still have the groovy implementations of those two methods? can you add a comment either here or there explaining, e.g. if there's some current limitation that makes this not possible, or if it's just a TODO to be implemented?

instrumentation/restlet/restlet-1.0/testing/src/main/groovy/io/opentelemetry/instrumentation/restlet/v1_0/AbstractRestletServerTest.groovy
```diff
@@ -156,6 +157,11 @@ abstract class AbstractRestletServerTest extends HttpServerTest<Server> {
     true
   }
 
+  @Override
+  boolean testErrorBody() {
+    false
+  }
+
```
trask, line 164: just curious what I missed in the PR that affected this?

trask (approved):
🥳 !!

I think you read @anuraaga's mind, he was just talking about this yesterday 😁

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
laurit: The main issue is that I don't know how to completely unify java and groovy asserts. Currently groovy can call java asserts but not the other way around.
Here I just didn't add stuff that wasn't needed by the grails test. These can be added later when needed.

instrumentation/netty/netty-3.8/javaagent/src/test/groovy/Netty38ServerTest.groovy
```diff
@@ -66,7 +67,7 @@ class Netty38ServerTest extends HttpServerTest<ServerBootstrap> implements Agent
         if (msg.getMessage() instanceof HttpRequest) {
           def request = msg.getMessage() as HttpRequest
           def uri = URI.create(request.getUri())
-          HttpServerTest.ServerEndpoint endpoint = forPath(uri.path)
+            ServerEndpoint endpoint = forPath(uri.path)
```
laurit, line 70:
```suggestion
          ServerEndpoint endpoint = forPath(uri.path)
```

instrumentation/restlet/restlet-1.0/testing/src/main/groovy/io/opentelemetry/instrumentation/restlet/v1_0/AbstractRestletServerTest.groovy
laurit: This isn't obvious. The original groovy code for this assert was
```groovy
    if (testErrorBody()) {
      response.contentUtf8() == ERROR.body
    }
```
which apparently doesn't really work. Due to `if` it should explicitly use `assert`. So it previously passed only because this assertion wasn't really verified.

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
```diff
@@ -0,0 +1,709 @@
...
+    if (server == null) {
+      echo(getClass().getName() + " can't stop null server");
+      return;
+    }
+    stopServer(server);
+    server = null;
+    echo(getClass().getName() + " http server stopped at: http://localhost:" + port + "/");
+  }
+
+  private static void echo(String message) {
```
anuraaga, line 113: Use logger instead of this method

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
```diff
@@ -0,0 +1,709 @@
...
+    // CHECKSTYLE:OFF
+    System.out.println(message);
+    // CHECKSTYLE:ON
+  }
+
+  public URI buildAddress() {
```
anuraaga, line 119: Can this be `protected`?

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
```diff
@@ -0,0 +1,709 @@
...
+            assertThat(attrs)
+                .containsEntry(
+                    "servlet.request.parameter.test_parameter", new String[] {"test value õäöü"});
+          }
+        });
+
+    return span;
+  }
+
+  private static void assertException(
```
anuraaga, line 628: Can you use `SpanDataAssert.hasException`?

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
```diff
@@ -0,0 +1,709 @@
...
+            }
+
+            trace.hasSpansSatisfyingExactly(spanAssertions);
+          });
+    }
+
+    testing.waitAndAssertTraces(assertions);
+  }
+
+  public SpanDataAssert assertControllerSpan(
```
anuraaga, line 491: Let's make `protected` where possible. Also surprised errorprone isn't suggesting making these static, thought we had that configured

anuraaga (commented):
Yay!

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
laurit: I didn't use `SpanDataAssert.hasException` because it requires an instance of `Throwable` but here we only have exception class and error message. As actually we use only a few exception classes and all of them have a suitable constructor I could try constructing the exception instance reflectively.

trask: > we use only a few exception classes and all of them have a suitable constructor I could try constructing the exception instance reflectively

I think this approach makes the test harder to read, @anuraaga what do you think about something like `SpanDataAssert.hasExceptionSatisying` for the most flexibility here?

@laurit I'm ok merging with any approach here and we can revisit

anuraaga: I think that could be useful, but in this case, can't the places passing an exception class and String error message pass a Throwable instead by calling `new Class(message)`?

mateuszrzeszutek (approved):

testing-common/src/main/java/io/opentelemetry/instrumentation/testing/junit/http/AbstractHttpServerTest.java
laurit: I think it should be done in a separate pr as this pr is already fairly large.

PR #5504, opened 2022-03-04
Convert azure sdk tests to java

trask:


instrumentation/azure-core/azure-core-1.14/javaagent/src/test/java/AzureSdkTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Test
+  void testHelperClassesInjected() {
+
+    assertThat(TracerProxy.isTracingEnabled()).isTrue();
+
+    List<HttpPipelinePolicy> list = new ArrayList<>();
+    HttpPolicyProviders.addAfterRetryPolicies(list);
+
+    assertThat(list).hasSize(1);
```
anuraaga, line 35: Consider doing size and element check together

```
assertThat(list).satisfiesExactly(
  item -> assertThat(item.getClass()...
````

anuraaga (approved):

Issue #5499, opened 2022-03-04
Exception transforming javaagent class


anuraaga:
While debugging a customer issue, I found this debug logging. Is this expected / harmless?

```
[otel.javaagent 2022-03-03 20:08:36:330 +0000] [AD Thread Pool-Global1] DEBUG io.opentelemetry.javaagent.tooling.AgentInstaller$TransformLoggingListener - Failed to handle io.opentelemetry.javaagent.instrumentation.servlet.AsyncRunnableWrapper for transformation on classloader java.net.URLClassLoader@2f4e40d7
net.bytebuddy.pool.TypePool$Resolution$NoSuchTypeException: Cannot resolve type description for io.opentelemetry.javaagent.instrumentation.servlet.ServletHelper
```

mateuszrzeszutek:
Hmm, why would `AsyncRunnableWrapper` even get transformed? It's an injected helper class, I thought we were explicitly skipping these.

anuraaga:
@mateuszrzeszutek Exactly why this line stood out more than other type description errors :)

laurit:
@anuraaga could you add stack trace. Maybe seeing where this class is defined from helps to understand why this happens.  As far as I know we don't explicitly exclude injected classes from transformation but rely on classes defined while transforming a class not triggering a nested transform.

Issue #5493, opened 2022-03-03
Matching conditions can cause instrumentation to not load


anuraaga:
The AWS distro includes log4j instrumentation that relies on the [upstream one](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/instrumentation/log4j/log4j-context-data/log4j-context-data-2.16/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/log4j/v2_16/Log4j2InstrumentationModule.java) to also be loaded. However, when updating from upstream 1.10 to 1.11, for some reason the upstream log4j 2.16 instrumentation could not be applied anymore in one of our smoke tests. I found that changing the type matcher to be the same as the upstream one fixed it

https://github.com/aws-observability/aws-otel-java-instrumentation/pull/156/files#diff-6e33314935e8c5617a790f5014a8fa4f091fc22d5b151bf41f6b7e24d3304212R64

Doing this is not a problem at all and actually makes the code better. But this made me think there is possibly a bug in the agent, I can't reason out why before this change the upstream instrumentation was not applied while after the change it is as I would expect them to be run independent of each other.

Sorry if this is some cache-related red herring that I couldn't figure out, but filing in case there is an issue.

trask:
Could possibly be related to #5466

trask:
discussed with @anuraaga, sounds like the ADOT instrumentation on `org.apache.logging.log4j.core.impl.ThreadContextDataInjector` was injecting a subclass of `org.apache.logging.log4j.core.util.ContextDataProvider`, which triggered loading of `org.apache.logging.log4j.core.util.ContextDataProvider` inside of the `transform()` method, which suppressed the otel java instrumentation which was changed in 1.11.0 to run on loading of `org.apache.logging.log4j.core.util.ContextDataProvider` (since classes loaded inside of `transform()` do not themselves get transformed).

laurit:
We could experiment with not immediately defining the injected classes during the transform, but doing it when the surrounding `defineClass` exits. Or load the from `loadClass` where we currently do the boot delegation when the class is really needed.

trask:
@anuraaga and I were hoping you would have an idea for this ❤️

[...]

PR #5528, opened 2022-03-08
Define helper classes in loadClass

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5499
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5493
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5466
Instead of injecting helper classes during transformation instrument `loadClass` methods of class loaders to load them. This should avoid defining helper class triggering loading additional classes that won't get transformed.


instrumentation/spring/spring-web-3.1/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/springweb/WebApplicationContextInstrumentation.java
```diff
@@ -65,14 +65,16 @@ public static void onEnter(@Advice.Argument(0) ConfigurableListableBeanFactory b
...
-          // Now attempt to load our injected instrumentation class.
-          Class<?> clazz =
+          Class<?> dispatcherServletClass =
               beanFactory
                   .getBeanClassLoader()
+                  .loadClass("org.springframework.web.servlet.DispatcherServlet");
+
+          // Now attempt to load our injected instrumentation class from the same class loader as
+          // DispatcherServlet
+          Class<?> clazz =
```
laurit, line 75: We have a test with wildfly where loading our injected class from `getBeanClassLoader()` fails. Previously it worked because it hit `findLoadedClass` during class loading.
If we do it this way we can probably put `OpenTelemetryHandlerMappingFilter` back into our own package. We needed to put in spring package because we set up the bean definition in the wrong way. We set both name and class that are actually kept in the same field in spring bean definition so it always tried to use loadClass which previously worked only if the class was in spring package.

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/ExposeAgentBootstrapListener.java
```diff
@@ -37,14 +40,21 @@ public void onTransformation(
       JavaModule javaModule,
       boolean b,
       DynamicType dynamicType) {
-    if (javaModule != JavaModule.UNSUPPORTED
-        && javaModule.isNamed()
-        && !javaModule.canRead(agentBootstrapModule)) {
-      logger.debug("Adding module read from {} to {}", javaModule, agentBootstrapModule);
+    exposeModule(javaModule, agentBootstrapModule);
+    if (classLoader != null) {
+      exposeModule(javaModule, JavaModule.of(classLoader.getUnnamedModule()));
```
laurit, line 45: This was previously done in `HelperInjector`.

muzzle/src/main/java/io/opentelemetry/javaagent/tooling/muzzle/HelperReferenceWrapper.java
```diff
@@ -152,20 +152,27 @@ public String toString() {
...
     public HelperReferenceWrapper create(ClassRef reference) {
       return new ReferenceType(reference);
     }
 
     private HelperReferenceWrapper create(String className) {
-      Resolution resolution = classpathPool.describe(className);
-      if (resolution.isResolved()) {
-        return new ClasspathType(resolution.resolve());
+      if (!helperClassPredicate.isHelperClass(className)) {
+        Resolution resolution = classpathPool.describe(className);
```
laurit, line 172: Looking up an injected helper records it in `AgentCachingPoolStrategy` as not found. If that helper is defined before it gets evicted there will be a stack trace about resource not found.

trask: this is helpful explanation, can you turn it into code comment? (especially since breaking it won't fail tests, only log annoying debug messages)

trask (approved):
Nice! I think this is an exciting improvement 🎉

instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/BootDelegationInstrumentation.java
```diff
@@ -100,7 +100,8 @@ public void transform(TypeTransformer transformer) {
   public static class LoadClassAdvice {
 
     @Advice.OnMethodEnter(skipOn = Advice.OnNonDefaultValue.class)
-    public static Class<?> onEnter(@Advice.Argument(0) String name) {
+    public static Class<?> onEnter(
+        @Advice.This ClassLoader classLoader, @Advice.Argument(0) String name) {
```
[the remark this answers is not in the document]
laurit: Thanks remove.

mateuszrzeszutek (approved):

muzzle/src/test/groovy/io/opentelemetry/javaagent/tooling/muzzle/ReferenceMatcherTest.groovy
```diff
@@ -209,7 +212,7 @@ class ReferenceMatcherTest extends Specification {
   def "should fail #desc helper classes that does not implement all abstract methods"() {
     given:
     def reference = ClassRef.builder(className)
-      .setSuperClassName(TestHelperClasses.HelperSuperClass.name)
+      .setSuperClassName(TestAbstractSuperClass.name)
```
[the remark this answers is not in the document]
laurit: This was needed because of the change in `HelperReferenceWrapper` that added `!helperClassPredicate.isHelperClass(className)` check. Now classes that are in `io.opentelemetry.instrumentation` are matched by that check which made a couple of tests fail. I believe the intent of the test was to test injected helper extending a class from application. To ensure that the class that the test extends is treated as an application class and not as a helper I had to copy it into a different package. I also tried moving `TestHelperClasses` but that made a bunch of other tests fail.

PR #5537, opened 2022-03-10
Implement "Prometheus mode" for better micrometer->OTel->Prometheus support

mateuszrzeszutek:
This is another part of the solution to #5292. It does not completely fix that issue, there are still two problems remaining:
* Micrometer histogram buckets should be translated into hints, once the metric hint API is available. For now, users will have to configure SDK views to get their desired buckets.
* `LongTaskTimer` does not really match any OTel instrument well - we're not using an OTel histogram for it (because OTel histogram is for finished records, while this one is supposed to show the ones still inflight). This one instrument would benefit from Bogdan's `MetricProducer` idea, but that's almost impossible to implement as javaagent instrumentation. I don't really have any good idea on how to handle that, so for now we'll have to accept that it's not a perfect shim.

I'll copy that PR over to the SDK repo once https://github.com/open-telemetry/opentelemetry-java/pull/4258 is merged.

### DIFF
--- a/instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistry.java
+++ b/instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistry.java
@@ -53,13 +53,18 @@ public final class OpenTelemetryMeterRegistry extends MeterRegistry {
   private final io.opentelemetry.api.metrics.Meter otelMeter;
 
   OpenTelemetryMeterRegistry(
-      Clock clock, TimeUnit baseTimeUnit, io.opentelemetry.api.metrics.Meter otelMeter) {
+      Clock clock,
+      TimeUnit baseTimeUnit,
+      boolean prometheusMode,
+      io.opentelemetry.api.metrics.Meter otelMeter) {
     super(clock);
     this.baseTimeUnit = baseTimeUnit;
     this.otelMeter = otelMeter;
 
+    NamingConvention namingConvention =
+        prometheusMode ? PrometheusModeNamingConvention.INSTANCE : NamingConvention.identity;
     this.config()
-        .namingConvention(NamingConvention.identity)
+        .namingConvention(namingConvention)
         .onMeterRemoved(OpenTelemetryMeterRegistry::onMeterRemoved);
   }
 
--- a/instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistryBuilder.java
+++ b/instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistryBuilder.java
@@ -10,6 +10,7 @@ import io.micrometer.core.instrument.MeterRegistry;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.instrumentation.api.config.Config;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 
 /** A builder of {@link OpenTelemetryMeterRegistry}. */
 public final class OpenTelemetryMeterRegistryBuilder {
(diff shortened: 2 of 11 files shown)


instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistryBuilder.java
```diff
@@ -38,12 +49,27 @@ public OpenTelemetryMeterRegistryBuilder setBaseTimeUnit(TimeUnit baseTimeUnit)
...
+  /**
+   * Enables the "Prometheus mode" - this will simulate the behavior of Micrometer's {@code
+   * PrometheusMeterRegistry}. The instruments will be renamed to match Micrometer instrument
+   * naming, and the base time unit will be set to seconds.
+   *
+   * <p>Set this to {@code true} if you are using the Prometheus metrics exporter.
+   */
+  public OpenTelemetryMeterRegistryBuilder setPrometheusMode(boolean prometheusMode) {
+    this.prometheusMode = prometheusMode;
+    return setBaseTimeUnit(TimeUnit.SECONDS);
```
trask, line 61: instead of setting baseTimeUnit here, which makes the final result dependent on the ordering of calling the setters, what do you think of initializing baseTimeUnit to `null` and doing the calculation in `build()`

trask (approved):

instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistry.java
```diff
@@ -53,13 +53,18 @@ public static OpenTelemetryMeterRegistryBuilder builder(OpenTelemetry openTeleme
...
+      Clock clock,
+      TimeUnit baseTimeUnit,
+      boolean prometheusMode,
+      io.opentelemetry.api.metrics.Meter otelMeter) {
     super(clock);
     this.baseTimeUnit = baseTimeUnit;
     this.otelMeter = otelMeter;
 
+    NamingConvention namingConvention =
+        prometheusMode ? PrometheusModeNamingConvention.INSTANCE : NamingConvention.identity;
```
anuraaga, line 65: By the way random note, I noticed our shim defaults to `identity` but micrometer seems to default to `snake_case` IIUC. Any reason ours is different?

https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L142

anuraaga: I guess because we expect the prometheus exporter to do the conversion

instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/PrometheusModeNamingConvention.java
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package io.opentelemetry.instrumentation.micrometer.v1_5;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.NamingConvention;
+import javax.annotation.Nullable;
+
+enum PrometheusModeNamingConvention implements NamingConvention {
```
anuraaga, line 12: AFAICT, the difference between this and the upstream naming convention is we don't append _total to counters, since the prometheus exporter does it. We should probably document that

instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistry.java
mateuszrzeszutek: Yep, that's one of the reasons. The other is that names using dots as the "word separator" seemsto match the spec names the closest (the spec doesn't explicitly mention concatenating with dots as the official strategy, but all instrument names follow this)

instrumentation/micrometer/micrometer-1.5/library/src/main/java/io/opentelemetry/instrumentation/micrometer/v1_5/OpenTelemetryMeterRegistryBuilder.java
mateuszrzeszutek: Good point, will do 👍

PR #5543, opened 2022-03-11
Do not propagate gRPC deadline when propagating OTel context via javaagent.

### LINKED ISSUE #4169: io.grpc.Context getting cancelled
**Describe the bug**
Seeing grpc context timeouts with 1.4.1 and 1.5.3.

**Steps to reproduce**
No specific recipe to share at this point. But this appears to be related to few other issues with Google client APIs.

GRPC context timeout not getting affected by Otel instrumentation.

**What did you see instead?**
```
CANCELLED: io.grpc.Context was cancelled without error.
Stack: io.grpc.Status.asRuntimeException(Status.java:535),
io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:478),
interceptor-for-ClientCall.Listener<RespT>: onClose()
interceptor-for-ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>: onClose()
...(truncated)...
(TracingClientInterceptor.java:139),
(DelayedClientCall.java:463),
(DelayedClientCall.java:427),
(DelayedClientCall.java:460),
(ClientCallImpl.java:553),
(ClientCallImpl.java:68),
(ClientCallImpl.java:212),
(ContextRunnable.java:37),
```

Tried with both 1.4.1 and 1.5.3 with the same result. Issue does not happen without instrumentation.

**Environment**
JDK11

anuraaga:
Fixes #4169 

Some background below but have worked around with this PR

I have been able to add a test that reproduces the problem in the issue, but I'm still not being able to figure out a fix. One problem is that per my reasoning, a cancellation error seems to actually make sense. gRPC server cancels the context when the server request is `onCompleted()` as expected

https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ServerCallImpl.java#L365

For this pattern of early return, where the server request is ended but business logic continues, it seems correct for that business logic to have been cancelled by default (imagine if it wasn't an early return pattern but some async callback sequence continuing after a 10s request deadline, cancellation propagation exists mostly to handle such a case), and explicit opt in to this pattern by calling `Context.fork()` in the business logic seems expected. So I'm not sure why without instrumentation there is no error.

I'm not convinced that there is no bug in the instrumentation, but in my head the instrumented behavior makes sense and the non-instrumented behavior doesn't, so stuck on how to dig deeper. @amitgud-doordash @ryandens do you happen to have any clues on this?

### DIFF
--- a/instrumentation/grpc-1.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_6/GrpcSingletons.java
+++ b/instrumentation/grpc-1.6/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/grpc/v1_6/GrpcSingletons.java
@@ -20,7 +20,7 @@ public final class GrpcSingletons {
 
   public static final ServerInterceptor SERVER_INTERCEPTOR;
 
-  public static final Context.Storage STORAGE = new ContextStorageBridge();
+  public static final Context.Storage STORAGE = new ContextStorageBridge(false);
 
   static {
     boolean experimentalSpanAttributes =
--- a/instrumentation/grpc-1.6/library/src/main/java/io/grpc/override/ContextStorageOverride.java
+++ b/instrumentation/grpc-1.6/library/src/main/java/io/grpc/override/ContextStorageOverride.java
@@ -14,7 +14,7 @@ import io.opentelemetry.instrumentation.grpc.v1_6.internal.ContextStorageBridge;
  */
 public final class ContextStorageOverride extends Context.Storage {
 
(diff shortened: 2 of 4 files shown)


anuraaga:
Came up with a possible conclusion, adding to issue

ryandens:
Thanks for the detailed analysis! I tried out a local build with this change on my reproducer project [ryandens/otel-grpc-context](https://github.com/ryandens/otel-grpc-context) and confirmed the error no longer shows up.

trask (approved):
Thanks @anuraaga @amitgud-doordash @chenkaiyin1201 @ryandens!!

PR #5546, opened 2022-03-11
Change the way how SpanKeys are extracted in instrumentation-api

mateuszrzeszutek:
This PR reverses the logic of how `SpanKey`s are being determined; instead of having `implements` checks somewhere in the `InstrumenterBuilder`, the `AttributesExtractor`s can now implement `SpanKeyProvider` and return all the related keys.

This PR makes it possible to split out all the semantic conventions related subpackages, should we decide to do so. I.e. we could release stable `instrumentation-api` without the `...http`, `...net` packages, and keep those in the alpha `instrumentation-api-semconv` module (that'd get the stable reelase after `opentelemetry-semconv` becomes stable).


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/SpanKeyProvider.java
```diff
@@ -0,0 +1,21 @@
...
+ * Returns {@link SpanKey}s associated with the {@link AttributesExtractor} that implements this
+ * interface.
+ *
+ * <p>This class is internal and is hence not for public use. Its APIs are unstable and can change
+ * at any time.
+ */
+public interface SpanKeyProvider {
+
+  Stream<SpanKey> internalGetSpanKeys();
+}
```
trask, line 21: I like the flip.

two thoughts/questions:

* maybe ok to assume at most one SpanKey and return `@Nullable SpanKey`?
* is it internal just out of caution? seems like a nice hook for anyone defining their own semantic conventions

cc @lmolkova

mateuszrzeszutek: > * is it internal just out of caution? seems like a nice hook for anyone defining their own semantic conventions

It is a nice hook - but since `SpanKey` is internal I wanted to be doubly sure and made this method internal too.

trask: makes sense, we can think about making it public later

wdyt about assuming at most one SpanKey instead of returning `Stream`?

mateuszrzeszutek: Ah, I somehow missed that in you previous comment - yes, that makes sense; I don't think we'll ever need to associate >1 span keys with a single attributes extractor.

trask (approved):

Issue #5291, opened 2022-02-01
Refactor standard AttributesExtractor implementations and split out *Getter interfaces


mateuszrzeszutek:
**Is your feature request related to a problem? Please describe.**
Factor out `*Getter` interfaces out of existing instrumentation-api `AttributeExtractor` implementations. Continuation of #5030

* [x] `CodeAttributesExtractor` #5342
* [x] `DbAttributesExtractor`/`SqlAttributesExtractor` #5354
* [ ] `MessagingAttributesExtractor` #5626
* [x] `RpcAttributesExtractor` #5548

PR #5548, opened 2022-03-11
Split out RpcAttributesGetter

mateuszrzeszutek:
Part of #5291

I thought it would be useful to split the extractors into `RpcClientAttributesExtractor` and `RpcServerAttributesExtractor` because they will have different `SpanKey`s associated to them; there's only one getter interface though, cause both extractors use it to retrieve the same set of attributes.


trask (approved):
> I thought it would be useful to split the extractors into `RpcClientAttributesExtractor` and `RpcServerAttributesExtractor` because they will have different `SpanKey`s associated to them; there's only one getter interface though, cause both extractors use it to retrieve the same set of attributes.

Do you think it makes sense to proactively split the getter as well (similar to http client/server getter) in case rpc client/server attributes diverge in the future?

mateuszrzeszutek:
> Do you think it makes sense to proactively split the getter as well (similar to http client/server getter) in case rpc client/server attributes diverge in the future?

I thought about it too and in the end I couldn't decide whether to do it or not -- I'll try it out and push the change here, let's see if it looks manageable.

anuraaga:
I think unlike HTTP RPC has a stronger tendency to have the same data in server / client so we could probably avoid splitting now.

trask:
> I think unlike HTTP RPC has a stronger tendency to have the same data in server / client so we could probably avoid splitting now.

sounds good, IIRC our current thinking is to not stabilize a semantic attributes extractor until that semantic convention is declared stable, so that gives us some time to see if anything new emerges for rpc

trask:
@mateuszrzeszutek heads up conflicts

PR #5582, opened 2022-03-15
Migrate AkkaActorTest to scala

anuraaga:
In #5571 I had the random idea of writing tests for code that isn't in Java in that language. This is mostly just to understand what it looks like to use our testing libraries in that language - it is not a goal to be idiomatic for that language but it can still be useful to just know what the state of things is and be able to point to it in case the language user asks about it. Given there is significant e.g. scala code outside of the test code, we don't escape the maintenance cost of that language either way.

We don't have to merge this just an idea.


instrumentation/akka/akka-actor-2.5/javaagent/src/test/scala/io/opentelemetry/instrumentation/akkaactor/AkkaActorTest.scala
```diff
@@ -0,0 +1,128 @@
...
+
+    val assertions = (1 to count)
+      .map(_ =>
+        ((trace: TraceAssert) => {
+          trace.hasSpansSatisfyingExactly(
+            ((span: SpanDataAssert) => {
+              span
+                .hasName("parent")
+                .hasAttributes(Attributes.empty())
+              ()
```
anuraaga, line 43: A relatively easy and significant improvement to Scala users might be adding support for `UnaryOperator` assertions. Not something I plan on adding but an interesting takeaway from writing this PR.

mateuszrzeszutek (approved):

mateuszrzeszutek:
I like the idea - as you mentioned, we have to maintain fragments of code in languages other than Java anyway (especially for library instrumentation tests); and we can test other languages' quirks and syntactic sugar this way (which is often not possible in plain Java).

trask (approved):

PR #5585, opened 2022-03-16
Fix possible deadlock

trask:
May want to consider a patch for this, I ran into deadlock locally.

### DIFF
--- a/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
+++ b/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
@@ -432,6 +432,9 @@ public class AgentClassLoader extends URLClassLoader {
   private static class JdkHttpServerClassLoader extends ClassLoader {
 
     static {
+      // this class loader doesn't load any classes, so this is technically unnecessary,
+      // but included for safety, just in case we every change Class.forName() below back to
+      // super.loadClass()
       registerAsParallelCapable();
     }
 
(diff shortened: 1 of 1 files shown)


javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -451,4 +441,41 @@ public long getContentLengthLong() {
       return -1;
     }
   }
+
+  private static class JdkHttpServerClassLoader extends ClassLoader {
+
+    static {
+      registerAsParallelCapable();
```
trask, line 448: this is the fix

anuraaga (approved):

mateuszrzeszutek (approved):

javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/AgentClassLoader.java
```diff
@@ -451,4 +428,44 @@ public long getContentLengthLong() {
...
+      super(null);
+    }
+
+    @Override
+    protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+      // prometheus exporter uses jdk http server, load it from the platform class loader
+      if (name != null && name.startsWith("com.sun.net.httpserver.")) {
+        return platformClassLoader.loadClass(name);
+      }
+      return Class.forName(name, resolve, null);
```
laurit, line 453:
```suggestion
      return Class.forName(name, false, null);
```

laurit (approved):

PR #5586, opened 2022-03-16
Migrate akka http test to scala

anuraaga:


instrumentation/akka/akka-http-10.0/javaagent/src/test/scala/io/opentelemetry/javaagent/instrumentation/akkahttp/AkkaHttpClientInstrumentationTest.scala
```diff
@@ -0,0 +1,102 @@
...
+  val materializer: ActorMaterializer = ActorMaterializer.create(system)
+
+  override protected def buildRequest(
+      method: String,
+      uri: URI,
+      h: util.Map[String, String]
+  ): HttpRequest =
+    HttpRequest(HttpMethods.getForKey(method).get)
+      .withUri(uri.toString)
+      .addHeaders(
```
anuraaga, line 50: Not a big deal but no matter how much I rattled I couldn't get the scaladsl version of this (`withHeaders`) to work, I was expecting `toSeq` instead of `asJava` to work. In case anyone wants to test their scala chops :P

trask (approved):

mateuszrzeszutek (approved):

PR #5594, opened 2022-03-16
Muzzle CI experiment

trask:


.github/workflows/muzzle.yml
```diff
@@ -41,16 +26,8 @@ jobs:
...
-        uses: nick-invision/retry@v2.6.0
+        uses: gradle/gradle-build-action@v2
         with:
-          # timing out has not been a problem, these jobs typically finish in 2-3 minutes
-          timeout_minutes: 15
-          # give maven central some time to hopefully recover (120 seconds has not proven to be enough)
-          retry_wait_seconds: 300
-          max_attempts: 5
-          command: ./gradlew ${{ matrix.module }}:muzzle
+          arguments: ${{ matrix.task }} --max-workers=4
```
anuraaga, line 32: Can we try the split jobs without max-workers?

instrumentation/build.gradle.kts
```diff
@@ -18,19 +20,29 @@ subprojects {
...
-          .map { it.path }
-          .forEach { out.println(it) }
-      }
-    }
+  plugins.withId("io.opentelemetry.instrumentation.muzzle-check") {
+    muzzleTasks.add(subProj.tasks.named("muzzle").get())
   }
 }
+
+// TODO (trask) how to know that muzzleTasks populated above before used below?
```
anuraaga, line 29: If the muzzle tasks are actually being run then maybe it's OK but indeed this pattern doesn't look quite reliable. I don't know if it's important for us to split the list into `1111222233334444`, if `123412341234` is OK I would create the tasks first and then in the `plugins.withId` callback `AtomicInteger.getAndIncrement() % 4` to pick the task to register into.

Alternatively we would need to use `evaluationDependsOn` and `afterEvaluate`, but reducing evaluation order dependencies tends to help understand the build.

trask: > if `123412341234` is OK I would create the tasks first and then in the `plugins.withId` callback `AtomicInteger.getAndIncrement() % muzzleTasks.size()` to pick the task to register into.

ya, I like that, do you know if the subprojects ordering is guaranteed to be the same across multiple invocations (on same git sha)?

.github/workflows/muzzle.yml
trask:
```suggestion
          arguments: ${{ matrix.task }}
```

instrumentation/build.gradle.kts
anuraaga: As far as I know Gradle project evaluation order is deterministic (a BFS) which would make this deterministic too, though I can't say with 100% confidence without trying

.github/workflows/muzzle.yml
trask: no real difference 👍

instrumentation/build.gradle.kts
trask: hm, this worries me: https://www.google.com/search?q=gradle+%22the+task+execution+order+is+not+deterministic%22

anuraaga: Here we're relying on project evaluation order, not task evaluation order. Luckily, project evaluation order is defined by Gradle

https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#N14CB4

anuraaga (approved):

laurit (approved):

mateuszrzeszutek (approved):

PR #5597, opened 2022-03-17
Add vm name to deadlock detector file name

laurit:
otherwise hotspot and openj9 will use the same file name


mateuszrzeszutek (approved):

laurit:
actually the deadlock detector doesn't work on openj9 as it uses `jcmd` so we'll get dumps only from gradle that is running with hotspot not the actual test

trask (approved):

PR #5598, opened 2022-03-17
Run tests with jdk17

laurit:


instrumentation/rmi/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rmi/context/client/RmiClientContextInstrumentation.java
```diff
@@ -58,6 +64,30 @@ public void transform(TypeTransformer transformer) {
...
+            Instrumentation instrumentation = InstrumentationHolder.getInstrumentation();
+            ClassInjector.UsingInstrumentation.redefineModule(
+                instrumentation,
+                module,
+                Collections.emptySet(),
+                Collections.emptyMap(),
+                Collections.singletonMap(
+                    "sun.rmi.transport",
+                    // AgentClassLoader is in unnamed module of the bootstrap class loader which is
+                    // where helper classes are also
```
trask, line 84: 👍

trask (approved):
🎉🎉🎉

can also update https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/supported-libraries.md#jvms-and-operating-systems

trask:
merging b/c I'm doing some work on yml files and don't want to cause conflicts, will send separate PR to update doc

PR #5604, opened 2022-03-17
Increase finatra startup timeout

laurit:
There have been a lot of finatra timeouts lately like https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5595 I believe this happens because during scala conversion timeout was reduced.


trask (approved):
👍

anuraaga (approved):

PR #5605, opened 2022-03-17
Handle failed request in http server concurrency tests

laurit:
Count down the latch even for failed request to ensure that the test doesn't completely hang in case there is a failed request. Probably this cause the test to time out in https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5595


trask (approved):
👍

PR #5608, opened 2022-03-17
Update workflow names

trask:
to help organize the workflows a bit on https://github.com/open-telemetry/opentelemetry-java-instrumentation/actions


.github/workflows/ci.yml
```diff
@@ -9,14 +9,14 @@ on:
 
 jobs:
   build:
-    uses: ./.github/workflows/build.yml
+    uses: ./.github/workflows/reusable-build.yml
```
mateuszrzeszutek, line 12: Would these reusable workflows still be visible on the actions page if we put them into a subdirectory? E.g. `.github/workflows/shared/reusable-build.yml`

mateuszrzeszutek (approved):

.github/workflows/ci.yml
trask: I initially tried this, but github actions won't run them anywhere other than directly inside the workflows directory 😞

Issue #5612, opened 2022-03-17
Tests may not be running on openj9 after #5598


trask:
from @laurit:

> I suspect that openj9 tests don’t actually run with openj9. Probably the part that I copied from Nikitas temurin pr that changed the vm element name, I guess these really need to be hotspot and openj9 for gradle to pick them up.

PR #5613, opened 2022-03-17
Fix running tests with openj9

laurit:
Resolves https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5612


trask (approved):
thx for catching this 👍

PR #5619, opened 2022-03-18
Run check during testing and assemble during building

anuraaga:
The `test.dependsOn` pattern requires non-intuitive boiler plate related to not failing when no tests match. The Gradle docs have an example of using `check.dependsOn` when adding test tasks and we can assume it's idiomatic

https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#declare_an_additional_test_suite

`check` will also run spotless, checkstyle, but I think these static checks are fast enough that we don't need to worry. If we did, we could exclude them later if we find this to be a performance issue.

Also, we run the same tests during build and test right now for one Java version, we may as well not test in the build step since the test step takes care of it.


trask (approved):
nice

Issue #5077, opened 2022-01-11
Convert all logging statements from slf4j to jul


trask:
Discussed in #5053

* we should use jul in the instrumentation-api and library instrumentation to be consistent with SDK and limit dependencies
* we should use jul in javaagent instrumentation because using two different logging libraries feels icky

For now at least, the javaagent will continue to shade all jul usage to [PatchLogger.java](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-bootstrap/src/main/java/io/opentelemetry/javaagent/bootstrap/PatchLogger.java) (which will still use slf4j/slf4jsimple under the hood) for the reasons @laurit mentioned above https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/5053#issuecomment-1008411602, and we will explore redirecting logging to user application logging separately in #5059

PR #5628, opened 2022-03-18
Convert all logging statements from slf4j to jul - part 1

mateuszrzeszutek:
Half of #5077

Not included in this PR:
- tooling, muzzle -- will do that in the next PR
- testing modules -- as far as I understand, we'll probably want to keep using slf4j here; in many tests, we're using testcontainers' built-in facilities to redirect container logs to slf4j logger

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetrics.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetrics.java
@@ -67,13 +67,10 @@ public final class HttpClientMetrics implements RequestListener {
   public void end(Context context, Attributes endAttributes, long endNanos) {
     State state = context.get(HTTP_CLIENT_REQUEST_METRICS_STATE);
     if (state == null) {
-      if (logger.isLoggable(Level.FINE)) {
-        logger.log(
-            Level.FINE,
-            "No state present when ending context "
-                + context
-                + ". Cannot record HTTP request metrics.");
-      }
+      logger.log(
+          Level.FINE,
+          "No state present when ending context {0}. Cannot record HTTP request metrics.",
+          context);
       return;
     }
     duration.record(
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetrics.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetrics.java
@@ -17,8 +17,8 @@ import io.opentelemetry.instrumentation.api.annotations.UnstableApi;
 import io.opentelemetry.instrumentation.api.instrumenter.RequestListener;
 import io.opentelemetry.instrumentation.api.instrumenter.RequestMetrics;
 import java.util.concurrent.TimeUnit;
-import org.slf4j.Logger;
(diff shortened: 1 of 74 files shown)


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpClientMetrics.java
```diff
@@ -67,8 +67,13 @@ public Context start(Context context, Attributes startAttributes, long startNano
...
     if (state == null) {
-      logger.debug(
-          "No state present when ending context {}. Cannot record HTTP request metrics.", context);
+      if (logger.isLoggable(Level.FINE)) {
+        logger.log(
+            Level.FINE,
+            "No state present when ending context "
+                + context
+                + ". Cannot record HTTP request metrics.");
+      }
```
trask, line 76: I think can parameterize using `{0}` and avoid many of the conditionals

mateuszrzeszutek: AFAIR slf4j (which we're _actually_ using underneath) only recognizes `{}` placeholders - it will treat `{0}` (`{anything}`) literally.

mateuszrzeszutek: Oh, looks like `PatchLogger` already handles that -- my bad, I'll change that PR to use placeholders.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/http/HttpServerMetrics.java
```diff
@@ -79,8 +79,12 @@ public Context start(Context context, Attributes startAttributes, long startNano
...
     State state = context.get(HTTP_SERVER_REQUEST_METRICS_STATE);
     if (state == null) {
-      logger.debug(
-          "No state present when ending context {}. Cannot reset HTTP request metrics.", context);
+      if (logger.isLoggable(Level.FINE)) {
+        logger.log(
+            Level.FINE,
+            "No state present when ending context {0}. Cannot record HTTP request metrics.",
+            context);
+      }
```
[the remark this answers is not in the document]
mateuszrzeszutek: Oops, totally missed that -- thanks!

instrumentation/internal/internal-class-loader/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/internal/classloader/BootDelegationInstrumentation.java
```diff
@@ -89,8 +90,8 @@ public void transform(TypeTransformer transformer) {
         //noinspection unchecked
         return (List<String>) methodHandle.invokeExact();
       } catch (Throwable e) {
-        LoggerFactory.getLogger(Holder.class)
-            .warn("Unable to load bootstrap package prefixes from the bootstrap CL", e);
+        Logger.getLogger(Holder.class.getName())
+            .log(WARNING, "Unable to load bootstrap package prefixes from the bootstrap CL", e);
```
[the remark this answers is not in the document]
mateuszrzeszutek: 👍 
I didn't really have an opinion on static import vs no import, so I just went with the default - let's try it out.

PR #5648, opened 2022-03-21
Fix http concurrency test with large responses

laurit:
While trying to reproduce https://ge.opentelemetry.io/s/cii3nazylwp3y/tests/:instrumentation:google-http-client-1.19:javaagent:test/io.opentelemetry.javaagent.instrumentation.googlehttpclient.GoogleHttpClientAsyncTest/highConcurrency() I made http concurrency tests use a large response which caused some test to fail. Hopefully these changes will make it easier in case a similar test needs to be carried out in the future.

### DIFF
--- a/instrumentation/netty/netty-3.8/javaagent/src/test/groovy/Netty38ClientTest.groovy
+++ b/instrumentation/netty/netty-3.8/javaagent/src/test/groovy/Netty38ClientTest.groovy
@@ -72,7 +72,7 @@ class Netty38ClientTest extends HttpClientTest<Request> implements AgentTestTrai
 
   @Override
   void sendRequestWithCallback(Request request, String method, URI uri, Map<String, String> headers, AbstractHttpClientTest.RequestResult requestResult) {
-    // TODO: context propagation into callbacks is not implemented
+    // TODO: context is not automatically propagated into callbacks
     Context context = Context.current()
     // TODO(anuraaga): Do we also need to test ListenableFuture callback?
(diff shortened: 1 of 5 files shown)


instrumentation/netty/netty-3.8/javaagent/src/test/groovy/Netty38ClientTest.groovy
```diff
@@ -70,17 +72,23 @@ class Netty38ClientTest extends HttpClientTest<Request> implements AgentTestTrai
 
   @Override
   void sendRequestWithCallback(Request request, String method, URI uri, Map<String, String> headers, AbstractHttpClientTest.RequestResult requestResult) {
+    // TODO: context propagation into callbacks is not implemented
```
laurit, line 75: I guess this would need to be corrected with `async-http-client` instrumentation, but we don't instrument `async-http-client` 1.8.

trask:
```suggestion
    // TODO: context is not automatically propagated into callbacks
```

trask (approved):
> I made http concurrency tests use a large response which caused some test to fail. Hopefully these changes will make it easier in case a similar test needs to be carried out in the future.

this sounds like a great test, I'm in favor of adding it

anuraaga (approved):

mateuszrzeszutek (approved):

PR #5655, opened 2022-03-21
Add missing InstrumenterBuilder.addRequestListener()

### LINKED ISSUE #5652: requestListeners In InstrumenterBuilder
How to add requestListener to InstrumenterBuilder,I can not find it in version 1.11.x

trask:
Resolves #5652

### DIFF
--- a/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
+++ b/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
@@ -232,10 +232,11 @@ public class Instrumenter<REQUEST, RESPONSE> {
 
     if (!requestListeners.isEmpty() || !requestMetricListeners.isEmpty()) {
       long endNanos = getNanos(endTime);
-      for (RequestListener requestListener : requestListeners) {
+      // TODO (trask) call end in the reverse order that start was called?
+      for (RequestListener requestListener : requestMetricListeners) {
         requestListener.end(context, attributes, endNanos);
       }
-      for (RequestListener requestListener : requestMetricListeners) {
+      for (RequestListener requestListener : requestListeners) {
         requestListener.end(context, attributes, endNanos);
       }
     }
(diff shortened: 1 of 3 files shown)


laurit (approved):

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -232,10 +232,11 @@ public void end(
 
     if (!requestListeners.isEmpty() || !requestMetricListeners.isEmpty()) {
       long endNanos = getNanos(endTime);
-      for (RequestListener requestListener : requestListeners) {
+      // TODO (trask) call end in the reverse order that start was called?
```
mateuszrzeszutek, line 235: We probably don't have to do that - unlike `start()`, the `end()` method cannot modify the context; also, we're passing the timestamp as an argument, so the exact time of invocation should not matter too.

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -232,10 +232,11 @@ public void end(
...
+      for (RequestListener requestListener : requestMetricListeners) {
         requestListener.end(context, attributes, endNanos);
       }
-      for (RequestListener requestListener : requestMetricListeners) {
+      for (RequestListener requestListener : requestListeners) {
```
mateuszrzeszutek, line 239: Nit: do we actually need two lists of listeners? Can't we merge `requestMetricListeners` into `requestListeners`?

mateuszrzeszutek (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
trask: ya, I just always think of start/end pairs telescoping :)

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
trask: I'll send that in a separate PR b/c it moves the requestListeners to after the span is started (which I think is ok, but I'd like to get eyes specifically on that change)

PR #5659, opened 2022-03-21
Update required status check

trask:
this will require updating the required status check for the repo

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -71,15 +71,16 @@ jobs:
           sh ./install-misspell.sh
           ./bin/misspell -error ./**/*
 
-  accept-pr:
+  required-status-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -71,15 +71,16 @@ jobs:
           sh ./install-misspell.sh
           ./bin/misspell -error ./**/*
 
-  accept-pr:
+  required-status-check:
```
anuraaga, line 74: FWIW this PR doesn't seem to require a change to this name, though also don't have a strong preference for either, am used to the current name now though.

anuraaga (approved):

.github/workflows/pr.yml
trask: ya, this is a good point, I changed it back for now at least

mateuszrzeszutek (approved):

PR #5667, opened 2022-03-22
Consolidate request listener lists

trask:
follow-up from @mateuszrzeszutek's https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/5655#discussion_r831930029

This PR moves context customizer execution and request listener start execution from right before the span is started until right after the span is started.

This makes the newly create span available in the Context, which the metric listeners need for generating exemplars, and seems like could be useful for other listeners.

I'm thinking it doesn't matter much one way or the other for context customizers, so seems to make some sense to me to keep them with the listeners (after span start), so we can think of all context updates happening after span start.


mateuszrzeszutek (approved):
Thanks! 👍

PR #5668, opened 2022-03-23
Migrate Guava tests to Java

anuraaga:
Removes AbstractPromiseTest since guava was the only implementation. If we find a reason to abstract out tests for futures later, we can restore something like it then


mateuszrzeszutek (approved):

instrumentation/guava-10.0/javaagent/src/test/java/io/opentelemetry/javaagent/instrumentation/guava/ListenableFutureTest.java
```diff
@@ -0,0 +1,198 @@
...
+          }
+        },
+        executor);
+    testing.runWithSpan("other", () -> future.set(value));
+
+    assertThat(Futures.getUnchecked(future)).isEqualTo(value);
+
+    // TODO: There appears to be a context propagation bug. There is no logical reason for other to
+    // be the parent of
+    // callback in this test but not in callWithParent.
```
laurit, line 188:
```suggestion
    // be the parent of callback in this test but not in callWithParent.
```

PR #5674, opened 2022-03-23
Convert all logging statements from slf4j to jul

### LINKED ISSUE #5077: Convert all logging statements from slf4j to jul

mateuszrzeszutek:
Closes #5077

### DIFF
--- a/javaagent-tooling/build.gradle.kts
+++ b/javaagent-tooling/build.gradle.kts
@@ -50,7 +50,6 @@ dependencies {
   implementation("io.opentelemetry:opentelemetry-sdk-extension-jaeger-remote-sampler")
 
   api("net.bytebuddy:byte-buddy-dep")
-  implementation("org.slf4j:slf4j-api")
 
   annotationProcessor("com.google.auto.service:auto-service")
   compileOnly("com.google.auto.service:auto-service-annotations")
--- a/javaagent-tooling/build.gradle.kts
+++ b/javaagent-tooling/build.gradle.kts
@@ -63,6 +63,21 @@ dependencies {
   testImplementation("com.google.guava:guava")
 }
 
+testing {
+  suites {
+    val testExceptionHandler by registering(JvmTestSuite::class) {
+      dependencies {
+        implementation(project(":javaagent-bootstrap"))
+        implementation(project(":javaagent-tooling"))
+        implementation("net.bytebuddy:byte-buddy-dep")
+
+        // Used by byte-buddy but not brought in as a transitive dependency.
+        compileOnly("com.google.code.findbugs:annotations")
+      }
+    }
+  }
+}
+
 // Here we only include autoconfigure but don"t include OTLP exporters to ensure they are only in
 // the full distribution. We need to override the default exporter setting of OTLP as a result.
 tasks {
@@ -76,6 +91,10 @@ tasks {
     options.errorprone {
       isEnabled.set(false)
     }
+  }
+
+  check {
+    dependsOn(testing.suites)
   }
 }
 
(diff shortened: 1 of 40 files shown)


javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/bytebuddy/ExceptionHandlers.java
```diff
@@ -41,8 +36,7 @@ public boolean isValid() {
 
               // writes the following bytecode:
               // try {
-              //   org.slf4j.LoggerFactory.getLogger((Class)ExceptionLogger.class)
-              //     .debug("exception in instrumentation", t);
+              //   ExceptionLogger.logSuppressedError("exception in instrumentation", t);
```
trask, line 39: 👍

trask (approved):
I found reviewing a bit confusing in places where TransformSafeLogger is used since that api is now different from normal loggers, maybe worth sync'ing up that class.

javaagent-tooling/javaagent-tooling-java9/src/main/java/io/opentelemetry/javaagent/tooling/ExposeAgentBootstrapListener.java
```diff
@@ -52,7 +54,12 @@ private void exposeModule(JavaModule fromModule, JavaModule targetModule) {
     if (fromModule != JavaModule.UNSUPPORTED
         && fromModule.isNamed()
         && !fromModule.canRead(targetModule)) {
-      logger.debug("Adding module read from {} to {}", fromModule, targetModule);
+
+      logger.log(
+          Level.FINE,
+          "Adding module read from {0} to {1}",
+          new Object[] {fromModule, targetModule});
```
anuraaga, line 61: For this pattern we should probably check loggable to avoid array allocation when disabled

mateuszrzeszutek: Good point, done.

anuraaga (approved):

javaagent-tooling/build.gradle.kts
```diff
@@ -64,6 +63,21 @@ dependencies {
   testImplementation("com.google.guava:guava")
 }
 
+testing {
+  suites {
+    val testExceptionHandler by registering(JvmTestSuite::class) {
```
trask, line 68: does this target specific tests?

mateuszrzeszutek: Yes, just the `ExceptionHandlerTest` - I moved it to a separate source directory, along with all its dependencies. I found that when I run `./gradlew :javaagent-tooling:check` locally it was randomly failing on that test; probably because it interfered with other bytecode modifying tests in the `test` source root.

trask: oh yes, I missed that this is tied to directory name, thx!

PR #5679, opened 2022-03-24
Migrate reactor testing to Java

anuraaga:


instrumentation/reactor/reactor-3.1/testing/src/main/groovy/io/opentelemetry/instrumentation/reactor/AbstractReactorCoreTest.groovy
```diff
@@ -1,521 +0,0 @@
...
-      }
-    }
-
-    assert remainingIterations.isEmpty()
-  }
-
-  def runWithSpan(def publisherSupplier) {
-    runWithSpan("trace-parent") {
-      def tracer = GlobalOpenTelemetry.getTracer("test")
-      def span = tracer.spanBuilder("publisher-parent").startSpan()
```
anuraaga, line 466: I didn't include the publisher-parent span in the migrated tests since I couldn't reason out any special tested behavior by it, it's just a test-only span. But can add it back if it seems important

trask (approved):

PR #5692, opened 2022-03-25
Library instrumentation should read its version from a file

### LINKED ISSUE #2761: Library instrumentation should read its version from a file, not from the jar manifest

mateuszrzeszutek:
Resolves #2761. As discussed on the last SIG meeting, each instrumentation gets its separate version file.


instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/EmbeddedInstrumentationVersions.java
```diff
@@ -0,0 +1,75 @@
...
+  private static final Map<String, String> versions = loadVersions();
+
+  private static Map<String, String> loadVersions() {
+    Map<String, String> versions = new HashMap<>();
+
+    // preload this once
+    try {
+      List<URL> urls =
+          list(
+              new ResourceLoaderProxy()
```
laurit, line 38: Do I understand correctly that for library instrumentation this class needs to be in the same class loader as the library instrumentation so that it can find the versions file?

mateuszrzeszutek: Yep - which I think _should_ normally happen: library instrumentation triggers `Instrumenter` load, `Instrumenter` triggers this.

laurit: What if you are using a library instrumentation along with the agent. Then this class will be in boot loader and won't find the library versions.

mateuszrzeszutek: The shaded version of this class (provided by the agent) will be in the bootloader; the unshaded version (provided by the application itself) should still see its own versions file (+ all other versions files, since they get resource-injected into all classloaders).

laurit: Completely forgot that this would also get shaded. I still think it is possible to break it by placing instrumentation-api in parent class loader and library instrumentation in child class loader.

mateuszrzeszutek: Yes, that scenario would break this mechanism. I wonder how common it is though, you'd have to deliberately separate the instrumentation-api from the instrumentation.

anuraaga:
Since we expect either bundled javaagent or use of the BOM for aligned versions, can we just have one version instead of per-instrumentation? Can just be tied straight into instrumentation-api

instrumentation/jedis/jedis-4.0/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/jedis/v4_0/JedisSingletons.java
```diff
@@ -14,7 +14,7 @@
 import io.opentelemetry.instrumentation.api.instrumenter.net.NetClientAttributesExtractor;
 
 public final class JedisSingletons {
-  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.jedis-3.0";
+  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.jedis-4.0";
```
trask, line 17: 💯

instrumentation/ratpack/ratpack-1.7/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTelemetryBuilder.java
```diff
@@ -29,7 +29,7 @@
 /** A builder for {@link RatpackTelemetry}. */
 public final class RatpackTelemetryBuilder {
 
-  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.ratpack-1.4";
+  private static final String INSTRUMENTATION_NAME = "io.opentelemetry.ratpack-1.7";
```
trask, line 32: Is there a test now that fails when these are wrong?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/instrumenter/Instrumenter.java
```diff
@@ -80,6 +83,7 @@
    * different library versions it's easy to find out which instrumentations produced the telemetry
    * data.
    */
+  // TODO: add a setInstrumentationVersion method to the builder instead
```
trask, line 86: 👍

trask (approved):

conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
```diff
@@ -119,3 +121,35 @@ if (testLatestDeps) {
     }
   }
 }
+
+val otelInstrumentation = extensions.create<OtelInstrumentationExtension>("otelInstrumentation")
+
+tasks {
+  val generateInstrumentationVersionFile by registering {
+    val propertiesDir = File(project.buildDir, "generated/instrumentationVersion")
+    outputs.dir(propertiesDir)
```
anuraaga, line 130: We need to add version as `inputs`. Also consider registering the subdirectory to outputs instead of this top level one, which causes directories to be automatically created. So maybe start by copying from SDK repo :)

https://github.com/open-telemetry/opentelemetry-java/blob/main/buildSrc/src/main/kotlin/otel.java-conventions.gradle.kts#L143

conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
```diff
@@ -119,3 +121,35 @@ if (testLatestDeps) {
...
+      val version = otelInstrumentation.version.getOrElse(project.version as String)
+      val metaInfDir = File(propertiesDir, "META-INF/io/opentelemetry/instrumentation/")
+      metaInfDir.mkdirs()
+      File(metaInfDir, "$name.properties")
+        .writeText("version=$version")
+    }
+  }
+}
+
+fun Project.computeInstrumentationName(): String {
```
anuraaga, line 143: I've noticed people love adding extension methods when writing Kotlin, I don't know if I get it :P But in this case especially, a normal `fun` seems like it will be the exact same code without extension. Notice that you didn't use `this.name`

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/EmbeddedInstrumentationProperties.java
```diff
@@ -0,0 +1,72 @@
...
+      Properties parsed = new Properties();
+      parsed.load(in);
+      return parsed.getProperty("version");
+    } catch (IOException e) {
+      logger.log(FINE, "Failed to load embedded instrumentation properties file " + path, e);
+      return null;
+    }
+  }
+
+  private static final class EmbeddedPropertiesLoader extends ClassLoader {
```
anuraaga, line 55: Can we just provide a `set(ClassLoader)` method called from the agent instead of using instrumentation?

instrumentation/ratpack/ratpack-1.7/library/src/main/java/io/opentelemetry/instrumentation/ratpack/RatpackTelemetryBuilder.java
mateuszrzeszutek: No, there isn't, unfortunately -- I went over all usages of `Instrumenter#builder()` and fixed them all manually. Currently I don't have any good idea about how to verify this in a smart way that does not require adding assertions to every test.

conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
mateuszrzeszutek: If you have a hammer... 😄 
Yeah, this definitely works better as a normal function, I suppose I just suffered from Kotlin withdrawal symptoms 🙈

conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
mateuszrzeszutek: Nice, thanks!

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/EmbeddedInstrumentationProperties.java
mateuszrzeszutek: That's much simpler, thanks!

conventions/src/main/kotlin/io.opentelemetry.instrumentation.base.gradle.kts
```diff
@@ -119,3 +119,32 @@ if (testLatestDeps) {
     }
   }
 }
+
+tasks {
+  val generateInstrumentationVersionFile by registering {
+    val name = computeInstrumentationName()
```
anuraaga, line 125: I think we need to add this name to inputs too since it's not in the output dir name.

anuraaga (approved):

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/EmbeddedInstrumentationProperties.java
```diff
@@ -0,0 +1,83 @@
...
```
anuraaga, line 66: Do we still need this class instead of defaulting to `EmbeddedInstrumentationProperties.class.getClassLoader()`?

mateuszrzeszutek: It's probably a bit safer this way -- in case you'll end up with instrumentation-api in the bootstrap CL, without the agent (which is probably not a very likely scenario, but it _can_ happen I guess)

trask: to make this reason more explicit, how about only using EmbeddedPropertiesLoader on null class loader?

instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/internal/EmbeddedInstrumentationProperties.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    // Do not remove this method -- it is instrumented by the javaagent so that the
+    // instrumentation-api (which resides in the bootstrap CL) is able to load instrumentation
+    // properties from the agent class space
+    @SuppressWarnings("RedundantOverride")
+    @Override
+    public URL getResource(String name) {
+      return super.getResource(name);
+    }
```
trask, line 79:
```suggestion
```

mateuszrzeszutek: 🙈

trask:
❤️

